PR #2291, opened 2025-11-18
add FIM dataset support

dimapihtar:
# What does this PR do ?

Adds FIM dataset support.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/.github/workflows/_build_test_publish_wheel.yml
+++ b/.github/workflows/_build_test_publish_wheel.yml
@@ -129,6 +129,7 @@ jobs:
   publish-wheels:
     needs: [build-and-test-wheels]
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
     environment: ${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/r')) && 'main' || 'public' }}
     strategy:
       fail-fast: false
--- a/.github/workflows/build-test-publish-wheel.yml
+++ b/.github/workflows/build-test-publish-wheel.yml
@@ -18,8 +18,8 @@ on:
   push:
     branches:
       - main
-      - "pull-request/[0-9]+"
-      - "deploy-release/*"
+      - 'pull-request/[0-9]+'
+      - 'deploy-release/*'
   merge_group:
     types: [checks_requested]
 
@@ -63,7 +63,7 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
           GITHUB_RUN_ID: ${{ github.run_id }}
-          SKIPPING_IS_ALLOWED: ${{ needs.pre-flight.outputs.docs_only == 'true' || needs.pre-flight.outputs.is_deployment_workflow == 'true' || needs.pre-flight.outputs.is_merge_group == 'true' || needs.pre-flight.outputs.is_ci_workload == 'true' }}
+          SKIPPING_IS_ALLOWED: ${{ needs.pre-flight.outputs.docs_only == 'true' || needs.pre-flight.outputs.is_deployment_workflow == 'true' || needs.pre-flight.outputs.is_merge_group == 'true' || needs.pre-flight.outputs.is_ci_workload == 'true' || github.ref != 'refs/heads/main' }}
         run: |
           FAILED_JOBS=$(gh run view $GITHUB_RUN_ID --json jobs --jq '[.jobs[] | select(.status == "completed" and .conclusion != "success")] | length') || echo 0
 
--- a/.github/workflows/check_api_backwards_compatibility_workflow.yml
+++ b/.github/workflows/check_api_backwards_compatibility_workflow.yml
@@ -4,10 +4,6 @@ on:
   push:
     branches:
       - "pull-request/[0-9]+"
-    paths:
-      - 'megatron/core/**/*.py'
-      - '!megatron/core/tests/**'
-      - '!megatron/legacy/**'
 
   # Allow manual trigger
   workflow_dispatch:
@@ -17,7 +13,49 @@ on:
         required: true
 
 jobs:
+  pre-flight:
+    name: Pre-flight check
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.check_files.outputs.should_skip }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      
+      - name: Check if relevant files changed
+        id: check_files
+        run: |
+          # For manual triggers, never skip
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
+            echo "should_skip=false" >> $GITHUB_OUTPUT
+            echo "Manual trigger - will run compatibility check"
+            exit 0
+          fi
+          
+          # Check if any relevant files changed
+          # Use merge-base to find common ancestor with main
+          # This ensures we only detect changes actually made in this PR branch,
+          # not changes that happened in main after the branch was created
+          BASE_SHA=$(git merge-base origin/main HEAD)
+          echo "Comparing against merge-base: $BASE_SHA"
+          
+          # Check for changes in megatron/core Python files (excluding tests and legacy)
+          CHANGED_FILES=$(git diff --name-only "$BASE_SHA" HEAD -- 'megatron/core/**/*.py' ':!megatron/core/tests/**' ':!megatron/legacy/**' || echo "")
+          
+          if [ -z "$CHANGED_FILES" ]; then
+            echo "should_skip=true" >> $GITHUB_OUTPUT
+            echo "No relevant megatron/core files changed - will skip compatibility check"
+          else
+            echo "should_skip=false" >> $GITHUB_OUTPUT
+            echo "Relevant files changed:"
+            echo "$CHANGED_FILES"
+          fi
+
   check-compatibility:
+    needs: [pre-flight]
+    if: needs.pre-flight.outputs.should_skip != 'true'
     name: Check API Backward Compatibility
     runs-on: ubuntu-latest
     
@@ -168,4 +206,34 @@ jobs:
         if: steps.compat_check.outcome == 'success'
         run: |
           echo "::notice::✅ No breaking API changes detected!"
+
+  api-backward-compatibility-summary:
+    needs: [pre-flight, check-compatibility]
+    runs-on: ubuntu-latest
+    name: API Backward Compatibility Check Summary
+    if: always() && !cancelled()
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Validate workflow result
+        shell: bash -x -e -u -o pipefail {0}
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SKIPPING_IS_ALLOWED: ${{ needs.pre-flight.outputs.should_skip == 'true' }}
+        run: |
+          FAILED_JOBS=$(gh run view $GITHUB_RUN_ID --json jobs --jq '[.jobs[] | select(.status == "completed" and .conclusion != "success" and .name != "API Backward Compatibility Check Summary")] | length') || echo 0
+
+          if [ "${FAILED_JOBS:-0}" -eq 0 ] || [ "$SKIPPING_IS_ALLOWED" == "true" ]; then
+              if [ "$SKIPPING_IS_ALLOWED" == "true" ]; then
+                  echo "✅ Compatibility check was skipped (no relevant files changed)"
+              else
+                  echo "✅ All checks passed successfully"
+              fi
+              exit 0
+          else
+              echo "❌ Found $FAILED_JOBS failed job(s)"
+              gh run view $GITHUB_RUN_ID --json jobs --jq '.jobs[] | select(.status == "completed" and .conclusion != "success" and .name != "API Backward Compatibility Check Summary") | .name'
+              exit 1
+          fi
 
--- a/megatron/core/inference/contexts/dynamic_context.py
+++ b/megatron/core/inference/contexts/dynamic_context.py
@@ -337,7 +337,7 @@ class DynamicInferenceContext(BaseInferenceContext):
             self.kv_reduced_dim = kv_lora_rank + qk_pos_emb_head_dim
             self.block_size_bytes = (
                 dtype_size_bytes
-                * num_attention_layers
+                * self.num_attention_layers
                 * self.block_size_tokens
                 * self.kv_reduced_dim
             )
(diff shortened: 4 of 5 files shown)


megatron/training/datasets/fim_dataset.py
```diff
@@ -0,0 +1,292 @@
...
+            sample = np.concatenate([sample, np.full((-1 * diff), self.pad_tok_id)])
+
+        assert sample.shape[0] == sample_len
+
+        return (
+            np.array(sample, dtype=np.int64),
+            np.array(document_ids, dtype=np.int64),
+        )
+
+    def _fim_permute_sequence(self, sequence, np_rng, rate):
```
suiyoubi: why do we need to pass `np_rng` here ? could we directly infer from `self.np_rng` ?

megatron/training/datasets/fim_dataset.py
```diff
@@ -0,0 +1,292 @@
...
+            self.config.tokenizer,
+            truncate_or_pad=False,
+            suffix_tok_id=self.suffix_tok_id,
+            prefix_tok_id=self.prefix_tok_id,
+            middle_tok_id=self.middle_tok_id,
+            pad_tok_id=self.pad_tok_id,
+            no_fim_prefix=self.no_fim_prefix,
+        )
+
+    def _fim_split_and_permute_sequence(self, sequence, np_rng):
```
suiyoubi: why do we need to pass np_rng here ? could we directly infer from self.np_rng ?

megatron/training/datasets/fim_dataset.py
```diff
@@ -0,0 +1,292 @@
...
+        # Permute the segment after the last split token
+        permuted = self._fim_permute_sequence(sequence[curr_start_position:], np_rng, self.fragment_fim_rate)
+        new_samples.append(permuted)
+
+        return np.concatenate(new_samples)
+
+    def _permute(
+        self,
+        sample,
+        np_rng,
```
suiyoubi: why do we need to pass np_rng here ? could we directly infer from self.np_rng ?

tests/unit_tests/data/test_fim_dataset.py
```diff
@@ -0,0 +1,58 @@
...
+import torch
+
+from megatron.core.datasets.blended_megatron_dataset_builder import BlendedMegatronDatasetBuilder
+from megatron.core.datasets.utils import compile_helpers, get_blend_from_list
+from megatron.core.tokenizers import MegatronTokenizer
+from megatron.training.datasets.fim_dataset import GPTFIMDataset, GPTFIMDatasetConfig
+from tests.unit_tests.test_utilities import Utils
+
+
+def test_fim_gpt_dataset():
```
suiyoubi: I don't think this unit test is meaningful: For FIM we should test that we can detect that split: we should use the data in a way that mimic the FIM split: (e.g., set `rate=1.` and treat some tokens as "fake fim tokens" to force the FIM split )

megatron/training/datasets/README.md
```diff
@@ -0,0 +1,34 @@
+# Data Pipeline
+
+## FIM dataset
```
suiyoubi, line 3: Maybe not necessarily belong to this PR, but we should at least note in the readme that in order to use FIM training, your pretrain dataset needs to be preprocessed with the special tokens. We might need to add support for data preprocessing script as well ( could be separate PR)

megatron/training/datasets/fim_dataset.py
```diff
@@ -0,0 +1,292 @@
...
+        indexed_dataset: IndexedDataset,
+        dataset_path: str,
+        indexed_indices: np.ndarray,
+        num_samples: int,
+        index_split: Split,
+        config: GPTFIMDatasetConfig,
+    ) -> None:
+        super().__init__(indexed_dataset, dataset_path, indexed_indices, num_samples, index_split, config)
+
+        self.indexed_dataset = indexed_dataset
```
asolergi-nv: `self.dataset` is defined already

https://github.com/NVIDIA/Megatron-LM/blob/00884a8765f2e67f880d962eaf3f231ed928fee2/megatron/core/datasets/megatron_dataset.py#L50

megatron/core/tokenizers/text/libraries/null_tokenizer.py
```diff
@@ -25,6 +25,14 @@ def ids_to_text(self, ids):
         text = [str(x) for x in ids]
         return ' '.join(text)
 
+    def tokens_to_ids(self, tokens):
+        """Converts tokens to ids."""
+        return [int(x) for x in tokens]
```
asolergi-nv, line 30: This will raise a issue when trying to convert `str` to `int`. 

If we just want to test this feature with the `NullTokenizer` works since the `extra_tokens` you defined are already numbers. I would suggest to properly support this method & also test other tokenizers

megatron/training/datasets/fim_dataset.py
```diff
@@ -0,0 +1,292 @@
...
+                # Add the sample part
+                offset = 0 if i > doc_index_beg else doc_index_beg_offset
+                length = None if i < doc_index_end else doc_index_end_offset + 1
+                sample_parts.append(self.indexed_dataset.get(self.document_index[i], offset=offset, length=length))
+
+        sample = np.concatenate(sample_parts)
+
+        sample_len = sample.shape[0]
+        segment_breaks = np.argwhere(sample == self.eod_tok_id)
+        np_rng = self.np_rng
```
asolergi-nv:
```suggestion
```

megatron/training/datasets/fim_dataset.py
```diff
@@ -0,0 +1,292 @@
...
+        if doc_index_beg == doc_index_end:
+            # Add the document id
+            document_ids.append(self.document_index[doc_index_beg])
+
+            # Add the entire sample
+            sample_parts.append(
+                self.indexed_dataset.get(
+                    self.document_index[doc_index_beg],
+                    offset=doc_index_beg_offset,
+                    length=doc_index_end_offset - doc_index_beg_offset + 1,
```
asolergi-nv, line 133:
```suggestion
                    length=doc_index_end_offset - doc_index_beg_offset + self.config.add_extra_token_to_sequence,
```

megatron/training/datasets/fim_dataset.py
```diff
@@ -0,0 +1,292 @@
...
+
+        # Sample spans multiple documents
+        else:
+            for i in range(doc_index_beg, doc_index_end + 1):
+                # Add the document id
+                document_ids.append(self.document_index[i])
+
+                # Add the sample part
+                offset = 0 if i > doc_index_beg else doc_index_beg_offset
+                length = None if i < doc_index_end else doc_index_end_offset + 1
```
asolergi-nv, line 145:
```suggestion
                length = None if i < doc_index_end else doc_index_end_offset + self.config.add_extra_token_to_sequence
```

megatron/core/tokenizers/text/libraries/null_tokenizer.py
dimapihtar: > This will raise a issue when trying to convert `str` to `int`.
> 
> If we just want to test this feature with the `NullTokenizer` works since the `extra_tokens` you defined are already numbers. I would suggest to properly support this method & also test other tokenizers

`Null Tokenizer` doesn't support strings. That's a mock tokenizer for tests/debugging which takes only strings of integers.

megatron/training/datasets/fim_dataset.py
dimapihtar: I don't think we want to introduce new config value here since this extra token is expected to be added by default.

megatron/training/datasets/fim_dataset.py
dimapihtar: I don't think we want to introduce new config value here since this extra token is expected to be added by default.

megatron/training/datasets/README.md
dimapihtar: agree, let's add it later i na separate PR.

megatron/training/datasets/fim_dataset.py
```diff
@@ -0,0 +1,288 @@
...
+            # we should rather truncate at the context-level
+            if truncate_or_pad:
+                # need to make same length as the input. Take the 3 sentinel tokens into account
+                new_length = suffix.shape[0] + prefix.shape[0] + middle.shape[0] + 3
+                diff = new_length - sample.shape[0]
+                if diff > 0:  # too long
+                    if (
+                        suffix.shape[0] <= diff
+                    ):  # if there's no space to truncate the suffix: stop and report it. atm i should have stopped this from happening
+                        return sample, self.np_rng
```
suiyoubi: this return does not seem right to me, should we just do `return sample`?

megatron/training/datasets/fim_dataset.py
```diff
@@ -0,0 +1,288 @@
...
+        """
+        If self.fim_split_sample is not None, split the sequence.
+        Then apply FIM on the fragments, or the whole sequence if self.fim_split_sample is None.
+        """
+        if self.fim_split_sample is None:
+            return self._fim_permute_sequence(sequence, self.fim_rate)
+        # fim_split_sample is set: split the sample on this token and permute each fragment separately.
+        # Typically, if each sample is a repository, then we split again on the file level.
+        # Each fragment is a file, and we permute the files.
+        fragment_breaks = np.argwhere(sequence == self.fim_split_sample)
```
suiyoubi, line 208: Let's `assert` to make sure that `self.fim_split_sample` is not a vector (I remember we had some issue with starcoder for this) . We can do the assertion in the init func, something like this: 
```
ids = tokenizer.tokens_to_ids(split_sample)
assert isinstance(ids, int) or len(ids) == 1
self.fim_split_sample = ids if isinstance(ids, int) else ids[0]
```

megatron/training/datasets/README.md
suiyoubi: yeah something like: 

```
fim_tokens["prefix"], fim_tokens["middle"], fim_tokens["suffix"], fim_tokens["pad"], fim_tokens["eod"]

These tokens must exist in the tokenizer vocab.

The dataset must be pre-tokenized with those tokens.
```

pretrain_gpt.py
```diff
@@ -172,26 +173,49 @@ def core_gpt_dataset_config_from_args(args):
...
+        "allow_ambiguous_pad_tokens": args.allow_ambiguous_pad_tokens,
+    }
+
+    # add FIM args to the config
+    if args.fim_data:
+        extra_tokens = {
+            "prefix": args.fim_prefix_token,
+            "middle": args.fim_middle_token,
+            "suffix": args.fim_suffix_token,
+            "pad": args.fim_pad_token,
```
yaoyu-33, line 203: will these be optional to add? will fim always has a different pad / eod token than original pad/ eod? and it's always required?

megatron/training/datasets/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+**Attributes**
+
+- `rate`: Probability of converting a sample into a FIM example.
+- `spm_rate`: Probability of using the SPM FIM pattern (vs PSM).
+- `extra_tokens`: Dictionary containing the FIM special tokens: {"prefix", "middle", "suffix", "pad", "eod"}.
+- `split_sample`: Optional token around which samples are split before applying FIM.
+- `fragment_rate`: Probability of applying FIM to each fragment when split_sample is used.
+- `no_prefix`: If the decoded sequence starts with this prefix, FIM is skipped.
+`GPTFIMDataset` dataset class that loads token sequences from an `IndexedDataset` and applies FIM transformations before returning each sample.
```
yaoyu-33, line 19: how will seq length change in this case? in many cases IndexedDataset has constant seq length?

megatron/training/datasets/README.md
```diff
@@ -0,0 +1,34 @@
...
+`GPTFIMDataset` extends Megatron-Core’s `GPTDataset` to support **Fill-in-the-Middle (FIM)** data augmentation.
+It probabilistically converts samples into FIM format using configurable rates, with support for both PSM and SPM patterns, fragment-level splitting, and length-preserving output.
+
+`GPTFIMDatasetConfig` provides the configuration needed to enable this behavior.
+`GPTFIMDatasetConfig` configuration object extending `GPTDatasetConfig` to enable FIM preprocessing.
+
+**Attributes**
+
+- `rate`: Probability of converting a sample into a FIM example.
+- `spm_rate`: Probability of using the SPM FIM pattern (vs PSM).
```
yaoyu-33: what does this mean? 0.3 means 30% spm and 70% psm? be more clear.

megatron/training/datasets/README.md
```diff
@@ -0,0 +1,34 @@
...
+- `extra_tokens`: Dictionary containing the FIM special tokens: {"prefix", "middle", "suffix", "pad", "eod"}.
+- `split_sample`: Optional token around which samples are split before applying FIM.
```
yaoyu-33: can you add an example, not fully understand.

megatron/training/datasets/README.md
```diff
@@ -0,0 +1,34 @@
...
+- `fragment_rate`: Probability of applying FIM to each fragment when split_sample is used.
+- `no_prefix`: If the decoded sequence starts with this prefix, FIM is skipped.
```
yaoyu-33, line 18: `fim_skip_prefix` is better?

megatron/training/datasets/fim_dataset.py
```diff
@@ -0,0 +1,308 @@
...
+from megatron.core.datasets.utils import Split
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class GPTFIMDatasetConfig(GPTDatasetConfig):
+    """Configuration object for Megatron Core GPT FIM datasets"""
+
+    rate: float = None
```
yaoyu-33: i would add `fim_` prefix to configs, so when debugging and print all configs it's more clear what it is.

megatron/training/datasets/README.md
dimapihtar: seq length will not be changed

pretrain_gpt.py
dimapihtar: not always, but it can be different to original pad / eod. It's up to user.

megatron/training/datasets/README.md
dimapihtar: I'd leave it as it is since it's implemntation from NeMo1

PR #2296, opened 2025-11-19
Changes to support latent MoEs

deepakn94:


kvareddy (approved):

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -272,9 +298,26 @@ def forward(self, hidden_states: torch.Tensor):
...
             output, mlp_bias = self.routed_experts_compute(dispatched_input, probs, residual)
+
+            if self.config.moe_latent_size and mlp_bias is not None:
+                output = output + mlp_bias
+                mlp_bias = None
             output = self.combine(output, shared_expert_output)
+            # Project the output back from latent dimension to hidden dimension after combine
+            # in latent dimension.
+            if self.config.moe_latent_size:
+                output, _ = self.fc2_latent_proj(output)
```
venmugil: `fc2_latent_proj` might return a bias if `self.config.add_bias_linear` is set.

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -272,9 +298,26 @@ def forward(self, hidden_states: torch.Tensor):
...
+                ), "Shared expert overlap not supported when MoE latent projections are used."
+                hidden_states, _ = self.fc1_latent_proj(hidden_states)
+
             dispatched_input, probs = self.dispatch(hidden_states, probs)
             output, mlp_bias = self.routed_experts_compute(dispatched_input, probs, residual)
+
+            if self.config.moe_latent_size and mlp_bias is not None:
+                output = output + mlp_bias
+                mlp_bias = None
             output = self.combine(output, shared_expert_output)
```
venmugil, line 314: `output` will be in latent dimension, while `shared_expert_output` will be in hidden dimension here. We may have to move `self.fc2_latent_proj` inside `self.combine` before the addition of `output` and `shared_expert_output`.

venmugil (changes requested):

megatron/core/transformer/moe/moe_layer.py
deepakn94: Good point. But what should be the right behavior here if this layer does have a bias?

megatron/core/transformer/moe/moe_layer.py
deepakn94: Good point, moved.

yanring:
Thanks for the work. Could you also add UT and integration tests covering this feature combined with EP/TP?

megatron/core/transformer/transformer_config.py
```diff
@@ -216,6 +216,9 @@ class TransformerConfig(ModelParallelConfig):
     """Number of SMs to use for HybridEP. In pure NVL scenarios, 
     16 SMs can generally achieve good bandwidth."""
 
+    moe_latent_size: Optional[int] = None
+    """Latent projection dimension for MoE. If None, MoE latent projections are not used."""
```
yanring: Could you elaborate on it a bit here?

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -272,9 +302,22 @@ def forward(self, hidden_states: torch.Tensor):
...
+            if self.config.moe_latent_size:
+                assert (
+                    not self.shared_expert_overlap
+                ), "Shared expert overlap not supported when MoE latent projections are used."
+                hidden_states, _ = self.fc1_latent_proj(hidden_states)
+
             dispatched_input, probs = self.dispatch(hidden_states, probs)
             output, mlp_bias = self.routed_experts_compute(dispatched_input, probs, residual)
+
+            if self.config.moe_latent_size and mlp_bias is not None:
```
yanring: Please document the change here to explain the bias handling update.

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -123,6 +123,32 @@ def __init__(
         # Initialize router
         self.router = TopKRouter(config=self.config, pg_collection=pg_collection)
 
+        # Initialize latent projections
+        if self.config.moe_latent_size:
+            assert HAVE_TE
```
yanring: Perhaps `assert HAVE_TE, "TransformerEngine is required for MoE latent projections."`

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -123,6 +123,32 @@ def __init__(
...
+                is_expert=False,
+            )
+            self.fc2_latent_proj = TELinear(
+                self.config.moe_latent_size,
+                self.config.hidden_size,
+                parallel_mode="duplicated",
+                config=self.config,
+                init_method=self.config.output_layer_init_method,
+                bias=self.config.add_bias_linear,
+                skip_bias_add=True,
```
venmugil: Perhaps we could set `skip_bias_add=False` so that any necessary bias addition is handled internally by TELinear.

megatron/core/transformer/transformer_config.py
Skylion007: Yeah, do you have reference for latent MoEs?

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -272,9 +302,22 @@ def forward(self, hidden_states: torch.Tensor):
         def custom_forward(hidden_states):
             shared_expert_output = self.shared_experts_compute(hidden_states)
             hidden_states, probs, residual = self.router_and_preprocess(hidden_states)
+
+            # Project the hidden_states from hidden dimension down to latent dimenion.
+            if self.config.moe_latent_size:
+                assert (
+                    not self.shared_expert_overlap
+                ), "Shared expert overlap not supported when MoE latent projections are used."
+                hidden_states, _ = self.fc1_latent_proj(hidden_states)
```
yfw: I believe this projection needs to happen before we call `self.token_dispatcher.dispatch_preprocess` (in `router_and_preprocess`). Otherwise the `hidden_shape` in the token dispatcher gets set to the original hidden size instead of the latent size. This will result in a shape error when trying to apply `fc2_latent_proj`

venmugil: Fixed, thanks.

megatron/core/transformer/moe/moe_layer.py
venmugil: Done!

PR #2306, opened 2025-11-19
[training migration] add training config dataclass and arg generation utility

maanug-nv:
# What does this PR do ?

1. Create dataclass for settings related to the training loop.
2. Create an extensible utility to generate argparse arguments from dataclasses

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


yanring:
Hey @maanug-nv @ericharper, do we have any design doc for this?

megatron/training/config.py
```diff
@@ -0,0 +1,116 @@
...
+    rampup_batch_size: Optional[list[int]] = field(default=None, metadata={"argparse_meta": {"nargs": 3}})
+    """Batch size ramp up with the following values: <start batch size>, <batch size increment>,
+    <ramp-up samples>
+    For example:
+        rampup-batch-size = [16, 8, 300000]
+        global-batch-size 1024
+    will start with global batch size 16 and over (1024 - 16) / 8 = 126 intervals will increase
+    the batch size linearly to 1024. In each interval we will use approximately
+    300000 / 126 = 2380 samples.
+    """
```
ananthsub: can you document the `argparse_meta` metadata schema in more detail in megatron/training/argument_utils.py so it can be used as a reference? otherwise other developers would have to look example usages from these configs

megatron/training/config.py
```diff
@@ -0,0 +1,116 @@
+# Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+from dataclasses import dataclass, field
+import signal
+from typing import Optional, Literal
+
+@dataclass(kw_only=True)
+class TrainingConfig:
+    """Configuration settings related to the training loop."""
+
+    micro_batch_size: Optional[int] = None
```
ananthsub: nit: might as well use the new style of typehints for the new code
```suggestion
    micro_batch_size: int | None = None
```

megatron/training/config.py
```diff
@@ -0,0 +1,116 @@
...
+    """Exit the program after the iteration is divisible by this value."""
+
+    exit_duration_in_mins: Optional[int] = None
+    """Exit the program after this many minutes."""
+
+    exit_signal_handler: bool = False
+    """Dynamically save the checkpoint and shutdown the training if SIGTERM is received"""
+
+    exit_signal: int = int(signal.SIGTERM)
+    """Signal for the signal handler to detect."""
```
ananthsub: exit-signal in the existing argparse is a str, and the conversion to the enum happens here: https://github.com/NVIDIA/Megatron-LM/blob/f42623006694108a7e20850a57e3730294053373/megatron/training/dist_signal_handler.py#L57-L59

tests/unit_tests/test_argument_utils.py
```diff
@@ -0,0 +1,579 @@
...
+        assert args.tags == []
+        assert args.numbers == [1, 2, 3]
+
+        # Provided values
+        args = parser.parse_args(['--tags', 'tag1', 'tag2', 'tag3', '--numbers', '10', '20', '30'])
+        assert args.tags == ['tag1', 'tag2', 'tag3']
+        assert args.numbers == [10, 20, 30]
+
+
+class TestArgumentGroupFactoryLiteral:
```
ananthsub, line 282: could you also add a test for a field typed with union to test the fallback on the argparse meta?

megatron/training/config.py
maanug-nv: Ah, I think I missed this when rebasing. I do like passing `--exit-signal SIGTERM` over `--exit-signal 15`. But I like how `AttnBackend` is handled [here](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/arguments.py#L1580). Basically user can provide a string value for the enum, and at parse time it will be converted to the enum type. 
I could also do this automatically for any enum type in the factory. wdyt?

megatron/training/config.py
maanug-nv: Looks like this still passes argument utils unit tests :+1:

megatron/training/config.py
maanug-nv: Added to the docstring. Let me know if it's sufficient

megatron/training/config.py
maanug-nv: Updated per our discussion. Marking this as resolved.

megatron/training/argument_utils.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+    def _build_argparse_kwargs_from_field(self, attribute: Field) -> dict[str, Any]:
+        """Assemble kwargs for add_argument().
+
+        Args:
+            attribute: dataclass attribute
+        """
+        argparse_kwargs = {}
+        argparse_kwargs["arg_names"] = [self._format_arg_name(attribute.name)]
```
Wohox, line 143: Wondering why list is used here, is there any case 1 attribute maps to >1 arg?

megatron/training/config.py
```diff
@@ -0,0 +1,116 @@
+# Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+from dataclasses import dataclass, field
+import signal
+from typing import Literal
+
+@dataclass(kw_only=True)
+class TrainingConfig:
```
Wohox, line 7: The dataclass looks similar to https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/training/config.py#L522, but different in 2 aspects:
* Train and Validation configs are separated classes in MCore, should MBridge use the same design?
* Please add validations like `finalize` or use `__post_init__`, it's cleaner to have independent checks per dataclass whenever possible. The cross-class validations can be kept the same.

megatron/training/arguments.py
```diff
@@ -1987,41 +1988,14 @@ def _add_rl_args(parser):
     return parser
 
 def _add_training_args(parser):
-    group = parser.add_argument_group(title='training')
+    from megatron.training.config import TrainingConfig
+
+    train_factory = ArgumentGroupFactory(TrainingConfig)
```
lmcafee-nvidia, line 2050: why do we need to build a factory before building the group? Can this be condensed into a single line such as `group = ArgumentGroupFactory.build_group(parser, TrainingConfig, "training")`?

megatron/training/arguments.py
```diff
@@ -1987,41 +1988,14 @@ def _add_rl_args(parser):
     return parser
 
 def _add_training_args(parser):
```
Wohox, line 2047: Why not including all the args into `TrainingConfig`?

megatron/training/argument_utils.py
maanug-nv: I think currently only these [two](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/arguments.py#L2386-L2398). 

But this PR needs to avoid breaking anyone's scripts. Also might be useful in the future to use this for deprecated arg names.

megatron/training/config.py
maanug-nv: 1. Yes we plan to split Train and Validation in MBridge soon
2. We determined that much of the validation/assertions in `arguments.py` should be split up into the dataclasses `finalize()` funcs. But this PR is large enough as is, and that task will be easier once all the dataclasses are merged, so I think we can address that in a future PR. Is that fine? I will note that in the design document as well.

megatron/training/arguments.py
maanug-nv: These are already captured by other existing dataclasses in MCore.

* Activation recompute settings (already captured by `TransformerConfig`)
* `clone_scatter_output_in_embedding` (already captured by `TransformerConfig`)
* `calculate_per_token_loss` (already captured by `TransformerConfig`)
* Fusion settings, flash attn, `persist_layer_norm` (already captured by `TransformerConfig`)
* qkv bias, linear bias (already captured by `TransformerConfig`)
* `use_cpu_initialization` (already captured by `ModelParallelConfig`)
* `deterministic_mode` (already captured by `ModelParallelConfig`)
* `async_tensor_model_parallel_allreduce` (already captured by `ModelParallelConfig`)
* `sequence_parallel`, `pipeline_model_parallel_comm_backend` (already captured by `ModelParallelConfig`)
* Check for spiky/nan in grad (already captured by `DDPConfig`)
* optimizer settings (already captured by `OptimizerConfig`)

In Megatron Bridge, these are in different configs.

* Nsys/Pytorch profiler settings (contained in `ProfilingConfig` in Megatron Bridge)
* tensorboard and logging settings (contained in `LoggerConfig` in Megatron Bridge)
* `dataloader_type` (contained in `DataloaderConfig` in Megatron Bridge)
* Gloo process groups (contained in 'DistributedInitConfig\` in Megatron Bridge)
* `high_priority_stream_groups` (contained in `DistributedInitConfig` in Megatron Bridge)

maanug-nv: We want to keep the dataclass structure similar to Megatron Bridge to reduce changes there. 

And we are working on upstreaming the rest of the dataclasses as well: https://github.com/NVIDIA/Megatron-LM/issues/2313

megatron/training/arguments.py
maanug-nv: Is the issue just how many lines it takes to call? then that can be reduced even now with `group = ArgumentGroupFactory(TrainingConfig).build_group(parser, "training")` . 

I think building the factory first also allows you to modify the docstrings.

What's the benefit of merging `__init__()` and `build_group()` ?

maanug-nv: ~~I can add a classmethod version of `build_group()` that accomplishes this, and it can simply call init and build group sequentially. Does that work?~~

Ok I didn't realize that it's not possible to overload like this in python. Having instance methods and class methods with the same name is not supported. So I'd rather not do this. 

If there's some benefit you see to merging init and build group, I can redesign ArgumentGroupFactory that way, but if it's just a matter of preference, I'd like to avoid changing all the unit tests that use the existing interface.

megatron/training/argument_utils.py
Wohox: - Correct me if I'm wrong: The `ArgumentGroupFactory` maps config to args, during the process `_`  will be replaced with `-`, so in order for `ArgumentGroupFactory` to map 1 attribute to 2 args(forward compatibility), you would need to hack one of the `-`  to `_` and I don't find the corresponding code in this MR, in another words, I don't think this MR can achieve what you described here and should break scripts anyway.
- Should we fix these two arguments instead?

megatron/training/config.py
Wohox: Sure, this makes sense to me.

megatron/training/arguments.py
Wohox: Wondering do you plan to move these args to their corresponding dataclasses in the future? It seems some of the args shouldn't reside in `_add_training_args`.

megatron/training/argument_utils.py
maanug-nv: Here's how i achieve that using metadata: https://github.com/NVIDIA/Megatron-LM/pull/2400/files#diff-cce0b7f6f5aa1a135ba21b6a3ac980e3c4a45c25b076a4fd317dcd740a7c4b13R154

I don't have any opinion on whether there should only be one of these flags or not. I just want to get these PRs merged as quickly as possible, so I'm trying my best to preserve current behavior.

megatron/training/arguments.py
maanug-nv: Yeah, we are planning to migrate all Megatron Bridge dataclasses (maybe some exceptions) and also generate from MCore dataclasses in the future. We're splitting this effort over several PRs to keep each one small.

maanug-nv: Once we can generate from the MCore dataclasses, esp. TransformerConfig, arguments.py will be a lot smaller. At that point, would be good to do some cleanup, e.g. i don't know if the `_add_<group>_args()` functions will be necessary.

yanring (approved):
LGTM

megatron/training/arguments.py
lmcafee-nvidia: it is about reducing the amount of code, but I guess that's just my preference

ericharper (approved):

PR #2309, opened 2025-11-19
Flag to patch TE GEMMs with recalculate verification.

kwyss-nvidia:
Co-authored by: Cyril Meurillon <cmeurillon@nvidia.com>

# What does this PR do ?

Adds an option to verify GEMM results are repeatable.

This can identify bugs that are either faults in the determinism of the BLAS or issues with a device.
It has the potential to triage misbehaving devices.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/extensions/gemm_patch.py
+++ b/megatron/core/extensions/gemm_patch.py
@@ -83,7 +83,7 @@ def new_general_gemm(A, B, *args, out=None, accumulate=False, bias=None, **kwarg
 
 
 def new_general_grouped_gemm(
-    A, B, out, out_dtype, *args, grad=False, accumulate=False, bias=None, **kwargs
+    A, B, out, *args, grad=False, accumulate=False, bias=None, **kwargs
 ):
     """
     A wrapper for general_grouped_gemm that calls the original method twice
@@ -106,7 +106,6 @@ def new_general_grouped_gemm(
             A,
             B,
             out_first_iter if is_first_iter else out,
-            out_dtype,
             *args,
             grad=grad if not is_first_iter else False,
             accumulate=accumulate,


kwyss-nvidia:
Tagging @deepakn94

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,187 @@
...
+    if _GLOBAL_IS_PATCHED_ is not None and _GLOBAL_IS_PATCHED_:
+        return
+    transformer_engine.pytorch.cpp_extensions.gemm.general_gemm = new_general_gemm
+    transformer_engine.pytorch.cpp_extensions.gemm.general_grouped_gemm = new_general_grouped_gemm
+
+    rank = int(os.environ.get("RANK", 0))
+    for module in sys.modules:
+        if 'general_gemm' in dir(sys.modules[module]):
+            if rank == 0:
+                print(f"PATCHING general_gemm IN MODULE: {module}")
```
deepakn94: No `print`s in `megatron/core`. Use Python's `logging` module instead.

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,187 @@
...
+def patch_te_gemms() -> None:
+    global _GLOBAL_IS_PATCHED_
+    if _GLOBAL_IS_PATCHED_ is not None and _GLOBAL_IS_PATCHED_:
+        return
+    transformer_engine.pytorch.cpp_extensions.gemm.general_gemm = new_general_gemm
+    transformer_engine.pytorch.cpp_extensions.gemm.general_grouped_gemm = new_general_grouped_gemm
+
+    rank = int(os.environ.get("RANK", 0))
+    for module in sys.modules:
+        if 'general_gemm' in dir(sys.modules[module]):
```
deepakn94, line 145: Is this the cleanest way to monkey patch this?

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,187 @@
...
+    elif isinstance(t, (NVFP4Tensor, NVFP4TensorStorage)):
+        return "nvfp4"
+    elif isinstance(t, torch.Tensor):
+        return DTYPES.get(t.dtype, "torch-unknown")
+    else:
+        return "unknown"
+
+
+def new_general_gemm(
+    A,
```
deepakn94: Do we actually need the full argument lists here and below? Can we make this `*args, **kwargs`?

megatron/core/extensions/gemm_patch.py
deepakn94: You can use https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/blob/main/megatron/core/utils.py?ref_type=heads#L737 to just log on a single rank.

kwyss-nvidia: Done.

megatron/core/extensions/gemm_patch.py
kwyss-nvidia: I'm not sure. There are two parts to this. Future import statements should see the result of

```
    transformer_engine.pytorch.cpp_extensions.gemm.general_gemm = new_general_gemm
```

The sys.modules patching is to make sure that any existing resolutions to these names also get updated.

Perhaps it's possible to have the patch ordered before any transformer engine imports run, but that seems precarious.

Do you have an idea of a cleaner way? Monkey patching is not a very clean pattern, but this does work.

megatron/core/extensions/gemm_patch.py
kwyss-nvidia: This would rely on TE calls to general_gemm using keyword arguments and not positional arg resolution for keyword arguments.

TE style seems pretty consistent about it, so that looks like a reliable assumption at first glance.

kwyss-nvidia: Is this what you have in mind? I will try running with this arg after these changes to check I didn't break anything.

megatron/training/training.py
```diff
@@ -614,6 +614,10 @@ def pretrain(
     args = get_args()
     timers = get_timers()
 
+    if args.repeat_te_gemms_check:
+        from megatron.core.extensions.gemm_patch import patch_te_gemms
```
jaredcasper, line 620: Should make sure that we are using TE?

megatron/core/extensions/gemm_patch.py
jaredcasper: How confident are we there isn't another "general_gemm" in there that isn't the TE gemm?

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,170 @@
...
+            A, B, workspace, out=out_arg, accumulate=accumulate, bias=bias_arg, **kwargs
+        )
+        out_res = res[0]
+        is_rerun = run_idx == 1
+        if will_rerun or is_rerun:
+            cksum[run_idx] = out_res.sum(dtype=torch.float32).item()
+        if will_rerun:
+            rerun_res = out_res
+            continue
+        if is_rerun:
```
jaredcasper: Why all this will_rerun/is_rerun logic? Why not just put this stuff after the loop?

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,170 @@
...
+        else:
+            out_arg = out
+            bias_arg = bias
+
+        res = orig_general_gemm(
+            A, B, workspace, out=out_arg, accumulate=accumulate, bias=bias_arg, **kwargs
+        )
+        out_res = res[0]
+        is_rerun = run_idx == 1
+        if will_rerun or is_rerun:
```
jaredcasper: How will this ever not be true?

megatron/core/extensions/gemm_patch.py
cyme: Re: patching before any TE imports -- AFAIK it's impossible b/c of a chicken-and-egg situation. We need to import TE in order to patch general_gemm, and importing TE triggers additional imports that expose general_gemm to other modules before we have a chance to perform the patch.

Agreed monkey patching is not a clean pattern. But I'm not aware of a clean and effective solution short of implementing this capability in TE itself. We can take that as a next step.

Re: risk of "another general_gemm in there that isn't the TE gemm" -- it's definitely a risk, at least in theory. We can protect against it by checking if the symbol we're patching resolves to the TE GEMM.

megatron/core/extensions/gemm_patch.py
kwyss-nvidia: These aren't referring to a coarse grain rerun. will_rerun/is_rerun are whether this is the first or second invocation of the GEMM in a loop.

megatron/core/extensions/gemm_patch.py
kwyss-nvidia: Good point. It can be elided.

megatron/core/extensions/gemm_patch.py
kwyss-nvidia: This is a debugging feature. How much extra complexity is warranted to check against name collisions and ensure TE use if the user has opted into a whitebox flag?

megatron/training/training.py
kwyss-nvidia: As in "transformer_impl"==TE?

megatron/core/extensions/gemm_patch.py
kwyss-nvidia: Renamed for clarity.

megatron/core/extensions/gemm_patch.py
kwyss-nvidia: Added a check that we only patch the TE modules to increase safety.

megatron/training/training.py
kwyss-nvidia: Done.

megatron/core/extensions/gemm_patch.py
deepakn94: Looks good, resolving.

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,175 @@
...
+        res = orig_general_gemm(
+            A, B, workspace, out=out_arg, accumulate=accumulate, bias=bias_arg, **kwargs
+        )
+        out_res = res[0]
+        is_second_iter = run_idx == 1
+        cksum[run_idx] = out_res.sum(dtype=torch.float32).item()
+        if is_first_iter:
+            rerun_res = out_res
+            continue
+        if is_second_iter:
```
jaredcasper: If this only runs on the second iteration, can't it just run outside of the loop after it finishes? Even with the name changes this loop logic with is_first_iter and is_second_iter is unnecessarily complex.

deepakn94: Yeah, I think that's a good suggestion.

kwyss-nvidia: Moving the check and return outside of the loop is equivalent, and is arguably easier to read (at least to my judgment). Thanks for the feedback.

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,182 @@
...
+        )
+        out_res = res[0]
+        cksum[run_idx] = out_res.sum(dtype=torch.float32).item()
+        if is_first_iter:
+            rerun_res = out_res
+            continue
+
+    # Check the consistency of the first and second run and return.
+    same_cksum = math.fabs(cksum[1] - cksum[0]) / (math.fabs(cksum[1]) + EPSILON) <= CKSUM_TOLERANCE
+    if not same_cksum:
```
cyme: What is the value of doing a cksum comparison if we do a bit-wise comparison anyway?

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,182 @@
...
+        logger.log(
+            logging.ERROR,
+            f"RANK {get_rank_info()}: DIFFERENT CHECKSUM ON 2ND RUN for dtype "
+            f"{get_dtype(A)}. {cksum[1]} != {cksum[0]}, "
+            f"shapes A:{list(A.size())},B:{list(B.size())},out:{list(out_res.shape)}",
+        )
+    elif not torch.equal(rerun_res, out_res):
```
cyme:
```suggestion
    elif not torch.equal(rerun_res.view(dtype=torch.uint8), out_res.view(dtype=torch.uint8))):
```
Doing the equality test in the int domain ensure a bit-wise comparison. This correctly handles NaNs, which are never equal in the FP domain.

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,182 @@
...
+        logger.log(
+            logging.ERROR,
+            f"RANK {get_rank_info()}: DIFFERENT BITWISE on 2ND RUN for dtype "
+            f"{get_dtype(A)}. "
+            f"shapes A:{list(A.size())},B:{list(B.size())},out:{list(out_res.shape)}",
```
cyme:
```suggestion
            f"RANK {get_rank_info()}: DIFFERENT BITWISE on 2ND RUN FOR OPERATION "
            f"A:({get_dtype(A)},{list(A.size())}) x B:({get_dtype(B)},{list(B.size())}) "
            f"-> OUT:({get_dtype(out_res)},{list(out_res.size())})",
```
Logging this too as it's very useful

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,182 @@
...
+            **kwargs,
+        )
+        output = res[0]
+        if is_first_iter:
+            saved_output = [t.clone().detach() for t in output]
+            continue
+
+    # Check comparison and return
+    for tidx in range(len(output)):
+        if not torch.equal(output[tidx], saved_output[tidx]):
```
cyme:
```suggestion
        if not torch.equal(output[tidx].view(dtype=torch.uint8), saved_output[tidx].view(dtype=torch.uint8)):
```
Same suggestion as above

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,182 @@
...
+            logging.log(
+                logging.ERROR,
+                f"RANK {get_rank_info()}: DIFFERENT GROUPED_GEMM RESULT ON 2ND RUN FOR OP "
+                f"GroupedGEMM group {tidx} = {get_dtype(A[tidx])}({list(A[tidx].size())}) "
+                f"x {get_dtype(B[tidx])}({list(B[tidx].size())}) -> "
+                f"{get_dtype(output[tidx])}({list(output[tidx].size())})",
```
cyme:
```suggestion
                f"RANK {get_rank_info()}: DIFFERENT GROUPED_GEMM RESULT ON 2ND RUN FOR "
                f"GROUP {tidx}, A:({get_dtype(A[tidx])},{list(A[tidx].size())}) "
                f"x B:({get_dtype(B[tidx])},{list(B[tidx].size())}) -> "
                f"OUT: ({get_dtype(output[tidx])},{list(output[tidx].size())})",
```
nit: made logging consistent with `general_gemm()` for easier parsing

megatron/core/extensions/gemm_patch.py
```diff
@@ -0,0 +1,182 @@
...
+def new_general_gemm(A, B, workspace, out=None, accumulate=False, bias=None, **kwargs):
+    """
+    A wrapper for general_gemm that calls the original method twice
+    and compares results for consistency.
+    """
+    cksum = [0, 0]
+    # Run twice for comparison.
+    for run_idx in range(2):
+        is_first_iter = run_idx == 0
+        need_output_copy = out is not None and (bias is out or accumulate)
```
cyme, line 56: I'm not sure `bias is out` is a legitimate use case, but it doesn't hurt to check

megatron/core/extensions/gemm_patch.py
cyme: Alternatively to making a copy of the first result and doing an actual bit-wise comparison, we can use a hash. This would reduce a bit the overhead. PyTorch2.9 introduced a `hash_tensor()` function. On previous versions, a possible poor man hash is a sum of the tensor viewed as int64s.

megatron/core/extensions/gemm_patch.py
kwyss-nvidia: It provides a bit of information whether the error was large in magnitude.

kwyss-nvidia: Arguably marginal value. Removed.

megatron/core/extensions/gemm_patch.py
kwyss-nvidia: Yep. Worth handling the edge case. Are you requesting a change?

megatron/core/extensions/gemm_patch.py
kwyss-nvidia: Done.

Phlip79:
**We are changing our review process and marking all open, unlabeled PRs as draft. This change will go in effect starting once https://github.com/NVIDIA/Megatron-LM/pull/3659 is merged.**

Moving forward, all PRs will be required to start as draft PRs. If you wish to get your PR merged, mark your PR as “Ready for review”. Read more about the new process at [submit.md](https://github.com/Phlip79/Megatron-LM/blob/d3763629bcdd04acca52477af7eb8795090426ff/docs/developer/submit.md).

PR #2346, opened 2025-11-21
Various small fixes for Megatron-FSDP.

cspades:
# What does this PR do ?

- **[_Delayed for future discussion. Will split off from this PR._]** Remove the `hybrid_fsdp_group` requirement when using HSDP without optimizer state sharding.
  - Main WAR is to avoid using `FSDPDistributedIndex.get_dp_group()` when not necessary because it refers to `FSDPDistributedIndex.hybrid_fsdp_group` even when `FSDPDistributedIndex.hsdp_outer_dp_shard=False` (such as during HSDP with DP-Replicate) and we don't need this group to compute the DP size for FSDP or HSDP. All other instances of `get_dp_group()` happen when `FSDPDistributedIndex.hybrid_fsdp_group` exists due to DP-Outer optimizer sharding in HFSDP.
- Fix a recent Torch 2.9 [`DeviceMesh._flatten` validation error](https://github.com/pytorch/pytorch/blob/main/torch/distributed/device_mesh.py#L723) when running `gather_uneven_dtensor_to_full_tensor` on a 1D sharding mesh. This unblocks BioNeMo unit tests that rely on `gather_uneven_dtensor_to_full_tensor` to test DCP checkpointing functionality.
  - Note that more serious backward-incompatible changes for DeviceMesh will be released soon: https://github.com/pytorch/pytorch/pull/164510
```
FAILED tests/test_distributed_checkpointing.py::test_final_model_save_mfsdp - RuntimeError: ("dp already exists for submesh of the DeviceMesh((dp=1, tp=1), device: 'cuda', stride: ...
```
- Fix a hanging gradient unit test (`tests/unit_tests/distributed/fsdp/test_mfsdp_fully_shard.py::TestMegatronFsdpFullyShard::test_fully_shard`) due to not asserting failure on every rank.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/.github/actions/check-nvidia-sso-membership/action.yml
+++ b/.github/actions/check-nvidia-sso-membership/action.yml
@@ -1,0 +1,139 @@
+name: 'Check NVIDIA SSO Membership'
+description: 'Check if a GitHub username exists in the NVIDIA SSO users list from github-audits'
+author: 'NVIDIA'
+
+inputs:
+  username:
+    description: 'GitHub username to check'
+    required: true
+  github_audits_repo:
+    description: 'Repository containing SSO users file'
+    required: false
+    default: 'NVIDIA-GitHub-Management/github-audits'
+  github_audits_version:
+    description: 'Release version tag'
+    required: false
+    default: 'v0.1.0'
+  sso_users_filename:
+    description: 'Filename of SSO users JSON'
+    required: false
+    default: 'users_sso.json'
+  github_token:
+    description: 'GitHub token with access to github-audits repo'
+    required: true
+
+outputs:
+  is_member:
+    description: 'Boolean - true if user is in NVIDIA SSO list, false otherwise'
+    value: ${{ steps.check-membership.outputs.is_member }}
+  is_org_member:
+    description: 'Boolean - true if user has NVIDIA or NVIDIA-NeMo in org_roles'
+    value: ${{ steps.check-membership.outputs.is_org_member }}
+  user_orgs:
+    description: 'Comma-separated list of orgs user is member of'
+    value: ${{ steps.check-membership.outputs.user_orgs }}
+  sso_file_available:
+    description: 'Boolean - true if SSO file was successfully downloaded'
+    value: ${{ steps.download-sso.outputs.sso_file_available }}
+  user_count:
+    description: 'Number of users in the SSO file (0 if download failed)'
+    value: ${{ steps.download-sso.outputs.user_count }}
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Download NVIDIA SSO users from github-audits
+      id: download-sso
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github_token }}
+      run: |
+        echo "Downloading ${{ inputs.sso_users_filename }} from ${{ inputs.github_audits_repo }} ${{ inputs.github_audits_version }} release..."
+
+        # Download the release asset using gh CLI
+        gh release download ${{ inputs.github_audits_version }} \
+          --repo ${{ inputs.github_audits_repo }} \
+          --pattern ${{ inputs.sso_users_filename }} \
+          --clobber 2>&1 || {
+            echo "ERROR: Failed to download ${{ inputs.sso_users_filename }} from github-audits release"
+            echo "sso_file_available=false" >> $GITHUB_OUTPUT
+            echo "user_count=0" >> $GITHUB_OUTPUT
+            exit 0
+          }
+
+        # Verify file was downloaded and is valid JSON
+        if [ ! -f ${{ inputs.sso_users_filename }} ]; then
+          echo "ERROR: ${{ inputs.sso_users_filename }} file not found after download"
+          echo "sso_file_available=false" >> $GITHUB_OUTPUT
+          echo "user_count=0" >> $GITHUB_OUTPUT
+          exit 0
+        fi
+
+        # Validate JSON structure
+        if ! jq -e 'type == "object"' ${{ inputs.sso_users_filename }} > /dev/null 2>&1; then
+          echo "ERROR: ${{ inputs.sso_users_filename }} is not a valid JSON object"
+          echo "sso_file_available=false" >> $GITHUB_OUTPUT
+          echo "user_count=0" >> $GITHUB_OUTPUT
+          exit 0
+        fi
+
+        USER_COUNT=$(jq 'length' ${{ inputs.sso_users_filename }})
+        echo "Successfully downloaded ${{ inputs.sso_users_filename }} with $USER_COUNT NVIDIA SSO users"
+        echo "sso_file_available=true" >> $GITHUB_OUTPUT
+        echo "user_count=$USER_COUNT" >> $GITHUB_OUTPUT
+
+    - name: Check if user is in SSO list
+      id: check-membership
+      shell: bash
+      run: |
+        USERNAME="${{ inputs.username }}"
+        SSO_FILE="${{ inputs.sso_users_filename }}"
+
+        echo "Checking if $USERNAME is in NVIDIA SSO users list..."
+
+        # Check if SSO file is available
+        if [ "${{ steps.download-sso.outputs.sso_file_available }}" != "true" ] || [ ! -f "$SSO_FILE" ]; then
+          echo "ERROR: $SSO_FILE not available - cannot check membership"
+          echo "is_member=false" >> $GITHUB_OUTPUT
+          echo "is_org_member=false" >> $GITHUB_OUTPUT
+          echo "user_orgs=" >> $GITHUB_OUTPUT
+          exit 0
+        fi
+
+        # Check if username exists as a key in the JSON object
+        if jq -e --arg user "$USERNAME" 'has($user)' "$SSO_FILE" > /dev/null 2>&1; then
+          echo "$USERNAME found in NVIDIA SSO users"
+          echo "is_member=true" >> $GITHUB_OUTPUT
+
+          # Extract and check org membership
+          IS_ORG_MEMBER=$(jq -r --arg user "$USERNAME" '
+            .[$user].org_roles // [] |
+            map(select(test("^(NVIDIA|NVIDIA-NeMo):Member$"))) |
+            length > 0
+          ' "$SSO_FILE")
+
+          USER_ORGS=$(jq -r --arg user "$USERNAME" '
+            .[$user].org_roles // [] |
+            map(split(":")[0]) |
+            unique |
+            join(",")
+          ' "$SSO_FILE")
+
+          echo "is_org_member=$IS_ORG_MEMBER" >> $GITHUB_OUTPUT
+          echo "user_orgs=$USER_ORGS" >> $GITHUB_OUTPUT
+
+          if [ "$IS_ORG_MEMBER" == "true" ]; then
+            echo "$USERNAME is a member of NVIDIA or NVIDIA-NeMo org"
+          else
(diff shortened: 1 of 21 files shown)


cspades:
/ok to test [b8f5682](https://github.com/NVIDIA/Megatron-LM/pull/2346/commits/b8f56825e8333a27946129c386e97604a6f99777)

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -311,12 +311,21 @@ def _init_fsdp_param_and_grad_buffer(self):
         else:
             if self.ddp_config.average_in_collective:
                 gradient_scaling_factor = 1.0
+                # Utilized to re-scale expert gradients to DP.
+                # (edp_size/dp_size) * (1/edp_size) = 1/dp_size
+                # FIXME(@cspades): Currently not used gradient_reduce_preprocessing()?
                 expert_gradient_scaling_factor = (
                     self.dist_index.get_dp_group(is_expert_parallel=True).size()
-                    / self.dist_index.get_dp_group().size()
```
shjwudp, line 316: Will the torch 2.9 check affect the behavior of `get_dp_group().size()`? Do we need to update the logic here?

megatron/core/distributed/fsdp/src/megatron_fsdp/fully_shard.py
```diff
@@ -142,12 +142,16 @@ def fully_shard_model(
             "zero_dp_strategy to use FSDP ('optim_grads_params', 3), because "
             "outer sharding is dependent on inner sharding."
         )
-    if (dp_outer_dim is None) ^ (hybrid_fsdp_group is None):
-        # XOR - HSDP requires both or neither of dp_outer_dim and hybrid_fsdp_group
-        # to be specified, so if XOR then raise an error.
+    if _outer_fsdp_sharding and hybrid_fsdp_group is None:
```
shjwudp: Do we need to change this line? My understanding is that we only need to handle [the new ValidateError](https://github.com/pytorch/pytorch/blob/main/torch/distributed/device_mesh.py#L723) introduced by the PyTorch 2.9 and perform the check before calling `device_mesh._flatten`.

shjwudp (changes requested):

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
cspades: I don't think this change is related to Torch version and upcoming (not released yet) DeviceMesh changes, it's related to our API requiring `FSDPDistributedIndex.hybrid_fsdp_group` which is used by `get_dp_group()`.

I don't see a clear way to flatten DP-Inner and DP-Outer without messing with the user's distributed environment, so I don't think there is a clear way to improve the implementation of `get_dp_group()` at this moment.

So to summarize, the issue is:
- Both HSDP and HFSDP use `get_dp_group() -> FSDPDistributedIndex.hybrid_fsdp_group`.
  - While HSDP does not "need" `FSDPDistributedIndex.hybrid_fsdp_group`, it makes sense for `get_dp_group()` to return the fully-flattened DP group when we are using HSDP, since that is the group that the function is designed to get.
- HSDP does not need this for core Megatron-FSDP functionality. Only used to compute `dp_group.size()`.

So I WAR the `size()` issue like this.

megatron/core/distributed/fsdp/src/megatron_fsdp/fully_shard.py
cspades: Similar to your previous comment, this change and the above change are related to (1), so not related to the new validation error. The original argument validation was that if you are using HSDP or HFSDP, in which case `dp_outer_dim is not None`, then you must pass `hybrid_fsdp_group`. Now I am relaxing it to only when you are using HFSDP.

While we are talking about this, do you think this is a reasonable relaxation? I currently cannot imagine any situation where we need the fully-flattened DP ranks when we are using HSDP, since HSDP's collectives seem simple and orthogonal.

The 3 things I'm doing in this PR:
1. Make HSDP easier to use.
2. Fix the DeviceMesh validation error.
3. Fix gradient unit tests.

megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py
```diff
@@ -272,7 +274,14 @@ def gather_uneven_dtensor_to_full_tensor(
...
     else:
-        process_group = device_mesh._flatten().get_group()
+        # Check if the fully-flattened mesh exists first.
+        full_flattened_mesh_dim_name = "_".join(device_mesh.mesh_dim_names)
+        if full_flattened_mesh_dim_name in get_mesh_names(device_mesh):
+            # Retrieve the existing flattened DeviceMesh ProcessGroup.
+            process_group = device_mesh[full_flattened_mesh_dim_name].get_group()
+        else:
+            # Create the _-separated flattened DeviceMesh ProcessGroup.
+            process_group = device_mesh._flatten().get_group()
```
cspades, line 284: @shjwudp Everything related to (2) is fixed in 3-4 lines of code here^^^

Before, we just immediately `_flatten()`. Going into Torch 2.9, they will not allow us to create new DeviceMesh that matches the flattened name of an existing DeviceMesh. So I just use our helper function `get_mesh_names()` which checks for sub- and flat- dimensions and if I find an existing flattened mesh, we just use that mesh.

I believe this still potentially has loopholes. If the user creates a flattened DeviceMesh dimension with the same name but different topology than our desired mesh, then it will use the user's setting instead. I feel like there is no way to fix this fundamental issue (but we can add a warning message, that may be a good idea), so it will be the user's responsibility to have reasonably-named flattened meshes, i.e. `dp_cp` ~ the flattening of the `dp` and `cp` dims, which is the default behavior (i.e. `"_".join([<mesh dims to flatten>])`) of `device_mesh._flatten()`.

Another thing to note is the `DTensor.device_mesh` here is a child/sub-mesh of the Megatron-FSDP root mesh. In future Torch versions, the flattened mesh will be a member of the DeviceMesh used to flatten. So there will be a lower chance of issues, because the user will likely call `root_mesh._flatten()` but not call `root_mesh[("dp_shard", "dp_outer")]._flatten()`, so we will be less likely to accidentally use the user's original DeviceMesh!

Skylion007:
Any feedback on the device mesh changes in upstream torch before they land btw?

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
shjwudp: Sorry, if the issue isn’t caused by changes in the environment (torch 2.9), why do we still need to modify the API? I understand this might improve the code’s robustness, but what tangible benefits would we actually gain?

megatron/core/distributed/fsdp/src/megatron_fsdp/fully_shard.py
shjwudp: Thanks for the explanation. Could we split this PR? Items 2 and 3 would make great contributions. As for the first one, I’m still a bit confuse.

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
cspades: Just usability. FSDP2 doesn't require you to create any custom groups to enable DP-Replicate HSDP, and I believe that the code changes necessary to make the HSDP pathway seamless aren't intrusive.

The modifications to the API are rather benign, because they are backwards compatible.

Before:
- HSDP: Must pass DP-Outer and flattened group.
- HFSDP: Must pass DP-Outer and flattened group.

Now:
- HSDP: Must pass DP-Outer.
  - You may pass the flattened DP group or not, it will not be used by Megatron-FSDP, so users can keep their old code without any change in functionality.
  - So you can either set `(dp_outer_dim="dp_outer", hybrid_fsdp_group=None)` or `(dp_outer_dim="dp_outer", hybrid_fsdp_group=device_mesh[("dp_shard", "dp_outer")]._flatten().get_group())`, both will work exactly the same way.
- HFSDP: Must pass DP-Outer and flattened group.

Thus, users will not encounter any migration or backwards incompatibility issues, because even when they were using HSDP before, Megatron-FSDP was only using the flattened group here.

cspades:
> Any feedback on the device mesh changes in upstream torch before they land btw?

@Skylion007 I think they are in the right direction. It was quite annoying that all DeviceMesh created during a program all shared the same global dictionaries that was used to retrieve flattened and/or sub-meshes, and I also encountered random bugs where a `backward` call would actually empty those dictionaries, which is why we have `FSDPDistributedIndex` to store references to everything we need. (I believe there is in-line commentary in the DeviceMesh source code that also points to this best practice but for other reasons.)

I didn't read all of the code, but if we are moving into the direction where a DeviceMesh's state is completely defined by its class, which also means that if you have multiple DeviceMesh instances that can create their own flattened DeviceMesh and track those instances within this instance, that's a good thing! I hope this also means we can possibly do deep-copies of DeviceMesh and any sub- or flattened instances that are constructed by it, so that we can maybe clone & snapshot a distributed environment that is completely separate from the user's original environment.

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
cspades: Maybe I can test both argument combinations or even `del hybrid_fsdp_group if outer_dp_sharding_strategy != "optim"`, but if you recall, the only time we need `get_dp_group()` is when we are fully-sharding the optimizer state (which calls the logical ranks function + DataParallelBuffer init), which requires `outer_dp_sharding_strategy="optim"`.

megatron/core/distributed/fsdp/src/megatron_fsdp/fully_shard.py
cspades: Sure, can do! Will split it tomorrow morning.

Let's hop on a call tomorrow evening for (1). Maybe we can plan ahead - when the flattened groups are tracked in DeviceMesh-specific dictionaries, that means it will be much safer for us to create new flattened DP groups during `MegatronFSDP.__init__`, and maybe we can even deprecate `hybrid_fsdp_group` entirely if we can deeply clone the user's DeviceMesh. We can wait until the DeviceMesh API is less volatile before doing (1).

.github/workflows/check_api_backwards_compatibility_workflow.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       # Default baseline for automatic PR checks
       # Can be: branch name (e.g., 'main'), commit hash, or tag
       # Will be resolved to commit hash during execution
-      DEFAULT_BASELINE: '712dff880cdf88e51289ad71e47d92f46d25a2d3'
+      DEFAULT_BASELINE: 'f7fb5ecbe218672719053fa304d91767ce30ffa1'
```
chtruong814: we recently bumped this on main.  If you merge main, does that address the issue?

cspades: Yup it did, thanks to @pablo-garay !

cspades: Oh, happened again, will do!

deepakn94 (approved):

PR #2363, opened 2025-11-23
Add MTP support for hybrid models

rkarimimahab:
# What does this PR do ?

(1) supporting to use hybrid mamba models as  mtp_model_layer.
(2) splitting the MTP loss calculation in the GPT model’s forward pass into a separate function.
(3) supporting MTP layer repetition.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/tests/unit_tests/ssm/test_mamba_hybrid_layer_allocation.py
+++ b/tests/unit_tests/ssm/test_mamba_hybrid_layer_allocation.py
@@ -160,10 +160,18 @@ class TestParseHybridPattern:
 
     def test_invalid_symbols_in_mtp_pattern(self):
         """Test that invalid symbols in MTP pattern raise ValueError."""
-        invalid_patterns = ["M*M*/MX", "M*M*/MM/Ma"]  # X is not valid  # a is not valid
-        for pattern in invalid_patterns:
-            with pytest.raises(ValueError, match="not a valid layer symbol"):
-                parse_hybrid_pattern(pattern)
+        # Single MTP depth with invalid symbol - should raise "not a valid layer symbol"
+        with pytest.raises(ValueError, match="not a valid layer symbol"):
+            parse_hybrid_pattern("M*M*/MX")  # X is not valid
+
+        # Multiple MTP depths with invalid symbol and matching patterns
+        with pytest.raises(ValueError, match="not a valid layer symbol"):
+            parse_hybrid_pattern("M*M*/Ma/Ma")  # a is not valid
+
+        # Multiple MTP depths with invalid symbol but mismatched patterns
+        # This raises "All MTP patterns must be identical" before checking symbols
+        with pytest.raises(ValueError, match="All MTP patterns must be identical"):
+            parse_hybrid_pattern("M*M*/MM/Ma")
 
     def test_empty_main_pattern_with_mtp(self):
         """Test pattern that starts with / (empty main pattern)."""


megatron/core/pipeline_parallel/schedules.py
```diff
@@ -382,7 +382,7 @@ def forward_step(
 
     if is_first_microbatch and hasattr(model, 'set_is_first_microbatch'):
         model.set_is_first_microbatch()
-    if current_microbatch is not None:
+    if config.external_cuda_graph and current_microbatch is not None:
```
BestJuly: This would change the original behavior logic. Is it a bugfix or any particular reason to change this?

megatron/core/models/mamba/mamba_layer_specs.py
```diff
@@ -17,6 +18,15 @@
 from megatron.core.transformer.mlp import MLP, MLPSubmodules
 from megatron.core.transformer.spec_utils import ModuleSpec
 from megatron.core.transformer.transformer_layer import TransformerLayer, TransformerLayerSubmodules
+from megatron.core.transformer.multi_token_prediction import MultiTokenPredictionBlockSubmodules
```
BestJuly: Duplicated import modules.

tests/unit_tests/transformer/test_multi_token_prediction.py
BestJuly, line 1: Current change is only about the naming part. Should add tests for a small hybrid model with mamba.

megatron/core/models/mamba/mamba_layer_specs.py
```diff
@@ -82,3 +108,58 @@
...
+    mtp_num_layers = config.mtp_num_layers if config.mtp_num_layers else 0
+    mtp_layer_specs = [mtp_layer_spec] * mtp_num_layers
+
+    offset = get_mtp_layer_offset(config)
+    # split the mtp layer specs to only include the layers that are built in this pipeline stage.
+    mtp_layer_specs = mtp_layer_specs[offset : offset + num_layers_to_build]
+    if len(mtp_layer_specs) > 0:
+        assert (
+            len(mtp_layer_specs) == config.mtp_num_layers
+        ), +f"currently all of the mtp layers must stage in the same pipeline stage."
```
BestJuly: Should not have `+`

megatron/core/models/mamba/mamba_layer_specs.py
```diff
@@ -25,6 +35,22 @@
     moe_use_legacy_grouped_gemm=False,
 )
 
+try:
+    from megatron.core.extensions.transformer_engine import TEFusedMLP, TENorm
+    from megatron.core.extensions.transformer_engine_spec_provider import TESpecProvider
+
+    HAVE_TE = True
```
BestJuly: Where would `HAVE_TE` and `HAVE_KITCHEN` be used? Also, `TENorm` has already been imported in line#6 in this file.

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -926,6 +963,81 @@ def forward(
         hidden_states = torch.cat(hidden_states_list, dim=0)
         return hidden_states
 
+    def process_loss(
```
BestJuly: We move the mtp loss calculation from `multi_token_prediction.py` to the `gpt_model.py` because in this way we can separate the mtp process and mtp loss calculation, and put them in different vpp stages for balance consideration. This change would also have conflict with the MTP standalone design. Can we still maintain the original code structure and change it in `gpt_model.py`?

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -433,14 +434,18 @@ def __init__(
...
-        )
+        self.mtp_hybrid_override_pattern = mtp_hybrid_override_pattern
+
+        if hasattr(self.submodules.mtp_model_layer.submodules, 'attention_layer'):
+            self_attention_spec = self.submodules.mtp_model_layer.submodules.attention_layer
+            if self_attention_spec.submodules.self_attention is not None:
+                    self_attention_spec = self_attention_spec.submodules.self_attention
+                    attn_mask_type = self_attention_spec.params.get('attn_mask_type', '')
+                    assert attn_mask_type in SUPPORTED_ATTN_MASK, (
+                        f"Multi-Token Prediction (MTP) is not jet supported with "
```
BestJuly: `jet` -> `yet`

megatron/core/ssm/mamba_block.py
```diff
@@ -146,7 +149,7 @@ def __init__(
                 elif layer_type == LayerSymbols.MOE:
                     # Transformer layers apply their own pp_layer_offset
                     layer = build_module(
-                        submodules.moe_layer, config=self.config, layer_number=i + 1
+                        submodules.moe_layer, config=self.config, layer_number=i + 1,
```
BestJuly:
```suggestion
                        submodules.moe_layer, config=self.config, layer_number=i+1,
```

megatron/core/models/common/language_module/language_module.py
```diff
@@ -244,12 +247,20 @@ def setup_embeddings_and_output_layer(self) -> None:
             LanguageModule.embedding_warning_printed = True
 
     def shared_embedding_or_output_weight(self) -> Tensor:
-        """Gets the emedding weight or output logit weights when share embedding and output weights set to True.
+        """Gets the emedding weight or output logit weights when share embedding and output weights set to True
```
BestJuly: typo `emedding`

megatron/core/models/mamba/mamba_model.py
```diff
@@ -237,9 +352,53 @@ def forward(
...
+                loss_mask=loss_mask,
+                output_layer=self.output_layer,
+                output_weight=output_weight,
+                runtime_gather_output=runtime_gather_output,
+                is_training=self.training,
+                compute_language_model_loss=self.compute_language_model_loss,
+            )
+
         # logits and loss
         output_weight = None
```
BestJuly: This is duplicated with the newly added code in line 356-358

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -433,14 +434,18 @@ def __init__(
...
-        self_attention_spec = self.submodules.transformer_layer.submodules.self_attention
-        attn_mask_type = self_attention_spec.params.get('attn_mask_type', '')
-        assert attn_mask_type in SUPPORTED_ATTN_MASK, (
-            f"Multi-Token Prediction (MTP) is not jet supported with "
-            + f"{attn_mask_type} attention mask type."
-            + f"The supported attention mask types are {SUPPORTED_ATTN_MASK}."
-        )
+        self.mtp_hybrid_override_pattern = mtp_hybrid_override_pattern
+
+        if hasattr(self.submodules.mtp_model_layer.submodules, 'attention_layer'):
```
BestJuly: This would break the original code path for transformer_layer when the module does not have the attribute `attention_layer`.

megatron/core/pipeline_parallel/schedules.py
rkarimimahab: This is for a bug fix, I shared the PR internally.

megatron/core/models/mamba/mamba_layer_specs.py
rkarimimahab: Thanks! Good catch. now fixed in the updated version.

megatron/core/models/mamba/mamba_layer_specs.py
rkarimimahab: thanks for all great comments. Fixed in the updated version.

megatron/core/transformer/multi_token_prediction.py
rkarimimahab: Thank you. Fixed.

megatron/core/models/common/language_module/language_module.py
rkarimimahab: thank you fixed in updated MR.

megatron/core/models/mamba/mamba_layer_specs.py
rkarimimahab: thank you. Fixed.

megatron/core/models/mamba/mamba_model.py
rkarimimahab: @BestJuly thank you, I did not get which part is duplicate, do you mind sharing the lines? This line is `self.mtp.process_loss` which looks to be called once. thanks!

megatron/core/transformer/multi_token_prediction.py
rkarimimahab: Hi @BestJuly  we discussed it on the slack. I added you to that conversation now. 
So the issue with bringing codes inside gpt is that for any other model using MTP a large block of code should be copied. So the goal is to do it in a way we can have the benefits of separating the mtp process and mtp loss calculation, and also keep the codes encapsulated inside the MTP module so every new model using MTP could use the codes without having multiple copy of the codes.
So we decided to add a function for it but still call it inside the gpt model to have both sides benefits.

megatron/core/ssm/mamba_block.py
rkarimimahab: thanks I corrected the spacing issue.

megatron/core/transformer/multi_token_prediction.py
rkarimimahab: Thank you. this is now corrected to cover both.

megatron/core/models/mamba/mamba_model.py
BestJuly: [Here](https://github.com/NVIDIA/Megatron-LM/blob/a05cbed31b812d56436f72bc7910ad14c5e2aa35/megatron/core/models/mamba/mamba_model.py#L356-L358) and [here](https://github.com/NVIDIA/Megatron-LM/blob/a05cbed31b812d56436f72bc7910ad14c5e2aa35/megatron/core/models/mamba/mamba_model.py#L403-L405)

rkarimimahab: thanks so much catching this, fixed.

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -193,7 +193,7 @@ def set_current_microbatch(model, microbatch_id):
             layer.current_microbatch = microbatch_id
         if hasattr(model_with_decoder, 'mtp'):
             for layer in model_with_decoder.mtp.layers:
-                layer.transformer_layer.current_microbatch = microbatch_id
+                layer.mtp_model_layer.current_microbatch = microbatch_id
```
deepakn94, line 218: Let's add an assertion that layer has a `mtp_model_layer` attribute?

santhnm2 (approved):

tests/unit_tests/transformer/test_multi_token_prediction.py
rkarimimahab: I have added this test thank you!

shifangx:
This pr implements three features:
(1) supporting to use Mamba layer as  mtp_model_layer.
(2) splitting the MTP loss calculation in the GPT model’s forward pass into a separate function.
(3) supporting MTP layer repetition.
I have no concerns about the implementation of the first two features.
As for the last feature, we need to consider how to support scenarios where multiple MTP layers are placed on different PP ranks for computation. These MTP layers share the same parameters value, but be placed onto different vpp stage, in order to make sure each vpp stage do not have much computation task.
Maybe we can use assert to guide users can not place mtp layers into different vpp stage currently, and support this feature in future.

deepakn94:
> This pr implements three features: (1) supporting to use Mamba layer as mtp_model_layer. (2) splitting the MTP loss calculation in the GPT model’s forward pass into a separate function. (3) supporting MTP layer repetition. I have no concerns about the implementation of the first two features. As for the last feature, we need to consider how to support scenarios where multiple MTP layers are placed on different PP ranks for computation. These MTP layers share the same parameters value, but be placed onto different vpp stage, in order to make sure each vpp stage do not have much computation task. Maybe we can use assert to guide users can not place mtp layers into different vpp stage currently, and support this feature in future.

This is a good point. Let's go with the assertion for now re: 3.

rkarimimahab:
> This is a good point. Let's go with the assertion for now re: 3.

I added the assert, thanks!

BestJuly (approved):

megatron/core/models/mamba/mamba_model.py
```diff
@@ -63,6 +74,7 @@ def __init__(
         hybrid_attention_ratio: float = 0.0,
         hybrid_mlp_ratio: float = 0.0,
         hybrid_override_pattern: str = None,
+        mtp_hybrid_override_pattern: str = None,
```
duncanriach: I think this should be called `mtp_layer_pattern`. But why is is not part of the overall layer pattern?

megatron/core/models/mamba/mamba_model.py
```diff
@@ -74,6 +86,8 @@ def __init__(
         scatter_embedding_sequence_parallel: bool = True,
         seq_len_interpolation_factor: Optional[float] = None,
         pg_collection: Optional[ProcessGroupCollection] = None,
+        vp_stage: Optional[int] = None,
+        mtp_block_spec: Optional[ModuleSpec] = None,
```
duncanriach: Why do we have a separate spec for the MTP block? Why is this not part of the hybrid model spec?

megatron/core/models/mamba/mamba_model.py
```diff
@@ -74,6 +86,8 @@ def __init__(
         scatter_embedding_sequence_parallel: bool = True,
         seq_len_interpolation_factor: Optional[float] = None,
         pg_collection: Optional[ProcessGroupCollection] = None,
+        vp_stage: Optional[int] = None,
```
duncanriach, line 91: Why do we need this? Virtual pipeline parallel will be getting added for hybrid models when internal MR 2506 is merged into main, which will hopefully be soon.

megatron/core/models/mamba/mamba_model.py
```diff
@@ -87,17 +101,24 @@ def __init__(
...
+        # Cache for RoPE tensors which do not change between iterations.
+        self.rotary_pos_emb_cache = {}
+
 
         # megatron core pipelining currently depends on model type
         # TODO: remove this dependency ?
         self.model_type = ModelType.encoder_or_decoder
 
-        if self.pre_process:
+        if self.pre_process or self.mtp_process:
```
duncanriach: Why is the embedding layer included specifically if we have MTP?

megatron/core/models/mamba/mamba_model.py
```diff
@@ -129,8 +150,13 @@ def __init__(
             pg_collection=self.pg_collection,
         )
 
+        if self.mtp_process:
+            self.mtp = MultiTokenPredictionBlock(
+                config=self.config, spec=self.mtp_block_spec, vp_stage=vp_stage, mtp_hybrid_override_pattern=self.mtp_hybrid_override_pattern
+            )
```
duncanriach: Why is MTP not another layer type that just gets instantiated in the block above? Why is it treated specially/differently?

megatron/core/models/mamba/mamba_model.py
```diff
@@ -237,14 +352,54 @@ def forward(
...
+            packed_seq_params = None
+            decoder_input, rotary_pos_emb, rotary_pos_cos, rotary_pos_sin, sequence_len_offset = self._preprocess(
+                input_ids=input_ids,
+                position_ids=position_ids,
+                decoder_input=decoder_input,
+                inference_context=inference_context,
+                packed_seq_params=packed_seq_params,
+            )
+
+
```
duncanriach: There shouldn't be more than one line here.

megatron/core/models/mamba/mamba_model.py
```diff
@@ -237,14 +352,54 @@ def forward(
...
+                rotary_pos_emb=rotary_pos_emb,
+                rotary_pos_cos=rotary_pos_cos,
+                rotary_pos_sin=rotary_pos_sin,
+                packed_seq_params=packed_seq_params,
+                sequence_len_offset=sequence_len_offset,
+                embedding=self.embedding,
+                **(extra_block_kwargs or {}),
+            )
+
         if not self.post_process:
```
duncanriach: Is it possible for `self.post_process` to be `False` but `self.mtp_process` to be `True`? Not sure that this logic here aligns with the logic in the initializer.

megatron/core/models/mamba/mamba_model.py
```diff
@@ -168,6 +194,94 @@ def set_input_tensor(self, input_tensor: Tensor) -> None:
         assert len(input_tensor) == 1, 'input_tensor should only be length 1 for gpt/bert'
         self.decoder.set_input_tensor(input_tensor[0])
 
+    def _preprocess(
```
duncanriach: Is this an MTP-specific pre-process? Should MTP be in its name?

I'm also wondering if it would make the code easier to read if this was defined after `forward`.

megatron/training/arguments.py
```diff
@@ -599,7 +602,18 @@ def validate_args(args, defaults={}):
...
+        mtp_segments = len(args.mtp_hybrid_override_pattern.split('|'))
+        expected_segments = args.pipeline_model_parallel_size * (
+            args.virtual_pipeline_model_parallel_size or 1
+        )
+        assert mtp_segments == expected_segments, (
+            f'MTP hybrid pattern has {mtp_segments} segments but expected {expected_segments} '
+            f'(PP={args.pipeline_model_parallel_size} * VP={args.virtual_pipeline_model_parallel_size or 1}). '
+            f'Use: {"|||" * (expected_segments - 1)}*E'
+        )
+
```
duncanriach: You've implemented some kind of virtual pipeline parallel support, but only for the MTP layers? When we planned out the layer syntax, the MTP layers were intended to be part of the layer pattern for the whole model.

megatron/training/arguments.py
```diff
@@ -1782,6 +1796,12 @@ def _add_network_size_args(parser):
                        'We compute the average of the MTP losses across all depths, '
                        'and multiply it the scaling factor to obtain the overall MTP loss, '
                        'which serves as an additional training objective.')
+    group.add_argument('--mtp-use-repeated-layer', action='store_true',
+                       help='Use a single MTP layer repeatedly instead of multiple separate layers. '
+                       'This is more parameter-efficient. When enabled, only 1 MTP layer is created '
+                       'and applied --mtp-num-layers times.')
+    group.add_argument('--mtp-num-layers-per-layer', type=int, default=None,
+                       help='Number of layers inside each MTP layer (e.g., for hybrid models).')
```
duncanriach: These arguments seem to be specifying something with additional command-line arguments that could be defined in the model's single layer pattern

megatron/training/arguments.py
```diff
@@ -3371,6 +3391,19 @@ def _add_experimental_args(parser):
                             'hybrid ratio arguments, then the number of each type'
                             'of layer in the override pattern must match number in'
                             'the overidden pattern')
+    group.add_argument("--mtp-hybrid-override-pattern", type=str, default=None,
+                       help='Force a specific hybrid layer pattern for MTP layers. '
+                       'If a value greater than 0.0 is supplied to any of the hybrid ratio '
+                       'arguments, then the number of each type of layer in the '
+                       'override pattern must match number in the overidden '
+                       'pattern')
```
duncanriach: I think we need to discuss this. The direction we're heading in, this would be part of `--hybrid-override-pattern` (which will become, or be replaced by, `--hybrid-layer-pattern`).

megatron/core/models/mamba/mamba_model.py
```diff
@@ -87,17 +101,24 @@ def __init__(
...
         self.hybrid_mlp_ratio = hybrid_mlp_ratio
         self.hybrid_override_pattern = hybrid_override_pattern
+        self.mtp_hybrid_override_pattern = mtp_hybrid_override_pattern
         self.post_process = post_process
         self.fp16_lm_cross_entropy = fp16_lm_cross_entropy
         self.parallel_output = parallel_output
         self.share_embeddings_and_output_weights = share_embeddings_and_output_weights
         self.position_embedding_type = position_embedding_type
+        self.mtp_block_spec = mtp_block_spec
+        self.mtp_process = mtp_block_spec is not None
```
duncanriach: Can we enable MTP if `self.mtp_hybrid_override_pattern` is not `None`?

megatron/core/models/mamba/mamba_model.py
duncanriach, line 322: Is it possible for `self.post_process` to be `False` but `self.mtp_process` to be `True`. In that case would the functionality here be correct?

megatron/core/models/mamba/mamba_model.py
```diff
@@ -287,3 +442,67 @@ def forward(
         loss = self.compute_language_model_loss(labels, logits)
 
         return loss
+
+    def shared_embedding_or_output_weight(self) -> Optional[Tensor]:
```
duncanriach: Should this be a private method of the class?

megatron/core/models/mamba/mamba_model.py
```diff
@@ -287,3 +442,67 @@ def forward(
...
+        """Gets the embedding weight or output logit weights when share input embedding and
+        output weights set to True or when use Multi-Token Prediction (MTP) feature.
```
duncanriach: The switch you're talking about is not in this function. Should it be in this function?

megatron/core/models/mamba/mamba_model.py
```diff
@@ -287,3 +442,67 @@ def forward(
...
+
+        Returns:
+            Optional[Tensor]: During pre processing or MTP process it returns the input embeddings weight.
+            Otherwise, during post processing it returns the final output layers weight.
```
duncanriach: What happens if the current pipeline stage has both pre-processing and post-processing?

Is this code part of the MTP solution to tie all the embeddings together between the main model and the MTP modules, and to independently tie all the output weights together between the main model and the MTP modules?

megatron/core/models/mamba/mamba_model.py
```diff
@@ -287,3 +442,67 @@ def forward(
...
+        """
+        if self.pre_process or self.mtp_process:
+            # Multi-Token Prediction (MTP) need both embedding layer and output layer.
+            # So there will be both embedding layer and output layer in the mtp process stage.
+            # In this case, if share_embeddings_and_output_weights is True, the shared weights
+            # will be stored in embedding layer, and output layer will not have any weight.
```
duncanriach: I think it will be less confusing to bring down the  if `self.share_embeddings_and_output_weights` conditional because you seem to be referring to a decision made outside this function. In this comment are you trying to express what will happen with both MTP and `share_embeddings_and_output_weights==True`? I.e. all of the embedding weights and all of the output weights in the model will be the same weights.

megatron/core/ssm/mamba_block.py
```diff
@@ -83,12 +86,17 @@ def __init__(
         device=None,
         dtype=None,
         pg_collection: ProcessGroupCollection = None,
+        vp_stage: Optional[int] = None,
+        num_layers: int = None,
```
duncanriach: Assuming this is the number of layers in this block, why is this number needed? Isn't it inferred by the pattern?

megatron/core/ssm/mamba_block.py
```diff
@@ -83,12 +86,17 @@ def __init__(
         device=None,
         dtype=None,
         pg_collection: ProcessGroupCollection = None,
+        vp_stage: Optional[int] = None,
```
duncanriach: Why are passing in vp_stage? This is a feature that is queued to be implemented elsewhere, and it doesn't require passing in the vp_stage

megatron/core/ssm/mamba_block.py
```diff
@@ -83,12 +86,17 @@ def __init__(
...
+        num_layers: int = None,
+        is_mtp_layer: bool = False,
```
duncanriach, line 89: Why does this block/stack need to know that it's an MTP block? Also, it's not a layer in the sense that blocks are made of layers in the terminology we're using here.

Does `is_mtp_layer` mean that this stack of layers is inside an MTP module? If so, why does that make a difference? Is it so that the weights can be shared between MTP modules?

megatron/core/ssm/mamba_block.py
duncanriach, line 48: Is this now being used not only for the regular stack of layers, also known as the decoder block, but also for the stack of layers inside an MTP module?

megatron/core/ssm/mamba_block.py
```diff
@@ -103,7 +111,7 @@ def __init__(
         self.hybrid_override_pattern = hybrid_override_pattern
 
         self.layer_type_list = allocate_layers(
-            self.config.num_layers,
+            self.config.num_layers if num_layers is None else num_layers,
             self.hybrid_attention_ratio,
             self.hybrid_mlp_ratio,
             self.hybrid_override_pattern,
```
duncanriach, line 125: Allocate layers converts from the pattern to a list of layers to implement. I'm wondering if this should be moved up to the model level and then it can provide a list of layers for the decoder and also for the MTP module(s)

megatron/core/ssm/mamba_block.py
```diff
@@ -147,7 +156,11 @@ def __init__(
                 elif layer_type == LayerSymbols.MOE:
                     # Transformer layers apply their own pp_layer_offset
                     layer = build_module(
-                        submodules.moe_layer, config=self.config, layer_number=i + 1
+                        submodules.moe_layer,
+                        config=self.config,
+                        layer_number=i + 1,
+                        pg_collection=pg_collection,
+                        is_mtp_layer=is_mtp_layer,
```
duncanriach: Why do these layers need to know that they're MTP layers? Is it so that weights can be shared between MTP modules?

megatron/core/ssm/mamba_block.py
```diff
@@ -287,6 +300,15 @@ def forward(
...
+                        hidden_states = layer(
+                            hidden_states=hidden_states,
+                            attention_mask=attention_mask,
+                            inference_context=inference_context,
+                            rotary_pos_emb=rotary_pos_emb,
+                            sequence_len_offset=sequence_len_offset,
+                        )
                     else:  # MambaLayer
                         hidden_states = layer(
                             hidden_states=hidden_states,
```
duncanriach: What is this block of code here for and how does it work? I don't understand how that elif is reached. It doesn't even seem like valid Python code

megatron/training/arguments.py
duncanriach: What does this argument specify? If it's the number of layers in the pattern, is that not clear from the pattern?

megatron/training/arguments.py
```diff
@@ -1782,6 +1796,12 @@ def _add_network_size_args(parser):
                        'We compute the average of the MTP losses across all depths, '
                        'and multiply it the scaling factor to obtain the overall MTP loss, '
                        'which serves as an additional training objective.')
+    group.add_argument('--mtp-use-repeated-layer', action='store_true',
+                       help='Use a single MTP layer repeatedly instead of multiple separate layers. '
+                       'This is more parameter-efficient. When enabled, only 1 MTP layer is created '
+                       'and applied --mtp-num-layers times.')
```
duncanriach: Do you mean that in a single MTP module, layers of the same type have shared weights?
Or do you mean that when there are multiple MTP modules, any given layer in any of the modules is using the same weights?

Looking at the description of `--mtp-num-layers` above, it seems that this refers to the number of MTP modules, each module containing a number of layers.

So we have `--mtp-num-layers` to choose the number of MTP modules, and then `--mtp-use-repeated-layer` to make them share weights.

Ideally we would just define the number of MTP modules, and the layers that are in them in the single layer pattern string for the model.

megatron/core/pipeline_parallel/schedules.py
duncanriach, line 1: Why does schedules.py need to know about the model having MTP?

megatron/core/models/mamba/mamba_layer_specs.py
```diff
@@ -146,3 +171,65 @@
...
+    if len(mtp_layer_specs) > 0:
+        expected_num_layers = 1 if config.mtp_use_repeated_layer else config.mtp_num_layers
+        assert (
+            len(mtp_layer_specs) == expected_num_layers
+        ), f"currently all of the mtp layers must stage in the same pipeline stage. Expected {expected_num_layers}, got {len(mtp_layer_specs)}"
+        mtp_block_spec = MultiTokenPredictionBlockSubmodules(layer_specs=mtp_layer_specs)
+    else:
+        mtp_block_spec = None
+
+    return mtp_block_spec
```
duncanriach: This whole section is concerning to me. It's very hard to understand what's going on here. Also we have vp_stage showing up in the spec. I don't think that should be happening. The init of the model itself needs to check what vp stage it's on so it knows which part of the layer pattern to instantiate, but it shouldn't be needed in the spec.

This reminds me of how the GPT model is constructed, with complex conditional code in the spec file. With the hybrid model, we moved away from this kind of complexity. You specify the layers in the spec and then you stamp them out according to the layer pattern. That's it

megatron/core/models/mamba/mamba_model.py
duncanriach: Also, how does this function relate to the one in language_module.py. It looks like the same code, and you modified that function there as well.

megatron/core/models/common/language_module/language_module.py
```diff
@@ -250,12 +254,20 @@ def setup_embeddings_and_output_layer(self) -> None:
...
         Returns:
-            Tensor: During pre processing it returns the input embeddings weight while during post processing it returns the final output layers weight
+            Tensor: During pre processing or MTP process it returns the input embeddings weight while during post processing it returns the final output layers weight
         """
-        if self.pre_process:
+        if self.pre_process or self.mtp_process:
+            # Multi-Token Prediction (MTP) need both embedding layer and output layer.
+            # So there will be both embedding layer and output layer in the mtp process stage.
+            # In this case, if share_embeddings_and_output_weights is True, the shared weights
+            # will be stored in embedding layer, and output layer will not have any weight.
```
duncanriach: I don't understand what this comment is trying to convey. Is this saying that if `share_embeddings_and_output_weights==True` in MTP modules, it will function as if that setting is `False`

megatron/core/models/mamba/mamba_model.py
duncanriach, line 179: So this block encapsulates and handles however many MTP modules we have (which could be specified in the model layer pattern) and also instantiates the hybrid layer blocks (MambaStacks) inside each. And ties the weights as necessary and appropriately. Correct?

So from a configuration standpoint, there is either MTP or not, a specific number of identical MTP modules, a layer pattern that is the same in all the MTP modules, and the modules potentially share weights. Is that correct?

This is the kind of high-level info that would be useful in a comment.

megatron/core/models/mamba/mamba_layer_specs.py
duncanriach: Looking at this again. It's really not good that layer stacking is being done here. The design pattern we have in the hybrid model is a simple/single layer pattern that gets unrolled and used to stamp out modules that have been specified non-dynamically in the spec file.

mamba_builders.py
```diff
@@ -22,6 +23,15 @@ def mamba_builder(args, pre_process, post_process, vp_stage=None, config=None, p
...
 
+    mtp_block_spec = None
+    if args.mtp_num_layers is not None:
+        use_te = args.transformer_impl == "transformer_engine"
+        assert args.mtp_spec is not None
+        assert args.mtp_hybrid_override_pattern is not None, "We need to set the override hybrid pattern for MTP layers!"
+        mtp_mamba_stack_spec = import_module(args.mtp_spec)
+        mtp_block_spec = get_mamba_mtp_block_spec(
+            config, mtp_mamba_stack_spec, use_transformer_engine=use_te, vp_stage=vp_stage
+        )
```
duncanriach: We should not have an additional command line argument to load in the MTP spec, but which in fact loads the hybrid spec. We don't need another command line argument.

Also the MambaModel itself can pull the MTP spec out of the hybrid model spec itself.

This whole block of code should not be necessary. If MTP is specified in the layer pattern, then the model should create an MTP module with an appropriate number of sub-modules, each containing the layers specified in the layer pattern. Any spec information needed for that build can be pulled out of the hybrid model spec and passed into the MTP module.

megatron/core/models/mamba/mamba_layer_specs.py
```diff
@@ -146,3 +171,65 @@
...
+
+    mtp_layer_spec = get_mtp_layer_spec_for_backend(
+        mtp_model_layer_spec=spec, backend=backend
+    )
+    mtp_num_layers = config.mtp_num_layers if config.mtp_num_layers else 0
+
+    if config.mtp_use_repeated_layer:
+        mtp_layer_specs = [mtp_layer_spec]
+    else:
+        mtp_layer_specs = [mtp_layer_spec] * mtp_num_layers
```
duncanriach: So with repeated layers, there really are just the same module re-used again and again. Okay, so instead of the data passing through multiple modules, it gets recycled through the same module. Correct?

It sounds like there's an assumption that the whole MTP process cannot be split over multiple pipeline stages. Is that correct?

megatron/training/arguments.py
```diff
@@ -3371,6 +3391,19 @@ def _add_experimental_args(parser):
...
+                       'arguments, then the number of each type of layer in the '
+                       'override pattern must match number in the overidden '
+                       'pattern')
+    group.add_argument('--mtp-spec', type=str, default=None, nargs='*',
+                       help='Specify the <module_location function_name> pair '
+                       'that returns a spec for mtp layer to customize a model,' 
+                       'transformer block, or transformer layer, depending on '
+                       'the use case. To use local spec specify local as the argument.'
+                       'For more details, see the model class, '
+                       'TransformerBlock class, or TransformerLayer class.')
```
duncanriach: This should not be needed. It's already included in the hybrid spec

megatron/training/checkpointing.py
```diff
@@ -1347,6 +1347,7 @@ def _set_arg(arg_name, old_arg_name=None, force=False):
     _set_arg('hidden_dropout', force=True)
 
     _set_arg('hybrid_override_pattern', force=True)
+    _set_arg('mtp_hybrid_override_pattern', force=True)
```
duncanriach, line 1423: I'm not happy with these additional arguments, but aren't we missing some from here?

megatron/training/training.py
```diff
@@ -178,6 +178,11 @@ def calculate_layer_counts():
             for layer_type in args.hybrid_override_pattern:
                 if layer_type in counts:
                     counts[layer_type] += 1
+            if args.mtp_hybrid_override_pattern:
+                assert args.mtp_num_layers is not None
+                for layer_type in args.mtp_hybrid_override_pattern:
+                    if layer_type in counts:
+                        counts[layer_type] += args.mtp_num_layers
```
duncanriach: In an ideal world, the number of MTP modules (each containing a number of model layers) would be easily countable from the `hybrid_override_pattern`, and that process would be encapsulated inside a function in `core/ssm`.

duncanriach (commented):
This is my first-pass review. I see a lot of issues. I don't know how to resolve all of this but I think it's good to start the discussion.

megatron/core/models/mamba/mamba_layer_specs.py
duncanriach: We should not have this logic here in the spec. It should be in the instantiation code for the model

megatron/core/models/mamba/mamba_model.py
sancha: we need this only for passing it into the multi_token_prediction module which already has support for this, we can bring this change later as well.

megatron/core/models/mamba/mamba_model.py
sancha: This is correct, except its no longer MambaStacks anymore. This is documented in the docstring of the MultiTokenPredictionBlock class

megatron/core/models/mamba/mamba_model.py
sancha: no, mtp-process will only be true in the post-process stage.

duncanriach:
Temporarily closing, to see if it resolves issues.

duncanriach:
Opening again. Hopefully this fixes the issue.

megatron/core/models/mamba/mamba_model.py
duncanriach: Presumably the MTP module just needs it to error check that it not in a VPP stage other than the last one.

megatron/core/models/mamba/mamba_model.py
duncanriach: So MTP is a special kind, or mode or version, of post processing

megatron/training/arguments.py
```diff
@@ -537,6 +537,23 @@ def validate_args(args, defaults={}):
                 args.global_batch_size), flush=True)
     assert args.global_batch_size > 0
 
+    # Deprecation warnings for legacy MTP arguments
+    if args.mtp_hybrid_override_pattern is not None:
+        warn_rank_0(
+            "--mtp-hybrid-override-pattern is deprecated. "
+            "For new Mamba+MTP models, use unified --hybrid-override-pattern instead. "
```
duncanriach: For new hybrid models with MTP

megatron/training/arguments.py
```diff
@@ -537,6 +537,23 @@ def validate_args(args, defaults={}):
...
+            "Example: 'M*M*/MM/MM' means main='M*M*', MTP pattern='MM' with 2 depths. "
+            "This argument is kept only for loading old checkpoints.",
+            args.rank,
+        )
+    if args.mtp_spec is not None:
+        warn_rank_0(
+            "--mtp-spec is deprecated. "
+            "For new Mamba+MTP models, use unified --hybrid-override-pattern instead. "
```
duncanriach: For new hybrid models with MTP

megatron/training/arguments.py
```diff
@@ -1262,13 +1279,56 @@ def validate_args(args, defaults={}):
...
 
+    # For Mamba/hybrid models with unified pattern, infer mtp_num_layers from pattern
+    if args.hybrid_override_pattern and '/' in args.hybrid_override_pattern:
+        from megatron.core.ssm.mamba_hybrid_layer_allocation import parse_hybrid_pattern
+        parsed = parse_hybrid_pattern(args.hybrid_override_pattern)
+        if parsed.mtp_pattern and parsed.mtp_num_depths > 0:
+            inferred_mtp_num_layers = parsed.mtp_num_depths
+            if args.mtp_num_layers is None:
+                args.mtp_num_layers = inferred_mtp_num_layers
+            elif args.mtp_num_layers != inferred_mtp_num_layers:
```
duncanriach: I think it would be better in this case to simply require that `--mtp-num-layers` is not specified. It should not be getting used at all. Allowing it to be used, but then error checking that it matches what it should be is not nipping this explosion of redundant args in the bud.

megatron/training/arguments.py
```diff
@@ -3423,6 +3487,18 @@ def _add_experimental_args(parser):
                             'hybrid ratio arguments, then the number of each type'
                             'of layer in the override pattern must match number in'
                             'the overidden pattern')
+    # DEPRECATED: Legacy MTP arguments kept for checkpoint compatibility.
+    # For new Mamba+MTP models, use unified pattern in --hybrid-override-pattern instead.
```
duncanriach: new hybrid models with MTP

megatron/core/models/mamba/mamba_layer_specs.py
```diff
@@ -146,3 +153,51 @@
...
+        >>> mtp_block = MultiTokenPredictionBlock(
+        ...     config=config,
+        ...     spec=mtp_spec,
+        ...     mtp_layer_pattern="MM",
+        ...     mtp_num_depths=2,
+        ...     mamba_submodules=mamba_stack_spec.submodules,
+        ... )
+    """
+    norm = TENorm if use_te else TENorm  # Fallback to TENorm for now
+    linear = TELayerNormColumnParallelLinear if use_te else TELayerNormColumnParallelLinear
```
duncanriach: Why include use_te as an argument when it doesn't do anything? Shouldn't that all just be stripped out? The spec as-is requires TE. I think that's also currently/still true for the main body of the hybrid model.

megatron/core/ssm/layer_builder.py
```diff
@@ -0,0 +1,139 @@
+# Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
```
duncanriach: 2026?

megatron/core/ssm/layer_builder.py
```diff
@@ -0,0 +1,139 @@
...
+
+    Example:
+        >>> submodules = mamba_stack_spec.submodules
+        >>> layers = build_layers_from_pattern("M*M", submodules, config, pg_collection)
+        >>> # Returns [MambaLayer, TransformerLayer, MambaLayer]
+    """
+    layers = nn.ModuleList()
+
+    for i, layer_type in enumerate(pattern):
+        layer_number = i + 1 + layer_offset
```
duncanriach: Please don't create this variable and then use it only once. It adds confusion because `layer_number` currently means different things for the mamba layer versus other layers.

megatron/core/models/mamba/mamba_model.py
duncanriach: Note to self: `vp_stage` was added to `GPTModel` only to pass to the MTP logic. This is why we have it here now in the hybrid model. This will probably also be used for implementing flexible virtual pipeline parallel for the hybrid model, rather than getting it from `parallel_state` (as in internal 2506), which doesn't seem to be compatible with the new, decoupled process groups approach.

megatron/core/ssm/mamba_hybrid_layer_allocation.py
```diff
@@ -29,9 +30,129 @@ class Symbols:
...
+        mtp_pattern: The MTP layer pattern per depth (e.g., "MM"), or None if no MTP
+        mtp_num_depths: Number of MTP prediction depths (0 if no MTP)
+    """
+
+    main_pattern: Optional[str]
+    mtp_pattern: Optional[str]
+    mtp_num_depths: int
+
+
+def parse_hybrid_pattern(pattern: Optional[str]) -> ParsedHybridPattern:
```
duncanriach, line 65: This is surprisingly complex function given what it does.

megatron/core/models/mamba/mamba_model.py
sancha: yes, we have this assertion.

megatron/core/models/mamba/mamba_model.py
sancha: yes, we could say that

shifangx:
With main branch, for mtp_num_layers=2, we can make each mtp layer into a standalone vpp stage, such as the following setting:

case-1 with vpp
```
INFO:megatron.core.transformer.pipeline_parallel_layer_layout:Parse pipeline model parallel layout Ett|(t|)*4m|m|L to:
                   VPP rank 0                 VPP rank 1
PP rank 0          embedding,decoder*2        decoder
PP rank 1-2        decoder                    mtp
PP rank 3          decoder                    loss
```

case-2 without vpp
```
INFO:megatron.core.transformer.pipeline_parallel_layer_layout:Parse pipeline model parallel layout Ett|(t|)*4m|m|L to:
                     VPP rank 0               
PP rank 0          embedding,decoder*2 
PP rank 1-4        decoder 
PP rank 5-6        mtp 
PP rank 7           loss
```


@rkarimimahab can you test and make sure this feature can works well after this pr?

duncanriach:
> With main branch, for mtp_num_layers=2, we can make each mtp layer into a standalone vpp stage, such as the following setting

Note that we're not supporting flexible VPP in the hybrid model using the same mechanism as the GPT model. Once this current PR is merged, we'll be adding a unified hybrid-layer-pattern containing the pipes, which will allow VPP for the main part of the model and the MTP. I have made a note to revisit this discussion when we port those changes to main.

Regarding MTP layers being in different pipeline stages. How does this work when the layers are repeated (including sharing the same weights)?

megatron/core/models/mamba/mamba_model.py
```diff
@@ -134,8 +157,25 @@ def __init__(
             pg_collection=self.pg_collection,
         )
 
+        # MTP block (MambaModel creates it, MTP builds its own inner layers)
+        if self.mtp_process:
+            from megatron.core.models.mamba.mamba_layer_specs import get_mamba_mtp_block_spec
```
duncanriach: There's a potential issue here. I'm not sure whether or how it should be resolved.

The pattern we had before adding MTP was basically that we can either pass in a spec for the hybrid model on the command line (using the spec argument), or `mamba_builders.py` would use the default. This is also how the GPT model works, and it makes sense and is simple. Ideally there would be one spec for the whole model, and then the relevant parts would be pulled from it, as needed, and passed down for sub-module building: (1) the sub-spec for the decoder, (2) the sub-spec for MTP. I guess there is somewhat of a legacy issue in that what gets passed in on the command line is currently called the "stack" spec, but the intention is really for that to to be the spec for the whole model. If the spec passed in on the command-line, or passed down to the model, covers more than just the "stack" (or decoder) then it should perhaps have a broader name (e.g. model_spec).

The way this is right now, the part of the model spec for the MTP layers is hard-coded in the top level of the hybrid model. This is not really how it should be.

However, if what the argument passed to `--spec` represents changes in scope, then it would presumably result in a backward-compatibility issue with respect to existing scripts and checkpoints.

Note that our scripts have been specifying `--spec` for a long time, but we've had the intention of making that redundant (by adding the default option in `mamba_builders.py`), which was added recently.

duncanriach (commented):
This comment seems significant enough that I want to submit it immediately.

megatron/core/models/mamba/mamba_model.py
duncanriach: As it stands. This seems kind of hacked-in.

Perhaps in the spec file, mamba_stack_spec, should contain the spec info for both the hybrid layer types and also the MTP layer type (and any other spec info needed for the model). Perhaps that name needs to be there for command-line backwards compatibility, but going into the model, and in mamba_builders.py, it can be called model_spec. Perhaps also in the spec file mamba_stack_spec could be aliased to model_spec, so either could be used when referencing the model spec. Perhaps there could be a note or a deprecation warning if the symbol mamba_stack_spec was used.

I also notice that there is an MoE spec thing outside the model spec object in the hybrid spec file, which is probably not correct/ideal. I must have let that slip through, unfortunately.

megatron/core/ssm/mamba_block.py
```diff
@@ -102,57 +108,36 @@ def __init__(
         self.hybrid_attention_ratio = hybrid_attention_ratio
         self.hybrid_mlp_ratio = hybrid_mlp_ratio
         self.hybrid_override_pattern = hybrid_override_pattern
+        self.pg_collection = pg_collection
+
+        # Parse pattern - extract only main pattern, MTP handled by MambaModel
+        parsed = parse_hybrid_pattern(hybrid_override_pattern)
+        main_pattern = parsed.main_pattern  # MTP pattern ignored here
```
duncanriach: Why parse this out again here? This can be done up at the model level (at the same time as for the MTP) and the pattern for the decoder can be passed down into the decoder as hybrid_override_pattern

megatron/core/ssm/mamba_block.py
```diff
@@ -102,57 +108,36 @@ def __init__(
...
 
         self.layer_type_list = allocate_layers(
-            self.config.num_layers,
+            self.config.num_layers if self.config.num_layers is not None else len(main_pattern),
```
duncanriach: Not sure that this forcing of num_layers is, or should be, needed here.

megatron/core/models/mamba/mamba_layer_specs.py
```diff
@@ -146,3 +153,51 @@
...
+
+    return ModuleSpec(
```
duncanriach: I think this should be another item in what is now called mamba_stack_spec, which is really the model_spec (and it's name should be updated at some point). Perhaps the key should be `mtp_layer`.

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -518,6 +525,41 @@ def get_mtp_layer_offset(config: TransformerConfig, vp_stage: Optional[int] = No
...
 
+class MTPModelLayerContainer(MegatronModule):
+    """Container for MTP inner layers built from a pattern.
+
+    This container wraps the inner layers (e.g., Mamba, Attention layers) built
+    from a pattern string. It provides proper state dict key structure:
+    `mtp_model_layer.layers.{i}.*` for checkpoint compatibility.
+
+    This is used when MTP builds its own layers using the shared layer_builder,
+    instead of receiving a pre-built MambaStack.
```
duncanriach: MTP can still be built by being passed an instance of MambaStack?

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -658,17 +721,38 @@ def __init__(
...
-        )
-        self.transformer_layer = build_module(
-            self.submodules.transformer_layer,
-            config=self.config,
-            vp_stage=vp_stage,
-            layer_number=self.layer_number + diff_transformer_layer_offset,
-        )
+        # Build inner layers: three possible paths
+        # 1. New Mamba path: build from pattern using shared layer_builder
+        # 2. Legacy Mamba path: build MambaStack with mtp_layer_pattern
```
duncanriach: Why does this path #2 need to be preserved? I also don't see this path in the code below.

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -788,19 +871,30 @@ def _proj_and_transformer_layer(
...
-                    rotary_pos_emb=rotary_pos_emb,
-                    rotary_pos_cos=rotary_pos_cos,
-                    rotary_pos_sin=rotary_pos_sin,
-                    attention_bias=attention_bias,
-                    inference_params=inference_params,
-                    packed_seq_params=packed_seq_params,
-                    sequence_len_offset=sequence_len_offset,
-                )
+                if self.mtp_layer_pattern is not None:
+                    # New Mamba path: run through pattern-based inner layers
```
duncanriach: hybrid path?

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -658,17 +721,38 @@ def __init__(
...
-            self.config, vp_stage
-        )
-        self.transformer_layer = build_module(
-            self.submodules.transformer_layer,
-            config=self.config,
-            vp_stage=vp_stage,
-            layer_number=self.layer_number + diff_transformer_layer_offset,
-        )
+        # Build inner layers: three possible paths
+        # 1. New Mamba path: build from pattern using shared layer_builder
```
duncanriach: hybrid path?

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -820,6 +914,52 @@ def _postprocess(self, hidden_states: torch.Tensor):
...
 
+    def _run_pattern_layers(
+        self,
+        hidden_states: torch.Tensor,
+        attention_mask: Optional[torch.Tensor] = None,
+        rotary_pos_emb: Optional[torch.Tensor] = None,
+        inference_params: Optional[InferenceParams] = None,
+        packed_seq_params: Optional[PackedSeqParams] = None,
+    ) -> torch.Tensor:
+        """Run through pattern-based inner layers (Mamba path).
```
duncanriach: hybrid path

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -820,6 +914,52 @@ def _postprocess(self, hidden_states: torch.Tensor):
 
         return hidden_states
 
+    def _run_pattern_layers(
```
duncanriach: Should this function be over in the ssm folder?

Should it also be called by MambaStack forward?

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -1064,15 +1237,64 @@ def build_layer(layer_spec, layer_number):
...
             return module
 
-        self.layers = torch.nn.ModuleList(
-            [
-                build_layer(layer_spec, i + 1)
-                for i, layer_spec in enumerate(self.submodules.layer_specs)
-            ]
-        )
+        def build_layer_with_pattern(layer_spec, layer_number, mtp_layer_pattern, mamba_submodules):
+            """Build layer using pattern-based approach (new Mamba path)."""
```
duncanriach, line 1227: hybrid path. The hybrid model started out as a mamba model, but soon became a hybrid model. The word mamba is often found in places where it should really say hybrid. I think it's preferable to refer to hybrid where possible so as not to keep propagating an out-of-date symbol

megatron/training/training.py
```diff
@@ -170,21 +198,34 @@ def destroy_global_state():
...
+            if parsed.main_pattern:
+                for layer_type in parsed.main_pattern:
+                    if layer_type in counts:
+                        counts[layer_type] += 1
+            # Count MTP layers (pattern repeated mtp_num_depths times)
+            if parsed.mtp_pattern and parsed.mtp_num_depths > 0:
+                for layer_type in parsed.mtp_pattern:
+                    if layer_type in counts:
+                        counts[layer_type] += parsed.mtp_num_depths
             return counts['*'], counts['M'], counts['-'], counts['E']
```
duncanriach, line 237: I'm not sure this should be addressed in this PR, but I wanted to note that all this hybris-layer-type-specific knowledge, including the independent hard-coding of symbols used for layer types, should really be abstracted into a function in the ssm directory (file currently called mamba_hybrid_layer_allocation.py) that could be called from here.

There was a small amount of code here, but the PR grows it in amount and complexity.

megatron/core/transformer/moe/router.py
```diff
@@ -453,11 +469,22 @@ def apply_z_loss(self, logits):
...
+            # accumulating the z_loss multiple times, we scale it by 1/mtp_num_layers so the total loss is correct.
+            if self.is_mtp_layer and self.config.mtp_use_repeated_layer and self.config.mtp_num_layers is not None:
+                z_loss = z_loss / self.config.mtp_num_layers
+
             num_layers = self.config.num_layers
             if self.config.mtp_num_layers is not None:
                 num_layers += self.config.mtp_num_layers
+
+            if self.is_mtp_layer:
+                layer_number = self.layer_number + self.config.num_layers
```
duncanriach, line 542: Would it make sense to number the layers in the MTP layers sequentially after the decoder layers?

duncanriach (commented):
Some more comments

megatron/core/models/mamba/mamba_layer_specs.py
yfw: Should this be `TEColumnParallelLinear` instead of `TELayerNormColumnParallelLinear`? From my understanding, `eh_proj` seems to be a simple linear projection without layernorm.

sancha: you are right, thanks for catching this 🙇

sancha:
> With main branch, for mtp_num_layers=2, we can make each mtp layer into a standalone vpp stage, such as the following setting:
> 
> case-1 with vpp
> 
> ```
> INFO:megatron.core.transformer.pipeline_parallel_layer_layout:Parse pipeline model parallel layout Ett|(t|)*4m|m|L to:
>                    VPP rank 0                 VPP rank 1
> PP rank 0          embedding,decoder*2        decoder
> PP rank 1-2        decoder                    mtp
> PP rank 3          decoder                    loss
> ```
> 
> case-2 without vpp
> 
> ```
> INFO:megatron.core.transformer.pipeline_parallel_layer_layout:Parse pipeline model parallel layout Ett|(t|)*4m|m|L to:
>                      VPP rank 0               
> PP rank 0          embedding,decoder*2 
> PP rank 1-4        decoder 
> PP rank 5-6        mtp 
> PP rank 7           loss
> ```
> 
> @rkarimimahab can you test and make sure this feature can works well after this pr?

Hi @shifangx thanks for question, I tried this and see that there is an assert in [`megatron/core/transformer/pipeline_parallel_layer_layout.py`](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/pipeline_parallel_layer_layout.py#L130) which forbids MTP modules from residing on different VPP stages, so neither of these configs would be supported. I have further added an assert that it needs to be on the last stage, is this an issue?

megatron/core/transformer/moe/router.py
sancha: this will complicate things it looks like, the router is the only place which needs a global index, so removing this logic will have it show up elsewhere ..

megatron/core/ssm/mamba_block.py
duncanriach, line 127: If in MTP context, switch off the logging inside here.

duncanriach: Can we remove this layer of abstraction. It's not needed anymore, and complicates the design

duncanriach: This code should be bypassed in the MTP context.

megatron/core/ssm/mamba_block.py
```diff
@@ -149,7 +160,7 @@ def __init__(
             )
 
     def _select_layers_for_pipeline_parallel(self, layer_type_list):
-        num_layers_per_pipeline_rank = self.config.num_layers // self.pp_group.size()
+        num_layers_per_pipeline_rank = len(layer_type_list) // self.pp_group.size()
```
duncanriach: Revert this change. It's not relevant to MTP and is incorrect for the decoder. The way the hybrid stuff works at the moment, self.config.num_layers should be used. I can explain in more detail why if you want

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -795,21 +762,22 @@ def __init__(
         )
 
         # Build inner layers: two possible paths
-        # 1. New Mamba path: build from pattern using shared layer_builder
+        # 1. Mamba path: use MambaStack for hybrid pattern support (and future PP)
```
duncanriach: I don't understand what you mean by "(and future PP)." I think best to remove

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -795,21 +762,22 @@ def __init__(
...
-                config=config,
+                hybrid_override_pattern=mtp_layer_pattern,
+                pre_process=True,       # Always receives input from eh_proj
+                post_layer_norm=False,  # MTP has its own final_layernorm
+                post_process=True,      # MTP layer is self-contained
                 pg_collection=pg_collection,
-                layer_offset=0,
-                is_mtp_layer=True,
+                vp_stage=vp_stage,
+                is_mtp_layer=True,      # Mark as MTP for router aux loss scaling
```
duncanriach: Is this comment correct? I don't know what router aux loss scaling is, but does it occur inside here? We're passing in is_mtp_layer for other reasons though. Wondering if this comment is a red herring

megatron/core/ssm/mamba_block.py
```diff
@@ -29,6 +29,9 @@
 from megatron.core.transformer.spec_utils import ModuleSpec, build_module
 from megatron.core.transformer.transformer_layer import TransformerLayer
 from megatron.core.transformer.utils import sharded_state_dict_default
+
+from megatron.core.ssm.mlp_layer import MLPLayer
+
```
duncanriach: Not used? Remove

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -22,9 +23,14 @@
 from megatron.core.transformer.enums import AttnMaskType, LayerType
 from megatron.core.transformer.module import MegatronModule
 from megatron.core.transformer.spec_utils import ModuleSpec, build_module
+
+# Type hint imports for Mamba support
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from megatron.core.ssm.mamba_block import MambaStackSubmodules
```
duncanriach: This seems weird and unusual. Why is this complexity needed?

megatron/training/arguments.py
```diff
@@ -582,6 +582,93 @@ def validate_args(args, defaults={}):
...
+                )
+                args.mtp_num_layers = inferred_mtp_num_layers
+
+    # MTP validation
+    if args.mtp_num_layers:
+        assert not args.use_legacy_models, "The legacy Megatron models does not support Multi-Token Prediction (MTP)."
+        assert args.position_embedding_type == "rope" or args.position_embedding_type == "none", (
+            f"Multi-Token Prediction (MTP) is not supported with {args.position_embedding_type} position embedding type."
+            + f"The supported position embedding types are rope and none."
+        )
```
duncanriach, line 550: I don't think that `args.mtp_num_layers` should be used as the measure of whether MTP is going to be present in the model. For hybrid models, in the future, using this argument will throw an exception. Whether the MTP is instantiated in the current model pipeline stage is determined during model init, which calls MTP init, if needed. These asserts could, and perhaps should, be in the MTP init.

megatron/training/arguments.py
```diff
@@ -582,6 +582,93 @@ def validate_args(args, defaults={}):
...
+            if args.mtp_num_layers is None:
+                args.mtp_num_layers = inferred_mtp_num_layers
+            elif args.mtp_num_layers != inferred_mtp_num_layers:
+                warn_rank_0(
+                    f"--mtp-num-layers ({args.mtp_num_layers}) conflicts with "
+                    f"MTP depth count ({inferred_mtp_num_layers}) in pattern '{args.hybrid_override_pattern}'. "
+                    f"Using the inferred value ({inferred_mtp_num_layers}).",
+                    args.rank
+                )
+                args.mtp_num_layers = inferred_mtp_num_layers
```
duncanriach, line 542: I'm wondering why is this not done in the hybrid model, as part of the layer pattern parsing process that is already there?

megatron/training/arguments.py
```diff
@@ -582,6 +582,93 @@ def validate_args(args, defaults={}):
...
+        getattr(args, 'mtp_hybrid_override_pattern', None) is not None
+        and args.mtp_num_layers is not None
+        and args.mtp_num_layers > 0
+        and (args.hybrid_override_pattern is None or sep not in args.hybrid_override_pattern)
+    ):
+        main_pattern = args.hybrid_override_pattern or ''
+        mtp_pattern = args.mtp_hybrid_override_pattern
+        args.hybrid_override_pattern = main_pattern + sep + sep.join([mtp_pattern] * args.mtp_num_layers)
+        args.mtp_hybrid_override_pattern = None
+        print_rank_0(f"Converted legacy MTP pattern to unified: {args.hybrid_override_pattern}")
```
duncanriach, line 526: Why is this not encapsulated in the hybrid model?

megatron/training/arguments.py
```diff
@@ -582,6 +582,93 @@ def validate_args(args, defaults={}):
...
+        # Non-hybrid (GPT) model MTP validation
+        if args.mtp_hybrid_override_pattern is not None:
+            warn_rank_0(
+                "--mtp-hybrid-override-pattern is for Mamba/hybrid models only. "
+                "For GPT models, MTP replicates the main transformer layer structure. "
+                "This argument will be ignored.",
+                args.rank
+            )
+        if args.mtp_spec is not None:
+            warn_rank_0(
```
jaredcasper: This seems like it should be an error, not a warning.

megatron/training/arguments.py
```diff
@@ -3580,6 +3671,18 @@ def _add_experimental_args(parser):
...
+    # Example: "M*M*/MM/MM" means main="M*M*", MTP pattern="MM" with 2 depths.
+    group.add_argument("--mtp-hybrid-override-pattern", type=str, default=None,
+                       help='DEPRECATED: Use unified --hybrid-override-pattern instead. '
+                       'Legacy argument for loading old checkpoints. '
+                       'Force a specific hybrid layer pattern for MTP layers.')
+    group.add_argument('--mtp-spec', type=str, default=None, nargs='*',
+                       help='DEPRECATED: Use unified --hybrid-override-pattern instead. '
+                       'Legacy argument for loading old checkpoints. '
+                       'Specify the <module_location function_name> pair '
+                       'that returns a spec for mtp layer.')
```
duncanriach: Since we never want anyone to set these from the command-line, can we just not include them here.

megatron/training/arguments.py
```diff
@@ -582,6 +582,93 @@ def validate_args(args, defaults={}):
...
+            "This argument is kept only for loading old checkpoints.",
+            args.rank,
+        )
+    if args.mtp_spec is not None:
+        warn_rank_0(
+            "--mtp-spec is deprecated. "
+            "For new hybrid models with MTP models, use unified --hybrid-override-pattern instead. "
+            "This argument is kept only for loading old checkpoints.",
+            args.rank,
+        )
```
duncanriach: If these will only ever be getting set by a checkpoint, then do we need a deprecation warning?

megatron/training/arguments.py
```diff
@@ -582,6 +582,93 @@ def validate_args(args, defaults={}):
...
+                "This argument will be ignored.",
+                args.rank
+            )
+        if args.mtp_spec is not None:
+            warn_rank_0(
+                "--mtp-spec is for Mamba/hybrid models only. "
+                "For GPT models, use the standard layer spec. "
+                "This argument will be ignored.",
+                args.rank
+            )
```
duncanriach: If these will only be getting set by a hybrid checkpoint, then I don't think we need these checks

megatron/training/arguments.py
duncanriach, line 1: An overall comment on this file. I've been pretty hung up on it today. I'm not sure how to proceed with a bunch of things here. I think we need to discuss.

megatron/training/arguments.py
sancha: Its a style question. I didnt want to pollute the mamba_model with legacy handling code. It looks cleaner now, than with this arg-parsing logic in there.

megatron/training/arguments.py
sancha: I am setting `args.mtp_num_layers` in the hybrid codepath for training.py to avoid parsing the hybrid pattern multiple times.

duncanriach: I've been looking for what you're referring to in `training.py`. All I see there is another parsing of the layer pattern, an expansion of the existing (not-well-encapsulated) parsing that was going on there. But I understand the basic problem here to be that `args.mtp_num_layers` is used in various places and so it has to be set early, which means that if the number of MTP layers it inferred from the pattern, the pattern must be parsed early. So that ends up happening in arguments.py, rather than (only) down inside the hybrid model.

megatron/training/arguments.py
duncanriach: I'm concerned that it's going to mean that I have to implement some or all of the new `--hybrid-layer-functionality` also in `arguments.py`, when it would ideally all be encapsulated inside the hybrid model.

megatron/training/arguments.py
```diff
@@ -3580,6 +3657,13 @@ def _add_experimental_args(parser):
                             'hybrid ratio arguments, then the number of each type'
                             'of layer in the override pattern must match number in'
                             'the overidden pattern')
+    # DEPRECATED: Legacy MTP arguments kept for checkpoint compatibility.
+    # For new hybrid models with MTP models, use unified pattern in --hybrid-override-pattern instead.
+    # Example: "M*M*/MM/MM" means main="M*M*", MTP pattern="MM" with 2 depths.
+    group.add_argument("--mtp-hybrid-override-pattern", type=str, default=None,
+                       help='DEPRECATED: Use unified --hybrid-override-pattern instead. '
+                       'Legacy argument for loading old checkpoints. '
+                       'Force a specific hybrid layer pattern for MTP layers.')
```
duncanriach: Are you sure we still need to add this here? Is this needed in order to load checkpoints that set this argument?

sancha: yes, we need this here, otherwise I have to invent a non trivial way to handle args in the ckpt but not in the args.py and it starts to look ugly.

duncanriach (approved):
Amazing work, Sanjeev. Thank you. Approving

ericharper (approved):

deepakn94:
/ok to test [d355d67](https://github.com/NVIDIA/Megatron-LM/pull/2363/commits/d355d67c06ed8d89d93c3149890012c090ba2b29)

sancha:
/ok to test [ea3c2bd](https://github.com/NVIDIA/Megatron-LM/pull/2363/commits/ea3c2bd8e2db7c88a090fff76900d3454fd980d3)

sancha:
/ok to test [19a56e5](https://github.com/NVIDIA/Megatron-LM/pull/2363/commits/19a56e5c9b3dc2e512af5fa7a17f29577e3d4616)

sancha:
/ok to test [e50409b](https://github.com/NVIDIA/Megatron-LM/pull/2363/commits/e50409bb290abf742a69bdd76d739448cb6b2eb9)

sancha:
/ok to test [906e513](https://github.com/NVIDIA/Megatron-LM/pull/2363/commits/906e51356bf32ab22c44c696766f085b6bc900cf)

sancha:
/ok to test [4bfc5ec](https://github.com/NVIDIA/Megatron-LM/pull/2363/commits/4bfc5ec42668f6f409d63990e7988bc66a5210ac)

sancha:
/ok to test [9cc1668](https://github.com/NVIDIA/Megatron-LM/pull/2363/commits/9cc1668ca68dac49027e7ef40dd13eb675def782)

deepakn94:
For posterity, this PR was re-merged as https://github.com/NVIDIA/Megatron-LM/pull/3207 with some bugfixes.

PR #2378, opened 2025-11-24
Improve performance of request_metadata logic

tdene:
# What does this PR do ?
The current dynamic sampling code relies on code that surrounds the `request_metadata` tensor in the dynamic context.    
Profiling, and A/B `timeit` testing, shows that this `request_metadata` code is poorly optimized.

This PR will address 3 issues found by @mathemakitten 

1. Copying & dtype-casting GPU tensors takes miliseconds, even with `non_blocking=True`.
2. Bucketizing a tensor via `torch.unique` is orders of magnitude slower than doing it via list comprehension.
~~3. In order to fix 1) and 2), the "hashed sampling parameters" tracked metadata must be resurrected.~~

BENCHMARK INFO:
 * The "golden" benchmark prior to this PR returns an average value of `81.77`, with expected noise of +/- 2%
 * The "golden" benchmark after the commit, `Split request_metadata 2D tensor into 1D slices`, returns an average value of `81.41`, with expected noise of +/- 2%. This particular change had a risk of degrading performance, but it does not seem to do so.
 * The "golden" benchmark after the commit, `Avoid casting/copying request metadata`, returns an average value of `84.87`, with expected noise of +/- 2%. This particular change was intended to improve performance, and it does.
 * The "golden" benchmark after the commit, `Replace tensorized bucketization with for loop`, returns an average value of `85.95`, with expected noise of +/- 2%. This particular change was intended to improve performance, and it does.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

4. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


mathemakitten (approved):

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1482,14 +1487,17 @@ def add_request(self, req: DynamicInferenceRequest, chunk_length: Optional[int]
...
-            len(metadata) == self.num_request_metadata
-        ), "Request added to context with invalid metadata length"
-        self.request_metadata[current_id] = torch.tensor(
-            metadata, dtype=torch.float32, device=self.request_metadata.device
-        )
+            req.get_metadata_types() == self.request_metadata_types
+        ), "Request added to context with invalid metadata types"
+        metadata = req.tracked_metadata
+        metadata_types = req.get_metadata_types()
+        for m, m_type in zip(metadata, metadata_types):
```
kvareddy, line 1497: Teo, we are adding a lot of for loops. They might affect perf. Can we avoid them?

tdene: All of the new for loops inside the `dynamic_context` only loop over 7 items, so they are not very expensive.

They are a side-effect of splitting up a large 2D tensor into smaller 1D tensors, and cannot be avoided because of that.

Doing this appears to be necessary. Using a single large 2D tensor meant all the metadata needed to share the same datatype and device. That meant we were spending a lot of time casting into other datatypes and copying between GPU and CPU. Having smaller 1D tensors means that each one can already be the correct datatype.

This change is what led to most of the performance gains this PR creates (throughput gains between 5% and 12%, depending on the test).

tdene: There is also a new for loop inside `text_generation_controller`. That one's a funny story: using a for loop in Python is actually ~20x faster than using tensor operations. I have a 40-line standalone Python script that confirms this speed-up happens for both thousands of requests and small numbers of requests.

kvareddy: I see.

kvareddy: On a different note, why are are we tracking request meta data in dynamic context.py? Is request.metadata  used in context routines?

tdene: The context shuffles several tensors during `update_requests` to keep track of which requests are active/paused and what positions in the batch they hold. The request metadata needs to be shuffled at the same time (in `_move_book_keeping_tensors` and `_swap_book_keeping_tensors`). Doing so is computationally cheap. The alternative is having a for loop on every `step` that figures out which logits correspond to which request metadata, which was the original implementation and was very slow.

But I share this concern. The context never uses this metadata, it only shuffles it. I wrote a very long Slack message on November 5th about how putting this in `dynamic_context.py` makes me uncomfortable.

@lmcafee-nvidia convinced me that we should put it in the context for now, and that we can revisit this decision at a later time. Maybe after #2456 and #2457, we can talk about factoring code out into a separate class.

lmcafee-nvidia: I personally think it makes sense that anything that needs to be shuffled along with the requests should be put in the context, otherwise you take both a performance hit and added code complexity as Teo mentioned. I think `metadata_types` does a good job of abstracting metadata details from the context.

I actually had the same question about separate tensors vs. casting after one large 2D tensors. Thanks for comparing this @tdene .

PR #2414, opened 2025-11-26
[training migration] Add LoggerConfig dataclass

maanug-nv:
# What does this PR do ?

1. Create dataclass in MLM
2. Get MLM arguments to use dataclass

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/training/arguments.py
+++ b/megatron/training/arguments.py
@@ -2025,83 +2025,11 @@ def _add_config_logger_args(parser):
 
 
 def _add_logging_args(parser):
-    group = parser.add_argument_group(title='logging')
+    from megatron.training.training_config import LoggerConfig
 
-    group.add_argument('--log-params-norm', action='store_true',
-                       help='If set, calculate and log parameters norm.')
-    group.add_argument('--log-num-zeros-in-grad', action='store_true',
-                       help='If set, calculate and log the number of zeros in gradient.')
-    group.add_argument('--log-throughput', action='store_true',
-                       help='If set, calculate and log throughput per GPU.')
-    group.add_argument('--log-progress', action='store_true',
-                       help='If set, log progress (in terms of number of processed tokens and '
-                       'number of floating-point operations) to progress.txt file in checkpoint '
-                       'directory.')
-    group.add_argument('--timing-log-level', type=int,
-                       default=0, choices=range(0,3),
-                       help='Granularity level to measure and report timing. '
-                       '   0: report only iteration time and make sure timing '
-                       '      does not introduce extra overhead.'
-                       '   1: report timing for operations that are executed '
-                       '      very limited times (basically once) during '
-                       '      each iteration (such as gradient all-reduce) '
-                       '   2: report timing for operations that migh be '
-                       '      executed numerous times during each iteration. '
-                       'Note that setting the level to 1 or 2 might '
-                       'cause increase in iteration time.')
-    group.add_argument('--log-energy', action='store_true',
-                       help='If set, log energy consumption (in Joules)')
-    group.add_argument('--no-barrier-with-level-1-timing', action='store_false',
-                       help='If not set, use barrier with level 1 time '
-                       'measurements. Note that this is up to the user '
-                       'to make sure calling barrier with their timers '
-                       'will not result in hangs. This can happen if for '
-                       'example the user adds a level 1 timer that is not '
-                       'called by all ranks.',
-                       dest='barrier_with_L1_time')
-    group.add_argument('--timing-log-option', type=str, default='minmax',
-                       choices=['max', 'minmax', 'all'],
-                       help='Options for logging timing:'
-                       '  max: report the max timing across all ranks'
-                       '  minmax: report min and max timings across all ranks'
-                       '  all: report timings of all ranks.')
-    group.add_argument('--tensorboard-log-interval', type=int, default=1,
-                       help='Report to tensorboard interval.')
-    group.add_argument('--tensorboard-queue-size', type=int, default=1000,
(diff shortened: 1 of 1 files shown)


megatron/training/argument_utils.py
```diff
@@ -0,0 +1,250 @@
...
+import enum
+from dataclasses import Field, fields
+
+# TODO: support arg renames
+
+class TypeInferenceError(Exception):
+    """Custom exception type to be conditionally handled by ArgumentGroupFactory."""
+    pass
+
+class ArgumentGroupFactory:
```
Skylion007, line 21: This is basically reinventing hydra lol with slightly different syntax.

If you pass an arg parse object into OmegaConf.fromdotlist it should be equivalent

maanug-nv: `OmegaConf.from_dotlist()` requires a list of strings, so I'm not sure how you could pass an `argparse.Namespace` into it?

Intent here is to preserve the argparse interface and arguments for MegatronLM, since many users have scripts that are built on top of that (i.e., cannot switch to something like hydra). Using dataclasses in the training loop presents several benefits, but I'd like to avoid having to define the argument in 2 places, once in the dataclass definition and once in `arguments.py` with `add_argument()`.

megatron/training/arguments.py
```diff
@@ -1930,83 +1930,11 @@ def _add_config_logger_args(parser):
...
-                       'Ignore wandb by default.')
-    group.add_argument('--wandb-exp-name', type=str, default='',
-                       help='The wandb experiment name.')
-    group.add_argument('--wandb-save-dir', type=str, default='',
-                       help='Path to save the wandb results locally.')
-    group.add_argument('--logging-level', type=int, default=None,
-                       help='Set default logging level')
+    from megatron.training.training_config import LoggerConfig
+
+    log_factory = ArgumentGroupFactory(LoggerConfig, exclude = ["log_throughput_to_tensorboard", "throughput_window_size", "memory_keys", "log_l2_norm_grad_to_tensorboard", "log_runtime_to_tensorboard", "runtime_time_unit", "filter_warnings", "modules_to_filter", "set_level_for_all_loggers", "save_config_filepath"])
```
maanug-nv, line 2030: Several of these excluded settings are for metrics currently logged in Megatron Bridge but not in MegatronLM. I think we will eventually like to upstream logging of those metrics to MegatronLM.
cc @dimapihtar

megatron/training/arguments.py
```diff
@@ -2025,83 +2025,11 @@ def _add_config_logger_args(parser):
...
-                       '   1: report timing for operations that are executed '
-                       '      very limited times (basically once) during '
-                       '      each iteration (such as gradient all-reduce) '
-                       '   2: report timing for operations that migh be '
-                       '      executed numerous times during each iteration. '
-                       'Note that setting the level to 1 or 2 might '
-                       'cause increase in iteration time.')
-    group.add_argument('--log-energy', action='store_true',
-                       help='If set, log energy consumption (in Joules)')
-    group.add_argument('--no-barrier-with-level-1-timing', action='store_false',
```
Wohox, line 2054: Please correct me if I am wrong:
`--no-barrier-with-level-1-timing` defaults to True, which means `barrier_with_level_1_timing` defaults to False, which is not aligned in LoggerConfig.
* If user scripts doesn't contain `--no-barrier-with-level-1-timing`, then the default behavior changes.
* If user scripts contains `--no-barrier-with-level-1-timing`, then behavior is the same.

jaredcasper: My understanding:

The action here is `store_false`, which means the presence of this argument sets `barrier_with_l1_timing` (the `dest`) to false. Without this flag (i.e. default), `barrier_with_l1_timing` is True. 

But it does seems like the LoggerConfig gets it wrong, unless it somehow knows the argument's action should be `store_false` (which happens in a few places in our arguments).

maanug-nv:
```
from argparse import ArgumentParser


if __name__ == "__main__":
    parser = ArgumentParser()
    parser.add_argument('--no-barrier-with-level-1-timing', action='store_false', dest='barrier_with_L1_time')
    args = parser.parse_args()

    print(args)
```

```
maanug:~ ➜ python test-l1.py
Namespace(barrier_with_L1_time=True)
maanug:~ ➜ python test-l1.py --no-barrier-with-level-1-timing
Namespace(barrier_with_L1_time=False)
```

If user script does not contain the flag default is True. If user script contains the flag, changes to False.
More or less confirming what @jaredcasper said.

`ArgumentGroupFactory` sets `store_false` if default is True and `store_true` if default is False.
https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/argument_utils.py#L166-L172

maanug-nv: Here are the kwargs for add_argument() for this one:
```
{'arg_names': ['--no-barrier-with-level-1-timing'], 'dest': 'barrier_with_L1_time', 'help': 'If not disabled, use barrier with level 1 time measurements. Note that this is up to the user to make sure calling barrier with their timers will not result in hangs. This can happen if for example the user adds a level 1 timer that is not called by all ranks.', 'default': True, 'action': 'store_false'}
```

i'm just printing [this](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/argument_utils.py#L202) and running `pretrain_gpt.py --help`

PR #2431, opened 2025-11-29
[training migration] Add CheckpointConfig dataclass

maanug-nv:
# What does this PR do ?

1. Create dataclass in MLM
2. Get MLM arguments to use dataclass

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/training/arguments.py
+++ b/megatron/training/arguments.py
@@ -2516,9 +2516,6 @@ def _add_checkpointing_args(parser):
     ckpt_factory = ArgumentGroupFactory(CheckpointConfig, exclude=["most_recent_k", "save_tokenizer_assets", "save_optim", "save_rng", "load_optim", "load_rng"])
     group = ckpt_factory.build_group(parser, "checkpointing")
 
-    group.add_argument('--save-retain-interval', type=int, default=None,
-                       help='Number of iterations between retained checkpoints (other'
-                       'checkpoints _except the last checkpoint_ are automatically deleted).')
     group.add_argument('--no-save-optim', action='store_true', default=None,
                        help='Do not save current optimizer.')
     group.add_argument('--no-save-rng', action='store_true', default=None,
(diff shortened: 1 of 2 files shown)


megatron/training/arguments.py
```diff
@@ -2432,131 +2432,34 @@ def _add_learning_rate_args(parser):
...
-    group.add_argument('--save', type=str, default=None,
-                       help='Output directory to save checkpoints to.')
-    group.add_argument('--save-interval', '--persistent-save-interval', type=int, default=None,
-                       help='Number of iterations between persistent checkpoint saves.')
-    group.add_argument('--save-retain-interval', type=int, default=None,
-                       help='Number of iterations between retained checkpoints (other'
-                       'checkpoints _except the last checkpoint_ are automatically deleted).')
+    from megatron.training.config import CheckpointConfig
+
+    ckpt_factory = ArgumentGroupFactory(CheckpointConfig, exclude=["most_recent_k", "save_tokenizer_assets", "save_optim", "save_rng", "load_optim", "load_rng"])
```
maanug-nv, line 2516: `most_recent_k` and `save_tokenizer_assets` are features in Megatron Bridge that are not available in MegatronLM currently.

maanug-nv: save/load optim/rng flags are different in bridge and MLM. 
in bridge, save optim is referenced by `CheckpointConfig.save_optim` and is default True. 
in MLM, save optim is referenced by `args.no_save_optim` and is default False.

Not sure how to address this discrepancy right now.

Wohox: 1. Can you elaborate more on the discrepancy, the default value of save_optim is aligned in CheckpointConfig and args, both are set to True. The issue is when user adds `--no-save-optim` flag, there is no mapping to CheckpointConfig.
2. I think this PR should replace `self.config.no_save_optim` to `self.config.save_optim`, which won't influence the user args.

This is same for `--no-save-rng --no-load-optim --no-load-rng --no-initialization`

megatron/training/arguments.py
```diff
@@ -2591,131 +2591,34 @@ def _add_learning_rate_args(parser):
...
-    group.add_argument('--use-mp-args-from-checkpoint-args', action='store_true',
-                       help='Copy model parallelism command-line arguments from checkpoint')
-    group.add_argument('--no-use-tokenizer-model-from-checkpoint-args', action='store_false',
-                       dest='use_tokenizer_model_from_checkpoint_args',
-                       help='If set, do not use tokenizer model path from checkpoint')
-    group.add_argument('--exit-on-missing-checkpoint', action='store_true',
-                       help="If '--load' is set, but checkpoint is not found "
-                       "(e.g., path typo), then exit instead of random "
-                       "initialization.")
     group.add_argument('--use-dist-ckpt', action='store_true',
```
Wohox, line 2532: Why does this flag get kept?

maanug-nv: it's deprecated. I don't really see a reason to add already deprecated flags to the configs. 
If you have a strong opinion on this, I can add it to the config, it won't break Megatron Bridge.

megatron/training/arguments.py
maanug-nv: Confused by 2. `CheckpointConfig` currently contains `save_optim`, not `no_save_optim`. 
The basic issue is that `--no-save-optim` and others don't follow the more common pattern of setting `dest="save_optim"`. `ArgumentGroupFactory` is built around that more common pattern. In other words, for boolean settings that are default False, the factory always generates `--no-some-feature` with `dest=some_feature`. save/load optim/rng use `dest=no_save_optim`. 

The best solution is to replace all `args.no_save_optim` with `not args.save_optim`. I also think it makes more sense to do this when we start using the config classes in functions instead of args.

Wohox: > The best solution is to replace all args.no_save_optim with not args.save_optim. I also think it makes more sense to do this when we start using the config classes in functions instead of args.

I meant the user should still use `--no-save-optim` flag and the `args.no_save_optim` should be depreacted and value get mapped to `CheckpointConfig.save_optim` (like in #2896: `add_bias_linear: bool = field(
        default=True, metadata={"argparse_meta": {"arg_names": ["--disable-bias-linear"]}}`), then let's replace all the reference of `args/config.no_save_optim` to `not config/args.save_optim`.

PR #2450, opened 2025-12-02
Default UVM level to 0.

lmcafee-nvidia:
# What does this PR do ?
Default UVM level to 0 until we better understand UVM's impact on latency.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


lmcafee-nvidia:
@shanmugamr1992 @santhnm2 , can you please review?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -675,28 +675,38 @@ def from_config(
...
-            inference_config.inference_max_seq_length or model_config.max_sequence_length
-        )
+        # Max sequence length. (todo: why does inference config have default value
+        # for inference_max_seq_length?)
+        model_max_seq_len = model.get_max_sequence_length()
+        inf_max_seq_len = inference_config.inference_max_seq_length
+        if inf_max_seq_len:
+            max_sequence_length = min(model_max_seq_len, inf_max_seq_len)
+        else:
+            max_sequence_length = model_max_seq_len
```
santhnm2: I think this change is correct if we're solely concerned about accuracy but for benchmarking purposes I think we would want to be able to exceed the maximum model sequence length.

megatron/core/models/gpt/gpt_model.py
```diff
@@ -258,6 +258,14 @@ def __init__(
                 quant_config = get_quant_config_or_none(name, self.config.quant_recipe)
                 module.finish_init(quant_config)
 
+    def get_max_sequence_length(self) -> int:
```
santhnm2: Do we need this as a function? Can we just use `model.max_sequence_length` directly?

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: I'm not sure how this was working before, but in `test_static_engine.py` today, `inf_max_seq_len` is 2560 (the default config value), and `model_max_seq_len` is 64, leading to position ids that are too large and we fail during position embeddings. This failure is exposed when only using GPU memory, but disturbingly it worked with unified memory. I'm not sure what changes have taken place since UVM was merged that now cause this to fail, but the position ids will simply be too large. So, I'm curious why we want this to be larger for benchmarking purposes ... something's not feeling right here to me.

lmcafee-nvidia: I personally also feel that `InferenceWrapperConfig` should not have a default value at all for `inference_max_seq_length`, I think this value should always be less than or equal to the model's `max_sequence_length`.

santhnm2: > I'm not sure how this was working before, but in test_static_engine.py today, inf_max_seq_len is 2560 (the default config value), and model_max_seq_len is 64, leading to position ids that are too large and we fail during position embeddings. 

Where is the error triggered by having position ids that are too large? My understanding was that `model_max_seq_len` is just telling us what sequence length the model was trained with but there's no explicit requirement that we do inference at that maximum sequence length, it just means we'll get garbage outputs beyond the `model_max_seq_len` which is ok if we are only concerned with benchmarking speed.

lmcafee-nvidia: the error happens on this line: https://github.com/NVIDIA/Megatron-LM/blob/587a0ff13d4fa412dc0cc09032aff8b586f887d6/megatron/core/models/common/embeddings/language_model_embedding.py#L113

If you use `main` today, and switch from using UVM to GPU memory by using `--inference-dynamic-batching-unified-memory-level 0`, the `position_ids` at this point will have a max value of 513 during cuda graph warmup, but the model has a `max_sequence_length` of 64 and therefore the `position_embeddings` here have a shape of `(64, 32)`, which means that the values in `position_ids` are too large for `position_embeddings`. This is all taken from `test_static_engine.py`'s `test_generate_dynamic()` test.

megatron/core/models/gpt/gpt_model.py
lmcafee-nvidia: I'm okay going either way, but I made it a function because using `model.max_sequence_length` directly feels a little tailored toward GPT. I know we mostly on care about GPT inference, but I was trying to be generic here. But I can switch it.

Side note, but it was buggy before because originally it tried to use `model_config.max_sequence_length`, but there is no such thing `model_config.max_sequence_length` since the sequence length isn't stored at all in this config. (During training, it's stored in the dataset config.)

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Discussed with @lmcafee-nvidia offline, we came to the conclusion that there will be an error in the position embeddings calculation if any `position_id` value is greater than the model maximum sequence length, but globally enforcing the constraint that `inference_max_seq_length <= model.max_sequence_length` is perhaps too restrictive since some models don't even use position embeddings (or they use a version of position embeddings that is not monotonically increasing). So instead we will only enforce this constraint if we're using absolute position embeddings.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -682,15 +683,26 @@ def from_config(
...
-        # for inference_max_seq_length?)
-        model_max_seq_len = model.get_max_sequence_length()
+        # Max sequence length.
+        def get_model_attr(key):
+            try:
+                return getattr(model, key)
+            except Exception as e:
+                raise e
+
+        position_embedding_type = get_model_attr("position_embedding_type")
```
santhnm2: Can you use `get_attr_wrapped_model` from `megatron/core/utils.py`? https://github.com/NVIDIA/Megatron-LM/blob/d2bd9fa89fe2fb74f002d68ad17cd5e47181dfcc/megatron/core/utils.py#L575

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -682,15 +683,26 @@ def from_config(
...
+        model_max_seq_len = get_model_attr("max_sequence_length")
         inf_max_seq_len = inference_config.inference_max_seq_length
-        if inf_max_seq_len:
-            max_sequence_length = min(model_max_seq_len, inf_max_seq_len)
+        if position_embedding_type == "learned_absolute":
```
santhnm2, line 715: Can you add a comment explaining why we're only doing this check for `learned_absolute`?

santhnm2 (approved):
LGTM % some minor nits. Do any unit tests and/or functional tests need to be modified to reflect the change in UVM default?

PR #2452, opened 2025-12-02
Enable training cudagraphs for RL

mathemakitten:
# What does this PR do ?
Cudagraphs will be enabled for training and inference when `--rl-training-cuda-graphs` is passed.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -551,12 +550,6 @@ def suspend(self):
         ):
             self.context.deallocate_all_tensors()
 
-        # Delete cuda graphs when not using unified memory at all (level 0). For
-        # levels 1 and 2, the context's tensors maintain static memory addresses,
-        # so the cuda graphs are re-used.
-        if self.unified_memory_level == 0:
-            delete_cuda_graphs()
-
```
tdene, line 574: @mathemakitten I'm pretty certain this is necessary. This codepath is used when KV$ is manually offloaded / onloaded, so we need to rebuild CGs to point to the new KV$ memory address.

mathemakitten: I've gated this behind `--rl-training-cuda-graphs` now with an arg description that says `Use only when all training and inference cudagraphs and the KV cache fit on device.`

megatron/core/transformer/cuda_graphs.py
```diff
@@ -175,6 +175,40 @@ def _determine_if_first_last_layer_of_this_vp_chunk(base_module):
...
 
+def _clone_nested_tensors(value: Any) -> Any:
+    """Recursively clone tensors inside nested containers."""
+    if torch.is_tensor(value):
+        return value.clone()
+    if isinstance(value, (tuple, list)):
+        return type(value)(_clone_nested_tensors(v) for v in value)
+    if isinstance(value, dict):
+        return {k: _clone_nested_tensors(v) for k, v in value.items()}
+    return value
```
yobibyte, line 192: Seems a bit dangerous to have Any and not throw an exception for unsupported data types. For instance, if I send a set of tensors, this will not clone them but return a set of tensors which a user of the function would expect to have cloned.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1324,7 +1372,7 @@ def __call__(self, megatron_module, args, kwargs):
                 # Now replay the graph
                 out = runner.replay_graph_capture(self.is_first_microbatch, args, kwargs)
 
-            elif self.training:
+            elif self.training and torch.is_grad_enabled():
```
wdykas: Not related specifically to this line or meant to be address in this MR, but do we ever need to be able to distinguish between no_grad and inference_mode in the code? If two forwards have the same signature but one is no grad and another is inference mode I wonder if it will error at all in inlace ops

megatron/core/transformer/cuda_graphs.py
mathemakitten: We actually don't want to throw an exception, but to pass-through for unsupported datatypes. This helper function is scoped very tightly within the cudagraphs code such that the "send set of tensors" case would *not be realistic, in my opinion.

yobibyte: I did not think of that! Got it now. The only q I have is why if I send a tuple/list of tensors we want to clone them, but if I send a set of tensors we don't clone them?

megatron/core/transformer/cuda_graphs.py
mathemakitten: I've made this unnecessary by flipping to `model.eval()` and then back to `model.train()` in `logprobs_forward_step` so it will never trigger cudagraph capture.

mathemakitten: But yes you're right — right now we rely on cudagraph runner matching guardrails like correct shapes and arguments (e.g. inference passes some arguments which training does not). If we ever have graphs with the exact same shapes and arguments but differing only between inference mode and no_grad mode we may have issues.

megatron/core/transformer/cuda_graphs.py
mathemakitten: Sets of tensors are never used as cudagraph input/output containers — it wouldn't make sense because inputs are ordered, duplicates are meaningful, etc. I would be more concerned if we were sending in a set of tensors at all so I will just throw an error.

This function does see other types (e.g. generators) which need to pass through so I'm not going to modify the typing hints.

PR #2454, opened 2025-12-02
Migrate SelfAttention and CrossAttention off of ModuleSpec

nschank:
# What does this PR do ?

Demonstrates what a migration away from ModuleSpec would look like. Draft is mostly for demonstration - can break it up or target the first updates differently depending on request.

Associated design doc: https://docs.google.com/document/d/1shyv0iKEzRdevLOlouF_NktbdJazvWifqxUwPXFigQE/edit?tab=t.0#heading=h.uwes2zo47yg6

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/extensions/transformer_engine_spec_provider.py
+++ b/megatron/core/extensions/transformer_engine_spec_provider.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
 import warnings
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Union
 
 from megatron.core.extensions.transformer_engine import (
     TEActivationOp,
@@ -45,7 +45,9 @@ class TESpecProvider(BackendSpecProvider):
         """Which module for sequential layernorm and linear"""
         return TELayerNormColumnParallelLinear
 
-    def layer_norm(self, rms_norm: bool = False, for_qk: bool = False) -> type:
+    def layer_norm(
+        self, rms_norm: bool = False, for_qk: bool = False
+    ) -> Union[type[FusedLayerNorm], type[TENorm]]:
         """Which module to use for layer norm"""
         if for_qk and not is_te_min_version("1.9.0"):
             # TENorm significantly harms convergence when used
--- a/megatron/core/models/backends.py
+++ b/megatron/core/models/backends.py
@@ -2,7 +2,7 @@
 
 import warnings
 from abc import abstractmethod
-from typing import Optional, Protocol, Tuple
+from typing import Optional, Protocol, Tuple, Union
 
 from megatron.core.tensor_parallel.layers import ColumnParallelLinear, RowParallelLinear
 from megatron.core.transformer.dot_product_attention import DotProductAttention
@@ -101,7 +101,9 @@ class LocalSpecProvider(BackendSpecProvider):
         """Which module for sequential layernorm and linear"""
         return None
 
-    def layer_norm(self, rms_norm: bool = False, for_qk: bool = False) -> type:
+    def layer_norm(
+        self, rms_norm: bool = False, for_qk: bool = False
+    ) -> Union[type['FusedLayerNorm'], type[WrappedTorchNorm]]:
         """Which module to use for layer norm"""
         if rms_norm:
             # Matching get_gpt_layer_local_spec.
@@ -157,7 +159,9 @@ class InferenceSpecProvider(BackendSpecProvider):
         """Which module for sequential layernorm and linear"""
         return InferenceLayerNormColumnParallelLinear
 
-    def layer_norm(self, rms_norm: bool = False, for_qk: bool = False) -> type:
+    def layer_norm(
+        self, rms_norm: bool = False, for_qk: bool = False
+    ) -> Union[type['FusedLayerNorm'], type[TENorm]]:
         """Which module to use for layer norm"""
         if for_qk and not is_te_min_version("1.9.0"):
             # TENorm significantly harms convergence when used
--- a/megatron/core/models/gpt/heterogeneous/heterogeneous_layer_specs.py
+++ b/megatron/core/models/gpt/heterogeneous/heterogeneous_layer_specs.py
@@ -116,10 +116,13 @@ def _get_heterogenous_attention_spec(
             linear_qkv = TELayerNormColumnParallelLinear
             core_attention = TEDotProductAttention
             linear_proj = TERowParallelLinear
+
         else:
             linear_qkv = ColumnParallelLinear
             core_attention = DotProductAttention
             linear_proj = RowParallelLinear
+        from transformer_engine import pytorch as te_pytorch
+
         self_attention = ModuleSpec(
             module=SelfAttention,
             params={"attn_mask_type": AttnMaskType.causal},
--- a/megatron/core/transformer/attention.py
+++ b/megatron/core/transformer/attention.py
@@ -27,7 +27,6 @@ from megatron.core.parallel_state import (
 from megatron.core.process_groups_config import ProcessGroupCollection
 from megatron.core.transformer.identity_op import IdentityOp
 from megatron.core.transformer.module import MegatronModule
-from megatron.core.transformer.spec_utils import ModuleSpec, build_module
 from megatron.core.utils import (
     deprecate_inference_params,
     divide,
@@ -223,6 +222,18 @@ class LinearProjBuilder(Protocol):
     ) -> LinearProj: ...
 
 
+class LayerNorm(Protocol):
+    """Protocol for a layernorm layer."""
+
+    def forward(self, input: Tensor, /) -> Tensor: ...
+
+
+class LayerNormBuilder(Protocol):
+    """Protocol for building layernorm layers."""
+
+    def __call__(self, *, hidden_size: int, config: TransformerConfig, eps: float) -> LayerNorm: ...
+
+
 @dataclass
 class SelfAttentionSubmodules:
     """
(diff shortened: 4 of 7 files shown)


Skylion007:
FYI, this typing utility might be useful. I need to clean it up a bit though: https://github.com/pytorch/pytorch/pull/163418/files let's you copy signatures form super classes and validate it.

examples/multimodal/layer_specs.py
```diff
@@ -122,12 +124,16 @@ def get_layer_spec_te(is_vit=False, padding=False) -> ModuleSpec:
...
     )
 
+
 def get_mamba_layer_spec_te(padding=False) -> ModuleSpec:
     attn_mask_type = AttnMaskType.causal
     # Padding mask is needed for e.g. Context Parallel.
     if padding:
         attn_mask_type = AttnMaskType.padding_causal
 
+    assert TELayerNormColumnParallelLinear is not None
```
Skylion007, line 134: Creating a not_none typing utility (or just using the one in PyTorch) might make your lives a bit easier.

megatron/core/extensions/transformer_engine_spec_provider.py
```diff
@@ -33,19 +33,21 @@ def column_parallel_linear(self) -> type:
...
 
-    def column_parallel_layer_norm_linear(self) -> Optional[type]:
+    def column_parallel_layer_norm_linear(self) -> type[TELayerNormColumnParallelLinear]:
         """Which module for sequential layernorm and linear"""
         return TELayerNormColumnParallelLinear
 
-    def layer_norm(self, rms_norm: bool = False, for_qk: bool = False) -> type:
+    def layer_norm(
+        self, rms_norm: bool = False, for_qk: bool = False
+    ) -> Union[type[FusedLayerNorm], type[TENorm]]:
```
Skylion007, line 50: Yikes, have a TypeAlias here that references all known LayerNorm classes here if you aren't going to have a parent class.

megatron/core/models/backends.py
```diff
@@ -89,19 +89,21 @@ def column_parallel_linear(self) -> type:
...
 
-    def column_parallel_layer_norm_linear(self) -> Optional[type]:
+    def column_parallel_layer_norm_linear(self) -> None:
         """Which module for sequential layernorm and linear"""
         return None
 
-    def layer_norm(self, rms_norm: bool = False, for_qk: bool = False) -> type:
+    def layer_norm(
+        self, rms_norm: bool = False, for_qk: bool = False
+    ) -> Union[type['FusedLayerNorm'], type[WrappedTorchNorm]]:
```
Skylion007, line 106: Import this other class Is it defined? Why is it stringified is it defined here too? Type checker is not gooing to like this and fall back to Any, wont it?

megatron/core/models/gpt/heterogeneous/heterogeneous_layer_specs.py
```diff
@@ -106,13 +109,27 @@ def _get_heterogenous_attention_spec(
         )
     else:
         ln = _get_qk_layernorm(use_te, normalization) if qk_layernorm else IdentityOp
+        if use_te:
+            assert TELayerNormColumnParallelLinear is not None
+            assert TEDotProductAttention is not None
+            assert TERowParallelLinear is not None
+            linear_qkv = TELayerNormColumnParallelLinear
+            core_attention = TEDotProductAttention
+            linear_proj = TERowParallelLinear
```
Skylion007, line 118:
```suggestion
            linear_qkv = not_none(TELayerNormColumnParallelLinear)
            core_attention = not_none(TEDotProductAttention)
            linear_proj = not_none(TERowParallelLinear)

File: megatron/core/models/gpt/heterogeneous/heterogeneous_layer_specs.py [hunk 5]
```diff
@@ -106,13 +109,27 @@ def _get_heterogenous_attention_spec(
...
+
+        else:
+            linear_qkv = ColumnParallelLinear
+            core_attention = DotProductAttention
+            linear_proj = RowParallelLinear
```
Skylion007, line 123: Use not_none from torch typing utilities or copy it here
```suggestion
        linear_qkv = not_none(TELayerNormColumnParallelLinear)
        core_attention = not_none(TEDotProductAttention)
        linear_proj = not_none(TERowParallelLinear)

File: megatron/core/transformer/attention.py [hunk 6]
```diff
@@ -107,17 +108,143 @@
...
+        self,
+        input_size: int,
+        output_size: int,
+        /,
+        *,
+        config: TransformerConfig,
+        init_method: Callable[[Tensor], None],
+        gather_output: bool,
+        bias: bool,
+        skip_bias_add: bool,
```
Skylion007, line 130: boolean trap, bias coould be an enum that had a skip_bias_add_option

megatron/core/transformer/spec_utils.py
```diff
@@ -24,7 +24,16 @@ class ModuleSpec:
...
+    submodules: object = None
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        """Builds an instance of the module from the spec.
+
+        Args:
+            *args: Positional arguments to be passed to the module init.
+            **kwargs: Keyword arguments to be passed to the module init.
+        """
+        return build_module(self, *args, **kwargs)
```
Skylion007, line 36: Just copy the signature from build_module or do:
```suggestion
        __call__ = build_module

File: megatron/core/typed_torch.py [hunk 8]
```diff
@@ -0,0 +1,31 @@
...
+R_co = TypeVar('R_co', covariant=True)
+
+
+class _Module(Generic[P, R_co], Protocol):
+    """Protocol allowing us to unwrap `forward`."""
+
+    def forward(self, *args: P.args, **kwargs: P.kwargs) -> R_co: ...
+
+
+def apply_module(m: _Module[P, R_co], *, check_subclass: bool = True) -> Callable[P, R_co]:
```
Skylion007, line 18: You can use a typing overload here to have check_subclass=True be a static assertion here. Now if check_subclass is True it will require it to follow the _TrueModule Protocol and complain if it doesn't. Mypy, pyright, etc prefer the more specific overload so this should work.

```suggestion
class _TrueModule(torch.nn.Module, _Module, Protocol):
    pass

@typing.overload
def apply_module(m: _TrueModule[P, R_co], *, check_subclass: Literal[True] = True) -> Callable[P, R_co]:
    pass
    
@typing.overload
def apply_module(m: _Module[P, R_co], *, check_subclass: Literal[True] = True) -> typing.Never:
    pass
```

nschank:
@Skylion007 I just saw this, sorry - I was mostly sending this PR as an example while the doc was being discussed. I broke out the first few commits into a prep PR: https://github.com/NVIDIA/Megatron-LM/pull/2668 happy to discuss further there!

megatron/core/typed_torch.py
nschank: This is a good idea, I'll add this!

megatron/core/models/backends.py
nschank: Type checker doesn't care about strings vs. not - I stringified because if the apex import fails, then this type is unbound.

However, this is all more of a draft PR just kind of showing the broader ideas here - I'd like to clean up the imports instead when I actually update this, to avoid needing to be weird like this.

examples/multimodal/layer_specs.py
nschank: Yeah that'd be fine! I don't actually want to do it this way (it's clearly sorta ugly), it was just a cheap way to get the type-checker to pass locally while I was drafting. Sorry for the confusion :)

megatron/core/extensions/transformer_engine_spec_provider.py
nschank: Ah so this is something that will take some thinking when we get to it - I'm taking advantage of covariance here and simply having the subclass specify what it would actually return, specifically because it allows for the type-checker to actually examine the underlying types for the correct signatures in practice. This is not a desired end-state, but was needed (at least while I was playing around) to actually ensure type-checking occurred at all the call-sites where implementations and `Protocols` needed to be matched up.

In reality, we wouldn't want the `BackendProvider` interface to specify a base class or a union of classes (which would prevent users from providing custom implementations without subclassing), we would want it to actually have methods which return the specific `Protocol` that needs to be satisfied; but this would require a bit more of a refactor since each individual method right now probably needs to be split up in order to describe the different ways in which the layers are used. For the moment, since there are only a few `BackendProviders` and they are mostly used non-polymorphically, it was cheaper to simply push the subclasses to declare the exact types they were returning, since that made ~almost all callers then type-check that the return types actually satisfy the `Protocols`.

In coming PRs, expect this to be done more cleanly

megatron/core/transformer/attention.py
nschank: I am explicitly not going to try to fix the existing interfaces in these PRs - these `Protocols` will be exact replicas of the current way in which the underlying modules will be used, to ensure backwards compatibility. (I'm happy to _also_ help fix things, but I don't want to get caught in the trap of doing both at once and confusing everyone)

megatron/core/transformer/spec_utils.py
nschank: What do you mean 'copy the signature'? Happy to do it either way since they're equivalent, but the existing methods has no type hints or docstring!

nschank:
> FYI, this typing utility might be useful. I need to clean it up a bit though: https://github.com/pytorch/pytorch/pull/163418/files let's you copy signatures form super classes and validate it.

You're doing god's work there, but unfortunately that wouldn't help me in the short-term - until all existing modules (both inside and outside of Megatron, plus all custom modules provided by users) are consistently annotated correctly, I will still need a way to figure out the correct signature of a module nonintrusively. I view the typed_torch library as a necessary evil which can be discarded easily when the time comes!

megatron/core/typed_torch.py
nschank: Oh wait no I tried this when I originally wrote it - `Protocols` cannot extend non-`Protocol` classes. Unfortunately python typing is fairly limited and doesn't allow for higher-kinded or constrained types :(

PR #2484, opened 2025-12-03
Add optional post-MLP LayerNorm in Transformer block for improved training stability

kmswin1:
**PR Description**

This PR adds an optional post-MLP LayerNorm inside the Transformer block, similar to stabilization techniques used in Gemma-style models. When enabled, the MLP output is normalized before the residual connection, which helps smooth activation scales and improves training stability under large batch sizes, short warmup schedules, or aggressive learning-rate regimes.

**Key Changes**

Add a new post_mlp_layernorm option and wire it through:

gpt_builders.py

GPT layer specs (gpt_layer_specs.py)

Transformer layer implementation (transformer_layer.py)

Training arguments (--post-mlp-layernorm)

Insert a LayerNorm (or IdentityOp when disabled) right after the MLP output.

Add recompute bookkeeping for post-MLP LayerNorm.

Update sharded state-dict key mappings to include the new layernorm parameters.

Default behavior remains unchanged unless the flag is enabled.

**Motivation**

Normalizing the MLP output provides more stable gradients and prevents activation blow-ups downstream, improving robustness in large-scale or high-LR training. This pattern has been adopted in recent transformer variants to improve convergence stability.

**Compatibility**

Fully backward-compatible: checkpoints without the new parameters load correctly (IdentityOp used when disabled).

Controlled by the boolean flag --post-mlp-layernorm.

**Testing Notes**

Verified forward outputs and shapes match between enabled/disabled settings.

Smoke-tested with distributed + recompute paths.

No regression observed in baseline training.

### DIFF
--- a/gpt_builders.py
+++ b/gpt_builders.py
@@ -35,6 +35,7 @@ def gpt_builder(args, pre_process, post_process, vp_stage=None, config=None):
             parallel_output=True,
             pre_process=pre_process,
             post_process=post_process,
+            post_mlp_layernorm=args.post_mlp_layernorm,
         )
     else:  # using core models
         if args.spec is not None:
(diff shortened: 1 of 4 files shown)


cuichenx:
Hi @kmswin1 , we recommend training open source models in Megatron-Bridge. For example, post layernorm in Gemma models fully implemented [here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/models/gemma/gemma3_provider.py#L409)

PR #2486, opened 2025-12-03
fix: Fix crash in validation when no main replicas on rank

ferrarojoseph7-amzn:
## Description
Fixes a crash in `validation.py` when resharding checkpoints with different tensor parallelism sizes.

## Problem
When `_validate_sharding_for_key_flattened()` runs on a rank that doesn't hold any main replicas, `all_slices` remains empty, causing `zip(*unique_slices)` to crash with `ValueError`.

## Solution
Add early return when `all_slices` is empty, allowing validation to safely skip on ranks without main replicas.

## Reproduction
1. Train model with TP=2 and save checkpoint
2. Load checkpoint with TP=4
3. Observe crash: `ValueError: zip() argument after * must be a sequence`

## Testing
Tested with various TP size changes during checkpoint resharding.

### DIFF
--- a/megatron/core/dist_checkpointing/validation.py
+++ b/megatron/core/dist_checkpointing/validation.py
@@ -518,6 +518,9 @@ def _validate_sharding_for_key_flattened(tensors_by_shard):
 
         all_slices.append((sharding.flattened_range.start, sharding.flattened_range.stop))
 
+    if not all_slices:
+        return
+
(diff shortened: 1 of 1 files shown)


gautham-kollu:
@ferrarojoseph7-amzn  there are some merge conflicts

dimapihtar:
Hi, thank you for the contribution but looks like this PR is not relevant anymore since flattened range support was removed some time ago from dist checkpointing.

Closing this PR.

PR #2492, opened 2025-12-04
Refactoring of PP implementation for hybrid models to support virtual pipeline parallelism

deepakn94:


megatron/post_training/model_builder.py
```diff
@@ -229,29 +229,21 @@ def modelopt_gpt_mamba_builder(args, pre_process, post_process, vp_stage=None, c
         mamba_stack_spec = get_mamba_stack_modelopt_spec(
             remap_te_layernorm=args.export_te_mcore_model
         )
-        model_kwargs = {
```
AAnoosheh: We re-use `model_kwargs` below these lines again

deepakn94:
Subsumed by #3377

PR #2569, opened 2025-12-05
[Dev] Refactor MoE loss logging

yanring:
# What does this PR do ?
main pr #3431 

### TL;DR

Refactors MoE metric logging from ad-hoc global state into `MoEMetricsTracker` -- a structured, per-config tracker that centralizes metric collection, distributed reduction, aggregation, and logging. The singleton default preserves existing behavior; multi-model workloads can opt into isolation by assigning separate tracker instances to each `TransformerConfig`.

---

### Problem Statement

MoE logging logic was spread across mutable global dictionaries and standalone utility functions (`save_to_aux_losses_tracker`, `reduce_aux_losses_tracker_across_ranks`, `track_moe_metrics`). The reduction semantics (which groups to reduce over, whether to DP-average) were implicit and easy to get wrong when adding new metrics. The global dict also made multi-model isolation impossible.

---

### Goals

- Centralize the full MoE metric lifecycle (record, sync, aggregate, log, clear) in one component.
- Make distributed reduction semantics explicit and configurable per metric via `needs_dp_avg`.
- Enable per-model metric isolation through `TransformerConfig` without changing training-level call sites.
- Preserve legacy entry points (`save_to_aux_losses_tracker`, `track_moe_metrics`) as deprecated wrappers.

### Non-Goals

- Change MoE loss formulas or optimizer/training semantics.
- Introduce new user-facing CLI flags.

---

### High-Level Design

```
Forward pass (per layer)         Step end (once)
─────────────────────────        ──────────────────────────────────────
router.record(name, value,  ──►  tracker.report(loss_scale, iteration, ...)
  layer_number, num_layers,        ├─ _sync_metrics()     PP → reduce_group → avg_group → DP
  reduce_group, needs_dp_avg)      ├─ _aggregate()        per-layer values → scalar means
                                   ├─ _log_scalars()      TensorBoard / W&B
                                   ├─ _log_per_layer()    optional per-layer detail
                                   ├─ _format()           console log string
                                   └─ clear()             zero buffers for next step
```

**Ownership model:**

| Layer | How it accesses the tracker | Isolation? |
|---|---|---|
| `megatron/core/` (router) | `self.config.moe_metrics_tracker` | Per-config |
| `megatron/training/` (training loop) | `MoEMetricsTracker.get_instance()` | Singleton |

- **Single model (default):** `TransformerConfig.__post_init__` sets `moe_metrics_tracker` to the global singleton. All paths point to the same instance -- behavior is identical to before.
- **Multi model:** User creates separate `MoEMetricsTracker()` instances and assigns them to each config.

---

### Per-Metric Metadata (`MetricEntry`)

Each metric name maps to a `MetricEntry` dataclass:

| Field | Type | Purpose |
|---|---|---|
| `values` | `torch.Tensor` | Per-layer accumulator (size = `num_layers`) |
| `reduce_group` | `ProcessGroup?` | Sum all-reduce group (e.g. TP+CP) |
| `avg_group` | `ProcessGroup?` | Average all-reduce group |
| `needs_dp_avg` | `bool` | Whether to average across DP after the above reductions |

This replaces the old `reduce_group_has_dp` flag (inverted semantics: `needs_dp_avg = not reduce_group_has_dp`).

---

### Reduction Semantics

For each metric, `_sync_metrics` applies reductions in this fixed order:

1. **PP all-reduce** (sum) -- collect across pipeline stages
2. **`reduce_group` all-reduce** (sum, optional) -- e.g. TP+CP group
3. **`avg_group` all-reduce** (avg, optional)
4. **DP all-reduce** (avg, if `needs_dp_avg=True`) -- average across data parallel ranks

Example: `global_load_balancing_loss` uses `reduce_group=tp_dp_cp_group` which already includes DP, so `needs_dp_avg=False` to avoid double-averaging.

---

### Key File Changes

| File | Change |
|---|---|
| `moe_logging.py` (new) | `MoEMetricsTracker` class with `record()`, `report()`, `clear()`, and private sync/aggregate/log helpers |
| `transformer_config.py` | Added `moe_metrics_tracker` field (defaults to singleton via lazy init in `__post_init__`) |
| `router.py` | Calls `self.config.moe_metrics_tracker.record(...)` instead of `save_to_aux_losses_tracker(...)` |
| `training.py` | Calls `MoEMetricsTracker.get_instance().report(...)` instead of `track_moe_metrics(...)` |
| `cuda_graphs.py` | Calls `MoEMetricsTracker.get_instance().clear()` instead of `clear_aux_losses_tracker()` |
| `moe_utils.py` | `save_to_aux_losses_tracker`, `track_moe_metrics` retained as deprecated wrappers forwarding to the tracker |

### DIFF
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -169,6 +169,7 @@ from megatron.core.num_microbatches_calculator import (
     get_num_microbatches,
     update_num_microbatches
 )
+from megatron.core.datasets.data_schedule import wrap_data_iterator
 
 from .async_utils import maybe_finalize_async_save
 from .utils import (
@@ -225,7 +226,7 @@ def print_datetime(string, override_timestamp=None):
         time_str = datetime.fromtimestamp(override_timestamp).strftime('%Y-%m-%d %H:%M:%S.%f')
     print_rank_0(f'[{string}] datetime: {time_str} ')
 
-def num_floating_point_operations(args, batch_size):
+def num_floating_point_operations(args, seqlen_sum_this_global_batch, seqlen_squared_sum_this_global_batch):
     def calculate_layer_counts():
         """Calculate the number of attention, Mamba, and MLP layers."""
         if args.hybrid_override_pattern:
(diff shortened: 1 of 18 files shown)


megatron/core/transformer/moe/moe_utils.py
```diff
@@ -914,14 +911,18 @@ def apply_router_token_dropping(
...
+)
 def save_to_aux_losses_tracker(
     name: str,
     loss: torch.Tensor,
     layer_number: int,
     num_layers: int,
     reduce_group: Optional[torch.distributed.ProcessGroup] = None,
     avg_group: Optional[torch.distributed.ProcessGroup] = None,
-    reduce_group_has_dp: bool = False,
+    layer_percentiles: Optional[List[float]] = None,
```
Victarry: Used variable `layer_percentiles`

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -0,0 +1,362 @@
...
+        tracker = MoEMetricsTracker.get_instance()
+        tracker.record("load_balancing_loss", loss, layer_number=1, num_layers=32)
+        log_str = tracker.report(loss_scale=1/8, iteration=100, writer=tb_writer,
+                                 num_layers=32)
+    """
+
+    _instance: ClassVar[Optional['MoEMetricsTracker']] = None
+
+    def __init__(self):
+        self._metrics: Dict[str, MetricEntry] = {}
```
Victarry, line 83: In this design, I think one `MoEMetricsTracker` should only server one MoE model? 
If so, how about make `num_layers/moe_layer_freq/mtp_num_layers` as the class member. Such that we don't need to pass that variables in other places.

yanring: Good point!

megatron/training/training.py
buptzyb:
```suggestion
                config.moe_metrics_tracker.clear()
```

megatron/training/training.py
```diff
@@ -142,7 +142,8 @@ def set_startup_timestamps(program_start=None, main_entry=None):
 from megatron.core.datasets.data_schedule import HybridCPDataLoaderWrapper
 from megatron.core.optimizer_param_scheduler import OptimizerParamScheduler
 from megatron.core.transformer.moe import upcycling_utils
-from megatron.core.transformer.moe.moe_utils import track_moe_metrics, clear_aux_losses_tracker
+from megatron.core.transformer.moe.moe_logging import MoEMetricsTracker
+from megatron.core.transformer.moe.moe_utils import clear_aux_losses_tracker
```
buptzyb:
```suggestion
```

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1014,8 +1011,9 @@ def clone_ten(ten):
                     param.main_grad.copy_(main_grad_copy)
 
         if is_moe:
-            for name in tracker:
-                tracker[name]["values"].copy_(cached_aux_losses[name])
+            for name, cached_values in cached_aux_losses.items():
+                if name in moe_metrics_tracker.metrics:
+                    moe_metrics_tracker.metrics[name].values.copy_(cached_values)
```
buptzyb:
```suggestion
                assert name in moe_metrics_tracker.metrics, "cached metrics must be found in the tracker."
                moe_metrics_tracker.metrics[name].values.copy_(cached_values)
```

yanring:
/ok to test [26a5025](https://github.com/NVIDIA/Megatron-LM/pull/2569/commits/26a5025220aa4bb4180fc89b541f273f36d9827b)

megatron/core/transformer/cuda_graphs.py
```diff
@@ -819,13 +819,10 @@ def create_fwd_graph(self, args, kwargs, outputs=None, clone_inputs=True):
...
         if is_moe:
-            from megatron.core.transformer.moe.moe_utils import get_moe_layer_wise_logging_tracker
-
-            tracker = get_moe_layer_wise_logging_tracker()
+            moe_metrics_tracker = self.base_module.config.moe_metrics_tracker
             cached_aux_losses = {}
-            for name in tracker:
-                if "values" in tracker[name]:
-                    cached_aux_losses[name] = torch.clone(tracker[name]["values"])
+            for name, entry in moe_metrics_tracker.metrics.items():
```
jiemingz, line 826: could this please be refactored as well?

megatron/core/transformer/moe/moe_logging.py
yanring: After more thought, I decided to keep the tracker as a module-level global variable (following the parallel_state / global_vars pattern) rather than putting it on TransformerConfig.                                                                                                
                                                                                                                                              
Storing a stateful, mutable object on a dataclass config feels like it breaks the convention that config fields should be declarative/static. The tracker is fundamentally runtime state, not configuration.

For the multi-model concern: running multiple models of different sizes in a single process isn't a common pattern today. Running multiple models of the same size should work fine (metrics just merge into the same log). If true per-model isolation is ever needed, we can revisit.

yanring:
/ok to test [aa234a3](https://github.com/NVIDIA/Megatron-LM/pull/2569/commits/aa234a36b821ba95ce09ed20977324cb280241a7)

megatron/core/transformer/moe/moe_logging.py
Victarry: I see. That's reasonable.

PR #2571, opened 2025-12-05
Fix world size mismatch causing distributed init deadlock (Issue #2458)

### LINKED ISSUE #2458: Incorrect world_size logic in run_simple_mcore_train_loop.py causes deadlock
**Describe the bug**

In `run_simple_mcore_train_loop.py`, the script derives `world_size` from `torch.cuda.device_count()`. When launching with fewer processes than available GPUs, this results in a mismatch between `world_size` and `WORLD_SIZE` provided by `torchrun`, causing `init_process_group` to deadlock.

Example: running with `--nproc_per_node=2` on an 8-GPU machine hangs during initialization.

1. Use a machine with more GPUs than the number of processes to launch.

2. Run:

```bash
torchrun --nproc_per_node=2 examples/run_simple_mcore_train_loop.py
```

3. The script hangs at:

```python
torch.distributed.init_process_group(world_size=world_size, rank=rank)
```

**Expected behavior**

The example should initialize without deadlock.

Using:

```python
world_size = int(os.environ["WORLD_SIZE"])
```

avoids the mismatch.

**Additional context**

This may be related to issue #881.

CodersAcademy006:
This PR fixes Issue #2458 by removing the incorrect world-size calculation based on the number of CUDA devices and replacing it with the correct environment-derived values (`WORLD_SIZE`, `RANK`, `LOCAL_RANK`) provided by `torchrun`, ensuring that `init_process_group` initializes properly without deadlocking when the number of launched processes is smaller than the number of available GPUs. Fixes #2458.


Skylion007:
Why not just let torch auto detect by setting init method to env and use the get_rank etc helpers?

PR #2572, opened 2025-12-05
Various CUDA graph improvements on capture time, replay time, memory footprint

jiemingz:
# What does this PR do ?
- improves training capture time
- supports cudagraphing module functions (for partial moe cudagraphs)
- supports partially cudagraphing moe router and postprocess
- generalizes cudagraph input output buffer reuse
- supports cudagraph input and output buffer recycling and deallocation to reduce memory overhead for training
- misc improvements to reduce cudagraph replay overhead 

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -1688,10 +1688,11 @@ class TransformerConfig(ModelParallelConfig):
                         self.cuda_graph_scope.append(CudaGraphScope.moe_preprocess)
 
             # Check cuda graph scopes
-            assert CudaGraphScope.full_iteration not in self.cuda_graph_scope, (
-                "To use full iteration cuda graph, please use "
-                "cuda_graph_impl=local instead of cuda_graph_impl=transformer_engine."
-            )
+            if self.cuda_graph_impl == "transformer_engine":
+                assert CudaGraphScope.full_iteration not in self.cuda_graph_scope, (
+                    "To use full iteration cuda graph, please use "
+                    "cuda_graph_impl=local instead of cuda_graph_impl=transformer_engine."
+                )
             assert (
                 CudaGraphScope.moe not in self.cuda_graph_scope
                 or CudaGraphScope.moe_router not in self.cuda_graph_scope


megatron/core/ssm/mamba_layer.py
```diff
@@ -84,6 +84,11 @@ def __init__(
         self.mamba_bda = build_module(submodules.mamba_bda)
         self.bias_dropout_add_exec_handler = torch.enable_grad
 
+    def create_mcore_cudagraph_manager(self, config):
+        from megatron.core.transformer.cuda_graphs import CudaGraphManager
+        if "full" in self.config.cuda_graph_scope or "mamba" in self.config.cuda_graph_scope:
```
gautham-kollu: Can we use CudaGraphScope ? Possibly map "full" -> to a combination of the layers defined

megatron/core/transformer/transformer_layer.py
```diff
@@ -406,6 +406,19 @@ def __init__(
         # self.bias_dropout_add_exec_handler = nullcontext if use_nvfuser else torch.enable_grad
         self.bias_dropout_add_exec_handler = torch.enable_grad
 
+    def create_mcore_cudagraph_manager(self, config):
+        from megatron.core.transformer.cuda_graphs import CudaGraphManager
+        # If full scope, just cudagraph the entire layer
+        if "full" in self.config.cuda_graph_scope:
```
mathemakitten:
```suggestion
        if "full" in self.config.cuda_graph_scope or self.config.cuda_graph_scope == []:
```

megatron/core/transformer/cuda_graphs.py
```diff
@@ -599,282 +581,326 @@ def __init__(
...
-                kwargs_copy = dict(kwargs)
-                kwargs_copy['hidden_states'] = None
-                _CudagraphGlobalRecord.record_fwd_graph(self, args, kwargs_copy)
-            else:
-                _CudagraphGlobalRecord.record_fwd_graph(self, args, kwargs)
-
-            self.fwd_graph_recorded = True
+        for t in self.get_tensors(args, kwargs):
+            t.is_cudagraph_input = True
+            if hasattr(t, "is_cudagraph_output"):
```
mathemakitten: Inference cudagraphs are now created as they are encountered (instead of waiting on a record-and-wait-until-create-cudagraphs-call like in training), which is needed because recording them along with training graphs and then calling `create_cudagraphs` all together was very buggy. Is it possible to move this logic into a helper which is called within or ahead of `create_fwd_graph`?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1327,11 +1351,8 @@ def __call__(self, megatron_module, args, kwargs):
                     # If the layer is frozen, we need to set the runner to eval mode.
                     runner.eval()
                 out = runner.record_graph_capture(args, kwargs)
-            else:
```
mathemakitten, line 1390: Why do we remove the eager fallback? The case where validation executes before training graphs get captured will now fail without this.

mathemakitten:
Can we throw an error or at least a warning if `HAVE_TE_GRAPHS = False` now since none of the new weakref-based code works without HAVE_TE_GRAPHS? Some of the older containers silently error on `from transformer_engine.pytorch.utils import make_weak_ref` but proceed like usual and cudagraphs will not execute but the logs provide no obvious reason why.

megatron/core/transformer/cuda_graphs.py
jiemingz: I placed this code into a `apply_cudagraph_artifacts`
record_graph_capture is only used from training, since inference cudagraphs are created as they are encountered as you say. So inference cudagraphs should still be using the previous reuse logic which will be unaffected.

megatron/core/ssm/mamba_layer.py
mathemakitten: This similarly needs the `or self.config.cuda_graph_scope = []` case for backward compatibility

megatron/core/transformer/cuda_graphs.py
```diff
@@ -90,6 +96,26 @@ def _set_capture_end():
...
+class CudagraphArtifacts:
+    is_cudagraph_input: bool = False
+    is_cudagraph_output: bool = False
+    input_use_count: int = 0
+    cudagraph_reuse_ref_count: int = 0
+    capture_reuse_count: int = 0
+    fwd_cudagraph_buffer: torch.Tensor = None
+    bwd_cudagraph_buffer: torch.Tensor = None
+    
+    g_id = -1
```
mathemakitten: nit: can we make this `meta_id` or something for readability? Not really sure what's optimal here but I think `meta_id` would be an improvement over `g_id`

megatron/core/transformer/cuda_graphs.py
```diff
@@ -90,6 +96,26 @@ def _set_capture_end():
     _IS_GRAPH_CAPTURING = False
 
 
+artifact_count = 0
+
+@dataclass
+class CudagraphArtifacts:
```
mathemakitten: I find this naming confusing — this object is really per‑tensor CUDA‑graph I/O buffer/usage metadata, not really an "artifact", right? And each class instance is only attached to a single tensor as opposed to a collection, right? So technically this is a singular `CudagraphArtifact`? Or maybe more accurately,  `CudagraphTensorMetadata` or `GraphTensorMetadata`? Or `CudagraphBufferMetadata` or `GraphBufferMetadata`? Lots of options which are more informative than "artifact".

megatron/core/transformer/cuda_graphs.py
```diff
@@ -90,6 +96,26 @@ def _set_capture_end():
...
+    capture_reuse_count: int = 0
+    fwd_cudagraph_buffer: torch.Tensor = None
+    bwd_cudagraph_buffer: torch.Tensor = None
+    
+    g_id = -1
+
+    def __init__(self):
+        global artifact_count
+        self.g_id = artifact_count
+        artifact_count += 1
```
mathemakitten: Similarly here, something like `tensor_meta_count` would be way more readable

megatron/core/transformer/cuda_graphs.py
```diff
@@ -225,22 +272,10 @@ def create_cudagraphs(cls):
...
-        optimize_transformer_layer_graph_buffers = all(
-            [g[0].reuse_input_output_buffer for g in cls.cudagraph_record]
-        )
-        if optimize_transformer_layer_graph_buffers:
-            prev_fwd_hidden_state_output = None
-            prev_bwd_hidden_state_inputgrad = None
+        else:
+            logging.warning("Transformer Engine was not detected while capturing training " \
+            "cudagraphs. As a result cudagraph memory overhead may significantly increase " \
+            "as Transformer Engine is needed to remove redundant cudagraph memory usage.")
```
mathemakitten: Can we drop a note about the weakref thing here so at least if you encounter this you know what part is actually affected by TE? Specifically, `make_weak_ref` is only available in new-ish TE versions.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -978,100 +1113,85 @@ def check(val, ref, context):
...
-    def __init__(
-        self,
-        config: TransformerConfig,
-        share_cudagraph_io_buffers: bool = True,
-        vp_stage: Optional[int] = None,
+    def __init__(self, 
+        config: TransformerConfig, 
+        base_module = None,
+        function_name = None,
+        need_backward = True,
```
mathemakitten: Why is this needed when we have `base_module.training` and/or `torch.is_grad_enabled()` available in the same scope?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -275,39 +310,16 @@ def format_mem_bytes(mem_bytes):
             elif g_idx % 100 == 0 or g_idx == len(cls.cudagraph_record) - 1:
                 logger.info(f"{g_idx}/{len(cls.cudagraph_record)}. {progress_str}")
```
deepakn94: Not directly related to this PR, but we can only log these on the first / last rank to reduce log clutter?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -275,39 +310,16 @@ def format_mem_bytes(mem_bytes):
...
-                if graph_type == 'fwd':
-                    args, kwargs = g[2:]
-                    runner.create_fwd_graph(args, kwargs)
-                else:
-                    runner.create_bwd_graph()
+                assert fwd_buffer_reuse_ref_count == 0
+                runner.create_bwd_graph()
+
+        # TODO: (jiemingz) see [interaction with recompute]
+        # assert bwd_buffer_reuse_ref_count == 0
```
deepakn94: Dead code.

megatron/core/transformer/cuda_graphs.py
jiemingz: This is for when you want to capture something that doesnt have a backward pass but still is run inside a training loop. For instance, we might want to graph over a `log_grad_norms` or something. Hypothetically there isn't a backward pass to graph over. So this is a flag to do that, without having to set `base_module.training and/or torch.is_grad_enabled()` everytime we call that function

megatron/core/transformer/cuda_graphs.py
jiemingz: I just got rid of it- I don't think this attribute is actually going to get used moving forward besides for debugging.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -276,39 +356,13 @@ def format_mem_bytes(mem_bytes):
             elif g_idx % 100 == 0 or g_idx == len(cls.cudagraph_record) - 1:
                 logger.info(f"{g_idx}/{len(cls.cudagraph_record)}. {progress_str}")
```
deepakn94: Can we also only log this on rank 0?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -324,6 +378,7 @@ def format_mem_bytes(mem_bytes):
                 - mem_stats_start["reserved_bytes.all.current"]
             ),
         }
+
         logger.info(
```
deepakn94: And this!

kvareddy (approved):

deepakn94 (approved):

megatron/training/arguments.py
```diff
@@ -1254,13 +1254,14 @@ def validate_args(args, defaults={}):
...
             args.te_rng_tracker = True
             warn_rank_0("te_rng_tracker is not enabled, enabling it for CUDA graphs.", args.rank)
-        assert (
-            "expandable_segments:True" not in os.getenv("PYTORCH_CUDA_ALLOC_CONF", "")
-            or os.getenv("NCCL_GRAPH_REGISTER", "") == "0"
-        ), (
-            "Setting NCCL_GRAPH_REGISTER=0 to avoid illegal memory access when using "
-            "CUDA Graph with PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True."
-        )
+        if args.cuda_graph_impl == "transformer_engine":
```
buptzyb, line 1342: local implementation won't run into illegal memory access?

megatron/core/models/mamba/mamba_layer_specs.py
```diff
@@ -78,8 +82,7 @@
             ),
         ),
         moe_layer=ModuleSpec(
-            # TODO (rwaleffe): change this to be an "MoELayer" to work with CudaGraphs?
-            module=TransformerLayer,
+            module=MoETransformerLayer,
```
buptzyb, line 85: Can using MoETransformerLayer still support te cudagraph partial capturing?

megatron/core/transformer/transformer_layer.py
```diff
@@ -723,6 +748,14 @@ def _forward_post_mlp(self, mlp_output_with_bias, residual):
         Returns:
             output (Tensor): Transformed hidden states of shape [s, b, h].
         """
+
+        if self.recompute_pre_mlp_layernorm:
```
buptzyb, line 783: Why move this to `_forward_post_mlp`? TE cudagraph capturing won't reach here so this will make it miss the pre mlp output discard.

megatron/training/arguments.py
jiemingz: yes we were able to successfully train with expandable segments

megatron/core/models/mamba/mamba_layer_specs.py
jiemingz: yes since MoeTransformerLayer is a child class of TransformerLayer

megatron/core/transformer/transformer_layer.py
jiemingz: I've made a change so that this behavior is programatically identical to whats currently there:
on capture, it discards the entire mlp output as before
on replay, the discard is disabled in the `_forward_post_mlp`

megatron/training/arguments.py
buptzyb: Wow, you have adopted this solution successfully? https://github.com/pytorch/pytorch/issues/158029#issuecomment-3535048207

megatron/core/transformer/transformer_layer.py
```diff
@@ -1006,7 +1039,16 @@ def _te_cuda_graph_replay(self, *args, **kwargs):
...
             nvtx_range_pop(suffix="mlp")
 
+            # If we early returned, layernorm recompute hooks were attached to the output buffer
+            # of the cudagraph, so disable the recompute hooks inside _forward_post_mlp
+            recompute_pre_mlp_layernorms = self.recompute_pre_mlp_layernorm
+            self.recompute_pre_mlp_layernorm = False
+
             output = self._forward_post_mlp(mlp_output_with_bias, residual)
+
+            recompute_pre_mlp_layernorms = self.recompute_pre_mlp_layernorm
```
buptzyb:
```suggestion
```

megatron/training/arguments.py
jiemingz: we didn't have this...but in the case of GB200/B200, we've never observed errors from expandable segments over very thorough testing so it seems ok?

megatron/core/tensor_parallel/random.py
```diff
@@ -624,11 +660,13 @@ def checkpoint(self, run_function, *args):
     def _recompute(self, _):
         """Used as a hook to recompute the output."""
 
-        if self.ctx is None:
```
buptzyb, line 631: why remove this condition?

megatron/core/tensor_parallel/random.py
```diff
@@ -624,11 +660,13 @@ def checkpoint(self, run_function, *args):
...
-            # The recomputation has been triggered already. Just return.
+        # Handle cudagraphs, do nothing if currently in graph warmup
+        from megatron.core.transformer.cuda_graphs import is_graph_capturing, is_graph_warmup
+
+        if is_graph_warmup():
             return
 
-        if not torch.autograd._is_checkpoint_valid():
+        if not torch.autograd._is_checkpoint_valid() and not is_graph_capturing():
```
buptzyb, line 674: Are you sure you want to continue capturing in an invalid checkpointing status? Is this safe?

megatron/core/transformer/transformer_layer.py
```diff
@@ -1006,6 +1034,10 @@ def _te_cuda_graph_replay(self, *args, **kwargs):
             self.mlp.cudagraph_tensor_store.clear()
             nvtx_range_pop(suffix="mlp")
 
+            # If we early returned, layernorm recompute hooks were attached to the output buffer
+            # of the cudagraph, so disable the recompute hooks inside _forward_post_mlp
+            self.recompute_pre_mlp_layernorm = False
```
buptzyb, line 1089: Would this be unsafe if it's set to False and then the next forward somehow run into the normal codepath instead of cudagraph replay? To this end I would prefre restore the value after `_forward_post_mlp`.

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -290,19 +294,23 @@ def routed_experts_compute(self, hidden_states: torch.Tensor, probs: torch.Tenso
...
         return output, mlp_bias
 
-    def combine(self, output: torch.Tensor, shared_expert_output: Optional[torch.Tensor]):
+    def combine(self, output: torch.Tensor):
         """Combines expert outputs via communication and adds shared expert output.
 
         This method uses the token dispatcher to combine the outputs from different
-        experts (e.g., via an All-to-All communication). It then adds the output
-        from the shared expert if it exists.
+        experts (e.g., via an All-to-All communication)
```
buptzyb:
```suggestion
        experts (e.g., via an All-to-All communication).
```

megatron/core/tensor_parallel/random.py
jiemingz: yes this is safe because of how we capture the graph and how recompute handles the fwd passes
The condition this is trying to avoid does not apply for our construction of cudagraphs:
[https://github.com/pytorch/pytorch/blob/main/torch/autograd/__init__.py#L558](https://github.com/pytorch/pytorch/blob/main/torch/autograd/__init__.py#L558)

megatron/core/transformer/transformer_layer.py
jiemingz: I've reverted this to what I orginally had then.

megatron/core/tensor_parallel/random.py
jiemingz: thanks this was a rebase error

sidsingh-nvidia:
@jiemingz This PR is breaking mcore-inference with nanov3. Interestingly, the branch works with qwen3-30b-a3b, so I think it's a mamba+moe specific issue. I see the following error during cuda-graph capture. 

`[rank7]: RecursionError: maximum recursion depth exceeded`

I will send you a reproducer over slack. 

cc: @kvareddy @mathemakitten

jiemingz:
> @jiemingz This PR is breaking mcore-inference with nanov3. Interestingly, the branch works with qwen3-30b-a3b, so I think it's a mamba+moe specific issue. I see the following error during cuda-graph capture.
> 
> `[rank7]: RecursionError: maximum recursion depth exceeded`
> 
> I will send you a reproducer over slack.
> 
> cc: @kvareddy @mathemakitten

I pushed a fix to resolve this, the reason this occurs is because nanov3 uses the new hybrid model spec, which eventually pipes down into the partial cudagraph code path for training, which isn't compatible with inference.

yanring (approved):

mathemakitten (approved):

mathemakitten (approved):
LGTM

ericharper:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/2572/commits/5d0fc629af755ec0309999a61fdc269059294806

PR #2580, opened 2025-12-06
Various quality-of-life improvements in training loop

deepakn94:
- Skip optimizer and main_grad creation to save memory when --skip-train argument is provided
- Enforce that --async-save is called with the right accompanying arguments
- Only log warning messages on rank 0

### DIFF
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -1147,24 +1147,28 @@ def setup_model_and_optimizer(
     timers = get_timers()
     one_logger = get_one_logger()
 
-    model = get_model(model_provider_func, model_type)
+    wrap_with_ddp = not args.skip_train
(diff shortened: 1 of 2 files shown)


Skylion007:
As a more meta point, it would be great if some of this validation also applied to the configs coming from Megatron-Bridge or YAML parser. Long term wish there was a nice Hydrabackend for this.

megatron/training/training.py
```diff
@@ -1149,24 +1147,28 @@ def setup_model_and_optimizer(
     timers = get_timers()
     one_logger = get_one_logger()
 
-    model = get_model(model_provider_func, model_type)
+    wrap_with_ddp = not args.skip_train
```
ananthsub, line 1150: will this be an issue if megatron FSDP is used, since the fsdp_dtensor ckpt format is expected then?

PR #2632, opened 2025-12-11
Pipeline parallelism fix in RL and sequence packing rewriting

jalbericiola:
# What does this PR do ?

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/rl/rl_utils.py
+++ b/megatron/rl/rl_utils.py
@@ -189,50 +189,57 @@ def align_unpacked_inference_logprobs(
     Returns:
         Aligned inference logprobs tensor
     """
-    # Get first occurrence of a generation token
-    # In get_logprobs() we chop off the first token -> the generation mask is shifted by one
-    gen_masks_for_alignment = generation_masks
-    first_gen_tok = gen_masks_for_alignment.int().argmax(dim=1) - 1
+    nvtx_range = get_nvtx_range()
+    
+    with nvtx_range("rl/align-logprobs/total", log_level=2):
+        # Get first occurrence of a generation token
+        # In get_logprobs() we chop off the first token -> the generation mask is shifted by one
+        gen_masks_for_alignment = generation_masks
+        first_gen_tok = gen_masks_for_alignment.int().argmax(dim=1) - 1
 
-    # Align inference logprobs with old_logprobs
-    # Note: We use old_logprobs_for_data as template since it has correct shape
-    padded_inference_logprobs = old_logprobs_for_data.clone()
+        # Align inference logprobs with old_logprobs
+        # Note: We use old_logprobs_for_data as template since it has correct shape
+        with nvtx_range("rl/align-logprobs/clone", log_level=3):
+            padded_inference_logprobs = old_logprobs_for_data.clone()
 
-    # We need to align old_logprobs and inference logprobs as the latter are only for generations
-    for i, inf_logprobs in enumerate(inference_logprobs):
-        first_gen_idx = first_gen_tok[i]
-        # We subtract -1 here because we append eod token on the train side, and we do not
-        # get it from the inference. For the eod token, we reuse old_logprobs value.
-        end_idx = min(first_gen_idx + len(inf_logprobs), padded_inference_logprobs.shape[1])
-        actual_len = end_idx - first_gen_idx
-        if actual_len > 0:
-            padded_inference_logprobs[i, first_gen_idx:end_idx] = inf_logprobs[:actual_len]
+        # We need to align old_logprobs and inference logprobs as the latter are only for generations
+        with nvtx_range("rl/align-logprobs/loop", log_level=3):
+            for i, inf_logprobs in enumerate(inference_logprobs):
+                first_gen_idx = first_gen_tok[i]
+                # We subtract -1 here because we append eod token on the train side, and we do not
+                # get it from the inference. For the eod token, we reuse old_logprobs value.
+                end_idx = min(first_gen_idx + len(inf_logprobs), padded_inference_logprobs.shape[1])
+                actual_len = end_idx - first_gen_idx
+                if actual_len > 0:
+                    padded_inference_logprobs[i, first_gen_idx:end_idx] = inf_logprobs[:actual_len]
 
-    # Create truncated mask for statistics
-    if old_logprobs_for_data.shape[1] + 1 < gen_masks_for_alignment.shape[1]:
-        gen_masks_for_alignment = gen_masks_for_alignment[:, : old_logprobs_for_data.shape[1] + 1]
+        # Create truncated mask for statistics
+        with nvtx_range("rl/align-logprobs/mask", log_level=3):
+            if old_logprobs_for_data.shape[1] + 1 < gen_masks_for_alignment.shape[1]:
+                gen_masks_for_alignment = gen_masks_for_alignment[:, : old_logprobs_for_data.shape[1] + 1]
 
-    truncated_mask = gen_masks_for_alignment[:, 1:].bool()
+            truncated_mask = gen_masks_for_alignment[:, 1:].bool()
 
-    # Final safety check
-    if truncated_mask.shape != old_logprobs_for_data.shape:
-        if truncated_mask.shape[1] > old_logprobs_for_data.shape[1]:
-            truncated_mask = truncated_mask[:, : old_logprobs_for_data.shape[1]]
-        elif truncated_mask.shape[1] < old_logprobs_for_data.shape[1]:
-            pad_size = old_logprobs_for_data.shape[1] - truncated_mask.shape[1]
-            truncated_mask = torch.nn.functional.pad(truncated_mask, (0, pad_size), value=False)
+            # Final safety check
+            if truncated_mask.shape != old_logprobs_for_data.shape:
+                if truncated_mask.shape[1] > old_logprobs_for_data.shape[1]:
+                    truncated_mask = truncated_mask[:, : old_logprobs_for_data.shape[1]]
+                elif truncated_mask.shape[1] < old_logprobs_for_data.shape[1]:
+                    pad_size = old_logprobs_for_data.shape[1] - truncated_mask.shape[1]
+                    truncated_mask = torch.nn.functional.pad(truncated_mask, (0, pad_size), value=False)
 
-    # Sanity check: Two probability values cannot be more than 1.0 apart
-    abs_diffs = (old_logprobs_for_data.exp() - padded_inference_logprobs.exp()).abs()[truncated_mask]
-    assert all(abs_diffs <= 1.0)
+        # Sanity check: Two probability values cannot be more than 1.0 apart
+        abs_diffs = (old_logprobs_for_data.exp() - padded_inference_logprobs.exp()).abs()[truncated_mask]
+        assert all(abs_diffs <= 1.0)
 
-    # Update group statistics using common helper
-    update_inference_logprobs_group_stats(
-        old_logprobs=old_logprobs_for_data,
-        inference_logprobs=padded_inference_logprobs,
-        mask=truncated_mask,
-        group_stats=group_stats,
-    )
+        # Update group statistics using common helper
+        with nvtx_range("rl/align-logprobs/stats", log_level=3):
+            update_inference_logprobs_group_stats(
+                old_logprobs=old_logprobs_for_data,
+                inference_logprobs=padded_inference_logprobs,
+                mask=truncated_mask,
+                group_stats=group_stats,
+            )
 
     return padded_inference_logprobs
 
@@ -254,37 +261,39 @@ _INFERENCE_INTERFACE = None
 
 def get_inference_interface(args, loop, model):
     global _INFERENCE_INTERFACE
+    nvtx_range = get_nvtx_range()
     if _INFERENCE_INTERFACE is None:
-        rank = torch.distributed.get_rank()
-        if rank == 0 and args.langrl_external_server:
-            if args.langrl_inference_server_type == 'inplace_megatron':
-                _INFERENCE_INTERFACE = loop.run_until_complete(
-                    InferenceInterfaceServer.launch(MegatronLocal, model=model[0])
-                )
-            elif args.langrl_inference_server_type == 'inplace_megatron_chat':
-                _INFERENCE_INTERFACE = loop.run_until_complete(
-                    InferenceInterfaceServer.launch(
-                        MegatronChatLocal,
-                        model=model[0],
-                        conversation_template=args.langrl_inference_server_conversation_template,
+        with nvtx_range("rl/inference-mode/get-interface/launch", log_level=2):
+            rank = torch.distributed.get_rank()
+            if rank == 0 and args.langrl_external_server:
+                if args.langrl_inference_server_type == 'inplace_megatron':
+                    _INFERENCE_INTERFACE = loop.run_until_complete(
+                        InferenceInterfaceServer.launch(MegatronLocal, model=model[0])
                     )
-                )
-            else:
-                raise ValueError(f"Unknown inference_server_type {args.inference_server_type}")
-        else:
-            if args.langrl_inference_server_type == 'inplace_megatron':
-                _INFERENCE_INTERFACE = loop.run_until_complete(MegatronLocal.launch(model[0]))
-            elif args.langrl_inference_server_type == 'inplace_megatron_chat':
-                _INFERENCE_INTERFACE = loop.run_until_complete(
-                    MegatronChatLocal.launch(
-                        model[0],
-                        conversation_template=args.langrl_inference_server_conversation_template,
+                elif args.langrl_inference_server_type == 'inplace_megatron_chat':
+                    _INFERENCE_INTERFACE = loop.run_until_complete(
+                        InferenceInterfaceServer.launch(
+                            MegatronChatLocal,
+                            model=model[0],
+                            conversation_template=args.langrl_inference_server_conversation_template,
+                        )
                     )
-                )
+                else:
+                    raise ValueError(f"Unknown inference_server_type {args.inference_server_type}")
             else:
-                raise ValueError(
-                    f"Unknown inference_server_type {args.langrl_inference_server_type}"
-                )
+                if args.langrl_inference_server_type == 'inplace_megatron':
+                    _INFERENCE_INTERFACE = loop.run_until_complete(MegatronLocal.launch(model[0]))
+                elif args.langrl_inference_server_type == 'inplace_megatron_chat':
+                    _INFERENCE_INTERFACE = loop.run_until_complete(
+                        MegatronChatLocal.launch(
+                            model[0],
+                            conversation_template=args.langrl_inference_server_conversation_template,
+                        )
+                    )
+                else:
+                    raise ValueError(
+                        f"Unknown inference_server_type {args.langrl_inference_server_type}"
+                    )
     return _INFERENCE_INTERFACE
 
 
@@ -293,21 +302,24 @@ _ROLLOUT_GENERATOR = None
 
 def get_rollout_generator(args, inference_interface, n_prompts, samples_per_group):
     global _ROLLOUT_GENERATOR
+    nvtx_range = get_nvtx_range()
     if not args.rl_partial_rollouts or _ROLLOUT_GENERATOR is None:
-        agent = get_agent(args)
+        with nvtx_range("rl/rollout-collection/create-agent", log_level=2):
+            agent = get_agent(args)
         # Collect Rollouts
-        request = GroupedRolloutRequest(
-            num_groups=-1 if args.rl_partial_rollouts else n_prompts,
-            rollouts_per_group=samples_per_group,
-            inference_interface=inference_interface,
-            generation_args={
-                'temperature': args.grpo_default_temperature,
-                'max_tokens': args.inference_max_seq_length,
-                'top_p': args.grpo_default_top_p,
-            },
-            filter_groups_with_same_reward=args.grpo_filter_groups_with_same_reward,
-        )
-        _ROLLOUT_GENERATOR = agent.get_grouped_rollouts(request)
+        with nvtx_range("rl/rollout-collection/create-request", log_level=2):
+            request = GroupedRolloutRequest(
+                num_groups=-1 if args.rl_partial_rollouts else n_prompts,
+                rollouts_per_group=samples_per_group,
+                inference_interface=inference_interface,
+                generation_args={
+                    'temperature': args.grpo_default_temperature,
+                    'max_tokens': args.inference_max_seq_length,
+                    'top_p': args.grpo_default_top_p,
+                },
+                filter_groups_with_same_reward=args.grpo_filter_groups_with_same_reward,
+            )
+            _ROLLOUT_GENERATOR = agent.get_grouped_rollouts(request)
     return _ROLLOUT_GENERATOR
 
 
@@ -326,14 +338,15 @@ def get_environment_rollouts(
     """
     args = get_args()
     nvtx_range = get_nvtx_range()
+    timers = get_timers()
 
     assert (
         n_prompts % mpu.get_expert_data_parallel_world_size() == 0
     ), "n_prompts must be divisible by data_parallel_world_size"
 
-    with nvtx_range("rollout-collection"):
+    with nvtx_range("rl/rollout-collection", log_level=1):
         loop = get_asyncio_loop()
-        with megatron_rl_inference_mode(
+        with nvtx_range('rl-enter-inference-mode', log_level=1), megatron_rl_inference_mode(
             model,
             optimizer,
             args.cuda_graph_impl,
@@ -343,7 +356,7 @@ def get_environment_rollouts(
             args.rl_remove_kv_cache_during_training,
         ) as inference_interface:
 
-            with nvtx_range("inference-setup"):
+            with nvtx_range("rl/inference-mode/setup", log_level=1):
                 # Asyncronously run inference and rollout collection
                 rollout_generator = get_rollout_generator(
                     args, inference_interface, n_prompts, samples_per_group
@@ -351,7 +364,7 @@ def get_environment_rollouts(
 
             # NOTE(jbarker): we need to double check this when using PP>1
             rank = torch.distributed.get_rank()
-            with nvtx_range("collect-rollouts"):
+            with nvtx_range("rl/rollout-collection/generate", log_level=1):
                 if rank == 0:
                     log_single_rank(
                         logger,
@@ -372,7 +385,9 @@ def get_environment_rollouts(
                     # Just set up space to collect the rollouts
                     rollouts = [[None for _ in range(samples_per_group)] for _ in range(n_prompts)]
 
-        with nvtx_range("sync-rollouts"):
+            pass  # exit inference mode (nvtx_range handles timing at context manager level)
+
+        with nvtx_range("rl/rollout-collection/sync", log_level=1):
             # Wait for Rollouts to be collected
             # TODO(jbarker): double check why this isn't causing rank 0 memory allocations
             torch.distributed.broadcast_object_list(rollouts, src=0)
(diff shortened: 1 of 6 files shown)


megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -686,7 +686,7 @@ def _add_request(
             request.sampling_params.return_log_probs
             and not request.sampling_params.skip_prompt_log_probs
         ):
-            assert not self.context.materialize_only_last_token_logits, (
+            assert not self.materialize_only_last_token_logits, (
```
tdene: Remove this diff.

megatron/core/inference/sampling_params.py
```diff
@@ -50,7 +50,7 @@ def _sync_prompt_logprobs_fields(self):
             assert (
                 self.skip_prompt_log_probs
             ), "return_prompt_top_n_logprobs requires skip_prompt_log_probs to be False"
-        if self.top_n_logprobs > 0:
+        if self.return_log_probs:
```
tdene: Remove this diff

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1189,7 +1189,7 @@ async def async_bookkeep(
                 )
 
         # Print context state.
-        if verbose:
+        if verbose and self.inference_logging_step_interval and step_count % self.inference_logging_step_interval == 1:
```
tdene: This should be `if verbose and (self.inference_logging_step_interval == 0 or step_count % self.inference_logging_step_interval == 1):`

Before this PR, `verbose=True` alone would turn on logging.

After this PR, the user needs to specify both `verbose=True` and `inference_logging_step_interval=1` to turn on logging. This changes pre-existing flows, and I see that as unnecessary.

tdene: @lmcafee-nvidia is inference okay with making the engine's `verbose` argument not print on every step, but instead be controller by `inference_logging_step_interval`?

Previously, `inference_logging_step_interval` only controlled the interval at which we would perform wandb logging. This change makes it affect all verbose logging.

examples/rl/model_configs/common.sh
```diff
@@ -22,7 +22,7 @@ COMMON_OPTIONS="\
     --attention-backend flash \
     --timing-log-level 1 \
     --log-timers-to-tensorboard \
-    --save-retain-interval 120 \
+    --save-retain-interval 128 \
```
yobibyte: This will potentially break several commands in the justfile. Do we need this change?

tests/unit_tests/test_rl_utils.py
```diff
@@ -305,72 +313,6 @@ def test_grpo_loss_truncation():
...
-    rollouts = [[r1, r2]]
-    data_iter = rl_utils.prepare_data_for_update([model], {}, rollouts, tokenizer)
-
-    _, _, old_logprobs, _, _, _, _ = next(data_iter)
-    # All logits are ones in the MockModel.
-    # All probabilities should be uniform.
-    torch.testing.assert_close(old_logprobs.exp(), torch.ones_like(old_logprobs) / VOCAB)
-
-
 def test_sequence_packing_basic():
```
yobibyte: This test is failing on the PR branch:

```
FAILED megatron-rl/tests/unit_tests/test_rl_utils.py::test_sequence_packing_basic - AttributeError: 'list' object has no attribute 'shape'
```

yobibyte: test_sequence_packing_with_generation_masks fails too, please, rerun unit tests.

tests/unit_tests/test_rl_utils.py
```diff
@@ -305,72 +313,6 @@ def test_grpo_loss_truncation():
     torch.testing.assert_close(truncated_from_below, torch.tensor([[False, True], [False, False]]))
 
 
-@patch('megatron.rl.rl_utils.mpu')
```
yobibyte, line 308: I think it is unreasonable to get rid of this test.

We can use the test Utils to initialize process groups:

Add this to imports:
```
os.environ['WANDB_MODE'] = "disabled"
os.environ['LOG_TO_WANDB'] = "false"
os.environ['WORLD_SIZE'] = "1"
os.environ['LOCAL_RANK'] = "0"
from tests.unit_tests.test_utilities import Utils
```
Add this after the model creation:
```
Utils.initialize_model_parallel()
```

And this to the end of test:
```
```

You might need to add a couple of attributes to the MockModel, but I think we can salvage this test.

train_rl.py
```diff
@@ -199,70 +204,24 @@ def forward_step(data_iterator, model: GPTModel, loss_only: bool = False):
...
+            old_logprobs,
+            loss_mask,
+            position_ids,
+            ref_logprobs,
+            inference_logprobs,
+            seq_starts,
+            seq_lengths,
+            seq_indices,
+            packed_seq_params,
+        ) = load_packed_data_by_index(bin_tensor.item(), runtime_state.packing_context)
```
yobibyte: NICE

megatron/rl/rl_utils.py
```diff
@@ -1326,751 +838,235 @@ def prepare_data_for_update(
                 (mpu.get_expert_data_parallel_rank() + 1) * data_split_size,
             )
             rollouts = rollouts[data_split_range[0] : data_split_range[1]]
+            rewards = torch.tensor([[r.reward for r in group] for group in rollouts], device='cpu')
```
yobibyte: This place looks a bit weird to me, but maybe I don't understand something. Above we prepared rewards/advantages and put them to tensors. Now we go over groups/rewards again and make rewards/advantages and put them to .cuda() again.

megatron/rl/rl_utils.py
```diff
@@ -1326,751 +838,235 @@ def prepare_data_for_update(
...
-                get_ltor_masks_and_position_ids(
-                    original_trajs,
+            with nvtx_range("sequence_packing", time=True):
+                runtime_state.packing_context = packing_context = pack_all_trajectories(trajs, generation_masks, inference_logprobs, global_advantages)
+    
+                compute_trajs = packing_context.packed_trajs
+                compute_position_ids = packing_context.packed_position_ids
+                # Use batch_size=1 for packed computation to enable proper attention masking
+                # via PackedSeqParams (TE needs cu_seqlens per bin)
+                logprobs_batch_size = 1
```
yobibyte: nit: do we need this var at all?

megatron/rl/rl_utils.py
```diff
@@ -1326,751 +838,235 @@ def prepare_data_for_update(
...
-            # Use packed sequences if packing is enabled for performance benefits
-            if args.rl_use_sequence_packing and 'packed_trajs' in packing_context:
-                compute_trajs = packing_context['packed_trajs']
-                compute_position_ids = packing_context['packed_position_ids']
-                compute_attention_mask = packing_context['packed_attention_mask']
-                use_packed_computation = True
-            else:
-                compute_trajs = original_trajs
+                original_loss_mask[~generation_masks] = 0.0
+                compute_trajs = trajs
```
yobibyte, line 947: nit: do we need this var? We can just put trajs into the TensorDataset.

megatron/rl/rl_utils.py
```diff
@@ -1326,751 +838,235 @@ def prepare_data_for_update(
...
-                b_packed_seq_params = None
-                b_packed_seq_len = 0
-                if use_packed_computation and 'cached_packed_seq_params' in packing_context:
-                    b_packed_seq_params, b_packed_seq_len = packing_context['cached_packed_seq_params'][batch_idx]
-
-                logprobs = get_logprobs(
-                    model, b_trajs, b_posids, b_attn_mask, no_grad=True,
-                    packed_seq_params=b_packed_seq_params,
-                    packed_seq_len=b_packed_seq_len
+                logprobs_batch_size = args.micro_batch_size
```
yobibyte: nit: get rid of the var, pass args.micro_batch_size directly.

megatron/rl/rl_utils.py
```diff
@@ -1326,751 +838,235 @@ def prepare_data_for_update(
...
-                # Get attention mask slice
-                if compute_attention_mask is not None:
-                    start_idx = batch_idx * logprobs_batch_size
-                    end_idx = min(
-                        start_idx + logprobs_batch_size, compute_attention_mask.shape[0]
+            def logprobs_forward_step(data_iterator, model):
+                if args.rl_use_sequence_packing:
+                    # When using sequence packing, the data iterator returns a tuple with a single element, the bin index.
+                    bin_tensor = next(data_iterator)[0]
+                    (b_trajs, _, _, _, b_posids, _, _, _, _, _, b_packed_seq_params) = (
```
yobibyte, line 971: This becomes more and more dangerous. Can you add a todo to return a named tuple instead?

megatron/rl/rl_utils.py
```diff
@@ -1326,751 +838,235 @@ def prepare_data_for_update(
...
+                        load_packed_data_by_index(bin_tensor.item(), packing_context)
```
yobibyte: Nice, it's so much more readable now.

megatron/rl/rl_utils.py
```diff
@@ -1326,751 +838,235 @@ def prepare_data_for_update(
...
                 )
 
-                target_sequences_per_step = args.global_batch_size
-                dp_world_size = max(1, mpu.get_data_parallel_world_size())
+            with torch.no_grad(), nvtx_range("compute_old_logprobs", time=True):
 
-                total_bins_needed = max(
-                    1, math.ceil(target_sequences_per_step / global_avg_seqs_per_bin)
+                dtype = (
+                    torch.bfloat16 if args.bf16 else (torch.float16 if args.fp16 else torch.float32)
```
yobibyte: What if we want fp8?

megatron/rl/rl_utils.py
```diff
@@ -1326,751 +838,235 @@ def prepare_data_for_update(
...
                 }
-
-                if args.micro_batch_size != 1:
-                    print_rank_0(
-                        f"WARNING: micro_batch_size={args.micro_batch_size} but sequence packing expects 1. Using 1."
+                model.load_state_dict(ref_state_dict)
+                ref_logprobs = []
+
+                # Compute reference logprobs
+                data_iterator = iter(data_loader)
```
yobibyte: Should we factor out the old logprobs and ref logprob computation to a function? the whole pp logic looks identical to me.

megatron/rl/rl_utils.py
```diff
@@ -1326,751 +838,235 @@ def prepare_data_for_update(
...
+                    inference_logprobs = align_unpacked_inference_logprobs(
+                        inference_logprobs=inference_logprobs,
+                        old_logprobs_for_data=old_logprobs,
+                        generation_masks=generation_masks,
+                        group_stats=group_stats,
+                    )
+                    # We run the above to fill in the inference/train side mismatch stats.
+                    # We do the above for logging purposes.
+                    # Nullify logprobs if not used in IS correction,
+                    if not args.rl_inference_logprobs_is_correction:
```
yobibyte, line 1166: I think this brings back the bug we had before. If we do not use is correction, we still want to keep the inference logprobs for logging purposes. I fixed it for the no-sequence-packing case, but this will break it again.

I think, we need to fix it. For both, packing or not packing, we need inference logprobs to log the mismatch between train and inference side.

megatron/rl/rl_utils.py
```diff
@@ -2127,56 +1123,16 @@ def setup_grpo_data_iterator(
...
     """
     args = get_args()
+    runtime_state = get_rl_runtime_state()
 
     # We collect new rollouts when we've gone over the collected data 'grpo_iterations' times.
     if (
-        iteration
-        % (args.grpo_iterations * ((args.grpo_samples_per_iteration) // args.global_batch_size))
-        == 0
+        buffered_rollouts is None or
```
yobibyte, line 1248: We need to set buffered_rollouts = train_data_iterator for this to work for GRPO_ITERATIONS>1

megatron/rl/inference/megatron.py
```diff
@@ -168,8 +171,9 @@ def get_dynamic_inference_engine(args: Namespace, model: MegatronModule, inferen
     return DynamicInferenceEngine(
         controller=text_generation_controller,
         context=inference_context,
-        enable_cuda_graph=enable_cuda_graph,
         random_seed=args.seed,
+        track_paused_request_events=args.inference_dynamic_batching_track_paused_request_events,
```
yobibyte, line 175: Is this used anywhere?

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1109 @@
...
+        seq_starts: Dict mapping bin index to list of start positions for each sequence in that bin
+        seq_lengths: List of all original sequence lengths (indexed by global sequence index)
+        seq_to_bin_idx: List mapping each global sequence index to its bin index
+        packing_algo: Algorithm used for distributing bins ('fifo' or 'round-robin')
+    """
+    bin_seq_indices: List[List[int]]
+    seq_starts: Dict[int, List[int]]
+    seq_lengths: List[int]
+    seq_to_bin_idx: List[Optional[int]]
+    packing_algo: str = 'fifo'
```
yobibyte: Let's use typing Literal['fifo', 'round-robin']

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1109 @@
...
+
+
+@dataclass
+class PackingContext:
+    """Context containing all information needed for sequence packing during training.
+    
+    Attributes:
+        bin_size: Maximum size of each bin (in tokens)
```
yobibyte, line 48: nit: let's sort the docstrings args the same way the dataclass fields are defined

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1109 @@
...
+        original_inference_logprobs: Inference logprobs for all sequences before packing (optional)
+        packed_trajs: Packed trajectories tensor [num_bins, bin_size]
+        packed_position_ids: Position IDs for packed sequences [num_bins, bin_size]
+        packed_attention_mask: Attention mask for packed sequences [num_bins, 1, bin_size, bin_size]
+        packed_loss_mask: Loss mask for packed sequences [num_bins, bin_size]
+        bin_advantages: List of advantage tensors for each bin
+        cached_packed_seq_params: Pre-computed PackedSeqParams for each bin
+    """
+    bin_size: int
+    packer: 'SequencePacker'
```
yobibyte, line 62: Do we need a string here or we can use SequencePacker?

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1109 @@
...
+    packed_trajs: torch.Tensor
+    packed_position_ids: torch.Tensor
+    packed_attention_mask: torch.Tensor
+    packed_loss_mask: torch.Tensor
+    original_inference_logprobs: Optional[torch.Tensor] = None
+    bin_advantages: List[torch.Tensor] = field(default_factory=list)
+    cached_packed_seq_params: List[Optional[PackedSeqParams]] = field(default_factory=list)
+
+
+def load_packed_data_by_index(bin_idx: int, packing_context: PackingContext):
```
yobibyte: We only use args in `args.rl_inference_logprobs_is_correction`, let's make this a boolean flag and pass it from the outside.

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1109 @@
...
+    # Get cached PackedSeqParams for proper attention masking in Transformer Engine
+    # These were pre-computed in prepare_data_for_update to avoid repeated tensor allocations
+    packed_seq_params = packing_context.cached_packed_seq_params[bin_idx]
+
+    # Extract packed data for this bin (already on GPU)
+    tokens = packing_context.packed_trajs[idx]
+    position_ids = packing_context.packed_position_ids[idx]
+
+    # Check if we have old_logprobs and ref_logprobs as attributes
+    old_logprobs = getattr(packing_context, 'old_logprobs', None)
```
yobibyte, line 96: Why not just use packing_context.old_logprobs?

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1109 @@
...
+    if old_logprobs is not None:
+        old_logprobs = old_logprobs[idx]
+    
+    ref_logprobs = getattr(packing_context, 'ref_logprobs', None)
```
yobibyte, line 100: same

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1109 @@
...
+    if ref_logprobs is not None:
+        ref_logprobs = ref_logprobs[idx]
+        
+    loss_mask = packing_context.packed_loss_mask[idx, 1:]
```
yobibyte, line 106: Could you add a comment explaining why 1: is here?

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1109 @@
...
+                logging.INFO,
+                f"[Sequence Packing]  - Average sequences per rank: {avg_seqs_per_rank:.1f}",
+            )
+            log_single_rank(
+                logger,
+                logging.INFO,
+                f"[Sequence Packing]  - Max deviation from average: {max_deviation:.0f} sequences ({max_deviation/avg_seqs_per_rank*100:.1f}%)",
+            )
+
+def get_actual_sequence_lengths(sequences: torch.Tensor, pad_token: int) -> List[int]:
```
yobibyte, line 286: Let's add a new file to the unit test where we have test_rl_utils.py (test_sequence_packing_utils.py), and unit test this function.

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1109 @@
...
+
+    return (
+        empty_trajs,
+        empty_position_ids,
+        empty_loss_mask,
+        empty_attention_mask,
+        empty_packing_info_entries,
+    )
+
+def create_packed_seq_params(packing_context: PackingContext):
```
yobibyte, line 429: Do we actually need this function? 

Can be a simple list comprehension at a call place:

```
cached_packed_seq_params = [
	create_packed_seq_params_for_bin(
            packing_info=packing_context.packing_info,
            bin_idx=bin_idx,
            bin_size=packing_context.bin_size,
            device=packing_context.packed_trajs.device,
        ) for bin_idx in range(len(packing_context.packed_trajs))
]
```

megatron/rl/sequence_packing_utils.py
jalbericiola: OK, but now the ugliness will be in the two calls in rl_utils and train_rl

megatron/rl/sequence_packing_utils.py
jalbericiola: neat

examples/rl/model_configs/common.sh
jalbericiola: it was not compatible with the mrl_internal branch we were using. Will change it

tests/unit_tests/test_rl_utils.py
jalbericiola: fixed

megatron/rl/sequence_packing_utils.py
jalbericiola: done

megatron/rl/sequence_packing_utils.py
ArEsKay3: FWIW, I actually like the old way because it moves the introspection of the underlying packing_context to the seqeunce_packing_utils.py file.

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1104 @@
...
+        bin_size: Size of each bin
+
+    Returns:
+        Packed inference logprobs tensor of shape [num_bins, bin_size - 1]
+    """
+    num_bins = len(packing_info.bin_seq_indices)
+
+    # Create packed inference logprobs tensor (logprobs are 1 token shorter than sequences)
+    packed_inference_logprobs = torch.zeros(
+        (num_bins, bin_size - 1), dtype=torch.float32, device='cpu'
```
yobibyte: Is float32 the correct dtype here?

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1104 @@
...
+        sequences_tensor = torch.stack(sequences)
+
+        seq_lengths = get_actual_sequence_lengths(sequences_tensor, self.pad_token)
+
+        # Trim sequences to actual lengths
+        sequences = [sequences_tensor[i, :length] for i, length in enumerate(seq_lengths)]
+
+        sorted_indices = sorted(range(len(sequences)), key=lambda i: seq_lengths[i], reverse=True)
+
+        args = get_args()
```
yobibyte: the only thing we use from args is seq_length, let's have the seq_length field in SequencePacker and send it to the constructor from the outside. get_args() is a disease that we should stop from spreading.

tdene: agreed

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1147 @@
...
+class SequencePacker:
+    """Packs multiple sequences into bins to minimize padding and improve GPU utilization."""
+
+    def __init__(self, bin_size: int, pad_token: int, max_sequences_per_bin: int = 16):
+        self.bin_size = bin_size
+        self.pad_token = pad_token
+        self.max_sequences_per_bin = max_sequences_per_bin
+
+    def pack_sequences(
+        self, trajs: torch.Tensor, generation_masks: torch.Tensor = None
```
yobibyte: generation masks type should be Optional[torch.Tensor]

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1147 @@
...
+    packed_trajs: torch.Tensor,
+    packed_position_ids: torch.Tensor,
+    packed_attention_mask: torch.Tensor,
+    packed_loss_mask: torch.Tensor,
+    packing_info: PackingInfo,
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, PackingInfo]:
+    """Distribute packed bins across the data parallel ranks."""
+    rank = mpu.get_expert_data_parallel_rank()
+    world_size = mpu.get_expert_data_parallel_world_size()
+    args = get_args()
```
yobibyte: Let's dump get_args() and use sequence_packing_algo as an arg to this function, please.

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1147 @@
...
+
+
+def get_microbatch_dataloader(packing_context: PackingContext) -> Tuple[DataLoader, int]:
+    args = get_args()
+    num_bins_this_rank = len(packing_context.packed_trajs)
+    dp_world_size = mpu.get_data_parallel_world_size()
+
+    # Ratio of collected sequences to the global batch size
+    pct_of_sequences_per_batch = len(packing_context.packing_info.seq_lengths) / args.global_batch_size
+    pct_of_bins_per_batch = pct_of_sequences_per_batch
```
yobibyte: this line is a bit weird, let's just use pct_of_sequences_per_batch in the math.ceil expr

megatron/rl/rl_utils.py
ArEsKay3: I think that happens outside this right?
[Here](https://github.com/jalbericiola/Megatron-LM/blob/jalbericiola/pp_spack_rewrite/megatron/training/training.py#L2419)

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1147 @@
...
+
+    # Ceiling division means we will reuse some bins
+    # If we did floor we would leave some behind
+    local_bins_per_step = math.ceil(pct_of_bins_per_batch * num_bins_this_rank)
+    effective_global_batch_size = local_bins_per_step * dp_world_size
+
+    # Store packing plan in runtime state for the training loop to use
+    optimizer_steps = -(-num_bins_this_rank // local_bins_per_step)
```
yobibyte, line 1094: is this some dark magic with 2 minus signs or a typo?

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1147 @@
...
+
+    from megatron.core.num_microbatches_calculator import (
```
yobibyte: let's please move the imports on top of the file

megatron/rl/rl_utils.py
ArEsKay3: First we calculate them on a global level and then we split and recalculate on a local level.
This is strange because sequence packing and reporting needs it global but non-packing wants it local.
We can probably rewrite variable names to be a bit cleaner (i.e. use global_advantages, global_rewards exclusively for the global version).

megatron/rl/sequence_packing_utils.py
ArEsKay3: This is ceiling division :), used to have a comment to that effect must have accidentally dropped it.

megatron/rl/sequence_packing_utils.py
ArEsKay3: Wasn't me this time!

megatron/rl/rl_utils.py
yobibyte: I see! I'm fine with variable names, but let's add your response from above as a comment for the future generation.

megatron/rl/rl_utils.py
yobibyte: I was a bit lost and actually thought we are at the place you linked. You are right, I'm closing this.

yobibyte (approved):
Hey folks, that's a monumental work, thanks for doing this!

I approve, but let's address the unresolved conversation, please.

megatron/training/initialize.py
```diff
@@ -71,12 +71,13 @@ def initialize_megatron(
...
 
     if args.use_checkpoint_args or args_defaults.get("use_checkpoint_args", False):
-        assert args.load is not None, "--use-checkpoint-args requires --load argument"
+        assert args.load is not None or args.pretrained_checkpoint is not None, "--use-checkpoint-args requires --load or --pretrained-checkpoint argument"
         assert args.non_persistent_ckpt_type != "local", (
             "--use-checkpoint-args is not supported with --non_persistent_ckpt_type=local. "
             "Two-stage checkpoint loading is not implemented, and all arguments must be defined "
             "before initializing LocalCheckpointManager."
         )
+        load_args_from_checkpoint(args, load_arg='pretrained_checkpoint')
```
tdene, line 80: Why is this line duplicated now?

tdene: Nevermind, I understand.

megatron/rl/sequence_packing_utils.py
yobibyte: We will eventually move it to the top level =)

megatron/rl/sequence_packing_utils.py
```diff
@@ -0,0 +1,1143 @@
...
+
+    cu_seqlens = torch.tensor(cu_seqlens_list, dtype=torch.int32, device=device)
+
+    # Pad cu_seqlens to bin_size by repeating the last value (creates zero-length ghost sequences)
+    # This ensures a fixed tensor size for CUDA graph compatibility
+    if len(cu_seqlens) < bin_size:
+        padding_size = bin_size - len(cu_seqlens)
+        last_value = cu_seqlens[-1]
+        padding = torch.full((padding_size,), last_value, dtype=torch.int32, device=device)
+        cu_seqlens = torch.cat([cu_seqlens, padding])
```
mathemakitten: This does indeed run but seems quite slow, I think because of the `concat` requiring a new memory allocation. Can we try to sub with a tensor instantiated with the padding value already like below? In a vacuum it's a bit faster though probably still suboptimal.

```
    if len(cu_seqlens) < bin_size:
        out = cu_seqlens.new_full((bin_size,), cu_seqlens[-1])
        out[:len(cu_seqlens)] = cu_seqlens
        cu_seqlens = out
```

jalbericiola:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/2632/commits/4f1711270fc41d2f61f9df76f6e473c6243e76c9

PR #2685, opened 2025-12-17
Adding stop word support

shanmugamr1992:
# What does this PR do ?

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


kvareddy:
@santhnm2 @tdene can you please take a look a this MR.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -911,8 +935,100 @@ def post_process_requests(
                     else:
                         request.generated_top_n_logprobs.append(logit_dict)
 
+        # Clear the stop word being finished set after processing
+        self.stop_word_being_finished_ids.clear()
+
         return active_request_ids, finished_request_records
 
+    def _check_stop_words_for_request(self, request_id: int, new_token: int) -> bool:
```
tdene: This method is no longer used anywhere.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -74,9 +74,23 @@ def __init__(
         self.sampling_rng = torch.Generator(device=torch.cuda.current_device())
         self.sampling_rng.manual_seed(model_config.inference_sampling_seed)
 
+        # Callback to get request IDs that should be marked as finished due to stop words
+        self._get_stop_word_finished_ids_callback = None
```
tdene: nit: I think this should go inside `self._init_dynamic_sampling_tensors()` because it's only used in dynamic inference.

But it's not very important where it goes.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -716,6 +723,11 @@ def _add_request(
             request.status = Status.FAILED
             request.add_event_error_nontransient(TokenOverflowError(request_id))
 
+        # Tokenize stop words if provided
+        if request.sampling_params.stop_words:
+            stop_word_ids = [self.controller.tokenize_prompt(stop_word, add_BOS=False) for stop_word in request.sampling_params.stop_words]
+            request.sampling_params.stop_word_ids = stop_word_ids
```
tdene: I would prefer if `stop_word_ids` was kept directly inside the `request` object, instead of added to the `sampling_params` object.

megatron/core/inference/sampling_params.py
```diff
@@ -30,6 +30,8 @@ class SamplingParams:
     top_n_logprobs: int = 0
     return_prompt_top_n_logprobs: bool = False  # Deprecated field for backwards compatibility
     add_BOS: bool = False
+    stop_words: Optional[List[str]] = None  # List of strings that will stop generation when produced
+    stop_word_ids: Optional[List[List[int]]] = None  # Tokenized stop words (populated internally)
```
tdene: I would prefer if `stop_word_ids` was kept directly inside the `request` object, instead of added to the `sampling_params` object (see above).

megatron/core/inference/engines/dynamic_engine.py
shanmugamr1992: Good catch removed

megatron/core/inference/text_generation_controllers/text_generation_controller.py
shanmugamr1992: Done

wdykas:
More just a general question.
Do stop words have to be a single token to work? Like what if my stop word gets broken up into multiple tokens?

tdene:
> More just a general question. Do stop words have to be a single token to work? Like what if my stop word gets broken up into multiple tokens?

@wdykas In this PR, the stop word that is passed in as an argument is pre-tokenized, and the stop word only comes into effect if all of those tokens are seen in sequence. So this works with multi-token stop words.

But that's only as long as the stop word that is passed in as an argument is tokenized the same way as the stop word in the sequence. For example, passing in the stop word `stop word` will not trigger on the sequence `i ate a chicken and a stop word`, while the stop word ` stop word` would have triggered on that sequence.

I think this question of ambiguous tokenization is a bigger one that can be answered later.

tdene (approved):

wdykas:
> > More just a general question. Do stop words have to be a single token to work? Like what if my stop word gets broken up into multiple tokens?
> 
> @wdykas In this PR, the stop word that is passed in as an argument is pre-tokenized, and the stop word only comes into effect if all of those tokens are seen in sequence. So this works with multi-token stop words.
> 
> But that's only as long as the stop word that is passed in as an argument is tokenized the same way as the stop word in the sequence. For example, passing in the stop word `stop word` will not trigger on the sequence `i ate a chicken and a stop word`, while the stop word ` stop word` would have triggered on that sequence.
> 
> I think this question of ambiguous tokenization is a bigger one that can be answered later.

cool. sounds good, wasn't meant to be a critique of the MR just more curiosity.

shanmugamr1992:
/ok to test [ff8b62b](https://github.com/NVIDIA/Megatron-LM/pull/2685/commits/ff8b62bd7e4a8b48568419e85814131847754925)

shanmugamr1992:
/ok to test [5926445](https://github.com/NVIDIA/Megatron-LM/pull/2685/commits/592644541e6fd67be74fcdc7b9d98326230219e7)

PR #2689, opened 2025-12-17
[M-FSDP] Fix double buffering not working with activation recompute

shjwudp:
# What does this PR do ?

During activation recompute, the forward pass is interleaved within the backward pass. Previously, to prevent forward unsharded parameters from being reused by the backward function, parameter release was skipped during the re-forward phase ([megatron_fsdp.py#L781](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L781)). However, this caused the memory allocated by the double-buffer allocator to not be properly reclaimed in time.

This PR fixes the problem by adopting a lazy parameter release on re-forward function.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -495,10 +495,10 @@ def _register_fsdp_hooks(self, root_module):
         """
         fsdp_unit_modules = self.fsdp_unit_modules
 
-        def release_module_parameters(module, *unused):
+        def release_module_parameters(module, lazy=False, *unused):
```
youngeunkwon0405: Could you please add a comment on when `lazy` is used here?

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -3344,11 +3344,16 @@ def wait_bucket_ready(self, bucket_id, empty_ok=False):
         mark_bucket_ready_to_use()
 
     @torch.no_grad()
-    def release_bucket(self, bucket_id: int):
+    def release_bucket(self, bucket_id: int, lazy: bool = False):
```
youngeunkwon0405: Can you also add a comment about the `lazy` argument? When is it used and how does it work?

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
cspades: ➕

shjwudp: The issue here is that what we need is a lazy release, not no release.

Thanks for the suggestion — I’ll add some comments.

youngeunkwon0405: I think the comment is not that informative. 

Maybe

`lazy: If True, will release parameter buffer at [Specific release timing (TBH, I currently don't know)], otherwise it is released [Specific release timing]. This argument is used when activation recomputation is used.`

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
youngeunkwon0405: Here as well. I currently don't have a clear idea of when and how the bucket is released. Could you please provide more specific information?

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
shjwudp: We can always use lazy release; the all-gather pipeline guarantees memory release at the correct time. To be conservative, we won't modify the default behavior. Therefore, `default lazy` is `False`.

I think the question is why, in some cases, the post-forward operation doesn't release model parameters. Do you think the explanation in the description is sufficient?

youngeunkwon0405: Having a default to False is totally fine. The point is that the comment is not so informative. 

`lazy: If True, release the parameters lazily.` This provides zero additional information. Could you please provide more informative comments about this argument? 

In the above comment, I was trying to provide an example form of the comment. I guess my intention wasn't delivered well to you.

shjwudp: > Could you please add a comment on when lazy is used here?

Going back to the beginning of the question, I think I've explained it clearly: we can use lazy at any time.

There is a question about why there is a mandatory lazy release in the post-forward function (when activation recompute used). The reason is that parameters can be reused during the subsequent backward function, so it is preferable to delay releasing them. I've updated the comments for post-backward; does that answer your question?

youngeunkwon0405: I see. Thanks for the explanation! I think my question was also not that clear, sorry about that.
I had two questions.

1. When should the `lazy` be used?
--> I think you've answered this clearly.
2. If `lazy` is used, **when** is the buffer released?
--> I can understand that it might be delayed compared to `lazy=False`, but how much will it be delayed? In this case, specifically, when will it be released? This part is not easy to infer from the code, since it is just marked as `self.bucket_can_be_released[bucket_id] = True` and handled from a different part of the code. If this part is hard to explain, I think you can leave it as is.

youngeunkwon0405 (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
shjwudp: > when is the buffer released?

Buffer release occurs before all-gather pipeline requests a new buffer, and is implemented by calling `recycle_unused_buckets`.

youngeunkwon0405: Thank you for the explanation. It will be appreciated if we can have that information somewhere near this comment.

jaredcasper (approved):

cspades:
/ok to test [c30a332](https://github.com/NVIDIA/Megatron-LM/pull/2689/commits/c30a33200153c9041d6fa05d4b01faf9462f53dc)

cspades:
So TL;DR, we're basically just trying to free some buckets in Buffer 1 at Layer N earlier (than the [post-backward](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L579) -> [`release_bucket`](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L513)) so that we can avoid synchronization delays when we allocate from Buffer 1 again at Layer N +/- 2?

PR #2690, opened 2025-12-17
[Dev] Paged Stashing

nanz-nv:
Main PR: #4247 

Main contributors (Equal Contribution, sorted alphabetically): Nan Zheng (@nanz-nv), Vasudevan Rengasamy (@vasunvidia)
Other contributors (sorted alphabetically): Dennis Liu(@Victarry), Hongbin Liu(@lhb8125), Qi Zhang(@QiZhangNV), Robin Zhang(@buptzyb), Tong Liu(@Autumn1998), Zijie Yan(@yanring)

# Background
In token-dropless MoE training, the number of tokens received by each expert might vary, resulting in dynamic shaped tensors. Dynamic shaped tensors are naturally supported by PyTorch, thanks to its eager mode nature. This is done by creating a tensor lazily when the shape of the tensor is known at run-time. Albeit working well in eager mode, dynamic shaped tensor poses challenges for CUDA graphs because the the size of a tensor cannot be dynamically adjusted at runtime without the intervene of the host. In order to remove the sync and enable CUDA graph, one solution is to oversize the buffer in the expert part. This however causes significantly higher memory consumption compared to the eager-mode baseline through the form of memory fragmentation.

image

# Idea overview
To address this problem, paged stashing decouples the need of oversized buffers for compute and the need of a properly sized buffer for storing activations for the backward pass. Paged stashing achieves this through adding one level of indirection: stashing and restoring. The stash operation copies the activation from the oversized static buffer to a pre-allocated stashing buffer after the forward for that module is done, and the restore operation does the reverse operation during the backward pass.

<p align="center">
image
</p>

The key of saving memory lies in the fact that the stash operation packs the variable-size activation into a contiguous stashing buffer to reduce memory fragmentation. For simple scheduling where the activation allocation and deallocation follows a first-in-last-out pattern, stash and restore can be done easily in a bump-allocation manner. To accommodate complicated scheduling schedules, e.g. pipeline parallel, paging can be used, hence the name paged stashing.

# page management
To accomodate complex scheduling such as that needed in pipeline parallelism, activations are partitioned into pages and a light-weight memory management kernel is in charge of allocate and deallocate pages for stashing. Pages are managed by lightweight GPU memory management kernels that can be fused with the stash/restore GPU kernels. It maintains a freelist which is implemented as a circular buffer. Each freelist keeps track of one type of pages.

# CPU offloading
Paged stashing naturally supports offloading. When the stashing buffer is a pinned CPU tensor, the activation is offloaded to the host memory during forward and is reloaded to the GPU during backward.
Furthermore, one can easily extend the paging management system to accommodate partial offloading or on-demand offloading. This feature is currently WIP.

# scheduling
Overlapping stashing and restore operations with compute can be implemented by inserting two autograd functions before and after the expert compute layer: pre-scheduler and post-scheduler that schedules stash and restore operations. The roles of these autograd functions are enumerated below:
- Pre-scheduler forward: Wait for previous stash op. to complete, free the max-capacity sized temporary activations for the completed stash op. The wait is performed here instead of Post-scheduler forward to reduce the peak memory usage since the following expert compute layer will allocate another set of max-capacity sized temporary activations.
- Post-scheduler forward: Since this is after experts compute, stashing operations for the current layer activations are scheduled here. If the next layer in the execution is a backward pass layer, schedule restore operations for the next layer.
Additionally, in case of pipeline parallelism, this can be used to record the pipeline schedule during the first iteration.
- Post-scheduler backward: Wait for previous stash op. to complete, free the max-capacity sized temporary activations for the completed stash op. The wait is performed here instead of Pre-scheduler backward to reduce the peak memory usage since the following expert compute BPROP layer will allocate another set of max-capacity sized temporary activations.
Wait for restore operation for the current layer to complete. Additionally, in case of pipeline parallelism, this can be used to record the pipeline schedule during the first iteration.
- Pre-scheduler backward: If the next layer in the execution is a backward pass layer, schedule restore operations for the next layer.


megatron/core/transformer/moe/experts.py
```diff
@@ -860,15 +872,24 @@ def forward(
...
         if self.config.fp8 or self.config.fp4:
-            actual_tokens_per_expert = tokens_per_expert
-            permuted_local_hidden_states, tokens_per_expert = self.quantization_padding(
-                permuted_local_hidden_states, tokens_per_expert
-            )
-            permuted_probs, _ = self.quantization_padding(
-                permuted_probs.unsqueeze(-1), actual_tokens_per_expert
-            )
+            if self.config.fp8 and self.config.moe_use_device_initiated_grouped_gemm:
+                assert self.config.moe_router_padding_for_fp8, "Should set --moe-router-padding-for-fp8 to use router padding for fp8 when enabled device-initiated grouped gemm."
```
QiZhangNV:
```suggestion
                assert self.config.moe_router_padding_for_fp8, "Should set --moe-router-padding-for-fp8 to use router padding for fp8 when enabled --moe-use-device-initiated-grouped-gemm."
```

megatron/core/transformer/moe/experts.py
```diff
@@ -860,15 +872,24 @@ def forward(
...
+            else:
+                assert not self.config.moe_use_device_initiated_grouped_gemm, "--moe-use-device-initiated-grouped-gemm not supported for FP4"
+                permuted_local_hidden_states, tokens_per_expert = self.quantization_padding(
+                    permuted_local_hidden_states, tokens_per_expert
+                )
+                permuted_probs, _ = self.quantization_padding(
+                    permuted_probs.unsqueeze(-1), actual_tokens_per_expert
+                )
         else:
             permuted_probs = permuted_probs.unsqueeze(-1)
```
QiZhangNV: Could we also add an assert here since device grouped gemm does not support it

megatron/core/transformer/moe/experts.py
```diff
@@ -915,6 +942,10 @@ def bias_act_func(intermediate_parallel, bias_parallel, permuted_probs):
                         bias_parallel,
                         permuted_probs,
                         self.config.activation_func_fp8_input_store,
+                        tokens_per_expert.sum()
+                        if (isinstance(tokens_per_expert, torch.Tensor) and tokens_per_expert.is_cuda)
```
QiZhangNV:
```suggestion
                        if self.config.moe_use_device_initiated_grouped_gemm
```

megatron/core/transformer/transformer_config.py
```diff
@@ -622,6 +622,10 @@ class TransformerConfig(ModelParallelConfig):
     GEMM feature introduced since CUTLASS 2.8 (https://github.com/fanshiqing/grouped_gemm).
     """
 
+    moe_use_device_initiated_grouped_gemm: bool = False
+    """Use the cutlass grouped gemm kernel, which allows for the token_per_expert tensor on GPU. This can prevent the GPU-CPU synchronization during the grouped gemm."""
```
QiZhangNV: cutlass -> device initiated, since there may be other backends
```suggestion
    """Use the device initiated grouped gemm kernel, which allows for the token_per_expert tensor on GPU. This can prevent the GPU-CPU synchronization during the grouped gemm."""
```

megatron/training/arguments.py
```diff
@@ -3204,6 +3212,8 @@ def _add_moe_args(parser):
                        'Only effective when moe-shared-expert-intermediate-size is set.')
     group.add_argument('--moe-grouped-gemm', action='store_true',
                        help='When there are multiple experts per rank, launch multiple local GEMM kernels in multiple streams to improve the utilization and performance with GroupedLinear in TransformerEngine.')
+    group.add_argument('--moe-use-device-initiated-grouped-gemm', action='store_true',
+                       help='Use the cutlass grouped gemm kernel, which allows for the token_per_expert tensor on GPU. This can prevent the GPU-CPU synchronization during the grouped gemm.')
```
QiZhangNV:
```suggestion
                       help='Use the device initiated grouped gemm kernel, which allows for the token_per_expert tensor on GPU. This can prevent the GPU-CPU synchronization during the grouped gemm.')
```

megatron/training/utils.py
```diff
@@ -561,7 +561,8 @@ def _broadcast(item):
         def _broadcast_cu_seqlens(cu_seqlens):
             dev = torch.cuda.current_device()
             n = 0 if cu_seqlens is None else int(cu_seqlens.numel())
-            n_tensor = torch.tensor(n, dtype=torch.int64, device=dev)
+            n_tensor = torch.ones(1, dtype=torch.int64, device=dev) * n
+            # n_tensor = torch.tensor(n, dtype=torch.int64, device=dev)
```
QiZhangNV: 👀

megatron/core/transformer/transformer_config.py
```diff
@@ -1245,6 +1275,30 @@ def __post_init__(self):
...
+            assert not invalid_modules, (
+                f'Invalid choices for stash_modules: {invalid_modules}. '
+                f'Allowed modules are: {allowed_modules}'
+            )
+            assert (
+                self.moe_expert_rank_capacity_factor is not None
+            ), "moe_expert_rank_capacity_factor must be set when moe_paged_stash is enabled."
+
+        # Check that no module is both stashed and offloaded
+        if self.stash_modules and self.offload_modules:
```
vasunvidia: When will this condition be true? Shouldn't offload_modules be empty when paged stashing is enabled?

megatron/core/transformer/moe/experts.py
```diff
@@ -966,20 +1015,48 @@ def glu(x):
...
+                    name="moe_act",
+                    max_num_tokens=fc1_output.shape[0],
+                    num_tokens_tensor=tokens_per_expert.sum(),
+                )
+            else:
+                offload_context = get_fine_grained_offloading_context(self.offload_moe_act)
+            with offload_context:
                 bias_act_output = bias_act_func(fc1_output, bias_parallel, permuted_probs)
+        if self.offload_moe_act:
+            (bias_act_output,) = fine_grained_offloading_group_commit(
```
vasunvidia: Why fine_grained_offloading_group_commit has been moved ?

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1401,9 +1418,9 @@ def _initialize_metadata(self, routing_map: torch.Tensor, probs: torch.Tensor) -
             .expand(-1, -1, self.tp_size, -1)
             .reshape(num_local_tokens, world_size, self.num_local_experts)
         ).contiguous()
+
         return routing_map, probs
 
-    @jit_fuser
```
vasunvidia, line 1434: Why is this change required?

megatron/training/utils.py
```diff
@@ -561,7 +561,8 @@ def _broadcast(item):
         def _broadcast_cu_seqlens(cu_seqlens):
             dev = torch.cuda.current_device()
             n = 0 if cu_seqlens is None else int(cu_seqlens.numel())
-            n_tensor = torch.tensor(n, dtype=torch.int64, device=dev)
+            n_tensor = torch.ones(1, dtype=torch.int64, device=dev) * n
```
vasunvidia: Will this cause any issue during CG replay? Is it safer to use n_tensor.fill_(n) instead?

megatron/core/transformer/transformer_config.py
nanz-nv: My expectation is that we can have both fine-grained offloading working on attention part and paged stashing working on the expert part.

megatron/core/transformer/moe/experts.py
lhb8125: The group_commit should be placed after ```self.activation_checkpoint.discard_output_and_register_recompute()```.

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -0,0 +1,951 @@
...
+                    stash_buffer = self.stash_buffers[paged_tensor.dtype][paged_tensor.hidden_size]
+                    paged_tensor.reload_from_stash(stash_buffer)
+            else:
+                pass
+            assert len(self.paged_tensors_to_reload[pp_schedule_layer]) == 0, (
+                f"paged_tensors_to_reload {pp_schedule_layer} is not empty "
+                f"{self.paged_tensors_to_reload[pp_schedule_layer]}"
+            )
+
+    def allocate_stash_buffers(self, stash_buffer_size_factor=1.10):
```
jianyuh: Curious how stash_buffer_size_factor is going to be determined? Is 1.10 be reasonable enough?

nanz-nv: Whether 1.1 is enough depends on the distribution of the token distribution in each layer.
One can use the remaining GPU memory after fitting the model/activation as the stashing buffer size. Or this can be done through some iterative trial, similar to deciding the best sharding/microbatch size that can fit on GPU with load imbalance.

megatron/training/arguments.py
```diff
@@ -1409,6 +1409,7 @@ def validate_args(args, defaults={}):
         if is_te_min_version("2.10.0"):
             assert os.getenv("NVTE_CPU_OFFLOAD_V1", "0") == "1", \
                 "For fine-grained activation offloading with TE >= 2.10.0, NVTE_CPU_OFFLOAD_V1 should be set to 1 to avoid offloading weights."
+        assert not args.moe_paged_stash, "Fine-grained activation offloading and paged stash cannot be enabled at the same time"
```
hxbai: Why is this assertion added?

megatron/core/transformer/transformer_config.py
```diff
@@ -770,10 +770,13 @@ class TransformerConfig(ModelParallelConfig):
     block interleaved format. Instead of interpreting the input tensor
     as a concatenation of gates and linear units, it will be
     interpreted as alternating blocks of gates and linear units.
-
     This data format is experimental and primarily intended to enable
     advanced fused kernels."""
 
+    moe_expert_rank_capacity_factor: Optional[float] = None
+    """moe_expert_rank_capacity_factor (float): The capacity factor for each expert, None means no token
```
hxbai: for each rank?

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -0,0 +1,1241 @@
...
+            num_tokens_tensor,
+            self.page_record,
+            self.spilled_to_host,
+            paged_stash_buffer.overflow,
+            paged_stash_buffer.free_list_cuda,
+            paged_stash_buffer.free_list_host,
+            paged_stash_buffer.free_list_tail,
+            paged_stash_buffer.free_list_capacity,
+            new_free_list_tail,
+            PAGE_SIZE=self.page_size,
```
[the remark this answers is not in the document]
nanz-nv: This was intentional because by design when overflow happens all the triton kernels will not try to load/store anything other than the overflow flag. However, this has changed in the process of attempting to optimize the kernel. So this is indeed a problem, and we have fixed it. Thanks for catching that!

megatron/core/transformer/moe/experts.py
```diff
@@ -1001,9 +1056,28 @@ def forward(
...
+                )
+                offload_context = get_paged_stash_context(
+                    name="expert_fc1",
+                    max_num_tokens=max_num_tokens,
+                    num_tokens_tensor=tokens_per_expert.sum(),
+                    avg_num_tokens=avg_num_tokens,
+                )
+            else:
+                offload_context = nullcontext()
+            with offload_context:
```
buptzyb: Can we use a different name? This is confusing, especially together with the `off_interface`... I believe they are two unrelated features, right?

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -0,0 +1,1299 @@
...
+            HIDDEN_SIZE=self.hidden_size,
+            BLOCK_SIZE=BLOCK_SIZE,
+        )
+
+        paged_stash_buffer.free_list_tail.copy_(new_free_list_tail)
+
+
+class PP_PreScheduleFunction(torch.autograd.Function):
```
buptzyb: Is this good with PEP8? (PascalCase + "_")

How about `PipelinePreScheduleFunction` or `PrePipelineScheduleFunction`?

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -0,0 +1,1299 @@
...
+
+def check_paged_stash_host_spill():
+    """True if any activation was stashed to pinned host (successful spill, not overflow path)."""
+    stash_manager = PagedStashManager.get_instance()
+    if not stash_manager.enabled or stash_manager.host_spill is None:
+        return torch.zeros(1, dtype=torch.bool, device='cuda')
+    return stash_manager.host_spill.ne(0)
+
+
+class PagedStashRunner:
```
buptzyb, line 1166: I don't see `PagedStashRunner` get called anywhere. If we want to merge the current version (without overflow fallback), we'd better remove this part.

megatron/core/full_cuda_graph.py
```diff
@@ -98,10 +99,18 @@ class FullCudaGraphWrapper:
     cuda_graph = {'training': None, 'validation': None}
     result = {'training': None, 'validation': None}
 
-    def __init__(self, forward_backward_func, cuda_graph_warmup_steps=1):
+    def __init__(
+        self,
+        forward_backward_func,
+        cuda_graph_warmup_steps=1,
+        moe_paged_stash=False,
```
buptzyb: Is this argument useful?

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1070,12 +1084,16 @@ def dispatch(
                 pad_multiple=self.pad_multiple,
             )
         )
+        if self.moe_expert_rank_capacity_factor is not None:
+            over_budget = self.handle[8] != 0  # this is overflow_flag
+            self.over_budget |= over_budget
```
buptzyb, line 1088: So, it's forever over budget once it goes over budget in one step. Never got a chance to recover to normal?

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -1295,6 +1295,22 @@ def get_align_size_for_quantization(config: TransformerConfig) -> int:
...
+    Returns True when padding is already applied by the router or the
+    HybridEP dispatcher.
+    """
+    if config.moe_router_padding_for_quantization:
+        return True
+    if (
+        config.moe_token_dispatcher_type == "flex"
+        and config.moe_flex_dispatcher_backend == "hybridep"
+    ):
+        return True
```
buptzyb, line 1358: Always skip padding for hybridep? Why put this in the paged stashing PR? It affects the non-paged-stashing logic as well.

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -590,6 +591,8 @@ def forward_backward_no_pipelining(
     if config.timers is not None:
         config.timers('forward-backward', log_level=1).start(barrier=config.barrier_with_L1_time)
 
+    paged_stash_reset(enabled=config.moe_paged_stash and not forward_only, config=config)
```
buptzyb:
```
if moe_paged_stash:
    paged_stash_reset(...)
```
To ensure the new things don't mess up the old codepath.

megatron/core/transformer/moe/moe_utils.py
nanz-nv: This is a minor fix from @zhongbozhu. It is to correct the behavior introduced in this PR: https://github.com/NVIDIA/Megatron-LM/pull/3890
Yes currently hybridEP unconditionally pad the input if it is FP8/FP4.

megatron/core/transformer/moe/paged_stash.py
nanz-nv: It is used in MBridge. @vasunvidia could you add the usage of `PagedStashRunner` to Megatron-LM as well?

megatron/core/transformer/moe/token_dispatcher.py
nanz-nv: It is reset here: https://github.com/NVIDIA/Megatron-LM/blob/21ff4bc23d357f6aaa489330f93e299103476394/megatron/core/transformer/moe/paged_stash.py#L1197-L1202
When a fallback rerun is triggered, the `over_budget` flag is reset.

megatron/core/transformer/moe/moe_utils.py
buptzyb: So I suppose there will be a dedicated PR from zhongbo handling this, instead of mixing in this PR?

megatron/core/full_cuda_graph.py
nanz-nv: removed

megatron/core/transformer/moe/experts.py
```diff
@@ -534,6 +541,36 @@ def backward_dw(self):
         """
         pass
 
+def _pad_unpad(inp, pad):
+    if pad:
+        if inp.ndim == 2:
+            result = torch.nn.functional.pad(inp, (0, 0, 0, 256))
+        elif inp.ndim == 1:
+            result = torch.nn.functional.pad(inp, (0, 256))
```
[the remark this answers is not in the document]
nanz-nv: removed

megatron/training/arguments.py
nanz-nv: It was there due to historical reasons. Just removed it. Thanks for catching that.

megatron/core/pipeline_parallel/schedules.py
nanz-nv: @vasunvidia

megatron/core/transformer/moe/paged_stash.py
nanz-nv: @vasunvidia

megatron/core/transformer/moe/moe_utils.py
zhongbozhu: I added a stand-alone PR for it: https://github.com/NVIDIA/Megatron-LM/pull/4071, can you take a look?

buptzyb: LGTM!

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -0,0 +1,1299 @@
...
+    assert num_tokens_tensor is not None and isinstance(num_tokens_tensor, torch.Tensor)
+    stash_manager.num_tokens_tensor = num_tokens_tensor
+    stash_manager.set_current_layer_name(name) if name is not None else None
+    pack_unpack_context = PagedStashContext(stash_manager)
+    return pack_unpack_context
+
+
+def paged_stash_group_commit(tensor, name=None):
+    """Mark the end of a layer group and prepare for stash/reload."""
+    rank = torch.distributed.get_rank()
```
Victarry: **[IMPORTANT Unused Variable]** `rank = torch.distributed.get_rank()` is computed but never used.

This is called on every expert layer forward pass. While the overhead is small, it's unnecessary dead code.

**Suggestion:** Remove this line.

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -0,0 +1,1299 @@
...
+            current_stream = torch.cuda.current_stream()
+            self.unpack_stream.wait_stream(current_stream)
+
+        with torch.cuda.stream(self.unpack_stream):
+            if self.status == 'captured':
+                self._unpack_stream_status = 'reloading'
+                count = 0
+                for item in self.paged_tensors_to_reload:
+                    if len(self.paged_tensors_to_reload[item]) > 0:
+                        count += 1
```
Victarry: **[IMPORTANT Unused Variable]** `count` is computed by iterating over all reload queues but is never read or used.

**Suggestion:** Remove these 4 lines (720-723).

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -0,0 +1,1299 @@
...
+        """Get the unpack stream."""
+        return self._unpack_stream
+
+    def set_current_layer_name(self, name):
+        """Set the current layer name."""
+        self._current_layer_name = name
+
+    def get_schedule_layer(self, vp_stage, layer_no, microbatch_no):
+        """Get the schedule layer."""
+        return vp_stage * 1000000 + layer_no * 1000 + microbatch_no
```
Victarry, line 665: **Suggestion:** Add an assertion to validate the ranges:
```python
assert layer_no < 1000 and microbatch_no < 1000, "Schedule encoding overflow"
```

megatron/core/transformer/transformer_block.py
```diff
@@ -892,7 +892,7 @@ def forward(
                         mhc_manager.is_last_layer_in_recompute_block = (
                             mhc_is_last_in_recompute_block[l_no]
                         )
-
+
```
Victarry: **[SUGGESTION]** This line changed from an empty line to a line with trailing whitespace — unintentional diff noise. Consider reverting to keep the diff clean.

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -0,0 +1,1299 @@
...
+        # Add more custom logic after entering if needed
+        return result
+
+    def __exit__(self, *args: Any):
+        # Call the underlying context manager's __exit__
+        result = self.saved_tensors_context.__exit__(*args)
+        from megatron.core.extensions.transformer_engine import cpu_offload
+
+        if cpu_offload is not None:
+            cpu_offload.CPUOffloadEnabled = False
```
Victarry: **[SUGGESTION]** `PagedStashContext` unconditionally sets `CPUOffloadEnabled = True` on enter and `False` on exit without saving/restoring the original value.

Given that `transformer_config` already asserts `moe_paged_stash` cannot coexist with `cpu_offloading`, the intent of toggling `CPUOffloadEnabled` here is unclear. If this is needed for TE internal behavior, a comment explaining why would help. If not, consider removing it to avoid accidentally enabling CPU offload in unexpected contexts.

If it must stay, save and restore the original value:
```python
def __enter__(self):
    if cpu_offload is not None:
        self._prev_offload = cpu_offload.CPUOffloadEnabled
        cpu_offload.CPUOffloadEnabled = True
    ...
def __exit__(self, *args):
    if cpu_offload is not None:
        cpu_offload.CPUOffloadEnabled = self._prev_offload
```

megatron/core/transformer/moe/experts.py
```diff
@@ -931,19 +937,38 @@ def _fused_forward(
...
+            permuted_local_hidden_states = paged_stash_group_start(permuted_local_hidden_states)
+            max_num_tokens = permuted_local_hidden_states.shape[0]
+            # Average/expected tokens is a pre-padding estimate used by paged stashing heuristics.
+            # moe_expert_rank_capacity_factor is required when moe_paged_stash is enabled.
+            cap_factor = self.config.moe_expert_rank_capacity_factor
+            avg_num_tokens = (
+                int(max_num_tokens // cap_factor) if cap_factor is not None and cap_factor > 0 else None
+            )
+            stash_context = get_paged_stash_context(
+                name="expert_fc1_fused",
```
Victarry: It seems the activation save for fc2 is also included in the stash context?
Such that we could change the context name `expert_fc1_fused` to `grouped_mlp`

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -0,0 +1,1299 @@
...
+        if self.status == 'captured':
+            self.paged_tensors_to_stash.append(paged_tensor)
+        else:
+            pass
+
+    def remove_paged_tensor_from_stash(self):
+        """Remove all paged tensors from the stash list."""
+        if self.status == 'captured':
+            while len(self.paged_tensors_to_stash) > 0:
+                paged_tensor = self.paged_tensors_to_stash.pop(0)
```
Victarry:
```suggestion
            self.paged_tensors_to_stash.clear()
```

megatron/core/transformer/moe/paged_stash.py
nanz-nv: Removed

megatron/core/transformer/moe/paged_stash.py
nanz-nv: Removed

megatron/core/transformer/moe/paged_stash.py
nanz-nv: @vasunvidia

megatron/core/transformer/transformer_block.py
nanz-nv: removed

megatron/core/transformer/moe/paged_stash.py
nanz-nv: removed

megatron/core/transformer/moe/experts.py
nanz-nv: Thanks for the suggestion. Renamed

megatron/core/transformer/moe/paged_stash.py
nanz-nv: @vasunvidia

buptzyb:
Does paged stashing support partial cudagraph? We have many cases where attention has dynamic shapes (varlen, dynamic cp, KDA, ...), so only the moe part is capturable.

megatron/core/transformer/moe/paged_stash.py
vasunvidia: Done

megatron/core/transformer/moe/paged_stash.py
vasunvidia: Done

megatron/core/transformer/moe/paged_stash.py
nanz-nv: This has been added to Megatron too now.

PR #2693, opened 2025-12-17
feat: add routing replay for Mcore

litianjian:
# What does this PR do ?
This PR introduces a "Router Replay" feature for Mixture-of-Experts (MoE) layers. This functionality provides a deterministic routing mechanism, which is essential for debugging, controlled experimentation, and reproducing model behavior.

Inspired by recent approaches in stabilizing MoE models [Router Replay(R2)](https://github.com/THUDM/slime/pull/387) and [Rollout Router Replay(R3)](https://arxiv.org/abs/2510.11370), `RouterReplay` implementation allows developers to easily save and set the router's replay information, providing precise control over the expert selection process to mitigate routing inconsistencie
### Implementation Details:

1. Configuration Flag:

- A new boolean flag, enable_routing_replay , has been added to TransformerConfig . This allows users to enable or disable the feature globally.
2. RouterReplay Class:
- A new class, `RouterReplay` , is introduced in `moe_utils.py` to manage the state and data for the replay functionality.
- It operates in three modes, defined by the RoutingMode enum:
  - None : The default state, where standard routing occurs.
  - RECORD : The router records the expert indices ( topk_ids ) for each token.
  - REPLAY : The router bypasses the standard Top-K logic and instead uses the previously recorded expert indices to route tokens.
3. Integration with `TopKRouter` :
- In router.py , the TopKRouter now initializes a RouterReplay instance if config.enable_routing_replay is True .
- This instance is then passed to the `topk_routing_with_score_function` during the routing process.
4. Core Logic in moe_utils.py :

- The `topk_routing_with_score_function` has been updated to handle the router_replay object.
- When the mode is RECORD , it captures the topk_ids after they are computed.
- When the mode is REPLAY , it retrieves the stored topk_ids from the RouterReplay object and uses them to construct the routing gates and probabilities, effectively bypassing the dynamic routing calculation.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


ISEEKYAN:
@yanring

megatron/core/transformer/transformer_config.py
```diff
@@ -551,6 +551,9 @@ class TransformerConfig(ModelParallelConfig):
     moe_router_topk: int = 2
     """Number of experts to route to for each token."""
 
+    enable_routing_replay: bool = False
```
yanring, line 554: Could you also include this arg in arguments.py?

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -527,6 +526,140 @@ def pad_routing_map(routing_map: torch.Tensor, pad_multiple: int) -> torch.Tenso
     return routing_map
 
 
+class RouterReplayAction(Enum):
```
yanring: It would be nice if we could move this function to a new file, e.g., router_replay.py.

docs/source/api-guide/router_replay.md
```diff
@@ -0,0 +1,174 @@
+# Design Document: MoE Router Replay Feature
```
yanring, line 1: Thanks for the detailed documentation 👍

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -527,6 +526,140 @@ def pad_routing_map(routing_map: torch.Tensor, pad_multiple: int) -> torch.Tenso
...
+
+class RouterReplay:
+    """
+    A class to manage the recording and replaying of MoE routing decisions.
+    It holds all router instances and provides static methods to globally
+    control recording and replaying.
+    """
+
+    # Static variable to hold all router instances, one per MoE layer.
+    global_router_replay_instances: List['RouterReplay'] = []
```
yanring: Hey, I think there might be a memory leak here with `global_router_replay_instances`.

The list keeps growing because instances get appended in `__init__` but are never removed:

```python
def __init__(self):
    ...
    RouterReplay.global_router_replay_instances.append(self)
```

I see there's `clear_global_router_replay_action()` and `clear_global_indices()`, but those only clear the data *on* each instance — they don't actually remove instances from the list.

So if someone does something like:

```python
model1 = create_moe_model()  # say 32 MoE layers → 32 instances
del model1
model2 = create_moe_model()  # now we have 64 instances in the list...
```

The old instances stick around forever. I noticed the unit tests work around this by manually calling `.clear()` on the list.

One option would be to use `weakref.WeakSet` instead, that way when the router gets deleted, the instance automatically disappears from the set without needing manual cleanup. Just a thought!

**Suggested fix:** Use `weakref.WeakSet` instead of a regular list:

```python
import weakref

class RouterReplay:
    _global_instances: weakref.WeakSet['RouterReplay'] = weakref.WeakSet()
    
    @classmethod
    @property
    def global_router_replay_instances(cls) -> List['RouterReplay']:
        """Returns a list of all active RouterReplay instances."""
        return list(cls._global_instances)
    
    def __init__(self):
        ...
        RouterReplay._global_instances.add(self)
```

With `WeakSet`, the set only holds **weak references** that don't prevent garbage collection. When a `TopKRouter` (and its `RouterReplay`) is deleted, Python's GC automatically reclaims the object and the `WeakSet` automatically removes the stale reference.

Alternatively, if you prefer explicit control, you could add a public `clear_global_instances()` method and document that users should call it when recreating models.

docs/source/api-guide/router_replay.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+### 6. Minimal Demo
+
+Here is a minimal code example showing how to use RouterReplay for recording and replaying:
+
+```python
+import torch
+import torch.distributed as dist
+from megatron.core.config import TransformerConfig
```
yanring: Should be `from megatron.core.transformer.transformer_config `

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -527,6 +526,140 @@ def pad_routing_map(routing_map: torch.Tensor, pad_multiple: int) -> torch.Tenso
...
+    def get_replay_topk(
+        self,
+        scores: torch.Tensor,
+        topk: int,
+        num_groups: Optional[int] = None,
+        group_topk: Optional[int] = None,
+        default_compute_topk: Callable[
+            [torch.Tensor, int, Optional[int], Optional[int]], torch.Tensor
+        ] = None,
+    ) -> torch.Tensor:
```
yanring: The return type should be Tuple[torch.Tensor, torch.Tensor]

litianjian:
> Thanks for the great PR👍. I've left a few comments.

I am very honored to receive your reply and suggestions. I will follow your advice to modify the code.

megatron/core/transformer/moe/moe_utils.py
litianjian: Your suggestion is excellent， it would be better to add a method `clear_global_router_replay_instances`.

yanring (approved):

yanring:
/ok to test [745cb8f](https://github.com/NVIDIA/Megatron-LM/pull/2693/commits/745cb8f5c30e44ca79ff77781a8d55ae4d699df6)

ISEEKYAN (approved):

PR #2630, opened 2025-12-11
MoE Overlap: Add 2 CUDA Events to Synchronize Computation & Communication Operations for Forward and Backward Respectively

yanminjia:
# Problem Description

Please refer to issue [#2180](https://github.com/NVIDIA/Megatron-LM/issues/2180). MoE communication & computation cannot overlap completely.  

# Root Cause Analysis

Basically, forward computation of a micro-batch overlaps with backward communication of a different micro-batch and vice versa during interleaving phase. Therefore, the forward or backward process with respect to a transformer layer is split into a couple of  communication and computation modules which are managed by TransformerLayerSchedulePlan (megatron/core/model/common/mode_trunk_schedule_plan.py). Additionally, based on TransformerLayerSchedulePlan, the schedule plan of a model trunk is generated by TransformerModelChunkSchedulePlan (megatron/core/model/common/mode_trunk_schedule_plan.py). 

Roughly, in the interleaving phase, a forward process of a micro-batch over a transformer layer moves forward side by side with a backward process of a different micro-batch over a different transformer layer within a model trunk. For example, backward combine (communication) is scheduled to go in parallel with forward attention (computation) on different CUDA streams, forward dispatch (communication) is conducted wth backward mlp.

To synchronize the operations on communication stream and computation stream, one CUDA event is used to manage the dependencies of the sub modules in forward pass or backward pass. The computation and communication in forward pass are independent of the computation and communication in backward pass and vice versa. With only one CUDA event for synchronization, a computation operation in forward pass may wait for a communication operation in backward pass mistakenly.

# Solution

Add 2 CUDA events for communication & computation synchronization with regard to forward pass and backward pass respectively. A CUDA event is used for forward synchronization and a different CUDA is used for backward synchronization. 

# Test

As shown by below screen shot, forward attention overlaps with moe_combine backward.

5e060f2c-324f-4c87-addc-3418d7099226

Issue #2180, opened 2025-11-07
[QUESTION] MoE communication & computation can only overlap partially


yanminjia:
**In case of MoE communication & computation overlapping, why can only part of combine or dispatch be hidden by the forward/backward attention & mlp computation?**

I'm using nsight to learn the detailed MoE communication & computation overlapping. By checking the latest Megatron code, I understand that the overlapping happens during 1F1B phase of an interation. Basically, the forward communication overlaps with the backward computation and vice versa. But I'm confused that only part of communication can be hidden by the computation. As illustrated as the following figure, the last kernel function (triton_poi_fused_add_0) of forward combine overlaps with the first kernel function (vectorized_elementwise_kernel) of backward post-atten. Would be highly appreciated if any clue. Many thanks.
 
![Image](https://github.com/user-attachments/assets/54c3a953-dfea-4f58-8c78-7e314aa3c34f)

LiJunscs:
Hi, my megatron version is 0.14, I think the following codes in TransformerLayerSchedulePlan may cause some issues.
```
        if b_layer is not None:
            b_grad = b_layer.mtp_post_process.backward(b_grad)
            b_grad = b_layer.moe_combine.backward(b_grad)

        if f_layer is not None:
            with f_layer.get_fp8_context():
                f_input = f_layer.attn.forward(f_input)
                f_input = f_layer.post_attn.forward(f_input)

        if b_layer is not None:
            b_grad = b_layer.mlp.backward(b_grad)

        if f_layer is not None:
            with f_layer.get_fp8_context():
                f_input = f_layer.moe_dispatch.forward(f_input)

        if b_layer is not None:
            b_layer.mlp.backward_dw()
            b_grad = b_layer.moe_dispatch.backward(b_grad)

        if f_layer is not None:
            with f_layer.get_fp8_context():
                f_input = f_layer.mlp.forward(f_input)
```
This is my profile result.

<img width="1522" height="305" alt="Image" src="https://github.com/user-attachments/assets/febdca27-f258-4b35-954a-1c229b43cf70" />

In my opinion, the f_layer.moe_dispatch.forward should be execued before the b_layer.mlp.backward. I modify it. But due to I encounter another issue #1810 , I do not profile it for now. Could you provide me with the profile result if you can run it successfully? Thank you.

yanminjia:
[...]

PR #2730, opened 2025-12-21
[Dev](moe):Refine A2A overlap under CUDA_DEVICE_MAX_CONNECTIONS=1

Baidu-AIAK:
## Problem Description

The current overlap_moe_expert_parallel_comm in Megatron-LM requires CUDA_DEVICE_MAX_CONNECTIONS to be set to a relatively large value (https://github.com/NVIDIA/Megatron-LM/issues/2180#issuecomment-3515690690).

However, in certain scenarios, ensuring the correctness of communication and computation scheduling requires setting CUDA_DEVICE_MAX_CONNECTIONS=1. This constraint can weaken the parallelism of All-to-All (A2A) overlap in practice, leading to a noticeable gap between the achieved optimization benefits and the theoretical expectations.

## Root Cause Analysis

The A2A overlap optimization works by splitting modules and overlapping the computation of one micro-batch with the communication of another, thereby hiding EP communication latency. The current scheduling logic can be summarized as follows:

| Stream        | Phase 1                      | Phase 2                         | Phase 3        |
|--------------|------------------------------|----------------------------------|----------------|
| comm_stream  | combine_bwd                  | dispatch_fwd → dispatch_bwd      | combine_fwd    |
| comp_stream  | attn_fwd → post_attn_fwd     | mlp_bwd → mlp_bwd_dw → mlp_fwd   | post_attn_bwd → attn_bwd |

The theoretical execution timeline of the overlap optimization is illustrated below:

base-Theoretical

However, in memory-constrained scenarios where both TP and CP are greater than 1, we are forced to set CUDA_DEVICE_MAX_CONNECTIONS=1 to ensure the correctness of synchronization. Under this constraint, the effective execution timeline of the overlap becomes as follows:

base-Current

As can be observed, the original overlap logic is completely disrupted. This issue is also discussed in https://github.com/NVIDIA/Megatron-LM/pull/2630#issuecomment-3663307510 . We attribute this behavior to setting CUDA_DEVICE_MAX_CONNECTIONS=1, which enforces a serialized kernel submission model. Under this model, the launch order of computation and communication kernels is forced to be consistent across all devices, effectively eliminating the intended concurrency between computation and communication.

- We initially observed that when CUDA_DEVICE_MAX_CONNECTIONS=1 is set, achieving overlap between computation and communication requires launching the communication first, followed by the computation.

  - In the current implementation, the launch of mlp_bwd occurs before dispatch_fwd. As a result, under CUDA_DEVICE_MAX_CONNECTIONS=1, it is impossible to overlap mlp_bwd with dispatch_fwd. Although this overlap can be enabled by reordering the launch sequence, doing so degrades the overlap behavior of several subsequent modules. This issue is also discussed in https://github.com/NVIDIA/Megatron-LM/issues/2180#issuecomment-3507435596 .

  - As illustrated in the figure above, this launch-order constraint causes subsequent A2A communication modules to overlap with the next computation module instead, leading to a misaligned (shifted) overlap pattern.

- For combine_fwd and post_attn_bwd → attn_bwd, regardless of which is launched first, overlap cannot be achieved when CUDA_DEVICE_MAX_CONNECTIONS=1. This behavior is also reported in https://github.com/NVIDIA/Megatron-LM/issues/2180#issue-3599953469
  - We believe this is due to inherent data dependencies: the computation of mlp_bda must occur after mlp.combine, which forces combine_fwd, post_attn_bwd → attn_bwd, and PP_fwd to execute serially.

  - In the current implementation, combine_fwd is launched before post_attn_bwd → attn_bwd. However, because combine_fwd ends with a bda operation, it cannot overlap with post_attn_bwd → attn_bwd.

Furthermore, since the launch of PP_fwd occurs after post_attn_bwd → attn_bwd, PP_fwd cannot be overlapped either.

## Our Solution

Our solution can be summarized in two steps. First, we split the original combine module into two separate stages, combine and post_combine, thereby decoupling communication from computation within the combine phase. Second, by further adjusting the scheduling logic, we are able to achieve overlap between all A2A communications and PP communications, even when CUDA_DEVICE_MAX_CONNECTIONS=1 is enforced.

The newly designed scheduling logic can be summarized as follows:

| Stream        | Phase 1                                   | Phase 2        | Phase 3        | Phase 4                                             | Phase 5   | Phase 6   |
|--------------|-------------------------------------------|----------------|----------------|-----------------------------------------------------|-----------|-----------|
| comm_stream  | combine_bwd                               | dispatch_fwd   | dispatch_bwd   | combine_fwd                                        | PP_fwd    | PP_bwd    |
| comp_stream  | post_combine_bwd → attn_fwd → post_attn_fwd | mlp_bwd        | mlp_fwd        | mlp_bwd_dw → post_attn_bwd → post_combine_fwd       | attn_bwd  | attn_bwd_dw |

The theoretical execution timeline of the overlap optimization is illustrated below:

ours

## Evaluation

Finally, we evaluated the performance of the DeepSeek-V3.1 model on 64 Hopper-architecture GPUs using this optimization. With DeepEP optimization and A2A overlap enabled, and with CUDA_DEVICE_MAX_CONNECTIONS=1 set, our approach **achieves more than a 10% performance improvement in our target scenario compared to the current A2A overlap implementation in Megatron-LM**.

### Nsight Systems Comparison

Before Optimization (Nsight Systems):

image-2

As shown above, the actual Nsight Systems trace matches the previously illustrated behavior of the current implementation. The overlap logic is misaligned, resulting in exposed (non-overlapped) communication phases for both combine_fwd and PP_fwd.

After Optimization (Nsight Systems):

image-1

With our optimization applied, all A2A communications and PP communications are successfully overlapped.

## Summary

Overall, our optimization provides a more effective All-to-All (A2A) overlap solution for scenarios in which CUDA_DEVICE_MAX_CONNECTIONS=1 must be enforced. In other words, our approach makes A2A overlap optimization no longer heavily dependent on setting CUDA_DEVICE_MAX_CONNECTIONS=32.

### DIFF
--- a/megatron/core/models/common/model_chunk_schedule_plan.py
+++ b/megatron/core/models/common/model_chunk_schedule_plan.py
@@ -281,7 +281,7 @@ class TransformerLayerSchedulePlan:
                     f_input = f_layer.moe_dispatch.forward(f_input)
 
         if b_layer is not None:
-                b_grad = b_layer.mlp.backward(b_grad)
+            b_grad = b_layer.mlp.backward(b_grad)
 
         if not early_comm_launch:
             if f_layer is not None:


Baidu-AIAK:
Hi @Wohox. I'd really appreciate any feedback when you have time, especially if there are design or implementation concerns I should address. Thanks a lot!

megatron/core/models/common/model_chunk_schedule_plan.py
```diff
@@ -208,72 +217,156 @@ def get_fp8_context(self):
...
+            if b_layer is not None:
+                b_layer.mlp.backward_dw()
+                b_grad = b_layer.post_attn.backward(b_grad)
+
+            if f_layer is not None:
+                with f_layer.get_fp8_context():
+                    f_input = f_layer.post_combine.forward(f_input)
+                    f_input = f_layer.mtp_post_process.forward(f_input)
+
+            if is_last_layer:
```
Wohox: `post_forward` and `post_backward` shouldn't be called in layer schedule, you can wait for current stream in chunk schedule.

megatron/core/models/common/model_chunk_schedule_plan.py
```diff
@@ -208,72 +217,156 @@ def get_fp8_context(self):
...
+                if f_schedule_plan is not None and post_forward is not None:
+                    f_schedule_plan.wait_current_stream()
+                    post_forward(f_input, f_schedule_plan.vp_stage)
+            
+            if b_layer is not None:
+                b_grad = b_layer.attn.backward(b_grad)
+
+            if is_last_layer:
+                if b_schedule_plan is not None and post_backward is not None:
```
Wohox: Same as above

megatron/core/models/common/model_chunk_schedule_plan.py
```diff
@@ -208,72 +217,156 @@ def get_fp8_context(self):
...
+
         Args:
             f_layer (TransformerLayerSchedulePlan): Forward layer (for current microbatch)
             b_layer (TransformerLayerSchedulePlan): Backward layer (for previous microbatch)
             f_input (Tensor): Input for forward computation
             b_grad (Tensor): Gradient for backward computation
             is_last_layer_in_bwd (bool):
                 Whether the current layer is the last layer in the backward pass.
-
+            fine_grained_overlap (bool):
```
Wohox: - Can you run experiments to show spilt combine is always no less efficient than the original design(I think so even if max conn = 32), if so, there is no need to distinguish between the two.

megatron/core/models/common/model_chunk_schedule_plan.py
```diff
@@ -208,72 +217,156 @@ def get_fp8_context(self):
...
+                Enable fine-grained communication / computation overlap
+            post_forward (callable or None):
```
Wohox: Please remove `post_forward`, `post_backward`, `f_schedule_plan` and `b_schedule_plan` from function signature.

megatron/core/models/gpt/fine_grained_callables.py
```diff
@@ -474,8 +482,12 @@ def submodule_combine_forward(
         # Need to record residual to comm stream, since it's created on comp stream
         node.layer_state.residual.record_stream(torch.cuda.current_stream())
 
-        # release tensor reference after use
+        # release tensor references after use
+        if shared_expert_output is not None:
+            shared_expert_output.untyped_storage().resize_(0)
```
Wohox, line 488: Wondering what's special about `shared_expert_output` here, why can't it be deallocated like other detached tensors?

lhb8125:
Setting the CUDA_DEVICE_MAX_CONNECTIONS to 1 or 32 depends on which one is the most prominent factor when TP comm and EP comm both exist. The chance of TP comm overlapping is reduced but not impossible when setting it to 32 at TP=2, and in many cases, EP comm is the most prominent factor, so we think setting the env to 32 should be good as of now.

Could you perform some experiments on the DSV3 model to show the necessity of this PR?

Baidu-AIAK:
@lhb8125 Thank you for your response. Indeed, in most cases, setting CUDA_DEVICE_MAX_CONNECTIONS to 32 can achieve very good overlap for EP communication.

However, the motivation of this PR targets memory-constrained post-training scenarios with long sequences on a relatively small number of GPUs. For example, when training DeepSeek-V3 with long sequences (32K, 64K, etc.) on 128 GPUs using a parallel strategy of TP=8 and EP=16, TP communication accounts for about 5–6% of the runtime, while EP communication accounts for roughly 10%. In such cases, we usually aim to overlap both TP and EP communications to achieve optimal performance. Our approach provides an optimization opportunity for this specific scenario.

In addition, we have observed that setting CUDA_DEVICE_MAX_CONNECTIONS to 32 can conflict with certain optimizations, potentially leading to training instability, such as gradient norm becoming NaN.

We are preparing the required machines and will supplement this PR with additional experimental results soon to further demonstrate the necessity of this approach.

megatron/core/models/common/model_chunk_schedule_plan.py
Baidu-AIAK: Our original intention was to make minimally invasive changes to the original code and provide an incremental optimization. As you pointed out, if our approach can achieve performance that is no worse than the existing Megatron-LM implementation when CUDA_DEVICE_MAX_CONNECTIONS = 32, it would be reasonable to appropriately merge the two to reduce code redundancy. After careful consideration, we have revised the implementation accordingly and provide the following theoretical analysis and experimental results. If there are any edge cases that we have not considered, we would greatly appreciate your feedback.

**When CUDA_DEVICE_MAX_CONNECTIONS = 1**, our implementation outperforms the existing Megatron-LM implementation; theoretical analysis and experimental evidence have already been provided in the PR description.

**When CUDA_DEVICE_MAX_CONNECTIONS = 32:**

- **Theoretical analysis**

From a theoretical perspective, the computation–communication overlap execution timelines of the Megatron-LM implementation and ours are shown below. Although some execution orders differ, the overall effects are equivalent.

  1. Megatron-LM:
base-Theoretical

  2. Ours:
32-理论-ours


- **Experimental results**

We conducted experiments on DeepSeek-V3 with different configurations, including EP16 and EP32. The results show that the performance of our approach is almost identical to that of the existing Megatron-LM implementation, and in some cases even slightly better (as shown in item 3 below).

- **Additional observations**

Finally, we observed that in certain cases, even with CUDA_DEVICE_MAX_CONNECTIONS = 32, our approach outperforms the Megatron-LM implementation. As shown in the two figures below, because the combine_fwd communication in our approach is launched relatively earlier, it can exhibit better robustness in scenarios where CPU overhead is present.

  1. Megatron-LM:
image-mg


  2. Ours:
image-ours

njdongchenling: is post_combine module placed in the communication stream?

Baidu-AIAK: Yes. The post_combine module is executed on the communication stream.

PR #2738, opened 2025-12-23
Dynamic Inference | Evict and re-compute context requests.

lmcafee-nvidia:
# What does this PR do ?
Evict requests from context that overflow the paused buffer, and re-compute prefill.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1663,6 +1676,166 @@ def get_index_of_chunked_prefill_request(self) -> int:
...
+            newly_paused_request_ids.
+        """
+
+        # Assign released blocks to paused requests.
+        # todo: @shanmugamr, un-pause requests using FIFO, rather than LIFO.
+        resume_request_count = 0
+        if self.paused_request_count > 0:
+            active_block_count_avail = self.block_allocator.get_active_avail()
+            paused_block_counts = self.request_kv_block_counts[: self.paused_request_count]
+            paused_block_counts = paused_block_counts.flip(dims=[0])
```
santhnm2, line 1749: Is this meant to do the LIFO -> FIFO conversion?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1663,6 +1676,166 @@ def get_index_of_chunked_prefill_request(self) -> int:
...
+            paused_block_counts += 1  # +1 for newly added block
```
santhnm2, line 1758: Why is there a newly added block here?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1663,6 +1676,166 @@ def get_index_of_chunked_prefill_request(self) -> int:
...
+            paused_block_counts_cumsum = paused_block_counts.cumsum(dim=0)
+            resume_request_count = min(
```
santhnm2, line 1760: Can this be 0?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1663,6 +1676,166 @@ def get_index_of_chunked_prefill_request(self) -> int:
...
+        # Update tracking vars.
+        self.paused_request_count -= evict_request_count
+        self.total_request_count -= evict_request_count
+
+        # Reset unused block ids.
+        self.request_to_kv_block_ids[self.total_request_count:(self.total_request_count + evict_request_count)] = -1
+
+        return evict_request_ids
+
     # TODO: see if we can compile this function
```
santhnm2: Not related to this PR but we should remove this TODO since the plan is to do this asynchronously rather than compiling it / writing a custom kernel

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -384,36 +390,25 @@ def __init__(
...
         buffer_size_bytes = int(buffer_size_gb * 1024**3)
-        block_count_total = buffer_size_bytes // (
+        paused_buffer_size_bytes = (
+            0
+            if paused_buffer_size_gb is None else
+            int(paused_buffer_size_gb * 1024**3)
+        )
+        block_count = buffer_size_bytes // (
+            self.block_size_bytes + mamba_states_memory_per_request
+        )
```
santhnm2: Can you add a TODO here to revisit making this allocation decision more flexible with respect to how much memory we want to give to KV cache versus Mamba state? I believe vLLM and/or SGLang have a parameter that lets you control the fraction assigned to each.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1663,6 +1676,166 @@ def get_index_of_chunked_prefill_request(self) -> int:
...
+        evict_end_idx = self.paused_request_count
+        evict_request_idxs = torch.arange(
+            evict_start_idx,
+            evict_end_idx,
+            device=torch.cuda.current_device(),
+        )
+        evict_request_ids = self.request_ids[evict_start_idx:evict_end_idx].clone()
+
+        # Release memory.
+        self.release_memory_blocks_from_request_indexes(evict_request_idxs)
```
santhnm2, line 1838: I don't think we're releasing Mamba memory here which would cause issues

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -259,6 +264,7 @@ def __init__(
         num_attention_heads: int,
         max_sequence_length: int,
         buffer_size_gb: float,
+        paused_buffer_size_gb: float,
```
sidsingh-nvidia: is this only relevant for uvm=1?

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: nope, this doesn't impact the LIFO order at all, this flip is here simply because we resume paused requests from right to left, and to compute how many requests we have space for, we must do a right to left cumsum to determine how many total blocks will be used by any potentially resumed paused requests.

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: the +1 is because anytime we resume a paused request, that unpaused request will be starting a new memory block. For context, pausing happens after a request has generated the final token of a memory block (i.e., token 256 of that block), which means the very next token (whenever that request gets unpaused) will be in a new block. So, when we resume a paused request, we have to account for the fact that it will need an extra block beyond the ones that it has already used.

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: yeah `resume_request_count` can be zero, and that shouldn't cause any issues

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: please take a look at my updates.

Inside `release_memory_blocks_from_request_indexes()`, I call `self.mamba_metadata.free_slots()` [here](https://github.com/NVIDIA/Megatron-LM/pull/2738/files#diff-aa4eebdd4ac24bc9e7937775652f49028a0c648aab781dcc8ff0f243a4abaf1cR1716).

And inside `evict_overflow_paused_requests()`, I reset mamba block table entries to -1 [here](https://github.com/NVIDIA/Megatron-LM/pull/2738/files#diff-aa4eebdd4ac24bc9e7937775652f49028a0c648aab781dcc8ff0f243a4abaf1cR1876) for evicted requests.

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: it's relevant regardless of uvm's value ... but the difference is that:
- if uvm=0, the paused buffer is contained within the specificed `buffer_size_gb`, so the total buffer size is `buffer_size_gb` and the active buffer size is `buffer_size_gb - paused_buffer_size_gb`
- if uvm=1, the paused buffer is added onto the specified `buffer_size_gb`, so the total buffer size is `buffer_size_gb + paused_buffer_size_gb`, and the active buffer size is `buffer_size_gb`

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -390,7 +396,11 @@ def __init__(
 
         # Initialize block allocator.
         buffer_size_bytes = int(buffer_size_gb * 1024**3)
-        block_count_total = buffer_size_bytes // (
+        paused_buffer_size_bytes = (
+            0 if paused_buffer_size_gb is None else int(paused_buffer_size_gb * 1024**3)
+        )
+        block_count = buffer_size_bytes // (self.block_size_bytes + mamba_states_memory_per_request)
```
santhnm2, line 404: Maybe outside the scope of this PR, but should we consider separating the Mamba memory pool from the KV cache memory pool?

megatron/core/inference/contexts/dynamic_context.py
santhnm2: I see, could you add a comment summarizing this?

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Makes sense, could you also add a comment about this?

megatron/core/inference/contexts/dynamic_context.py
santhnm2: I think logically this looks good, you may want to run an end-to-end eval test with a small buffer size to confirm though

lmcafee-nvidia: I have run my own test with nano-v3 with a small buffer and confirmed that some of the requests are getting evicted, but their generations still look good. Is this the kind of eval you have in mind?

PR #2753, opened 2025-12-24
[main] feat(moe): Support apply wd to qk layernorm for Qwen3-Next (4/4)

yuzhongw-nvidia:
# What does this PR do ?

[MR to dev](https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/merge_requests/4070#note_41285641).

[Design doc](https://docs.google.com/document/d/1mGJDsqw4eej3965Ks65JQXuAqC8c_4Msnkm3btToHU4/edit?tab=t.0#heading=h.3ol9bs4x7jpw)

Qwen3-Next functionality PRs.
* https://github.com/NVIDIA/Megatron-LM/pull/1989
* https://github.com/NVIDIA/Megatron-LM/pull/2751
* https://github.com/NVIDIA/Megatron-LM/pull/2752

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/core/optimizer/optimizer_config.py
```diff
@@ -41,6 +41,15 @@ class OptimizerConfig:
     weight_decay: float = 0.01
     """Weight decay coefficient for L2 regularization."""
 
+    no_weight_decay_cond_type: Optional[str] = None
```
Phlip79: Why not make this a bool?

tests/unit_tests/test_optimizer.py
```diff
@@ -598,3 +598,74 @@ def test_get_megatron_optimizer_custom_process_groups_validation():
...
+        TransformerConfig(num_attention_heads=1, num_layers=1), ddp_config, model
+    )
+
+    # Create optimizer config with no_weight_decay_cond_type='apply_wd_to_qk_layernorm'
+    optimizer_config = OptimizerConfig(
+        optimizer='adam',
+        lr=0.01,
+        bf16=True,
+        use_distributed_optimizer=False,
+        no_weight_decay_cond_type_type='apply_wd_to_qk_layernorm',
```
Phlip79:
```suggestion
        no_weight_decay_cond_type='apply_wd_to_qk_layernorm',
```

megatron/core/optimizer/optimizer_config.py
yuzhongw-nvidia: Because we may have other types in the future, and if we make it a bool, we need to have a new flag next time, which is bloated.

ericharper (approved):

megatron/training/arguments.py
```diff
@@ -2032,6 +2045,12 @@ def _add_regularization_args(parser):
                        'numerical stability')
     group.add_argument('--sgd-momentum', type=float, default=0.9,
                        help='Momentum factor for sgd')
+    group.add_argument('--no-weight-decay-cond-type', type=str, choices=['apply_wd_to_qk_layernorm'],
```
jaredcasper, line 2127: Why are we adding a deprecated argument? Is this for checkpoint compatibility or something?

yuzhongw-nvidia: It is for compatibility, because it has been added to the dev branch for about half a year.

megatron/core/optimizer_param_scheduler.py
```diff
@@ -61,16 +63,30 @@ def combine_param_group_overrides(
...
+            if key.startswith("_"):
+                continue
             if key in combined_override:
                 if combined_override[key] != value:
                     raise ValueError(
                         f"Conflicting overrides for {key}: {combined_override[key]} and {value}"
                     )
             combined_override[key] = value
+
+    # Overrides that force overrides.
```
deepakn94: What does this mean?

megatron/core/optimizer_param_scheduler.py
```diff
@@ -33,6 +33,8 @@ class ParamGroupOverride(TypedDict):
     end_wd: float
     wd_mult: float
 
+    _force_override: bool = False
```
deepakn94: Why does this have an underscore at the front?

megatron/training/arguments.py
```diff
@@ -2038,6 +2049,8 @@ def _add_regularization_args(parser):
                        help='Dropout probability for hidden state transformer.')
     group.add_argument('--weight-decay', type=float, default=0.01,
                        help='Weight decay coefficient for L2 regularization.')
+    group.add_argument('--apply-wd-to-qk-layernorm', action='store_true',
```
deepakn94, line 2091: I'm not sure I understand what this option does. In particular, "as a special case"? What's the general case then?

megatron/training/arguments.py
```diff
@@ -2071,6 +2084,12 @@ def _add_regularization_args(parser):
                        help='How to perform NS calculation for tensor model parallel weights')
     group.add_argument('--muon-extra-scale-factor', type=float, default=1.0,
                        help='Additional scale factor for the muon update')
+    group.add_argument('--no-weight-decay-cond-type', type=str, choices=['apply_wd_to_qk_layernorm'],
+                       help='Type of no weight decay condition. Choices: '
+                       'DEPRECATED. Please use --apply-wd-to-qk-layernorm instead. '
```
deepakn94: Deprecation warning should go at the end of the help string.

jstjohn (changes requested):
I'm nervous about how this approach of a force_override will interact with multiple different kinds of merging. For example it seems to work well for the case of only overriding wd, but what if we also want to override LR (eg decoupled lr) and that has partial overlap with parameters that need to be wd unskipped?

Here is an alternative approach, please feel free to cherry pick the commit over: https://github.com/NVIDIA/Megatron-LM/pull/2968

The new design works by adding a new kind of predictate that handles the tuple of param,name. That is sufficient for modifying the weight decay skip rule with your filter for qk_layernorm. @FDecaYed came up with this idea which would have simplified his matching rule in megatron bridge for this same problem.

megatron/core/optimizer_param_scheduler.py
jstjohn: https://github.com/NVIDIA/Megatron-LM/pull/2968/ would get rid of this part of the logic, but the answer is that this current implementation adds wd overrides twice, once setting those qk_layernorm layers to wd=0, then later setting them to wd=1. Setting things twice breaks the old logic, so the work-around was to add a "force override" concept. The underscore at the beginning tells the optimizer loop to not include this key as a thing to override in the parameter group. I think the proposal I added in the PR above accomplishes the goal in a cleaner way.

megatron/training/arguments.py
jstjohn: Generally all len==1 (eg layernorm weights) or bias terms get added to the `wd=0` group. This says "do not add the q or k layernorm weights to the wd=0 group, leave them as wd=1".

yuzhongw-nvidia:
> I'm nervous about how this approach of a force_override will interact with multiple different kinds of merging. For example it seems to work well for the case of only overriding wd, but what if we also want to override LR (eg decoupled lr) and that has partial overlap with parameters that need to be wd unskipped?
> 
> Here is an alternative approach, please feel free to cherry pick the commit over: #2968, ([e57b2f5](https://github.com/NVIDIA/Megatron-LM/commit/e57b2f54738b329bc3aa23bceadafd352dd73a08))
> 
> The new design works by adding a new kind of predictate that handles the tuple of param,name. That is sufficient for modifying the weight decay skip rule with your filter for qk_layernorm. @FDecaYed came up with this idea which would have simplified his matching rule in megatron bridge for this same problem.

Thanks @jstjohn and @FDecaYed for your help. Your implementation is much cleaner, so I cherry-pick your changes. 

Hi @deepakn94 , could you please help take a look about the current version?

deepakn94:
/ok to test [2ababd2](https://github.com/NVIDIA/Megatron-LM/pull/2753/commits/2ababd26ae030cefd966d6f0f7d47bc773947b9e)

chtruong814:
fast merging since the functional tests on main were passing.  We had some issues with newer tests we were onboarding.  This one shoudl have merged earlier.

PR #2761, opened 2025-12-25
Replaces ModuleSpec with Protocols for some of the inputs to SelfAttention/CrossAttention

nschank:
# What does this PR do ?
Introduces `Protocols` for the first half of the inputs to `SelfAttention` and `CrossAttention`.

Stopped here just for ease of review/to keep things more focused; as more typing is corrected throughout the codebase, PRs will slowly get smaller since there's less to fix.

## Notes

* Added a `not_none` utility because the patterns used by config objects and imports make it a common need; it's optional though, I can drop it if it doesn't look useful.
* I had to update the return types and argument types of several fundamental modules. These should all be pretty trivial; `attention_mask` is the most interesting one, and I've verified that it is indeed allowed to be `None` in the implementations.
* Note that, without the odd `TYPE_CHECKING` escape hatch in the TE extension file, no type checking can be performed on any TE code because `te.pytorch.XXX` is not known to be a class so all "subclasses" in the file make the typechecker really unhappy. This doesn't affect any real code though.
* Apologies for the autoformatting in the examples - I can revert them if desired, but they all seemed correct so :shrug: 

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/extensions/transformer_engine.py
+++ b/megatron/core/extensions/transformer_engine.py
@@ -58,19 +58,19 @@ from megatron.core.utils import (
     is_torch_min_version,
 )
 
-if TYPE_CHECKING:
-    # For type checking, treat transformer_engine as always available.
+try:
     import transformer_engine as te
     from transformer_engine.pytorch.fp8 import FP8GlobalStateManager, fp8_autocast
 
     HAVE_TE = True
-else:
-    try:
+except ImportError:
+    if TYPE_CHECKING:
+        # For type checking, treat transformer_engine as always available.
         import transformer_engine as te
         from transformer_engine.pytorch.fp8 import FP8GlobalStateManager, fp8_autocast
 
         HAVE_TE = True
-    except ImportError:
+    else:
         from unittest.mock import MagicMock
 
         te = MagicMock()
--- a/megatron/core/transformer/attention.py
+++ b/megatron/core/transformer/attention.py
@@ -1154,18 +1154,6 @@ class Attention(MegatronModule, ABC):
                         attention_bias=attention_bias,
                         packed_seq_params=packed_seq_params,
                     )
-%%%%%%% Changes from base to side #2
--                core_attn_out = self.core_attention(
-+                core_attn_out = apply_module(self.core_attention)(
-                     query,
-                     key,
-                     value,
-                     attention_mask,
-                     attn_mask_type=attn_mask_type,
-                     attention_bias=attention_bias,
-                     packed_seq_params=packed_seq_params,
-                 )
->>>>>>> Conflict 1 of 1 ends
 
             else:
                 # Dynamic batching attention kernel.
--- a/megatron/core/typed_torch.py
+++ b/megatron/core/typed_torch.py
@@ -1,5 +1,6 @@
 # Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 """Utilities for improved type hinting with torch interfaces."""
+from __future__ import annotations
 
 from collections.abc import Callable
 from typing import Generic, ParamSpec, Protocol, TypeVar


nschank:
Rebased to capture [other PR](https://github.com/NVIDIA/Megatron-LM/pull/2766)

megatron/core/transformer/attention.py
```diff
@@ -280,7 +378,7 @@ def custom_forward(*inputs):
             attention_mask = inputs[3]
             attn_mask_type = inputs[5]
             attn_mask_type = AttnMaskType(attn_mask_type.item())
-            output_ = self.core_attention(
+            output_ = apply_module(self.core_attention)(
```
yashaswikarnati, line 401: @ericharper @jaredcasper  - for any thoughts on the usage of apply_module.

megatron/core/extensions/transformer_engine.py
```diff
@@ -61,16 +61,23 @@
     is_torch_min_version,
 )
 
-try:
+if TYPE_CHECKING:
```
yashaswikarnati: neat :)

yashaswikarnati:
Thank you for also fixing some incorrect type hints! Overall lgtm.

I'll let @jaredcasper @ericharper comment on if any thoughts on the usage of apply_module in the forward passes of modules

yashaswikarnati (approved):

megatron/core/transformer/attention.py
nschank: To give my thoughts on `apply_module`:

- Firstly, I won't be offended by trying to find an alternative (it works but I recognize it's odd)
- To me, the main drawbacks it has are that it can be distracting and ends up being in a bunch of places. However, I liken it to `cast` or little things like that - people used to the codebase very quickly start ignoring it and knowing when to use it without effort. (That's what happened in our old fork of Megatron anyway)
- As a major plus _during the migration_, it forces _me_ to show you every place where a particular module is being used because it gives me a reason to wrap every call, which might help spot unexpected diffs.
- In general, it makes it so each `Protocol` only needs to show exactly what it means without needing to follow any additional magical patterns (like the only potential alternative, shown below), and the weird complexity is instead wrapped in a unified interface that effectively becomes a 'code-base wide' central link back to a place that can explain what's going on.
- Finally, `apply_module` works even in cases where a module's type is known, while alternatives that rely on `Protocol` do not. For instance, in some submodules of `TransformerLayer`, the correct type is `BuilderProtocol | type[IdentityOp]`; `apply_module` is able to understand this, but if you try to use `__call__` in this case you will get an untyped result because `IdentityOp.__call__` is untyped. More broadly, using `apply_module` everywhere (even when you're using a non-configurable module with a fixed type) has nice typing benefits you can't easily get otherwise.

The only alternative idea that doesn't require providing the entire signature twice is to redefine `__call__` based on `forward` like this:

```python
class Unary(Protocol):
  def forward(self, x: torch.Tensor, /) -> torch.Tensor: ...

  __call__ = forward
```

I mentioned this might be workable with Yash, but after some research I think it's a bad idea. It works by happenstance, but has three pretty significant drawbacks:

 1. If you flip the order (define the signature for `__call__` and use `forward = __call__`) then this `Protocol` will silently not type-check correctly (it will accept any `Module` without complaint).
 2. It is doing something much more subtle than it appears - it is not equivalent to 'copying the signature of `forward` to `__call__`', it is the equivalent of _providing a default definition_ for `__call__` which is (in fact) possible to _use_. Just in broad terms, doing this would leave around a rather misleading/dangerous pattern for people to misunderstand.
 3. As an explicit example of where this could actually be a _bug_, if you explicitly extend that protocol as in `class Example(Unary, torch.nn.Module)` (note the order of subclasses), this would actually break the Module and override torch's builtin `__call__` behavior to simply call `forward` directly instead.

megatron/core/extensions/transformer_engine.py
```diff
@@ -2357,7 +2364,10 @@ def fused_apply_rotary_pos_emb_thd(
     pass
 
 try:
-    from transformer_engine.pytorch import Fp8Padding, Fp8Unpadding  # pylint: disable=unused-import
```
Phlip79, line 2385: This change fails the [linting test](https://github.com/NVIDIA/Megatron-LM/actions/runs/20767666139/job/59637134485?pr=2761).

megatron/core/transformer/attention.py
```diff
@@ -319,7 +417,9 @@ def _get_pp_layer_offset_for_inference(self):
         ), "Virtual pipeline parallelism is not supported for inference"
 
         # Import here to avoid circular imports
-        from megatron.core.transformer.transformer_layer import get_transformer_layer_offset
```
Phlip79, line 342: This change fails the [linting test](https://github.com/NVIDIA/Megatron-LM/actions/runs/20767666139/job/59637134485?pr=2761).

nschank:
Sigh, I'm still trying to get the formatting+import sorting working properly in my devbox, ruff+black are fighting so if I close a file early it sometimes does that. Fixed.

megatron/core/extensions/transformer_engine.py
nschank: Fixed thanks!

shanmugamr1992:
@yashaswikarnati  could you add tests

shanmugamr1992 (approved):

nschank:
Rebased to fix conflicts with changes to Attention module

yashaswikarnati:
/ok to test [77e87b7](https://github.com/NVIDIA/Megatron-LM/commit/77e87b74d66fa8a185c7b8ba1e9c9acb4f120427)

megatron/core/transformer/attention.py
```diff
@@ -110,14 +113,107 @@
...
+        gather_output: bool,
+        bias: bool,
+        skip_bias_add: bool,
+        is_expert: bool,
+        tp_comm_buffer_name: str,
+        tp_group: torch.distributed.ProcessGroup | None = None,
+    ) -> LinearQkv: ...
+
+
+class LinearLayer(Protocol):
```
jaredcasper: Should this be specific to attention? Would this same type apply across the board?

nschank: I'm sorta trying to take this migration incrementally and make each Protocol in place, without presuming its final location yet; presuming that two different modules can be reused early can end up backfiring because there can be subtle differences in their usage that only really become apparent once I start actually updating callsites. If I end up with duplicate Protocols in two places that seem they should be reused, we can make the call then whether to reuse them!

nschank:
A few small updates:
 * Renamed protocols for consistency with https://github.com/NVIDIA/Megatron-LM/pull/2891; will use this naming convention moving forward
 * Reordered the transformer_engine TYPE_CHECKING thing to make it a bit less likely people will accidentally only update the type-checking part.
 * Updated some type hints that were added to consistently use python3.12 syntax + `from __future__ import annotations` to allow for python3.10

megatron/core/transformer/attention.py
```diff
@@ -110,14 +113,107 @@
     HAVE_FUSED_QKV_ROPE = False
 
 
+class LinearQkvInterface(Protocol):
+    """Interface required for linear_qkv modules."""
+
+    def forward(self, input: Tensor, /) -> tuple[Tensor, object]:
```
yanring, line 123: neat: Maybe we could standardize on using either `Tuple` or `tuple` everywhere?

nschank: Yep! I'm actually using this as an excuse to move to python3.12 notation everywhere as I go; however since Megatron still offers support to 3.10 it requires `from __future__ import annotations` and I'm trying to avoid making too many unrelated changes in these meatier PRs. I'll send more cleanup/consistency-focused PRs as I go as well.

yashaswikarnati:
/ok to test [77e87b7](https://github.com/NVIDIA/Megatron-LM/commit/77e87b74d66fa8a185c7b8ba1e9c9acb4f120427)

yashaswikarnati:
/ok to test [5506f27](https://github.com/NVIDIA/Megatron-LM/pull/2761/commits/5506f27f1f979ab35d1f4420f04338779b4a9396)

nschank:
Rebased to fix conflicts!

yashaswikarnati:
/ok to test [2590b97](https://github.com/NVIDIA/Megatron-LM/commit/2590b973d78dfe281b732280d322ef293cdbefb6)

yashaswikarnati:
/ok to test [2590b97](https://github.com/NVIDIA/Megatron-LM/commit/2590b973d78dfe281b732280d322ef293cdbefb6)

yashaswikarnati:
/ok to test [94367ca](https://github.com/NVIDIA/Megatron-LM/commit/94367ca26311d414115aa044f28b25a27e5fe10d)

PR #2787, opened 2025-12-31
[draft] support async param gather in layer-wise optimizer

FDecaYed:
# What does this PR do ?
This is draft code now, so likely doesn't run. It's meant for demo how to support the feature before proper implementation.

## Current Architecture

DistributedOptimizer implements async param gather with these components:
- Bucket-based param organization - Parameters are grouped into buckets within _ParamAndGradBucketGroup
- start_param_sync() - Launches async all-gather using torch.distributed.all_gather with async_op=True
- finish_param_sync() - Waits for completion and dispatches the next bucket's all-gather
- Forward pre-hooks - Registered on modules to call finish_param_sync() when a module needs its params
- assume even sharding and allgather directly into tensor(param buffer)

## Changes needed for layer-wise

- let each bucket know which param is on which rank, and group them into lists(`lw_params_list` in addition to `params_list` that each bucket already holds, think it as per-bucket layerwise sharding)
- change `all_gather_into_tensor` to `all_gather(_v)`
- copy from all_gatherv output into param inside `finish_param_sync()`

### DIFF
--- a/megatron/core/distributed/param_and_grad_buffer.py
+++ b/megatron/core/distributed/param_and_grad_buffer.py
@@ -11,6 +11,7 @@ from typing import Dict, List, Optional
 
 import torch
 from torch.distributed import _coalescing_manager
+from torch._utils import _flatten_dense_tensors, _unflatten_dense_tensors
 
 import megatron.core.nccl_allocator as nccl_allocator
 from megatron.core import parallel_state
@@ -108,6 +109,12 @@ class _ParamAndGradBucket:
             self.param_to_index[param] = (offset, offset + param.numel())
             offset += param.numel()
 
+        self.lw_params_list = None
+        self.lw_param_flat_sizes = None
+
+    def set_lw_params_list(self, lw_params_list: List[List[torch.nn.Parameter]]):
+        self.lw_params_list = lw_params_list
+        self.lw_param_flat_sizes = [sum([p.numel() for p in param_list]) for param_list in lw_params_list]
 
 class _ParamAndGradBucketGroup:
     """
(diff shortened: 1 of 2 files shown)


megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -250,20 +257,41 @@ def start_param_sync(self, force_sync: bool = False):
...
-                    self.intra_distributed_optimizer_instance_rank
-                ]
-                dist_all_gather_func(
-                    bucket.param_data,
-                    local_data_view,
-                    group=self.intra_distributed_optimizer_instance_group,
-                    async_op=async_op,
-                )
+            else:
+                assert async_op, "Layer-wise optimizer requires overlap_param_gather=True"
```
mkhona-nvidia, line 276: does async allgather with layerwise still require use-distributed-optimizer? I think yes, to let DDP make the buckets right?

mkhona-nvidia (commented):
Another feature is that the current layerwise allows reshardable EP sizes in the middle of training. AFAIK, using adam's distopt does not allow this because of the way the DDP buckets are constructed. Can we avoid this and continue to allow reshardable EP sizes in the middle of training with layerwise?

megatron/core/optimizer/layer_wise_optimizer.py
```diff
@@ -45,6 +45,8 @@ def __init__(
         config: OptimizerConfig,
         pg_collection: Optional[ProcessGroupCollection] = None,
         init_state_fn_list: Optional[List[Callable]] = None,
+        model_chunks: Optional[List] = None,
+        async_allgather: Optional[bool] = False,
```
skyw, line 49: Reuse "overlap_param_gather" if the flag indicates the same thing, no need to introduce new names.

megatron/core/distributed/param_and_grad_buffer.py
FDecaYed: for simplicity to demo the idea, I didn't touch that part. currently if use-distributed-optimizer is off, then all async related functionality will be turned off and code errors out. 

But technically this is not required, we just need to change those check from if use-distributed-optimizer to if (use-distributed-optimizer or use-layer-wise)

FDecaYed:
> Another feature is that the current layerwise allows reshardable EP sizes in the middle of training. AFAIK, using adam's distopt does not allow this because of the way the DDP buckets are constructed. Can we avoid this and continue to allow reshardable EP sizes in the middle of training with layerwise?

my rough feeling is it should still be supported. the async feature is just moving param allgather from within optimizer.step into forward pre-hook calling bucket function. Checkpointing should not be affected

mkhona-nvidia:
> > Another feature is that the current layerwise allows reshardable EP sizes in the middle of training. AFAIK, using adam's distopt does not allow this because of the way the DDP buckets are constructed. Can we avoid this and continue to allow reshardable EP sizes in the middle of training with layerwise?
> 
> my rough feeling is it should still be supported. the async feature is just moving param allgather from within optimizer.step into forward pre-hook calling bucket function. Checkpointing should not be affected

I guess the issue is that we don't have a good understanding of why adamW distopt cannot allow resharding of EP in the middle of training. My feeling is that it is because of the way the DDP buckets are made (separate bucketing scheme for EP=true and separate bucketing scheme for rest of the network). If the layerwise distopt overlap_param_gather uses the same scheme again, I think we will run into the same issue

FDecaYed:
> > > Another feature is that the current layerwise allows reshardable EP sizes in the middle of training. AFAIK, using adam's distopt does not allow this because of the way the DDP buckets are constructed. Can we avoid this and continue to allow reshardable EP sizes in the middle of training with layerwise?
> > 
> > 
> > my rough feeling is it should still be supported. the async feature is just moving param allgather from within optimizer.step into forward pre-hook calling bucket function. Checkpointing should not be affected
> 
> I guess the issue is that we don't have a good understanding of why adamW distopt cannot allow resharding of EP in the middle of training. My feeling is that it is because of the way the DDP buckets are made (separate bucketing scheme for EP=true and separate bucketing scheme for rest of the network). If the layerwise distopt overlap_param_gather uses the same scheme again, I think we will run into the same issue

I think the EP buckets are separate from regular DP weight no matter if dist-opt is on or not. and async-gather doesn't affect it either.
Here is my guess why dist-opt doesn't support change EP(haven't check code recently, from memory so guess):
there are 2 view of the param buffer, the buffer view(every thing flattened) and the param view(list of param, with their .data backed by the cont buffer). the way dist-opt support torch_dist ckpting is just save the 'buffer view' as it is, which is tightly coupled with EP and cannot be changed. but layerwise safe by each parameter(param view), thus is not affected. 
and reason dist-opt have to save it that way is each master parameter/state can be sharded to multiple rank. this is not a problem for layerwise

FDecaYed:
implemented in https://github.com/NVIDIA/Megatron-LM/pull/3524
closing

PR #2816, opened 2026-01-06
Supporting inference when called within an asyncio loop

shanmugamr1992:
Export-Deploy faced this issue 
```
WARNING:pytriton.proxy.inference:Exception while performing inference on requests=00000002: Traceback (most recent call last):
  File "/opt/venv/lib/python3.12/site-packages/pytriton/proxy/inference.py", line 393, in _handle_requests
    async for responses in self._model_callable(requests):
  File "/opt/venv/lib/python3.12/site-packages/pytriton/proxy/inference.py", line 85, in _callable
    yield inference_callable(requests)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/pytriton/decorators.py", line 213, in batch
    outputs = wrapped(*args, **new_kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/pytriton/decorators.py", line 672, in wrapper
    return wrapped(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/Export-Deploy/nemo_deploy/llm/megatronllm_deployable.py", line 390, in triton_infer_fn
    output_infer = self._infer_fn(
                   ^^^^^^^^^^^^^^^
  File "/opt/Export-Deploy/nemo_deploy/llm/megatronllm_deployable.py", line 472, in _infer_fn
    results = self.generate(prompts, inference_params)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/Export-Deploy/nemo_deploy/llm/megatronllm_deployable.py", line 243, in generate
    results = self.mcore_engine.generate(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/megatron-lm/megatron/core/inference/engines/static_engine.py", line 323, in generate
    return self.generate_using_dynamic_engine(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/megatron-lm/megatron/core/inference/engines/static_engine.py", line 232, in generate_using_dynamic_engine
    return self.dynamic_engine.generate(prompts=prompts, sampling_params=sampling_params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/megatron-lm/megatron/core/inference/engines/dynamic_engine.py", line 1286, in generate
    result = self.step_modern()
             ^^^^^^^^^^^^^^^^^^
  File "/opt/megatron-lm/megatron/core/inference/engines/dynamic_engine.py", line 1255, in step_modern
    return self._loop.run_until_complete(self.async_step())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 663, in run_until_complete
    self._check_running()
  File "/usr/lib/python3.12/asyncio/base_events.py", line 624, in _check_running
    raise RuntimeError(
RuntimeError: Cannot run the event loop while another loop is running
```

This is a known issue with Python asyncio - you cannot call run_until_complete() from within an already running event loop. The error occurs because:
PyTriton runs its own async event loop internally
When it calls Megatron's generate() → step_modern(), the code tries to run self._loop.run_until_complete(self.async_step())
This fails because Python's asyncio doesn't allow nested run_until_complete() calls

### DIFF
--- a/.github/actions/action.yml
+++ b/.github/actions/action.yml
@@ -126,7 +126,7 @@ runs:
         IS_CI_WORKLOAD: ${{ inputs.is_ci_workload }}
       run: |
         PR_NUMBER=${{ fromJSON(steps.get-pr-info.outputs.pr-info || '{}').number }}
-        HAS_RUN_FUNCTIONAL_TESTS_LABEL=$(gh pr view $PR_NUMBER --json labels | jq '[.labels[].name] | any(. == "Run functional tests")')
+        HAS_RUN_FUNCTIONAL_TESTS_LABEL=$(gh pr view $PR_NUMBER --json labels | jq '[.labels[].name] | any(. == "Run functional tests")') || echo "$IS_CI_WORKLOAD"
         HAS_RUN_FUNCTIONAL_TESTS_LABEL=${HAS_RUN_FUNCTIONAL_TESTS_LABEL:-$IS_CI_WORKLOAD}
         echo "main=$HAS_RUN_FUNCTIONAL_TESTS_LABEL" | tee -a $GITHUB_OUTPUT
 
--- a/.github/oncall_schedule.json
+++ b/.github/oncall_schedule.json
@@ -1,9 +1,5 @@
 [
     {
-        "user": "Phlip79",
-        "date": "2026-01-07"
-    },
-    {
         "user": "BoxiangW",
         "date": "2026-01-14"
     },
@@ -26,5 +22,29 @@
     {
         "user": "janEbert",
         "date": "2026-02-18"
+    },
+    {
+        "user": "maanug-nv",
+        "date": "2026-02-25"
+    },
+    {
+        "user": "BoxiangW",
+        "date": "2026-03-04"
+    },
+    {
+        "user": "Phlip79",
+        "date": "2026-03-11"
+    },
+    {
+        "user": "asolergi-nv",
+        "date": "2026-03-18"
+    },
+    {
+        "user": "dimapihtar",
+        "date": "2026-03-25"
+    },
+    {
+        "user": "gautham-kollu",
+        "date": "2026-04-01"
     }
 ]
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -406,6 +406,7 @@ jobs:
           is_unit_test: "true"
           PAT: ${{ secrets.PAT }}
           container-image: ${{ env.container-registry }}/megatron-lm:${{ github.sha }}
+          is_ci_workload: ${{ needs.pre-flight.outputs.is_ci_workload }}
 
   cicd-parse-integration-tests:
     runs-on: ubuntu-latest
--- a/megatron/core/transformer/attention.py
+++ b/megatron/core/transformer/attention.py
@@ -1,5 +1,6 @@
 # Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 import copy
+import inspect
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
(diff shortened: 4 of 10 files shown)


kvareddy (approved):

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1331,11 +1332,29 @@ async def async_step(
...
+    def _run_coroutine_sync(self, coro):
+        """Run a coroutine synchronously, handling the case when already in an event loop.
+
+        This method safely runs an async coroutine from synchronous code, even when
+        called from within an already running event loop (e.g., when used with async
+        frameworks like pytriton).
+        """
+        try:
+            # Check if there's already a running event loop
+            asyncio.get_running_loop()
```
santhnm2, line 1371: Can we leverage this existing code somehow instead? @shanmugamr1992 @tdene 

https://github.com/NVIDIA/Megatron-LM/blob/a56a0b001dcb6cb50636691f60499a2598c351e4/megatron/core/utils.py#L2213-L2228

tdene: I wondered the same thing, but no, I do not see how it can be done.

The existing code handles the situation "We're in a synchronous context, and want to run async code. Let's make sure we run it in the correct loop!"

This PR handles the situation "We're in an asynchronous loop, and want to run more async code. Let's make sure we don't try to re-enter the loop we're already inside of." This is a new situation we have not previously encountered.

tdene:
@shanmugamr1992 @santhnm2 I am continuing this discussion in a separate thread.

This is the Pythonic way to implement this fix: #2831. Here are the differences between #2816 and #2831.

 * This #2816 has `generate` work no matter how the user calls it. #2831 instead throws an error (`Please call the async_generate method instead.`) if the user calls `generate` in an async context, asking the user to modify their code.
 * This #2816 is blocking, preventing the user's other work from running in parallel in the same event loop, potentially slowing down the user. #2831 instead is non-blocking, allowing the user's other work to run while the CPU idles during the forward pass.



After thinking about this yesterday, I decided on this current #2816 for its simplicity.     
Perhaps it's worth having a brief conversation with the users though: why are they running inference from inside an asynchronous context, and do they actually expect to run other workloads in parallel in the same event loop?

megatron/core/inference/engines/dynamic_engine.py
shanmugamr1992: That was my first attemp, but didnt quite work. So the problem is not about which loop to use. Its more around run_until_complete(). run_until_complete() is blocking, and if this loop is already running (drivin by something like pytriton) then youc ant have two things trying to drive the same loop.

santhnm2 (approved):

shanmugamr1992:
/ok to test [12abb38](https://github.com/NVIDIA/Megatron-LM/pull/2816/commits/12abb38b118c9a3d27482c5aa04fd9efb413a51e)

shanmugamr1992:
/ok to test [3c7a941](https://github.com/NVIDIA/Megatron-LM/pull/2816/commits/3c7a9416155fbd50e6c92ff9bd627853acae0cd9)

PR #2827, opened 2026-01-06
Fix: Skip JIT warmup when fusion is disabled via arguments

kisseternity:
# What does this PR do ?
This PR adds conditional checks to _warmup_jit_function in initialize.py to ensure JIT fusions (Bias+GeLU/SwiGLU and Bias+Dropout+Add) are only warmed up when they are enabled via arguments (e.g., checks args.bias_gelu_fusion, args.bias_dropout_fusion). This prevents unnecessary warmup execution when users explicitly disable these fusions using flags like --no-bias-gelu-fusion.

### Pre-checks

- [done ] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [done  ] I have added relevant unit tests
- [done  ] I have added relevant functional tests
- [done  ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [done  ] I have added relevant documentation
- [done  ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR


megatron/training/initialize.py
```diff
@@ -478,55 +478,57 @@ def _warmup_jit_function():
...
-    # prop and recomputation
-    for bias_grad, input_grad in zip([True, True], [False, True]):
-        bias.requires_grad, input.requires_grad = bias_grad, input_grad
-        for _ in range(5):
-            if args.swiglu:
-                output = bias_swiglu(input, bias)
-            else:
-                output = bias_gelu(bias, input)
-    del bias, input, output
+    if (args.swiglu and args.bias_swiglu_fusion) or (not args.swiglu and args.bias_gelu_fusion):
```
asolergi-nv: Some conditions are missing like `use_te_activation_func`. Can we add ALL the conditions to make sure we are only warming up the functions when necessary? You can use the [MLP](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/mlp.py#L163-L213) & [shared experts](https://github.com/NVIDIA/Megatron-LM/blob/43db8c1cf98f59d7c2fb84fbbb39c1ab05a68c6c/megatron/core/transformer/moe/shared_experts.py#L190-L210) as a reference

asolergi-nv (commented):
Looks nice, but we should check exhaustively the conditions that trigger the torch.compiled functions. Could you also add geglu? Thanks!

megatron/training/initialize.py
kisseternity: Thanks, I've added the missing conditions:
 use_te_activation_func check (skip warmup when TE activation is used)

kisseternity:
> Looks nice, but we should check exhaustively the conditions that trigger the torch.compiled functions. Could you also add geglu? Thanks!

Thanks for the feedback! I've updated the code to add exhaustive condition checks based on the MLP & shared experts implementation. Please let me know if any additional conditions should be considered.

megatron/training/initialize.py
```diff
@@ -470,56 +471,145 @@ def _warmup_jit_function():
...
+    #   - activation_func=F.gelu + gated_linear_unit=True  -> bias_geglu_impl
+    #   - activation_func=F.gelu + gated_linear_unit=False -> bias_gelu_impl
+    #   - activation_func=F.silu + gated_linear_unit=True  -> bias_swiglu_impl
+    #
+    # Args mapping:
+    #   - args.swiglu=True       -> gated_linear_unit=True, activation_func=F.silu
+    #   - args.quick_geglu=True  -> gated_linear_unit=True, activation_func=quick_gelu (no fusion warmup needed)
+    #   - default (neither set)  -> gated_linear_unit=False, activation_func=F.gelu
+
+    # gated_linear_unit can be set via YAML config or other means
```
asolergi-nv: Remove

megatron/training/initialize.py
```diff
@@ -470,56 +471,145 @@ def _warmup_jit_function():
...
-    input = torch.rand(
-        (
-            args.seq_length // args.context_parallel_size,
-            args.micro_batch_size,
-            args.ffn_hidden_size // args.tensor_model_parallel_size,
-        ),
-        dtype=dtype,
-        device="cuda",
+
+    # Warmup bias_gelu: non-gated gelu activation (F.gelu without GLU)
```
asolergi-nv: Remove

megatron/training/initialize.py
```diff
@@ -470,56 +471,145 @@ def _warmup_jit_function():
...
     )
-    # Warmup JIT fusions with the input grad_enable state of both forward
-    # prop and recomputation
-    for bias_grad, input_grad in zip([True, True], [False, True]):
-        bias.requires_grad, input.requires_grad = bias_grad, input_grad
-        for _ in range(5):
-            if args.swiglu:
+
+    # Warmup bias_geglu: gated gelu activation (F.gelu + GLU)
+    # This is triggered when gated_linear_unit=True with gelu activation
```
asolergi-nv: Remove

megatron/training/initialize.py
```diff
@@ -470,56 +471,145 @@ def _warmup_jit_function():
...
+    warmup_bias_geglu = (
+        not use_te_activation_func
+        and not args.swiglu
+        and not getattr(args, 'quick_geglu', False)
+        and gated_linear_unit
+        and args.bias_gelu_fusion
+    )
+
+    # Warmup fused bias+swiglu
```
asolergi-nv: Remove

megatron/training/initialize.py
```diff
@@ -470,56 +471,145 @@ def _warmup_jit_function():
...
+        # Warmup JIT fusions with the input grad_enable state of both forward
+        # prop and recomputation
+        for bias_grad, input_grad in zip([True, True], [False, True]):
+            bias.requires_grad, input.requires_grad = bias_grad, input_grad
+            for _ in range(5):
                 output = bias_swiglu(input, bias)
-            else:
+        del bias, input, output
+
+    # Warmup fused bias+gelu (non-gated)
```
asolergi-nv: Remove

megatron/training/initialize.py
```diff
@@ -470,56 +471,145 @@ def _warmup_jit_function():
...
+        # Warmup JIT fusions with the input grad_enable state of both forward
+        # prop and recomputation
+        for bias_grad, input_grad in zip([True, True], [False, True]):
+            bias.requires_grad, input.requires_grad = bias_grad, input_grad
+            for _ in range(5):
                 output = bias_gelu(bias, input)
-    del bias, input, output
+        del bias, input, output
+
+    # Warmup fused bias+geglu (gated gelu)
```
asolergi-nv: Remove

asolergi-nv (commented):
Just remove some comments & we are ready!

kisseternity:
> Just remove some comments & we are ready!

Addressed, thanks!

asolergi-nv:
/ok to test [201e816](https://github.com/NVIDIA/Megatron-LM/pull/2827/commits/201e81637423c216a30cb1201b92a1479c87b343)

asolergi-nv (approved):

kisseternity:
[@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) Hello, could you please approve the pending workflow run so the CI checks can proceed? The branch has been updated and is ready for testing. Thanks!

asolergi-nv:
/ok to test [11eb966](https://github.com/NVIDIA/Megatron-LM/pull/2827/commits/11eb966ea8c4ce05f49b1597841d2e360ee50bf6)

kisseternity:
[@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) Sorry to bother, but the CI seems to be failing again. I've checked the failing tests and they appear to be unrelated to the code changes in this PR.

kisseternity:
The following CI tests are failing and I am not sure why:
gpt/gpt3_mcore_te_tp1_pp4_resume_torch_dist_persistent_disable_bias_linear - latest
gpt/gpt3_mcore_te_tp1_pp4_resume_torch_dist_disable_bias_linear - latest
moe/gpt3_moe_mcore_te_ep8_resume_torch_dist_dist_muon - latest
moe/gpt3_moe_mcore_te_ep8_resume_torch_dist_dist_optimizer - latest
moe/gpt3_mcore_te_tp2_pp1_te_a2a_ovlp_8experts_etp1_ep4 - latest
moe/gpt3_moe_mcore_te_ep8_resume_torch_dist_muon - latest
The failures seem to be caused by mismatches in the loss comparison, but I am not sure why the loss values are failing to match.

gautham-kollu:
@kisseternity  request to add a unit-test

gautham-kollu:
Requesting @jaredcasper  for review

asolergi-nv (approved):

jaredcasper (approved):

asolergi-nv:
/ok to test [712951e](https://github.com/NVIDIA/Megatron-LM/pull/2827/commits/712951e1797e5e7173bd63e7242bda82409e8fbd)

asolergi-nv:
/ok to test [c61e8a5](https://github.com/NVIDIA/Megatron-LM/pull/2827/commits/c61e8a5c2c8a3f48fd6651e0e74714839f835fdc)

cspades:
@kisseternity There seems to be golden value failures for this PR during merge queue:

- https://github.com/NVIDIA/Megatron-LM/actions/runs/25146837002/job/73709929588
- https://github.com/NVIDIA/Megatron-LM/actions/runs/25146837002/job/73709929623

Not sure why, maybe has to do with the warmup skip and these two functional tests need to be updated, you can find them here: https://github.com/NVIDIA/Megatron-LM/tree/main/tests/functional_tests/test_cases/gpt

Can you take a look and tell me what's going on? You can update the golden values as well but we need to know why.

Phlip79:
@kisseternity please fix failing tests

guihong-nv:
@kisseternity May you please fix failing tests?

ilml:
/ok to test https://github.com/NVIDIA/Megatron-LM/commit/d3af9b3d620182590fde158050b12bc4af7c2449

ilml:
waiting on @kisseternity

janEbert:
I'm assuming that the contribution may have changed seeding, since certain `torch.rand` calls aren't being executed. Therefore, the golden values may also need to be updated.

kisseternity:
> I'm assuming that the contribution may have changed seeding, since certain `torch.rand` calls aren't being executed. Therefore, the golden values may also need to be updated.

@cspades @janEbert — figured out what's going on and pushed a fix: 8b94191.

**Root cause.** The PR didn't change the seed itself, but it changed how many
`torch.rand(..., device="cuda")` calls `_warmup_jit_function` makes. Every
warmup draw consumes the **default** CUDA RNG stream, and warmup executes
*after* seeding: `pretrain()` → `initialize_megatron()`
(training/training.py:1084) → `_set_random_seed` →
`model_parallel_cuda_manual_seed()` (which calls `torch.cuda.manual_seed()` on
the default generator), then `set_jit_fusion_options()` (training.py:1110) →
`_warmup_jit_function()`. Downstream code also relies on that default stream,
e.g. dropout in non-tensor-parallel regions (see the docstring of
`model_parallel_cuda_manual_seed` in core/tensor_parallel/random.py).

Previously the warmup executed 5 unconditional draws. With the new conditional
checks, configs that end up with a fusion arg disabled skip some draws —
explicit `--no-bias-gelu-fusion` / `--no-bias-dropout-fusion` / `--quick-geglu`,
and also `--disable-bias-linear`, because validate_args then forces
`args.bias_gelu_fusion = False` ("Disable bias gelu fusion if we are disabling
bias altogether", arguments.py:1410). The new GEGLU warmup additionally draws
2×-sized tensors for `--gated-linear-unit` configs. So the RNG stream position
at training start shifts — exactly as @janEbert suspected.

This is now concrete for the two failing jobs: both ran
`gpt3_mcore_te_tp1_pp4_resume_torch_dist(_persistent)_disable_bias_linear`,
which set `--disable-bias-linear: true` → `bias_gelu_fusion = False`, so the 2
activation-warmup draws that previously ran unconditionally are skipped (the 3
bias-dropout draws still run) — precisely the shift behind the golden-value
mismatch. It also explains why only these two jobs failed: the fusion flags
default to True (`store_false`), so configs without `--no-*-fusion` /
`--disable-bias-linear` / `--quick-geglu` / `--gated-linear-unit` still draw
byte-identically to before.

**Fix.** Instead of regenerating golden values, the new commit makes the warmup
side-effect free: snapshot the CUDA RNG state on entry
(`torch.cuda.get_rng_state_all()`) and restore it before returning
(`torch.cuda.set_rng_state_all()`), guarded by `torch.cuda.is_available()` so
CPU-only runs are unaffected. Skipping arbitrary warmup blocks can no longer
perturb training determinism, so the existing golden values should pass
unchanged. +9 lines in `megatron/training/initialize.py` only.

(The 2× sizing in the GEGLU warmup itself is intentional and kept: GEGLU
splits the last dim into two halves (`torch.chunk(y, 2, -1)`), so the previous
code was warming up with wrong shapes. With the RNG snapshot/restore in place
it no longer affects determinism either way.)

Could you re-trigger the merge queue / functional tests when you get a chance?
If the two disable-bias-linear gpt jobs go green with the golden values
untouched, that closes this out.

maanug-nv:
@kisseternity 
I guess we're still waiting on functional tests to confirm, but won't this last commit break a lot of functional tests? 
As you mentioned, the warmup previously executed unconditional draws. The golden values for all functional tests are depending on that. In other words, the numerics of the tests expect those unconditional draws in jit warmup.
Before 8b94191d5e8e7aefb67f6970926e4d693dda0164, some tests would skip those draws (again as you mentioned), affecting the numerics for those tests. After that commit, all tests are skipping those draws, correct? So, the numerics for all tests will change. After that commit.
The result of 8b94191d5e8e7aefb67f6970926e4d693dda0164 is that the new numerics are independent of RNG draws in jit warmup, which is a good design choice. however, if my understanding is correct and it requires updating golden values for many many more tests, i'm not sure if that's worthwhile. i dont think we see changes to the jit warmup that often.

Let me know if there's any flaw in that understanding. I guess the functional tests run on this commit will confirm/refute.

maanug-nv:
@kisseternity tests are passing, so this looks ready to merge. can you fix the DCO signoff? you'll just need to run `git rebase --signoff` for the commits that don't have a 'Signed off by' suffix.

kisseternity:
> @kisseternity tests are passing, so this looks ready to merge. can you fix the DCO signoff? you'll just need to run `git rebase --signoff` for the commits that don't have a 'Signed off by' suffix.

Thanks, DCO now fixed — ran `git rebase --signoff` onto the merge base. History is now linear with 6 commits, original authors kept, and every commit carries a Signed-off-by trailer (the bot/merge commits went away with the linearization).

PR #2836, opened 2026-01-07
Add retry loop with exponential backoff in dataloader as a form of in-application fault tolerance

deepakn94:


megatron/core/datasets/indexed_dataset.py
```diff
@@ -451,17 +452,45 @@ def read(self, dtype: Type[numpy.number], count: int, offset: int) -> numpy.ndar
...
+                with msc.open(self._bin_path, mode="rb", buffering=0) as bin_buffer_file:
+                    bin_buffer_file.seek(offset)
+                    bin_buffer_file.readinto(sequence)
+            else:
+                with open(self._bin_path, mode="rb", buffering=0) as bin_buffer_file:
+                    bin_buffer_file.seek(offset)
+                    bin_buffer_file.readinto(sequence)
+            return sequence
+
+        NUM_MAX_RETRIES = 3
```
jkamalu: Why are these in all caps? Why don't we make these __init__ args, or at least class args?

megatron/core/datasets/indexed_dataset.py
```diff
@@ -451,17 +452,45 @@ def read(self, dtype: Type[numpy.number], count: int, offset: int) -> numpy.ndar
...
+            try:
+                return _read()
+            except Exception as e:
+                time_str = datetime.now().strftime('%Y-%m-%d %H:%M:%S.%f')
+                if i == NUM_MAX_RETRIES:
+                    logger.warning(
+                        f"[{time_str}] {NUM_MAX_RETRIES+1} total tries to read data item failed; "
+                        f"going to abort and re-raise exception \"{e}\"..."
+                    )
+                    # Re-raise exception if in last iteration of for loop.
```
jkamalu, line 487: nit. can remove this comment probably.

megatron/core/datasets/indexed_dataset.py
```diff
@@ -451,17 +452,45 @@ def read(self, dtype: Type[numpy.number], count: int, offset: int) -> numpy.ndar
...
+                    raise e
+                logger.warning(
+                    f"[{time_str}] Attempt {i+1}/{NUM_MAX_RETRIES+1} to read data item failed "
+                    f"with exception \"{e}\"; going to sleep for {sleep_duration} seconds and "
+                    "then re-try..."
+                )
+                time.sleep(sleep_duration)
+                sleep_duration = sleep_duration * 2
+
+        raise RuntimeError("Should not reach here!")
```
jkamalu, line 497: nit. verbose.

jkamalu (changes requested):
Naming and scope of certain magic numbers and then some nits (which you can choose to ignore)

megatron/core/datasets/indexed_dataset.py
deepakn94: Added, though I didn't thread the two new arguments through all the classes from the `DatasetConfig` class since I think that's overkill.

jkamalu (approved):

PR #2837, opened 2026-01-07
Support DDP overlap for models with repeated parameters

deepakn94:


megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -165,7 +163,22 @@ def __init__(
         if self.ddp_config.reduce_scatter_with_fp32_accumulation:
             dist_reduce_scatter_func = reduce_scatter_with_fp32_accumulation
 
-        self.reset()
```
kvareddy, line 168: Why is the reset removed here?

deepakn94: Because I added some other functionality to `reset` (see below). I was calling `reset` earlier to initialize internal state, now just doing this directly in the constructor below (lines 176-179).

ericharper (approved):

PR #2857, opened 2026-01-07
Use global user buffer when the bucket size does not fit FixedPoolAllocator

shengf-nv:
# What does this PR do ?

FixedPoolAllocator maintains a double buffer so that one buffer holds the data for the current layer and another buffer is pre-fetching the data for the next layer, it can overlap computing and communication. The size of the double buffer in LLM is usually decided by the size of the transformer layers since their size are all the same. However, the embedding layer and output layer usually have different size. FixedPoolAllocator will fall back to dynamic memory allocation if the size does not fit. The dynamic allocated tensor can not be used as NCCL user buffer. In our SHARP performance test with Llama3 70B model,  the large NCCL collectives that do not use user buffer got split into many small collectives before sending down to SHARP. We observed 10x slow performance because of the user buffer issue.

To fix this issue, this PR is to use the buffer in global memory pool with a unique buffer name as the following:

{FixedPoolAllocator name}_not_fit_in_fixed_pool_{bucket_id}_{size}_{dtype}_{device}"

This fix will keep unique buffers for the buckets, which do not fit the size of FixedPoolAllocator, in the global memory pool. In LLM, they are the embedding layer and output layer. It will increase the peak memory size. In our Llama3 70B test, it is around 1.5%.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @megatron-oncall.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

Feel free to message or comment the @megatron-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: If we aren't using the backup allocator anymore, we can also remove this. @shengf-nv and just to make sure, the 1.5% memory increase is persistently buffering the unsharded embedding and output layers for 70B?

Nit: Best add some control or documentation or logging that informs users that any FSDP units that deviate from the memory profile of the double-buffer size will not be sharded / deallocated, so they will not be surprised that their asymmetric model architecture does not work / does not shard with `FixedPoolAllocator`.

shjwudp (changes requested):
Hi @shengf-nv , since we intend to treat persistent buffers in FSDP with great care and caution. Could you please add a switch to make this optimization optional?

shengf-nv:
@shjwudp and @cspades , thanks for your comments. 

@cspades, regarding your question "the 1.5% memory increase is persistently buffering the unsharded embedding and output layers for 70B?", embedding and output layers are still sharded, but these layers have dedicated persistent memory. If the layer size fits the double buffer size, this layer will reuse the shared memory in FixedPoolAllocator, which is the case for all transformer layers.

I added a new flag trainer.strategy.ddp.fsdp_db_fallback_dynamic_alloc to allow user to choose if they want to fallback to dynamic allocator, or use persistent memory. The default is True, so no change for existing models. 

I also added a log for the layers that do not fit the double buffer size, the following is an example log:
[NeMo I 2026-01-15 23:17:17 param_and_grad_buffer:675] FSDP unit (id = None) does not fit in FixedPoolAllocator
[NeMo I 2026-01-15 23:17:17 param_and_grad_buffer:677] It will fall back to dynamic memory allocator, NCCL user buffer is not supported

User can find the details of these layers (like parameter name, shapes) in the logs from self._log_parameter_groups() with FSDP unit id

Please help to review it. Thanks.

megatron/core/distributed/distributed_data_parallel_config.py
```diff
@@ -122,7 +122,15 @@ class DistributedDataParallelConfig:
       This option will cause additional memory overhead, however, it is necessary for
       to register user buffer (nccl_ub=True) for the Megatron FSDP. 
       This option will be automatically set to True when nccl_ub=True.
-   """
+    """
+
+    fsdp_db_fallback_dynamic_alloc: bool = True
```
shjwudp: Could we rename this flag to indicate that the double‑buffer fallback uses a persistent buffer?

By default, when double‑buffer allocation fails, we fall back to dynamic temporary memory allocation. For users who need the most aggressive optimization, enabling the flag (e.g. `use_persistent_buffer_on_fsdp_alloc_fail`) ensures that the communication memory NOT covered by double buffering is also allocated as persistent memory, so it can benefit from the performance improvements brought by NCCL‑UBR.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -678,6 +679,18 @@ def __init__(self, name: str, fsdp_param_groups: List["ParameterGroup"], size: i
         ), "Found no FSDP units to use fixed-size buffering"
         self.fsdp_double_buffer_units = fsdp_units_to_double_buffer
 
+        for bucket_id, param_group in enumerate(fsdp_param_groups):
+            if (param_group.fsdp_unit_id == -1 or param_group.fsdp_unit_id is None or 
+                param_group.fsdp_unit_id not in self.fsdp_double_buffer_units):
+                logging.info(f"FSDP unit (id={param_group.fsdp_unit_id}) does not fit in FixedPoolAllcator")
```
shjwudp: To avoid log flooding, it’s better not to print logs on every rank. You can wrap the logging statement with if `torch.distributed.get_rank() == 0`: so that only rank 0 prints logs.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -782,9 +799,10 @@ def free(self, bucket_id: int):
             self.idle_buffer.append(self.using_buffer[bucket_id])
             del self.using_buffer[bucket_id]
             return
-        # If not managed by fixed pool allocator, delegate to the backup allocator.
-        logging.debug(f"[FSDP] Free from the backup allocator for {bucket_id} {fsdp_unit_id}")
-        self.backup_allocator.free(bucket_id)
+        if self.fallback_to_dynamic_alloc is True:
+            # If not managed by fixed pool allocator, delegate to the backup allocator.
+            logging.debug(f"[FSDP] Free from the backup allocator for {bucket_id} {fsdp_unit_id}")
```
shjwudp, line 833: Do not print debug logs.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -1835,10 +1853,12 @@ def _init_each_parameter_group_buffers(self, meta_device_init_fp8_params):
...
             UB_BUFFER_NUM = 2
             self.weight_alloc = FixedPoolAllocator(
-                name="fsdp_params", fsdp_param_groups=self.parameter_groups, size=UB_BUFFER_NUM
+                name="fsdp_params", fsdp_param_groups=self.parameter_groups, size=UB_BUFFER_NUM, 
+                fallback_to_dynamic_alloc = self.ddp_config.fsdp_db_fallback_dynamic_alloc
             )
             self.main_grad_alloc = FixedPoolAllocator(
-                name="fsdp_grads", fsdp_param_groups=self.parameter_groups, size=UB_BUFFER_NUM
+                name="fsdp_grads", fsdp_param_groups=self.parameter_groups, size=UB_BUFFER_NUM,
+                fallback_to_dynamic_alloc = self.ddp_config.fsdp_db_fallback_dynamic_alloc
```
shjwudp: The code format here is incorrect; you can use `autoformat` tool to check it.

```bash
export PATH=".venv/bin:$PATH"
export GITLAB_ENDPOINT=github.com
export CI_PROJECT_NAMESPACE=NVIDIA
export BASE_REF="main"
export SKIP_DOCS=false
export CHECK_ONLY=true
bash tools/autoformat.sh
```

megatron/core/distributed/distributed_data_parallel_config.py
shengf-nv: @shjwudp , It is a good idea to have persistent buffer in the flag so user might be aware of the cost of this optimization. At the same time, I want to have fsdp_dp as a prefix, so fsdp related flags can be grouped together. How about fsdp_db_use_persist_buf_on_alloc_fail?

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
shengf-nv: will fix it.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
shengf-nv: will fix it

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
shengf-nv: This log exists in the version from main branch. Why do you want to remove it? It wont's show if the logger level is INFO.

megatron/core/distributed/distributed_data_parallel_config.py
shjwudp: Sounds good.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
shjwudp: OK, I misunderstood; we can keep it.

shengf-nv:
@shjwudp @cspades , I fixed the code review comments, please take a look again. Thanks

megatron/core/distributed/distributed_data_parallel_config.py
```diff
@@ -122,7 +122,16 @@ class DistributedDataParallelConfig:
       This option will cause additional memory overhead, however, it is necessary for
       to register user buffer (nccl_ub=True) for the Megatron FSDP. 
       This option will be automatically set to True when nccl_ub=True.
-   """
+    """
+
+    fsdp_db_use_persist_buf_on_alloc_fail: bool = True
```
youngeunkwon0405: Hi @shengf-nv, could you please change this default to False?

shengf-nv: Great catch! Fixed.

shengf-nv:
@youngeunkwon0405 , I fixed your comment, please take a look again. Thanks.

dimapihtar (approved):
LGTM. Thank you!

shengf-nv:
@jaredcasper @ericharper , can you help review this PR? Thanks

yaoyu-33 (approved):

PR #2878, opened 2026-01-08
RL refit pipelining support

wdykas:
# What does this PR do ?
This enables changing the pipeline size between training and inference for RL. It also fixes a bug that the wrong model was being passes into the eval section when using refit causing it to be off.

Here are some verification runs that were done for comparing changing PP and TP to the baseline no refit

image

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

Feel free to message or comment the @megatron-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


wdykas:
/ok to test [ef0e340](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/ef0e340a53f799478d2b5159ef26fbf45420a7fb)

wdykas:
/ok to test [74d6cd2](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/74d6cd2c0132cced3f61dc3642a0389de15a0ed8)

wdykas:
/ok to test [74d6cd2](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/74d6cd2c0132cced3f61dc3642a0389de15a0ed8)

wdykas:
/ok to test [23dd22d](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/23dd22dce68a53f5d63f25b6fdafa8580d11e420)

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -264,6 +265,7 @@ def __init__(
         block_size_tokens: int = 256,
         tensor_model_parallel_size: Optional[int] = None,
         pipeline_model_parallel_size: Optional[int] = None,
+        pg_collection: Optional[ProcessGroupCollection] = None,
```
wdykas, line 268: @santhnm2 For RL refit we are going to need to pass the pg_collection to grab the PP group, but this is starting to get a bit weird with like tensor_model_parallel_size, pipeline_model_parallel_size, and pg_collection all being passed. We could have this context only take in a pg_collection or we could just leave as is. Do you have any thoughts on how these should be handled?

wdykas:
/ok to test [d825eab](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/d825eab1f4bb8284017787639d40912c9fcf3e49)

wdykas:
/ok to test [d825eab](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/d825eab1f4bb8284017787639d40912c9fcf3e49)

wdykas:
/ok to test [596e7db](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/596e7dbb7d52dd2749018f5682f79c835cfc998a)

megatron/core/inference/contexts/dynamic_context.py
santhnm2: I would be ok with just passing in the `pg_collection` and getting rid of the separate `tensor_model_parallel_size` and `pipeline_model_parallel_size` arguments. We would just have to be careful to update all the places where we initialize the context. Btw, I think the eventual long-term solution is moving to a model where we just pass the `TransformerConfig` (and/or an even finer-grained `InferenceConfig`) which would anyway remove the need for these individual arguments.

wdykas: ok we haven't unified everywhere we set this right so I need to go find it everywhere?
Also we still need to pass the pg_collection even if we pass TransformerConfig because this file uses the actual group it seems and not just the size

santhnm2: Yea still need to unify all the calls :/ 
And yea I agree, we would still need to pass the `pg_collection`

santhnm2: Maybe we can merge this in as is and then do a follow-up PR to clean up all the `DynamicInferenceContext` initializations + remove the TP / PP args?

wdykas: maybe let me see how big the blast radius is if I try fixing partially in the is MR. If its too big then I think its own clean up MR makes more sense or else this MR would grow to large.

wdykas:
/ok to test [fbb89e4](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/fbb89e4a5c561cc7705fd27df35cbc08eb956ee5)

megatron/core/resharding/utils.py
```diff
@@ -157,12 +175,81 @@ def assign_resolved_name_inplace(meta: ParameterMetadata) -> None:
...
+    layer module's `layer_number`.
+
+    Implementation:
+    - Cache a `{local_prefix -> global_prefix}` map on the root module.
+    - Perform a longest-prefix match replacement so we only rewrite the module path portion.
+    """
+    prefix_map = getattr(module, '_reshard_layer_module_prefix_map', None)
+    if prefix_map is None:
+        prefix_map = _build_layer_module_prefix_map(module)
+        setattr(module, '_reshard_layer_module_prefix_map', prefix_map)
```
tdene: Instead of passing `module` in through the execution stack and running `_build_layer_module_prefix_map(module)` down here, I think it'd be a lot cleaner to grab this map up in `build_centralized_reshard_plan` where you already have the models and you can build the maps before you start looping over all the params.

Unless I'm misunderstanding, the `prefix_map` building looks to be param-agnostic, so it can be built up there?

megatron/rl/inference/megatron.py
```diff
@@ -150,8 +154,8 @@ def get_dynamic_inference_engine(
         buffer_size_gb=args.inference_dynamic_batching_buffer_size_gb,
         max_requests=args.inference_dynamic_batching_max_requests,
         max_tokens=args.inference_dynamic_batching_max_tokens,
-        tensor_model_parallel_size=inference_tp_size,
-        pipeline_model_parallel_size=inference_pp_size,
+        # TP/PP sizes are derived from the model's pg_collection.
+        pg_collection=pg_collection,
```
tdene: complete nit: personally I strongly dislike full-line comments inside argument lists, and would rather see
```
pg_collection=pg_collection, # TP/PP sizes are derived from the model's pg_collection.
```

megatron/rl/parallel_utils.py
```diff
@@ -29,11 +29,11 @@ def build_inference_pg_collection(
...
 
     Args:
-        tp_size: Tensor model parallel size for the inference model.
         world_size: Total world size (number of ranks).
+        tp_size: Tensor model parallel size for the inference model. Defaults to current MPU value if None.
+        pp_size: Pipeline parallel size. Defaults to current MPU value if None.
         use_tp_pp_dp_mapping: If True, use 'tp-cp-ep-pp-dp' order; otherwise 'tp-cp-ep-dp-pp'.
         cp_size: Context parallel size. Defaults to current MPU value if None.
-        pp_size: Pipeline parallel size. Defaults to current MPU value if None.
         ep_size: Expert parallel size. Defaults to current MPU value if None.
```
tdene, line 36: nit: reorder arg docstrings in the same order as the args themselves

megatron/rl/rl_utils.py
```diff
@@ -1671,9 +1670,11 @@ def megatron_rl_inference_mode(
 
 
 def rl_inference_interface_shutdown():
+    global _INFERENCE_INTERFACE
     if _INFERENCE_INTERFACE is not None:
         loop = get_asyncio_loop()
         loop.run_until_complete(_INFERENCE_INTERFACE.kill())
+        _INFERENCE_INTERFACE = None
```
tdene, line 1677: Agreed! Thank you!

megatron/training/training.py
```diff
@@ -863,7 +874,9 @@ def pretrain(
         prefix = f'iteration {iteration} on validation set'
         if getattr(args, 'perform_rl_step', False):
             rl_utils.evaluate_and_print_results_rl(
-                valid_data_iterator, model, optimizer,
+                valid_data_iterator,
+                inference_model if inference_model is not None else model,
+                optimizer,
```
tdene, line 889: If we're changing this to use the `inference_model`, then we assume that we have already run `swap_model_weights`, correct?

I do not see how `swap_model_weights` could have been run at this point in time, right after we come out of `train()`.

megatron/training/arguments.py
```diff
@@ -2115,6 +2115,12 @@ def _add_rl_args(parser):
                             'round-robin: distribute bins cyclically across ranks for better load balancing')
     group.add_argument('--rl-inference-tensor-model-parallel-size', type=int, default=None,
                        help='Degree of tensor model parallelism for inference for RL.')     
+    group.add_argument(
+        '--rl-inference-pipeline-model-parallel-size',
+        type=int,
+        default=None,
+        help='Degree of pipeline model parallelism for the separate RL inference model.',
```
yobibyte: Nit: replace 'separate'. From this descrption it's unclear, if we have another inference model for RL or something else.

megatron/core/resharding/utils.py
```diff
@@ -157,12 +175,81 @@ def assign_resolved_name_inplace(meta: ParameterMetadata) -> None:
...
+    name = meta.name if base_name is None else base_name
+    if module is not None:
+        name = _resolve_global_layer_number_in_name(name, module)
+    assign_ep_resolved_name_inplace(meta, base_name=name)
+
+
+def _build_layer_module_prefix_map(module: torch.nn.Module) -> dict[str, str]:
+    """Build a mapping `{local_module_prefix -> global_module_prefix}` for PP layer modules.
+
+    Megatron assigns a *global*, 1-indexed `layer_number` to each transformer layer module at
```
yobibyte: Not sure how stable this convention. Should we add a test/assert so that in the future when the 'global/1-indexed layer number' is changed by megatron we could quickly identify the issue?

E.g. if megatron switches to a zero-indexed numeration, we will get a negative number in here. We can check for that. 

But again, depends on how stable we believe the numeration convention is.

wdykas: that indexing numeration will not change. It had been standard for a long time and would break a ton of things in Megatron if it was ever changed since its used in a ton of places

megatron/training/arguments.py
wdykas: removed to match TP rl arg

megatron/training/training.py
wdykas: yeah your right, I guess it looks ok in my runs since it would at most be off by a single optimizer step. I have added the swapping of weights properly in both places we do rl evaluation. Seems like I had missed the other call site as well so this was a good catch.

megatron/rl/parallel_utils.py
wdykas: done

megatron/rl/inference/megatron.py
wdykas: yup agreed. Done

megatron/core/resharding/utils.py
wdykas: yeah like this. I think I did what you are referring to

wdykas:
/ok to test [aa09290](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/aa092907f47e7f4faa788ed5986702e6aab30774)

megatron/core/resharding/utils.py
yobibyte: sounds good to me

wdykas:
/ok to test [298ba7c](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/298ba7cd9d213007ed66aefa244894d884d813e2)

wdykas:
/ok to test [486f88b](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/486f88bcdf191527fccee554156d4198abe83fcb)

wdykas:
/ok to test [0eca251](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/0eca25152ac8b242b6e187b8ff3cc50c33890f0f)

wdykas:
/ok to test [0eca251](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/0eca25152ac8b242b6e187b8ff3cc50c33890f0f)

wdykas:
/ok to test [7a2fcdb](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/7a2fcdbc448951851af81a81b6c381cec632b53b)

wdykas:
/ok to test [71fd639](https://github.com/NVIDIA/Megatron-LM/pull/2878/commits/71fd639db1d087262540d21525f7dcf5ac2ad09f)

PR #2895, opened 2026-01-10
fix(dist_ckpt): Remove step tensor from sharded param state to fix MoE/EP checkpoint validation

jreiml:
## Summary

Fixes false checkpoint validation warnings when using Expert Parallelism (EP) with MoE models.

The `step` tensor is skipped during ShardedTensor conversion in `_get_param_state_sharded_tensors` but not removed from the tensors dict. This causes the plain tensor to leak into `common_state_dict` during `extract_sharded_base()`.

With Expert Parallelism, different ranks have different numbers of parameters (experts are distributed), resulting in different `param_idx` counts. The leaked step tensors create keys like `('optimizer', 0, 'param_state', N)` in `common_state_dict`, where N varies by rank. This triggers false warnings in `_validate_common_state_dict` because `common_state_dict` is expected to be identical across ranks:

```
WARNING: Rank 4 common state dict differs from rank 0 common state dict.
Keys only on rank 0: [], Keys only on 4: [('optimizer', 0, 'param_state', 1), ...]
```

The step is already properly saved via `param_groups` (as noted in the existing comment), so removing it from the per-param tensors dict is safe and correct.


Phlip79:
Please add a unit test that will cover this exception.

jaredcasper (approved):

jreiml:
@Phlip79 @jaredcasper I added the test.

ericharper (approved):

gautham-kollu:
@jreiml  can you look the failing unit tests ?

Phlip79:
@jreiml your unit test is failing. Please fix.

yaoyu-33 (approved):

PR #2896, opened 2026-01-10
Generate arguments from TransformerConfig

maanug-nv:
# What does this PR do ?
Generate argparse arguments from `TransformerConfig`. 

All changes to `TransformerConfig` or `ModelParallelConfig` are in an attempt to match existing arguments in `arguments.py`. Four categories:

- Tweak the type annotation in the dataclass to help auto-generating arguments
- Change the defaults in the dataclasses to match `arguments.py` so that auto-generated arguments exactly match existing arguments' behavior
- Add metadata for argument generation, mainly tweaks to CLI flag name ('dest' should always match the name of the attr in the dataclass)
- Tweak docstring. Sometimes adding info from the argument helpstring, sometimes making existing docstring more compatible as a helpstring (since docstrings will be used for helpstrings in auto-generated arguments).

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/core/model_parallel_config.py
```diff
@@ -57,7 +57,9 @@ class ModelParallelConfig:
     """Distributes Moe Experts across sub data parallel dimension."""
 
     expert_tensor_parallel_size: Optional[int] = None
-    """Intra-layer tensor model parallelsm for expert layer. Splits tensors across GPU ranks."""
+    """Intra-layer tensor model parallelsm for expert layer. Splits tensors across GPU ranks.
```
lmcafee-nvidia: 'parallelsm' -> 'parallelism'

megatron/core/transformer/transformer_config.py
```diff
@@ -738,8 +769,10 @@ class TransformerConfig(ModelParallelConfig):
     inference_sampling_seed: int = 42
     """ Random seed to use for sampling during inference. """
 
-    symmetric_ar_type: Optional[str] = None
-    """Type of symmetric all reduce to use"""
+    symmetric_ar_type: Optional[Literal['two_shot', "one_shot", "multimem_all_reduce"]] = None
+    """What type of symmetric all reduce to use. The default is None
+    which is no use of symetric memory.
```
lmcafee-nvidia: 'symetric' -> 'symmetric'

megatron/training/arguments.py
```diff
@@ -2689,10 +2447,6 @@ def _add_distributed_args(parser):
     group.add_argument('--no-scatter-gather-tensors-in-pipeline', action='store_false',
                        help='If not set, use scatter/gather to optimize communication of tensors in pipeline.',
                        dest='scatter_gather_tensors_in_pipeline')
-    group.add_argument('--use-ring-exchange-p2p', action='store_true',
```
lmcafee-nvidia, line 2511: did this arg get moved to the dataclass?

megatron/training/arguments.py
```diff
@@ -1922,10 +1865,6 @@ def _add_ft_package_args(parser):
 
 
 def _add_config_logger_args(parser):
```
lmcafee-nvidia: this function is empty ... we should remove it.

megatron/core/transformer/transformer_config.py
```diff
@@ -42,14 +42,22 @@ class TransformerConfig(ModelParallelConfig):
     # model architecture
     ####################
 
-    num_layers: int = 0
+    num_layers: int = field(default=0, metadata={"argparse_meta": {"default": None}})
```
Wohox, line 45: Why 2 different default values? I noticed this is true for many flags, 1 default 1 metadata default, what's the difference?

megatron/core/transformer/transformer_config.py
```diff
@@ -42,14 +42,22 @@ class TransformerConfig(ModelParallelConfig):
...
-    """Number of Multi-Token Prediction (MTP) Layers."""
+    """Number of Multi-Token Prediction (MTP) Layers.
+    MTP extends the prediction scope to multiple future tokens at each position.
+    This MTP implementation sequentially predict additional tokens
+    by using D sequential modules to predict D additional tokens.
+    """
 
-    mtp_loss_scaling_factor: Optional[float] = None
-    """Weighting factor of Multi-Token Prediction (MTP) loss."""
+    mtp_loss_scaling_factor: Optional[float] = 0.1
```
Wohox, line 55: Why change the default value to 0.1?

megatron/training/arguments.py
```diff
@@ -1678,36 +1595,98 @@ def _add_retro_args(parser):
 
 
 def _add_network_size_args(parser):
+    exclude = [
```
Wohox, line 1529: Can you explain more details for these reasons you listed, since I am not sure excluding them here is the correct action:
* cannot provide callables over CLI
* types affect docstring
* no CLI argument exists for these
* deprecated and no CLI arg exists
* already generated by another config
* incompatible defaults in dataclass

megatron/core/model_parallel_config.py
maanug-nv:
```suggestion
    """Intra-layer tensor model parallelism for expert layer. Splits tensors across GPU ranks.
```

megatron/core/transformer/transformer_config.py
maanug-nv:
```suggestion
    which is no use of symmetric memory.
```

megatron/training/arguments.py
maanug-nv: Yes it's in `ModelParallelConfig` [here](https://github.com/maanug-nv/Megatron-LM/blob/5e49b2dfa5296acfbd0c039a7e299a9651e7d14e/megatron/core/model_parallel_config.py#L296)

megatron/core/transformer/transformer_config.py
maanug-nv: The default of `--num-layers` argument is 'None', while `TransformerConfig` default is 0. Keeping argparse defaults are critical to ensure functional tests and especially user scripts do not break. So I would have liked to update `TransformerConfig` defaults to match argparse. However, doing so breaks many unit tests which directly initialize `TransformerConfig` without using argparse, using the dataclass defaults.
Two sets of defaults is a problem that needs to be addressed, but that's outside the scope of this PR.

megatron/core/transformer/transformer_config.py
maanug-nv: To match argparse defaults. 
I mentioned unit test issues in another comment, but looks like this change didn't break any, so I changed the dataclass default to match.
Ignoring unit tests, since most users of MLM use argparse interface, I think those defaults are the ones that really matter. `TransformerConfig` will get initialized with the argparse defaults during a real training from this [function](https://github.com/maanug-nv/Megatron-LM/blob/maanug/transfcfg-generated-args-v2/megatron/training/arguments.py#L1330)

megatron/training/arguments.py
maanug-nv: All of these are attributes of TransformerConfig for which either there is no corresponding argparse argument, or it it's too difficult to generate one right now (causes some other kind of issue). So I am excluding them from auto-generation.

I think these are not common cases and addressing them is not a high priority, there is a lot more work to do to upstream improvements from Megatron Bridge and unify the training loop. So I'd rather not block on this. This doesn't create any breaking changes either (at least that's my intent).

- cannot provide callables over CLI
There is no existing argparse arg for these. I'm not sure how a user would provide something for these over CLI, they are all `typing.Callable`.

- types affect docstring
These have some complicated type, either in the dataclass or in arguments or both. Some are Unions of several types in the dataclass which the factory does not support currently. Some have a function/lambda for the type in argparse. For most of these, since the type differs, the docstring also differs.

- no CLI argument exists for these
For all of these, there is no exactly corresponding argparse arg. e.g `enable_autocast` is not touched by megatron/training anywhere. 
Some of these are controlled by a completely different flag, e.g `params_dtype` is affected by '--fp16' and '--bf16'. But I don't see a simple way to generate this behavior.

- deprecated and no CLI arg exists
No need to generate CLI args for these.

- already generated by another config
I believe these two are generated by `RNGConfig`. 

- incompatible defaults in dataclass
These are all bool flags where the default in dataclass is the opposite of the default in args. Trying to match the defaults breaks several tests. And for these bool flags, using the 'argparse_meta' override method would require overriding almost all the kwargs which seems no different than excluding from generation and keeping the argument in arguments.py. Real solution is to fix the mismatching defaults and corresponding tests.

Wohox: I agree we can keep it this way for now, thanks~

lmcafee-nvidia (approved):

jaredcasper (approved):

PR #2920, opened 2026-01-13
Optimizing post-processing of requests

sidsingh-nvidia:
# What does this PR do ?
1. Optimizes serialization logic for finished requests
2. Overlaps detokenization with the engine step.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

3. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
@@ -1105,7 +1105,7 @@ class MegatronFSDP(torch.nn.Module):
         """
         self.param_and_grad_buffer.update_main_grads()
 
-    def finish_grad_sync(self):
+    def finish_grad_sync(self, force_all_reduce: Optional[bool] = False):
         """
         Finishes grad sync (all-reduce or reduce-scatter) communication operations
         for all model gradients. Call prior to the optimization step to resolve
@@ -1114,6 +1114,9 @@ class MegatronFSDP(torch.nn.Module):
         When overlap_grad_reduce is set to True, waits for asynchronous communication
         calls to complete. When overlap_grad_reduce is set to False, calls synchronous
         communication ops.
+
+        NOTE: force_all_reduce is included as an argument to maintain API compatibility
+        with DDP.force_grad_sync.
         """
         # Synchronize gradient reduce-scatter operations for all model gradients.
         self.synchronize_gradient_reduce()


megatron/core/inference/inference_request.py
```diff
@@ -294,10 +305,15 @@ def deserialize(cls, obj: dict) -> "DynamicInferenceRequest":
         Returns:
             (DynamicInferenceRequest) Deserialized request.
         """
-        request = super().deserialize(obj)
-        request.events = [DynamicInferenceEvent.deserialize(e) for e in obj["events"]]
+        request = cls(**obj)
+        request._post_deserialize(obj)
+
```
tdene: The `deserialize` method is the same in the child class as it is in the parent class: there's no need to redefine it. `classmethod`s are inherited by the child class fine, so the entire repeated method can be removed.

megatron/core/inference/inference_request.py
```diff
@@ -99,17 +99,21 @@ def serialize(self) -> dict:
             (dict) A dictionary representation of the instance suitable for
                 serialization.
         """
-
         # Dataclass to dict.
-        obj = asdict(self)
+        # do not use asdict(self) - it has very high CPU overheads
+        # and if there are tensors, it will try to deepcopy them
+        obj = self.__dict__  # shallow dict copy
```
tdene: This is not a shallow copy, it's the internal state dict that backs the object. You still need a `self.__dict__.copy()`.

Same thing with the other uses of `__dict__` in this PR.

megatron/core/inference/inference_request.py
```diff
@@ -280,8 +289,10 @@ def serialize(self):
             (dict) A dictionary representation of the instance suitable for
                 serialization.
         """
```
tdene, line 308: Can you add a `-> dict` return typehint on this method as well?

megatron/core/inference/inference_request.py
```diff
@@ -508,8 +524,10 @@ def serialize(self) -> dict:
...
-        obj = asdict(self)
-        obj["requests"] = [r.serialize() for r in self.requests]
+        torch.cuda.nvtx.range_push("DynamicInferenceRequestRecord.serialize")
+        self.requests = [r.serialize() for r in self.requests]
```
tdene: This destroys `self.requests` of the object itself. In every other class, `serialize` does not modify the underlying object; this should also be the case in this class.

Serialize them after you copy the `__dict__`?

megatron/core/inference/inference_request.py
```diff
@@ -99,17 +99,21 @@ def serialize(self) -> dict:
...
-
         # Dataclass to dict.
-        obj = asdict(self)
+        # do not use asdict(self) - it has very high CPU overheads
+        # and if there are tensors, it will try to deepcopy them
+        obj = self.__dict__  # shallow dict copy
         obj["status"] = self.status.name if self.status else None
+        obj["sampling_params"] = self.sampling_params.serialize() if self.sampling_params else None
+        obj["inference_parameters"] = (
+            self.inference_parameters.serialize() if self.inference_parameters else None
```
tdene, line 109: These don't get deserialized in `deserialize`, where `Status` gets deserialized.

megatron/core/inference/data_parallel_inference_coordinator.py
```diff
@@ -65,7 +65,7 @@ class DataParallelInferenceCoordinator:
         next_request_id (int): A counter for generating unique server-side request IDs.
     """
 
-    def __init__(self, inference_coordinator_port: int, data_parallel_size: int):
+    def __init__(self, inference_coordinator_port: int, data_parallel_size: int, tokenizer):
```
tdene, line 68: Is passing a tokenizer object into a newly-spawned process actually robust? The objects need to be pickle-able; I'm just surprised this would work.

I would expect that instead the tokenizer needs to be re-initialized inside the new process, but if it works that's good!

megatron/core/inference/inference_request.py
```diff
@@ -15,33 +14,34 @@
...
     """
-    buffer = io.BytesIO()
-    torch.save(tensor, buffer)
-    buffer.seek(0)
-    tensor_bytes = buffer.read()
-    return tensor_bytes
+    torch.cuda.nvtx.range_push("serialize_tensor")
 
+    # simply convert tensor into a list
+    tensor = tensor.cpu().tolist()
```
tdene, line 29: This loses dtype; float16 becomes float32 and int32 becomes int64.

I don't think this would be a problem today, but it could be in the future. Is this significantly faster than the old implementation?

Note that in the original implementation, `seek` and `read` are not necessary (you want something like that for higher efficiency when you run over multiple tensors in a for loop, but this method doesn't allow for doing that kind of code-golfing). You can instead just do
```
buffer = io.BytesIO()
torch.save(tensor, buffer)
tensor_bytes = buffer.getbuffer()
```
because ZMQ is fine with `memoryview` objects, and this saves you an extra read.

megatron/core/inference/inference_request.py
```diff
@@ -99,17 +99,21 @@ def serialize(self) -> dict:
             (dict) A dictionary representation of the instance suitable for
                 serialization.
         """
-
         # Dataclass to dict.
-        obj = asdict(self)
+        # do not use asdict(self) - it has very high CPU overheads
```
tdene, line 103: Good catch! Could you also fix the `asdict` on line 462 in this PR?

megatron/core/inference/inference_request.py
tdene:
```
from dataclasses import dataclass

@dataclass(kw_only=True)
class D:
    x: int

d = D(x=1)
attempted_copy = d.__dict__
attempted_copy["x"] = 10
attempted_copy["y"] = 11

assert d.x == 10; assert d.y == 11
```

megatron/core/inference/data_parallel_inference_coordinator.py
siddharth9820: This works!

megatron/core/inference/inference_request.py
siddharth9820: Good point.

megatron/core/inference/inference_request.py
siddharth9820: True, btw I'm not even sure why we have sampling parameters as well as inference parameters of the same class type

siddharth9820:
@Phlip79 we have functional tests in the suite for both values of the self.use_coordinator boolean.

megatron/core/inference/inference_request.py
siddharth9820: Good catch!

megatron/core/inference/inference_request.py
siddharth9820: This works because we serialize finished requests, so the destroyed object doesn't matter. But yes it's not clean code per se.

megatron/core/inference/inference_request.py
sidsingh-nvidia: Fixed

megatron/core/inference/inference_request.py
sidsingh-nvidia: Yeah I agree that this is not a shallow copy, and my comment is wrong. That being said I am wondering if we even need one? Returning this dict should be enough. Msgpack will create a serialized copy of it anyway. What do you think?

megatron/core/inference/inference_request.py
sidsingh-nvidia: Fixed!

megatron/core/inference/inference_request.py
sidsingh-nvidia: You are correct. 

However, I think the old implementation was wrong because it was losing dtype information anyway.

sidsingh-nvidia: Resolved offline - we checked that there is no information loss in bf16 -> fp32 -> bf16

megatron/core/inference/inference_request.py
sidsingh-nvidia: I switched to using .copy() - the performance penalty is very tiny and .copy() is safer

megatron/core/inference/inference_request.py
sidsingh-nvidia: Fixed

tdene (approved):

kvareddy (approved):

PR #2924, opened 2026-01-13
[Dev] Add E2E support for THD format

xiaoyao0115:
## Description

This PR adds **Sequence Packing (THD format)** E2E support to MCore. Main branch PR:https://github.com/NVIDIA/Megatron-LM/pull/3386

The core missing functionalities of THD in MCore are:
* data iterator cannot handle THD meta data, like cu_seqlens, max_seqlens.
* num_microbatches is fixed.
* PackParams are not passing between PP ranks.

#### 1. Add a data_iterator wrapper (megatron/core/datasets/data_schedule.py::wrap_dataloader)

A wrapper function that intercepts the data iterator to perform scheduling and packing:

- **Schedule & Pack**: Extracts data from the data iterator, schedules sequences across DP×CP ranks, and packs them into microbatches with `cu_seqlens` metadata.
- **Returns packing results**: Returns the packed `num_microbatches` along with two parameters for FLOPs calculation: `num_total_tokens_this_global_batch` and `sequence_square_sum_this_global_batch`.
- **TP broadcast**: Broadcasts `num_microbatches` and FLOPs parameters across TP ranks since only TP rank 0 has access to the data iterator.
- **PP broadcast**: When using PP, middle PP stages (not first or last) require metadata (`cu_seqlens`, `cu_seqlens_padded`, `max_seqlen`, etc.) to be broadcast from PP rank 0 for correct computation.

#### 2. Mock SFT Dataset Support

Supports mock datasets for testing and benchmarking with configurable sequence length distributions.
There are two modes of mock sft dataset:
- **File mode**: Load sequence lengths from an external CSV, example json:
  ```json
  {"mode": "file", "path": "/path/to/seqlens.csv"}
  ```
- **Distribution mode**: Generate sequence lengths from a distribution (currently supports lognormal), example json:
  ```json
  {"mode": "distribution", "type": "lognormal", "min_seq_len": 1024, "max_seq_len": 8192, "mean_seq_len": 4096, "lognormal_sigma": 1.1}
  ```

#### Before vs After

```mermaid
graph LR
    subgraph Before
        A1[DataIterator] --> B1[get_batch]
        B1 --> C1[forward_backward]
        C1 --> D1[Fixed seq_len FLOPs]
    end
    subgraph After
        A2[DataIterator] --> W[wrap_dataloader]
        W -->|schedule + pack| B2[PackedDataIterator]
        W -->|broadcast| M[num_microbatches + flops_params]
        B2 --> C2[get_batch_for_sequence_packing]
        C2 --> D2[forward_backward]
        D2 --> E2[Dynamic FLOPs]
        M   --> E2
    end
```

#### Execution Flow

```mermaid
sequenceDiagram
    participant Train as training.py
    participant Schedule as schedules.py
    participant Wrap as wrap_iterator_helper
    participant DataSched as data_schedule.py
    participant GetBatch as get_batch_for_seq_packing

    Train->>Schedule: forward_backward_*(data_iterator)
    Schedule->>Wrap: wrap_iterator_helper(config, data_iterator)
    Wrap->>DataSched: wrap_dataloader(data_iterator, scheduler_type)
    
    Note over DataSched: 1. Gather global seqlens across DP
    Note over DataSched: 2. Scheduler assigns sequences to microbatches
    Note over DataSched: 3. All-to-all redistribute samples
    Note over DataSched: 4. Pack into microbatches
    Note over DataSched: 5. Broadcast to TP/PP ranks
    
    DataSched-->>Schedule: (packed_iter, num_mbs, total_tokens, seq_sq_sum)
    
    loop for each microbatch
        Schedule->>GetBatch: get_batch_on_this_rank_for_sequence_packing
        Note over GetBatch: Broadcast tokens/labels to TP group
        Note over GetBatch: Partition for CP if needed
        GetBatch-->>Schedule: (tokens, labels, loss_mask, pos_ids, packed_seq_params)
    end
    
    Schedule-->>Train: forward_data_store + [total_tokens, seq_sq_sum]
```

### New Arguments

| Argument | Type | Description |
|----------|------|-------------|
| `--sequence-packing` | flag | Enable sequence packing (THD format) for training |
| `--sequence-packing-scheduler` | str | Scheduler type: `default` or `empty` |
| `--sft-mock-dataset-config-json` | str | JSON config for mock dataset |

### Changes

| File | Description |
|------|-------------|
| `megatron/core/datasets/data_schedule.py` | Core scheduling and packing logic |
| `megatron/core/pipeline_parallel/schedules.py` | Integration with forward/backward schedules |
| `megatron/training/training.py` | Updated FLOPs calculation for variable-length sequences |
| `megatron/training/datasets/sft_dataset.py` | Mock dataset support |
| `megatron/training/arguments.py` | New CLI arguments |
| `megatron/core/model_parallel_config.py` | Configuration options |
| `tests/unit_tests/test_sequence_packing.py` | Unit tests |

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/tokenizers/text/libraries/sft_tokenizer.py
+++ b/megatron/core/tokenizers/text/libraries/sft_tokenizer.py
@@ -108,7 +108,6 @@ class SFTTokenizer:
 
         self._prompt_format = prompt_format
 
-
     def tokenize_conversation(
         self, conversation: List[Dict], return_target: bool, add_generation_prompt: bool
     ):


megatron/core/model_parallel_config.py
```diff
@@ -72,6 +72,23 @@ class ModelParallelConfig:
...
+    sequence_packing_scheduler: Optional[str] = None
+    """
+    Scheduler for sequence packing and hybrid context parallel.
+    naive_sequence_packing: default naive sequence packing scheduler(just THD, no Hybrid-CP, this 
+    is just for comparison with default hybrid-cp scheduler, not recommended for production)
+    default_hybrid_cp: default hybrid-cp scheduler for hybrid context parallel provided by MCore.
+    empty_scheduler_with_packing: scheduling is already handled by the data sampler,
+    this scheduler only performs packing.
+    empty_scheduler_no_packing: scheduling and packing are already handled by the data sampler,
+    this scheduler only returns the batch.
```
lhb8125: The naming looks ambiguous and complicated. The first two could be like ```naive``` and ```dynamic_cp```, and do we really need the last two modes?
BTW, we renamed the hybrid cp as dynamic cp, right?

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -512,6 +513,30 @@ def check_first_val_step(first_val_step, forward_only, cond):
...
+    data_iterator: Union[Iterator, List[Iterator]],
+    num_microbatches: int,
+    pg_collection: Optional[ProcessGroupCollection] = None,
+):
+    """Warp data iterator for sequence packing if needed."""
+    if config.sequence_packing:
+        num_total_tokens_this_global_batch, sequence_square_sum_this_global_batch = None, None
+        scheduler_type_map = {
+            'default': PackingScheduler.DEFAULT,
+            'empty': PackingScheduler.EMPTY,
```
lhb8125: Does it align with the comments in ```model_parallel_config.py```?

megatron/core/datasets/data_schedule.py
```diff
@@ -295,7 +310,1049 @@ def __next__(self) -> Any:
...
+        dp_group = parallel_state.get_data_parallel_group()
+        tp_group = parallel_state.get_tensor_model_parallel_group()
+        pp_group = parallel_state.get_pipeline_model_parallel_group()
+    else:
+        dp_cp_group = pg_collection.dp_cp
+        dp_group = pg_collection.dp
+        tp_group = pg_collection.tp
+        pp_group = pg_collection.pp
+    assert (
+        dp_cp_group is not None and dp_group is not None and tp_group is not None
```
kunlunl: Missing pp_group

megatron/core/datasets/data_schedule.py
```diff
@@ -295,7 +310,1049 @@ def __next__(self) -> Any:
...
+            if config.virtual_pipeline_model_parallel_size is None
+            else config.microbatch_group_size_per_vp_stage
+        ),
+    )
+
+    (
+        new_data_iterator,
+        num_micro_batches,
+        num_total_tokens_this_global_batch,
+        sequence_square_sum_this_global_batch,
```
kunlunl: Use consistent name as "num_total_tokens_this_global_batch"?

megatron/core/datasets/data_schedule.py
```diff
@@ -295,7 +310,1049 @@ def __next__(self) -> Any:
...
+
+    def get_require_sample_keys(self):
+        """Return the required key of each batch."""
+        raise NotImplementedError
+
+    def get_groups_and_subsamples(self, sample_id_seqlens):
+        """schedule the samples into groups"""
+        raise NotImplementedError
+
+    def _get_global_seqlens(self, subsample_seqlens: torch.Tensor, dp_group) -> List[int]:
```
kunlunl: Add @staticmethod to all stateless function?

megatron/core/datasets/data_schedule.py
```diff
@@ -295,7 +310,1049 @@ def __next__(self) -> Any:
...
+        num_micro_batches,
+        num_total_tokens_this_global_batch,
+        sequence_square_sum_this_global_batch,
+        pp_group,
+        tp_group,
+        dev,
+    ):
+        """Broadcast scheduling info to middle PP stages."""
+
+        def _broadcast_to_pp(item):
```
kunlunl: Unify with _boradcast_to_tp()

megatron/core/datasets/data_schedule.py
```diff
@@ -295,7 +310,1049 @@ def __next__(self) -> Any:
...
+        global_ids_this_rank = global_ids[
+            offsets[dp_rank] : offsets[dp_rank] + num_local_subsamples
+        ]
+
+        return global_id_seqlens, global_ids_this_rank
+
+    def _broadcast_to_pp_group(
+        self,
+        new_samples,
+        num_micro_batches,
```
kunlunl: Make it more general, don't need to hardcode this three scalar.

megatron/core/model_parallel_config.py
xiaoyao0115: yeah, this has been removed. BTW, i will rename hybrid-cp to dynamic-cp in MR2000, not in this MR.

megatron/core/pipeline_parallel/schedules.py
xiaoyao0115: I've fixed it.

megatron/core/datasets/data_schedule.py
xiaoyao0115: what about `seqlen_sum_this_global_batch` and `seqlen_squared_sum_this_global_batch`?

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +314,1045 @@ def __next__(self) -> Any:
...
+    }
+
+    # Convert string to enum if needed
+    if isinstance(scheduler_type, str):
+        try:
+            scheduler_type = PackingScheduler[scheduler_type.upper()]
+        except KeyError:
+            available_scheduler = ", ".join([scheduler.name for scheduler in PackingScheduler])
+            raise ValueError(
+                f"Unknown packing scheduler: {scheduler_type}. "
```
xiaoyao0115: remove it to model_parallel_config post_init

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +314,1045 @@ def __next__(self) -> Any:
...
+
+
+class PackingScheduler(enum.Enum):
+    """Enum for supported sequence packing algorithms."""
+
+    EMPTY = "empty"
+    DEFAULT_SEQUENCE_PACKING = "default_sequence_packing"
+
+
+class BaseScheduler:
```
xiaoyao0115: missing def run()

megatron/training/datasets/sft_dataset.py
```diff
@@ -73,131 +73,254 @@ def build_low_level_dataset(dataset_path: str, config: GPTDatasetConfig) -> LowL
...
+    
+
+    def __init__(self, config: Dict) -> None:
+        np.random.seed(self.seed)
+        # either choose to load sequence lengths from external file, or generate random sequence lengths
+        
+        assert "mode" in config, f"mode must be set, either 'file' or 'distribution'"
+        
+        if config["mode"] == "file":
+            self.sequence_lengths = np.array(pd.read_csv(config["path"])).flatten()
```
xiaoyao0115: add default config

megatron/core/model_parallel_config.py
```diff
@@ -72,6 +72,17 @@ class ModelParallelConfig:
     Please set max_seqlen_per_dp_cp_rank when using hybrid_context_parallel.
     """
 
+    sequence_packing_scheduler: Optional[str] = None
```
ISEEKYAN: just a suggestion, can we open an API for the users to register their own schedulers and choose it by this option?

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +314,969 @@ def __next__(self) -> Any:
...
+    """Base class for sequence packing schedulers."""
+
+    def __init__(
+        self,
+        max_seqlen_per_dp_cp_rank: int,
+        cp_size: int,
+        dp_size: int,
+        microbatch_group_size_per_vp_stage: Optional[int],
+    ):
```
lhb8125, line 324: Better add comments for the inputs, the same below.

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +314,969 @@ def __next__(self) -> Any:
...
+                            seqlen_squared_sum_this_global_batch,
+                        ],
+                        dtype=torch.float32,
+                    ).cuda()
+                ]
+                for sample in new_samples:
+                    tensor_list.append(sample["max_seqlen"].unsqueeze(0))
+                for sample in new_samples:
+                    tensor_list.append(sample["cu_seqlens"])
+                    tensor_list.append(sample["cu_seqlens_padded"])
```
lhb8125: Why is separating ```max_seqlen``` and ```cu_seqlens/cu_seqlens_padded``` intentionally?

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +314,969 @@ def __next__(self) -> Any:
...
+    Args:
+        data_iterator: The original data_iterator to wrap around
+        config: The config object containing the max_seqlen_per_dp_cp_rank
+        dp_cp_group: Data parallel context parallel group.
+        pg_collection: The process group collection.
+    """
+
+    scheduler_map: Dict[PackingScheduler, Type[BaseScheduler]] = {
+        PackingScheduler.DEFAULT_SEQUENCE_PACKING: DefaultSequencePackingScheduler
+    }
```
lhb8125: Why do we need a ```scheduler_map``` here, considering we already have ```PackingScheduler```?

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +314,969 @@ def __next__(self) -> Any:
...
+            assert tex is not None and is_te_min_version("1.10.0"), (
+                "Please update Transformer Engine to >= 1.10 to use "
+                "Context Parallel with THD format data"
+            )
+            total_tokens = batch['tokens'].size(0)
+            # Transformer Engine has a bug of cu_seqlens, we must treat cu_seqlens_padded as
+            # cu_seqlens to get the correct result.
+            # TODO: Revert this workaround once TE fixes the issue.
+            cu_seqlens = batch["cu_seqlens_padded"]
+            index = tex.thd_get_partitioned_indices(cu_seqlens, total_tokens, cp_size, cp_rank)
```
lhb8125: Please move the TE-related calls to ```transformer_engine.py```

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +314,969 @@ def __next__(self) -> Any:
...
+    Args:
+        data_iterator (Iterator): The data iterator to get the batch from.
+        mtp_on_this_rank (bool): Whether to use multi-token prediction.
+        vp_stage (Optional[int]): The stage of the pipeline.
+    Returns:
+        tuple of (tokens, labels, loss_mask, attention_mask, position_ids, packed_seq_params)
+    """
+
+    tp_src_rank = parallel_state.get_tensor_model_parallel_src_rank()
+    tp_group = parallel_state.get_tensor_model_parallel_group()
```
lhb8125: I think getting TP/PP rank/size from ```parallel_state``` is not recommended now, try ```pg_collection``` instead.

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +314,969 @@ def __next__(self) -> Any:
...
+            batch['max_seqlen'] = torch.tensor(batch['max_seqlen'], dtype=torch.int32, device=dev)
+        else:
+            assert batch['max_seqlen'].dtype == torch.int32
+            assert batch['max_seqlen'].numel() == 1
+    else:
+        batch['cu_seqlens'] = torch.empty([cu_seqlen_size], dtype=torch.int32, device=dev)
+        batch['cu_seqlens_padded'] = torch.empty([cu_seqlen_size], dtype=torch.int32, device=dev)
+        batch['max_seqlen'] = torch.empty(1, dtype=torch.int32, device=dev)
+
+    # Broadcast batch inside TP group.
```
lhb8125, line 822: Why don't we broadcast batch inside PP group?

megatron/core/transformer/transformer_config.py
```diff
@@ -1982,6 +1982,38 @@ def __post_init__(self):
...
+                raise ValueError(
+                    "SFT sequence packing requires Transformer Engine >= 2.9.0 "
+                    f"but got {get_te_version()} (TE < 2.9.0 may have convergence issues)."
+                )
+
+            # Needed for passing variable sequences between pp stages.
+            self.variable_seq_lengths = True
+
+            # TODO(tailaim): add support for other dispatcher types
+            warnings.warn("Setting moe_token_dispatcher_type to alltoall for sft sequence packing.")
```
lhb8125: Warning could be ignored. Throwing an error for other dispatcher_type is better.

megatron/core/transformer/transformer_config.py
```diff
@@ -1982,6 +1982,38 @@ def __post_init__(self):
...
+            self.moe_token_dispatcher_type = "alltoall"
+
+            if self.sequence_packing_scheduler is None:
+                self.sequence_packing_scheduler = 'default_sequence_packing'
```
lhb8125: 'default' is enough since the kbob's name already indicates the "sequence_packing"?

megatron/training/arguments.py
```diff
@@ -3629,4 +3649,8 @@ def _add_sft_args(parser):
     group.add_argument('--sft', action="store_true", help='Megatron SFT training')
     group.add_argument('--sft-tokenizer-prompt-format', type=str, default="nemotron-h-aligned",
                        help='SFT prompt format.')
+    group.add_argument('--sequence-packing', action='store_true',
+                       help='use sequence packing(thd format) for training')
+    group.add_argument('--sft-mock-dataset-config-json', type=str, default=None, 
+                       help='This config provides the necessary information for the mock dataset. You can either specify a CSV file that contains sequence lengths, where each line stores the length of a sequence, for example: {"mode":"file","path":"/path/to/file"}. Alternatively, you can specify a distribution (currently only supporting lognormal distribution) along with the required parameters, for example, {"mode":"distribution","type":"lognormal","min_seq_len":1024,"max_seq_len":2048,"mean_seq_len":1536,"lognormal_sigma":1.1}, where sigma controls the variability of the lognormal distribution.')
```
lhb8125: Please let the user know the default option.

tests/unit_tests/test_sequence_packing.py
```diff
@@ -0,0 +1,322 @@
...
+        (2, 2, 1, False),  # TP + PP
+        (1, 1, 2, False),  # CP only
+        (2, 1, 2, False),  # TP + CP
+        (1, 2, 2, False),  # PP + CP
+        (1, 4, 1, False),  # Has middle pp stage
+        (1, 1, 1, True),  # Hybrid CP enabled (CP=1 with hybrid groups)
+        (2, 1, 1, True),  # TP + Hybrid CP
+    ],
+)
+def test_get_batch_on_this_rank_for_sequence_packing(tp, pp, cp, hybrid_cp):
```
lhb8125: Besides testing this function, I suggest testing the correctness of e2e training and the compatibility of other features, which could also be done in ut in addition to ft.

megatron/core/model_parallel_config.py
xiaoyao0115: yeah, we can do it in the next pr~

megatron/core/datasets/gpt_dataset.py
```diff
@@ -79,6 +79,14 @@ class GPTDatasetConfig(BlendedMegatronDatasetConfig):
     context_parallel_size: Optional[int] = None
     """The size of the context parallel group. Needed for padding in packed sequences."""
 
+    sft_mock_dataset_config_json: Optional[str] = None
+    """This config provides the necessary information for the mock dataset."""
```
jkamalu, line 83: Anti-pattern? Optional dataset config within a dataset config? Why aren't we using inheritance here?

lhb8125:
/ok to test [b223308](https://github.com/NVIDIA/Megatron-LM/pull/2924/commits/b2233087cb1a8aa78b689a219faf2c0e29a9f1da)

megatron/core/datasets/data_scheduler_utils.py
```diff
@@ -0,0 +1,492 @@
...
+    batch = []
+    for item in batch_list:
+        if isinstance(item, dict):
+            batch.append(item)
+        elif isinstance(item, list):
+            batch.extend(item)
+        else:
+            raise ValueError(f"Invalid item type: {type(item)}")
+
+    # This unpack step is redundant: in sft_dataset.py, sequences are already packed before
```
lhb8125: Why it is redundant? The following comments say it's essential in mcore's sft_dataset.

megatron/training/arguments.py
```diff
@@ -2932,4 +2962,10 @@ def _add_sft_args(parser):
     group.add_argument('--sft', action="store_true", help='Megatron SFT training')
     group.add_argument('--sft-tokenizer-prompt-format', type=str, default="nemotron-h-aligned",
                        help='SFT prompt format.')
+    group.add_argument('--sequence-packing', action='store_true',
```
jkamalu: So, we already do sequence packing in megatron core. I think this term is being overloaded here.

For example, we already refer to our samples as packed sequences, so --sequence-packing is confusing? I think this MR should explicitly reference what is actually being supported not just here but all over i.e. thd qkv_format with transformer engine using sample accurate cu_seqlens and max_seqlen info that will actually inhibit cross document attention.

jkamalu: This makes the whole MR hard to interpret for me, actually. E.g. BaseScheduler being referred to as the base class for "sequence packing schedulers". BaseScheduler sounds indifferent to THD. So does sequence packing. But I get the sense that sequence packing is meant to be read as THD specific. Is the base scheduler?

megatron/training/arguments.py
```diff
@@ -921,6 +914,32 @@ def validate_args(args, defaults={}):
...
+            args.sft_mock_dataset_config_json = json.dumps(
+                {
+                    "mode": "distribution",
+                    "type": "lognormal",
+                    "min_seq_len": args.seq_length // 2,
+                    "max_seq_len": args.seq_length,
+                    "mean_seq_len": args.seq_length // 4 * 3,
+                    "lognormal_sigma": 1.1,
+                }
+            )
```
lhb8125: How about moving it to transformer_config.py

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,564 @@ def __next__(self) -> Any:
             batch, global_ids_this_rank, global_id_seqlens, sample_id_groups, offsets
         )
         return samples_this_rank_with_id, sample_id_groups
+
+
+class BaseScheduler:
```
jkamalu: Can we have a more verbose name for this base class?

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,564 @@ def __next__(self) -> Any:
...
+
+
+class DpBalancedScheduler(BaseScheduler):
+    """Packs sequences in their original order until reaching the max limit of sequence length."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.max_seq_len_all_ranks = self.max_seqlen_per_dp_cp_rank * self.cp_size
+
+    def get_require_sample_keys(self):
```
jkamalu: nit: required

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,564 @@ def __next__(self) -> Any:
...
+            new_data_iterator: The new data iterator (or list for VPP).
+            num_micro_batches: Number of micro batches after scheduling.
+            seqlen_sum_this_global_batch: Total tokens for FLOPs calculation.
+            seqlen_squared_sum_this_global_batch: Sum of squared seqlens for FLOPs.
+        """
+        raise NotImplementedError
+
+
+class DpBalancedScheduler(BaseScheduler):
+    """Packs sequences in their original order until reaching the max limit of sequence length."""
```
jkamalu, line 380: Can you be more specific about whether we're truncating or padding here?

In SFT land, I assume we'd be padding. In pre-training land, we're likely truncating. Is this specific to one or the other? If so, can this be made clearer in the scheduler name?

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,564 @@ def __next__(self) -> Any:
...
+
+        return (
+            new_data_iterator,
+            num_micro_batches,
+            seqlen_sum_this_global_batch,
+            seqlen_squared_sum_this_global_batch,
+        )
+
+
+class PackingScheduler(enum.Enum):
```
jkamalu: Name should change. DpBalancedScheduler is a BaseScheduler but PackingScheduler is an Enum?

Maybe PackingSchedulerEnum or PackingSchedulerMethod or PackingSchedulerAlgorithm. 

And on that note, Maybe BaseScheduler -> BasePackingScheduler.

^^all of these suggestions are still made with it being said that "packing" is an overloaded term

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,564 @@ def __next__(self) -> Any:
...
+    return (
+        new_data_iterator,
+        num_micro_batches,
+        seqlen_sum_this_global_batch,
+        seqlen_squared_sum_this_global_batch,
+    )
+
+
+def get_batch_on_this_rank_for_sequence_packing(
+    data_iterator,
```
jkamalu, line 695: nit: typing.

We also need a way of knowing what a well implemented data iterator is. I see below a comment which explains the iterator should returns a dict containing certain keys.

So that would suggest this scheduler is only made for SFT datasets which contain this information? This should be made apparent somehow. For example, if I'm working on pre-training, I might mistakenly try to use this function when I'm working with GPTDatasets.

megatron/training/arguments.py
jkamalu: Also, how is this scoped? Can I do sequence packing for BERT? For non-SFT GPT pretraining? How do I know when this option will do anything?

megatron/core/datasets/data_scheduler_utils.py
```diff
@@ -0,0 +1,492 @@
...
+
+from typing import Dict, List
+
+import numpy as np
+import torch
+
+from megatron.core.rerun_state_machine import RerunDataIterator
+
+
+def _unpack_batch(batch):
```
jkamalu: nit: typing

megatron/core/datasets/data_schedule.py
jkamalu, line 1: These additions need a comprehensive section in the megatron/core/datasets/ read me

megatron/core/transformer/transformer_config.py
```diff
@@ -2040,6 +2040,40 @@ def __post_init__(self):
...
+
+            supported_schedulers = ['dp_balanced']
+            if (
+                self.sequence_packing_scheduler is not None
+                and self.sequence_packing_scheduler not in supported_schedulers
+            ):
+                raise ValueError(
+                    f"Unknown scheduler: {self.sequence_packing_scheduler}. "
+                    f"Available schedulers: {supported_schedulers}"
+                )
```
jkamalu, line 2111: Why do we need to keep track of the supported schedulers here? Can we make it a part of the data class? 

nit: and in the error message, it's more that the scheduler isn't supported as opposed to being unknown

megatron/training/datasets/sft_dataset.py
```diff
@@ -201,3 +224,155 @@ def extend_with_padding(tokens, targets, positions, pad_len):
...
+    seed: int = 0
+    """The hard-coded random seed to use to set the NumPy RNG"""
+
+    size: int = 1000000
+    """The hard-coded number of sequence to generate"""
+    
+    # This is to maintain consistency with the SFT dataset that uses real data. In the real dataset, an element in the low-level dataset often contains multiple sequences. So here, each element in the mock low-level dataset also contains num_sequence_per_sample sequences. This will be made more reasonable in the future.
+    
+
+    def __init__(self, config: Dict) -> None:
```
jkamalu: Why not args or kwargs with default values?

Difficult to read + know what config should have in it.

megatron/training/datasets/sft_dataset.py
```diff
@@ -201,3 +224,155 @@ def extend_with_padding(tokens, targets, positions, pad_len):
...
+
+    def __len__(self) -> int:
+        return self.size
+
+    def __getitem__(self, idx: int) -> List[np.ndarray]:
+        length = self.sequence_lengths[idx % self.size]
+        # the length of sample is 'length', but only length-1 elements are generated here, 
+        # because an eod token will be appended at the end later in SFTDataset
+        sample = np.arange(1, length, dtype=np.int64)
+        return sample
```
jkamalu: nit: newline

megatron/training/arguments.py
xiaoyao0115: Apologies — --sequence-packing was supposed to be removed. I missed deleting it in this MR. Will clean it up.

megatron/core/datasets/data_schedule.py
xiaoyao0115: ok, I'll add it

megatron/training/arguments.py
xiaoyao0115: `sft_mock_dataset_config_json` is used to construct `GPTDatasetConfig` in `gpt_dataset.py`. We could move it to `post_init` in `GPTDatasetConfig` but another problem is that variables `mock_data` are not defined there, we need to add `mock_data` first.

megatron/core/datasets/data_schedule.py
xiaoyao0115: No truncation or padding is performed here.
The responsibility of this component is to pack sequences, broadcast the required runtime information (metadata, num_microbatches, and FLOPs-related data), and reformat data to the required format for advanced features.

megatron/training/arguments.py
xiaoyao0115: > So, we already do sequence packing in megatron core. 

Although sft_dataset does return packed sequences, that alone is not sufficient to enable THD to run end-to-end. There are several fundamental issues:
1. Metadata must be broadcast across PP ranks.
Metadata such as `cu_seqlens` and `max_seqlen` are not available on intermediate PP stages. Running THD requires these values to be visible across all PP ranks, which means we must explicitly broadcast them.
2. FLOPs accounting must change under THD.
For a packed sequence, FLOPs should be computed per (unpacked) sequence and then aggregated. The current implementation does not reflect this behavior. This is why `data_iterator_wrapper` returns two additional variables related to FLOPs calculation.
3. Sometimes `num_microbatches` is not fixed, especially for RL and multi-modal.
MCore currently assumes a fixed num_microbatches. However, in scenarios such as RL training, the number of microbatches per iteration may vary depending on sample lengths. Therefore, we should make sure Mcore can run correctly with dynamic `num_microbatches`.

These concerns cannot be cleanly addressed inside `sft_dataset`. They involve cross-PP-rank coordination, runtime scheduling decisions, and global training-state management (e.g., microbatch configuration and FLOPs accounting). 

For this reason, we think it's a good way to resolve this issues in `data_iterator_wrapper`.

megatron/core/datasets/data_schedule.py
xiaoyao0115: Good idea, i will rename them: BaseScheduler -> BasePackingScheduler and PackingScheduler -> PackingSchedulerEnum

megatron/training/datasets/sft_dataset.py
xiaoyao0115: You are right, i'll fix it

megatron/core/datasets/data_schedule.py
xiaoyao0115: We only reach this code path when the user sets `--sequence-packing-scheduler`. In pre-training the flag is `None` by default, so there is no risk of accidentally using this with GPTDataset iterators.

megatron/core/transformer/transformer_config.py
xiaoyao0115: ok, i'll make the changes.

megatron/core/datasets/gpt_dataset.py
xiaoyao0115: The mock/SFT dataset config has a large number of parameters, and threading them one-by-one into the  `gpt_dataset` would be noisy.

megatron/training/arguments.py
xiaoyao0115: In addition, several upcoming features will require similar structural handling. For example, Hybird-CP requires rescheduling sequences, selecting different CP sizes dynamically, and redistributing data across ranks. In this process, `num_microbatches` also become dynamic.
Similarly, THD + CUDA Graph requires further manipulation of both sequences and their metadata.
This MR can be viewed as Step 1 toward enabling those features. In the future, these features will corresponding to different scheduler implementations.

megatron/core/datasets/data_scheduler_utils.py
xiaoyao0115: i've make the change

pretrain_gpt.py
```diff
@@ -255,6 +266,8 @@ def core_gpt_dataset_config_from_args(args):
         "data_parallel_size": args.data_parallel_size,
         "sequence_parallel_size": args.tensor_model_parallel_size*args.sequence_parallel,
         "hybrid_context_parallel": args.hybrid_context_parallel,
+        "sft_mock_dataset_config_json":args.sft_mock_dataset_config_json,
+        "sequence_packing_scheduler": args.sequence_packing_scheduler,
```
kunlunl: This attribute is not used. Please delete it.

megatron/training/training.py
```diff
@@ -137,7 +137,6 @@ def set_startup_timestamps(program_start=None, main_entry=None):
 from megatron.training.initialize import set_jit_fusion_options
 from megatron.training.utils import get_batch_on_this_cp_rank, get_batch_on_this_tp_rank
 from megatron.training.datasets.data_samplers import build_pretraining_data_loader
-from megatron.core.datasets.data_schedule import HybridCPDataLoaderWrapper
```
kunlunl, line 142: This shouldn't be removed.

megatron/training/training.py
```diff
@@ -1661,14 +1664,35 @@ def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_sch
...
+            (
+                data_iterator,
+                num_microbatches,
+                seqlen_sum_this_global_batch,
+                seqlen_squared_sum_this_global_batch,
+            ) = wrap_data_iterator(data_iterator, config, get_num_microbatches())
+        else:
+            # data_iterator unchanged
+            num_microbatches = get_num_microbatches()
+            seqlen_sum_this_global_batch = args.seq_length * args.micro_batch_size * args.data_parallel_size * num_microbatches
```
kunlunl: `args.micro_batch_size * args.data_parallel_size * num_microbatches` is equal to `args.global_batch_size`?

megatron/training/training.py
```diff
@@ -1699,9 +1723,10 @@ def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_sch
         # iteration is 0-indexed, move to 1-indexed for checkpoint name and logging.
         save_grads(args.save, state_dict, iteration + 1, "wgrads")
 
+
```
kunlunl: Delete this line.

megatron/training/training.py
```diff
@@ -1699,9 +1723,10 @@ def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_sch
...
     should_checkpoint, should_exit, exit_code = rerun_state_machine.should_checkpoint_and_exit()
     if should_exit:
-        return {}, True, should_checkpoint, should_exit, exit_code, None, None, 0
+        return {}, True, should_checkpoint, should_exit, exit_code, None, None, seqlen_sum_this_global_batch, seqlen_squared_sum_this_global_batch, 0
```
kunlunl: Wrong position of `seqlen_sum_this_global_batch` and `seqlen_squared_sum_this_global_batch`, should be after `0`?

megatron/training/arguments.py
```diff
@@ -921,6 +914,29 @@ def validate_args(args, defaults={}):
...
+    # during pipeline parallelism, it should not be set if sequence length
+    # is constant during training.
+    args.variable_seq_lengths = False
+    if args.sequence_packing_scheduler is not None:
+        args.variable_seq_lengths = True
+        assert args.context_parallel_size * args.max_seqlen_per_dp_cp_rank >= args.seq_length, \
+            f'Packed sequence buffer size ({args.context_parallel_size * args.max_seqlen_per_dp_cp_rank}) ' \
+            f'must be >= single sequence max length ({args.seq_length})'
+        if args.mock_data and args.sft_mock_dataset_config_json is None:
+            args.sft_mock_dataset_config_json = json.dumps(
```
kunlunl: Maybe this default value should be moved into MockSFTDataset?

megatron/training/arguments.py
```diff
@@ -1640,6 +1656,9 @@ def _add_network_size_args(parser):
         "persist_layer_norm",
         "bias_dropout_fusion",
         "apply_rope_fusion",
+        "max_seqlen_per_dp_cp_rank",
+        "hybrid_context_parallel",
```
kunlunl: Why need this change?

megatron/training/arguments.py
```diff
@@ -2377,6 +2396,14 @@ def _add_distributed_args(parser):
...
                        'specify separated types for each layer like '
                        '--cp-comm-type p2p p2p a2a a2a a2a+p2p a2a+p2p')
+    group.add_argument('--max-seqlen-per-dp-cp-rank', type=int, default=None,
+                       help='Maximum sequence length per CP rank. This is used to calculate the '
+                       'number of sub-samples assigned to each CP rank when using heterogeneous context parallel.')
+    group.add_argument('--hybrid-context-parallel', action='store_true', default=False,
+                       help='Enables hybrid context parallel. This is used to balance the workload '
+                       'of each CP rank when we use packed samples with variable sequence lengths. '
+                       'Requires --max-seqlen-per-dp-cp-rank to be set.')
+    group.add_argument('--sequence-packing-scheduler', type=str, default=None, choices=['dp_balanced'])
```
kunlunl: Should we directly add them in TransformerConfig

megatron/training/tokenizer/tokenizer.py
```diff
@@ -870,7 +870,10 @@ def eos(self):
     @property
     def additional_special_tokens_ids(self):
         return None
-
+    
+    @property
+    def pad(self):
+        return self._eod_id - 1
```
kunlunl: Why use eod_id - 1?

megatron/training/training.py
kunlunl, line 3187: Also need to add data loader wrapper for evaluate function?

megatron/training/datasets/sft_dataset.py
```diff
@@ -47,7 +50,6 @@ def __len__(self) -> int:
     def __getitem__(self, idx: int) -> list:
         return self.dataset[idx]["messages"]
 
-
```
kunlunl, line 50: Don't delete this line.

megatron/training/datasets/sft_dataset.py
```diff
@@ -61,6 +63,8 @@ def __init__(
         config: GPTDatasetConfig,
     ) -> None:
         super().__init__(dataset, dataset_path, indices, num_samples, index_split, config)
+        # Pre-calculate padding divisor to avoid redundant computation in get_item
+        self.padding_divisor = self._calculate_padding_divisor()
```
kunlunl: Don't need to cache this attribute.

megatron/training/datasets/sft_dataset.py
```diff
@@ -201,3 +224,158 @@ def extend_with_padding(tokens, targets, positions, pad_len):
             'cu_seqlens': cu_seqlens,
             'max_seqlen': max_seqlen,
         }
+
```
kunlunl, line 216: Missing one empty line.

megatron/training/datasets/sft_dataset.py
```diff
@@ -201,3 +224,158 @@ def extend_with_padding(tokens, targets, positions, pad_len):
...
+        max_seqlen = adjacent_diffs.max()  # max_seqlen is a 0-D tensor
+
+        return {
+            'tokens': input_ids,
+            'labels': labels,
+            'loss_mask': loss_mask,
+            'position_ids': position_ids,
+            'cu_seqlens': cu_seqlens,
+            'max_seqlen': max_seqlen,
+        }
```
kunlunl: Missing empty line at the end of file.

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -552,6 +552,9 @@ def forward_backward_no_pipelining(
         pg_collection.dp_cp = parallel_state.get_data_parallel_group(
             with_context_parallel=True, partial_data_parallel=False
         )
+        pg_collection.dp = parallel_state.get_data_parallel_group(
```
kunlunl: Don't need this line?

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -1009,6 +1012,9 @@ def forward_backward_pipelining_with_interleaving(
...
```
kunlunl: Don't need this line?

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -2158,6 +2164,9 @@ def forward_backward_pipelining_without_interleaving(
...
```
kunlunl: Don't need this line?

megatron/core/datasets/gpt_dataset.py
```diff
@@ -79,6 +79,14 @@ class GPTDatasetConfig(BlendedMegatronDatasetConfig):
...
+
+    sequence_packing_scheduler: Optional[str] = None
```
kunlunl: This attribute is not used?

megatron/core/datasets/data_schedule.py
```diff
@@ -57,7 +69,6 @@ def get_global_seqlens(self, subsample_seqlens: torch.Tensor) -> List[int]:
         Gathers the sequence lengths of all subsamples from all DP ranks.
         Each DP rank loads the same number of microbatches but each microbatch
         may have a different number of subsamples.
-
```
kunlunl, line 60: Dont delete this line.

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,564 @@ def __next__(self) -> Any:
...
+
+        return (
+            new_data_iterator,
+            num_micro_batches,
+            seqlen_sum_this_global_batch,
+            seqlen_squared_sum_this_global_batch,
+        )
+
+
+class PackingSchedulerEnum(enum.Enum):
```
kunlunl: Maybe don't need this abstraction?

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,564 @@ def __next__(self) -> Any:
...
+    """Enum for supported sequence packing algorithms."""
+
+    DP_BALANCED = "dp_balanced"
+
+
+scheduler_map: Dict[PackingSchedulerEnum, Type[BasePackingScheduler]] = {
+    PackingSchedulerEnum.DP_BALANCED: DpBalancedScheduler
```
kunlunl: Maybe just `"dp_balanced": DpBalancedScheduler`

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,555 @@ def __next__(self) -> Any:
...
+                data_iterator = data_iterator[0]
+            elif pp_group.rank() == pp_group.size() - 1:
+                # the last stage
+                data_iterator = data_iterator[-1]
+            else:
+                data_iterator = None
+
+        # data_iterator is not None when TP rank 0, with PP stage 0 or -1.
+        if data_iterator is not None:
+            assert tp_group.rank() == 0 and (
```
parthmannan: @xiaoyao0115 One concern I have come to realize is the assumption that only TP rank 0 should have the data iterator and it always broadcasts information to other TP ranks.
This is true in Megatron-LM but does not hold true for Nemo-RL for example. In Nemo-RL, the head node distributes data to all the workers and each rank has the data_iterator.

It might be difficult to identify which regime we are in but maybe we need to support a mode where we don't perform broadcasts between TP ranks etc. This applies to other places in the code where we make this assumption as well.

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,555 @@ def __next__(self) -> Any:
...
+
+    return (
+        new_data_iterator,
+        num_micro_batches,
+        seqlen_sum_this_global_batch,
+        seqlen_squared_sum_this_global_batch,
+    )
+
+
+def get_batch_on_this_rank_for_sequence_packing(
```
parthmannan, line 694: Similar thing here. We should keep data sharding (i.e. get_thd_partitioned_indices etc.) separate from broadcasting logic instead of combining into 1 large function.

For example, in [`pretrain_gpt.py`](https://github.com/NVIDIA/Megatron-LM/blob/main/pretrain_gpt.py#L64), we call `get_batch_on_this_tp_rank` if needed and then we can use `get_thd_batch_on_this_cp_rank` for THD format or `get_batch_on_this_cp_rank` for SBHD format.

Calling it out here in case you have plans to remove those independent functions in favor of this single utility.

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,555 @@ def __next__(self) -> Any:
...
+    # TODO: Revert this workaround once TE fixes the issue.
+    packed_seq_params = PackedSeqParams(
+        qkv_format="thd",
+        cu_seqlens_q=cu_seqlens_padded,
+        cu_seqlens_kv=cu_seqlens_padded,
+        cu_seqlens_q_padded=cu_seqlens_padded,
+        cu_seqlens_kv_padded=cu_seqlens_padded,
+        max_seqlen_q=max_seqlen,
+        max_seqlen_kv=max_seqlen,
+        local_cp_size=None,
```
parthmannan, line 851: Where are we setting `local_cp_size` and `cp_group` for Dynamic CP? Is that expected to go into a follow-up PR?

xiaoyao0115: Yeah, we will set `local_cp_size` and `cp_group` in the next PR for supporting Dynamic-CP

megatron/core/datasets/data_schedule.py
xiaoyao0115: Yeah — different frameworks make different assumptions about `data_iterator`. Megatron-LM typically assumes only TP rank 0 owns the iterator and broadcasts to the other TP ranks, while in frameworks like NeMo-RL and verl are different.
For this change, I think it’s best to keep Megatron-LM’s behavior, rather than trying to cover every external data-loading regime. Supporting “no TP broadcast” is very framework-specific and is better implemented in the integration layer (e.g., NeMo-RL / verl), by adapting their dataloader to produce the same final inputs that Megatron expects.

megatron/core/datasets/data_schedule.py
xiaoyao0115: Folding THD into `get_batch_on_this_tp_rank` would add a lot of format-specific branching since THD carries extra keys like `max_seqlen`, `cu_seqlens`, etc.
Also consistent with my earlier point: other frameworks (NeMo-RL / verl) should implement their own data processing to match Megatron’s expected inputs rather than having Megatron cover every data-iterator regime.
For CP sharding, `get_batch_on_this_cp_rank` in data_schedule.py (around L749–L763) already makes the sharding logic explicit. In the current flow I also shard first then broadcast on TP, which reduces broadcast volume and makes the broadcast logic clearer.

parthmannan: Oh I didn't mean to suggest that we should fold THD into `get_batch_on_this_tp_rank` but the opposite. We are folding CP sharding and TP broadcast into 1 function here but I was wondering if we could keep them separate so that other frameworks can use the functions they need instead of re-implementing everything.
Today, we do
```
batch = get_batch_on_this_tp_rank
if thd:
    batch = get_thd_batch_on_this_cp_rank
else:
    batch = get_batch_on_this_cp_rank
```
This allows us to use them separately based on the need.

megatron/core/datasets/data_schedule.py
parthmannan: I understand that the default codepath here should be what Megatron-LM expects but I am thinking if we can reduce the friction while maintaining the expectation. We can keep Scheduler and MCore training logic separate. I am just brainstorming here to ease integration into other frameworks - 

For example, for the `DpBalancedScheduler`, this change has
```
if data_iterator is not None:
            assert tp_group.rank() == 0 and (
                pp_group.rank() == 0 or pp_group.rank() == pp_group.size() - 1
            ), f"Only TP rank 0 and PP stage 0 or -1 should have data_iterator"
```
This assert would mean Nemo-RL etc. have to write an entirely new `DpBalancedScheduler` or add complicated logic to schedule only on TP rank0 and then exchange information. This exchange would be unnecessary as in Nemo-RL, each rank has the data_iterator already.
Can we not separate these assumptions and keep the scheduling logic separate from the data processing logic?
As an idea, we could separate these and inside `wrap_data_iterator`, call these separately
```
scheduler.process_data #Run the checks needed about data_iterator and TP/PP ranks
if tp_rank == 0:
    scheduler.run #Takes data_iterator and schedules
scheduler.broadcast_pp_tp #Perform broadcasts
scheduler.create_new_iterator #Create new iterator
```

Nemo-RL could implement it's own `wrap_data_iterator` where
```
scheduler.run #Takes data_iterator and schedules
scheduler.create_new_iterator #Create new iterator
```

It prevents creating a duplicate scheduler in other frameworks and would help prevent code drifting apart as we update/improve logic.

megatron/core/datasets/readme.md
```diff
@@ -192,6 +192,28 @@ To query the `BlendedDataset` for the _k_-th sample we do the following
...
 
+## Packing Scheduler
+
+The packing scheduler re-schedules variable-length sequences across DP×CP ranks to improve GPU utilization. It is built around the following modules:
+
+### `data_schedule`
+
+This module contains the high-level scheduling logic and entry points:
+
+- **`HybridCPDataLoaderWrapper`**: A wrapper class for hybrid context parallel (CP) scheduling. For every `__next__` call, it: (1) pulls a batch of packed samples from each DP rank, (2) gathers sequence lengths across the DP group, (3) schedules sub-samples using the `BalancedCPScheduler`, (4) reroutes sub-samples to the correct DPxCP ranks via all-to-all communication.
```
lhb8125: Where does the ```HybridCPDataLoaderWrapper``` come from? Is it from another PR?

megatron/core/datasets/readme.md
```diff
@@ -192,6 +192,28 @@ To query the `BlendedDataset` for the _k_-th sample we do the following
...
+
+- **`BasePackingScheduler`**: Abstract base class for packing schedulers. Defines the interface for `get_groups_and_subsamples()` (scheduling algorithm) and `run()` (full scheduling pipeline including fetch, schedule, reroute, pack, broadcast, and VPP handling).
+
+- **`DpBalancedScheduler`**: A concrete scheduler that packs sequences in their original order until reaching the max sequence length limit per DPxCP rank. Supports aligning the number of microbatches to DP size and VPP stage multiples.
+
+- **`wrap_data_iterator()`**: Top-level entry point that wraps an existing `data_iterator`. It creates the appropriate scheduler, runs the scheduling pipeline, broadcast metadata and new num_microbatches, returns a new data iterator along with the updated number of microbatches and FLOPs statistics.
+
+- **`get_batch_on_this_rank_for_sequence_packing()`**: Fetches and broadcasts a single packed microbatch for the current rank. Handles TP/PP broadcasting, constructs `PackedSeqParams` (with `cu_seqlens`, `max_seqlen`, `qkv_format=thd`), and optionally partitions sequences across CP ranks using Transformer Engine's `thd_get_partitioned_indices`.
+
```
lhb8125, line 242: Could you make the functions more structural and hierarchical? Maybe drawing a diagram showing the relationship among the function calls?

lhb8125 (approved):
Overall LGTM except for the docs.

megatron/core/datasets/readme.md
xiaoyao0115: yeah, i will delete this part.

megatron/core/datasets/readme.md
xiaoyao0115: ok, i'll make the changes

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +310,555 @@ def __next__(self) -> Any:
...
+    # Step1: Prepare "tokens", "position_ids" on all ranks.
+    if is_first_stage or mtp_on_this_rank:
+        if is_tp_rank_0:
+            assert batch['tokens'].dtype == torch.int64
+            assert batch['position_ids'].dtype == torch.int64
+            batch['tokens'] = batch['tokens'].view(1, total_tokens)
+            batch['position_ids'] = batch['position_ids'].view(1, total_tokens)
+        else:
+            batch['tokens'] = torch.empty([1, total_tokens], dtype=torch.int64, device=dev)
+            batch['position_ids'] = torch.empty([1, total_tokens], dtype=torch.int64, device=dev)
```
yanring, line 785: `total_tokens` undefined when `mtp_on_this_rank=True` on a middle PP stage
In get_batch_on_this_rank_for_sequence_packing, when mtp_on_this_rank is True but is_first_or_last_stage is False (possible with custom pipeline layouts that place MTP layers on middle stages), the variable total_tokens is never defined.

megatron/core/datasets/data_schedule_utils.py
```diff
@@ -0,0 +1,492 @@
...
+            if pp_group.rank() != pp_group.size() - 1:
+                # middle PP stages receive the broadcasted info and unpack it
+                info_numpy = info_to_broadcast.cpu().numpy()
+                num_micro_batches = int(info_numpy[0])
+                seqlen_sum_this_global_batch = info_numpy[1]
+                seqlen_squared_sum_this_global_batch = info_numpy[2]
+                max_seqlens = info_to_broadcast[3 : 3 + num_micro_batches]
+                cu_seqlens_list = []
+                cu_seqlens_padded_list = []
+                indices = np.where(info_numpy == 0)[0]
```
yanring, line 242: `broadcast_to_pp_group` uses `np.where(info_numpy == 0)` to find `cu_seqlens` boundaries, which is fragile and incorrect if any metadata value is zero
The receiver side uses np.where(info_numpy == 0) to locate cu_seqlens boundaries (since cu_seqlens always start with 0). Will this silently produce incorrect results if any other value in the tensor happens to be 0?

megatron/training/training.py
yanring, line 524: qq:

In the old code, the structure was:
`total = batch_size * seq_length * (... + self_attn_term ...)`
where both linear_self_attn_term and standard_self_attn_term were per-token values, and the outer batch_size * seq_length scaled them both.
In the new code, the structure is:
`total = seqlen_sum_this_global_batch * (MLP + MTP + Logit) + self_attn_term`
where standard_self_attn_term now includes seqlen_sum_this_global_batch internally (lines 416, 446), but linear_self_attn_term (lines 504-522) does NOT. Since self_attn_term = linear_self_attn_term * num_linear_attention_layers + standard_self_attn_term * num_standard_attention_layers, the linear attention FLOPs are underestimated by a factor of seqlen_sum_this_global_batch.

megatron/training/arguments.py
```diff
@@ -921,6 +914,18 @@ def validate_args(args, defaults={}):
...
         assert args.calculate_per_token_loss, 'Hybrid context parallelism must be used with --calculate-per-token-loss'
 
+    # Support for variable sequence lengths across batches/microbatches.
+    # set it if the dataloader supports generation of variable sequence lengths
+    # across batches/microbatches. Due to additional communication overhead
+    # during pipeline parallelism, it should not be set if sequence length
+    # is constant during training.
+    args.variable_seq_lengths = False
+    if args.sequence_packing_scheduler is not None:
+        args.variable_seq_lengths = True
```
yanring: `variable_seq_lengths` is currently set in both args and transformer_config. Should we consider keeping it only in transformer_config?

megatron/core/datasets/data_schedule.py
xiaoyao0115: Good catch, i will fix this

megatron/core/datasets/data_schedule_utils.py
xiaoyao0115: Yeah,  this is pretty hacky. However, in practice the other metadata values  should never be 0 — `num_micro_batches`, `seqlen_sum_this_global_batch`, `seqlen_squared_sum_this_global_batch` and `max_seqlen` are always positive, and `cu_seqlens` always increases after the initial 0. So using `np.where(info == 0)` to locate `cu_seqlens` boundaries works correctly.
I can either add a comment clarifying the assumption, or add an explicit broadcast for the `cu_seqlens` length to make it more robust. Which would you prefer?

megatron/training/training.py
xiaoyao0115: I will fix this.

megatron/training/arguments.py
xiaoyao0115: okay

megatron/core/datasets/data_schedule_utils.py
yanring: lets add some comments

PR #2456, opened 2025-12-02
FlashInfer sampling

tdene:
# What does this PR do ?

:warning: For major changes (either in lines of code or in its impact), please make sure to first share discuss a design-doc with the team.  

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
        C1[Expert Review] --> C2[Final Review]
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [ ] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

The following process is enforced via the CODEOWNERS file for changes into `megatron/core`. For changes outside of `megatron/core`, it is up to the PR author whether or not to tag the Final Reviewer team.

<details>
<summary>For MRs into `main` branch</summary>

#### (Step 2): Collect the expert reviewers reviews

1. Attach the `Expert Review` label when your PR is ready for review.
2. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

:warning: Only proceed to the next step once all reviewers have approved, merge-conflict are resolved and the CI is passing.  
Final Review might get declined if these requirements are not fulfilled.

#### (Step 3): Final Review

1. Add `Final Review` label
2. GitHub auto-assigns final reviewers based on your changes. They will get notified and pick up your PR soon.

#### (Optional Step 4): Cherry-pick into release branch

If this PR also needs to be merged into `core_r*` release branches, after this PR has been merged, select `Cherry-pick` to open a new PR into the release branch.

</details>

<details>
<summary>For MRs into `dev` branch</summary>
The proposed review process for `dev` branch is under active discussion.

MRs are mergable after one approval by either `eharper@nvidia.com` or `zijiey@nvidia.com`.
</details>

### Merging your PR

Any member of [core-adlr](https://github.com/orgs/teams/NVIDIA/core-adlr) and [`core-nemo`](https://github.com/orgs/teams/NVIDIA/core-nemo) will be able to merge your PR.

PR #2929, opened 2026-01-13
Reorganize order of operations in inference context and text generation controller

tdene:
# What does this PR do ?
This PR cleans up the order of operations in order to easily enable future optimizations.

In addition, this PR adds some common sugar that is used by multiple such optimizations.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


Phlip79 (changes requested):
Please add tests for the new logic.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -866,20 +886,69 @@ def cu_kv_lengths(self) -> Tuple[Tensor, Tensor, int]:
...
+        """Pad the active slices of specific tensors."""
+        # Some tensors need to be padded at the token level.
+        padding_token_slice = slice(self.active_token_count, self.padded_active_token_count)
+
+        self.token_to_block_idx[padding_token_slice] = self.block_allocator.dummy_block_idx
+        self.token_to_local_position_within_kv_block[padding_token_slice] = 0
+        self.token_to_position_in_request[padding_token_slice] = 0
+
+        # Other tensors need to be padded at the request level.
+        padding_request_slice = slice(
```
lmcafee-nvidia: not sure if I'm missing something, but is this used?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -634,6 +634,26 @@ def allocate_all_tensors(self, *, is_init: bool) -> None:
         self.token_to_position_in_request = torch.empty_like(self.token_to_input_ids)
         self.token_to_local_position_within_kv_block = torch.empty_like(self.token_to_input_ids)
 
+        # Static tensor addresses of active slices to enable fast inference kernels.
```
lmcafee-nvidia, line 863: what's your timeline for merging this .... this might be a good argument for re-ordering the context tensors and even simplifying what we already have."

I really don't like the idea of continually adding duplicated tensors, including the ones we already have, if all we need to do is bite the bullet and redesign things.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -866,20 +886,69 @@ def cu_kv_lengths(self) -> Tuple[Tensor, Tensor, int]:
...
-        return self.request_output_lengths[self.paused_request_count : self.total_request_count]
-
     def get_active_request_count(self):
         """Returns the current number of active requests."""
         return self.total_request_count - self.paused_request_count
 
+    def build_active_slices(self, batch_size: int):
+        """Build the active slices of specific tensors. This is run on every forward step.
+
+        If the context is reordered to active -> paused -> finished, this can be graphed.
```
lmcafee-nvidia, line 1073: this method can be graphed? Or we wouldn't need this method at all?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -866,20 +886,69 @@ def cu_kv_lengths(self) -> Tuple[Tensor, Tensor, int]:
...
+        padded_slice = slice(self.paused_request_count, self.paused_request_count + batch_size)
+
+        # Request metadata all needs to be sliced.
+        for label, _, _ in self.request_metadata_types:
+            self.active_request_metadata[label][:batch_size].copy_(
+                self.request_metadata[label][padded_slice], non_blocking=True
+            )
+
+        # The following tensor slices are used in various kernels.
+        self.active_request_ids[:batch_size].copy_(self.request_ids[padded_slice])
```
lmcafee-nvidia: if we re-order active and paused, can all these copies be avoided?

megatron/core/inference/contexts/dynamic_context.py
tdene: At a high-level, I see this PR as the base that enables several bite-sized optimization PRs and benchmarks them to see how much improvement each optimization leads to - in order to prove their value in a clean apples-to-apples comparison.

One of these bite-sized optimization PRs is the one that reorders the context tensors.

That is, we could reorder the context tensor today, but we will not have a clear idea of how much throughput/latency we gain by doing so. Going through the work of building things on top of this PR would clearly prove how much of a practical win we can get from reordering the context tensor.

tdene: But your point is valid. I'm saying that we should do things in order so that we have proper ablation studies. We do not have to actually merge things in that order though; I can do the ablation study, come back with results, and then say "Okay, let's just skip #2929 and go straight to re-ordering the context tensors."

megatron/core/inference/contexts/dynamic_context.py
tdene: Exactly!!!

megatron/core/inference/contexts/dynamic_context.py
tdene: Everything listed in this PR, yes.

But all PRs that built on top of this one end up needing to add things to `pad_active_slices` - and that method is used to pad tensors on the interval `active_request_count : padded_active_request_count`. The need to pad means that we do need copies of the tensor, so that we don't overwrite real data.

We see that today in the attention metadata tensors, which currently do a copy up to `padded_active_request_count` followed by a pad on the interval `active_request_count : padded_active_request_count`. Exactly the same pattern is needed for graphed logprobs, graphed sampling, and graphing time-critical parts of `_dynamic_step_context_bookkeeping`.

tdene: So, in a nutshell, here's what's happening:
 * This method, as it is in this PR, copies several tensors that are used in CUDA graphs in follow-up PRs. All of these copies go away if we reorder the context tensors.
 * There are also several tensors that are copied & padded in follow-up PRs. If we reorder the context tensors, those copies do not go away, but they become graphable.

megatron/core/inference/contexts/dynamic_context.py
tdene: Correct.

This PR has gone through many iterations: I have several different optimizations, and I've been trying to extract the code changes they have in common.

Each one of the follow-up PRs adds `self.active_tensor[padding_request_slice] = 0` lines in this method, so each one needs this `padding_request_slice` that is calculated here. If this calculation was not included in this PR, it'd be included in every follow-up PR separately.

But when I tried to make this PR as small as possible, I ended up removing all uses to `padding_request_slice` in this PR itself, so it looks like orphaned code.

lmcafee-nvidia: makes sense! thanks

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: so will your follow-up PRs be relatively independent from this duplicated-tensor style? As in, if we re-order active & paused requests in the near future, will that require changing all of your follow-up PRs significantly?

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: exactly == this method can be graphed?
-or-
exactly == we won't need this method at all?

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: orthogonal to your PRs and possibly less practical, but I wonder if we can merge regular and padded tensors also and just have a single tensor. We care mostly about the cuda graph case, or at least that's our gold standard for performance, so it seems like we could just used the padded tensors in the non-cuda-graph case as well for simplicity

Phlip79 (approved):

megatron/core/inference/contexts/dynamic_context.py
tdene: It will not.

megatron/core/inference/contexts/dynamic_context.py
tdene: Most of this method goes away.

The small amount that's still necessary can be graphed.

megatron/core/inference/contexts/dynamic_context.py
tdene: That is fair. Will consider.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1062,6 +1067,23 @@ def get_active_request_count(self):
...
+
+        # Request metadata all needs to be sliced.
+        for label in self.request_metadata:
+            self.active_request_metadata[label][:batch_size].copy_(
+                self.request_metadata[label][padded_slice], non_blocking=True
+            )
+
+    def pad_active_slices(self):
+        """Pad the active slices of specific tensors."""
+        pass
```
santhnm2, line 1085: Why do we have this function if it's empty?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -400,37 +400,53 @@ def create_cuda_graphs(self, reset_context: bool = True):
...
+            # Note that some steps of the pipeline may capture multiple different variant graphs.
+            with torch.inference_mode():
+                for setup_variant in controller.graph_capture_variants():
+                    setup_variant(context)
+                    controller._pre_forward_bookkeeping_stream.wait_stream(
+                        torch.cuda.current_stream()
+                    )
+                    # Launch bookkeeping on a side stream so it overlaps with forward.
+                    with torch.cuda.stream(controller._pre_forward_bookkeeping_stream):
+                        controller._pre_forward_bookkeeping_event.record()
```
santhnm2: What is actually being overlapped here? It's confusing to me that we have `_pre_forward_bookkeeping` code that overlaps with `_forward` itself.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -400,37 +400,53 @@ def create_cuda_graphs(self, reset_context: bool = True):
...
-                                ),
-                                next_token_ids=torch.zeros((1, n), device=device, dtype=torch.long),
-                                position_ids=torch.zeros((1, n), device=device, dtype=torch.int64),
-                                depth=depth,
-                            )
+            # Capture all relevant graphs in the pipeline.
+            # Note that some steps of the pipeline may capture multiple different variant graphs.
+            with torch.inference_mode():
+                for setup_variant in controller.graph_capture_variants():
+                    setup_variant(context)
```
santhnm2: What does `setup_variant` do?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -138,6 +138,10 @@ def _init_dynamic_sampling_tensors(self):
         """Initialize tensors needed for dynamic sampling."""
         context = self.inference_wrapped_model.inference_context
         max_requests = context.max_requests
+        if context.config.materialize_only_last_token_logits:
+            max_logits = max_requests
+        else:
+            max_logits = context.max_tokens
```
santhnm2, line 145: Is this correct even with MTP (i.e., `num_speculative_tokens > 0`? Or is this code not exercised in the speculative case?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -689,13 +688,24 @@ def _dynamic_step_forward_logits(self, input_ids: Tensor, position_ids: Tensor)
         """
         context = self.inference_wrapped_model.inference_context
         active_request_count = context.total_request_count - context.paused_request_count
+        logits_seq_len = (
+            active_request_count
+            if context.config.materialize_only_last_token_logits
+            else context.padded_active_token_count
+        )
```
santhnm2: Same comment about correctness with MTP

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -724,22 +734,26 @@ def _dynamic_step_forward_logits(self, input_ids: Tensor, position_ids: Tensor)
                 pp_group=self.pp_group,
             )
 
-        return logits
+        # Copy logits to contiguous buffer.
+        if self._enable_cuda_graph:
+            self._all_logits_cuda[:, :logits_seq_len, :].copy_(logits)
```
santhnm2: Probably not worth the complexity, but I wonder if we could pipe this buffer into the output layer and have it directly write into the buffer rather than doing an extra copy. Feel free to ignore for now, or defer with a TODO comment.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -1059,6 +1073,7 @@ def _dynamic_step_sample_logits_and_verify_tokens(self, logits: Tensor, input_id
         # These indices are always needed for input_ids slicing and tracking
         # accepted sequence positions, even when logits are pre-sliced.
         nvtx_range_push("mtp-spec-decoding/verify/logit-indices")
+        logits = self._all_logits_cuda
```
santhnm2, line 1060: Add a comment that we're using the pre-allocated buffer for CUDA graph compatibility

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -1253,20 +1265,23 @@ def _router_record_bookkeeping(self) -> Optional[np.ndarray]:
...
-    def _dynamic_step_calculate_log_probs(self, logits: Tensor) -> Optional[Tensor]:
+    def _dynamic_step_calculate_log_probs(self) -> Optional[Tensor]:
         """Calculate log probs from logits."""
         context = self.inference_wrapped_model.inference_context
         active_request_count = context.total_request_count - context.paused_request_count
+        logits_seq_len = (
+            active_request_count
+            if context.config.materialize_only_last_token_logits
+            else context.padded_active_token_count
+        )
```
santhnm2, line 1264: Same comment about correctness with MTP

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -1300,6 +1312,7 @@ def _dynamic_step_calculate_log_probs_speculative(
         num_decode_requests = active_request_count - num_prefill_requests
 
         only_last = context.config.materialize_only_last_token_logits
+        logits = self._all_logits_cuda
```
santhnm2, line 1304: Add a comment that we're using the pre-allocated buffer for CUDA graph compatibility

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -1585,6 +1594,18 @@ def _dynamic_step_calculate_top_n_logprobs(
...
+        """Yield context-setup callables for each graph-capture variant.
+
+        During graph warmup, the engine runs the full step pipeline once per yielded callable.
+        Each callable exercises a different kernel path.
+        """
+        if not self._enable_cuda_graph:
+            yield lambda context: None
+            return
+
+        yield lambda context: None
```
santhnm2: This function's purpose is not clear to me, what does this do?

megatron/core/inference/contexts/dynamic_context.py
tdene: It's a common function used by all following PRs, such as #2456 #2457

I wanted to have it to minimize the diff / conflicts on those two PRs; that way the following PRs stay relatively independent.

I can absolutely take it out if you'd like; let me know.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
tdene: Oh we absolutely can, and we will.

#4380 was doing exactly this! But then it had a really nasty collision with a different PR, and I have to re-write it from scratch.

megatron/core/inference/engines/dynamic_engine.py
tdene: Here's an example:
```
            # Launch indexing on the pre-forward side stream so it overlaps with the forward pass.
            self._pre_forward_bookkeeping_stream.wait_stream(torch.cuda.current_stream())
            with torch.cuda.stream(self._pre_forward_bookkeeping_stream):
                self._dynamic_step_log_probs_indexing()
                self._pre_forward_bookkeeping_event.record()
```

It's for kernels that need to be launched BEFORE the forward pass, but their completion is only needed AFTER the forward pass.

I hide D2H syncs by making sure this kind of work overlaps with the forward pass, instead of just running it after the forward pass and having to pay the sync cost.

I've gone ahead and also removed it from this base PR; graphed sampling and graphed logprobs can each re-implement it.

megatron/core/inference/engines/dynamic_engine.py
tdene: It abstracts away things like the `for depth in mtp_warmup_depths:` block below.    
Basically, every time we want to capture multiple variants of a graph, we would add it to `graph_capture_variants`, and that would set up all the variants for you.    
The idea is that eventually, this code is just
```
with torch.inference_mode():
    for setup_variant in controller.graph_capture_variants():
        setup_variant(context)
        controller.generate_output_tokens_dynamic_batch()
```
with no other fluff.

Today itt's only used by the graphed sampling PR actually, so it's not shared code. Originally I was using this for the graphed logprobs PR as well, but I no longer need to.

I've gone ahead and removed it from this base PR, and graphed sampling can re-implement it.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
tdene: See above; but also, I've gone ahead and removed it from this base PR.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
tdene: Done!

megatron/core/inference/text_generation_controllers/text_generation_controller.py
tdene: Done

megatron/core/inference/text_generation_controllers/text_generation_controller.py
tdene: Added a comment addressing why this cannot be an issue with MTP.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
tdene: It was not correct; fixed.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
tdene: It was not correct; fixed.

megatron/core/inference/contexts/dynamic_context.py
santhnm2: I see, let's leave it in for now since we're planning on merging in the other PRs soon anyway

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -689,13 +684,27 @@ def _dynamic_step_forward_logits(self, input_ids: Tensor, position_ids: Tensor)
...
         active_request_count = context.total_request_count - context.paused_request_count
+        if context.config.materialize_only_last_token_logits:
+            if self.num_speculative_tokens > 0:
+                # Under MTP, each decode request emits (num_speculative_tokens + 1) logit rows.
+                logits_seq_len = (
+                    context.num_decode_requests * (self.num_speculative_tokens + 1)
+                    + context.num_prefill_requests
+                )
+            else:
+                logits_seq_len = active_request_count
```
santhnm2: Can we consolidate this with the `last_token_logits` function in the context? I suspect either this code or that code will be redundant now but I'm not sure what the exact implications are

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: Can you add an assertion confirming this?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
tdene: I believe this is done.

santhnm2 (approved):

kvareddy (approved):

shanmugamr1992 (approved):

lmcafee-nvidia (approved):

PR #2938, opened 2026-01-14
Enable phase transition iterations

jkamalu:
# What does this PR do ?

Enables phase transition iterations for multi-phase training with the correct token horizons per phase

Also fixes the string concatenation for the IndexedDataset FileNotFoundError


Phlip79 (approved):

jkamalu:
> Approved, but would be nice to have some more documentation around this. Can we add some more docs in a future PR?
> 
> Also, do we need rampup batch size still?

@ericharper Were you thinking more comments? Or a beefier help message in arguments.py? Both?

And for batch size ramp-up... I don't know if we do. It'd be nice to deprecate so we didn't have to account for it in new features, in the data samplers, etc.

ericharper:
> > Approved, but would be nice to have some more documentation around this. Can we add some more docs in a future PR?
> > Also, do we need rampup batch size still?
> 
> @ericharper Were you thinking more comments? Or a beefier help message in arguments.py? Both?
> 
> And for batch size ramp-up... I don't know if we do. It'd be nice to deprecate so we didn't have to account for it in new features, in the data samplers, etc.

I was thinking user guide for multi-phase training.

jkamalu:
> I was thinking user guide for multi-phase training.

I see. Yes, this can come in a future PR.

PR #2942, opened 2026-01-14
Fix VPP redundant dataset initialization: share data iterators across…

### LINKED ISSUE #2860: VPP causes redundant dataset initialization leading to mmap virtual memory exhaustion
# Describe the bug
When using Virtual Pipeline Parallelism (--virtual-pipeline-model-parallel-size > 1), the training script redundantly initializes the dataset and data iterators once per virtual stage (i.e., len(model) times). This leads to multiple independent np.memmap calls on the same large data files (e.g., 16GB .bin files), causing excessive virtual address space consumption. On systems with TB-scale datasets, this results in:

OSError: [Errno 12] Cannot allocate memory

during mmap.mmap() even when physical memory and vm.max_map_count are sufficient.
This issue does not occur when VPP is disabled, confirming that the redundant initialization—not the data itself—is the root cause.
Tagging [@megatron-oncall](https://github.com/orgs/NVIDIA/teams/megatron-oncall) for attention.
# Steps/Code to reproduce bug
1. Prepare a dataset consisting of one or more large binary files (e.g., 16GB each) loaded via np.memmap in a custom Dataset.
2. Launch training with VPP enabled, e.g.:

python pretrain_gpt.py \
  --virtual-pipeline-model-parallel-size 4 \
  --pipeline-model-parallel-size 2 \
  ... # other PP/MP settings

3. Observe that build_train_valid_test_data_iterators is called 4 times (once per virtual model).
4. Training fails during dataset initialization with OSError: [Errno 12] Cannot allocate memory.
Minimal code pattern causing the issue (from pretrain_*.py):

if args.virtual_pipeline_model_parallel_size is not None:
    train_data_iterator = []
    for i in range(len(model)):
        # ❌ Called N times (N = VPP size)
        iterators = build_train_valid_test_data_iterators(provider)
        train_data_iterator.append(iterators[0])

# Expected behavior
Data iterators should be initialized only once per rank, regardless of VPP configuration. All virtual pipeline stages should share the same data iterator instance, as they consume micro-batches sequentially from a single global data stream.
The correct pattern should be:

train_iter, valid_iter, test_iter = build_train_valid_test_data_iterators(provider)

if args.virtual_pipeline_model_parallel_size is not None:
    # Share the same iterator across all virtual stages
    train_data_iterator = [train_iter] * len(model)
    # ... similarly for valid/test
This avoids redundant dataset initialization and prevents virtual memory exhaustion.

# Additional context
Observed in Megatron-LM versions 0.5 through 0.15.
The problem becomes severe with large mmap-backed datasets (common in LLM pretraining).
vm.max_map_count is already set to a very high value (2147483642), ruling out mmap region limits.
The fix is backward-compatible and aligns with how other frameworks (e.g., DeepSpeed, ColossalAI) handle VPP data loading.
This is a resource usage bug, not a correctness bug—training works if memory permits, but unnecessarily fails under realistic large-data conditions.

CodersAcademy006:
#### Summary
This PR fixes an issue where enabling Virtual Pipeline Parallelism (VPP) causes the training script to initialize datasets and data iterators once per virtual pipeline stage. This results in multiple independent `np.memmap` mappings of the same large dataset files, leading to virtual address space exhaustion and training failures with large mmap-backed datasets.

The fix ensures that data iterators are initialized **once per rank** and **shared across all virtual pipeline stages**, which aligns with how VPP consumes microbatches sequentially on a single rank.

#### Problem
When `--virtual-pipeline-model-parallel-size > 1`, the current code path calls  
`build_train_valid_test_data_iterators()` inside a loop over `len(model)`, creating multiple datasets and memory maps per rank. With large datasets (e.g., tens or hundreds of GB), this leads to: `OSError: [Errno 12] Cannot allocate memory` even when physical memory and `vm.max_map_count` are sufficient.

#### Solution
- Build train/valid/test data iterators **once per rank**
- Reuse the same iterator instance across all virtual pipeline stages
- Preserve existing behavior for non-VPP runs

This avoids redundant dataset initialization, prevents excessive `mmap` usage, and is fully backward-compatible.

#### Additional Notes
- Virtual pipeline stages consume microbatches sequentially on the same rank, so sharing a single iterator instance is safe and correct.
- This behavior matches how other distributed training frameworks handle VPP data loading.

Fixes #2860

### DIFF
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -763,24 +763,14 @@ def pretrain(
     # Data stuff.
     app_metrics['app_build_dataiters_start_time'] = one_logger_utils.get_timestamp_in_ms()
     timers('train/valid/test-data-iterators-setup', log_level=0).start(barrier=True)
+    # VPP fix: Only initialize data iterators once and share across all virtual pipeline stages.
     if args.virtual_pipeline_model_parallel_size is not None:
-        train_data_iterator = []
-        valid_data_iterator = []
-        test_data_iterator = []
-        for vp_stage in range(len(model)):
-            dataset_provider_parameters = inspect.signature(train_valid_test_dataset_provider).parameters
-            assert "vp_stage" in dataset_provider_parameters, \
-                "vp_stage must be a kwarg in train_valid_test_dataset_provider when using virtual pipeline parallelism"
-            vp_stage_train_valid_test_dataset_provider = \
-                functools.partial(train_valid_test_dataset_provider, vp_stage=vp_stage)
-            if getattr(train_valid_test_dataset_provider, 'is_distributed', False):
-                vp_stage_train_valid_test_dataset_provider.is_distributed = True
-            iterators = build_train_valid_test_data_iterators(
-                vp_stage_train_valid_test_dataset_provider
-            )
-            train_data_iterator.append(iterators[0])
-            valid_data_iterator.append(iterators[1])
-            test_data_iterator.append(iterators[2])
+        # Build iterators once
+        train_iter, valid_iter, test_iter = build_train_valid_test_data_iterators(train_valid_test_dataset_provider)
+        train_data_iterator = [train_iter] * len(model)
+        valid_data_iterator = [valid_iter] * len(model)
+        test_data_iterator = [test_iter] * len(model)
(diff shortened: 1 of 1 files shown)


Phlip79:
Please see response: https://github.com/NVIDIA/Megatron-LM/issues/2860#issuecomment-3750946019. Closing this PR for now.

PR #2951, opened 2026-01-14
Nemo RL upstream

yaoyu-33:
Cherry-pick of NVIDIA/Megatron-LM#2013 onto main.

### DIFF
--- a/megatron/core/distributed/param_and_grad_buffer.py
+++ b/megatron/core/distributed/param_and_grad_buffer.py
@@ -806,7 +806,7 @@ class _ParamAndGradBuffer:
                     device=torch.cuda.current_device(),
                     requires_grad=False,
                 )
-   
+
         self.grad_data_size = 0
         self.param_data_size = 0
         self.param_data_cpu = None


megatron/core/pipeline_parallel/schedules.py
```diff
@@ -226,6 +226,7 @@ def forward_step_calc_loss(
     forward_data_store,
     cp_group_size=None,
     is_last_stage=None,
+    do_not_average_loss=False,
```
jaredcasper, line 229: Please no double negatives in arguments. This should be `average_loss`.

yashaswikarnati, line 229: Im not sure i easily follow what is do_not_average_loss doing.. also having calculate per token loss in config and this other arg here in the function signature is a bit confusing

yaoyu-33: This is from RL. Lemme check if the per token loss they can leverage here.

yaoyu-33: good point will fix

yaoyu-33: RL is handling there own way to do average essentially

yaoyu-33:
was able to update in RL to fulfill the feat.

PR #2955, opened 2026-01-15
Miscellaneous inference cleanup

santhnm2:
# What does this PR do ?

This PR makes the following quality-of-life improvements to inference:
- Removes `InferenceWrapperConfig`.
- Consolidates all the redundant `TransformerConfig` arguments in `DynamicInferenceContext` into a single `model_config` argument.
- Adds an `InferenceConfig` object that has all the inference-related configuration parameters from the command line arguments.
- Adds `megatron/inference` as a location for convenience functions to initialize an `InferenceConfig` or `DynamicInferenceEngine` using the global args.
- Updates all driver scripts to use these convenience functions.
- Moves some commonly used utility functions out of `gpt_dynamic_inference.py` into `megatron/inference/utils.py`.
- Deprecates the `pipeline_parallel_large_batch` function in `AbstractModelInferenceWrapper`
- Removes `simple_text_generation_controller.py` 
- Ensures that the static engine `generate` function always returns `InferenceRequest` objects instead of `DynamicInferenceRequestRecord` objects.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -227,6 +227,43 @@ def __init__(
         # Create cuda graphs.
         self.create_cuda_graphs()
 
+    @classmethod
+    def from_model_and_args(
+        cls,
+        model,
+        args,
```
wdykas: Do we take arg objects into core in other places like this?  It technically doesn't violate the idea of not argument parsing in core but feels a bit weird if someone were to use this outside Megatron-lm like in Nemo

wdykas: I think related to this, I see seed and cuda_graph_impl in the transformer config. I think the model will always have the config attached where we can access these vars from there since Megatron module keeps a copy of config https://github.com/wdykas/Megatron-LM/blob/nvshmem-refit/megatron/core/transformer/module.py#L41

megatron/core/inference/model_inference_wrappers/abstract_model_inference_wrapper.py
```diff
@@ -78,40 +59,18 @@ def __init__(
...
-    def inference_params(self, value):
-        """Setter for deprecated `inference_params`."""
-        warnings.warn(
-            "`inference_params` renamed to `inference_context`, and will be removed in `megatron-core` 0.13."
-        )
-        self.inference_context = value
+        # TODO(ksanthanam): Add support for fp4
 
-    def prep_model_for_inference(self, prompts_tokens: Optional[torch.Tensor] = None):
+    def prep_model_for_inference(self):
```
wdykas, line 67: Is this just general clean up where prompt_tokens was never used?

megatron/core/inference/model_inference_wrappers/abstract_model_inference_wrapper.py
```diff
@@ -368,34 +232,25 @@ def run_one_forward_step(
...
-            ):
-                return self.forward_pass_with_pipeline_parallel_large_input_batch(
-                    inference_input, recv_buffer_seq_len
-                )
-            else:
-                # If input batch is very small we can do a simple forward pass on the entire global batch
-                return self.forward_pass_with_pipeline_parallel_small_input_batch(
-                    inference_input, recv_buffer_seq_len
-                )
+            return self.forward_pass_with_pipeline_parallel(inference_input, recv_buffer_seq_len)
```
wdykas, line 266: Love that we are getting rid of the above if statements! Thanks for this clean up.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -71,13 +84,52 @@ def __init__(
             is_pipeline_first_stage(self.pp_group) and is_pipeline_last_stage(self.pp_group)
         )
 
-        model_config = get_model_config(self.inference_wrapped_model.model)
+        # Use padded vocab size because tokenizer vocab size might pad to nearest power of 2.
+        try:
```
wdykas: Can we do this without a try: except: or is this the cleanest way?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -71,13 +84,52 @@ def __init__(
...
+        self.sampling_rng.manual_seed(self.model_config.inference_sampling_seed)
 
         if self.inference_wrapped_model.inference_context.is_dynamic_batching():
             self._init_dynamic_sampling_tensors()
 
+    @classmethod
+    def from_model_and_args(
+        cls,
+        model,
+        args,
```
wdykas: What is your thinking on passing in args here and then building the tokenizer? I think args is only used for building the tokenizer so what about just accepting tokenizer as an arg

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -71,13 +84,52 @@ def __init__(
...
+            model_inference_wrapper_cls (type[AbstractModelInferenceWrapper]): The class
+                used to wrap the model for inference. Defaults to GPTInferenceWrapper.
+
+        Returns:
+            TextGenerationController: The initialized text generation controller.
+        """
+        tokenizer = build_tokenizer(args)
+        # TODO(ksanthanam): Condition this on model type?
+        model = model_inference_wrapper_cls(model, context)
+        model.model_is_pipeline_parallel = not (
```
wdykas: Can we use non-global parallel state here please? This also seems like a weird way to check for pipeline parallel, could we just grab from the model config the pp size?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -779,28 +764,42 @@ def from_config(
...
-            num_cuda_graphs=num_cuda_graphs,
-            use_flashinfer_fused_rope=None,
-            mamba_inference_state_config=mamba_inference_state_config,
-            unified_memory_level=unified_memory_level,
-        )
+
+        mamba_inference_state_config = get_mamba_inference_state_config_from_model(model)
+        pg_collection = get_attr_wrapped_model(model, "pg_collection")
+
+        kwargs = {
```
wdykas: Related to the above talking points about having arg object passed into core. Is this allowed and do you think an outside user like NeMo will not like this pattern? Makes it nice for us but may confuse someone not using Megatron-lm

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: Good point, I'll explicitly check the model type

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: Yea I thought about this but I wanted to keep the `from_model_and_args` name / interface consistent across all of the difference classes. One drawback of this approach though is that this won't work with the legacy tokenizer anymore because that requires calling `get_tokenizer` from MLM. Open to discussion here

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: This was just copied from existing functionality but yea good point, I'll clean this up

megatron/core/inference/engines/dynamic_engine.py
santhnm2: Ah yea I can use the config for these then.

examples/inference/gpt/gpt_dynamic_inference.py
```diff
@@ -546,7 +424,13 @@ def escape_str(s):
             # ---- Prompt summary line ----
             prompt_len = len(requests[request_idxs[0]].prompt_tokens)
             escaped_prompt_text = escape_str(prompt_text)
+<<<<<<< HEAD
+            print(
+                f"{unique_idx+1}/{len(unique_prompt_map)} [n {len(request_idxs)}, l {prompt_len}] {escaped_prompt_text}"
+            )
+=======
             print(f"\n{unique_idx+1}/{len(unique_prompt_map)} [n {len(request_idxs)}, l {prompt_len}] {escaped_prompt_text}")
+>>>>>>> upstream/main
```
tdene: reminder: uncleaned PR merge

examples/inference/gpt/gpt_dynamic_inference.py
```diff
@@ -566,16 +450,20 @@ def escape_str(s):
...
                     o_hash = "--"
                     o_len = 0
                     escaped_output_text = "--"
+<<<<<<< HEAD
+                print(
+                    f"  >>>> [n {len(output_request_idxs)}, {o_len} tokens, hash {o_hash}] {escaped_output_text}"
+                )
+=======
                 print(f"  >>>> [n {len(output_request_idxs)}, {o_len} tokens, hash {o_hash}{', <evicted>' if evicted else ''}] {escaped_output_text}")
+>>>>>>> upstream/main
```
tdene: reminder: uncleaned PR merge

megatron/rl/inference/megatron.py
```diff
@@ -66,131 +55,43 @@ def get_static_inference_engine(args: Namespace, model: MegatronModule) -> Abstr
...
-        track_paused_request_events=args.inference_dynamic_batching_track_paused_request_events,
-        enable_chunked_prefill=not args.disable_chunked_prefill,
-        inference_logging_step_interval=inference_logging_step_interval,
-        pg_collection=pg_collection,
+    context = DynamicInferenceContext.from_model_and_args(
+        model, args, overrides={"metrics_writer": metrics_writer}
     )
+    controller = TextGenerationController.from_model_and_args(model, args, context)
+    engine = DynamicInferenceEngine.from_model_and_args(model, args, controller, context)
+    return engine
```
tdene: Since `metrics_writer` is only used inside the engine anyway (the context just stores it, but does nothing with it), I think that you should just go ahead and move the `metrics_writer` arg out of the context and directly into the engine.

Then you can do a 1-line `DynamicInferenceEngine.from_model_and_args(model, args, overrides)` and let the engine auto-initialize the controller and context. It gets rid of the entire `get_dynamic_inference_engine` method too.

santhnm2: Done, thanks

examples/inference/gpt/gpt_dynamic_inference.py
santhnm2: Fixed, thanks

examples/inference/gpt/gpt_dynamic_inference.py
santhnm2: Fixed, thanks

megatron/core/inference/engines/static_engine.py
```diff
@@ -69,55 +67,58 @@ def __init__(
...
 
                 self.dynamic_engine = DynamicInferenceEngine(
                     controller=self.controller,
-                    random_seed=self.random_seed,
                     context=dynamic_context,
-                    enable_cuda_graph=True,
+                    random_seed=self.random_seed,
                 )
         except Exception as e:
+            torch.distributed.breakpoint(0)
```
lmcafee-nvidia: remove breakpoint

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -227,6 +230,69 @@ def __init__(
...
+                    logging.WARNING,
+                    "WARNING: --rl-inference-logging-step-interval is set but no metrics writer "
+                    "wandb module is available. Inference logging will be disabled.",
+                )
+
+        return cls(
+            controller,
+            context,
+            enable_cuda_graph=model.config.cuda_graph_impl == "local",
+            random_seed=amodel.config.seed,
```
lmcafee-nvidia: `amodel` -> `model`

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -748,25 +744,25 @@ def round_up_tokens(cls, value, tp_size=None):
...
-        cls,
-        inference_config: InferenceWrapperConfig,
-        model,
-        max_batch_size: int,
-        buffer_size_gb: float = 40,
-        num_cuda_graphs: int = None,
-        mamba_inference_state_config: Optional[MambaInferenceStateConfig] = None,
-        unified_memory_level: int = 0,
-    ):
+    def from_model_and_args(cls, model, args, overrides: Optional[Dict[str, Any]] = None):
```
lmcafee-nvidia: I'm guessing this method has so far only been used in the static engine, but if it was more widespread, we might want to properly deprecate `from_config()`.

examples/inference/gpt/gpt_dynamic_inference_with_coordinator.py
```diff
@@ -212,31 +209,13 @@ async def main(
...
-        # Inference engine.
-        engine = DynamicInferenceEngine(
-            controller,
-            context,
-            enable_cuda_graph=args.cuda_graph_impl == "local",
-            random_seed=args.seed,
-            enable_chunked_prefill=not args.disable_chunked_prefill,
-            inference_logging_step_interval=args.inference_logging_step_interval,
-        )
+        engine = DynamicInferenceEngine.from_model_and_args(model, args)
```
lmcafee-nvidia: something about passing global args makes me really uncomfortable ... I feel like we should probably have an `DynamicInferenceContextConfig` (or just `InferenceContextConfig`), which would be primarily for the dynamic context at this point

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -259,37 +245,56 @@ class DynamicInferenceContext(BaseInferenceContext):
...
+            tensor_model_parallel_size,
+            pipeline_model_parallel_size,
+            cache_mla_latent,
+            kv_lora_rank,
+            qk_pos_emb_head_dim,
+        ]
+        if any(param is not None for param in deprecated_params):
+            raise TypeError(
+                "Passing `TransformerConfig` arguments directly is deprecated. "
+                "Please pass `model_config` instead."
```
shanmugamr1992: Please pass TransformerConfig directly to model_config

megatron/core/inference/engines/static_engine.py
santhnm2: Thanks, fixed

megatron/core/inference/engines/dynamic_engine.py
santhnm2: Thanks, fixed

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Fixed

examples/inference/gpt/gpt_dynamic_inference_with_coordinator.py
santhnm2: I see your point, but I think ideally we should have a `from_args` function somewhere or else we'll be back at the same place we're at now where we have to manually initialize the `DynamicInferenceContextConfig` in multiple places. Maybe one compromise option is we bring back `megatron/inference` as a folder that exists for any inference utilities that shouldn't live in `megatron/core/inference` (analogous to current `megatron/training`), and then put the factory function `get_dynamic_inference_context_config_from_args` there. What do you think?

santhnm2: We currently use `examples/inference/gpt` as the location of similar utility functions but I think it would be nice to have a more central place

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -748,25 +744,25 @@ def round_up_tokens(cls, value, tp_size=None):
...
-        unified_memory_level: int = 0,
-    ):
+    def from_model_and_args(cls, model, args, overrides: Optional[Dict[str, Any]] = None):
         """
-        Instantiate a `DynamicInferenceContext` from a `TransformerConfig` and an `InferenceWrapperConfig`.
+        Instantiate a `DynamicInferenceContext` from the model and args.
+
+        Args:
+            model: The Megatron model instance.
+            args: The arguments object.
```
lmcafee-nvidia: nit: add type hints for `model` and `args` ... but more importantly as someone reading this docstring, I think I'd be like "what the heck is this SimpleNamespace named args?!"

examples/inference/gpt/gpt_dynamic_inference_with_coordinator.py
santhnm2: Also is passing the `MegatronModule` directly into core still ok?

EDIT: It's defined in `megatron/core/transformer/module.py` so I think this should be fine

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Yea I think we can safely assume the only use was in the static engine

examples/inference/gpt/gpt_dynamic_inference_with_coordinator.py
lmcafee-nvidia: I like the idea of a utility method to do `context_config = from_args(args)` that lives outside of core. Great question about a `megatron/inference` directory ... it feels like `examples/inference` has become the de-facto place for non-core inference. I think it makes sense to start `megatron/inference`, and I'd imagine we can even move some sub-components of the example scripts in there.

santhnm2: Ok sounds good, I'll make these changes. Thanks!

santhnm2: cc @kvareddy @jaredcasper for viz

megatron/core/inference/model_inference_wrappers/abstract_model_inference_wrapper.py
```diff
@@ -121,23 +80,13 @@ def prep_model_for_inference(self, prompts_tokens: Optional[torch.Tensor] = None
 
         self.inference_context.reset()
 
-    @abc.abstractmethod
-    def prep_inference_input(self, prompt_tokens) -> Dict[str, Any]:
```
lmcafee-nvidia, line 125: `prep_inference_input()` was removed here, but not in `GPTInferenceWrapper` ... should it be removed there also?

examples/inference/gpt/gpt_dynamic_inference_with_coordinator.py
santhnm2: I have now added `DynamicInferenceConfig` and a `get_dynamic_inference_config_from_model_and_args` function in `megatron/inference/utils.py`

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -626,7 +582,7 @@ def _dynamic_step_forward_logits(self, input_ids: Tensor, position_ids: Tensor)
         if self.model_is_pipeline_parallel:
             logits_seq_len = (
                 active_request_count
-                if context.materialize_only_last_token_logits
+                if context.inference_config.materialize_only_last_token_logits
```
tdene: Grab this from `self.inference_config`

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -682,7 +638,7 @@ def _dynamic_step_sample_logits(self, logits: Tensor):
 
         # Last token logits.
         context = self.inference_wrapped_model.inference_context
-        if context.materialize_only_last_token_logits:
+        if context.inference_config.materialize_only_last_token_logits:
```
tdene: Grab this from `self.inference_config`

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -727,7 +683,7 @@ def _dynamic_step_calculate_log_probs(self, logits: Tensor) -> Optional[Tensor]:
         return context.calculate_log_probs(
             logits,
             self._sampled_tokens_cuda[:active_request_count],
-            only_last_token_logits=context.materialize_only_last_token_logits,
+            only_last_token_logits=context.inference_config.materialize_only_last_token_logits,
```
tdene: Grab this from `self.inference_config`

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -755,7 +711,7 @@ def _dynamic_step_calculate_top_n_logprobs(
         active_request_slice = slice(context.paused_request_count, context.total_request_count)
 
         # Handle decode-only mode (only last token)
-        if context.materialize_only_last_token_logits or context.is_decode_only():
+        if context.inference_config.materialize_only_last_token_logits or context.is_decode_only():
```
tdene: Grab this from `self.inference_config`

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -1231,7 +1187,7 @@ def generate_all_output_tokens_static_batch(
                     or not (sampling_params.return_log_probs or sampling_params.top_n_logprobs > 0)
                 )
                 inference_context = self.inference_wrapped_model.inference_context
-                inference_context.materialize_only_last_token_logits = (
+                inference_context.inference_config.materialize_only_last_token_logits = (
```
tdene: Grab this from `self.inference_config`

megatron/core/inference/config.py
```diff
@@ -0,0 +1,205 @@
...
+
+    request_metadata_types: Optional[List[Tuple[str, torch.dtype, bool]]] = None
+    """
+    A list of the per-request metadata types to track. Each entry is a tuple consisting of the string
+    label, the target dtype, and whether to store the data on GPU.
+    """
+
+
+@dataclass
+class StaticInferenceConfig(DynamicInferenceConfig):
```
lmcafee-nvidia: is there a plan to add fields to this class? If not, I wonder if we should just have a single `InferenceConfig`?

megatron/inference/utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+    # Get metrics writer if logging is enabled and on the logging rank
+    # Use the same rank convention as training (last rank logs)
+    metrics_writer = None
+    if (
+        inference_logging_step_interval > 0
+        and log_inference_wandb
+        and args.rank == (args.world_size - 1)
+    ):
+        metrics_writer = get_wandb_writer()
```
lmcafee-nvidia, line 268: the following things are missing their imports:
- `get_wandb_writer`
- `log_single_rank`
- `logging`

and `logger` needs to be defined with a `logger = logging.getLogger(...)` somewhere

megatron/inference/utils.py
```diff
@@ -0,0 +1,311 @@
...
+        # When using absolute position embeddings, it is critical that the
+        # context's `max_sequence_length` is less than or equal to the model's
+        # `max_sequence_length`. Otherwise, the context's `position_ids` will
+        # contain ids greater than the dimension of the position embedding
+        # tensor, which will result in an index error.
+        if inf_max_seq_len:
+            max_sequence_length = min(model_max_seq_len, inf_max_seq_len)
+        else:
+            max_sequence_length = model_max_seq_len
+        assert max_batch_size <= model_max_seq_len
```
lmcafee-nvidia: `max_batch_size` hasn't been defined yet

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -205,129 +183,38 @@ class DynamicInferenceContext(BaseInferenceContext):
...
-        use_flashinfer_fused_rope: bool = False,
-        unified_memory_level: Optional[int] = 0,
-        cuda_graph_max_tokens: Optional[int] = None,
-        cuda_graph_mixed_prefill_count: Optional[int] = 16,
-        metrics_writer: Optional['WandbModule'] = None,
-        request_metadata_types: Optional[List[Tuple[str, torch.dtype, bool]]] = None,
-        persist_cuda_graphs: Optional[bool] = False,
-    ):
-        super().__init__(materialize_only_last_token_logits=materialize_only_last_token_logits)
+    def __init__(self, model_config: TransformerConfig, inference_config: DynamicInferenceConfig):
```
lmcafee-nvidia: looks good!

megatron/core/inference/contexts/base_context.py
```diff
@@ -10,13 +12,11 @@ class BaseInferenceContext(abc.ABC):
...
 
-    def __init__(self, materialize_only_last_token_logits: bool):
+    def __init__(self, inference_config: BaseInferenceConfig):
         """
         Args:
-            materialize_only_last_token_logits (bool):
-                If True, only the last-token logits will be extracted during decode
         """
-        self.materialize_only_last_token_logits = materialize_only_last_token_logits
+        self.inference_config = inference_config
```
lmcafee-nvidia: I vote to shorten this to just `config`. I don't think we store `model_config` at all, and it's also a little weird in places where we have `inference_context.inference_config`

santhnm2: Done

megatron/inference/utils.py
santhnm2: Thanks, fixed

megatron/inference/utils.py
santhnm2: Thanks, fixed

megatron/core/inference/config.py
santhnm2: Just simplified to only having `InferenceConfig`

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: I ended up not saving `self.inference_config` since it wasn't used anywhere else. But per Lawrence's suggestion I have shortened this to `inference_context.config` instead of `inference_context.inference_config`

megatron/core/inference/model_inference_wrappers/abstract_model_inference_wrapper.py
santhnm2: Good catch, I don't think I should have removed it so I've brought it back now

santhnm2:
> Hey! This currently fails the test_rl_utils.py due to imports problems:
> 
> ```
>     from megatron.inference import get_dynamic_inference_engine
> ImportError: cannot import name 'get_dynamic_inference_engine' from 'megatron.inference' (/lustre/fsw/portfolios/llmservice/projects/llmservice_nlp_fm/data-quality
> /yobibyte/dev/mrl_internal/megatron-rl/megatron/inference/__init__.py)
> ```
> 
> The same error is when I run internal functional tests.

Thanks for pointing this out! I just pushed a fix, testing now

santhnm2:
@yobibyte could you take another look at this when you get a chance?

yobibyte:
> @yobibyte could you take another look at this when you get a chance?

Hey @santhnm2 , the unit tests are passing now, thanks for fixing!

Can you, please, update the `--inference-max-batch-size` flag that was renamed? IIRC all the usage is in the examples/rl/model_configs scripts, but, please, grep to be on the safe side.

PR #2963, opened 2026-01-15
cp: `ci(fix): GB200 racecondition (2962)` into `main`

ko3n1g:
We're repeating tests 5 times. each loop is:
* delete old logs;
* training (that generates logs)
* evaluate logs;

Since GB testing happens on two nodes, one of the node was now finishing the evaluation faster than the other one. the faster node already deleted logs as part of the next loop before the slower node started evaluating


ko3n1g:
/ok to test [9c44101](https://github.com/NVIDIA/Megatron-LM/pull/2963/commits/9c441010496e8a1e7a09db6c18b971369f68c650)

PR #2966, opened 2026-01-15
Multiturn rollout support prep

yobibyte:
# What does this PR do ?

This is a first PR to prep for the full multiturn rollout support in megatron. We treat each turn here as it's own trajectory, each turn will share the full rollout reward.

- Main API change: Rollout/TokenRollout contains a list of trajs/logprobs/generation mask (one rollout has multiple turns).
- RL utils are ready to deal with multiturn, including sequence packing:
  - adjusted num microbatches per step
  - proper advantage computation (not to skew stats by averaging over the same turn rewards)
- sequence packing:
  - refactored the batch calculator logic that is much cleaner and testable now.
  - fixed a bug that resulted in wrong schedule for data sampling in the cases when we generate more data than we consume in a single step
- Added multiple unit test for sequence packing logic and multiturn traj prep.

### DIFF
--- a/tests/unit_tests/rl/test_rl_utils.py
+++ b/tests/unit_tests/rl/test_rl_utils.py
@@ -273,15 +273,16 @@ class TestRLUtils:
         """Test that getting logprobs at least does not crash."""
         world_size, dp, tp, pp = initialize_model_parallel
         # Here I assume that we will be consuming all data in one step.
+        group_size = 2
         self.create_test_args(
             micro_batch_size=2,
             seq_length=4,
             curr_iteration=1,
             tensor_model_parallel_size=tp,
             pipeline_model_parallel_size=pp,
-            global_batch_size=Utils.world_size * 2,
-            grpo_prompts_per_step=Utils.world_size,
-            grpo_group_size=2,
+            global_batch_size=dp * 2,
+            grpo_prompts_per_step=dp,
+            grpo_group_size=group_size,
         )
 
         model = MockModel()
@@ -304,7 +305,7 @@ class TestRLUtils:
             problem_id="2",
         )
 
-        rollouts = [[r1, r2] for _ in range(Utils.world_size)]
+        rollouts = [[r1, r2] for _ in range(dp)]
         try:
             rl_utils.prepare_data_for_update(
                 [model], {}, rollouts, tokenizer, sequence_packing=False, is_correction=False
@@ -329,7 +330,7 @@ class TestRLUtils:
             env_id='MEGAENV',
             problem_id="2",
         )
-        rollouts = [[r1, r2] for _ in range(Utils.world_size)]
+        rollouts = [[r1, r2] for _ in range(dp)]
         data_iter = rl_utils.prepare_data_for_update(
             [model], {}, rollouts, tokenizer, sequence_packing=False, is_correction=False
         )


megatron/rl/agent/reward_only_agent.py
```diff
@@ -104,16 +104,16 @@ async def rollout_from_response(
                 for x in range(len(response.token_ids))
             ]
             rollout = TokenRollout(
-                trajectory=response.token_ids,
+                trajectory=[response.token_ids],
```
yobibyte, line 107: reward_only_agent stuff stays inherently single-turn, wrapping vars in lists is to conform to the new API, as single-turn is a specific case of the multi-turn.

megatron/rl/rl_utils.py
```diff
@@ -964,11 +962,92 @@ def prepare_trajectories(
...
+            sequence_packing=b_packed_seq_params is not None, 
+            packed_seq_params=b_packed_seq_params,
+        ),
+        None,
+    )
+    model.train()
+    return logprobs
+
+
+def _compute_logprobs_batch(
```
jalbericiola, line 1007: cleaner in this way!

megatron/rl/rl_utils.py
```diff
@@ -1690,3 +1734,32 @@ def get_iteration_sequence_count(args):
     if torch.distributed.is_initialized():
         torch.distributed.all_reduce(sequences_tensor, group=mpu.get_data_parallel_group())
     return int(sequences_tensor.item())
+    
+def _pad_nonnull_with_zeros(data: list[Optional[torch.Tensor]], max_len: int) -> torch.Tensor:
```
jalbericiola, line 1751: Couldn't we have this function as
   `torch.stack([F.pad(t, (0, max_len - t.size(0))) for t in data if t is not None])`

megatron/rl/sequence_packing_utils.py
```diff
@@ -969,33 +967,20 @@ def distribute_packed_bins(
...
-        # Gather all generation masks
-        generation_masks = generation_masks.cuda()
-        masks_list = [torch.empty_like(generation_masks) for _ in range(data_parallel_world_size)]
-        torch.distributed.all_gather(
-            masks_list, generation_masks, group=mpu.get_data_parallel_group()
-        )
-        generation_masks = torch.cat(masks_list, dim=0)
-
-        # Gather inference logprobs if present
+        def _gather(data):
```
jalbericiola, line 974: nice

megatron/rl/sequence_packing_utils.py
```diff
@@ -969,33 +967,20 @@ def distribute_packed_bins(
...
+        trajs = _gather(trajs)    
+        generation_masks = _gather(generation_masks) 
         if inference_logprobs is not None:
-            inference_logprobs = inference_logprobs.cuda()
-            logprobs_list = [torch.empty_like(inference_logprobs) for _ in range(data_parallel_world_size)]
-            torch.distributed.all_gather(
-                logprobs_list, inference_logprobs, group=mpu.get_data_parallel_group()
-            )
-            inference_logprobs = torch.cat(logprobs_list, dim=0)
+            inference_logprobs = _gather(inference_logprobs)
```
jalbericiola, line 983: nice

megatron/rl/rl_utils.py
```diff
@@ -998,59 +1079,50 @@ def prepare_data_for_update(
...
+
+        # Let's expand rollouts getting rid of the groups.
+        # We need this to correctly split the rollouts across dp groups.
+        # And we do not actually need them grouped in anything below anyways.
+        rollouts = [r for g in rollouts for r in g]
+        total_turns_sampled = len(rollouts)
+
+        # We might sample more than we consume in one step.
+        samples_ratio_per_step = args.global_batch_size // (args.grpo_prompts_per_step * args.grpo_group_size)
+        assert samples_ratio_per_step <= 1, "You cannot use more data than you sampled."
```
jalbericiola, line 1113: Should be fine after changing the type of division. Right now it'll be always 0

megatron/rl/rl_utils.py
```diff
@@ -998,59 +1079,50 @@ def prepare_data_for_update(
...
+        example_group = rollouts[0]
+
+        # Let's expand rollouts getting rid of the groups.
+        # We need this to correctly split the rollouts across dp groups.
+        # And we do not actually need them grouped in anything below anyways.
+        rollouts = [r for g in rollouts for r in g]
+        total_turns_sampled = len(rollouts)
+
+        # We might sample more than we consume in one step.
+        samples_ratio_per_step = args.global_batch_size // (args.grpo_prompts_per_step * args.grpo_group_size)
```
jalbericiola: I think it has to be normal division instead of integer

tests/unit_tests/rl/test_sequence_packing_utils.py
```diff
@@ -407,3 +410,56 @@ def test_compute_packed_inference_logprobs_stats_shape_mismatch():
...
+        (
+            0.5,
+            1,
+            8,
+            8,
+        ),  # 0.5 means we use half of all seqs per step, they all fit 1 bin -> we should reuse
+        (1 / 3, 4, 8, 16),  # third of the data per step, nonint division
+    ],
+)
+def test_get_bins_bs_and_steps(ratio, local_bins, world, expected_bs):
```
jalbericiola, line 429: This test is passing because we patch the values of the function.
But the ones for prepare_data_for_update shouldn't be passing?

megatron/rl/rl_utils.py
yobibyte: That is a great catch, thank you!

megatron/rl/rl_utils.py
yobibyte: Now when division is non-integer, the assert might fire.

tests/unit_tests/rl/test_sequence_packing_utils.py
yobibyte: Yeah, that test is currently skipped and I did not catch the bug you saw.

megatron/rl/rl_utils.py
yobibyte: I do not think that your code is equivalent to what we are doing there. E.g. if the chunk (or t in your case) is None, we append zeros.

jalbericiola: You are right, forgot the else, it should be:
if all(el is None for el in data):
        raise ValueError("At least one element of the data list should be not None.")
    return torch.stack([F.pad(t, (0, max_len - len(t))) if t is not None else torch.zeros(max_len) for t in data])

yobibyte: This also does not have a check for max len overflow. I would rather keep it as is, tbh, unless you do not like the current version too much.

jalbericiola:
LGTM after the unit test in progress is included

megatron/rl/sequence_packing_utils.py
```diff
@@ -1073,62 +1058,66 @@ def pack_all_trajectories(trajs, generation_masks, inference_logprobs, global_ad
 
     return packing_context
 
+def update_microbatch_calculator(
+    samples_ratio_per_step: float,
```
ArEsKay3, line 1062: Can we add a comment explaining what this is?

yobibyte: Added a docstring to the function.

PR #2979, opened 2026-01-16
Feature/amem nccl rl

CodersAcademy006:
### Motivation

Reinforcement learning workflows in Megatron-LM often alternate between training and inference-heavy rollout phases. During rollout, a significant amount of GPU memory can be consumed by NCCL-allocated buffers that are idle from a compute perspective but cannot be reclaimed.

This PR introduces optional support for the AMem NCCL plugin to allow NCCL-allocated GPU memory to be transparently offloaded during RL rollout phases and restored when needed. The goal is to reduce peak GPU memory pressure during inference-dominated steps without changing model semantics or training behavior.

### What this PR does

- Adds an optional AMem NCCL integration for RL training paths.
- Introduces explicit hooks to pause NCCL memory usage during rollout and resume it before training resumes.
- Integrates with the native RL execution path (`perform-rl-step`).
- Exposes opt-in CLI flags to control behavior:
  - Enable or disable AMem NCCL support
  - Configure AMem process group ID
  - Control whether offloading occurs during rollout
- Keeps the feature fully disabled by default and backward compatible.

### Design principles

- **Opt-in only**: No behavior changes unless explicitly enabled.
- **No semantic changes**: Model execution, gradients, and optimizer behavior remain unchanged.
- **Minimal surface area**: Reuses existing RL control flow and initialization paths.
- **Fail-safe**: If AMem is unavailable or misconfigured, training proceeds normally without offloading.

### Expected impact

- Reduced peak GPU memory usage during RL rollout and inference phases.
- Increased memory headroom for larger batch sizes or longer sequences in RL workloads.
- No impact on non-RL training paths.

### Scope limitations

- This PR does not introduce new checkpointing, scheduling, or allocator behavior.
- This PR does not generalize AMem usage beyond RL workflows.
- Performance characteristics depend on the underlying AMem NCCL implementation and system configuration.

### Testing

- Validated on RL training flows with AMem enabled and disabled.
- Confirmed that disabling AMem restores existing behavior.
- Exercised error handling paths when AMem is unavailable.

### DIFF
--- a/DESIGN_AMEM_NCCL.md
+++ b/DESIGN_AMEM_NCCL.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-This PR integrates optional AMem (Asynchronous Memory) NCCL-based offloading support for Megatron RL training workflows. The feature enables offloading GPU memory to CPU/host memory during RL rollout phases to reduce GPU memory pressure.
+This PR integrates optional AMem (Asynchronous Memory) NCCL-based offloading support for Megatron RL training workflows. The feature enables releasing NCCL-allocated communicator buffers during RL rollout phases to reduce GPU memory pressure.
 
 ## When AMem Hooks Are Enabled
 
@@ -23,24 +23,28 @@ During rollout:
 - Large batches of sequences are generated
 - Activation tensors accumulate rapidly
 - GPU memory becomes a bottleneck before compute saturation
-- Model parameters can be temporarily offloaded without performance penalty
+- NCCL buffer memory (communicator allocations) is idle during rollout generation
 
 AMem enables:
-- Offloading idle model parameters during rollout
-- Fetching them back asynchronously before training phase
-- Overlapping CPU-GPU transfers with rollout computation
+- Releasing GPU memory allocated by NCCL (communication buffers)
+- Restoring NCCL memory before training collective operations resume
+- Reducing peak GPU memory during rollout phases where NCCL buffers are otherwise idle
 
 ## What Memory Is Offloaded
 
-When enabled, AMem NCCL offloads:
-- **Model parameters** during rollout generation
-- **Optimizer states** (optional, based on configuration)
-- **Gradient buffers** when not actively being computed
+AMem enables:
+- Offloading GPU memory allocated by NCCL (communication buffers)
+- Restoring NCCL memory before training collective operations resume
+- Reducing peak GPU memory during rollout phases where NCCL buffers are otherwise idle
 
-Explicitly **not offloaded**:
-- Active computation tensors
-- Attention KV caches during generation
-- Critical activations needed for current forward pass
+AMem does NOT offload:
+- Model parameters
+- Optimizer states
+- Gradient buffers
+- Activations
+
+Parameter and optimizer offloading are handled separately via
+`--rl-offload-optimizer-during-inference` and related flags.
 
 ## Implementation Details
 
--- a/megatron/rl/rl_utils.py
+++ b/megatron/rl/rl_utils.py
@@ -1829,16 +1829,21 @@ def megatron_rl_inference_mode(
     loop = get_asyncio_loop()
     nvtx_range = get_nvtx_range()
 
-    amem_offload_during_rollout = getattr(args, 'rl_amem_offload_during_rollout', True)
+    amem_offload_during_rollout = getattr(args, 'rl_amem_offload_during_rollout', False)
+
+    # Resolve use_amem here, not lazily
+    use_amem = False
+    amem_nccl = None
     if amem_offload_during_rollout:
         try:
             from megatron.core import amem_nccl
-            if amem_nccl.is_amem_available():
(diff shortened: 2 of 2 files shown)


HaochenYuan:
The changes in `megatron/core/transformer/moe/grouped_gemm_util.py` don't seem to be related to `amem`.

CodersAcademy006:
Hey @HaochenYuan, I have made the relevant changes. Please tell me if anything else has to be changed. Thank you for the review!!

Also i have made a follow-up pr for the rest of the changes #2988 please review that as well.

megatron/rl/rl_utils.py
```diff
@@ -1654,6 +1673,14 @@ def megatron_rl_inference_mode(
         if cuda_graph_impl != "none" and not args.rl_training_cuda_graphs:
             toggle_cuda_graphs(lang_module, 'none', reset_cuda_graphs=reset_cuda_graphs)
 
+        # Restore NCCL memory after inference if AMem is enabled
+        if use_amem and amem_offload_during_rollout:
```
HaochenYuan, line 1914: Where is `use_amem` defined?

megatron/rl/rl_utils.py
```diff
@@ -1601,6 +1612,14 @@ def megatron_rl_inference_mode(
...
                 optimizer.offload_to_cpu()
 
+        # Offload NCCL memory before inference if AMem is enabled
+        if use_amem and amem_offload_during_rollout:
+            with nvtx_range("amem-nccl-pause-before-inference"):
+                if amem_nccl.nccl_pause():
+                    logger.info(f"[{dist.get_rank()}:DP] Successfully offloaded NCCL memory")
+                else:
+                    logger.warning(f"[{dist.get_rank()}:DP] Failed to offload NCCL memory")
+
```
HaochenYuan, line 1895: The inference process seems to be the same as training's, if you offload nccl memory before inference while the inference process also needs the comm groups, there may be runtime error.

DESIGN_AMEM_NCCL.md
```diff
@@ -0,0 +1,92 @@
...
+2. **Training Phase**: Update model parameters (compute-heavy)
+
+During rollout:
+- Large batches of sequences are generated
+- Activation tensors accumulate rapidly
+- GPU memory becomes a bottleneck before compute saturation
+- Model parameters can be temporarily offloaded without performance penalty
+
+AMem enables:
+- Offloading idle model parameters during rollout
```
HaochenYuan: Amem cannot offload parameters, it can only offload memory used by nccl.

HaochenYuan:
Before proceeding, could you confirm if you have tested this change with a compiled amem library?

During some earlier beta testing, I observed that amem is currently only able to release NCCL memory for communication groups that are created via the loaded `libamem_nccl.so libnccl.so`. However, in Megatron, most communication groups are created directly through torch.distributed.new_group, which may fall outside its current memory release scope.

To better evaluate this PR, it would be very helpful if you could share:

- The released memory size observed with specified model configuration and parallel setup used for testing/
- If you have tested it , can you provide the `libamem_nccl.so libnccl.so` & tested environment config?

Thx!

CodersAcademy006:
@HaochenYuan — thanks for the detailed feedback. I've tested the fallback 
paths (AMem unavailable, env vars not set, graceful degradation) and the 
argument parsing, but I don't currently have access to hardware that supports 
NCCL_CUMEM_ENABLE to validate the full integration against a compiled 
libamem_nccl.so.

Given your point about torch.distributed.new_group comm groups potentially 
falling outside AMem's release scope, I'd like to understand that constraint 
better before pushing further. Would it make sense to convert this to a draft 
PR until I can get access to validate it properly, or is there a test 
environment within NVIDIA I could coordinate with to verify this?

I'll fix the design doc regardless — AMem offloads NCCL memory, not model 
parameters. That was wrong.

Sorry for being late.

HaochenYuan:
> @HaochenYuan — thanks for the detailed feedback. I've tested the fallback paths (AMem unavailable, env vars not set, graceful degradation) and the argument parsing, but I don't currently have access to hardware that supports NCCL_CUMEM_ENABLE to validate the full integration against a compiled libamem_nccl.so.
> 
> Given your point about torch.distributed.new_group comm groups potentially falling outside AMem's release scope, I'd like to understand that constraint better before pushing further. Would it make sense to convert this to a draft PR until I can get access to validate it properly, or is there a test environment within NVIDIA I could coordinate with to verify this?
> 
> I'll fix the design doc regardless — AMem offloads NCCL memory, not model parameters. That was wrong.
> 
> Sorry for being late.

Hi @CodersAcademy006, since [NCCL v2.29.7 release](https://github.com/NVIDIA/nccl/releases) has supported dynamic memory offload, the Amem integration maybe not be necessary now. Thx for your contribution!

CodersAcademy006:
@HaochenYuan thanks for your explanation and review. If there is any other issue to work on, i would be happy to take that. Thank You.

PR #3029, opened 2026-01-21
Move tensor offload/onload out of RL code

tdene:
# What does this PR do ?

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/tests/unit_tests/inference/engines/test_dynamic_engine.py
+++ b/tests/unit_tests/inference/engines/test_dynamic_engine.py
@@ -1682,10 +1682,7 @@ class TestDynamicInferenceEngine:
         # Suspend.
         engine.suspend()
         assert engine.is_suspended
-        if deallocates:
-            assert not context.is_tensor_state_allocated
-        else:
-            assert context.is_tensor_state_allocated
+        assert not context.is_tensor_state_allocated
 
         gc.collect()
         torch.cuda.empty_cache()


megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -717,13 +729,17 @@ def deallocate_all_tensors(self):
             return
         self.is_tensor_state_allocated = False
 
-        # Delete all tensor attributes.
+        # Delete or offload all tensor attributes.
         # TODO(@lmcafee): check that device == 'cuda'?
```
tdene: This TO-DO becomes vital with this PR. If we are allowing for offloading of GPU tensors here, we specifically need to check whether a to-be-offloaded tensor lives on GPU.

Will be fixed in a follow-up commit on this PR.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -433,6 +433,19 @@ def __init__(
                         "Unified memory requested but not available; defaulting to GPU memory."
                     )
                 self.unified_memory_level = 0
+        self.persist_cuda_graphs = persist_cuda_graphs
```
yobibyte: nit: I would keep this line where it was in case, in the future, someone adds a function call inside the if branch above that could depend on the self.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -433,6 +433,19 @@ def __init__(
...
+        if self.persist_cuda_graphs and self.unified_memory_level == 0:
+            assert (
+                HAVE_TORCH_MEMORY_SAVER
+            ), "Can only persist CUDA graphs without UVM if torch_memory_saver is available"
+        self.offload_kv_cache = offload_kv_cache
```
yobibyte: Same with these folks.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -433,6 +433,19 @@ def __init__(
...
+        self.remove_kv_cache = remove_kv_cache
+        assert not (offload_kv_cache and remove_kv_cache), "Cannot both offload and remove KV$."
+        # For tensor offload via storage resize trick (when not using torch_memory_saver)
```
yobibyte: Add link to the PR?

tests/functional_tests/test_cases/gpt/gpt_grpo_tp4_pp1_dp2_8b_suspend_offload_kv/model_config.yaml
```diff
@@ -0,0 +1,104 @@
...
+  --empty-unused-memory-level: 2
+  --seed: 42
+  --calculate-per-token-loss: true
+  --rl-use-sequence-packing: true
+  --rl-sequence-packing-algo: fifo
+  --rl-offload-optimizer-during-inference: true
+  --timing-log-level: 1
+  --cuda-graph-impl: local
+  --micro-batch-size: 1
+  --global-batch-size: 2
```
yobibyte: The global batch size should be group_size * prompts_per_step to avoid the error we have before and non-uniform per-step lengths. This way, you will be sampling every second iteration.

tests/functional_tests/test_cases/gpt/gpt_grpo_tp4_pp1_dp2_8b_suspend_remove_kv/model_config.yaml
```diff
@@ -0,0 +1,105 @@
...
+  --empty-unused-memory-level: 2
+  --seed: 42
+  --calculate-per-token-loss: true
+  --rl-use-sequence-packing: true
+  --rl-sequence-packing-algo: fifo
+  --rl-offload-optimizer-during-inference: true
+  --timing-log-level: 1
+  --cuda-graph-impl: local
+  --micro-batch-size: 1
+  --global-batch-size: 2
```
yobibyte: Same here.

tests/unit_tests/rl/test_rl_suspend.py
```diff
@@ -0,0 +1,1782 @@
...
+            remove_kv_cache=True,
+            unified_memory_level=0,
+            num_cuda_graphs=num_cuda_graphs,
+        )
+        enable_cuda_graph = num_cuda_graphs is not None and num_cuda_graphs > 0
+
+        engine = self._build_engine(config, enable_cuda_graph=enable_cuda_graph)
+
+        # Check initial CUDA graph state
+        initial_cuda_graph_count = len(_CudagraphGlobalRecord.cudagraph_record)
```
yobibyte: What do you actually check by this?

tests/unit_tests/rl/test_rl_suspend.py
```diff
@@ -0,0 +1,1782 @@
...
+
+            # Resume
+            engine.resume()
+            assert not engine.is_suspended
+
+    @pytest.mark.internal
+    @pytest.mark.skipif(
+        not is_fa_min_version("2.7.3"), reason="need latest flash attn for dynamic batching"
+    )
+    @pytest.mark.parametrize("num_cuda_graphs", [None, 1, 2])
```
yobibyte: Why do we need 1 and 2 here?

tests/unit_tests/rl/test_rl_suspend.py
```diff
@@ -0,0 +1,1782 @@
...
+            cycle_memories.append(
+                {'with_requests': memory_with_requests, 'after_cycle': memory_after_cycle}
+            )
+
+        # Check that memory is roughly consistent across cycles (allowing 10% variance)
+        for i in range(1, len(cycle_memories)):
+            prev = cycle_memories[i - 1]['after_cycle']
+            curr = cycle_memories[i]['after_cycle']
+            variance = abs(curr - prev) / max(prev, 1)
+            assert variance < 0.1, (
```
yobibyte: How did you pick 10%? Potentially flaky.

megatron/rl/inference/megatron.py
```diff
@@ -304,11 +305,25 @@ async def suspend(self):
...
         await self._inference_engine.paused.wait()
+        self._inference_engine.suspend()
+
+    def mark_for_recompute(self):
+        """Mark all request entries for KV cache recomputation.
+
+        Called before resume when using partial rollouts with KV cache removal,
+        so the engine knows to recompute the KV cache for all in-flight requests.
+        """
+        args = get_args()
```
yobibyte: I would personally turned those into class fields and only used get_args() in the constructor of a class. This spread of get_args() everywhere should stop.

megatron/rl/inference/megatron.py
```diff
@@ -304,11 +305,25 @@ async def suspend(self):
         if dist.get_rank() == 0:
             await self._client.pause_engines()
         await self._inference_engine.paused.wait()
+        self._inference_engine.suspend()
+
+    def mark_for_recompute(self):
+        """Mark all request entries for KV cache recomputation.
+
+        Called before resume when using partial rollouts with KV cache removal,
```
yobibyte: If we call it before resume(), do we actually need it as a separate method?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -116,6 +116,7 @@ class RequestEntry:
 
     record: DynamicInferenceRequestRecord
     future: asyncio.Future
+    recompute_soon: bool = False
```
yobibyte: Please, add a comment what it means, or (better), rename the variable so that it is clear what 'soon' means.

megatron/rl/rl_utils.py
```diff
@@ -1587,7 +1588,6 @@ def megatron_rl_inference_mode(
         None: this context manager does not return a value.
 
     """
-    args = get_args()
```
yobibyte, line 1675: NICE

megatron/core/transformer/utils.py
```diff
@@ -391,7 +391,7 @@ def find_modules_with_attrs(module):
     find_modules_with_attrs(model_modules)
 
 
-def toggle_cuda_graphs(model, set_to="none", reset_cuda_graphs=True):
+def toggle_cuda_graphs(model, set_to="none", reset_cuda_graphs=False):
```
yobibyte: Why change the default?

megatron/training/arguments.py
```diff
@@ -327,29 +327,97 @@ def validate_args(args, defaults={}):
...
+        #    ** With `--rl-partial-rollouts`, it expects the engine to resume its state.
+        #  - Resuming the engine makes the context restore/rebuild the KV$.
+        #    ** With `--rl-offload-kv-cache-during-training`, the KV$ is reloaded from CPU memory.
+        #    ** With `--rl-remove-kv-cache-during-training`, the KV$ is reallocated.
+        #    ** With neither argument, it no-ops, leaving the KV$ as-is.
+        #  - The engine proceeds to recompute the KV$ of any request records marked as "recompute".
+        #    ** If using both partial rollouts and KV$ removal, "recompute" must mark all requests.
+        # ------------------------------------------------
+
+        # TODO: Add support for `torch_memory_saver` with async model weight offload.
```
yobibyte: TODO(tdene)?

megatron/training/arguments.py
```diff
@@ -327,29 +327,97 @@ def validate_args(args, defaults={}):
...
+        # We just assert that every single request record is marked for recomputation.
+        # ------------------------------------------------
+        # CGs are handled as follows:
+        #  - The inference engine's init builds inference CGs, if desired.
+        #    ** This is controlled by `--cuda-graph-impl`.
+        #  - Training is graphed on the first training step, if desired.
+        #    ** This is controlled by `--rl-training-cuda-graphs`.
+        #  - `megatron_rl_inference_mode` inside rl_utils toggles CGs on/off
+        #      if inference is graphed but training is not.
+        #      Only the RL loop is aware that there's both training and inference.
```
yobibyte: What does it mean? 
- What is 'the RL loop'?
- What do you want to say by 'only the loop is aware'?

megatron/training/arguments.py
```diff
@@ -327,29 +327,97 @@ def validate_args(args, defaults={}):
...
+        if args.rl_offload_kv_cache_during_training:
+            if args.inference_dynamic_batching_unified_memory_level > 0:
+                warn_rank_0(
+                    "--inference-dynamic-batching-unified-memory-level and "
+                    "--rl-offload-kv-cache-during-training are both set. "
+                    "This will not cause issues currently, but it is not recommended or supported."
+                )
+        # We definitely cannot remove the KV$ if we are offloading it.
+            assert not args.rl_remove_kv_cache_during_training, (
+                "Cannot use both --rl-remove-kv-cache-during-training and "
```
yobibyte: Should we turn this into --rl-kv-cache-train-mode and make it accept 'noop', 'remove', 'offload'? We won't need to check the validity in this case.

megatron/training/arguments.py
```diff
@@ -1914,6 +1982,9 @@ def _add_rl_args(parser):
                        help='Remove KV cache during training to save GPU memory')
     group.add_argument('--rl-reset-cuda-graphs', action=argparse.BooleanOptionalAction, type=bool, default=False,
                        help='Reset CUDA graphs between inference/training to save GPU memory')
+    group.add_argument('--rl-training-cuda-graphs', action=argparse.BooleanOptionalAction, type=bool,
+                       default=False,
+                       help='If set, do not call `delete_cuda_graphs` or `toggle_cuda_graphs` when the inference engine is suspended.')
```
yobibyte: This message is confusing. Doesn't it just mean 'use cuda graphs for train side'? This message reads, that if I use this flag, I should somehow avoid removing/toggling the cuda graphs. 

If you describe the implementation detail, probably we need to move this explanation to the actual function that does this. As a user, I do not care about this.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -734,25 +790,31 @@ def deallocate_all_tensors(self):
...
-        if self.unified_memory_level != 0:
-            return
-
         # Mark deallocated.
         if not self.is_tensor_state_allocated:
             return
         self.is_tensor_state_allocated = False
 
-        # Delete all tensor attributes.
-        # TODO(@lmcafee): check that device == 'cuda'?
```
yobibyte, line 722: Was this an important TODO?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -734,25 +790,31 @@ def deallocate_all_tensors(self):
...
-            value = getattr(self, key)
-            if isinstance(value, torch.Tensor):
-                delattr(self, key)
+        # UVM and `torch_memory_saver` do not require explicit deallocation.
+        if self.unified_memory_level != 0:
+            return
+        if self.persist_cuda_graphs and HAVE_TORCH_MEMORY_SAVER:
+            torch_memory_saver.backup_region(tag="inference_context")
+            return
+        # If offloading manually, copy tracked tensors to CPU and resize storage to 0.
```
yobibyte: I do not like this comment as it describes the internal implementation, the implementation might change in the future if resize(0) starts behaving differently.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -734,25 +790,31 @@ def deallocate_all_tensors(self):
...
+        if self.offload_kv_cache:
+            self._offload_tracked_tensors_to_cpu()
+
+        # Delete tensor attributes if explicitly deallocating.
+        if self.remove_kv_cache:
+            keys = list(vars(self).keys())
```
yobibyte: nit: inline

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -734,25 +790,31 @@ def deallocate_all_tensors(self):
...
+            for key in keys:
+                value = getattr(self, key)
+                if isinstance(value, torch.Tensor):
+                    # Keep tracked tensors if offloading (mutually exclusive with remove, but here for clarity)
+                    if self.offload_kv_cache and key in self._offloadable_tensor_names:
+                        continue
+                    delattr(self, key)
```
yobibyte, line 776: Is it actually safe to do? Imagine I create an object and expect it to have some attributes. Then, somewhere in my code I run `if object.attribute_i_want is None`, this will throw an error, right?

tests/unit_tests/rl/test_rl_suspend.py
```diff
@@ -0,0 +1,1782 @@
...
+        initial_memory = get_gpu_memory_allocated()
+
+        # Deallocate tensors
+        context.deallocate_all_tensors()
+
+        assert not context.is_tensor_state_allocated
+
+        # Memory should not change because deallocate is a no-op without remove_kv_cache
+        # (when not using UVM or torch_memory_saver and not explicitly removing)
+        current_memory = get_gpu_memory_allocated()
```
yobibyte: I don't think we check anything here.

tests/unit_tests/rl/test_rl_suspend.py
```diff
@@ -0,0 +1,1782 @@
...
+            assert context.is_tensor_state_allocated
+
+            # Reset context for next cycle
+            context.reset()
+
+    @pytest.mark.internal
+    @pytest.mark.skipif(
+        not is_fa_min_version("2.7.3"), reason="need latest flash attn for dynamic batching"
+    )
+    def test_context_deallocate_idempotent(self):
```
yobibyte: Are we checking that it won't crash here or what?

tests/unit_tests/rl/test_rl_suspend.py
```diff
@@ -0,0 +1,1782 @@
...
+            persist_cuda_graphs=False,
+            offload_kv_cache=False,
+            remove_kv_cache=True,
+            unified_memory_level=0,
+        )
+        engine = self._build_engine(config, enable_cuda_graph=False)
+
+        # Add requests
+        requests = self._add_requests_to_engine(engine, config)
+        original_request_ids = set(engine.requests.keys())
```
yobibyte: I don't think you need to convert everything to sets. https://docs.python.org/3/library/stdtypes.html#dict-views

tests/unit_tests/rl/test_rl_suspend.py
```diff
@@ -0,0 +1,1782 @@
...
+        Utils.destroy_model_parallel()
+        force_gc()
+
+    @pytest.mark.internal
+    @pytest.mark.skipif(
+        not is_fa_min_version("2.7.3"), reason="need latest flash attn for dynamic batching"
+    )
+    def test_mark_for_recompute_basic(self):
+        """Test basic mark_for_recompute functionality on request entries."""
+        from megatron.core.inference.engines.dynamic_engine import RequestEntry
```
yobibyte: Please, move imports to the top of the file.

tests/unit_tests/rl/test_rl_suspend.py
```diff
@@ -0,0 +1,1782 @@
...
+        # Create a mock request and entry
+        request = DynamicInferenceRequest(
+            request_id=0,
+            prompt_tokens=torch.tensor([1, 2, 3], device='cuda'),
+            sampling_params=SamplingParams(num_tokens_to_generate=5),
+        )
+        record = DynamicInferenceRequestRecord.from_request(request)
+        entry = RequestEntry(record=record, future=asyncio.Future())
+
+        # Initially recompute_soon should be False
```
yobibyte: This test looks nonsensical to me. Why do we check the attributes we just set?

megatron/training/arguments.py
tdene: Yeah, I've been wanting to. I will make that commit inside this PR in a bit, and see what other reviewers think.

megatron/training/arguments.py
tdene: TODO(peter) currently, but I'm not sure who will get around to it first.

wdykas: https://github.com/NVIDIA/Megatron-LM/pull/3170 you mean this?

megatron/training/arguments.py
tdene: Hm, what I'm trying to point out here is that toggling the CG on/off needs to happen outside of inference, while deleting/creating the CG needs to happen inside of inference. My wording's not good, I will try again.

megatron/core/transformer/utils.py
tdene: We no longer use `reset_cuda_graphs=True` anywhere, because as I argue in that section of the big comment that you asked about, deleting cuda graphs is now only done at the inference level and never done up in the RL abstraction layer.

Since we never use `reset_cuda_graphs=True` anywhere, it seems bad to have the default be set to that.

megatron/core/inference/contexts/dynamic_context.py
tdene: Addressed

megatron/training/arguments.py
tdene: Addressed!

tests/functional_tests/test_cases/gpt/gpt_grpo_tp4_pp1_dp2_8b_suspend_offload_kv/model_config.yaml
tdene: Addressed!

tests/functional_tests/test_cases/gpt/gpt_grpo_tp4_pp1_dp2_8b_suspend_remove_kv/model_config.yaml
tdene: Addressed!

megatron/core/inference/contexts/dynamic_context.py
tdene: Addressed

megatron/core/inference/contexts/dynamic_context.py
tdene: I think I addressed this, let me know

megatron/core/inference/contexts/dynamic_context.py
tdene: ach, I was trying to intentionally preserve it, but I lost it in some commit. Brought it back.

Addressed.

megatron/core/inference/contexts/dynamic_context.py
tdene: Addressed

megatron/core/inference/contexts/dynamic_context.py
tdene: Addressed

megatron/core/inference/contexts/dynamic_context.py
tdene: True, it will throw an error. But attempting to use the engine or context in this state is already illegal. In the engine there is an explicit `EngineSuspendedError` raised when doing this illegal action; the context is not supposed to ever be called directly by the user, it should only be called through the engine, so the user will encounter `EngineSuspendedError` first before any AttributeError can kick in.

megatron/rl/inference/megatron.py
tdene: Addressed

megatron/core/inference/engines/dynamic_engine.py
tdene: Addressed

megatron/rl/inference/megatron.py
tdene: The idea is that `mark_for_recompute` will be the method where we decide how stale the KV$ is allowed to get before we recompute it, which means that it will grow to have complicated logic. I'm making it a separate method early so that it does not pollute the codebase down the road.

tests/test_utils/recipes/h100/gpt-grpo.yaml
```diff
@@ -79,3 +79,13 @@ products:
       - environment: [dev]
         scope: [mr-broken]
         platforms: [dgx_h100]
+  - test_case: [gpt_grpo_tp4_pp1_dp2_8b_suspend_remove_kv]
```
jon-barker: Could we not test suspend and remove/offload KV in a unit test rather than a functional test?

megatron/training/arguments.py
tdene: Yes, exactly.

tests/test_utils/recipes/h100/gpt-grpo.yaml
tdene: Yes, and once I clean up the un-slop the Claude-generated unit tests, they will be doing exactly that.

However, I am concerned that unit tests may not be enough. I'd like to at least have 1 combination of arguments in a functional test; something that hits all the scary code-paths. I'll come back to this and revise.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -433,6 +439,16 @@ def __init__(
                         "Unified memory requested but not available; defaulting to GPU memory."
                     )
                 self.unified_memory_level = 0
+        if self.persist_cuda_graphs and self.unified_memory_level == 0:
+            assert (
+                HAVE_TORCH_MEMORY_SAVER
+            ), "Can only persist CUDA graphs without UVM if torch_memory_saver is available"
```
mathemakitten: This is not true. CUDA graphs can be persisted without UVM if inference and training activations can co-exist within HBM alongside the KV cache.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -710,20 +763,29 @@ def allocate_mamba_states():
...
-        # Allocate `ctx_manager`-managed buffers. (For currently unknown reasons,
-        # `ctx_manager` can only be used once.)
-        ctx_manager = (
-            torch.cuda.use_mem_pool(self.unified_memory_mempool)
-            if self.unified_memory_level > 0
-            else nullcontext()
-        )
+        # Allocate large non-graphed buffers.
+        # - If UVM is turned on, we use UVM unconditionally.
+        # - If CGs persist and we lack UVM, we need to use `torch_memory_saver`.
```
mathemakitten: same comment here as above - we can persist if it all fits within HBM

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -710,20 +763,29 @@ def allocate_mamba_states():
...
+        # - If offloading KV$, we prefer to use `torch_memory_saver` over manual offloading.
+        # - If removing the KV$, we must re-allocate every time, not just at init.
+        ctx_manager = nullcontext()
+        offload_kv = self.kv_cache_management_mode == "offload"
+        remove_kv = self.kv_cache_management_mode == "remove"
+        if self.unified_memory_level > 0:
```
mathemakitten: is it not just 0 or 1?

megatron/training/arguments.py
```diff
@@ -327,29 +327,89 @@ def validate_args(args, defaults={}):
...
-
+        # CUDA graph and KV$ handling support matrix is as follows:
+        # ------------------------------------------------
+        # Resetting CGs only makes sense if we build any CGs
+        has_cg = args.cuda_graph_impl != "none" or args.rl_training_cuda_graphs
+        assert not args.rl_reset_cuda_graphs or has_cg, (
+            "--rl-reset-cuda-graphs is set but no CUDA graphs are being built."
+        )
+        if not args.rl_reset_cuda_graphs:
+            # Persisting CGs requires either torch_memory_saver or UVM
```
mathemakitten: same note as above re: can fit in HBM

megatron/training/arguments.py
```diff
@@ -327,29 +327,89 @@ def validate_args(args, defaults={}):
...
+            try:
+                from torch_memory_saver import torch_memory_saver
+            except ImportError:
+                assert args.inference_dynamic_batching_unified_memory_level > 0, (
+                    "--rl-persist-cuda-graphs requires either torch_memory_saver or "
+                    "--inference-dynamic-batching-unified-memory-level > 0."
+                )
+            # If we persist both training and inference CGs, and keep the KV$,
```
mathemakitten: I don't understand the need for this — we don't warn about expensive memory pressure or weird-but-permissible combinations of flags in other parts of the codebase and I wouldn't frame this as "more memory than expected" — it seems like the user is getting precisely what they asked for.

megatron/training/arguments.py
```diff
@@ -327,29 +327,89 @@ def validate_args(args, defaults={}):
...
+                )
+
+        # Offloading the KV$ is technically what UVM does.
+        # Having both arguments active does not cause any issues, but should be avoided.
+        if args.rl_kv_cache_management_mode == "offload":
+            if args.inference_dynamic_batching_unified_memory_level > 0:
+                warn_rank_0(
+                    "--inference-dynamic-batching-unified-memory-level and "
+                    "--rl-kv-cache-management-mode=offload are both set. "
+                    "This will not cause issues currently, but it is not recommended or supported."
```
mathemakitten: This seems like it should simply not disallowed for being nonsensical, even if the code runs? Setups should be clearly marked as "this is permissible" or "this is not supposed to be used this way", without ambiguity.

megatron/training/arguments.py
```diff
@@ -327,29 +327,89 @@ def validate_args(args, defaults={}):
...
+
+        # TODO: Add support for `torch_memory_saver` with async model weight offload.
         assert not (
             args.rl_offload_inference_model_weights_when_idle
             and args.rl_inference_model_unified_memory_level != 1
         ), (
             "--rl-offload-inference-model-weights-when-idle requires "
             "--rl-inference-model-unified-memory-level=1."
         )
+        # TODO: Add support for `torch_memory_saver` with async KV$ offload.
```
mathemakitten: I'd leave this comment out as it's a detail that probably only we care about (the TODO is tagged to me).

megatron/training/arguments.py
```diff
@@ -1908,12 +1968,17 @@ def _add_rl_args(parser):
                        help="Default top-k for model inference.")
     group.add_argument('--rl-offload-optimizer-during-inference', action='store_true',
                        help='Offload optimizer state to CPU during inference/rollout to save GPU memory')
-    group.add_argument('--rl-offload-kv-cache-during-training', action=argparse.BooleanOptionalAction, default=False,
-                       help='Offload KV cache to CPU during training to save GPU memory')
-    group.add_argument('--rl-remove-kv-cache-during-training', action=argparse.BooleanOptionalAction, default=False,
-                       help='Remove KV cache during training to save GPU memory')
+    group.add_argument('--rl-kv-cache-management-mode', type=str, default='keep',
+                       choices=['keep', 'offload', 'remove'],
```
mathemakitten: nit: I'd go with "persist" over "keep". where "persist" clearly means “keep this cache beyond the current request/session so it can be reused later,” i.e., across requests / rollout sessions; “keep” sounds like “do not touch it right now” but it is somewhat ambiguous as to what this means, really — “keep... in GPU? until some time? until when?”

megatron/training/arguments.py
```diff
@@ -1908,12 +1968,17 @@ def _add_rl_args(parser):
...
+                       help='KV cache management mode during RL training: '
+                            'keep: leave KV cache in GPU memory (default), '
+                            'offload: offload KV cache to CPU during training, '
+                            'remove: delete and reallocate KV cache each training/inference cycle')
     group.add_argument('--rl-reset-cuda-graphs', action=argparse.BooleanOptionalAction, type=bool, default=False,
                        help='Reset CUDA graphs between inference/training to save GPU memory')
+    group.add_argument('--rl-training-cuda-graphs', action=argparse.BooleanOptionalAction, type=bool,
+                       default=False,
+                       help='If set, prevents disabling of training CUDA graphs when switching from inference to training inside the RL loop.')
```
mathemakitten: I'd frame this as "Enables training cudagraphs" as opposed to "prevents disabling of cudagraphs" only for readability reasons.

megatron/core/inference/contexts/dynamic_context.py
tdene: ~~I also wrote code to support the 3rd option: do nothing and just let the KV$ sit there. It's not hard to implement and it can be useful for debugging.~~

tdene: Sorry, I misread :).

Throughout the codebase, this is how the flag is checked. This decision was made originally when the flag was first added, because the plan back then was to have choices of `[0, 1, 2]`. That never panned out, but I'm choosing to follow pre-existing convention in this PR.

megatron/training/arguments.py
tdene: Addressed?

megatron/training/arguments.py
tdene: Addressed.

megatron/training/arguments.py
tdene: Sounds good; addressed this.

megatron/training/arguments.py
tdene: Sounds good; addressed this.

megatron/training/arguments.py
tdene: Sounds good; addressed this.

megatron/training/arguments.py
tdene: Sounds good; addressed this.

megatron/training/arguments.py
tdene: Sounds good; addressed this.

megatron/training/arguments.py
tdene: Roger. Let's consolidate discussion to the first thread.

megatron/core/inference/contexts/dynamic_context.py
tdene: Roger. Let's consolidate discussion to the first thread.

megatron/core/inference/contexts/dynamic_context.py
tdene: You're right. I've made a commit which re-organizes things; let me know if this looks correct now.

In a nutshell, I am now checking against `not(rl_reset_cuda_graphs) and rl_kv_cache_management_mode!="persist"` instead of just checking against `rl_reset_cuda_graphs`.

The `rl_reset_cuda_graphs` flag controls whether we rebuild CGs, and that's it.    
The problem happens at the intersection: when we rebuild CGs, but we don't keep the KV$ memory addresses static. That's when we need to assert either UVM or TMS.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -421,9 +421,17 @@ def __init__(
...
+        # Unified memory and general tensor management.
         self.unified_memory_level = unified_memory_level
-        self.persist_cuda_graphs = persist_cuda_graphs
+        self.reset_cuda_graphs = reset_cuda_graphs
+        # KV cache management mode: "persist", "offload", or "remove"
+        assert kv_cache_management_mode in (
+            "persist",
+            "offload",
+            "remove",
+        ), f"Invalid kv_cache_management_mode: {kv_cache_management_mode}"
```
mathemakitten:
```suggestion
        ), f"Invalid kv_cache_management_mode: {kv_cache_management_mode}. Options are `persist`, `offload`, `remove`."
```

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -433,6 +441,19 @@ def __init__(
                         "Unified memory requested but not available; defaulting to GPU memory."
                     )
                 self.unified_memory_level = 0
+        # If CUDA graphs are not reset but KV$ memory address is not static, we need
```
mathemakitten:
```suggestion
        # If CUDA graphs are not reset and KV cache memory address is not static, we need
```
nit sorry :P

megatron/core/inference/contexts/dynamic_context.py
mathemakitten: I remember that. Can we change it back to [0, 1]? Someone asked me what UVM = 2 does last week and I had to go check if it had been implemented, so I suggest we phase it out to avoid future confusion.

megatron/core/inference/contexts/dynamic_context.py
mathemakitten: Yes, I think the `if not self.reset_cuda_graphs and self.kv_cache_management_mode != "persist":` block looks good now.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -593,21 +650,29 @@ def allocate_all_tensors(self, *, is_init: bool) -> None:
...
         self.is_tensor_state_allocated = True
 
+        # Only allocate tensors when not using either unified memory or torch_memory_saver.
+        # Otherwise, we do not need to perform any explicit allocations.
+        if not is_init:
+            if self.unified_memory_level != 0:
+                return
+            if len(self._offloadable_tensor_names) > 1:
+                torch_memory_saver.restore_region(tag="inference_context")
+                return
```
mathemakitten: Previously we only offloaded the KV cache, so the tag was scoped to `kv_cache`. This looks like it is intended to be more general. What is the intended universe of tensors which can be offloaded? Is it just the KV cache and Mamba states?

megatron/training/arguments.py
```diff
@@ -327,22 +327,66 @@ def validate_args(args, defaults={}):
...
-        assert not args.inference_dynamic_batching_unified_memory_level, "The KV cache should not be instantiated in unified memory when it is offloaded during training."
-
-    # Batch size checks if running RL.
     if args.perform_rl_step:
-        assert not (args.rl_remove_kv_cache_during_training and args.rl_offload_kv_cache_during_training), \
-            "Cannot use both remove-kv-cache-during-training and offload-kv-cache-during-training"
-
-        assert not (args.rl_partial_rollouts and args.rl_remove_kv_cache_during_training), \
-            "Cannot use both partial-rollouts and remove-kv-cache-during-training"
+        # CUDA graph and KV$ handling support matrix is as follows:
```
mathemakitten: We should default to "KV cache" or "KV$" but mixing it like this makes the codebase much less searchable in the editor.

megatron/rl/inference/megatron.py
```diff
@@ -249,6 +249,9 @@ async def launch(cls, model: GPTModel, **kwargs):
         args = get_args()
         tokenizer = get_tokenizer()
 
+        self.partial_rollouts = args.rl_partial_rollouts
+        self.kv_cache_management_mode = args.rl_kv_cache_management_mode
```
mathemakitten: This accidentally mixes `cls` and `self`.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -593,21 +650,29 @@ def allocate_all_tensors(self, *, is_init: bool) -> None:
...
-            return
-
         # Mark allocated.
         if self.is_tensor_state_allocated:
             return
         self.is_tensor_state_allocated = True
 
+        # Only allocate tensors when not using either unified memory or torch_memory_saver.
+        # Otherwise, we do not need to perform any explicit allocations.
+        if not is_init:
```
mathemakitten: There's a logical error here for iterations > 1; `request_ids` and the like are already allocated so they appear in `vars(self).keys()` and hit the `"All tensors should be allocated within `allocate_all_tensors()."`error.

megatron/core/inference/contexts/dynamic_context.py
tdene: That's an important detail, yeah.

I've discussed this with others, and currently it's just KV$ and mamba states because these are the largest tensors, memory-footprint-wise. If we add any large tensors in the future, we'd have to include those as well.    
We could include all GPU tensors today, but that could cause a performance hit due to all the small D2H / H2D copies. This can be a follow-up PR after we merge this and benchmark.

I have noted this inside a docstring as a TODO:
```
+        NOTE: This current tracks the KV cache and mamba states, as they are the largest tensors.
+        TODO: Benchmark the benefits of tracking additional tensors.
```

megatron/training/arguments.py
tdene: My HW background's leaking through, mb.

I've replaced all instances of "KV$" with "KV cache". Addressed!

megatron/rl/inference/megatron.py
tdene: Addressed by reverting to an earlier version. I need someone else's advice to clean this up properly...

megatron/rl/inference/megatron.py
tdene: Re-addressed.

megatron/rl/inference/megatron.py
tdene: Discussed with Robert, and gotten the go-ahead to keep it this way for now, and call it forward-facing code that we'll modify once we start doing complicated experiments.

megatron/rl/inference/megatron.py
tdene: Re-addressed.

megatron/core/transformer/utils.py
yobibyte: Do we need this argument at all then?

tests/unit_tests/rl/test_rl_suspend.py
tdene: Resolving this because it's caused by AI slop that I've since reverted.

tests/unit_tests/rl/test_rl_suspend.py
tdene: Resolving this because it's caused by AI slop that I've since reverted.

tests/unit_tests/rl/test_rl_suspend.py
tdene: Resolving this because it's caused by AI slop that I've since reverted.

megatron/core/transformer/utils.py
tdene: That's fair. Addressed. I'll ask for the opinion of @mathemakitten before final merge.

tests/unit_tests/rl/test_rl_suspend.py
tdene: Resolved; this is a comment on AI slop that has since been removed.

tests/unit_tests/rl/test_rl_suspend.py
tdene: Resolved; this is a comment on AI slop that has since been removed.

tests/unit_tests/rl/test_rl_suspend.py
tdene: Resolved; this is a comment on AI slop that has since been removed.

tests/unit_tests/rl/test_rl_suspend.py
tdene: Resolved; this is a comment on AI slop that has since been removed.

tests/unit_tests/rl/test_rl_suspend.py
tdene: Resolved; this is a comment on AI slop that has since been removed.

megatron/core/inference/contexts/dynamic_context.py
tdene: Addresed

megatron/core/inference/contexts/dynamic_context.py
tdene: I consider this addressed due to the TODO comment.

megatron/core/inference/config.py
```diff
@@ -102,8 +102,14 @@ class InferenceConfig:
     Eventually, additional levels will be included to control other tensors within the context.
     """
 
-    offload_kv_cache: bool = False
-    """If True, offload KV cache during RL training."""
+    kv_cache_management_mode: str = "persist"
```
lmcafee-nvidia: I feel like this should be an enum instead of string values. This can easily lead to bugs

megatron/core/inference/config.py
```diff
@@ -102,8 +102,14 @@ class InferenceConfig:
...
+    """
+    Mode used to determine how large tensors are handled by the allocate and deallocate methods.
+    Options are:
+        - "persist": do not deallocate and reallocate large tensors; keep them on GPU.
+        - "offload": offload large tensors to CPU during dellocation; onload during allocation.
+        - "remove": deallocate large tensors during dellocation; reallocate during allocation.
```
lmcafee-nvidia: nit: "remove" might be perfectly fine, but one other idea is "recompute" to emphasize that it'll be back later

megatron/core/inference/config.py
```diff
@@ -124,11 +130,10 @@ class InferenceConfig:
     Whether to use CUDA graphs for non-decode steps.
     """
 
-    persist_cuda_graphs: bool = False
+    reset_cuda_graphs: bool = False
```
lmcafee-nvidia: nit: the term 'reset' is slightly unclear to me, as in are they being deleted, or is their internal state simply being reset? I'd learn towards 'delete' or 'remove' to be more explicity

megatron/rl/inference/megatron.py
```diff
@@ -74,6 +74,7 @@ class MegatronLocal(InferenceServer, ReturnsTokens, ReturnsRaw):
 
     _client: InferenceClient = PrivateAttr(None)
     _inference_engine: DynamicInferenceEngine = PrivateAttr(None)
+    _rl_kv_cache_management_mode: str
```
lmcafee-nvidia: mentioned in other comment, but I feel like this really needs to be an enum, for robustness

megatron/rl/rl_utils.py
```diff
@@ -1665,27 +1663,21 @@ def megatron_rl_inference_mode(
...
-    offload_kv_cache_during_training: bool,
-    remove_kv_cache_during_training: bool,
 ):
     """Manage the model inference context when collecting rollouts.
 
     Args:
         model: model to prepare.
         optimizer: optimizer used to train the model.
         cuda_graph_impl: which cuda graph implementation to use.
-        reset_cuda_graphs: rebuild cuda graphs for each inference stage or not.
```
lmcafee-nvidia, line 1664: add an entry for `training_cuda_graphs`

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -729,6 +730,7 @@ def _add_request(
             request.sampling_params.num_tokens_to_generate = (
                 request.sampling_params.num_tokens_total - len(request.prompt_tokens)
             )
+            request.sampling_params.num_tokens_total = None
```
lmcafee-nvidia, line 740: why do we need to set this to `None`? was there a bug?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -609,9 +610,10 @@ def suspend(self):
         self.resume_request_ids = [*active_request_ids, *waiting_request_ids]
         self.waiting_request_ids.clear()
 
-        # Suspend requests objects.
+        # Suspend request objects that are marked for recompute.
         for request_id in active_request_ids:
-            self.requests[request_id].record.checkpoint()
+            if self.requests[request_id].recompute_upon_suspend:
```
lmcafee-nvidia: suspend and recompute are independent now? what has changed?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -596,11 +599,9 @@ def suspend(self):
             "suspended", unified_memory_level=self.unified_memory_level
         ):
             self.context.deallocate_all_tensors()
+            torch.cuda.synchronize()
 
-        # Delete cuda graphs when not using unified memory at all (level 0) and
-        # `--rl-training-cuda-graphs` is not passed. For UVM levels 1 and 2, the context's tensors
-        # maintain static memory addresses, so the cuda graphs are re-used.
-        if self.unified_memory_level == 0 and not self.persist_cuda_graphs:
+        if self.reset_cuda_graphs:
```
lmcafee-nvidia: as mentioned in the config, I think a config value of `delete_cuda_graphs` would align better name-wise

lmcafee-nvidia: still perusing the PR, but it seems like a lot of things are independently controllable now:
- unified memory
- delete cuda graphs
- recompute requests

is it generally true that all of these are completely decoupled now?

megatron/training/arguments.py
```diff
@@ -327,21 +327,64 @@ def validate_args(args, defaults={}):
...
+                    "--rl-kv-cache-management-mode=persist, UVM, or install torch_memory_saver."
+                )
+
+        # There's no need to manually offload the KV cache with UVM.
+        assert not (
+            args.inference_dynamic_batching_unified_memory_level > 0
+            and args.rl_kv_cache_management_mode == "offload"
+        ), "--rl-kv-cache-management-mode=offload is incompatible with UVM"
+        # ------------------------------------------------
+        # CGs are handled as follows:
```
lmcafee-nvidia: this is a great comment!

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -668,25 +708,32 @@ def deallocate_all_tensors(self):
...
-        # Delete all tensor attributes.
+        # UVM and `torch_memory_saver` do not require explicit deallocation.
+        if self.unified_memory_level != 0:
+            return
+        if len(self._offloadable_tensor_names) > 1:
+            torch_memory_saver.backup_region(tag="inference_context")
+            return
+
+        # Explicitly deallocate tensors and offload them.
+        if self.kv_cache_management_mode == "offload":
```
lmcafee-nvidia: less of a concern for us with 2+ TB of memory, but do we need special handling/warning/error if CPU memory OOMs?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -125,6 +125,7 @@ class RequestEntry:
 
     record: DynamicInferenceRequestRecord
     future: asyncio.Future
+    recompute_upon_suspend: bool = False
```
lmcafee-nvidia: just curious, since each individual request has attribute `recompute_upon_suspend`. Is there a future where this value is not equal for all requests? I.e., some requests remain live while others are suspended?

megatron/core/inference/engines/dynamic_engine.py
tdene: Yeah. The bug was with the checkpointing code, which would snapshot the sampling parameters. Then, if recomputing a request, it would re-submit it with the snapshotted sampling parameters.

This block of code sets `num_tokens_to_generate` based on the value of `num_tokens_total`, so upon request resubmit, the request would have instructions for both `num_tokens_to_generate` and `num_tokens_total`, which is forbidden.

The change here has always been necessary, we just didn't encounter it until now.

megatron/core/inference/contexts/dynamic_context.py
tdene: I do not think we need special handling; I think the CPU memory message will be self-explanatory. Is that fair?

megatron/core/inference/config.py
tdene: Addressed!

megatron/core/inference/config.py
tdene: Addressed!

megatron/core/inference/config.py
tdene: I changed it to `static_kv_memory_pointers`.

What it truly does is even more different than what you suggest.    
As far as the context is concerned, this flag determines whether the KV$ tensors should have static addresses, or whether their addresses might change.    
As far as the engine is concerned, this flag determines whether it /needs/ to rebuild the CGs; because if addresses are not static, we have no other choice.

megatron/rl/inference/megatron.py
tdene: Addressed!

megatron/rl/rl_utils.py
tdene: Addressed!

megatron/core/inference/engines/dynamic_engine.py
tdene: In a nutshell, the RL flow ALWAYS go through `suspend`. It's a unified code-path.

Sometimes `suspend` also does recompute on some requests - but not necessarily all requests.

megatron/core/inference/engines/dynamic_engine.py
tdene: Discussed in the other comment.

tdene: Correct!!!

megatron/core/inference/engines/dynamic_engine.py
tdene: Yes; that is in fact a necessary future, and the primary motivation behind this PR.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -518,23 +530,74 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
...
+                    self.mamba_conv_states, device="cpu"
+                ).pin_memory()
+                self._offloadable_tensor_names.append("mamba_ssm_states")
+                self._offloadable_cpu_backups["mamba_ssm_states"] = torch.empty_like(
+                    self.mamba_ssm_states, device="cpu"
+                ).pin_memory()
+        else:
+            self.mamba_metadata = None
 
+    def allocate_all_tensors(self) -> None:
```
tdene: Rename to `initialize_all_tensors`.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -689,7 +689,7 @@ def reallocate_large_tensors(self):
             return
 
         if self._uses_torch_memory_saver:
-            torch_memory_saver.restore_region(tag="inference_context")
+            torch_memory_saver.resume("inference_context")
```
yobibyte: I would use a constant for 'inference_context'.

yobibyte (approved):

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -628,103 +691,90 @@ def allocate_all_tensors(self, *, is_init: bool) -> None:
...
+            self._allocate_memory_buffer()
+            self._allocate_mamba_states()
 
         # Reset attention and Mamba state.
         self.reset_attention_state()
         self.reset_mamba_state()
 
-    def deallocate_all_tensors(self):
-        """Deallocate GPU state.
+    def reallocate_large_tensors(self):
```
tdene: Rename `large_tensors`

megatron/rl/rl_utils.py
```diff
@@ -488,10 +488,11 @@ def get_environment_rollouts(
...
             else:
                 logger.warning(
-                    "Gradient buffers will not be offloaded when training cudagraphs are enabled!")
+                    "Gradient buffers will not be offloaded when training cudagraphs are used!"
+                )
             optimizer.offload_to_cpu()
-             
-    # If we have separate training and inference models we to refit weights from the training model to the inference model.
+
+    # If we have seperate training and inference models we to refit weights from the training model to the inference model.
```
mathemakitten:
```suggestion
    # If we have separate training and inference models we to refit weights from the training model to the inference model.
```
nit sorry

megatron/training/arguments.py
```diff
@@ -327,21 +327,72 @@ def validate_args(args, defaults={}):
...
+
+        # There's no need to manually offload the KV cache with UVM.
+        assert not (
+            args.inference_dynamic_batching_unified_memory_level > 0
+            and args.rl_kv_cache_management_mode == "offload"
+        ), "--rl-kv-cache-management-mode=offload is incompatible with UVM"
+        # We currently cannot recapture CGs in offload mode.
+        assert not(
+            not args.rl_persist_cuda_graphs and args.rl_kv_cache_management_mode == "offload"
+        ), "Cannot recpature CUDA graphs while offloading KV cache."
```
mathemakitten:
```suggestion
        ), "Cannot recapture CUDA graphs while offloading KV cache."
```

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -582,7 +645,7 @@ def allocate_all_tensors(self, *, is_init: bool) -> None:
         for key in vars(self).keys():
             value = getattr(self, key)
             assert not isinstance(value, torch.Tensor), (
-                "All tensors should be allocated within `allocate_all_tensors()."
+                "All tensors should be allocated within `allocate_all_tensors()`. "
```
mathemakitten: You mean `initialize_all_tensors` now right?

tdene: Thanks! Addressed.

tdene:
Internal pipeline passes on all CI functional tests and outside-of-CI functional tests.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -556,23 +569,74 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
...
+                    self.block_allocator.total_count,
+                    self.block_size_tokens,
+                    self.num_attention_heads_per_partition,
+                    self.hidden_size_per_attention_head,
+                ),
+                dtype=self.params_dtype,
+                device=torch.cuda.current_device(),
+            )
+        if (
+            self.kv_cache_management_mode == KVCacheManagementMode.OFFLOAD
```
kvareddy, line 599: I guess the condition here should also include self.unified_memory_level == 0

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -556,23 +569,74 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
...
+                dtype=self.params_dtype,
+                device=torch.cuda.current_device(),
+            )
+            self.mamba_ssm_states = torch.empty(
+                (self.num_mamba_layers, self.max_requests) + self.mamba_ssm_states_shape,
+                dtype=self.params_dtype,
+                device=torch.cuda.current_device(),
+            )
+            if (
+                self.kv_cache_management_mode == KVCacheManagementMode.OFFLOAD
```
kvareddy, line 625: same as above

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -628,103 +692,90 @@ def allocate_all_tensors(self, *, is_init: bool) -> None:
...
-        # deallocations after the initial call to `__init__()`.
-        if self.unified_memory_level != 0:
+        if (
+            self.unified_memory_level != 0
+            and self.kv_cache_management_mode != KVCacheManagementMode.RECOMPUTE
+        ):
+            return
+
+        if self._uses_torch_memory_saver:
+            torch_memory_saver.resume("inference_context")
```
kvareddy: When we resume under recomputation, how do we guarantee initialized values for all tensors?

megatron/core/inference/contexts/dynamic_context.py
tdene: That is what I originally had written: UVM and `torch_memory_saver` are just advanced ways of doing `OFFLOAD`. Early reviewers on this PR disagreed, and said that UVM does not count as "offload".

In its current state, this PR asserts that UVM is incompatible with `OFFLOAD`:
```
        assert not (
            args.inference_dynamic_batching_unified_memory_level > 0
            and args.rl_kv_cache_management_mode == "offload"
        ), "--rl-kv-cache-management-mode=offload is incompatible with UVM"
```
So UVM is being treated as a way of persisting the KV cache inplace.

In the end, it's a semantics question: should we think of UVM as offloading/onloading, or should we think of UVM as just giving us a huge pool of memory. More than 1 reviewer told me that it's better to think of UVM as a huge pool of memory, so I followed their advice.

megatron/core/inference/contexts/dynamic_context.py
tdene: See above; it's a semantics issue that was discussed with early reviewers.

megatron/core/inference/contexts/dynamic_context.py
tdene: This specific case is covered by the engine running `self.context.reset()` on line 651, after we return out of `deallocate_inference_state_buffers`.

tdene: This case is complicated, so here's a more in-depth discussion.

`_uses_torch_memory_saver` is set if `HAVE_TORCH_MEMORY_SAVER and need_static_addr`. That `need_static_addr` is set only if we want to persist CUDA graphs. So we only enter this case when are using `RECOMPUTE`, persisting CUDA graphs, and have `torch_memory_saver` available.

Because we persist CUDA graphs, we cannot delete the tensors during suspend and then run `initialize_all_tensors` during resume. We have to keep the memory pointers consistent. So we use `torch_memory_saver` to free up the physical memory during suspend and bring it back during resume. All the non-KV-cache tensors are kept in-place, and then reset to their initialized values by `context.reset()`.

The `torch_memory_saver` setup has a `enable_cpu_backup` argument that determines whether it keeps the data around or not. In recompute mode, we set `enable_cpu_backup=False`, so we do not waste time keeping the KV cache data around - because we're going to recompute it at all anyway. We preserve just the memory pointers to avoid rebuilding CUDA graphs.

PR #3032, opened 2026-01-21
Add ability to save wgrads and dgrads

deepakn94:
wgrad saving requires that the DP collective is an all-reduce instead of reduce-scatter to make it easier to pull wgrads from DP replica 0. Some other gotchas:
- First gradient reduction (when metadata is being collected about when to launch collectives) should also be an all-reduce if needed
- start_grad_sync call should be a no-op if in first batch and collective has already been dispatched
- Don't try calling reduce_scatter_with_fp32_accumulation's .wait() method if using all-reduce in a particular iteration

### DIFF
--- a/megatron/training/dgrad_logging.py
+++ b/megatron/training/dgrad_logging.py
@@ -48,7 +48,10 @@ LINEAR_TYPES = _get_linear_types()
 
 
 class DataGradLogger:
-    """Captures and saves gradients from all linear layers using backward hooks."""
+    """Captures and saves gradients from all linear layers using backward hooks.
+    
+    NOTE: Right now, we only save the dgrads for the last microbatch in a batch on DP replica 0.
+    The code below would need to be extended to save dgrads for all microbatches in a batch."""
 
     def __init__(self, save_dir: str):
         self._save_dir = save_dir


megatron/core/pipeline_parallel/schedules.py
```diff
@@ -125,11 +125,15 @@ def forward_step(data_iterator, model):
         respective list of shapes. Thus it is not used in the other forward-backward functions
         which have different shape handling.
 
+    force_all_reduce (bool, optional): Run all-reduce for this iteration instead of maybe
+        reduce-scatter to ensure all ranks have fully reduced gradients.
```
erhoo82: `Run all-reduce for the gradient reduction instead of reduce-scatter in this iteration to ensure all data-parallel ranks have fully reduced gradients.`

megatron/training/training.py
```diff
@@ -1590,16 +1591,23 @@ def dummy_train_step(data_iterator):
...
-def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_scheduler, config, forward_backward_func):
+def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_scheduler, config, forward_backward_func, iteration=None):
     """Single training step."""
     args = get_args()
     timers = get_timers()
 
     rerun_state_machine = get_rerun_state_machine()
+    save_dgrads_in_this_iteration = (args.save_dgrads_interval is not None and
+                                     (iteration + 1) % args.save_dgrads_interval == 0)
+    # If saving main_grads in this iteration, then all-reduce instead of reduce-scatter.
```
erhoo82: I think this comment can move to row 1609 and replace the current comment on row 1609?

jaredcasper (approved):

megatron/training/checkpointing.py
```diff
@@ -432,6 +433,42 @@ def _build_sharded_state_dict_metadata(args: Namespace, dp_cp_group: Optional[to
...
+
+    print_rank_0(f"  [{datetime.now().strftime('%Y-%m-%d %H:%M:%S.%f')}] saving {grad_label} "
+                 f"from iteration {iteration:7d}")
+
+    if mpu.get_expert_data_parallel_rank() == 0:
+        # Create saving directory.
+        ep_rank = mpu.get_expert_model_parallel_rank()
+        pp_rank = mpu.get_pipeline_model_parallel_rank()
+        tp_rank = mpu.get_tensor_model_parallel_rank()
+        assert args.save is not None
```
maanug-nv: nit: if `args.save` is the only setting needed, can it be a function argument, and passed in by `train_step()` ?

maanug-nv:
i think this deserves tests, is it possible to add unit tests for this? for both the grad saving logic in training and the forced all reduce in core

PR #3036, opened 2026-01-22
Logging cleanup (only log on rank 0 if possible)

deepakn94:
If we don't do this, then logfiles for jobs with thousands of ranks can get very cluttered.

### DIFF
--- a/megatron/core/tokenizers/text/libraries/tiktoken_tokenizer.py
+++ b/megatron/core/tokenizers/text/libraries/tiktoken_tokenizer.py
@@ -2,6 +2,7 @@
 
 import base64
 import json
+import logging
 import os
 from pathlib import Path
 from typing import Dict, List, Optional
@@ -23,6 +24,9 @@ SPECIAL_TOKENS = ["<unk>", "<s>", "</s>", "<mask>", "<pad>", "<cls>", "<sep>"]
 SPECIAL_TOKEN_TEMPLATE = "<SPECIAL_{id}>"
 
 
+logger = logging.getLogger(__name__)
+
+
 def reload_mergeable_ranks(
     path: str, max_vocab: Optional[int] = None, num_special_tokens: Optional[int] = None
 ) -> Dict[bytes, int]:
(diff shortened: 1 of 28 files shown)


megatron/core/tokenizers/text/libraries/tiktoken_tokenizer.py
```diff
@@ -39,15 +39,18 @@ def reload_mergeable_ranks(
...
     assert path.endswith(".json")
+    from megatron.core.utils import safe_get_rank
 
     # reload vocab
     with open(path, "r") as f:
         vocab = json.load(f)
     assert isinstance(vocab, list)
-    print(f"Vocab size: {len(vocab)}")
+    if safe_get_rank() == 0:
+        print(f"Vocab size: {len(vocab)}")
```
deepakn94: No prints in `megatron/core`.

megatron/core/utils.py
```diff
@@ -45,6 +46,7 @@
 
 from megatron.core import config
 from megatron.core.package_info import __version__ as mcore_version
+from megatron.core._rank_utils import safe_get_rank
```
deepakn94: This was done to avoid circular imports. Apparently this is the cleanest way of doing this according to Opus 4.5 🤷🏾

megatron/training/arguments.py
```diff
@@ -1298,7 +1298,8 @@ def validate_args(args, defaults={}):
         assert args.replication_jump is not None, "--replication requires the value of --replication-jump!"
         assert args.non_persistent_ckpt_type == "local", f"--replication requires args.non_persistent_ckpt_type == 'local', but got: {args.non_persistent_ckpt_type}"
     elif args.replication_jump:
-        print("Warning: --replication-jump was specified despite not using replication. Ignoring.")
+        if args.rank == 0:
+            print("Warning: --replication-jump was specified despite not using replication. Ignoring.")
```
yaox12: Should we replace all `print` with either `warnings.warn` or `logging`?

pretrain_gpt.py
```diff
@@ -2,11 +2,20 @@
...
 """Pretrain and SFT GPT."""
 
-# Capture the true program start time BEFORE any heavy imports
+# Capture the true program start time BEFORE any heavy imports.
 import time
 _PROGRAM_START_TIME = time.time()
 
 import json
+
+# Suppress warnings on all ranks but rank 0.
```
deepakn94, line 11: Note that this is from other libraries.

megatron/training/arguments.py
deepakn94: Done!

PR #3039, opened 2026-01-22
Fuse MLA DOWN projection GEMMs

cjld:
dev PR: #2960 

# What does this PR do ?
Add cli args `--mla-down-proj-fusion` and `mla_down_proj_fusion` to the transformer config for the fusion.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


maanug-nv:
is there a PR to dev that we can link? i don't see the same changes on dev branch currently?

megatron/core/models/gpt/gpt_layer_specs.py
```diff
@@ -235,6 +242,31 @@ def get_gpt_layer_with_transformer_engine_spec(
 
     if multi_latent_attention:
         assert qk_l2_norm is False, "qk_l2_norm is not supported with MLA."
+
+        fuse_input_layernorm = (
```
maanug-nv: can `fuse_input_layernorm` just be an argument to this function? that seems a more consistent UX than adding environment variables for any adjustment to the spec.

megatron/core/models/gpt/gpt_layer_specs.py
```diff
@@ -171,6 +171,13 @@ def get_gpt_layer_with_inference_spec(
         )
 
 
+import os
+# NOTE: These toggles are intentionally environment-driven for experimentation.
+# They are expected to be replaced by explicit config options.
```
jaredcasper: Please replace them now. We do not use environment variables to configure anything in mcore.

maanug-nv:
Hi @cjld , are you able to update this PR to remove usage of env vars? I think that's the main blocker

megatron/core/models/gpt/gpt_layer_specs.py
cjld: ok, added fuse_mla_input_layernorm

megatron/core/models/gpt/gpt_layer_specs.py
cjld: ok, replaced -> --qkv-down-proj-fusion

cjld:
Hi, @maanug-nv , I updated the PR. If the PR still has problems, please tell me.

cjld:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3039/commits/e245e191724de4766e8e4b4f8af742a5810a7fa8

megatron/core/transformer/multi_latent_attention.py
```diff
@@ -411,29 +424,31 @@ def __init__(
...
-                tp_comm_buffer_name='q_down_proj',
-                skip_weight_param_allocation=False,
-                tp_group=(
-                    pg_collection.tp
-                    if q_down_proj_kwargs.get('parallel_mode') != 'duplicated'
-                    else None
-                ),
-                **q_down_proj_kwargs,
-            )
+            if not self.config.qkv_down_proj_fusion:
```
jaredcasper: Shouldn't this be handled by the spec?

megatron/core/transformer/multi_latent_attention.py
```diff
@@ -456,29 +471,55 @@ def __init__(
...
-            tp_comm_buffer_name='kv_down_proj',
-            skip_weight_param_allocation=False,
-            tp_group=(
-                pg_collection.tp
-                if kv_down_proj_kwargs.get('parallel_mode') != 'duplicated'
-                else None
-            ),
-            **kv_down_proj_kwargs,
-        )
+        if not self.config.qkv_down_proj_fusion:
```
jaredcasper: likewise here.

megatron/core/transformer/multi_latent_attention.py
cjld: fusion now control by spec, see [1d97497](https://github.com/NVIDIA/Megatron-LM/pull/3039/commits/1d97497ca0eafe7c0a06b18ee4621baedd1a061d)

megatron/core/transformer/multi_latent_attention.py
cjld: same as above

megatron/core/transformer/multi_latent_attention.py
```diff
@@ -460,29 +477,55 @@ def __init__(
...
-            tp_comm_buffer_name='kv_down_proj',
-            skip_weight_param_allocation=False,
-            tp_group=(
-                pg_collection.tp
-                if kv_down_proj_kwargs.get('parallel_mode') != 'duplicated'
-                else None
-            ),
-            **kv_down_proj_kwargs,
-        )
+        if submodules.linear_qkv_down_proj is None:
```
jaredcasper: Can we get rid of this if/else. We really want to avoid things like this in the modules. Either using the spec and IdentityOps, or abstracting things out into different classes derived from a common base (see for example SelfAttention vs CrossAttention). Maybe FusedMLASelfAttention or something?

cjld: Okay, I'll finish the refactoring after the Chinese new year

cjld: Hi, jaredcasper. 
Thanks for the review — I refactored the MLA fusion path to follow the same pattern as separate attention implementations (e.g., class-level specialization selected in the spec), instead of adding runtime branching in the core attention module.

What changed
* Added a dedicated `FusedMLASelfAttention` class and kept MLASelfAttention as the unfused baseline.
* Switched selection to spec-time via ModuleSpec in GPT TE layer spec (mla_fusion decides which class is instantiated).
* Kept fusion as a single best-effort flag (`--mla-fusion`): it enables fused q/kv down-projection and fused input layernorm when backend support exists, otherwise falls back gracefully.
* Added checkpoint compatibility logic so fused implementation can interoperate with legacy unfused checkpoint formats.

This should address the concern about runtime conditional complexity in the core module while preserving the intended fusion behavior.

megatron/core/models/gpt/gpt_layer_specs.py
```diff
@@ -248,6 +253,47 @@ def get_gpt_layer_with_transformer_engine_submodules(
             if qk_layernorm
             else backend.column_parallel_linear()
         )
+
+        if mla_down_proj_fusion:
+            fuse_input_layernorm = (
+                mla_down_proj_fusion and backend.column_parallel_layer_norm_linear() is not None
```
ericharper: do we need mla_down_proj_fusion here? we're already in the conditional?

ericharper:
Can we add unit tests for this feature?

cjld:
> Can we add unit tests for this feature?

OK

cjld:
Unittests added.

gpt_builders.py
```diff
@@ -137,6 +137,7 @@ def _get_transformer_layer_spec(use_te, config):
             use_te_activation_func=config.use_te_activation_func,
             use_kitchen_attention=config.use_kitchen_attention,
             kitchen_attention_backend=config.kitchen_attention_backend,
+            mla_down_proj_fusion=args.mla_down_proj_fusion,
```
gdengk: Can we make sure this option could be set through transformer_config as well?
mla_down_proj_fusion: bool = False in transformer_config.py is a dead field.

yaox12: Taking a closer look at the code, we should replace all these `args.` with `config.` right? That's to say, all these `num_experts`/`moe_grouped_gemm`/`qk_layernorm` ... options are only set through args but not `transformer_config.py`? I'm a bit shocked by this.

yaox12: Also, we need another PR to the Megatron-Bridge to correctly parse this option.

gdengk: Megatron Bridge calls the `get_gpt_decoder_block_spec`  (from megatron.core.models.gpt.gpt_layer_specs import get_gpt_decoder_block_spec). So actually it have no access to the function in gpt_builder.py. The func in gpt_builders.py is limited to Megatron-lm usage only so I think options starting with args. should still be fine.
Just talked with Dun, I think we need to pass the option correct in the get_gpt_decoder_block_spec from megatron/core/models/gpt/gpt_layer_specs.py .

cjld:
> Can we add unit tests for this feature?

Hi, @ericharper , could you please review this unittest?

Merged FusedMLA unit tests into `test_multi_latent_attention.py`, covering:

- **Constructor & weight shape**: Verifies `FusedMLASelfAttention` instantiation and fused down-projection weight dimensions (`q_lora_rank + kv_lora_rank + qk_pos_emb_head_dim`).
- **Forward pass**: Tests GPU forward in both fp32 and bf16, parametrized over `rope_type` (`yarn`/`rope`).
- **QKV split**: Validates `_qkv_down_projection` output shapes for q and kv branches.
- **Backward pass**: Confirms gradient flow through `linear_qkv_down_proj` and input.
- **State dict compatibility**: Loads unfused state dict into fused model and verifies weight fusion; checks `sharded_state_dict()` splits fused weight back into separate q/kv keys.
- **Error handling**: Asserts `q_lora_rank=None` raises `AssertionError`.

All 14 tests pass.

PR #3048, opened 2026-01-22
Refactor `rl_offload_kv_cache_during_training` to offload KV cache to CPU while retaining fixed virtual address 

mathemakitten:
# What does this PR do ?
Previously, setting `args.rl_offload_kv_cache_during_training` would offload the KV cache to CPU to save memory while switching to training mode and reload it back onto device. The new KV cache tensor would be instantiated with a new memory address, which necessitated rebuilding inference cudagraphs on every step. 

Orthogonally, training cudagraphs previously could only be used when the KV cache, inference activations, and training activations could all be co-located within HBM.

Now, we suspend the physical memory underlying the KV cache tensor and offload the tensor contents to CPU when transitioning from inference to training, using `torch_memory_saver`. Notably, the virtual memory address of the KV cache tensor is kept consistent. Training cudagraphs can now be captured while accounting for this offload: since the KV cache tensor has been offloaded with `torch_memory_saver.pause()`, its CUDA allocation is released back to PyTorch’s caching allocator, so it is safe to reuse this memory for other training-side tensors. 

When transitioning from training to inference, physical memory is simply re-bound to the same virtual address. This means that cudagraphs no longer need to be disabled between phases, since the KV cache has a consistent virtual address despite being offloaded/onloaded. This is compliant with partial rollouts. 

The result is a minor increase in max reserved memory, with lower peak allocated memory and a consistent memory usage pattern over many iterations. This _replaces_ the old behaviour of `args.rl_offload_kv_cache_during_training`.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -202,8 +202,28 @@ jobs:
       && needs.pre-flight.outputs.is_merge_group == 'false'
       && !cancelled()
     steps:
+      - name: Get PR info
+        id: get-pr-info
+        if: startsWith(github.ref, 'refs/heads/pull-request/')
+        uses: nv-gha-runners/get-pr-info@main
+
+      - name: Get merge commit sha
+        shell: bash -x -e -u -o pipefail {0}
+        id: sha
+        env:
+          IS_PR: ${{ startsWith(github.ref, 'refs/heads/pull-request/') }}
+        run: |
+          if [[ "$IS_PR" == "true" ]]; then
+            SHA=${{ fromJSON(steps.get-pr-info.outputs.pr-info || '{}').merge_commit_sha }}
+          else
+            SHA=${GITHUB_SHA}
+          fi
+          echo "main=${SHA}" | tee -a "$GITHUB_OUTPUT"
+
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.sha.outputs.main }}
 
       - name: Setup python
         uses: actions/setup-python@v5
@@ -215,11 +235,6 @@ jobs:
         run: |
           apt-get update
           apt-get install -y gh
-
-      - name: Get PR info
-        id: get-pr-info
-        if: startsWith(github.ref, 'refs/heads/pull-request/')
-        uses: nv-gha-runners/get-pr-info@main
 
       - name: Has lts label
         id: has-lts-label
--- a/tests/unit_tests/distributed/megatron_fsdp/test_mcore_fully_sharded_data_parallel.py
+++ b/tests/unit_tests/distributed/megatron_fsdp/test_mcore_fully_sharded_data_parallel.py
@@ -623,6 +623,7 @@ class TestMegatronFSDPE2E:
 
         return outputs
 
+    @pytest.mark.flaky_in_dev
     @pytest.mark.skipif(
         not is_torch_min_version("2.4.0"), reason="Test needs to be updated for torch >= 2.4.0"
     )


megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -337,6 +342,9 @@ def create_cuda_graphs(self, reset_context: bool = True):
 
         self.capture_stats = capture_stats
 
+        if HAVE_TORCH_MEMORY_SAVER:
```
wdykas, line 346: why do you pause here?

tdene:
For now, can we preserve both the UVM and torch memory saver functionality, since it seems like very minimal support to keep both?

You'd have to make the torch memory saver context not be used if UVM is turned on by directly changing this block in `dyamic_context.py`:
```
        ctx_manager = (
            torch.cuda.use_mem_pool(self.unified_memory_mempool)
            if self.unified_memory_level > 0
            else nullcontext()
        )
```
to make the `else` portion be `torch_memory_saver.region(tag="inference_context", enable_cpu_backup=True)`

And you'd have to add to the if statement in the argument validation:
```
    if args.rl_offload_kv_cache_during_training and not args.inference_dynamic_batching_unified_memory_level:
        try:
            from torch_memory_saver import torch_memory_saver
        except ImportError:
            raise AssertionError("To use offload-kv-cache-during-training, `torch_memory_saver` must be installed. See https://github.com/fzyzcjy/torch_memory_saver.")
```

EDIT: Mainly because I think we should run perf benchmarks with both options to check there's no unexpected behavior.

megatron/core/inference/engines/dynamic_engine.py
mathemakitten: When we enter inference mode in `megatron_rl_inference_mode` we always resume the tensor (which makes sense on iterations >= 2). To avoid special-casing the first iteration, we pause it after inference graphs are created. This was previously handled by the `.cuda()` call being a no-op.

megatron/training/arguments.py
```diff
@@ -385,6 +385,14 @@ def validate_args(args, defaults={}):
     total_model_size = args.tensor_model_parallel_size * args.pipeline_model_parallel_size * args.context_parallel_size
     args.data_parallel_size = args.world_size // total_model_size
 
+    # Assert that `torch_memory_saver` is installed if offloading KV cache during RL.
+    if args.rl_offload_kv_cache_during_training:
```
wdykas, line 332: Probably want to add import check somewhere in core as well since its used there in the context.

mathemakitten: We also do a check in `megatron/core/inference/contexts/dynamic_context.py`!

https://github.com/NVIDIA/Megatron-LM/pull/3048/files#diff-aa4eebdd4ac24bc9e7937775652f49028a0c648aab781dcc8ff0f243a4abaf1cR61

wdykas: yeah but shouldn't we raise an error if someone thinks we are using it? Seems like secretly moving to a null context if its not available will give users unexpected behavior.

mathemakitten: I've added an `assert not self.offload_kv_cache and self.unified_memory_level` in `megatron/core/inference/contexts/dynamic_context.py` for core. The check for `self.offload_kv_cache` is now sufficient; we need to keep the nullcontext path because we use it in the unified memory case.

Issue #2824, opened 2026-01-06
Feature Request: μP (Maximal Update Parameterization)


sbhavani:
## Summary
Request to add support for [μP (Maximal Update Parameterization)](https://arxiv.org/abs/2203.03466) to enable hyperparameter transfer across model scales.

## Motivation
μP allows tuning hyperparameters on small models and transferring them to larger models, reducing the cost of large-scale pretraining. It has been adopted in production models like [Falcon-H1](https://falcon-lm.github.io/blog/falcon-h1/).

## Requested Features
1. **Per-layer Learning Rate Scaling** - Different LR multipliers for specific layers
2. **Width-dependent Initialization** - μP-style init scaling based on model width

## References
- [MuP Paper](https://arxiv.org/abs/2203.03466)
- [Falcon-H1](https://falcon-lm.github.io/blog/falcon-h1/)

janEbert:
Other references to consider for automatic hyperparameter transfer:
- [Modular norm](https://arxiv.org/abs/2405.14813)
- [Modular norm 2](https://arxiv.org/abs/2409.20325)
- [Modular norm 3](https://arxiv.org/abs/2410.21265)
- [Scion optimizer (Muon generalization)](https://arxiv.org/abs/2502.07529)

BoxiangW:
Link to the older issue https://github.com/NVIDIA/Megatron-LM/issues/1263

Skylion007:
Ditto: would be great to add. We have an old fork here that tries to add it, but would be good to expand and properly integrate/test it: https://github.com/LLM360/Megatron-LM-muP

Skylion007:
> Other references to consider for automatic hyperparameter transfer:
> 
>     * [Modular norm](https://arxiv.org/abs/2405.14813)
> 
>     * [Modular norm 2](https://arxiv.org/abs/2409.20325)
> 
>     * [Scion optimizer (Muon generalization)](https://arxiv.org/abs/2502.07529)

Also two other more recent references for parameter transfer:
https://arxiv.org/abs/2505.01618
https://arxiv.org/abs/2512.22382

plugyawn:
Added a PR to address this. The HP transfer is tested on a small dataset, 4-layer transformer, on an A100. Plots and experiment details attached.

BoxiangW:
Help merging https://github.com/NVIDIA/Megatron-LM/pull/3058

PR #3058, opened 2026-01-23
μP: Maximal Update Parameterization 

plugyawn:
# What does this PR do ?
Adds support for Maximal Update Parameterization (μP) for optimal hyperparameter transfer across model widths. 
Addresses issue #2824 opened by @sbhavani.

The idea is to train multiple high-depth, low-width models to recover optimal HPs (i.e, reduced hidden_size), and then transfer to high-width models (i.e, high hidden_size). 

Automatic initialization scaling (σ / √(width_mult) for hidden layers) and automatic LR scaling: lr / width_mult for hidden layers (Adam only, not SGD) is also implemented. Embedding/output layers use base LR (no scaling), as in the original TP-V paper.

References:
  - https://arxiv.org/abs/2203.03466

Tagging the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall)

Functional tests and documentation in progress, unit tests added.
Some doubts: in the param_and_grad_buffer, I added an `is_embedding_parameter` even though there exists an  `is_embedding_or_output_parameter`. In TP-V, the fan-in of the output layer is interpreted as infinite-width, inlike the embedding layer, which has the fixed vocabulary (both according to the paper and [mutransformers](https://github.com/microsoft/mutransformers)). There seem to be conflicts about the case of Tied Embeddings (embedding and output layer share weights, see this [discussion](https://github.com/microsoft/mup/issues/55).

Some plots:

The following plots show the current functioning. In the first image, please note that MuP is on a different Y-axis scale than SP. In the second, I believe training more longer would make the MuP sharing optimal LR much clearer (it's currently on 500 steps). I only have access to an A100 at the moment, so these are character-level transformers trained on enwiki8. 

image
image
image

Experiment details:
| Parameter            | Default                         | Paper Reference                          |
|----------------------|----------------------------------|-------------------------------------------|
| `--widths`           | 128,256,512,1024,2048,4096,8192 | MuP paper Fig. 1                           |
| `--base-hidden-size` | 128                              | Base model for `width_mult`                |
| `--num-layers`       | 4                                | Transformer depth                          |
| `--lr-sweep-steps`   | 500                              | Steps per (width, LR) run                  |
| `--num-seeds`        | 3                                |  -                 |
| `--seq-len` (for LR sweep) | 128 | - | 
| `--batch_size` | 4 | - |

Plotting code can be accessed on fork `feature/mup-implementation` on my [fork of the code](https://github.com/plugyawn/Megatron-LM/tree/feature/mup-implementation). 

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [ ] I have added relevant functional tests [incoming]

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -149,14 +149,8 @@ def get_mup_config_overrides(
             "override those parameters."
         )
         if is_adam_optimizer:
-            message += (
-                " MuP Adam epsilon scaling remains applied to hidden matrix-like parameters."
-            )
-        log_single_rank(
-            logger,
-            logging.WARNING,
-            message,
-        )
+            message += " MuP Adam epsilon scaling remains applied to hidden matrix-like parameters."
+        log_single_rank(logger, logging.WARNING, message)
 
     if mup_width_mult == 1.0:
         # No scaling needed when width_mult is 1
@@ -188,9 +182,7 @@ def get_mup_config_overrides(
             return False
         return not is_vector_like_parameter(param, param_name)
 
-    def should_scale_vector_like_lr_with_mup(
-        param: torch.nn.Parameter, param_name: str
-    ) -> bool:
+    def should_scale_vector_like_lr_with_mup(param: torch.nn.Parameter, param_name: str) -> bool:
         if decoupled_lr_enabled and getattr(param, 'is_embedding_or_output_parameter', False):
             return False
         return is_vector_like_parameter(param, param_name)
@@ -236,15 +228,13 @@ def get_mup_config_overrides(
     if decoupled_lr_enabled:
         if lr_override:
             hidden_predicate = ParamWithNamePredicate(
-                name="mup_hidden_only_excluding_embedding_output",
-                fn=should_scale_lr_with_mup,
+                name="mup_hidden_only_excluding_embedding_output", fn=should_scale_lr_with_mup
             )
             mup_overrides[ParamKey(with_name_predicate=hidden_predicate)] = lr_override
 
         if eps_override:
             hidden_output_predicate = ParamWithNamePredicate(
-                name="mup_hidden_only_for_adam_eps",
-                fn=should_scale_eps_with_mup,
+                name="mup_hidden_only_for_adam_eps", fn=should_scale_eps_with_mup
             )
             mup_overrides[ParamKey(with_name_predicate=hidden_output_predicate)] = eps_override
     else:
@@ -253,8 +243,7 @@ def get_mup_config_overrides(
         combined_override.update(eps_override)
         if combined_override:
             hidden_output_predicate = ParamWithNamePredicate(
-                name="mup_hidden_and_output",
-                fn=should_scale_eps_with_mup,
+                name="mup_hidden_and_output", fn=should_scale_eps_with_mup
             )
             mup_overrides[ParamKey(with_name_predicate=hidden_output_predicate)] = combined_override
 
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
-import math
 import logging
+import math
 import warnings
 from dataclasses import dataclass, field
 from typing import Callable, List, Literal, Optional, Tuple, Union


plugyawn:
Note that this is just width-MuP (the original paper). There's also a new depth-MuP (which would help with not having to train "skinny" models with low-width/high-depth for transfer). A new paper, Complete(d)-P also exists, that I've not entirely gone through.

plugyawn:
Hi! @sbhavani, could you take a look?

sbhavani:
@plugyawn thanks for the contribution! Please bear with us as this will take some time to review since it touches a lot of areas in core

plugyawn:
Thank you, @sbhavani and the team!

megatron/core/transformer/transformer_config.py
```diff
@@ -314,6 +316,50 @@ class TransformerConfig(ModelParallelConfig):
...
+    """
+
+    mup_embedding_mult: float = 1.0
+    """
+    Multiplier for embedding layer output. Applied after the embedding lookup.
+    Default: 1.0 (no scaling).
+    """
+
+    mup_output_mult: float = 1.0
+    """
```
KyroChi, line 383: I don't think that this scaling actually occurs anywhere, so by default the outputs will be m times larger than we would expect. Maybe I am missing something?

plugyawn: My bad. Fixed.

plugyawn:
image

> You don't actually plot the minima in the figure of val loss vs lr on the right. Sometimes if not properly implemented you can in fact see the optimum shifting, but we won't be able to tell from this plot since we don't see the minima.

Plotted with the optimum marked. It's on log scale... so the shifted minima is,
| Width | SP opt LR | MuP opt LR |
  |-------|-----------|------------|
  | 128   | 9.77e-04  | 9.77e-04   |
  | 256   | 4.88e-04  | 9.77e-04   |
  | 512   | 2.44e-04  | 9.77e-04   |
  | 1024  | 1.22e-04  | 9.77e-04   |
  | 2048  | 6.10e-05  | 1.95e-03   |
Also 16 instead of 11 LRs, since the minima wasn't clear.

The shift in 2048 is worrying... but the minimum loss looks closeby.

Getting back with coordinate checks soon.

plugyawn:
image
Normalized logits. T is gradient steps after training.

KyroChi:
I don't think that the optimum shift for the 2048 model is concerning for two reasons:

1. The loss is almost identical, we expect mup to only hold on average, which means that occasionally the empirical optimum will shift a little bit between runs. The curvature of the lr vs. loss curves usually decreases as the model size increases which only adds to this issue.
2. The 2048 models are almost certainly undertrained which will ALWAYS favor a larger learning rate. 

Regarding this latter point, transfer is really only expected to occur at fixed TPP or something, but due to the asymptotic properties of mup we can usually just get away with optimally training our largest model and overtraining smaller models to demonstrate mu-transfer. You see only 4*128*500 =256,000 tokens during training, which is ~1 TPP for the smallest model and ~0.003 TPP for the largest model 😝 In my experience you usually need at least 2TPP to get good transfer plots, so in some sense this is already better than I would expect!

Regarding the residual power dependency in the logits plot: this could be because Megatron's default optimizer Adam eps is 10^{-8}, which is actually rather high for mup. Since your models are pretty small you can probably get away with setting this to 10^{-12} or even 10^{-15} and see if the coordinate check flattens. This is a quirk of all mup implementations unfortunately and is unlikely to indicate a bug IMO. See also [this paper](https://arxiv.org/abs/2505.01618).

BoxiangW:
Hi @plugyawn, thanks for your contribution here, I was actually implementing this on my branch but since you have all the experiments result ready, we can try merge your PR!

megatron/core/optimizer/__init__.py
```diff
@@ -105,6 +105,64 @@ def get_standard_config_overrides(config: OptimizerConfig) -> Dict[ParamKey, Par
...
+
+    Returns:
+        Dict[ParamKey, ParamGroupOverride]: MuP config overrides for hidden layers.
+    """
+    if mup_width_mult == 1.0:
+        # No scaling needed when width_mult is 1
+        return {}
+
+    # SGD is width-invariant, no scaling needed.
+    if optimizer_type.lower() == 'sgd':
```
BoxiangW: I think SGD also need to scale LM_HEAD by `1.0 / mup_width_mult`

megatron/core/transformer/transformer_config.py
BoxiangW: My understanding for `output_mult` is that they should be the same as `mup_width_mult` please correct me

megatron/core/transformer/transformer_config.py
```diff
@@ -314,6 +316,50 @@ class TransformerConfig(ModelParallelConfig):
...
+    Multiplier for output logits before softmax. Scaled by 1/mup_width_mult when
+    MuP is enabled to ensure output variance stability across widths.
+    Default: 1.0 (no scaling).
+    """
+
+    mup_attn_scale_power: float = 1.0
```
BoxiangW, line 398: Can you help remind me on this scaling?

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -2583,6 +2540,7 @@ def set_param_attribute():
                             "partition_dim",
                             "partition_stride",
                             "is_embedding_or_output_parameter",
+                            "is_embedding_parameter",
```
BoxiangW, line 2630: Can we actually just check the name `embedding.word_embeddings.weight`

megatron/core/transformer/transformer_config.py
```diff
@@ -1532,14 +1595,29 @@ def __post_init__(self):
                 self.embedding_init_method = self.init_method
 
         if self.init_method is None:
-            self.init_method = init_method_normal(self.init_method_std)
+            if self.use_mup:
+                # MuP: scale std by 1/sqrt(width_mult).
+                self.init_method = mup_init_method_normal(
+                    self.init_method_std, self.mup_width_mult
+                )
```
BoxiangW:
```suggestion
                self.init_method = init_method_normal(self.init_method_std/sqrt(self.mup_width_mult))
```

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
plugyawn: What if models have different layer names? Might be more brittle this way?

plugyawn:
> I don't think that the optimum shift for the 2048 model is concerning for two reasons:
> 
> 1. The loss is almost identical, we expect mup to only hold on average, which means that occasionally the empirical optimum will shift a little bit between runs. The curvature of the lr vs. loss curves usually decreases as the model size increases which only adds to this issue.
> 2. The 2048 models are almost certainly undertrained which will ALWAYS favor a larger learning rate.
> 
That makes sense! Thank you!

> Regarding this latter point, transfer is really only expected to occur at fixed TPP or something, but due to the asymptotic properties of mup we can usually just get away with optimally training our largest model and overtraining smaller models to demonstrate mu-transfer. You see only 4_128_500 =256,000 tokens during training, which is ~1 TPP for the smallest model and ~0.003 TPP for the largest model 😝 In my experience you usually need at least 2TPP to get good transfer plots, so in some sense this is already better than I would expect!
I did _not_ know we had good heuristics for when transfer happens! That makes sense!
I had some intuition, of course, that it must take _some_ time to saturate, but I thought 256,000 should be close to enough... given the smallness of the vocabulary? 

I was also reminded of https://arxiv.org/abs/2501.16975 and their scaling laws over vocabulary (min. loss decreases with log vocab), although that might be unrelated.

>
> Regarding the residual power dependency in the logits plot: this could be because Megatron's default optimizer Adam eps is 10^{-8}, which is actually rather high for mup. Since your models are pretty small you can probably get away with setting this to 10^{-12} or even 10^{-15} and see if the coordinate check flattens. This is a quirk of all mup implementations unfortunately and is unlikely to indicate a bug IMO. See also [this paper](https://arxiv.org/abs/2505.01618).

That makes sense!


Hahaha and also thanks this sent me into a rabbit hole, learned quite a few things!

megatron/core/transformer/transformer_config.py
plugyawn: It's just a way to switch between sqrt(d) softmax scaling (usual) and (d) scaling in MuP. Kept it to test between different scales if ablations are needed somewhere. It defaults to the MuP (d) if not set. Could remove it! It does feel a bit redundant.

For clarity, MLM already had the softmax_scale flag.

megatron/core/transformer/transformer_config.py
```diff
@@ -314,6 +316,50 @@ class TransformerConfig(ModelParallelConfig):
...
+    This value is automatically computed in __post_init__ when use_mup is enabled.
+    """
+
+    mup_base_hidden_size: Optional[int] = None
+    """
+    Base hidden size for MuP width scaling. This is the reference width from which
+    scaling factors are computed. Defaults to hidden_size if not specified (base model
+    case where width_mult=1.0). Set this to your base/proxy model's hidden size when
+    scaling up.
+    """
```
janEbert, line 374: I'd suggest also adding a `mup_base_head_dim`. That would allow at least the base model to be the same as its non-μP version. I found this immensely helpful to ensure compatibility with the existing ecosystem back when I used μP.

It would use the same trick as in the original μP repo README:
https://github.com/microsoft/mup/blob/19814971934ef91dd546f88e913fc963e096d11c/README.md#L71-L76

In this codebase's case, the `softmax_scale` you define below would be changed like this:
```patch
             # MuP attention scaling: 1/d_head instead of 1/sqrt(d_head).
             if self.softmax_scale is None:
-                self.softmax_scale = 1.0 / (self.kv_channels ** self.mup_attn_scale_power)
+                self.softmax_scale = self.mup_base_head_dim / (self.kv_channels ** self.mup_attn_scale_power)
```

Please double-check whether this makes sense, it's been some time since I used μP.

plugyawn: Gotcha, makes sense. Making these changes now.

megatron/core/optimizer/__init__.py
plugyawn: Oh; I was of the impression that the normalization was Adam/AdamW only. 
image
This is from Appendix B of the TP-V paper.

megatron/core/transformer/transformer_config.py
plugyawn: Added!

janEbert: Thank you! And sorry, I messed up the math in my suggestion, it should be `self.softmax_scale = self.mup_base_head_dim**0.5 / (self.kv_channels ** self.mup_attn_scale_power)`.

I.e., in your code,
```patch
                 base_head_scale = (
-                    1.0 if self.mup_base_head_dim is None else self.mup_base_head_dim
+                    1.0 if self.mup_base_head_dim is None else self.mup_base_head_dim**0.5
                 )
```

That way, you have `sqrt(d)/d = 1/sqrt(d)` for `d = base_head_dim`, ensuring non-μP-compatibility for that specific value.

janEbert: (Same applies to the nice docstring you added for the argument.)

janEbert: Thanks a lot for fixing it! One minor nit: the description in `megatron/training/arguments.py` would also need to be changed. :)

janEbert: Awesome, looks great!

plugyawn: Cool! Thanks!

plugyawn:
@chtruong814 I think all the review comments have been addressed.

BoxiangW:
Hi @plugyawn, that tag is automated, but can you help merge main and resolve conflicts? thx

plugyawn:
Got it! I'll also write out the functional tests.

plugyawn:
Just realized the code doesn't handle Mamba-based models. Added in a warning flag there.

plugyawn:
Updated functional tests. Checked for merge conflicts, there appear to be none.

plugyawn:
Fixed some semantic conflicts with optimizer-grouping cases (`shared_embeddings` params are now treated as embedding-class, so they use the Base LR now. In case of `decoupled_LR` and MuP being enabled at the same time, MuP now applies only to hidden layers).  

@BoxiangW, could you retrigger the CI?

plugyawn:
Ran some more tests on my side. Seems good.

plugyawn:
@chtruong814 a little unsure what the next changes should be...

plugyawn:
@chtruong814 @BoxiangW any changes that are required from my side?

BoxiangW:
@plugyawn, sorry for the confusion, this tag is automated and we are waiting for final reviews. Just pinged reviewers again

megatron/core/transformer/transformer_config.py
```diff
@@ -1594,14 +1675,29 @@ def __post_init__(self):
...
+                    self.num_layers,
+                    self.mup_width_mult,
+                    multiplier=2.0 if not self.is_hybrid_model else 1.0,
+                )
+            else:
+                self.output_layer_init_method = scaled_init_method_normal(
+                    self.init_method_std,
+                    self.num_layers,
+                    multiplier=2.0 if not self.is_hybrid_model else 1.0,
+                )
```
janEbert, line 1734: Could we add a warning that complains about `mup` possibly breaking when `self.use_mup is True` and either one of `self.init_method` or `self.output_layer_init_method` is not `None`?

janEbert:
From [muTransfer](https://arxiv.org/abs/2203.03466) Appendix B.3:

> All of our derivations here assume _ϵ_ is negligible in Adam. If it is set to a non-negligible number, then it needs to be scaled, for all parameters, like 1/`fan_in`² if it is added before the square root, or like 1/`fan_in` if it is added after the square root.

However, it has been shown (in [Small-scale proxies for large-scale Transformer training instabilities](https://arxiv.org/abs/2309.14322), Section 3.4) that ϵ is _not_ negligible and definitely needs to be scaled.

Thus, could you please add Adam ϵ scaling as well? This really impacts training stability when scaling to larger models.

janEbert:
I'd also love for us to have functional tests in some way or form; either coordinate checks or an actual optimal LR transfer test. However, I don't think we need the tests to be included in this PR.

plugyawn:
Re: [Small-scale proxies for large-scale Transformer training instabilities](https://arxiv.org/abs/2309.14322): thank you! Didn't know! Will make the changes, as you said.

janEbert:
Thanks a lot for keeping at this!

megatron/core/transformer/transformer_config.py
plugyawn: Yep, done!

plugyawn:
@janEbert made the changes! 

Also: there was a certain case where if there's a YAML config without the mup tag it crashes; patched that as well.

megatron/core/optimizer/__init__.py
```diff
@@ -162,30 +165,56 @@ def is_embedding_parameter(param: torch.nn.Parameter, param_name: str) -> bool:
...
+    else:
+        combined_override: ParamGroupOverride = {}
+        combined_override.update(lr_override)
+        combined_override.update(eps_override)
+        if combined_override:
+            hidden_output_predicate = ParamWithNamePredicate(
+                name="mup_hidden_and_output",
+                fn=should_scale_eps_with_mup,
+            )
+            mup_overrides[ParamKey(with_name_predicate=hidden_output_predicate)] = combined_override
```
janEbert: You special-case SGD already earlier in the function, but just for future-proofing, we should also handle potential non-Adam optimizers better here. The `optimizer_config` already has a pre-defined (albeit deprecated) `adam_eps` attribute, which is `not None` by default. So it could be that an `eps` of a non-Adam optimizer is modified/passed, even though it should not be modified/passed. E.g., for Muon, we don't want to pass `eps` for non-Embedding parameters.

plugyawn: Will do; I think a follow-up with #3177 would be necessary too. I'll make the changes, should be short.

megatron/core/transformer/transformer_config.py
```diff
@@ -1594,14 +1690,29 @@ def __post_init__(self):
...
-                multiplier=2.0 if not self.is_hybrid_model else 1.0,
-            )
+            if self.use_mup:
+                # MuP: depth and width scaling for output layers.
+                self.output_layer_init_method = mup_scaled_init_method_normal(
+                    self.init_method_std,
+                    self.num_layers,
+                    self.mup_width_mult,
+                    multiplier=2.0 if not self.is_hybrid_model else 1.0,
+                )
```
janEbert, line 1728: Sorry, maybe I'm going crazy; shouldn't output weights have constant init variance according to Table 8?

megatron/core/optimizer/__init__.py
```diff
@@ -105,6 +105,118 @@ def get_standard_config_overrides(config: OptimizerConfig) -> Dict[ParamKey, Par
...
+) -> Dict[ParamKey, ParamGroupOverride]:
+    """Get MuP config overrides for per-layer LR and Adam epsilon scaling.
+
+    In MuP, hidden layer learning rates are scaled by 1/width_mult to ensure
+    that parameter updates remain O(1) as model width increases. This enables
+    hyperparameter transfer from a smaller proxy model to a larger target model.
+
+    MuP optimizer scaling rules (as implemented here):
+    - Adam/AdamW:
+      - hidden+output lr = base_lr / width_mult
```
janEbert: Again, maybe you have a better understanding here. But shouldn't input and output weights match here according to Table 8? I.e., symmetric init std and LR scaling between input and output layers?

plugyawn: Sorry, genuinely overlooked this; these should be fixed now. I spent last night regenerating the plots to make sure this is resolved now. 

As expected, on the tied embeddings/output case, they behave the same:
image


(before and after the correction)

The untied case was going wrong before(!), but it works better now. 
image

These are over 20 steps on batch size=20, seq_length=256.

janEbert: Oof, I'm so happy I didn't cause you any redundant extra work. :sweat_smile:
Super nice, thank you so, so much for all the effort!

plugyawn: Not at all! Sorry for overlooking this in the first place!

plugyawn:
@janEbert @chtruong814 I think the remaining review threads have been resolved too.

plugyawn (commented):
Thank you! Very excited to have this merged!

janEbert:
1. @plugyawn could you quickly comment on [this](https://github.com/NVIDIA/Megatron-LM/pull/3058#discussion_r2846479854)?

  > Sorry, maybe I'm going crazy; shouldn't output weights have constant init variance according to Table 8?

2. And similarly, SGD LR isn't currently adjusted but (again referring to Table 8), it should be, right?

megatron/core/transformer/transformer_config.py
plugyawn: Should be handled by fa813b4, I think.

janEbert: Oh I'm sorry, I got confused by the naming; I thought this init method was applied to the final layer of the model instead of to the final layer of each sub-block.

megatron/core/models/bert/bert_model.py
```diff
@@ -141,7 +141,11 @@ def __init__(
                 config.hidden_size,
                 self.vocab_size,
                 config=config,
-                init_method=config.init_method,
+                init_method=(
+                    config.embedding_init_method
+                    if config.use_mup and not self.share_embeddings_and_output_weights
+                    else config.init_method
```
janEbert, line 141: Why do we use a different initialization here (and for the other models) when embeddings are not tied? Shouldn't it work the same in both settings? I.e., we shouldn't need to special-case unshared embeddings, right?

janEbert: Okay I get it; `embedding_init_method` is what the `init_method` used to be before being adjusted for μP, so we need this conditional to ensure that the same function is used for the output layer that was used to initialize the `Embedding`s as well.

plugyawn: Yep!

janEbert:
Thank you for the endless endurance and spirit! Let's get this merged. :)

janEbert:
Think you just need to rebase onto `main` and apply `tools/autoformat.sh`.

plugyawn:
Reran the plots for SGD as well, to be sure:
image

Autoformat's done, too!

> Thank you for the endless endurance and spirit! Let's get this merged. :)

It was very fun!

BoxiangW:
Just triggered one last CI, I will be merged if all passed, thanks @plugyawn !

PR #3060, opened 2026-01-23
Extract intermediate embeddings of transformer block

sajadn:
Enabling extraction of intermediate embeddings of TransformerBlock. This is particularly useful for diffusion models that use TransformerBlock in [DFM](https://github.com/NVIDIA-NeMo/DFM/tree/main).

### DIFF
--- a/tests/unit_tests/transformer/test_transformer_block.py
+++ b/tests/unit_tests/transformer/test_transformer_block.py
@@ -199,13 +199,83 @@ class TestParallelTransformerBlock:
         for intermediate in intermediate_hidden_states:
             assert intermediate.shape == (sequence_length, micro_batch_size, config.hidden_size)
 
+    def test_feature_extraction_full_checkpoint_block(self):
+        """Test feature extraction with full checkpoint using block method."""
+        transformer_config = self.transformer_config
+        config = transformer_config
+        config.recompute_granularity = 'full'
+        config.recompute_method = 'block'
+        config.recompute_num_layers = config.num_layers  # checkpoint all layers
+        block_transformer_block = TransformerBlock(
+            config, get_gpt_layer_with_transformer_engine_spec()
+        )
+        block_transformer_block.cuda()
+
+        sequence_length = 32
+        micro_batch_size = 2
+
+        # [sequence length, batch size, hidden size]
+        hidden_states = torch.ones((sequence_length, micro_batch_size, config.hidden_size))
+        hidden_states = hidden_states.cuda()
+        hidden_states.requires_grad = True
+
+        attention_mask = torch.ones((1, 1, sequence_length, sequence_length), dtype=bool).cuda()
+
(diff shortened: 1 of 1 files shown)


megatron/core/transformer/transformer_block.py
```diff
@@ -788,6 +842,9 @@ def forward(
         if not self.pre_process and len(self.layers) == 0 and not self.final_layernorm:
             hidden_states = hidden_states.clone()
 
+        if len(extract_layer_indices) > 0:
+            return hidden_states, intermediate_hidden_states
```
yaoyu-33, line 871: how is this working with PP?

sajadn: that's a good question. Currently it extracts the embeddings for the given indices from each chunk. 
I think it would be a bit complex if we want to extract the absolute index of each layer, unless megatron has some mechanism to extract chunk index.

sajadn: @yaoyu-33 I modified the code to work with pipeline parallelism. Please review.

megatron/core/transformer/transformer_block.py
```diff
@@ -563,9 +593,17 @@ def checkpoint_handler(forward_func):
...
+
+                # Feature extraction: collect hidden states at specified global layer indices
+                if (layer_idx + layer_offset) in extract_layer_indices:
+                    intermediate_hidden_states.append(hidden_states)
         else:
             raise ValueError("Invalid activation recompute method.")
 
+        # Return intermediate hidden states if feature extraction was requested
+        if len(extract_layer_indices) > 0:
+            return hidden_states, intermediate_hidden_states
```
yashaswikarnati, line 605: i wonder if this breaks pipelining as the return type is changed

sajadn: that's a fair concern. I verified this with a minimal pp example that you can find here: 
https://gist.github.com/sajadn/3d45bffdfcd9788f515ffee166ac3ad0
the thing is pp_communicator communicate the output of the forward_step not the output of the transformerBlock and within the forward_step i use the extracted embeddings and only return the hidden state as the output.

If you think there is a scenario that may haven't been covered, let me know!

yaoyu-33 (approved):

sajadn:
/ok to test [3610fb0](https://github.com/NVIDIA/Megatron-LM/commit/3610fb05a29f89dc0752780078d0259354250572)

ko3n1g:

You can track the progress here: https://github.com/NVIDIA/Megatron-LM/actions/runs/22004047479

ko3n1g:

You can track the progress here: https://github.com/NVIDIA/Megatron-LM/actions/runs/22006463411

ko3n1g:

You can track the progress here: https://github.com/NVIDIA/Megatron-LM/actions/runs/22007305312

PR #3063, opened 2026-01-23
Inference | KV prefix caching.

lmcafee-nvidia:
# What does this PR do ?

Implement KV cache prefix caching for dynamic batching inference, enabling requests with identical prompt prefixes to share computed KV cache blocks instead of recomputing them.

### Key Features

- **Block-level prefix sharing**: Requests with matching prompt prefixes share KV cache blocks, reducing redundant computation
- **Content-based hashing**: Blocks are identified by hash of their token content for efficient prefix matching
- **Request coordination**: Multi-rank coordination prevents redundant KV computation when multiple requests match the same prefix blocks
- **LRU eviction**: Cached blocks use LRU eviction based on timestamps when memory is constrained
- **CLI control**: New `--no-inference-dynamic-batching-enable-prefix-caching` flag to disable when needed

### Changes

**Block Allocator** (`megatron/core/inference/contexts/dynamic_block_allocator.py`):
- Add block hash tracking and content-based matching
- Implement hash-to-block-id mapping for prefix lookup
- Add reference counting for shared blocks
- LRU eviction for cached blocks

**Dynamic Context** (`megatron/core/inference/contexts/dynamic_context.py`):
- `_find_matching_prefix_blocks()` - Find cached blocks matching request prefix
- `_compute_block_hashes()` - Compute content hashes for blocks
- Block sharing logic in `add_request()`

**Dynamic Engine** (`megatron/core/inference/engines/dynamic_engine.py`):
- Request coordination to avoid redundant computation
- Pending block tracking during prefill
- Integration with chunked prefill scheduling

**Inference Request** (`megatron/core/inference/inference_request.py`):
- Add `enable_prefix_caching` flag
- Track matched prefix blocks per request

**Arguments** (`megatron/training/arguments.py`):
- Add `--no-inference-dynamic-batching-enable-prefix-caching` flag

### How It Works

1. When a request arrives, compute hashes for each block of prompt tokens
2. Look up existing cached blocks with matching hashes
3. Reuse matched blocks (increment ref count) instead of allocating new ones
4. Only compute KV for unmatched (new) tokens
5. After prefill, cache newly computed blocks for future requests
6. Coordinate across DP ranks to prevent redundant computation

## Test plan

- [x] Unit tests for block hash computation and matching (57 tests in `test_dynamic_prefix_caching.py`)
- [x] Unit tests for request coordination
- [x] Tests for LRU eviction
- [x] Tests for prefix caching enable/disable
- [x] Tests for edge cases (empty prompts, single tokens, exact block boundaries)
- [ ] End-to-end inference test with prefix sharing

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/transformer/moe/moe_utils.py
+++ b/megatron/core/transformer/moe/moe_utils.py
@@ -1036,13 +1036,20 @@ def track_moe_metrics(
     """
     # Aux loss logging
     tracker = get_moe_layer_wise_logging_tracker()
-    # Initialize the tracker if force_initialize is True
+    # Initialize the tracker if force_initialize is True.
+    # The values tensor size must match what the router creates in save_to_aux_losses_tracker,
+    # which uses (num_layers + mtp_num_layers). This is important for PP ranks that have no
+    # MoE layers (so the tracker is empty and force_initialize creates the entry); their tensor
+    # size must match ranks that do have MoE layers, otherwise all_reduce across PP will hang.
+    tracker_num_layers = num_layers
+    if mtp_num_layers is not None:
+        tracker_num_layers += mtp_num_layers
     if force_initialize:
         if track_names is not None:
             for key in track_names:
                 if key not in tracker:
                     tracker[key] = {}
-                    tracker[key]["values"] = torch.zeros(num_layers, device="cuda")
+                    tracker[key]["values"] = torch.zeros(tracker_num_layers, device="cuda")
                     tracker[key]["reduce_group"] = None
                     tracker[key]["avg_group"] = None
                     tracker[key]["reduce_group_has_dp"] = False
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -82,7 +82,7 @@ def ensure_test_data():
 
         try:
             # Download assets to /opt/data
-            download_and_extract_asset(assets_dir=str(data_path))
+            download_and_extract_asset(assets_dir=data_path)
 
             print("Test data downloaded successfully.")
 


megatron/rl/inference/megatron.py
```diff
@@ -167,6 +167,7 @@ def get_dynamic_inference_engine(
         cuda_graph_mixed_prefill_count=args.inference_dynamic_batching_cuda_graph_mixed_prefill_count,
         metrics_writer=metrics_writer,
         persist_cuda_graphs=args.rl_training_cuda_graphs,
+        enable_prefix_caching=getattr(args, 'inference_dynamic_batching_enable_prefix_caching', True),
```
yobibyte: Why use getattr instead of just args.inference_dynamic_batching_enable_prefix_caching?

megatron/training/arguments.py
```diff
@@ -1510,6 +1510,13 @@ def _add_inference_args(parser):
                        '1) allocate `memory_buffer` in unified memory. '
                        'Eventually, additional levels will be included to '
                        'control other tensors within the context.')
+    group.add_argument('--inference-dynamic-batching-disable-prefix-caching',
+                       action='store_false',
+                       dest='inference_dynamic_batching_enable_prefix_caching',
+                       help='Disable prefix caching for dynamic batching inference. '
+                       'When disabled, KV cache blocks cannot be shared between '
+                       'requests with identical prompt prefixes.')
+    group.set_defaults(inference_dynamic_batching_enable_prefix_caching=True)
```
tdene: Can we use `BooleanOptionalAction` instead? Something along the lines of
```
group.add_argument(
    '--inference-dynamic-batching-prefix-caching',
    dest='inference_dynamic_batching_enable_prefix_caching',
    action=argparse.BooleanOptionalAction,
    default=True,
    help=(
        'Enable/disable prefix caching for dynamic batching inference. '
        'When disabled, KV cache blocks cannot be shared between requests with '
        'identical prompt prefixes.'
    ),
)
```

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -37,6 +37,30 @@ def __init__(self, context: "DynamicInferenceContext", total_count: int, paused_
             self.total_count, dtype=torch.int32, device=torch.cuda.current_device()
         )
 
+        # Block hash tracking for prefix caching: -1 = uncomputed, positive = valid hash
```
tdene: I think most of these comments are unnecessary because the code style itself is self-documenting.

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +180,170 @@ def reset(self) -> None:
...
+        Called when a request starts using (sharing) existing cached blocks.
+
+        Args:
+            block_ids: Tensor of block IDs to increment.
+        """
+        if block_ids.numel() == 0:
+            return
+        self.block_ref_counts[block_ids] += 1
+
+    def decrement_ref_count(self, block_ids: Tensor) -> None:
```
tdene: This does not merit having its own helper method.

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +180,170 @@ def reset(self) -> None:
...
+        signaling that the KV cache for this block is ready for reuse.
+
+        Args:
+            block_id: The block ID to mark as computed.
+        """
+        if block_id in self._pending_block_hashes:
+            hash_value = self._pending_block_hashes.pop(block_id)
+            self.set_block_hash(block_id, hash_value)
+
+    def increment_ref_count(self, block_ids: Tensor) -> None:
```
tdene: This does not merit having its own helper method.

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +180,170 @@ def reset(self) -> None:
...
+        self.block_hashes.fill_(-1)
+
+        # Reset prefix caching state
+        self.hash_to_block_id.clear()
+        self._pending_block_hashes.clear()
+        self.block_ref_counts.fill_(0)
+        self.block_timestamps.fill_(0)
+        self.global_timestamp = 0
+
+    def set_block_hash(self, block_id: int, hash_value: int) -> None:
```
tdene: This does not merit having its own helper method.

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +180,170 @@ def reset(self) -> None:
...
+        """Set the hash for a specific block.
+
+        Args:
+            block_id: The block ID to set hash for.
+            hash_value: The hash value to store.
+        """
+        self.block_hashes[block_id] = hash_value
+
+    def get_block_hash(self, block_id: int) -> int:
```
tdene: This does not merit having its own helper method.

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +180,170 @@ def reset(self) -> None:
...
+        Returns:
+            Hash value (-1 if not computed).
+        """
+        return self.block_hashes[block_id].item()
+
+    # =========================================================================
+    # Prefix caching methods
+    # =========================================================================
+
+    def lookup_block_by_hash(self, block_hash: int) -> Optional[int]:
```
tdene: This does not merit having its own helper method.

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +180,170 @@ def reset(self) -> None:
...
+
+    def update_timestamps(self, block_ids: Tensor) -> None:
+        """Update LRU timestamps for accessed blocks.
+
+        Args:
+            block_ids: Tensor of block IDs that were accessed.
+        """
+        if block_ids.numel() == 0:
+            return
+        self.global_timestamp += 1
```
tdene: I don't agree that the global timestamp should be incremented every time an `update_timestamps` operation is run on some particular blocks. This should track the step count.

Right now, the global timestamp gets incremented just from `add_request` and that's very counter-intuitive.

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +180,170 @@ def reset(self) -> None:
...
+            return
+        self.block_ref_counts[block_ids] -= 1
+
+    def update_timestamps(self, block_ids: Tensor) -> None:
+        """Update LRU timestamps for accessed blocks.
+
+        Args:
+            block_ids: Tensor of block IDs that were accessed.
+        """
+        if block_ids.numel() == 0:
```
tdene: I don't think we need this check, and if that is true, this becomes a 1-line helper method which means it does not merit having its own helper method.

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +180,170 @@ def reset(self) -> None:
...
+        self.block_timestamps[block_ids] = self.global_timestamp
+
+    def get_evictable_block_count(self) -> int:
+        """Get count of cached blocks that can be evicted (ref_count == 0, hash set).
+
+        Returns:
+            Number of evictable cached blocks.
+        """
+        cached_mask = (self.block_ref_counts == 0) & (self.block_hashes != -1)
+        return cached_mask.sum().item()
```
tdene: In general, I think running `.item()` on a GPU computation result inside a helper method is very bad. We want to be able to enqueue GPU ops and not block on them.

However, specifically for this helper method, I don't think this programming pattern is dangerous right now.

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +180,170 @@ def reset(self) -> None:
...
+            return False  # Not enough cached blocks to evict
+
+        # Sort by timestamp (ascending = oldest first)
+        cached_timestamps = self.block_timestamps[cached_block_ids]
+        sorted_indices = torch.argsort(cached_timestamps)
+        blocks_to_evict = cached_block_ids[sorted_indices[:num_blocks_needed]]
+
+        # Remove from hash mapping
+        for block_id in blocks_to_evict:
+            block_id_int = block_id.item()
```
tdene: Can we vectorize this?

megatron/core/inference/inference_request.py
```diff
@@ -56,6 +56,39 @@ class Status(Enum):
...
+
+    Args:
+        parent_hash: Hash of parent block (0 for first block in sequence).
+        token_ids: Token IDs in this block, shape [block_size_tokens].
+
+    Returns:
+        Positive integer hash value (1 to HASH_PRIME).
+    """
+    block_size = token_ids.shape[0]
+    positions = torch.arange(block_size, device=token_ids.device, dtype=torch.int64)
```
kvareddy: Since block size is fixed, we can precompute positions and powers.

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -37,6 +37,30 @@ def __init__(self, context: "DynamicInferenceContext", total_count: int, paused_
...
         )
 
+        # Block hash tracking for prefix caching: -1 = uncomputed, positive = valid hash
+        self.block_hashes = torch.full(
+            (self.total_count,), -1, dtype=torch.int64, device=torch.cuda.current_device()
+        )
+
+        # Prefix caching data structures
+        # Hash-to-block mapping for O(1) prefix lookup
+        self.hash_to_block_id: Dict[int, int] = {}
```
kvareddy: should this be [int64, int]?

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: agreed, and now it tracks engine.step_count

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -75,43 +109,78 @@ def get_paused_avail(self):
...
+        if not self.enable_prefix_caching:
+            return False
+        if not self.block_evict_lru:
+            return False  # RZ: no cached blocks to evict
+        # Also count evictable cached blocks
+        evictable_count = self.get_evictable_block_count()
+        return (self.total_avail + evictable_count) >= num_blocks
 
-    def allocate_memory_blocks(self, num_blocks: int) -> Optional[Tensor]:
+    def allocate_memory_blocks(self, num_blocks: int, step_count: int = 0) -> Optional[Tensor]:
```
tdene: Should we just have `step_count` live in the context at this point?

It will simplify a lot of the logic. No one will need to pass it around in function headers, because everyone who wants it can just do `context.step_count`, unlike in the current situation, where the engine is at the top and no one else can run `engine.step_count`.

And I think it makes sense for the context to own it at this point.

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: This method is called from 2 places, and the method itself has grown to a few lines since we now handle 2 different eviction modes (LRU and ref_count==0), so I'll keep this as a method for now.

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: I agree, I was just reluctant to make that change. But I'll do it since you also think it's a good idea.

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: these are Python data types, so there is not int64, Python's int can handle the precision of torch's int64

megatron/core/inference/inference_request.py
lmcafee-nvidia: yep, fixed this by making `powers` a global variable (`_hash_powers`)

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -75,43 +109,78 @@ def get_paused_avail(self):
...
         Return:
             None
         """
-        num_blocks = blocks.size(dim=0)
-        self.block_bag[self.total_avail : (self.total_avail + num_blocks)] = blocks
-        self.total_avail += num_blocks
+        if blocks.numel() == 0:
+            return
+
+        if self.enable_prefix_caching:
```
kvareddy, line 182: What happens when prefix caching is not enabled?

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +202,130 @@ def reset(self) -> None:
...
+        num_blocks = block_ids.numel()
+        if num_blocks == 0:
+            return
+
+        # Remove from hash mappings
+        for block_id in block_ids:
+            block_id_int = block_id.item()
+
+            # Clean up pending hash if block was pending computation
+            if block_id_int in self._pending_block_hashes:
```
kvareddy: Why would we deregister pending blocks? all pending blocks need to be computed before even considered for deregistering, right?

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +202,130 @@ def reset(self) -> None:
...
+        Args:
+            block_ids: Tensor of block IDs to deregister.
+        """
+        num_blocks = block_ids.numel()
+        if num_blocks == 0:
+            return
+
+        # Remove from hash mappings
+        for block_id in block_ids:
+            block_id_int = block_id.item()
```
kvareddy: This could add lots of GPU/CPU syncs. Can we avoid loops?

lmcafee-nvidia: just vectorized all these syncs into a single `block_ids_list = block_ids.cpu()` before this loop

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1559,10 +1653,21 @@ def add_request(self, req: DynamicInferenceRequest, chunk_length: Optional[int]
...
             + req.sampling_params.num_tokens_to_generate
         )
-        if num_blocks_needed > 0:
+
+        # Assign blocks: matched blocks first, then newly allocated blocks
+        # For new requests: matched blocks at [0:num_matched], new at [num_matched:...]
+        # For chunked continuations: already_allocated at [0:already], new at [already:...]
+        new_block_start = already_allocated_blocks + num_matched_blocks
+        if num_matched_blocks > 0:
+            matched_tensor = torch.tensor(
```
kvareddy: matched_tensor is defined earlier. can we reuse it?

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: vectorized this to be before the loop

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1596,6 +1701,21 @@ def add_request(self, req: DynamicInferenceRequest, chunk_length: Optional[int]
...
+        # Register hashes for completely filled blocks (skip matched blocks)
+        if self.enable_prefix_caching and req.precomputed_block_hashes is not None:
+            total_tokens_after = req.finished_chunk_token_count + chunk_length
+            num_complete_blocks = total_tokens_after // self.block_size_tokens
+            previously_complete = req.finished_chunk_token_count // self.block_size_tokens
+
+            # Start from the first non-matched block
+            first_block_to_hash = max(previously_complete, num_matched_blocks)
+
+            for block_pos in range(first_block_to_hash, num_complete_blocks):
```
kvareddy: We have .item() in a loop here too.

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: yep, done

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: vectorized before loop to `block_id_list = ...`

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: great catch, thanks. this was a bug because I initially implemented the code to be prefix-caching only, and later added the arg to control prefix caching, and this non-prefix-caching block wasn't re-added.

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: I think these comments add a bit of clarity beyond the variable names alone, so leaving for now.

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: removed `.item()`, and deferred this sync to `is_memory_available()`.

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: for any given block id, the block's hash is stored in 1 of 2 places: 1) `_pending_block_hashes` if the block is pending, and 2) `block_hashes` if the block has been computed (`block_hashes` is initialized to -1 until the block is no longer pending). after computing the block, the block hash is removed from `_pending_block_hashes` and stored in `block_hashes`. so this code block is all about maintaining consistency between `_pending_block_hashes`, `block_hashes`, and `hash_to_block_id`, and not about deregistering a block that is depended on by waiting requests. if pending block hashes aren't deregistered, then `hash_to_block_id` will contain mappings of old hash -> old block id, which another matching request would use and find garbage data in that block. so this loop is really about iterating block ids, and then figuring out if the correct block hash is stored in `block_hashes` or `_pending_block_hashes`.

lmcafee-nvidia: also, it's okay if dependent requests are waiting in the engine and this pending request in the context is removed & its blocks are deregistered. each step in the engine, we poll to see if a request has any block hashes in a pending state, and if it does, then we don't add it to the context. if the pending request in the context does get removed for whatever reason from the context, then the dependent requests in the engine will see that none of their blocks are in a pending state, and then it's okay to add those previously-dependent requests to the context. so this all should work cleanly together.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1596,6 +1701,27 @@ def add_request(self, req: DynamicInferenceRequest, chunk_length: Optional[int]
...
         ] = (token_offset_range % self.block_size_tokens)
 
+        # Register hashes for completely filled blocks (skip matched blocks)
+        if self.enable_prefix_caching and req.precomputed_block_hashes is not None:
+            total_tokens_after = req.finished_chunk_token_count + chunk_length
+            num_complete_blocks = total_tokens_after // self.block_size_tokens
+            previously_complete = req.finished_chunk_token_count // self.block_size_tokens
+
+            # Start from the first non-matched block
+            first_block_to_hash = max(previously_complete, num_matched_blocks)
```
kvareddy: I guess we can just use the range "new_block_start : new_block_start + len(new_block_ids)"

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1092,35 +1124,141 @@ def _check_stop_words_for_request_post_append(self, request: DynamicInferenceReq
...
+            return False
+
+        block_allocator = self.context.block_allocator
+
+        # Batch GPU hash table lookup for all block hashes
+        block_ids = block_allocator.gpu_hash_table.lookup_batch_alloc(req.precomputed_block_hashes)
+
+        # GPU-only prefix mask: cumprod of hit mask gives consecutive-match mask
+        hit_mask = (block_ids != -1).to(torch.int32)
+        prefix_mask = hit_mask.cumprod(dim=0).bool()
```
kvareddy: Not sure why we need a prefix mask here? Let's discuss this.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1092,35 +1124,141 @@ def _check_stop_words_for_request_post_append(self, request: DynamicInferenceReq
...
+        return self.context.gpu_match_prefixes_batch(request_hashes_list)
+
     def schedule_non_chunked_prefill(self):
         """
         Perform the same original scheduling logic for non-chunked runs
         """
+        # Pre-compute batch prefix matching on GPU (one kernel for all waiting requests)
+        batch_prefix_result = self._gpu_precompute_prefix_matches()
+
+        # Pre-compute schedulable limit: how many consecutive non-pending requests from front
```
kvareddy: Why not schedule all non-pending requests?

megatron/core/inference/inference_request.py
```diff
@@ -294,11 +451,49 @@ class DynamicInferenceRequest(InferenceRequest):
     finished_chunk_token_count: int = 0
     stop_word_ids: Optional[List[List[int]]] = None  # Tokenized stop words (populated internally)
 
+    # Prefix caching fields
+    block_size_tokens: Optional[int] = None  # Block size for hash computation
+    enable_prefix_caching: bool = False  # Whether prefix caching is enabled
```
santhnm2, line 357: Is it allowed to pass `enable_prefix_caching` without `block_size_tokens`? If not can we add an assert to verify this? Or consolidate these so that only the `block_size_tokens` needs to be passed and that implies prefix caching is enabled?

megatron/core/inference/inference_request.py
```diff
@@ -56,6 +57,162 @@ class Status(Enum):
...
+    import triton.language as tl
+
+    HAVE_TRITON = True
+except ImportError:
+    HAVE_TRITON = False
+
+if HAVE_TRITON:
+
+    @triton.jit
+    def _parent_chain_hash_kernel(
```
santhnm2: Can you add a link to some documentation for the algorithm here? Also is there some known reference implementation we can validate the Triton kernel against? Would be nice to add some unit tests for this in isolation (e.g., validate computed hash is correct, validate edge cases are handled correctly without underflowing)

megatron/core/inference/contexts/gpu_hash_table.py
```diff
@@ -0,0 +1,367 @@
...
+    # Cast EMPTY to int64 to match TABLE_KEYS pointer type for atomic_cas
+    empty_i64 = tl.full([], EMPTY, dtype=tl.int64)
+
+    # Skip empty sentinel keys
+    if key != empty_i64:
+        slot = (key & CAPACITY_MASK).to(tl.int64)  # Fast modulo for power-of-2
+        done: tl.int32 = 0
+
+        # Linear probing with bounded search
+        for _ in range(256):
```
santhnm2: Why 256 here? Can we parametrize this somehow (either as a configurable argument or a global constant)? Also can insertion fail if collisions cannot be avoided? This is extremely unlikely of course but wondering if there should be some safety mechanism in place

megatron/core/inference/contexts/gpu_hash_table.py
```diff
@@ -0,0 +1,367 @@
...
+    return 1 << (n - 1).bit_length()
+
+
+# =========================================================================
+# Triton kernels
+# =========================================================================
+
+
+@triton.jit
+def _hash_table_insert_kernel(
```
santhnm2: Can we add unit tests for the hash table implementation in isolation? This should behave identically to a native Python dictionary given the same sequence of insertions and lookups right? Can we validate that explicitly by generating random sequences and comparing output?

megatron/core/inference/contexts/gpu_hash_table.py
```diff
@@ -0,0 +1,367 @@
...
+    TABLE_VALUES,
+    INSERT_KEYS,
+    INSERT_VALUES,
+    CAPACITY_MASK: tl.constexpr,
+    EMPTY: tl.constexpr,
+):
+    """Insert N key-value pairs with linear probing.
+
+    Each program handles one key-value pair. Collisions are resolved via
+    linear probing with atomic CAS on keys.
```
santhnm2: Where is the atomic CAS happening? Do we know that with a Triton grid size > 1 that there won't be race conditions here across parallel requests?

megatron/core/inference/contexts/gpu_hash_table.py
```diff
@@ -0,0 +1,367 @@
...
+    Keys: int64 (block hashes), Values: int32 (block IDs).
+    Empty sentinel: key = -1.
+    Capacity: next_power_of_2(2 * max_entries) for ~50% load factor.
+
+    Deletion strategy: Rather than implementing complex tombstone-free
+    backward-shift deletion in Triton, we use a "rebuild after eviction"
+    approach. Inserts happen individually via the insert kernel, but
+    deletions trigger a full rebuild from the source-of-truth tensors
+    (block_hashes in BlockAllocator). This is acceptable because eviction
+    is infrequent and batched.
```
santhnm2: > This is acceptable because eviction is infrequent

Is this true in practice? If requests are finishing often then wouldn't we have frequent deletion events?

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +242,167 @@ def reset(self) -> None:
...
+    def get_evictable_block_count(self) -> Tensor:
+        """Get count of cached blocks that can be evicted (ref_count == 0, hash set).
+
+        Returns:
+            Scalar tensor with the number of evictable cached blocks.
+        """
+        cached_mask = (self.block_ref_counts == 0) & (self.block_hashes != -1)
+        return cached_mask.sum()
+
+    def evict_lru_blocks(self, num_blocks_needed: int) -> bool:
```
santhnm2, line 296: Maybe rename to `try_evict_lru_blocks` since this could be a no-op?

megatron/core/inference/contexts/dynamic_block_allocator.py
```diff
@@ -133,3 +242,167 @@ def reset(self) -> None:
...
+        if num_cached < num_blocks_needed:
+            return False  # Not enough cached blocks to evict
+
+        # Use topk to find oldest timestamps. Mask non-cached blocks with MAX_TS
+        # so they sort to the end. topk output size is known at call time (no sync).
+        MAX_TS = torch.iinfo(torch.int64).max
+        masked_timestamps = torch.where(cached_mask, self.block_timestamps, MAX_TS)
+        _, evict_indices = torch.topk(
+            masked_timestamps, num_blocks_needed, largest=False, sorted=False
+        )
```
santhnm2: Can we add a safety check (or a unit test) that ensures active blocks won't be evicted here?

megatron/core/inference/engines/dynamic_engine.py
lmcafee-nvidia: the hash table can return scattered matches (e.g., blocks 0,1,3 hit but block 2 misses). we can only reuse a contiguous prefix starting from block 0, a gap means the KV cache chain is broken. `cumprod` on the hit mask zeros out everything after the first miss, giving exactly the valid contiguous prefix. without it, we'd incorrectly count block 3 as matched despite the gap at block 2

megatron/core/inference/inference_request.py
lmcafee-nvidia: the condition below already guards against this (line ~468):

```
        if (
            self.enable_prefix_caching
            and self.block_size_tokens is not None
            and self.prompt_tokens is not None
        ):
```

lmcafee-nvidia: added an explicit assert (line 472) that fails early with a clear message. the guard condition below remains as a safety net.

megatron/core/inference/inference_request.py
lmcafee-nvidia: added Rabin fingerprint reference (Karp & Rabin, 1987) to the kernel docstring. added `test_gpu_cpu_hash_cross_validation` that validates GPU Triton kernel output matches CPU implementation for 1, 3, 16, and 64 blocks. the existing `TestHashContract` class already covers determinism, range, parent chain integrity, and edge cases.

megatron/core/inference/contexts/gpu_hash_table.py
lmcafee-nvidia: extracted `MAX_PROBE_DISTANCE = tl.constexpr(256)` as a module-level constant, used by all three Triton kernels. at 50% load factor, expected probe length is ~2; 256 is far beyond any realistic chain. the `insert_batch` retry loop (up to 8 iterations) handles the safety mechanism, it now verifies both key presence and value correctness after each kernel launch, re-inserting any entries lost to thread races.

megatron/core/inference/contexts/gpu_hash_table.py
lmcafee-nvidia: added `TestGPUHashTableDictEquivalence` class with `test_random_insert_lookup_matches_dict`: 500 unique random keys in `[1, 10^18)` range, validates all lookups match a Python reference `dict`, and confirms missing keys return `-1`. the GPU hash table test file now has 25 tests across 8 classes.

megatron/core/inference/contexts/gpu_hash_table.py
lmcafee-nvidia: good catch, the docstring was stale. there is no atomic CAS; the kernel uses non-atomic stores with linear probing. races between threads can cause two issues: (1) a key is lost when two threads write to the same slot, and (2) a key-value "tear" where one thread's key gets another's value. the `insert_batch` retry loop verifies both key presence and value correctness after each kernel launch, re-inserting any affected entries. updated the docstring to accurately describe this design.

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: done, renamed to `try_evict_lru_blocks`.

megatron/core/inference/contexts/dynamic_block_allocator.py
lmcafee-nvidia: added assertion at line 406: `assert (self.block_ref_counts[evict_indices] == 0).all()`, verifies no active blocks (`ref_count > 0`) are in the eviction set. the existing `test_lru_protects_active_blocks` also validates this at the integration level.

megatron/core/inference/contexts/gpu_hash_table.py
lmcafee-nvidia: clarified the docstring. rebuild cost is `O(num_blocks)`, sub-millisecond for up to ~100K blocks on H100, approaching ~1ms near 1M blocks. in LRU mode, rebuilds only occur under memory pressure (free pool exhausted). in REF_ZERO mode (default), rebuilds occur on every release but remain cheap at typical block counts (1K-100K).

megatron/core/inference/engines/dynamic_engine.py
kvareddy: My point is that hit_mask is always a prefix mask. Why would there be a case with scattered matches?

kvareddy (approved):

asolergi-nv (approved):

asolergi-nv:
@greptile Please review

examples/inference/gpt/gpt_dynamic_inference_12b.sh
```diff
@@ -69,7 +69,7 @@ ARGS=" \
     --tiktoken-pattern v2 \
     --tokenizer-model ${TOKENIZER_MODEL} \
     --distributed-timeout-minutes 2400 \
-    --transformer-impl local \
```
shanmugamr1992, line 72: is this needed?

lmcafee-nvidia: remove this arg

PR #2278, opened 2025-11-17
Automatically choose available ports in ZMQ

tdene:
# What does this PR do ?
Automatically choose available ports in ZMQ.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
        C1[Expert Review] --> C2[Final Review]
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [ ] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

The following process is enforced via the CODEOWNERS file for changes into `megatron/core`. For changes outside of `megatron/core`, it is up to the PR author whether or not to tag the Final Reviewer team.

<details>
<summary>For MRs into `main` branch</summary>

#### (Step 2): Collect the expert reviewers reviews

1. Attach the `Expert Review` label when your PR is ready for review.
2. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

:warning: Only proceed to the next step once all reviewers have approved, merge-conflict are resolved and the CI is passing.  
Final Review might get declined if these requirements are not fulfilled.

#### (Step 3): Final Review

1. Add `Final Review` label
2. GitHub auto-assigns final reviewers based on your changes. They will get notified and pick up your PR soon.

#### (Optional Step 4): Cherry-pick into release branch

If this PR also needs to be merged into `core_r*` release branches, after this PR has been merged, select `Cherry-pick` to open a new PR into the release branch.

</details>

<details>
<summary>For MRs into `dev` branch</summary>
The proposed review process for `dev` branch is under active discussion.

MRs are mergable after one approval by either `eharper@nvidia.com` or `zijiey@nvidia.com`.
</details>

### Merging your PR

Any member of [core-adlr](https://github.com/orgs/teams/NVIDIA/core-adlr) and [`core-nemo`](https://github.com/orgs/teams/NVIDIA/core-nemo) will be able to merge your PR.

PR #3107, opened 2026-01-27
Move to using the Inference OpenAI API server

ArEsKay3:
# What does this PR do ?
This changes the RL loop to use the new inference server code that implements an OpenAI-esque API.
As a result we no longer need to worry about chat templating in the RL code (so all of that is removed).
Also, since the "non-chat" mode is a special case of the "chat" mode switch all RL code to use messages and add a fallback template that just concatenates requests.
In order to support output parsing it adds a new Parser class that takes a string and returns the remaining string and the processed metadata. Each inference library has their own versions of these, so we need our own. In the future hopefully we can rely on HF tokenizers like we do for chat templating.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


ArEsKay3:
Two things I still need to clean up

- Countdown and Math environments want to control their prompt based on what stage of alignment the model is at (chat_mode). I think we should throw this away?
- We are not properly using the "n" argument to generation backend anymore. I want to move to use it where possible.

examples/rl/model_configs/common.sh
```diff
@@ -22,7 +22,7 @@ COMMON_OPTIONS="\
     --attention-backend flash \
     --timing-log-level 1 \
     --log-timers-to-tensorboard \
-    --save-retain-interval 120 \
+    --save-retain-interval 160 \
```
yobibyte, line 25: This might break some of the justfile recipes we have.

megatron/training/arguments.py
yobibyte, line 1: Will require adjusting of just recipes.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
```diff
@@ -78,13 +76,15 @@ async def chat_completions():
                 return_log_probs=sampling_params.return_log_probs,
                 top_n_logprobs=sampling_params.top_n_logprobs,
                 num_tokens_to_generate=sampling_params.num_tokens_to_generate,
+                skip_prompt_log_probs=True,
```
yobibyte: Was this a bug before?

megatron/rl/agent/api.py
yobibyte, line 1: Nice!

megatron/rl/__init__.py
```diff
@@ -65,6 +65,7 @@ class GenericGenerationArgs(BaseModel):
     top_k: int | None = None
     top_p: float | None = None
     max_tokens: int | None = None
+    n: int | None = None
```
yobibyte: Could you add a comment explaining what this means?

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
ArEsKay3: Yes, I believe so. OpenAI chat completions only returns for output tokens. We may want to change this behavior in the future, but not returning logprobs for the prompt is in line with what people would expect.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -740,6 +741,7 @@ def _add_request(
         if request.status != Status.FAILED:
             self.waiting_request_ids.append(request_id)
         else:
+            logger.error(f"Request {request_id} failed to add to engine. {request.events}")
```
tdene: This is already merged into main. Would you mind undoing your changes to `dynamic_engine.py` to minimize the diff?

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
```diff
@@ -31,14 +32,14 @@ async def chat_completions():
...
         except AttributeError:
-            return (
-                "Tokenizer does not support 'apply_chat_template'. "
-                "Chat completions requires a tokenizer with a configured chat template."
-            ), 500
+            logger.warning("Tokenizer does not support 'apply_chat_template'. Using tokenize instead.")
+            prompt_tokens = tokenizer.tokenize("\n".join([message["content"] for message in messages]))
         except Exception as e:
+            import traceback
+            logger.error(f"{traceback.format_exc()}")
```
tdene: Please wrap this method in the `@trace_async_exceptions` decorator - which does exactly this - instead of adding manual traceback code.

megatron/rl/agent/api.py
tdene: +1

megatron/rl/inference/megatron.py
```diff
@@ -196,51 +195,45 @@ def get_dynamic_inference_engine(
...
 
-        if any(isinstance(p, LLMChatMessage) for p in request.prompt):
-            raise ValueError(
-                "MegatronLocal does not support chat requests."
-                "Use MegatronChatLocal to apply chat templating."
-            )
-        assert all(
-            isinstance(p, str) for p in request.prompt
-        ), "MegatronLocal only supports string prompts."
+        assert self._server_task is not None, "Infernce server is not initialized"
```
tdene: `Infernce` typo

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
```diff
@@ -78,13 +79,15 @@ async def chat_completions():
                 return_log_probs=sampling_params.return_log_probs,
                 top_n_logprobs=sampling_params.top_n_logprobs,
                 num_tokens_to_generate=sampling_params.num_tokens_to_generate,
+                skip_prompt_log_probs=True,
```
tdene: We should remove the rebuild of the sampling params. It only adds confusion.

Sampling params already go through a deepcopy when the request is initialized, before anything is allowed to modify them. There is no need to have `per_req_params`.

tdene: Please make `skip_prompt_log_probs` use `req.get` like `return_log_probs` does:

```
skip_prompt_log_probs = bool(req.get("skip_prompt_log_probs", True))
```

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
```diff
@@ -62,7 +63,7 @@ async def chat_completions():
                 top_p=top_p,
                 return_log_probs=return_log_probs,
                 top_n_logprobs=top_n_logprobs,
-                num_tokens_to_generate=int(req.get("max_tokens", 16)),
+                num_tokens_to_generate=int(max_tokens) if ( (max_tokens := req.get("max_tokens", None)) is not None ) else None,
             )
         except ValueError as e:
             return f"Invalid sampling parameter: {e}", 400
```
tdene, line 79: Let's add `add_BOS` and `termination_id` in here as well. It's not directly germane, but it should be done at some point.

megatron/rl/rl_utils.py
```diff
@@ -1687,6 +1657,12 @@ def rl_inference_interface_shutdown():
     else:
         logger.warning("No inference interface to shutdown. This should not happen.")
 
+    # TODO(rkirby): This is a hack to hard exit. There is a bug that is preventing us from using sys.exit(0).
+    # It seem the Flask server has non-daemon threads that are preventing the program from exiting.
+    # We need to find a way to gracefully complete all in progress requests and shutdown the Flask server.
+    import os
+    os._exit(0)
```
tdene, line 1801: General note: @tdene will take on this TODO, and related shutdown issues.

We should push this PR with `os._exit` and leave the fix for a follow-up.

megatron/rl/inference/megatron.py
```diff
@@ -282,14 +275,25 @@ async def launch(cls, model: GPTModel, **kwargs):
...
         if dist.get_rank() == 0:
-            # TODO: We have to do this only on the rank 0 process, should be fixed in the future when we have support for multiple inference clients. !2278
+            from megatron.core.inference.text_generation_server.dynamic_text_gen_server.flask_server import run_flask_server_on_client
+            loop = asyncio.get_event_loop()
             client = InferenceClient(inference_coordinator_port=41521)
             await client.start()
+            server_task = loop.create_task(run_flask_server_on_client(
+                client=client,
+                tokenizer=inference_engine.controller.tokenizer,
+                flask_port=8294,
```
tdene: This is an oversight; it should use `kwargs["port"]`.

megatron/rl/inference/megatron.py
```diff
@@ -196,51 +195,45 @@ def get_dynamic_inference_engine(
...
+        assert all(len(response.choices) == 1 for response in responses), "Still need to properly support requests with n > 1"
+
         return [
             InferenceResponse(
-                response=r.generated_text,
-                raw_text=p + r.generated_text,
-                token_ids=r.prompt_tokens.tolist() + r.generated_tokens,
-                logprobs=r.generated_log_probs,
-                prompt_length=len(r.prompt_tokens),
+                response=LLMChatMessage(**choice.message.model_dump(include={'role', 'content'})),
```
tdene: Could you include a `TODO` reminder to handle `message["tool_calls"]` / `message["reasoning"]` in the future, so that it's easy to remember when we get to that point?

megatron/rl/inference/megatron.py
```diff
@@ -196,51 +195,45 @@ def get_dynamic_inference_engine(
...
-        ]
-        records = await asyncio.gather(*requests)
-        responses = [record[-1] for record in records]
+            n=request.generation_args.n or 1,
+            logprobs=True,
+        ) for prompt in request.prompt ]
+
+        responses = await asyncio.gather(*generations)
+
+        assert all(len(response.choices) == 1 for response in responses), "Still need to properly support requests with n > 1"
```
tdene: What's still needed to support these? I see a `for response in responses for choice in response.choices` loop below.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/flask_server.py
```diff
@@ -33,7 +33,7 @@ def temp_log_level(level, logger=None):
 
 
 @trace_async_exceptions
-async def run_flask_server(coordinator_port: int, tokenizer, rank: int, flask_port: int):
+async def run_flask_server_on_client(client: InferenceClient, tokenizer, flask_port: int, parsers: list[str] = []):
```
tdene: `parsers: list[str] = []` is the dangerous Python pattern where you pass in a mutable default. See [this old blog post](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil).

It should not cause issues here. But the pattern should still be avoided on principle?

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
```diff
@@ -133,12 +136,31 @@ async def chat_completions():
...
-                    "message": {"role": "assistant", "content": text_output},
+                    "message": message,
+                    "prompt_token_ids": prompt_tokens,
+                    "generation_token_ids": result.generated_tokens,
+                    "generation_log_probs": result.generated_log_probs,
+                    "raw_text": result.prompt + result.generated_text,
                     # 'logprobs' in chat API is an object containing 'content'
                     "logprobs": {"content": logprobs_content} if logprobs_content else None,
-                    "finish_reason": "length",  # Original code hardcoded this.
+                    "finish_reason": "tool_calls" if metadata.get("tool_calls", []) else "stop",  # Original code hardcoded this.
```
tdene: Please add a `TODO` to fix this so that it distinguishes between `length` and `stop`.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
ArEsKay3: I don't think this is the same right? The point here is to log it and return.

megatron/core/tokenizers/text/parsers/base_parser.py
```diff
@@ -0,0 +1,15 @@
+class BaseParser:
+    @staticmethod
+    def parse(text: str, **kwargs) -> tuple[str, dict[str, str]]:
```
tdene: The type-hint here is off. Qwen3 parser, for example, can return `tuple[str, dict[str, list[dict]]]` if I'm reading the tool calls option correctly.

megatron/core/tokenizers/text/parsers/qwen3_coder_tool_parser.py
```diff
@@ -0,0 +1,295 @@
...
+        self.tool_call_regex = re.compile(
+            r"<tool_call>(.*?)</tool_call>|<tool_call>(.*?)$", re.DOTALL
+        )
+        self.tool_call_function_regex = re.compile(
+            r"<function=(.*?)</function>|<function=(.*)$", re.DOTALL
+        )
+        self.tool_call_parameter_regex = re.compile(
+            r"<parameter=(.*?)(?:</parameter>|(?=<parameter=)|(?=</function>)|$)",
+            re.DOTALL,
+        )
```
tdene: These get recompiled on every parse, because `_Qwen3CoderToolParser` gets re-initialized on every parse. Initialize this class only once in `Qwen3CoderToolParser`, or move the regex compilation to some other static place, so that it only runs once.

megatron/core/tokenizers/text/parsers/deepseek_r1_reasoning_parser.py
```diff
@@ -0,0 +1,26 @@
...
+
+        Args:
+            text (str): The text to parse.
+
+        Returns:
+            tuple[str, dict[str, str]]: A tuple containing the unprocessed text
+            and a dictionary with the extracted reasoning content.
+        """
+
+        if "</think>" in text:
```
tdene, line 24: Please add a `NOTE` that this can only parse the first set of `think` tokens, and any subsequent `think` blocks are treated as non-think.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/flask_server.py
```diff
@@ -68,9 +65,19 @@ def health_check():
     # Force logging level to INFO to ensure that hostname is printed
     with temp_log_level(logging.INFO, logger):
         logger.info(f"Starting Flask server on http://{hostname}:{flask_port}")
+        logger.info(f"Using tokenizer: {type(tokenizer)}")
+        logger.info(f"Using parsers: {parsers}")
+
+    await serve(app, config)
```
tdene, line 73: While it's not as important as for the ZMQ ports, we should still guard against the port being already in-use.

Annoyingly, there is no support for this in the ASGI standard ([see here](https://github.com/fastapi/fastapi/discussions/11331)).

So this suggestion would either involve something like
```
with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as sock:
    try:
        sock.bind((", flask_port))
    except OSError:
        sock.bind(('', 0))
        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        avail_port = str(s.getsockname()[1])

logger.info(f"Useful log message based on {avail_port == flask_port}")
config.bind = [f"0.0.0.0:{avail_port}"]
```

Or a more correct, but more painful:
```
    try:
        sock.bind((", flask_port))
    except OSError:
        sock.bind(('', 0))
        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        avail_port = str(s.getsockname()[1])

sock.set_inheritable(True)

logger.info(f"Useful log message based on {avail_port == flask_port}")
config.bind = [f"fd://{sock.fileno()}"]
```

megatron/rl/rl_utils.py
```diff
@@ -384,36 +383,12 @@ def get_agent(args, parallel_generation_tasks: int | None = None):
...
-                )
-            else:
-                raise ValueError(
-                    f"Unknown inference_server_type {args.langrl_inference_server_type}"
-                )
+        _INFERENCE_INTERFACE = loop.run_until_complete(
+            MegatronLocal.launch(
+                model[0], 
+                host='0.0.0.0', 
+                port=os.getenv('MEGATRON_RL_INFERENCE_SERVER_PORT', 8294))
```
tdene: Do we need to specify a particular port? There's a `logger.info(f"Starting Flask server on http://{hostname}:{flask_port}")` that can be scraped to obtain the valid port. Scraping the logs for the correct port is also the pattern that the non-flask coordinator uses as of #2278 

See the other related comment in `flask_server.py` for why we might need to scrape the logs anyway, regardless of whether or not we request a port.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
ArEsKay3: I'm not totally against this, but I want to point out logprobs is in the OpenAI API.
skip_prompt_logprobs is not.
We would be adding additional supported arguments on top of the API. We already do this for output (add addtnl info) so it's probably not the worst.

megatron/rl/agent/reward_only_agent.py
```diff
@@ -144,19 +138,9 @@ async def group_rollout(self, request: GroupedRolloutRequest) -> list[Rollout]:
...
-        ), "get_grouped_rollouts only requested a single group but got multiple groups"
-        responses = groups[0].responses
-
-        rollouts = await asyncio.gather(
-            *[self.rollout_from_response(request, response, golden) for response in responses]
-        )
 
-        return rollouts
+        responses = await asyncio.gather(*[request.inference_interface.agenerate(inference_request) for _ in range(request.rollouts_per_group)])
+        return [await self.rollout_from_response(request, response[0], golden) for response in responses]
```
tdene: I'm confused and don't fully understand. It seems like we're moving away from `inference_request.n` in here?

megatron/core/tokenizers/text/parsers/qwen3_coder_tool_parser.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+    def _generate_tool_call_id(self) -> str:
+        """Generate a unique tool call ID."""
+        return f"call_{uuid.uuid4().hex[:24]}"
+
+    def _get_arguments_config(
+        self, func_name: str, tools: list[ChatCompletionToolsParam] | None
```
tdene, line 45: Since `ChatCompletionToolsParam = dict[str, Any]`, none of the `hasattr` or `config.` dot-accessing will work, right?

Test script:
```
from typing import Any
MyParams = dict[str, Any]
def f(pars: list[MyParams]):
    [print(p.type) for p in pars]

f([{"type": "function"}])
```

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
tdene: Ah, you're right, my bad.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
tdene: @santhnm2 what are your thoughts on adding additional supported arguments like this?

megatron/core/tokenizers/text/parsers/qwen3_coder_tool_parser.py
ArEsKay3: Correct! Good catch. This was actually working because it was just returning everything as a string (which isn't a big deal because it gets json.dump'd anyway), but no actual conversion/validation was happening.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/flask_server.py
ArEsKay3: Gonna suggest we move this to another PR

megatron/core/tokenizers/text/parsers/base_parser.py
ArEsKay3: Correct.

megatron/rl/agent/reward_only_agent.py
ArEsKay3: This is one thing I still am cleaning up on the RL side. I'm not sure where we'll land. Stay tuned.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/flask_server.py
tdene: Decided to resolve in future PR

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -91,6 +90,9 @@
     HAVE_PSUTIL = False
 
 
+logger = logging.getLogger(__name__)
+
+
```
tdene: Please remove this diff as well.

megatron/rl/inference/megatron.py
```diff
@@ -282,14 +271,25 @@ async def launch(cls, model: GPTModel, **kwargs):
...
         if dist.get_rank() == 0:
-            # TODO: We have to do this only on the rank 0 process, should be fixed in the future when we have support for multiple inference clients. !2278
+            from megatron.core.inference.text_generation_server.dynamic_text_gen_server.flask_server import run_flask_server_on_client
+            loop = asyncio.get_event_loop()
             client = InferenceClient(inference_coordinator_port=41521)
             await client.start()
+            server_task = loop.create_task(run_flask_server_on_client(
+                client=client,
+                tokenizer=inference_engine.controller.tokenizer,
+                flask_port=kwargs.get('port', 8294),
```
tdene, line 95: Do you even need the default at all? Won't it always be provided?

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
santhnm2: In this particular case the `skip_prompt_log_probs` flag is just for performance optimization purposes. But perhaps more generally these kinds of flags should be read directly by the engine rather than as a parameter in the endpoint API.

ArEsKay3:
/ok to test [cb53a84](https://github.com/NVIDIA/Megatron-LM/pull/3107/commits/cb53a8450b01dce1693a0c8e5af9a713c0366ee3)

ArEsKay3:
/ok to test [79b7ef0](https://github.com/NVIDIA/Megatron-LM/pull/3107/commits/79b7ef0889c80f45d0ad9b2d8d58890c6e9d08ad)

ArEsKay3:
/ok to test [5182594](https://github.com/NVIDIA/Megatron-LM/pull/3107/commits/5182594d287bbf05caf7478fb984c665a3452b4a)

examples/rl/model_configs/common.sh
ArEsKay3: Yes I will follow with a clean up of mrl_internal next week.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
ArEsKay3: I'd rather not have add_BOS over the API, really we should do it on a tokenizer basis.
I think the stop condition should be stop string. Is that supported?

tdene: It is; `stop_words: Optional[List[str]]` inside the sampling params.

I think the stop conditions should be both stop string and termination ID. Stop string is less reliable and uses more compute.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
tdene: Regardless though, hardcoding `skip_prompt_log_probs` to `True` in the flask server with this PR will trip up other uses of the flask server who may be expecting prompt log probs.

We can fix things in a follow-up PR, but we should avoid having `True` as the default once this gets merged.

megatron/core/tokenizers/text/parsers/deepseek_r1_reasoning_parser.py
```diff
@@ -0,0 +1,33 @@
...
+from megatron.core.tokenizers.text.parsers.base_parser import BaseParser
+
+
+class DeepSeekR1ReasoningParser(BaseParser):
+    """Parser for DeepSeek R1 style reasoning output."""
+
+    @staticmethod
+    def parse(text: str, **kwargs) -> tuple[str, dict[str, str]]:
+        """
+        Extracts the reasoning content from the text using <think>...</think> tags.
```
yobibyte, line 11: Let's add a couple of unit tests checking that this behaviour is as written. e.g. add the second part of think tags. Have one opening no closing, one closing no opening.

This is simple so far, but parsers tend to grow and accumulate bugs.

yobibyte: This will actually break if your closing think is before opening think, please add the test for this too. 

```parse('</think>42<think>')```

File: megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py (hunk 8 above)
Comment (ArEsKay3) [reply megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py:79]: Over the OpenAI API you don't know the ID.

**Review by santhnm2 — APPROVED** [review]
Text gen server changes LGTM

**Review by kvareddy — APPROVED** [review]

**Comment by ArEsKay3** [comment]
/ok to test [338f36d](https://github.com/NVIDIA/Megatron-LM/pull/3107/commits/338f36d6167f3b6118734a9cf2cc2c06107c6f28)

File: megatron/core/tokenizers/text/parsers/deepseek_r1_reasoning_parser.py (hunk 24 above)
Comment (yobibyte) [reply megatron/core/tokenizers/text/parsers/deepseek_r1_reasoning_parser.py:11]: @ArEsKay3 pinging in case you missed this

**Comment by ArEsKay3** [comment]
/ok to test [0bd3202](https://github.com/NVIDIA/Megatron-LM/pull/3107/commits/0bd32029b791c345793d79b6070b5958c41e8638)

**Comment by ko3n1g** [comment]
/ok to test [3912fd7](https://github.com/NVIDIA/Megatron-LM/pull/3107/commits/3912fd7d5260ff99018b5bfb20797881f4ff67bb)

**Comment by chtruong814** [comment]
@ArEsKay3 This PR failed the merge queue.  The test was hanging on this failure.  We are now running the larger functional test suite in the queue.  By default on a PR, it will only run unit tests + some handful of functional tests.

Please look at this failure and you can re-run all functional tests by adding the "Run functional tests" label and running CI tests with "/ok to test <commit>" comment.

https://github.com/NVIDIA/Megatron-LM/actions/runs/21863513917/job/63100758345

```
9-task-1-0/0 [default0]:INFO:root:Inference Coordinator: waiting for connections from data parallel ranks...
9-task-1-0/0 [default0]:/usr/local/lib/python3.12/dist-packages/torch/distributed/distributed_c10d.py:4879: UserWarning: barrier(): using the device under current context. You can specify `device_id` in `init_process_group` to mute this warning.
9-task-1-0/0 [default0]:  warnings.warn(  # warn only once
9-task-1-0/0 [default0]:INFO:root:Inference Coordinator: Connected with data parallel ranks...
9-task-1-0/0 [default0]:INFO:root:Inference co-ordinator is ready to receive requests!
9-task-1-0/0 [default0]:INFO:root:Data parallel coordinator can be found at tcp://172.18.0.2:41521
9-task-1-0/0 [default0]:INFO:root:Client: Connecting to InferenceCoordinator...
9-task-1-0/0 [default0]:ERROR:megatron.core.utils:Exception in async function run_flask_server_on_client: Flask not available
9-task-1-0/0 [default0]:Traceback (most recent call last):
9-task-1-0/0 [default0]:  File "/opt/megatron-lm/megatron/core/utils.py", line 2364, in wrapper
9-task-1-0/0 [default0]:    return await fn(*args, **kwargs)
9-task-1-0/0 [default0]:           ^^^^^^^^^^^^^^^^^^^^^^^^^
9-task-1-0/0 [default0]:  File "/opt/megatron-lm/megatron/core/inference/text_generation_server/dynamic_text_gen_server/flask_server.py", line 41, in run_flask_server_on_client
9-task-1-0/0 [default0]:    raise RuntimeError(f"Flask not available")
9-task-1-0/0 [default0]:RuntimeError: Flask not available
```

ko3n1g:

You can track the progress here: https://github.com/NVIDIA/Megatron-LM/actions/runs/21920080099

ko3n1g:

You can track the progress here: https://github.com/NVIDIA/Megatron-LM/actions/runs/21932262559

ko3n1g:

You can track the progress here: https://github.com/NVIDIA/Megatron-LM/actions/runs/21960186453

ko3n1g:

You can track the progress here: https://github.com/NVIDIA/Megatron-LM/actions/runs/21962433107

ko3n1g:

You can track the progress here: https://github.com/NVIDIA/Megatron-LM/actions/runs/21965383906

PR #3116, opened 2026-01-28
[dev] feat(moe): Support Seq1F1B - an efficient extension of pipeline parallelism for long-context training

yuzhongw-nvidia:
# What does this PR do ?

## Background

Seq1F1B ([Sun et al., 2024](https://arxiv.org/abs/2406.03488)) is an efficient sequence-level one-forward-one-backward pipeline scheduling method designed for training LLMs on long sequences. The core idea is to split sequences along the sequence length dimension, decomposing batch-level schedulable units into finer sequence-level units, and then perform PP or interleaved PP. The main benefits of Seq1F1B are the reduction of PP bubble size and memory footprint, which are two of the most significant overheads of pipeline parallelism.

To implement Seq1F1B in Megatron-Core, we develop a feature called **Chunked Pipeline Parallel (Chunked PP)**, which provides the infrastructure for sequence-level pipeline scheduling. Furthermore, based on the Chunked PP infrastructure, we plan to implement SlimPipe ([Li et al., 2025](https://arxiv.org/abs/2504.14519)), a Seq1F1B variant in future works.

### Reference

- [Seq1F1B: Efficient Sequence-Level Pipeline Parallelism for Large Language Model Training](https://arxiv.org/abs/2406.03488) (Sun et al., 2024)
- [thunlp/Seq1F1B: Sequence-level 1F1B schedule for LLMs.](https://github.com/thunlp/Seq1F1B)

### Acknowledgements

This implementation is based on [thunlp/Seq1F1B](https://github.com/thunlp/Seq1F1B). We would like to thank [Ao Sun](https://github.com/MayDomine) from Beijing University of Posts and Telecommunications for his contribution.

---

### Comparison with Context Parallelism (CP)

Compared to Context Parallelism (CP), Seq1F1B (via Chunked PP) has advantages in the following aspects:
- **PP Bubble Optimization**: Effectively reduces PP bubble
- **Flexibility**: Does not occupy a dimension in 3D parallelism, making it very flexible. For example, with 32 GPUs, if you already have PP4EP8TP8, you can still set `chunked_pipeline_model_parallel_splits=4`, but it would not be possible to set CP=4
- **Linear attention**: Works well with linear attention because: (1) the workload imbalance issue does not exist, and (2) the KV cache becomes smaller.

Besides, Seq1F1B (via Chunked PP) is compatible with CP, and they can work together in specific scenarios to further improve performance or extend sequence length.

### Benchmarking

Achieves at most **1.15x** perf gain compared with CP with Qwen3-Next-80B-A3B model, training 32k context. However, we also observe significant PP bubbles due to the workload imbalance in some cases, making it underperform CP in these cases. Besides, chunked PP has a higher activation memory compared with CP, sometimes making it runs OOM.

Chunked PP does not support cuda graph for now, so it still cannot improve our best practice of Qwen3-Next-80B-A3B on Blackwell.

TODO: Will add more detailed benchmarking results comparing with CP.

---

### Command Line Arguments

Set in the launch script:
```bash
--chunked-pipeline-model-parallel-splits 4
```

### Code Configuration

Set in `TransformerConfig`:
```python
config = TransformerConfig(
    ...
    chunked_pipeline_model_parallel_splits=4,
)
```

---

### Configuration Parameters

A new configuration option is added to `TransformerConfig`:

```python
chunked_pipeline_model_parallel_splits: int = 1
"""Number of chunked pipeline model parallel splits."""
```

Chunked PP (Seq1F1B) is enabled when `chunked_pipeline_model_parallel_splits > 1`.

#### Pipeline Schedule Modifications (`megatron/core/pipeline_parallel/schedules.py`)

To support the Seq1F1B-style pipeline schedule, we have the following modifications to `forward_backward_pipelining_with_interleaving` and `forward_backward_pipelining_without_interleaving`:

1. Wrap data iterator with `ChunkedPipelineParallelDataIterator`.
   ```python
   if config.chunked_pipeline_model_parallel_splits > 1:
       data_iterator = ChunkedPipelineParallelDataIterator(data_iterator, config)
   ```
   - `ChunkedPipelineParallelDataIterator`: Data iterator adapter for micro-batch management (`megatron/core/chunked_pipeline_parallel_utils.py`).
        - Splits input tensors (tokens, labels, loss_mask, position_ids) along the sequence dimension
        - Each `__next__` call returns one slice, i.e., one chunked micro-batch
        - Provides `get_current_cached_prefix_params()` method to get current chunked PP parameters
        - Provides `mock_next()` method for intermediate PP stages that hold a null data iterator.

2. Use `ChunkedPipelineParallelQueue`:
   ```python
   if config.chunked_pipeline_model_parallel_splits > 1:
       config.variable_seq_lengths = True
       input_tensors = ChunkedPipelineParallelQueue(config.chunked_pipeline_model_parallel_splits)
       output_tensors = ChunkedPipelineParallelQueue(config.chunked_pipeline_model_parallel_splits)
   else:
       input_tensors = []
       output_tensors = []
   ```
   - `ChunkedPipelineParallelQueue`: Two-level queue data structure for tensor management (`megatron/core/chunked_pipeline_parallel_utils.py`).
        - Manages input/output tensors in chunked PP scenarios
        - Inner queues are grouped by `chunked_pp_splits`
        - Supports append and pop operations, pop removes from the queue head

3. Adjust micro-batch count calculation:
   - `total_num_microbatches = num_microbatches * num_model_chunks * chunked_pp_splits`
   - `num_warmup_microbatches` increased by `chunked_pp_splits - 1`

4. Minor changes:
   - When Chunked PP is enabled, `config.batch_p2p_comm = False`
   - When Chunked PP is enabled, `config.variable_seq_lengths = True`

#### Attention Modifications

The attention module needs to be modified to support saving KV cache for micro-batches, as well as forward and backward propagation with KV cache.

Here we take standard multi-head attention as an example to illustrate the modification. We will support chunked PP for more attention variants (e.g., MLA, GDN, etc.) in the future.

##### Chunked Attention Implementation (`megatron/core/transformer/attention_chunked.py`)

1. Class `ChunkedKVCacheForAttention`: Implements KV cache management
   -  Maintains `key_cache`, `value_cache` lists to store historical K/V
   -  Maintains `grad_key_cache`, `grad_value_cache` lists to store gradients
   -  Interface of forward with KV cache.
   ```python
    def forward_with_kv_cache(
        self,
        module: Callable,
        cached_prefix_params: CachedPrefixParams,
        query: torch.Tensor,
        key: torch.Tensor,
        value: torch.Tensor,
        attention_mask: Optional[torch.Tensor],
        attn_mask_type: AttnMaskType,
        attention_bias: Optional[torch.Tensor] = None,
        packed_seq_params: Optional[PackedSeqParams] = None,
    ):
        # ...
   ```

2. `AttentionFuncionWithChunkedPipelineParallel`: Custom `torch.autograd.Function`
    - Forward Logic:
        - Append current K/V to cache lists
        - Concatenate all historical K/V: `key_joined = torch.cat(key_cache, dim=seq_dim)`
        - **Switch attention mask type to bottom-right**
            - This is a key step to let the current query pay attention to the historical kv cache.
        - Execute core attention computation
        - Release concatenated K/V data to save memory

    - Backward Logic:
        - Recover concatenated K/V from cache
        - Compute gradients for query, key_joined, value_joined
        - Split K/V gradients by span and accumulate to historical gradients
        - Pop current span's cache and gradients

    - *Note: We save KV cache in chunks because we need to implement chunk-based context exchange for SlimPipe in the future.*

##### Attention Module (`megatron/core/transformer/attention.py`)

1. Enabling attention function with chunked KV cache management.

    ```python
    if self.config.chunked_pipeline_model_parallel_splits > 1:
        kv_cache = cached_prefix_params.kv_cache_pool.get_kv_cache(
            layer_idx=self.layer_number, kv_cache_cls=ChunkedKVCacheForAttention
        )
        core_attn_out = kv_cache.forward_with_kv_cache(
            module=self.core_attention,
            cached_prefix_params=cached_prefix_params,
            query=query, key=key, value=value,
            attn_mask_type=attn_mask_type,
            # ...
        )
    ```

    * `KVCachePool`: Manages KV cache instances for each micro-batch (`megatron/core/chunked_pipeline_parallel_utils.py`).
        ```python
        class KVCachePool:
            def __init__(self, config: TransformerConfig):
                self.kv_cache_pool = {}
            
            def get_kv_cache(self, layer_idx: int, kv_cache_cls: Type[KVCache]) -> KVCache:
                # Lazily instantiate and cache KV cache for each layer
        ```

#### RoPE Position Embedding Modifications (`megatron/core/models/common/embeddings/`)

1. New `get_pos_emb_on_this_chunked_pp_span` function (`rope_utils.py`)
    - Get position embedding on the current chunked PP span:

    ```python
    def get_pos_emb_over_cached_prefix(
        pos_emb: Tensor, seq_dim: int, cached_prefix_params: CachedPrefixParams
    ):
        """Get the position embedding over the cached prefix.
    
        Args:
            pos_emb (Tensor): Positional embedding tensor
            seq_dim (int): Sequence dimension
            cached_prefix_params (CachedPrefixParams): Cached prefix parameters
        """
    
        prefix_len = sum(cached_prefix_params.prefix_seqlens)
        this_chunk_len = cached_prefix_params.this_chunk_seqlen
        pos_emb = pos_emb.narrow(seq_dim, prefix_len, this_chunk_len)
        return pos_emb
    ```

2. Modified `RotaryEmbedding.forward()` (`rotary_pos_embedding.py`):
    - If chunked PP enabled, slices position embeddings to the current span

3. Modified `get_rotary_seq_len()`:
    - If chunked PP enabled, returns the sum of all span lengths

#### Chunked Pipeline Parallel Params

`CachedPrefixParams` is a dataclass to store the metainfo of the current chunk and prefixes of micro-batch (`megatron/core/chunked_pipeline_parallel_utils.py`).

```python
@dataclass
class CachedPrefixParams:
    prefix_seqlens: List[int]
    this_chunk_seqlen: int
    max_total_seqlen: Optional[int]
    kv_cache_pool: KVCachePool
    boundary_elements_for_mtp: Optional[torch.Tensor]
    is_terminal: bool
```

To pass this param to transformer and attention modules, we have the following modifications.

* `pretrain_gpt` Modifications (`pretrain_gpt.py`)
    * In `get_batch()` function:
        - For PP stages that don't hold data, call `data_iterator.mock_next(seq_length)` to simulate the next batch
        - Obtain `cached_prefix_params` via `data_iterator.get_current_cached_prefix_params()`
        - Return `cached_prefix_params` along with other batch data
    * In `forward_step()` function:
        - Receive `cached_prefix_params` from `get_batch()`
        - Pass `cached_prefix_params` to `model.forward(..., cached_prefix_params=cached_prefix_params)`

* `GPTModel` / `TransformerBlock` / `TransformerLayer` Modifications
    * In `GPTModel.forward()`, `TransformerBlock.forward()`, and `TransformerLayer.forward()`
        - Added `cached_prefix_params` parameter
        - Pass the parameter to self-attention modules

---

## Future Work

* **Variable-Length Sequence Chunks**: Seq1F1B introduces a variable-length sequence mechanism that adjusts the length of each sequence chunk to balance computational workload across different chunks, thereby reducing the performance impact of PP bubbles. We plan to support this feature in a future PR.

* **SlimPipe** ([Li et al., 2025](https://arxiv.org/abs/2504.14519)): SlimPipe is a variant of Seq1F1B that proposes context exchange and vocabulary parallelism to resolve the workload imbalance issue in another way. Based on the chunked PP infrastructure introduced in this PR, we plan to support SlimPipe for Megatron Core in future PRs.

* The current implementation has the following limitations, which are left as our future works:

    1. **Attention mask not supported**: `attention_mask` must be `None` for now.
    2. **Packed sequence not supported**: `packed_seq_params` must be `None` for now.
    3. **MLA (Multi-Latent Attention) not supported**: Incompatible with MLA for now.
    4. **Core attention offloading not supported**: `core_attn` cannot be in `offload_modules` for now.

---

### Unit Tests

1. **End-to-end tests** (`tests/unit_tests/pipeline_parallel/test_chunked_pipeline_model_parallel.py`):
   - Tests forward correctness under different TP/PP/VPP configurations
   - Validates result consistency by comparing against PP=1 baseline

2. **Attention module tests** (`tests/unit_tests/transformer/test_attention_chunked_pp.py`):
   - Tests output consistency between `chunked_pp_splits=1` and `chunked_pp_splits=4`
   - Supports testing with different TP sizes

3. **Utility class tests** (`tests/unit_tests/pipeline_parallel/test_chunked_pipeline_model_parallel.py`):
   - `TestChunkedPipelineParallelDataIterator`: Tests data splitting and iteration
   - `TestChunkedPipelineParallelQueue`: Tests queue operations

---

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/cached_prefix_utils.py
+++ b/megatron/core/cached_prefix_utils.py
@@ -20,6 +20,7 @@ class KVCachePool:
         self.config = config
         self.kv_cache_pool = {}
         self.block_prefixes = []
+        self.kv_cache_pool[tuple(self.block_prefixes)] = {}
 
     def block_range_push(self, suffix: str):
         """Push a block range onto the stack."""
--- a/megatron/core/pipeline_parallel/chunked_pipeline_parallel_utils.py
+++ b/megatron/core/pipeline_parallel/chunked_pipeline_parallel_utils.py
@@ -84,7 +84,14 @@ class ChunkedPipelineParallelDataIterator:
         return slice_data
 
     def _get_span(self, seq_length: int) -> List[int]:
-        spans = [seq_length // self.chunked_pp_splits] * self.chunked_pp_splits
+        if self.config.chunked_pipeline_model_parallel_split_lengths is not None:
+            assert sum(self.config.chunked_pipeline_model_parallel_split_lengths) == seq_length, (
+                f"Sum of {self.config.chunked_pipeline_model_parallel_split_lengths=} "
+                f"must be equal to {seq_length=}."
+            )
+            spans = self.config.chunked_pipeline_model_parallel_split_lengths
+        else:
+            spans = [seq_length // self.chunked_pp_splits] * self.chunked_pp_splits
         return spans
 
     def mock_next(self, seq_length: int):
--- a/megatron/core/pipeline_parallel/schedules.py
+++ b/megatron/core/pipeline_parallel/schedules.py
@@ -1118,7 +1118,6 @@ def forward_backward_pipelining_with_interleaving(
         input_tensors = [[] for _ in range(len(model))]
         output_tensors = [[] for _ in range(len(model))]
     else:
-        config.variable_seq_lengths = True
         input_tensors = [
             ChunkedPipelineParallelQueue(config.chunked_pipeline_model_parallel_splits)
             for _ in range(len(model))
@@ -2307,8 +2306,6 @@ def forward_backward_pipelining_without_interleaving(
 
     if not forward_only:
         if config.chunked_pipeline_model_parallel_splits > 1:
-            config.variable_seq_lengths = True
-            config.batch_p2p_comm = False
             input_tensors = ChunkedPipelineParallelQueue(
                 config.chunked_pipeline_model_parallel_splits
             )
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -90,6 +90,11 @@ class TransformerConfig(ModelParallelConfig):
     chunked_pipeline_model_parallel_splits: int = 1
     """Number of chunked pipeline model parallel splits."""
 
+    chunked_pipeline_model_parallel_split_lengths: Optional[List[int]] = None
+    """[Experimental] Custom chunk lengths for chunked pipeline model parallel.
+    If not set, the chunk lengths will be evenly distributed across all splits.
+    Note: This is an experimental flag and may change in the future."""
+
     moe_load_balancing_loss_mode_for_chunked_sequence: str = "reduce"
     """
     Mode for applying load balancing loss for chunked sequence in MoE layer.
@@ -1365,6 +1370,21 @@ class TransformerConfig(ModelParallelConfig):
             self.mtp_standalone = self.pipeline_model_parallel_layout.validate_layer_layout(
                 num_layers=self.num_layers, mtp_num_layers=self.mtp_num_layers
             )
+
+        # Chunked PP
+        if self.chunked_pipeline_model_parallel_splits > 1:
+            if self.chunked_pipeline_model_parallel_split_lengths is not None:
+                assert (
+                    len(self.chunked_pipeline_model_parallel_split_lengths)
+                    == self.chunked_pipeline_model_parallel_splits
+                ), (
+                    f"{self.chunked_pipeline_model_parallel_split_lengths=} "
+                    f"must be a list of length {self.chunked_pipeline_model_parallel_splits}."
+                )
+            # The sequence lengths vary across chunks, so we need to disable
+            # batch P2P comm and enable variable sequence lengths.
+            self.batch_p2p_comm = False
+            self.variable_seq_lengths = True
 
         # Uneven PP
         elif (
--- a/megatron/training/arguments.py
+++ b/megatron/training/arguments.py
@@ -2814,6 +2814,10 @@ def _add_distributed_args(parser):
                        help='Degree of pipeline model parallelism.')
     group.add_argument('--chunked-pipeline-model-parallel-splits', type=int, default=1,
                        help='Number of splits in chunked pipeline model parallel.')
+    group.add_argument('--chunked-pipeline-model-parallel-split-lengths', nargs='+', type=int, default=None,
+                       help='[Experimental] Custom chunk lengths for chunked pipeline model parallel.'
+                       'If not set, the chunk lengths will be evenly distributed across all splits.'
+                       'Note: This is an experimental flag and may change in the future.')
     group.add_argument('--moe-load-balancing-loss-mode-for-chunked-sequence', type=str, 
                        default="reduce",choices=["all_gather", "reduce"],
                        help='Mode for applying load balancing loss for chunked sequence in MoE layer.'


megatron/core/pipeline_parallel/chunked_pipeline_parallel_utils.py
```diff
@@ -0,0 +1,161 @@
...
+            prefix_seqlens=self.current_split_span[: self.current_span_idx],
+            this_chunk_seqlen=self.current_split_span[self.current_span_idx],
+            max_total_seqlen=self.current_seq_length,
+            kv_cache_pool=self.current_kv_cache_pool,
+            boundary_elements_for_mtp=self.current_boundary_elements_for_mtp,
+        )
+
+
+class ChunkedPipelineParallelQueue:
+    """A two-stage queue to store the chunked pipeline model parallel input or output tensors."""
```
Wohox, line 127: Please give more detail of how the two-stage queue work and a example is better if possible~

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -1101,15 +1114,32 @@ def enable_grad_sync():
 
     # Model chunk IDs with synchronized grads
     synchronized_model_chunks = set()
-
-    input_tensors = [[] for _ in range(len(model))]
-    output_tensors = [[] for _ in range(len(model))]
+    if config.chunked_pipeline_model_parallel_splits == 1:
```
Wohox, line 1117: Seems `List` is a special case of  `ChunkedPipelineParallelQueue` with `split=1`  , is it possible to unify these 2?

megatron/core/transformer/attention_cached_prefix.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # Save the context
+        ctx.seq_dim = seq_dim
+        ctx.key_cache = key_cache
+        ctx.value_cache = value_cache
+        ctx.grad_key_cache = grad_key_cache
+        ctx.grad_value_cache = grad_value_cache
+        ctx.save_for_backward(query, key_joined, value_joined, core_attn_out)
+
+        core_attn_out = core_attn_out.clone()
```
Wohox, line 138: Why create a clone here?

megatron/core/transformer/attention_cached_prefix.py
```diff
@@ -0,0 +1,189 @@
...
+
+import torch
+
+from megatron.core.cached_prefix_utils import CachedPrefixParams, KVCache
+from megatron.core.packed_seq_params import PackedSeqParams
+from megatron.core.transformer.enums import AttnMaskType
+from megatron.core.transformer.transformer_config import TransformerConfig
+
+
+class ChunkedKVCacheForAttention(KVCache):
```
Wohox, line 13: Maybe `ChunkedAttentionKVCache`?

megatron/core/transformer/attention_cached_prefix.py
```diff
@@ -0,0 +1,189 @@
...
+            self.grad_key_cache,
+            self.grad_value_cache,
+            attention_mask,
+            attn_mask_type,
+            attention_bias,
+            packed_seq_params,
+        )
+
+
+class AttentionFuncionWithChunkedKVCache(torch.autograd.Function):
```
Wohox, line 72: Maybe `ChunkedAttentionKVCacheFunction`?

megatron/core/pipeline_parallel/chunked_pipeline_parallel_utils.py
```diff
@@ -0,0 +1,161 @@
...
+            # Initialize the KV cache pool for the current micro batch
+            self.current_kv_cache_pool = KVCachePool(self.config)
+        else:
+            self.current_span_idx += 1
+
+        if seq_length != self.current_seq_length:
+            self.current_seq_length = seq_length
+            self.current_split_span = self._get_span(seq_length)
+
+    def get_current_cached_prefix_params(self) -> CachedPrefixParams:
```
Wohox, line 113: More like `build_current_cached_prefix_params` to me.

megatron/core/pipeline_parallel/chunked_pipeline_parallel_utils.py
```diff
@@ -0,0 +1,161 @@
...
+        """Get cached prefix params for the current chunk."""
+
+        return CachedPrefixParams(
+            prefix_seqlens=self.current_split_span[: self.current_span_idx],
+            this_chunk_seqlen=self.current_split_span[self.current_span_idx],
+            max_total_seqlen=self.current_seq_length,
+            kv_cache_pool=self.current_kv_cache_pool,
```
Wohox, line 120: What about moving creation of KVCachePool here?

megatron/core/transformer/transformer_block.py
```diff
@@ -740,6 +745,7 @@ def forward(
                     rotary_pos_emb=rotary_pos_emb,
                     attention_bias=attention_bias,
                     packed_seq_params=packed_seq_params,
+                    cached_prefix_params=cached_prefix_params,
```
Wohox, line 748: Does current implementation support full recompute now?

megatron/core/cached_prefix_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        self.config = config
+
+
+class KVCachePool:
+    def __init__(self, config: TransformerConfig):
+        self.config = config
+        self.kv_cache_pool = {}
+        self.block_prefixes = []
+
+    def block_range_push(self, suffix: str):
```
Wohox, line 25: If we use the module name as the key of kv cache, will KVCachePool be simpler? Something like `gpt_model.decoder.layer0.attention`, then we can get rid of `block_range_push/pop`.

megatron/core/transformer/enums.py
```diff
@@ -55,6 +55,7 @@ class AttnMaskType(enum.Enum):
     padding_causal = 4  # only used for thd attention
     arbitrary = 5
     causal_bottom_right = 6  # only used for TE
+    padding_causal_bottom_right = 7  # only used for TE
```
Wohox, line 58: Wondering does this PR require a specific TE version?

Wohox:
Nice work! Apart from the comments above, I have some small suggestions :
* Give detailed comments for newly added data structures. (AI-generated will do).
* 1F1B overlap is not supported due to missing input of `CachedPrefixParams` and `block_range_push/pop`, please add assertion in transformer_config as well.

megatron/core/pipeline_parallel/chunked_pipeline_parallel_utils.py
```diff
@@ -0,0 +1,161 @@
...
+                    'tokens': self.token_slices[self.current_span_idx + 1][:, 0],
+                    'labels': self.labels_slices[self.current_span_idx + 1][:, 0],
+                    'loss_mask': self.loss_mask_slices[self.current_span_idx + 1][:, 0],
+                    'position_ids': self.position_ids_slices[self.current_span_idx + 1][:, 0],
+                }
+
+        return slice_data
+
+    def _get_span(self, seq_length: int) -> List[int]:
+        spans = [seq_length // self.chunked_pp_splits] * self.chunked_pp_splits
```
Wohox: If we have uneven span, will this lead to lost tokens?

megatron/core/pipeline_parallel/chunked_pipeline_parallel_utils.py
yuzhongw-nvidia: `get_current_cached_prefix_params` may be called multiple times for one micro-batch, so I think we cannot create KVCachePool here.

megatron/core/transformer/transformer_block.py
yuzhongw-nvidia: No. Good catch. Let me add an assert in transformer config.

megatron/core/cached_prefix_utils.py
yuzhongw-nvidia: Yes. But how do we get the prefix inside each module? For example, how do we get the string `gpt_model.decoder.layer0.attention` in the `Attention` module?

Wohox: We can use `transformer_layer.layer_number` directly. You can double check the layer number calculation logic, since last I have checked, the `layer_number` in Decoder and MTP are continuous, for example on the last VPP stage with 1 decoder transformer layer and 1 MTP layer, the `layer_number` passed to `Attention` is like:
* `decoder.transformer_layer.attention.layer_number` = 0
* `mtp.layers[0].transformer_layer.attention.layer_number` = 1

jeromeku:
@yuzhongw-nvidia

Looking forward to this feature -- any estimates on when it will be ready?

PR #3160, opened 2026-01-30
HybridModel Examples [DO NOT SUBMIT]

Phlip79:
# What does this PR do ?

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/models/hl/common_layer_config.py
+++ b/megatron/core/models/hl/common_layer_config.py
@@ -199,7 +199,7 @@ class CommonLayerConfig:
     gated_linear_unit: bool = False
     """Use a gated linear unit for the first linear layer in the MLP."""
 
-    activation_func: Callable = F.gelu
+    activation_func: str = "gelu"
     """Activation function to use for the non-linearity in the MLP."""
 
     activation_func_fp8_input_store: bool = False
@@ -207,9 +207,28 @@ class CommonLayerConfig:
     The stored input is casted back to the original precision before backprop compuatation."""
 
     glu_linear_offset: float = 0.0
-    """Offset term in the GLU activation function: activation_func(x[0]) * (x[1] + offset). Only 
+    """Offset term in the GLU activation function: activation_func(x[0]) * (x[1] + offset). Only
     used when gated_linear_unit is True"""
 
-    activation_func_clamp_value: Optional[float] = None
+    activation_func_clamp_value: float | None = None
     """Clamp the output of the linear_fc1 in the activation function. Only used when activation_func
     is quick_gelu."""
+
+    ####################
+    # initialization
+    ####################
+    init_method: str | None = None
+    """Method to initialize weights. Note that bias is always set to zero. Should be a function that
+    takes a single Tensor and initializes it. If None, will be set to
+    megatron.core.utils.init_method_normal(init_method_std) which is torch nn init normal with
+    mean=0.0 and std=init_method_std."""
+
+    init_method_std: float = 0.02
+    """Standard deviation of the zero mean normal for the default initialization method, not used if
+    init_method and output_layer_init_method are provided."""
+
+    init_model_with_meta_device: bool = False
+    """
+    If True, initializes the model with the meta device. This is helpful for
+    training of very large models. This feature is only works when megatron fsdp is turned on.
+    """
--- a/megatron/core/models/hl/common_layer_config.py
+++ b/megatron/core/models/hl/common_layer_config.py
@@ -18,33 +18,6 @@ class CommonLayerConfig:
        (https://arxiv.org/abs/2205.05198) for more details.
     """
 
-    context_parallel_size: int = 1
-    """Splits network input along sequence dimension across GPU ranks."""
-
-    hierarchical_context_parallel_sizes: list[int] | None = None
-    """Degrees of the hierarchical context parallelism. Users should provide a list to specify
-       the sizes for different levels. Taking the a2a+p2p cp comm type as example, it contains
-       groups of two levels, so the first value of the list indicates the group size of the a2a
-       communication type, and the second value indicates the group size of the p2p communication
-       type.
-    """
-
-    max_seqlen_per_dp_cp_rank: int | None = None
-    """
-    Maximum sequence length per DPxCP rank. This is the maximum sequence length each rank
-    can handle without overflowing the memory. Typically, a good starting point is to set this
-    to maximum sequence length / context parallel size.
-    This is used to calculate the number and length of sub-samples assigned to
-    each rank when using hybrid_context_parallel.
-    """
-
-    hybrid_context_parallel: bool = False
-    """
-    If true, enables hybrid context parallel. This is used to balance the workload of
-    each CP rank when we use packed samples with variable sequence lengths.
-    Please set max_seqlen_per_dp_cp_rank when using hybrid_context_parallel.
-    """
-
     expert_model_parallel_size: int = 1
     """Distributes Moe Experts across sub data parallel dimension."""
 
--- a/megatron/core/models/hl/common_layer_config.py
+++ b/megatron/core/models/hl/common_layer_config.py
@@ -2,8 +2,6 @@
 
 from dataclasses import dataclass
 
-import torch
-
 
 @dataclass
 class CommonLayerConfig:
@@ -41,13 +39,10 @@ class CommonLayerConfig:
     ###################
     # Training
     ###################
-    fp16: bool = False
-    """If true, train with fp16 mixed precision training."""
+    mixed_precision_dtype: str = "fp32"
+    """dtype used for mixed-precision training."""
 
-    bf16: bool = False
-    """If true, train with bf16 mixed precision training."""
-
-    params_dtype: torch.dtype = torch.float32
+    params_dtype: str = "fp32"
     """dtype used when intializing the weights."""
 
     deterministic_mode: bool = False
@@ -57,7 +52,7 @@ class CommonLayerConfig:
     enable_autocast: bool = False
     """If true runs the forward step function inside torch.autocast context."""
 
-    autocast_dtype: torch.dtype | None = None
+    autocast_dtype: str | None = None
     """dtype to pass to torch.amp.autocast when enabled. If None, is set to pipeline_dtype."""
 
     num_microbatches_with_partial_activation_checkpoints: int | None = None
--- a/megatron/core/models/hl/examples/complex_model.py
+++ b/megatron/core/models/hl/examples/complex_model.py
@@ -119,12 +119,12 @@ PS = PipelineSplit()
 
 # Define each pipeline stage
 # Hybrid pattern with multiple layer types
-Stage1 = [Mamba, [LargeMoE, Mamba] * 2, GlobalAttention, [SmallMoE, Mamba] * 3, SlidingAttention]
+Stage1 = [Embedding, Mamba, [LargeMoE, Mamba] * 2, GlobalAttention, [SmallMoE, Mamba] * 3, SlidingAttention]
 Stage2 = [[LargeMoE, Mamba] * 3, SlidingAttention, [SmallMoE, Mamba] * 3, GlobalAttention]
 Stage3 = [[LargeMoE, Mamba] * 3, SlidingAttention, [SmallMoE, Mamba] * 4, GlobalAttention]
-Stage4 = [[LargeMoE, Mamba] * 4, SmallMoE]
+Stage4 = [[LargeMoE, Mamba] * 4, SmallMoE, Loss]
 
-layer_pattern = [Embedding, Stage1, PS, Stage2, PS, Stage3, PS, Stage4, Loss]
+layer_pattern = [Stage1, PS, Stage2, PS, Stage3, PS, Stage4]
 
 # =============================================================================
 # MODEL
--- a/megatron/core/models/hl/examples/complex_model.py
+++ b/megatron/core/models/hl/examples/complex_model.py
@@ -11,6 +11,7 @@ Demonstrates using multiple layer configurations:
 - SmallMoE: 32 experts (top-2), bf16, custom parallelism
 """
 
+from dataclasses import dataclass
 
 from megatron.core.models.hl import (
     AttentionLayerConfig,
@@ -19,11 +20,32 @@ from megatron.core.models.hl import (
     EmbeddingLayerConfig,
     HLModelConfig,
     LinearLayerConfig,
+    make_args_container,
     MambaLayerConfig,
     MTPLayerConfig,
     MoELayerConfig,
     PipelineSplit,
 )
+
+import tyro
+
+# =============================================================================
+# ARGUMENTS
+# =============================================================================
+
+@dataclass
+class ExtraArgs:
+    num_attention_heads: int = 32
+    large_moe_ffn_hidden_size: int = 1856
+
+
+ArgsContainer = make_args_container(
+    hl_model_config=HLModelConfig,
+    common_layer_config=CommonLayerConfig,
+    extra_args=ExtraArgs,
+)
+
+args = tyro.cli(ArgsContainer)
 
 # =============================================================================
 # COMMON CONFIGURATION
@@ -61,7 +83,7 @@ Mamba = MambaLayerConfig(
 
 GlobalAttention = AttentionLayerConfig(
     common_config=common_config,
-    num_attention_heads=32,
+    num_attention_heads=args.num_attention_heads,
     num_query_groups=2,
     kv_channels=128,
     use_flash_attention=True,
@@ -70,7 +92,7 @@ GlobalAttention = AttentionLayerConfig(
 
 SlidingAttention = AttentionLayerConfig(
     common_config=common_config,
-    num_attention_heads=32,
+    num_attention_heads=args.num_attention_heads,
     num_query_groups=2,
     kv_channels=128,
     use_flash_attention=True,
@@ -79,7 +101,7 @@ SlidingAttention = AttentionLayerConfig(
 
 LargeMoE = MoELayerConfig(
     common_config=moe_common_config,
-    ffn_hidden_size=1856,
+    ffn_hidden_size=args.large_moe_ffn_hidden_size,
     num_experts=128,
     top_k=6,
     router_score_function="sigmoid",
@@ -88,7 +110,7 @@ LargeMoE = MoELayerConfig(
     router_enable_expert_bias=True,
     router_dtype="fp32",
     moe_aux_loss_coeff=1e-4,
-    shared_expert_intermediate_size=3712,
+    shared_expert_intermediate_size=2 * args.large_moe_ffn_hidden_size,
     activation="squared_relu",
     token_dispatcher_type="allgather",
     grouped_gemm=True,
@@ -97,7 +119,7 @@ LargeMoE = MoELayerConfig(
 
 SmallMoE = MoELayerConfig(
     common_config=moe_common_config.update(expert_model_parallel_size=4),
-    ffn_hidden_size=3712,
+    ffn_hidden_size=2 * args.large_moe_ffn_hidden_size,
     num_experts=32,
     top_k=2,
     router_score_function="softmax",
--- a/megatron/core/models/hl/examples/complex_model.py
+++ b/megatron/core/models/hl/examples/complex_model.py
@@ -15,9 +15,9 @@ import torch
 
 from megatron.core.models.hl import (
     AttentionLayerConfig,
+    CommonLayerConfig,
     EmbeddingLayerConfig,
     HLModel,
-    LayerConfig,
     MambaLayerConfig,
     MoELayerConfig,
     PipelineSplit,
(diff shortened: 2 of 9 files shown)


megatron/core/models/hl/examples/complex_model.py
```diff
@@ -0,0 +1,147 @@
...
+"""
+
+from megatron.core.models.hl import (
+    HLModel,
+    HLModelConfig,
+    AttentionLayer,
+    MambaLayer,
+    MoELayer,
+    ParallelismConfig,
+)
```
janEbert, line 28: The `...Layer` classes should also be `...LayerConfig`s (or functions/lambdas) so that we don't instantiate any tensors just yet. Just to be very obvious to the user. Maybe it's also fine to create a `...hl.configs` module and import classes named `...Layer` from there. I need to think about what the best approach would be, I'm not quite sure yet. The former is more obvious, while the latter is more ergonomic. Functions/lambdas should probably not be considered due to serialization problems (if we care about those).

If we went with the `...Config` route, we'd want every layer to take its corresponding `...Config` as possibly its sole argument and instantiate from just that.

megatron/core/models/hl/examples/complex_model.py
```diff
@@ -0,0 +1,147 @@
...
+    vocab_size=131072,
+    max_sequence_length=8192,
+
+    # Hybrid pattern with multiple layer types
+    layer_pattern="""
+        M1 (E1 M1)x2 A1 (E2 M1)x3 A2 |
+        (E1 M1)x3 A2 (E2 M1)x3 A1    |
+        (E1 M1)x3 A2 (E2 M1)x4 A1    |
+        (E1 M1)x4 E2
+    """,
```
janEbert: I'm not sure if we want to stick with string patterns if we're already using Python. It seems arbitrarily limiting when we already have a complete language on our hands. :)
It also means we wouldn't have to implement a complete parser and users could benefit from LSPs and other linting to catch errors early.

E..g, with the below aliases defined (and an additional alias "|"→`PipelineSplit()`), this could be something like:
```python
[
    "M1", ["E1", "M1"] * 2, "A1", ["E2", "M1"] * 3, "A2", "|",
    ["E1", "M1"] * 3, "A2", ["E2", "M1"] * 3, "A1", "|",
    ["E1", "M1"] * 3, "A2", ["E2", "M1"] * 4, "A1", "|",
    ["E1", "M1"] * 4, "E2",
]
```

Alternatively, pipeline splits could be automatically placed whenever a second-level list ends:
```python
[
    ["M1", ["E1", "M1"] * 2, "A1", ["E2", "M1"] * 3, "A2"],
    [["E1", "M1"] * 3, "A2", ["E2", "M1"] * 3, "A1"]],
    [["E1", "M1"] * 3, "A2", ["E2", "M1"] * 4, "A1"],
    [["E1", "M1"] * 4, "E2"],
]
```

In the above examples, I used strings for the aliases; we should discuss whether this is worse than letting the user instantiate aliases as variables and use those instead to avoid all the quotes. However, at some point, serialization may become annoying to handle, so who knows.

megatron/core/models/hl/examples/complex_model.py
```diff
@@ -0,0 +1,147 @@
...
+            router_dtype="fp32",
+            moe_aux_loss_coeff=1e-2,
+            activation="swiglu",
+            token_dispatcher_type="alltoall",
+            grouped_gemm=True,
+            parallelism=ParallelismConfig(
+                tensor_parallel_size=8,
+                sequence_parallel=True,
+                expert_parallel_size=4,
+            ),
```
janEbert: Is it a thing to have two different, non-matching parallelization settings in the same job/model? I.e., here at the top, it has `tensor_parallel_size=8` with `expert_parallel_size=16` and `expert_tensor_parallel_size=8` vs. at the bottom `tensor_parallel_size=8` with `expert_parallel_size=4`.

If it's not a thing, the parallelism config should be separate from the model IMO.

megatron/core/models/hl/examples/complex_model.py
```diff
@@ -0,0 +1,147 @@
...
+    ParallelismConfig,
+)
+
+# =============================================================================
+# COMPLEX (IMAGINED) MODEL CONFIGURATION
+# =============================================================================
+
+model_config = HLModelConfig(
+    vocab_size=131072,
+    max_sequence_length=8192,
```
janEbert, line 76: Is there a reason not to have an explicit `Embedding` layer and prefer setting `vocab_size` separately?

About the `max_sequence_length`, I'm not sure whether it should remain here. On the one hand, pedantically, it should be passed to each position embedding layer like other arguments. On the other, it's always the same across the model, so no need to vary it for different position embedding layers. On yet another hand, it's also part of more than the model since it's used to provide matching training data.

megatron/core/models/hl/examples/complex_model.py
```diff
@@ -0,0 +1,147 @@
...
+            ),
+        ),
+    },
+
+    # Model settings
+    share_embeddings_and_output_weights=False,
+    position_embedding_type="none",
+    normalization="RMSNorm",
+    disable_bias_linear=True,
+    init_method_std=0.0173,
```
janEbert: Ideally, these would also be configured per-layer. They could still be referenced from a variable defined elsewhere, but having these be definable per-layer would be amazing for fine-grained customization, which is a goal we're pursuing. Especially initialization configuration is super important for more modern neural net training.

I'd also want us to add, similar to initialization settings, optional per-layer optimizer settings. Like `"lr": 2**0.5 * base_lr`. This would automatically be converted to the appropriate `param_groups` to pass to the optimizer.

megatron/core/models/hl/examples/complex_model.py
janEbert: Actually, for ergonomy, it doesn't really matter whether the user has to write `...Layer` or `...LayerConfig` since we assume that it'll be used to create an alias anyway.

megatron/core/models/hl/examples/complex_model.py
janEbert: Second version (implicit pipelining in second level) without quotes:
```python
[
    [M1, [E1, M1] * 2, A1, [E2, M1] * 3, A2],
    [[E1, M1] * 3, A2, [E2, M1] * 3, A1]],
    [[E1, M1] * 3, A2, [E2, M1] * 4, A1],
    [[E1, M1] * 4, E2],
]
```

janEbert: Explicit pipelining without quotes:

```python
PS = PipelineSplit()

[
    M1, [E1, M1] * 2, A1, [E2, M1] * 3, A2, PS,
    [E1, M1] * 3, A2, [E2, M1] * 3, A1, PS,
    [E1, M1] * 3, A2, [E2, M1] * 4, A1, PS,
    [E1, M1] * 4, E2,
]
```

janEbert: Also allows nested aliasing like
```python
PS = PipelineSplit()

P1 = [M1, [E1, M1] * 2, A1, [E2, M1] * 3, A2]
P2 = [[E1, M1] * 3, A2, [E2, M1] * 3, A1]
P3 = [[E1, M1] * 3, A2, [E2, M1] * 4, A1]
P4 = [[E1, M1] * 4, E2]

[P1, PS, P2, PS, P3, PS, P4]
```

megatron/core/models/hl/examples/complex_model.py
janEbert: `dtype`, too, actually.

Phlip79:
```
common_config = CommonConfig(dtype=bf16, ...)

M1 = MoELayer(..., config=common_config)

m2_config = common_config.update(dtype=fp8)
M2 = MoELayer(..., config=m2_config)

A1 = AttentionLayer(..., config=common_config)
```

megatron/core/models/hl/examples/complex_model.py
Phlip79: Updated.

megatron/core/models/hl/examples/complex_model.py
Phlip79: Done.

megatron/core/models/hl/examples/complex_model.py
Phlip79: Updated code.

megatron/core/models/hl/examples/complex_model.py
Phlip79: Discussed over chat. We want to support this.

megatron/core/models/hl/examples/complex_model.py
janEbert: For optimizer and initialization settings, we could create dedicated keys in the `CommonConfig`.

megatron/core/models/hl/examples/complex_model.py
```diff
@@ -0,0 +1,148 @@
...
+
+# =============================================================================
+# LAYER DEFINITIONS
+# =============================================================================
+
+Embedding = EmbeddingLayerConfig(
+    common_config=common_config,
+    vocab_size=131072,
+    max_sequence_length=8192,
+    position_embedding_type="none",
```
janEbert, line 77: Minor nit: I think we should have separate `VocabEmbedding` and `PositionEmbedding`.

As for `PositionEmbedding`, there would probably be two ways to pass it – as an argument to individual layers and at the beginning of the model for "globally" passing it to each layer (i.e., reusing the same pre-calculated pos embedding). This is something we also need to put more thought into, it's not clear to me yet how to properly approach the different options we need to provide here.

megatron/core/models/hl/examples/complex_model.py
```diff
@@ -0,0 +1,148 @@
...
+PS = PipelineSplit()
+
+# Define each pipeline stage
+# Hybrid pattern with multiple layer types
+Stage1 = [Mamba, [LargeMoE, Mamba] * 2, GlobalAttention, [SmallMoE, Mamba] * 3, SlidingAttention]
+Stage2 = [[LargeMoE, Mamba] * 3, SlidingAttention, [SmallMoE, Mamba] * 3, GlobalAttention]
+Stage3 = [[LargeMoE, Mamba] * 3, SlidingAttention, [SmallMoE, Mamba] * 4, GlobalAttention]
+Stage4 = [[LargeMoE, Mamba] * 4, SmallMoE]
+
+layer_pattern = [Stage1, PS, Stage2, PS, Stage3, PS, Stage4]
```
janEbert, line 156: `(Vocab)Embedding` should be part of the layer pattern. The previous design doc also had a `Loss` in the layer pattern (due to MTP), which we haven't considered yet. I guess it could be added by default at the end of the model if it's not inserted by the user, so that part would be fine here.

Not sure about `PositionEmbedding`, [see comment above](https://github.com/NVIDIA/Megatron-LM/pull/3160/changes#r2755835857).

megatron/core/models/hl/examples/complex_model.py
```diff
@@ -0,0 +1,148 @@
...
+    tensor_model_parallel_size=8,
+    sequence_parallel=True,
+)
+
+# MoE-specific configuration (copy of common_config with expert parallelism)
+moe_common_config = replace(
+    common_config,
+    expert_model_parallel_size=16,
+    expert_tensor_parallel_size=8,
+)
```
janEbert: Maybe prefer a method instead:
```python
moe_common_config = common_config.update(
    expert_model_parallel_size=16,
    expert_tensor_parallel_size=8,
)
```

I think this keeps things more nicely connected. A user immediately knows that a `Config`-related function is used.

megatron/core/models/hl/examples/complex_model.py
janEbert: Unresolving this to consider keeping this as an extension for later. I believe it will require extensive changes to `ParallelState` and friends and may violate existing assumptions. This probably requires a lot of care to get right.

megatron/core/models/hl/examples/complex_model.py
Phlip79: Done.

megatron/core/models/hl/common_layer_config.py
```diff
@@ -0,0 +1,234 @@
...
+    """If true, weights are initialized. This option can be useful when you know you are going to
+       load values from a checkpoint.
+    """
+
+    use_cpu_initialization: bool = False
+    """When set to False, we initialize the weights directly on the GPU. CPU initialization is the
+       same regardless of tensor model parallelism, but GPU initialization is not. Transferring
+       weights from CPU to GPU can take a significant amount of time for large models.
+    """
+
```
duncanriach: Why is TP not included here?
Why is the number of PP ranks not included here?

megatron/core/models/hl/common_layer_config.py
```diff
@@ -0,0 +1,234 @@
...
+    autocast_dtype: torch.dtype | None = None
+    """dtype to pass to torch.amp.autocast when enabled. If None, is set to pipeline_dtype."""
+
+    num_microbatches_with_partial_activation_checkpoints: int | None = None
+    """If int, set the number of microbatches where not all of the layers will be checkpointed and
+       recomputed. The rest of the microbatches within the window of maximum outstanding
+       microbatches will recompute all layers (either full recompute or selective recompute). If
+       None, the checkpoint and recompute will be left up to the forward_step function.
+
+    """
```
duncanriach, line 64: Thinking about the various recipes for low-precision (such as FP8), where the first N layers and the last M layers are in a higher precision. Wondering how things like this would be specified.

megatron/core/models/hl/examples/complex_model.py
```diff
@@ -0,0 +1,142 @@
...
+
+# =============================================================================
+# MODEL
+# =============================================================================
+
+complex_model = HLModelConfig(
+    common_config=common_config,
+    layer_pattern=layer_pattern,
+    tensor_model_parallel_size=8,
+    untie_embeddings_and_output_weights=True,
```
duncanriach, line 166: Need to specify here the number of physical pipeline parallel ranks there are (--pipeline-model-parallel-size), so that they stages specified above can be allocated across them. If pp size is 4, then there will be only one VPP rank. If pp size is 8, then there will be two VPP ranks.

megatron/core/models/hl/common_layer_config.py
```diff
@@ -0,0 +1,234 @@
...
+    # Model parallelism
+    ###################
+    sequence_parallel: bool = False
+    """Makes tensor parallelism more memory efficient for LLMs (20B+) by parallelizing layer norms
+       and dropout sequentially.  See Reducing Activation Recomputation in Large Transformer Models
+       (https://arxiv.org/abs/2205.05198) for more details.
+    """
+
+    context_parallel_size: int = 1
+    """Splits network input along sequence dimension across GPU ranks."""
```
duncanriach: Why is this a per-layer thing and not a per-model thing?

This is in CommonLayerConfig, but is the idea that this would be the same over all layers? I'm wondering why there is some config in CommonLayerConfig and some in HLModelConfig.

megatron/core/models/hl/common_layer_config.py
janEbert: Oh yes, sorry; the FP8 config values should be included and the `fp16` and `bf16` flags should also be merged to a common `dtype` flag. Then, a user could have a `fp8_config` that sets `dtype="fp8"` (and/or similar recipe-required flags) and a `bf16_config` that could be created like the following to ensure the same settings except for the `dtype`:
```python
bf16_config = fp8_config.update(dtype="bf16")
```

Then you should be able to do the following:
```python
first_half = list_of_layers_with_fp8_config * N
second_half = list_of_layers_with_bf16_config * M
layer_config = first_half + second_half
```

megatron/core/models/hl/examples/complex_model.py
janEbert: Great point; `pipeline_model_parallel_size` was supposed to be part of the `HLModelConfig`, i.e., a "global" argument, but I lost it along the way. Sorry about that.

megatron/core/models/hl/common_layer_config.py
janEbert: Somehow, I was theorizing that this setting could in theory be varied per-layer, but you're right, that doesn't make any sense and it should be a global value in `HLModelConfig`.

I have to re-check expert model parallelism as well; I was simply assuming based on my knowledge of TorchTitan that it was a subgroup of tensor model parallelism, but that may not be the case, thus it also would not be feasible to vary it per-layer.

The thinking with some settings being available here was to have any setting that, in theory and without immense implementation complexity, could be applied per-layer. We could still disallow having the settings varied per-layer when the implementation does not support varying them yet.

megatron/core/models/hl/common_layer_config.py
duncanriach: Oh yeah. This is an example of how being able to change these parameters per layer means that these things can be easily experimented with and set a particular way.

megatron/core/models/hl/hl_model_config.py
```diff
@@ -24,6 +24,22 @@ class HLModelConfig:
...
+
+       For example:
+       - When 4 pipeline stages are specified in the layer pattern (i.e., 3 `PipelineSplit`s are
+         present) and this is set to 1 (default), there will be 4 physical pipeline stages with no
+         virtual blocks.
+       - When 4 pipeline stages are specified in the layer pattern (i.e., 3 `PipelineSplit`s are
+         present) and this is set to 2, there will be 4 physical pipeline stages with 2 virtual
+         blocks in each physical stage, resulting in 8 total pipeline stages.
+    """
+
```
duncanriach: I don't think this should be explicitly provided. This is implicit from the pipelining in the layer pattern. The thing that needs to be specified is the number of pipeline ranks (--pipeline-model-parallel-size). Then the virtual pipeline model parallel size is calculated by num-pp-ranks/num-pp-stages. If there are two ranks but six stages, then vp size will be three.

janEbert: Ah I think I understand the advantage. By configuring `pipeline_model_parallel_size` instead, you can have _fewer_ physical pipeline ranks than pipeline stages. Is my understanding correct?

duncanriach: That's right. That's how VPP works. For PP, we have to specify the number of PP ranks to be able to configure the GPUs. VPP is used to reduce the pipeline bubble size by breaking the model into smaller chunks and interleaving the pipeline parallel. Before flexible virtual pipeline parallel, the number of layers in each stage was the same, and a parameter like virtual_pipeline_model_parallel_size could be used to specify that the model should be partitioned into stages to fill the PP ranks over all VPP ranks. In this HL model, the partitioning of the model into stages is done in the layer pattern. The number of stages then needs to be a multiple of the number of physical pp ranks.

I don't think we should support the old way of doing PP and VPP because they can be implemented using the new way, by specifying the partitioning in the layer pattern.

duncanriach: This is a feature that I'm quite familiar with. I'm concerned that there will be issues like this that are not caught on features that I don't know about. Need lots of review and discussion on this design

duncanriach: I've been looking at how I implemented fVPP in GitLab 2506. The layer pattern is used to define `virtual_pipeline_model_parallel_size` in `arguments.py`, and then this is used to have the right number of pipeline parallel model chunks and to schedule them (it's used by the shared pipeline parallel logic). I don't know if this parameter should exist at this level. I don't understand where API threshold is for this proposed model. If the number of pipeline segments is set in the pattern by the "user", then `virtual_pipeline_model_parallel_size` should not be set by the user, even though it might need to be calculated internally.

megatron/core/models/hl/examples/complex_model.py
```diff
@@ -0,0 +1,150 @@
...
+    activation="swiglu",
+    token_dispatcher_type="alltoall",
+    grouped_gemm=True,
+)
+
+MTP = MTPLayerConfig(
+    common_config=common_config,
+    eh_proj=LinearLayerConfig(common_config=common_config),
+    embedding=Embedding,
+)
```
duncanriach, line 138: This does not reflect how MTP is configured in the hybrid model. MTP is composed of one or more MTP layers, each of which can contain a number of model layers. At the moment, I believe that all the MTP layers have to contain the same pattern of layers (and they can optionally map to the same underlying layers). I also believe that all MTP layers have to be in the same pipeline stage.

For the hybrid model, MTP is currently specified like this in hybrid-override-pattern: "M-M-M*-/M-/M-". The / use used to delimit the start of an MTP layer.

For pipelining, I believe that the GPT model can put the MTP layers into a pipeline stage on their own, with the loss in its own pipeline stage. I am currently implementing fVPP for the hybrid model and I'm not sure if I'll add this functionality at the moment, possible only supporting pipelining the main part of the model, with the MTP always going in the final stage with the loss. However, this would prevent the implementation of the GPT model functionality with the hybrid model

Issue #3097, opened 2026-01-27
[QUESTION] Potential Logic Issue in `common.py` Test Validation


cailun01:
Hi, Megatron-LM team! 

I've been using Megatron-LM's regression testing framework to validate model convergence. During testing, I encountered a case where two training runs had significant differences (179 out of 200 steps failed to match), yet the `DeterministicTest` still passed.

**Test configuration:**
- Total steps: 200
- Allowed failing steps: 2 (calculated as `min(max(200 // 100, 1), 50)`)
- Actual matching steps: 21
- Actual failing steps: 179
- Match rate: 10.5%

**Expected behavior:** Test should fail (179 failures >> 2 allowed)  
**Actual behavior:** Test passed

### Question

Looking at the validation logic in [tests/functional_tests/python_test_utils/common.py](https://github.com/NVIDIA/Megatron-LM/blob/b47c376f5ceb8fa9f2cb1ba8510112b668e41366/tests/functional_tests/python_test_utils/common.py#L230) line 230:

```python
passing = np.mean(is_close) >= (num_failing_steps_allowed / total_steps_evaluated)
```

This checks if `match_rate >= allowed_failure_ratio` (e.g., 10.5% >= 1%).

Should this instead check if `match_rate >= (1 - allowed_failure_ratio)` to ensure the actual failing steps don't exceed the allowed limit?

### Analysis

The current logic:
```python
match_rate >= allowed_failure_ratio
→ (T - F) / T >= A / T
→ F <= T - A
```

For my case: `179 <= 198` → Pass (but seems incorrect)

The intended logic (I believe):
```python
match_rate >= (1 - allowed_failure_ratio)
→ (T - F) / T >= (T - A) / T
→ F <= A
```

For my case: `179 <= 2` → Fail (expected)

Where:
- T = total steps (200)
- A = allowed failing steps (2)
- F = actual failing steps (179)

### Proposed Fix

Adjust the comparison threshold
```python
passing = np.mean(is_close) >= (1 - num_failing_steps_allowed / total_steps_evaluated)
```

### Why This Wasn't Caught Earlier

In typical test scenarios with high match rates (98-100%), both logic variants return True, so the issue remained hidden. My case with a low match rate (10.5%) exposed this discrepancy.

### Environment

- Megatron-LM: [latest main branch](https://github.com/NVIDIA/Megatron-LM/blob/b47c376f5ceb8fa9f2cb1ba8510112b668e41366)
[...]

PR #3171, opened 2026-01-30
Logic issue in common.py test validation (#3097)

Ratheesh1104:
## What does this PR do?

Fixes a logic bug in deterministic regression test validation where runs with many mismatched steps could incorrectly pass. The validation now correctly fails when the number of failing steps exceeds the configured allowance, ensuring determinism and convergence checks behave as intended.

## Why is this change needed?

The previous logic allowed false positives when the mismatch rate was high. This issue went largely unnoticed because typical runs have >99% step match rates, but it becomes evident when match rates drop significantly. This fix improves the reliability and trustworthiness of regression test results.

## What changed?

-Corrected the validation logic to strictly enforce the configured failure threshold.
-Ensured test runs fail when actual failing steps exceed the allowed limit.

## Impact

-Prevents incorrect passing of non-deterministic or divergent runs.
-Improves confidence in regression test outcomes without affecting normal high-match scenarios.

### DIFF
--- a/tests/functional_tests/python_test_utils/common.py
+++ b/tests/functional_tests/python_test_utils/common.py
@@ -227,7 +227,9 @@ def pipeline(
                 is_close = np.isclose(actual, golden, rtol=test.rtol, atol=test.atol)
 
                 num_failing_steps_allowed = min(max(total_steps_evaluated // 100, 1), 50)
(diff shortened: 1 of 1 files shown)


Ratheesh1104:
Is it rejected?

Phlip79:
Please fix linting error and update to main.

Ratheesh1104:
@Phlip79 Sure , thanks

PR #3190, opened 2026-02-02
[M-FSDP] Refactor uneven dtensor to full tensor and add UT

shjwudp:
# What does this PR do ?

1. Refactor the `gather_uneven_dtensor_to_full_tensor` function to `uneven_dtensor_to_full_tensor`, removing its dependency on device mesh flattening for improved robustness.
2. Add unit tests for uneven dtensor functions.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

3. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


dimapihtar:
Hi @shjwudp, could you please fix unit tests?

shjwudp:
> Hi @shjwudp, could you please fix unit tests?

Hi @dimapihtar , thanks for the heads-up. It looks like the exception was caused by an OOM issue — I’ll try to fix it.

tests/unit_tests/distributed/megatron_fsdp/test_mfsdp_uneven_dtensor.py
```diff
@@ -0,0 +1,762 @@
...
+    # Example pattern: [Shard(0), _StridedShard(0, split_factor=mesh_shape[0])]
+    # so that combined, dim 0 is split across mesh_shape[0] * mesh_shape[1] ranks. [web:146]
+    placements = [Shard(0), _StridedShard(0, split_factor=mesh_shape[0])]
+
+    dtensor = distribute_tensor(global_tensor, mesh, placements)
+
+    gathered = uneven_dtensor_to_full_tensor(dtensor)
+
+    assert gathered.shape == global_tensor.shape
+    assert torch.allclose(gathered, global_tensor, rtol=1e-5, atol=1e-5)
```
cspades, line 413: Please correct me if I'm wrong, but shouldn't this be:
```
placements = [_StridedShard(0, split_factor=mesh_shape[1]), Shard(0)]
```
if we want right-to-left sharding instead of left-to-right sharding?

My current understanding is that `distribute_tensor` and `DTensor.redistribute` is a `for` loop that should go from left-to-right, and when it sees `_StridedShard`, instead of cutting the sharding dimension by half, it cuts it into `split_factor * mesh_shape[0]` pieces and combines the strided pieces so that the next `Shard(0)` can then cut those pieces as if we had sharding from right-to-left?

So for example, if `rows=16` and `(2,4)`, I _think_ instead of `[(0,1), (2,3), (4,5), (6,7)]` (on Rank (0, 0-3)) it will end up as `[(0,4), (1,5), (2,6), (3,7)]` because we have a split factor of 2 on the 0123 and 4567 shards. Wonder if this unit test passes because `rows=8`, in which case (after the initial sharding produces 8 / 2 = 4 elements per initial shard) the split factor has no effect: `[(0,), (1,), (2,), (3,)]` and `_StridedShard ~ Shard` in this case.

Maybe the uneven logic you wrote already accounts for this, since I think the `rows=8` and `(2,2)` would also fail for the same reason if my claim is true; have we tested this on 4 GPUs? 👀

cspades: I see: `gather_and_compute_chunk_metadata`
```
                if isinstance(p, _StridedShard):
                    # Re-position all _StridedShard to the beginning,
                    # because DTensor-based distribution functions
                    # assume left-to-right processing of _StridedShard.
                    reversed_shard_order.append(mesh_dim)
                    mesh_dims.remove(mesh_dim)
                    strided_shard_count += 1
```

shjwudp: Are you saying the unit test failed because the placement was incorrect? Interesting — I thought it was due to older versions of PyTorch not supporting distribute `_StridedShard`.

cspades: I was actually commenting on this unit test, I didn't read the old one. 👀 

I think the only things that could be good to check here to make sure we are on the right track are:

- What does the `distribute_tensor` output look like? Does it match our expectations for DP-TP right-to-left sharding?
- Test on 4 GPUs / `rows=8` or 8 GPUs / `rows=16` so we can be sure that we are also testing a non-trivial stride, i.e. to ensure that `distribute_tensor` and `uneven_dtensor_to_full_tensor` are inverse operations when operating on strided shards. (IIRC, `uneven_dtensor_to_full_tensor` requires the shards to be in the correct order, I think striding should break this test with the suggested arguments.)
- Add the same test for `[_StridedShard(0, split_factor=mesh_shape[1]), Shard(0)]` which is the native way to define 2-D strided / right-to-left sharding.

cspades: For TE-TP work I also did some tests to visualize this:
```
"""
What does this shard look like for a DP4 x TP2 mesh of GPU's?

Global Tensor: [0, 1, 2, 3, 4, 5, 6, 7]

DP4 x TP2 Sharding Strategy

┌───┬───┐
│ 0 │ 4 │ ← DP=0
├───┼───┤
│ 1 │ 5 │ ← DP=1
├───┼───┤          FSDP all-gather happens across the DP ranks, so we need to form the 0-3 and 4-7 TP shards!
│ 2 │ 6 │ ← DP=2
├───┼───┤
│ 3 │ 7 │ ← DP=3
└───┴───┘
  ↑   ↑
TP=0 TP=1

which is exactly what we want for TE, i.e. we shard on TP first, then fully_shard() on DP!
"""

# Global DP-TP Tensor
Global DP-TP Tensor: DTensor(local_tensor=tensor([0., 1., 2., 3., 4., 5., 6., 7.], device='cuda:0', dtype=torch.bfloat16), device_mesh=DeviceMesh((dp=4, tp=2), 'cuda', stride=(2, 1)), placements=(Replicate(), Replicate()))

# Redistribute to _StridedShard x Shard. 
Local DP-TP Shard (World Rank 0 / DP Rank 0 / TP Rank 0): DTensor(local_tensor=tensor([0.], device='cuda:0', dtype=torch.bfloat16), device_mesh=DeviceMesh((dp=4, tp=2), 'cuda', stride=(2, 1)), placements=(_StridedShard(dim=0, sf=2), Shard(dim=0)))
Local DP-TP Shard (World Rank 1 / DP Rank 0 / TP Rank 1): DTensor(local_tensor=tensor([4.], device='cuda:1', dtype=torch.bfloat16), device_mesh=DeviceMesh((dp=4, tp=2), 'cuda', stride=(2, 1)), placements=(_StridedShard(dim=0, sf=2), Shard(dim=0)))
Local DP-TP Shard (World Rank 2 / DP Rank 1 / TP Rank 0): DTensor(local_tensor=tensor([1.], device='cuda:2', dtype=torch.bfloat16), device_mesh=DeviceMesh((dp=4, tp=2), 'cuda', stride=(2, 1)), placements=(_StridedShard(dim=0, sf=2), Shard(dim=0)))
Local DP-TP Shard (World Rank 3 / DP Rank 1 / TP Rank 1): DTensor(local_tensor=tensor([5.], device='cuda:3', dtype=torch.bfloat16), device_mesh=DeviceMesh((dp=4, tp=2), 'cuda', stride=(2, 1)), placements=(_StridedShard(dim=0, sf=2), Shard(dim=0)))
Local DP-TP Shard (World Rank 4 / DP Rank 2 / TP Rank 0): DTensor(local_tensor=tensor([2.], device='cuda:4', dtype=torch.bfloat16), device_mesh=DeviceMesh((dp=4, tp=2), 'cuda', stride=(2, 1)), placements=(_StridedShard(dim=0, sf=2), Shard(dim=0)))
Local DP-TP Shard (World Rank 5 / DP Rank 2 / TP Rank 1): DTensor(local_tensor=tensor([6.], device='cuda:5', dtype=torch.bfloat16), device_mesh=DeviceMesh((dp=4, tp=2), 'cuda', stride=(2, 1)), placements=(_StridedShard(dim=0, sf=2), Shard(dim=0)))
Local DP-TP Shard (World Rank 6 / DP Rank 3 / TP Rank 0): DTensor(local_tensor=tensor([3.], device='cuda:6', dtype=torch.bfloat16), device_mesh=DeviceMesh((dp=4, tp=2), 'cuda', stride=(2, 1)), placements=(_StridedShard(dim=0, sf=2), Shard(dim=0)))
Local DP-TP Shard (World Rank 7 / DP Rank 3 / TP Rank 1): DTensor(local_tensor=tensor([7.], device='cuda:7', dtype=torch.bfloat16), device_mesh=DeviceMesh((dp=4, tp=2), 'cuda', stride=(2, 1)), placements=(_StridedShard(dim=0, sf=2), Shard(dim=0)))

# Redistribute back to un-sharded form. 
Global DP-TP Replicate Tensor: DTensor(local_tensor=tensor([0., 1., 2., 3., 4., 5., 6., 7.], device='cuda:0', dtype=torch.bfloat16), device_mesh=DeviceMesh((dp=4, tp=2), 'cuda', stride=(2, 1)), placements=(Replicate(), Replicate(
```
Code:
```
import torch
from torch.distributed.tensor import DTensor
from torch.distributed.tensor.placement_types import Shard, _StridedShard, Replicate

# Dist Init
torch.distributed.init_process_group()
WORLD_SIZE = torch.distributed.get_world_size()
rank = torch.distributed.get_rank()
torch.cuda.set_device(rank)

mesh = torch.distributed.init_device_mesh("cuda", mesh_shape=(4,2), mesh_dim_names=("dp", "tp"))

# Global Tensor
t = torch.tensor([i for i in range(WORLD_SIZE)], dtype=torch.bfloat16, device="cuda")

dt = DTensor.from_local(
    t,
    mesh,
    placements=(Replicate(), Replicate())
)

if torch.distributed.get_rank() == 0:
    print(f"Global DP-TP Tensor: {dt}")

# Shard to TP-DP.
sharded_dt = dt.redistribute(
    mesh,
    placements=(_StridedShard(0, split_factor=mesh["tp"].size()), Shard(0))
    # placements=(Shard(0), Shard(0))
)

print(f"Local DP-TP Shard (World Rank {rank} / DP Rank {mesh['dp'].get_local_rank()} / TP Rank {mesh['tp'].get_local_rank()}): {sharded_dt}")

# Shard TP, but do not shard DP. (Forward Pass) 
tp_sharded_dt = sharded_dt.redistribute(
    mesh,
    placements=(Replicate(), Shard(0))
)

if mesh['dp'].get_local_rank() == 0:
    print(f"Local TP Shard (World Rank {rank} / TP Rank {mesh['tp'].get_local_rank()}): {tp_sharded_dt}")

# Shard DP, but do not shard TP. (Never happens.) 
dp_sharded_dt = sharded_dt.redistribute(
    mesh,
    placements=(_StridedShard(0, split_factor=mesh['tp'].size()), Replicate())
)

if mesh['tp'].get_local_rank() == 0:
    print(f"Local DP Shard (World Rank {rank} / DP Rank {mesh['dp'].get_local_rank()}): {dp_sharded_dt}")

# Unshard again. 
dt = sharded_dt.redistribute(
    mesh,
    placements=(Replicate(), Replicate())
)

if torch.distributed.get_rank() == 0:
    print(f"Global DP-TP Replicate Tensor: {dt}")

torch.distributed.destroy_process_group()
```

shjwudp: Could you point out the parts of the code where the order of `_StridedShard` and `Shard` in placements actually matters? I still don’t quite understand this part.

From my understanding, the order of elements in placements shouldn’t matter as long as it correctly corresponds to the mesh dims.

Regarding `_StridedShard` and `Shard`, I understand that `_StridedShard` is always applied after Shard, just as described in the torch documentation: [link](https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/placement_types.py#L634).

megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py
```diff
@@ -247,148 +245,144 @@ def preprocess_state_dict_for_uneven_dtensor(state_dict: dict) -> dict:
     return state_dict
 
 
-def gather_uneven_dtensor_to_full_tensor(
-    dtensor: DTensor, target_device: Optional[torch.device] = None
-) -> DTensor:
+def uneven_dtensor_to_full_tensor(dtensor: DTensor) -> torch.Tensor:
```
cspades, line 256: Do we need to deprecate the optional `target_device`?
```
full_tensor = torch.zeros(dtensor.shape, dtype=dtensor.dtype, device=target_device)
```
I suppose the new recommendation is to `gather_uneven_dtensor_to_full_tensor(dtensor).cpu()/.cuda(device_rank)` if users want to move it to a different device?

megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py
```diff
@@ -247,148 +245,144 @@ def preprocess_state_dict_for_uneven_dtensor(state_dict: dict) -> dict:
...
+        offset = chunk_info["offset"]
+        slices = tuple(slice(o, o + s) for o, s in zip(offset, local_chunk.shape))
+        debug_slices.append(slices)
+
+    # Reconstruct the full tensor by placing chunks at their correct offsets
+    full_tensor = torch.zeros(dtensor.shape, dtype=dtensor.dtype, device=dtensor.device)
+    for chunk_info, local_chunk in zip(local_chunks_info, all_local_chunks):
+        offset = chunk_info["offset"]
+        slices = tuple(slice(o, o + s) for o, s in zip(offset, local_chunk.shape))
+        full_tensor[slices] = local_chunk
```
cspades: Nit: Debug copy of the for loop? Want to do anything with `debug_slices`?

megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py
```diff
@@ -247,148 +245,144 @@ def preprocess_state_dict_for_uneven_dtensor(state_dict: dict) -> dict:
...
+
+    return full_tensor
+
+
+def redistribute_uneven_dtensor_to_replicated(dtensor: DTensor) -> DTensor:
```
cspades, line 369: Nit: Do we need to change the name of the function (from `gather_uneven_dtensor_to_full_tensor`)? It looks like it has identical behavior to the original, and this is a public-facing API I'm currently using in BioNeMo, saves us a backwards incompatibility.

cspades (approved):
LGTM, mostly just need to make sure our tests are testing sensible cases, i.e. `(_StridedShard(dim=0, sf=k), Shard(dim=0))` and using more data to be able to see / verify the strides, that would be great!

🦀 One more use of `DeviceMesh._flatten()` is gone! 🦀 

How `_StridedShard` works is it first pre-partitions the data by the stride / shard factor, shards each part according to the placement strategy, and then concatenates the parts to get strided shards. Sharding is performed from left-to-right in the mesh dimension order, so `_StridedShard` is only sensible when it precedes other `Shard`(s). Let's test both cases at least, such that our gather code is not affected by non-conventional sharding strategies!

megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py
shjwudp: Since CPU and CUDA devices require different `process_group` → `device_mesh` configurations, we can’t actually select the target device directly. Therefore, this argument has been deprecated.

megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py
shjwudp: Yes, this is for debugging purposes; delete it.

megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py
shjwudp: Yes, should keep `gather_uneven_dtensor_to_full_tensor`, but `redistribute_uneven_dtensor_to_replicated` might be a better name. Mark `gather_uneven_dtensor_to_full_tensor` as deprecate in future.

cspades:
Rebased on main and rerunning, a ton of tests failed to TCP store error, likely random networking issue?

cspades:
/ok to test [3b9e020](https://github.com/NVIDIA/Megatron-LM/pull/3190/commits/3b9e0209996515e917b69af2b17d3d892d3b9a88)

PR #3254, opened 2026-02-04
Nano QAT/D fix with sft tokenizer and datasets

ChenhanYu:
# What does this PR do ?

**Megatron Core Fix:**
A change is made such that `expert_bias` is only updated when the module is in training mode. This is a special case where student is in training mode but the teacher is not when using Model-Optimizer for online knowledge distillation. 

**Others:**
1. The `SFTTokenizer` only supports `nemotron-h-aligned` prompt format to produce target that has answer-only loss mask. For example, we only want to include the CE loss for the assistant replies rather than the system or user prompts. It needs to identify the token indices what are not part of the assistant reply and replace with `IGNORE_INDEX` (where a loss_mask is computed later). HuggingFace tokenizer actually support this feature and provide the mask as `attention_mask` if specific `generation` key words are identified.
2. Adding `tool` role into the logic (treated as the same as `user`) where we replace the target tokens with `IGNORE_INDEX`.
3. This PR ports the `identity` option with an `identity_template` which assume that the message data has already been decorated by some template (e.g. nemotron-h chat template). We also adds an `default` option which will apply the `chat_template` from the tokenizer and assume OAI message format. For the `default` option, we return target directly without replacing any `IGNORE_INDEX`. As a result, all token losses are used.
4. `SFTDataset` uses packed data format where `thd` instead of `sbhd` qkv_format is used. This requires the original `attn_mask_type` to be `causal`. This PR remove the CP=1 specialization (using abitrary mask for speculative decoding TTT) in the `model_provider` to avoid `TEDotProductAttention` error.

**Integration Test**
1. Tested with `NVIDIA-Nemotron-3-Nano-30B-A3B-BF16` FP8/NVFP4 PTQ and QAT/D
2. TP=1, EP=4, CP={1,4}, PP={1,4}
3. Fake-quant checkpoints can be resumed to continue training. 0-shot MMLU is normal.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

6. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

#### (Step 3): Final Review

1. Add `Final Review` label
4. GitHub auto-assigns final reviewers based on your changes. They will get notified and pick up your PR soon.

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/training/tokenizer/sft_tokenizer.py
+++ b/megatron/training/tokenizer/sft_tokenizer.py
@@ -140,7 +140,7 @@ class SFTTokenizer(MegatronLegacyTokenizer):
             turn_len = len(turn_tokens)
 
             role = turn["role"].lower()
-            if role in ("system", "user"):
+            if role in ("system", "user", "tool"):
                 target[idx : idx + turn_len] = IGNORE_INDEX
             elif role == "assistant":
                 if self._prompt_config.assistant_prefix_len > 0:


megatron/training/tokenizer/sft_tokenizer.py
```diff
@@ -98,6 +106,9 @@ def tokenize_conversation(
 
         target = tokens.copy()
```
jenchen13, line 116: can you also update line 119
`assert conversation[turn_idx-1]["role"].lower() in ["user", "tool"]`

megatron/training/tokenizer/sft_tokenizer.py
```diff
@@ -58,6 +58,14 @@ def __init__(
                 has_bos=False,
                 has_system_role=True,
```
jenchen13, line 60: also add identity template at the top
`identity_template = """{% for message in messages %}{{ message['content'] }}{% endfor %}"""`
then change line 65 to `custom_chat_template=identity_template`

then line 109-110 should not be necessary

megatron/training/tokenizer/sft_tokenizer.py
ChenhanYu: Done.

megatron/training/tokenizer/sft_tokenizer.py
ChenhanYu: Done.

megatron/training/tokenizer/sft_tokenizer.py
```diff
@@ -58,6 +59,22 @@ def __init__(
...
             )
+        elif prompt_format == "identity":
+            self._prompt_config = PromptConfig(
+                assistant_prefix_len=0,
+                pad_token_id=tokenizer.convert_tokens_to_ids("<unk>"),
+                custom_chat_template=dentity_template,
```
jenchen13:
```suggestion
                custom_chat_template=identity_template,
```

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -299,7 +299,11 @@ def _update_router_expert_bias(model: List[torch.nn.Module], config: Transformer
     expert_bias_list = []
     for model_chunk in model:
         for module in get_attr_wrapped_model(model_chunk, 'modules')():
-            if hasattr(module, 'expert_bias'):
+            # Only update expert_bias if this module is in the training mode. There are special
+            # cases where only the student is in training mode but the teacher is in eval mode
+            # when using online knoweldge-distillation with Model-Optimizer. In this case, we want
+            # to avoid updating teacher's expert_bias.
+            if hasattr(module, 'expert_bias') and module.training:
```
AAnoosheh, line 306: I feel like we addressed something like this several months ago?

ChenhanYu: In a different manner? Or you mean the change to this line should have been ported already?

megatron/training/tokenizer/sft_tokenizer.py
ChenhanYu: fixed

megatron/core/distributed/finalize_model_grads.py
AAnoosheh: Not sure but I remember this conversation somewhere. Maybe with @sharathts ?

megatron/training/tokenizer/sft_tokenizer.py
```diff
@@ -118,7 +140,7 @@ def tokenize_conversation(
             turn_len = len(turn_tokens)
 
             role = turn["role"].lower()
-            if role in ("system", "user"):
+            if role in ("system", "user", "tool"):
```
jenchen13, line 143: is this right? I don't see tool calling tokens being masked out in the nano3 branch

ChenhanYu: The `tool` must be taken care here. We either treat it the same as `user` content (masked out loss) or `assistant` content (loss not masked out). I am surprised that you didn't see an error in the internal nano3 branch if the data has role `tool`.

megatron/core/distributed/finalize_model_grads.py
sharathts: We decided to use `if hasattr(module, 'expert_bias') and module.training:`

PR #3258, opened 2026-02-04
fix: None LR after restoring from ckpt

ko3n1g:
# What does this PR do ?

After loading from checkpoint, LR looks good:

Screenshot 2026-02-08 at 16 27 51

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -2956,18 +2956,11 @@ def train(
         if args.log_params_norm:
             params_norm = calc_params_l2_norm(model)
         learning_rate = None
-        # Find the learning rate from the default config param_group for logging.
-        # Note: In distributed training with pipeline parallelism, some ranks may only have
-        # param_groups with overrides (e.g., embedding params with decoupled_lr). We still
-        # want to log the base learning rate, so we look for default_config=True even if
-        # that param_group has no params on this rank (the 'lr' field is still valid).
         for param_group in optimizer.param_groups:
             if param_group.get('default_config', False):
                 learning_rate = param_group['lr']
                 break
 
-        # Fallback: if no default_config exists (e.g., old checkpoints or unusual configs),
-        # use the first non-empty param_group's learning rate.
         if learning_rate is None:
             for param_group in optimizer.param_groups:
                 if len(param_group['params']) > 0:


megatron/training/training.py
```diff
@@ -2949,8 +2949,16 @@ def get_e2e_base_metrics():
         for param_group in optimizer.param_groups:
             if len(param_group['params']) == 0:
                 continue
+            print(f"DEFAULT CONFIG VALUE: {param_group['default_config']}")
```
dimapihtar:
```suggestion
            logger.info(f"DEFAULT CONFIG VALUE: {param_group['default_config']}")
```

dimapihtar:
let's see if @BoxiangW also approves this change.

megatron/training/training.py
```diff
@@ -2957,10 +2957,16 @@ def get_e2e_base_metrics():
...
-            if param_group['default_config']:
+            if param_group.get('default_config', False):
                 learning_rate = param_group['lr']
+                break
+
+        if learning_rate is None:
+            for param_group in optimizer.param_groups:
+                if len(param_group['params']) > 0:
+                    learning_rate = param_group['lr']
+                    break
```
jstjohn, line 2968: This block would happen under two situations as far as I can tell. 

1. the learning rate was never updated in a param group prior to saving. In this case I suspect it would stay None. I wonder if it is safe to ask for the "lr" without a .get to support this case? For example if you call this line on a checkpoint saved before the first scheduler.step, then the LR would not be set yet unless the user overrides the LR manually through an override: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/optimizer_param_scheduler.py#L272-L286
2. the user has overridden the min/max LR for all parameters from the default in the optimizer config. Given this we can't really know which LR is the canonical one... Maybe we could pick the "lr" that is set on most parameters in that case? Generally the canonical LR is determined by looking at whether the user overrode an lr related parameter (typically min/max LR) on a particular param group: https://github.com/NVIDIA/Megatron-LM/blob/c072f894da962453bc91a6df62a36073a1422e87/megatron/core/optimizer/__init__.py#L204 https://github.com/NVIDIA/Megatron-LM/blob/c072f894da962453bc91a6df62a36073a1422e87/megatron/core/optimizer/__init__.py#L223

megatron/training/training.py
```diff
@@ -2957,10 +2957,16 @@ def get_e2e_base_metrics():
             params_norm = calc_params_l2_norm(model)
         learning_rate = None
         for param_group in optimizer.param_groups:
-            if len(param_group['params']) == 0:
-                continue
-            if param_group['default_config']:
+            if param_group.get('default_config', False):
```
jstjohn, line 2960: See my other comment about the case where the scheduler.step has not been called prior to the checkpoint being saved. In that case the lr would never have been set, so this could be a missing key.

megatron/training/training.py
jstjohn: This block makes me nervous. Please add a comment about why this case happens. 

Basically the goal here is to determine the canonical learning rate from the set of param groups. Typically this is done by looking at the param group with an LR schedule that has not been tweaked by the user vs the default config, and is found by looking for the `default_config=True` flag. If that is not the case then we need to pick one of the LRs and call it the canonical one. In that case the current rule in this implementation above is "pick one at random, whichever is first in the list".

PR #3265, opened 2026-02-05
Refactor Mamba chunked prefill

santhnm2:
# What does this PR do ?
Refactors Mamba chunked prefill to treat the _first_ prefill request as the chunked prefill request rather than the _last_ prefill request. Although the context ensures that the last request is always the designated chunked prefill request, what we actually care about is ensuring that any prefill request with non-zero initial states is executed through the chunked prefill path. There are three scenarios:

Scenario A: No prefill request has initial states yet but the last request is the designated chunked prefill request
Scenario B: There is exactly 1 prefill request which is a continuing chunked prefill request with non-zero initial states
Scenario C: There is a leftover chunked prefill request that is executing its last chunk, followed by additional prefill requests

In scenarios B and C the first prefill request has the relevant initial states. The implementation generalizes to scenario A as well where the first prefill request will be treated as if it has non-zero initial states but this is harmless since the initial states are 0.

This refactor allows us to not have to terminate scheduling early to maintain the invariant that the last request is the relevant chunked prefill request (i.e., the request with the non-zero initial states).

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/models/common/model_chunk_schedule_plan.py
+++ b/megatron/core/models/common/model_chunk_schedule_plan.py
@@ -122,14 +122,13 @@ class TransformerLayerSchedulePlan:
 
         # get flags for latter use
         is_mtp = isinstance(self.layer, MultiTokenPredictionLayer)
-        is_moe = (
-            isinstance(self.layer.mtp_model_layer.mlp, MoELayer)
-            if is_mtp
-            else isinstance(self.layer.mlp, MoELayer)
-        )
+        transformer_layer = self.layer.mtp_model_layer if is_mtp else self.layer
+        is_moe = isinstance(transformer_layer.mlp, MoELayer)
+        num_local_experts = transformer_layer.mlp.num_local_experts if is_moe else None
 
         extra_args["config"] = self.layer.config
         extra_args["is_moe"] = is_moe
+        extra_args["num_local_experts"] = num_local_experts
         extra_args["delay_wgrad_compute"] = self.layer.config.delay_wgrad_compute
         extra_args["is_mtp"] = is_mtp
 
--- a/megatron/core/models/gpt/fine_grained_callables.py
+++ b/megatron/core/models/gpt/fine_grained_callables.py
@@ -44,13 +44,14 @@ def weak_method(method):
 
 
 @internal_api
-def should_free_input(name, is_moe, config):
+def should_free_input(name, is_moe, config, num_local_experts):
     """Determine if the node should free its input memory.
 
     Args:
         name: Node name
         is_moe: Whether it's a MoE model
         config: TransformerConfig object
+        num_local_experts: Number of local experts in MoE module
 
     Returns:
         bool: Whether to free input memory
@@ -71,8 +72,19 @@ def should_free_input(name, is_moe, config):
     # when and how to free the input memory.
     # The input and output of A2A are not needed anymore after the forward pass,
     # so we can free the input memory after the forward pass.
+
+    # When low precision fp8/4 is enabled, the casted tensors are saved and the
+    # original bf16 tensors are safe to be freed.
+    free_mlp = config.fp8 is not None or config.fp4 is not None
+    if not free_mlp:
+        # AlltoAll dispatcher with local_num_experts=1 and HybridEP both use identity
+        # operation for `dispatch_postprocess`, hence the mlp inputs will be directly
+        # passed to GroupedGemm and should be saved for backward pass.
+        free_mlp = num_local_experts > 1 or config.moe_token_dispatcher_type != "alltoall"
+        free_mlp = free_mlp and not enable_hybridep
+
     free_input_nodes = {
-        "mlp": not enable_hybridep,
+        "mlp": free_mlp,
         "moe_combine": True,
         # For non-DeepEP and non-HybridEP dispatcher mode, the input is the un-dispatched tokens
         # and probs before dispatch A2A and it's not needed anymore after the forward pass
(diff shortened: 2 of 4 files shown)


tests/unit_tests/inference/contexts/attention_metadata/test_mamba_metadata.py
```diff
@@ -0,0 +1,428 @@
...
+        )
+        expected_seq_idx[:, :10] = 0
+        assert torch.equal(metadata_context.seq_idx, expected_seq_idx)
+
+    # -------------------------------------------------------------------------
+    # Scenario 4: Chunked Prefill
+    # -------------------------------------------------------------------------
+
+    @pytest.mark.internal
+    def _test_update_chunked_prefill_mixed_exact(self, metadata_context):
```
mathemakitten: Do you mean for these to not get picked up by pytest? I know they're marked internal but tests beginning with _ will just never be run.

megatron/core/ssm/mamba_mixer.py
```diff
@@ -587,12 +512,81 @@ def _dynamic_inference(self, hidden_states: torch.Tensor, context: DynamicInfere
...
+
+        # Merge chunked prefill and regular prefill results
+        if y_chunked is not None and y_regular is not None:
+            y_combined = torch.empty_like(y_regular)
+            tensor_merge(
+                y_chunked, y_regular, metadata.device_chunked_prefill, output_tensor=y_combined
+            )
+            return y_combined
+        elif y_chunked is not None:
+            y_prefill = torch.empty(
```
mathemakitten, line 580: Do we face any similar problem here as with the FP8 case where the `.empty` initialization (as opposed to .zeroes) causes unexpected behaviour, or is it fully masked out? I figure the latter but thought I'd check.

megatron/core/inference/contexts/attention_context/mamba_metadata.py
```diff
@@ -109,20 +112,58 @@ def update(
...
+        # There are three scenarios:
+        #
+        # Scenario A: No prefill request has initial states yet, but the last request
+        #             is the designated chunked prefill request (starting a new chunk).
+        #
+        #   [ ... Decode Requests ... ] [ Prefill (start) ]
+        #                               ^
+        #                               |--- First prefill request
+        #                                    Treated as having states.
+        #                                    Harmless because actual initial states are 0.
```
mathemakitten, line 147: Strictly speaking, this is a minor inefficiency, right? Every batch with any prefill requests will get run through chunked prefill path even though there is no actual chunking? I don't think we need to change anything just noting it.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -452,6 +452,7 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
         self.graph_attn_metadata = {}
         self.non_graph_attn_metadata = {}
         self.active_attn_metadata = None
+        self.is_recording_cuda_graphs = False
```
mathemakitten: "Recording" might be overloaded here, since it refers to a specific recording function in cuda_graphs.py (i.e. we record a graph then create the graph later). Inference cudagraphs, strictly speaking, don't get a separate "recording" phase — the graphs are created as they are encountered, so mixing the terminology is a little confusing. I think what is being referred to as recording here is actually graph creation.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1133,15 +1127,14 @@ def schedule_chunked_prefill(self):
                     # We break early for Mamba models running a final prefill chunk
```
mathemakitten: Update this comment?

tests/unit_tests/inference/contexts/attention_metadata/test_mamba_metadata.py
santhnm2: Good point, will update. Thanks!

megatron/core/ssm/mamba_mixer.py
santhnm2: Technically the kernel is supposed to write out 0s for the -1 values but I think I'll add the explicit 0ing out that we do for the embedding / attention layers.

megatron/core/inference/contexts/attention_context/mamba_metadata.py
santhnm2: Yea it is an inefficiency for any batch with no continuing prefill requests. Presumably that would only be the case in regimes where we don't have frequently arriving requests though so it should be rare (evals and RL will not hit this case except for the very first batch since all requests arrive up front).

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Maybe rename to `is_capturing_cuda_graphs`? Or `is_creating_cuda_graphs`?

megatron/core/inference/engines/dynamic_engine.py
santhnm2: Thanks, good catch

tests/unit_tests/inference/contexts/attention_metadata/test_mamba_metadata.py
santhnm2: I think I had disabled them for debugging purposes at some point but then forgot to re-enable them

megatron/core/inference/contexts/attention_context/mamba_metadata.py
santhnm2: I updated the comment to mention this inefficiency

megatron/core/ssm/mamba_mixer.py
santhnm2: Will do this in a follow-up PR along with some other quantization changes

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Renamed to `is_creating_cuda_graphs` for now

ko3n1g:

You can track the progress here: https://github.com/NVIDIA/Megatron-LM/actions/runs/21943803326

PR #3282, opened 2026-02-06
[dev] moe(feat): support chunkwise context parallelism for GDN.

yuzhongw-nvidia:
# What does this PR do ?

PR for main: https://github.com/NVIDIA/Megatron-LM/pull/6439

FLA gives another CP implementation for GDN https://github.com/fla-org/flash-linear-attention/pull/691, which is claimed to outperform the old A2A-style CP. Integrate the new one into MCore.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


tests/unit_tests/ssm/test_gated_delta_net.py
```diff
@@ -383,29 +448,34 @@ def test_parallel_gated_delta_net_correctness(tmp_path_dist_ckpt, sequence_packi
...
         atol=atol,
         rtol=rtol,
+        cosine_similarity_threshold=cosine_similarity_threshold,
         tp=tp,
         sp=sp,
         cp=cp,
         seed=123,
         sequence_length=256,
-        micro_batch_size=4,
+        micro_batch_size=1 if (cp_comm_type == "all_gather" and cp > 1) else 4,
```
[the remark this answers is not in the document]
yuzhongw-nvidia: Disable the path `if cp_size_all_gather > 1 and packed_seq_params is None and batch > 1`.

yuzhongw-nvidia:
/ok to test https://github.com/NVIDIA/Megatron-LM/commit/6315e1578d1bfb26b625b01fc3e72e50b6c5deb2

megatron/core/context_parallel_layout.py
```diff
@@ -0,0 +1,297 @@
...
+    pos_in_seq = positions - global_starts[seq_idx]
+    chunk_lens = (seq_lens // chunk_divisor)[seq_idx]
+    chunk = pos_in_seq // chunk_lens
+    offset = pos_in_seq - chunk * chunk_lens
+
+    if layout == "zigzag":
+        owner = torch.where(chunk < cp_size, chunk, 2 * cp_size - chunk - 1)
+        local_slot = torch.where(chunk < cp_size, torch.zeros_like(chunk), torch.ones_like(chunk))
+    else:
+        owner = chunk // 2
```
Victarry: **[CRITICAL Correctness]** Make the THD target layout match FLA's flattened rank-contiguous partition.

The root cause is that two different meanings of “contiguous CP layout” are being conflated. For SBHD with one logical sequence, assigning rank `r` chunks `[2r, 2r+1]` is equivalent to assigning it one contiguous interval of the global token stream. This implementation extends that rule independently to every packed sequence. FLA does not use that per-sequence partition: `build_cp_context()` computes `part_len = total_tokens // world_size` from the entire flattened THD buffer and assumes rank `r` physically owns `[r * part_len, (r + 1) * part_len)`. These layouts agree for a single sequence, which is why the current chunkwise packed test with `micro_batch_size=1` does not expose the bug, but they diverge as soon as the pack contains multiple sequences.

For example, with `cu=[0,16,40]` and `CP=2`, this code assigns rank 0 `[0:8] + [16:28]`, so its local rows contain 8 tokens from sequence 0 followed by 12 from sequence 1. FLA instead assumes rank 0 owns flat range `[0:20]` and derives `local_cu_seqlens=[0,16,20]`, interpreting the first 16 local rows as sequence 0. Consequently both causal conv and GDR consume incorrect local sequence boundaries. The state communication is also structurally incompatible: FLA only carries state for the first/last local sequence that crosses a flat rank boundary, whereas this per-sequence layout makes every packed sequence cross every rank.

**Suggested change:** keep the existing per-sequence mapping only for `layout == "zigzag"`, but define the THD `"contiguous"` indices directly from FLA's flat partition, e.g.:

```python
part_len = total_tokens // cp_size
rank_start = cp_rank * part_len
return positions[rank_start : rank_start + part_len]
```

This makes both `zigzag -> contiguous` and its inverse compatible with `build_cp_context()` without changing the A2A routing structure. Please update the current expected indices (`rank0 == range(0, 20)`, `rank1 == range(20, 40)` for this example) and add a distributed round-trip/GDN parity test with at least two unequal-length packed sequences; a single-sequence test cannot distinguish the two layouts.

megatron/core/ssm/gated_delta_net.py
```diff
@@ -45,6 +49,7 @@
 try:
     from fla.modules.convolution import causal_conv1d
     from fla.modules.l2norm import l2norm
+    from fla.ops.cp import build_cp_context
```
Victarry, line 52: **[IMPORTANT Compatibility]** Raise the FLA dependency floor for `fla.ops.cp`. The current `flash-linear-attention~=0.4.0` constraint permits 0.4.0 and 0.4.1, but neither version contains `fla.ops.cp`; the API used here appears in 0.4.2. Because this import shares the top-level `try`, those supported versions set `HAVE_FLA=False` and disable even non-chunkwise GDN. Please require `flash-linear-attention>=0.4.2,<0.5` and update the lock metadata.

megatron/core/transformer/transformer_config.py
```diff
@@ -1500,6 +1525,18 @@ def __post_init__(self):
                 f"{self.linear_num_value_heads=} must be a multiple of "
                 f"({self.tensor_model_parallel_size=} * {self.context_parallel_size=})."
             )
+            if self.context_parallel_size > 1:
+                assert self.linear_cp_mode in ("headwise", "chunkwise"), (
```
Victarry, line 1519: **[IMPORTANT Correctness]** Do not apply the headwise `TP * CP` divisibility requirement to chunkwise mode. The unconditional assertions immediately above this block still require both head counts to be divisible by `tensor_model_parallel_size * context_parallel_size`. That contradicts the new GDN initialization logic, where chunkwise CP keeps all TP-local heads and intentionally skips CP head splitting. Please require divisibility by TP only for chunkwise mode, and by `TP * CP` for headwise mode.

megatron/core/context_parallel_layout.py
```diff
@@ -0,0 +1,297 @@
...
+    if cp_size == 1:
+        return x
+    cp_rank = cp_group.rank()
+
+    if seq_dim != 0:
+        x = x.movedim(seq_dim, 0)
+    x = x.contiguous()
+
+    cu = cu_seqlens.to(device=x.device, dtype=torch.long)
+    source_by_rank = [
```
Victarry, line 145: **[IMPORTANT Performance]** Precompute and reuse THD routing metadata instead of applying `lru_cache` directly to the Tensor-keyed helper. Each swap calls this function for both layouts and every CP rank, so one GDN layer rebuilds `arange`/`bucketize`/`argsort` metadata roughly `4 * CP` times across its two conversions, with additional host synchronizations from `.item()` and Tensor-to-bool checks. A normal `lru_cache` is unsafe because Tensor keys are identity-based and may be mutated in static CUDA-graph buffers, and the cache retains CUDA tensors. Please build a scoped/versioned routing context once per immutable packed-sequence metadata and reuse it for both directions and across GDN layers.

megatron/core/ssm/gated_delta_net.py
```diff
@@ -417,14 +509,35 @@ def _forward_compute(
...
 
+        # Chunkwise CP expects the contiguous-time chunk layout (rank r holds chunks
+        # [2r, 2r+1]) inside conv1d / chunk_gated_delta_rule. Megatron attention CP
+        # feeds us the zigzag attention-load-balanced layout (rank r holds
+        # [r, 2*cp-r-1]), so reshuffle chunks over the CP group with a single
+        # all-to-all — no full-sequence gather required.
+        if cp_size_chunkwise > 1:
+            nvtx_range_push(suffix="zigzag_to_contiguous")
+            if packed_seq_params is not None and packed_seq_params.qkv_format == 'thd':
+                qkvzba = zigzag_to_contiguous_chunks(
```
Victarry, line 522: **[SUGGESTION Performance]** Consider moving the chunkwise layout conversions outside the input and output projections. Sequence-layout permutation commutes with tokenwise linear projections. For the TP=1 Qwen3.5 configurations, the current two A2As operate on summed per-token widths of 16,448 elements for 35B and 28,800 for 397B, whereas converting `hidden_states` before `in_proj` and after `out_proj` would operate on `2 * hidden_size`: 4,096 and 8,192 elements respectively. This reduces payload by approximately 4.02x and 3.52x. The implementation still needs to account for TP sequence-parallel all-gather/reduce-scatter changing the sequence distribution around these projections.

megatron/core/ssm/gated_delta_net.py
```diff
@@ -463,11 +598,26 @@ def _forward_compute(
...
+        norm_out = norm_out.reshape(batch, seq_len_post_headwise, -1)
         norm_out = norm_out.transpose(0, 1).contiguous()
 
+        # Inverse of the zigzag -> contiguous reshuffle performed before conv1d.
+        # Restores the Megatron attention-load-balanced layout that downstream
+        # layers and loss computation expect.
+        if cp_size_chunkwise > 1:
+            nvtx_range_push(suffix="contiguous_to_zigzag")
+            if packed_seq_params is not None and packed_seq_params.qkv_format == 'thd':
+                norm_out = contiguous_to_zigzag_chunks(
```
Victarry, line 614: **[IMPORTANT Performance]** Avoid restoring and re-entering the layout for every GDN in a consecutive linear-attention region. With `linear_attention_freq=4`, Qwen3.5 uses three consecutive GDN layers followed by SDPA, so the current module-local contract performs six layout A2As per region: one entry and one exit for each GDN. Normalization, residual/BDA, and dense MLP operations are tokenwise, so the block could enter the contiguous layout once before the first GDN and restore zigzag only before the following SDPA or another canonical-layout boundary. Simply removing this inverse is not sufficient because the residual currently remains zigzag; layout ownership and relevant per-token metadata need to be handled at the transformer-block/region boundary.

tests/unit_tests/transformer/test_attention.py
```diff
@@ -862,27 +863,37 @@ def get_tensor_on_this_rank(tensor):
...
+                torch.testing.assert_close(
+                    baseline,
+                    parallel,
+                    atol=atol,
+                    rtol=rtol,
+                    msg=lambda msg: f"Mismatch in {tensor_name}: {msg}",
+                )
+                return
+            except AssertionError as close_error:
+                if cosine_similarity_threshold is None:
```
Victarry, line 878: **[IMPORTANT Numerical Testing]** Do not use cosine similarity as the sole fallback after `assert_close` fails. Cosine similarity is insensitive to scale, so even `parallel = 100 * baseline` passes with a score of 1.0. This can hide material output and gradient errors in the new CP path. Please require an additional relative-L2/norm-ratio bound, or use explicitly relaxed elementwise tolerances together with cosine similarity.

megatron/core/context_parallel_layout.py
```diff
@@ -0,0 +1,297 @@
...
+    if cu_seqlens.dim() != 1:
+        raise ValueError(f"cu_seqlens must be 1-D, got shape {tuple(cu_seqlens.shape)}.")
+
+    cu = cu_seqlens.to(dtype=torch.long)
+    if cu.numel() == 0 or cu[0].item() != 0:
+        raise ValueError(f"cu_seqlens must start at 0, got {cu_seqlens}.")
+
+    seq_lens = torch.diff(cu)
+    chunk_divisor = 2 * cp_size
+    if torch.any(seq_lens <= 0):
```
xiaoyao0115: **[IMPORTANT Compatibility] Allow zero-length placeholder entries in statically padded THD `cu_seqlens`.**

In static THD CUDA Graph mode, Megatron pads `cu_seqlens` to `thd_max_packed_sequences + 1` entries by repeating the final offset. For example, a valid `[0, 16, 40]` may become `[0, 16, 40, 40, 40]`. The repeated trailing entries are placeholders used to keep the metadata shape static, and therefore produce zero-length segments.

The current `seq_lens <= 0` check rejects this valid representation, causing chunkwise CP to fail whenever the actual number of packed sequences is smaller than the configured maximum.

Please reject only decreasing boundaries, and ignore zero-length placeholder segments when validating divisibility and computing ownership, or otherwise normalize repeated boundaries without changing the static input contract. Please also add a test with trailing duplicate offsets.

megatron/core/ssm/gated_delta_net.py
yuzhongw-nvidia: Very good suggestion, but we plan to make this change in the upcoming refactor works. Will add a comment for this.

megatron/core/ssm/gated_delta_net.py
yuzhongw-nvidia: Similar to https://github.com/NVIDIA/Megatron-LM/pull/3282#discussion_r3490324819, I will leave a comment to explain.

megatron/core/context_parallel_layout.py
yuzhongw-nvidia: Good point. The routing is a function of cu_seqlens, so it is only reusable within the same packed microbatch. Since cu_seqlens changes across batches, a persistent cache would be unsafe unless keyed by the actual sequence-boundary contents, which is likely not worth the synchronization/complexity. I added a TODO to keep this as future work for a per-microbatch CP layout context instead of caching it here.

tests/unit_tests/transformer/test_attention.py
yuzhongw-nvidia: Very good point. Resolved.

megatron/core/context_parallel_layout.py
yuzhongw-nvidia: Thanks. Resolved.

megatron/core/transformer/transformer_config.py
yuzhongw-nvidia: Very good catch. Resolved.

megatron/core/ssm/gated_delta_net.py
yuzhongw-nvidia: Thanks. Resolved.

megatron/core/context_parallel_layout.py
yuzhongw-nvidia: Very good catch. Helps a lot! Resolved.

yuzhongw-nvidia:
/ok to test https://github.com/NVIDIA/Megatron-LM/commit/644098c7228b2c72187108f3a2a20394bf19df26

yuzhongw-nvidia:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3282/commits/df71688f58c82eeba6d6dbcdd723360b69a09282

Victarry (approved):
LGTM!Thanks for the great work~

yuzhongw-nvidia:
/ok to test https://github.com/NVIDIA/Megatron-LM/commit/6aaf4b78d6938feb835cc91f954c533a9f6cd2aa

yuzhongw-nvidia:
/ok to test https://github.com/NVIDIA/Megatron-LM/commit/0beec47dd0ebf7135be07ee1ece8157fb8eae308

xiaoyao0115 (approved):
lgtm！

ko3n1g (approved):

PR #1520, opened 2025-04-04
added fix to avoid overflow with new numpy casting behaviour (Issue: #1519)

Apsod:
Fix for issue #1519

Issue #1519, opened 2025-04-04
[BUG] Overflow in _sequence_pointers calculcation with numpy >= 2.0


Apsod:
**Describe the bug**
Changes to numpy (>= 2.0) data type promotion introduces an overflow in _sequence_pointers, leading to negative values in the merge_datasets script. 

**To Reproduce**
```
import numpy
from typing import *

def _sequence_pointers(sequence_lengths: List[int]) -> List[int]:
    """Build the sequence pointers per the sequence lengths and dtype size

    Args:
        sequence_lengths (List[int]): The length of each sequence

    Returns:
        List[int]: The pointer to the beginning of each sequence
    """
    itemsize = 4
    curr_ptr = 0
    list_ptr = []
    for length in sequence_lengths:
        list_ptr.append(curr_ptr)
        curr_ptr += length * itemsize
    return list_ptr

l1 = numpy.array([(1<<31)-100, 123], dtype=numpy.int32)
sl = []
sl.extend(l1)

print(_sequence_pointers(sl))
# outputs [0, np.int32(-400)]
```

**Expected behavior**
The expected output is to not overflow: 

`[0, 8589934192]`

**Environment (please complete the following information):**
 - Megatron-LM commit ID: 1730f6a0dde15a086808ec9f3ba9ac35a3225716

**Proposed fix**
Casting the np.int32 to a python int in the ptrs-calculation solves the issue: 

```
    for length in sequence_lengths:
        list_ptr.append(curr_ptr)
        curr_ptr += int(length) * itemsize
    return list_ptr
```

Apsod:
Commit 84e9c3a solved this issue. So I'm closing it.

PR #3306, opened 2026-02-08
fix: numpy overflow

ko3n1g:
# What does this PR do ?

Issue https://github.com/NVIDIA/Megatron-LM/issues/1519 and PR https://github.com/NVIDIA/Megatron-LM/pull/1520 already observed this.

During long convergence training, we ran into a zero byte during step ~31000.

I hot-fixed it, and @asolergi-nv later suggested the proper fix to avoid the overflow.

Presumably, this started after bumping numpy to ge 2.0.

We didnt notice this in the weekly runs as those stop after ~10000 steps. 

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/datasets/gpt_dataset.py
+++ b/megatron/core/datasets/gpt_dataset.py
@@ -361,7 +361,7 @@ class GPTDataset(MegatronDataset):
                     else doc_index_end_offset + self.config.add_extra_token_to_sequence
                 )
                 sample_parts.append(
-                    self.dataset.get(self.document_index[i], offset=offset, length=length)
+                    self.dataset.get(self.document_index[i], offset=int(offset), length=length)
                 )
         assert len(document_ids) == len(
             sample_parts


megatron/core/datasets/indexed_dataset.py
```diff
@@ -862,7 +862,8 @@ def get(
         sequence_pointer, sequence_length, sequence_mode = self.index[idx]
         if length is None:
             length = sequence_length - offset
-        sequence_pointer += offset * DType.size(self.index.dtype)
+        # Convert to Python int to avoid numpy int64 overflow
```
jkamalu: We're running into int64 overflow? As in, we're going outside the range of int64? If that's the case, something else is going wrong.

Can you provide some inputs and outputs that cause this to fail?

Some of these casts should be unnecessary, e.g. offset and the return from DType.size are both typed for int already. So either something is going wrong and they're not int, or the casts are unnecessary and sequence_pointer is the issue, but sequence_pointer should be numpy.int64

asolergi-nv: > offset and the return from DType.size **are both typed for int already**

Here's the problem: the type hint is misleading us. `offset` isn't an `int` but a `numpy` array [[1](https://github.com/NVIDIA/Megatron-LM/blob/30695913bdaef9deaafcf1c7ac617fc37f878811/megatron/core/datasets/gpt_dataset.py#L328-L329), [2](https://github.com/NVIDIA/Megatron-LM/blob/30695913bdaef9deaafcf1c7ac617fc37f878811/megatron/core/datasets/gpt_dataset.py#L592)]. We should cast this variable either here and change the typing or in [`_query_document_sample_shuffle_indices`](https://github.com/NVIDIA/Megatron-LM/blob/30695913bdaef9deaafcf1c7ac617fc37f878811/megatron/core/datasets/gpt_dataset.py#L301) just before calling this method. I prefer the latter.

[`sequence_pointer`](https://github.com/NVIDIA/Megatron-LM/blob/30695913bdaef9deaafcf1c7ac617fc37f878811/megatron/core/datasets/indexed_dataset.py#L293)'s are `numpy.int64` which shouldn't overflow and `DType.size` is an int value.

jkamalu: Agreed, I'd rather cast before the call than change the typing to be a numpy array and cast inside.

jaredcasper (approved):

PR #3360, opened 2026-02-11
Switch oncall

janEbert:
With @Phlip79. Thank you!


janEbert:
Automatic oncall rotation happened in the meantime causing a conflict, so I had to rebase. :/

PR #3364, opened 2026-02-11
Optimized local_multi_tensor_l2_norm and fixed double-counting bug

AkCodes23:
# What does this PR do ?
Optimizes local_multi_tensor_l2_norm by removing redundant norm computation and fixes a double-counting bug that caused incorrect results when _foreach_norm was unavailable.

Fixes a correctness bug where norms were computed twice when _foreach_norm was unavailable, resulting in sqrt(2) * true_norm.

Removes redundant tensor iteration and unnecessary all_tensors list construction.

Ensures _foreach_norm is applied only when dtype == torch.float32 to prevent potential FP16 overflow.

Adds unit test tests/unit_tests/test_local_multi_tensor_l2_norm_simple.py validating results against manual L2 norm computation.

### Pre-checks


Phlip79:
Please clean up PR before marking it as ready for review.

PR #3373, opened 2026-02-11
RL: Hybrid MoE training cudagraphs and fix training <-> inference transition

mathemakitten:
# What does this PR do ?
Hybrid MoE models require cudagraphs during training for usable performance but the present workflow has several bugs preventing a clean transition between training and inference:
* The existing hybrid sequence packing code is not cudagraphable because it has data-dependent shapes. We move the creation of seq_idx out of the cudagraphed forward pass, and instead pass in a static value ahead of time for each batch.
* The RL training loop alternates between inference (rollouts) and training (policy gradient update). MoE layers need full-layer graph capture during inference (with drop_and_pad=True so D2H copy is skipped) and partial capture       
during training (router + postprocess captured, expert dispatch with D2H runs eagerly).

We enable a cleaner transition between inference and training in RL by fixing the existing bugs in the workflow:
* Since inference runs before training, currently `create_mcore_cudagraph_manager` only runs at __init__ (with full-layer scope), so MoE layers permanently keep a full-layer cudagraph_manager and `use_partial_cudagraphs=False`.
* Similarly, because inference runs before training, its state leaks into training graph capture. If we do not carefully track cudagraph manager state, we can incorrectly toggle `drop_and_pad` during training depending on the last value of `drop_and_pad` during inference (set to True or False depending on whether the last inference step uses cudagraphs).                        
* We also make the `moe_router` and `moe_preprocess` scopes cudagraphable, which they previously were not.

Changes needed to cleanly transition between inference / training graphs:
* During training, `transition_moe_to_partial_cudagraphs()` iterates over all MoeTransformerLayer modules and sets use_partial_cudagraphs = True. Then it creates `cudagraph_manager_router` and `cudagraph_manager_postprocess` once upfront (we cache the created cudagraph managers and do not re-create the graphs between phases).
* During inference, `transition_moe_to_full_cudagraphs()` iterates over all MoeTransformerLayer modules and sets use_partial_cudagraphs = False, then asserts that the layerwise cudagraph_manager exists. As before, we never teardown graphs when swapping between phases.
* After inference, we need to reset `drop_and_pad` to avoid leaking inference settings into training capture by setting `set_decode_expert_padding(..., set_to=False)`.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/tools/trigger_internal_ci.md
+++ b/tools/trigger_internal_ci.md
@@ -17,18 +17,15 @@ To check existing remotes: `git remote -v`
 
 **The name of the origin will be required later!**
 
-**2. Obtain a pipeline trigger token:**
+**2. Obtain a personal access token:**
 
-1. Open the internal GitLab project in your browser.
-2. Go to **Settings → CI/CD → Pipeline trigger tokens**.
-3. Click **Add new token**, give it a description, and click **Create**.
-4. Copy the generated token (starts with `glptt-`).
-5. Store it in your environment to avoid passing it on every invocation:
-
-Reach out to @mcore-ci in case you don't have access to the settings page.
+1. Open your internal GitLab profile: **User menu → Edit profile → Access tokens**.
+2. Click **Add new token**, give it a description, set an expiry, and select the **`api`** scope.
+3. Click **Create personal access token** and copy the generated token (starts with `glpat-`).
+4. Store it in your environment to avoid passing it on every invocation:
 
 ```bash
-export GITLAB_TRIGGER_TOKEN=glptt-<your-token>
+export GITLAB_TOKEN=glpat-<your-token>
 ```
 
 **Tip: Store this in your .env or .bashrc file**
@@ -38,7 +35,7 @@ export GITLAB_TRIGGER_TOKEN=glptt-<your-token>
 ```bash
 python tools/trigger_internal_ci.py \
   --gitlab-origin gitlab \
-  [--trigger-token glptt-<your-token>] \
+  [--access-token glpat-<your-token>] \
   [--functional-test-scope mr] \
   [--functional-test-repeat 5] \
   [--functional-test-cases all] \
@@ -48,7 +45,7 @@ python tools/trigger_internal_ci.py \
 | Argument | Default | Description |
 |---|---|---|
 | `--gitlab-origin` | *(required)* | Git remote name for the internal GitLab |
-| `--trigger-token` | `$GITLAB_TRIGGER_TOKEN` | Pipeline trigger token |
+| `--access-token` | `$GITLAB_TOKEN` | Personal access token with `api` scope |
 | `--functional-test-scope` | `mr` | `FUNCTIONAL_TEST_SCOPE` pipeline variable |
 | `--functional-test-repeat` | `5` | `FUNCTIONAL_TEST_REPEAT` pipeline variable |
 | `--functional-test-cases` | `all` | `FUNCTIONAL_TEST_CASES` pipeline variable |
--- a/tools/trigger_internal_ci.py
+++ b/tools/trigger_internal_ci.py
@@ -18,6 +18,9 @@
 Pushes the current branch to the internal GitLab remote under the
 pull-request/<branch> naming convention and triggers a pipeline with
 the specified test configuration.
+
+Requires a GitLab personal access token with at least the 'api' scope.
+Set the GITLAB_TOKEN environment variable or pass --access-token.
 """
 
 import argparse
@@ -82,7 +85,7 @@ def git_push(origin, target_branch, dry_run=False):
     )
 
 
-def trigger_pipeline(gitlab_url, trigger_token, ref, pipeline_vars, dry_run=False):
+def trigger_pipeline(gitlab_url, access_token, ref, pipeline_vars, dry_run=False):
     """Trigger a GitLab pipeline on the given ref with the provided variables."""
     if dry_run:
         logger.info(
@@ -95,9 +98,10 @@ def trigger_pipeline(gitlab_url, trigger_token, ref, pipeline_vars, dry_run=Fals
     logger.info(
         "Triggering pipeline on https://%s project %s @ %s", gitlab_url, GITLAB_PROJECT_ID, ref
     )
-    gl = gitlab.Gitlab(f"https://{gitlab_url}")
+    gl = gitlab.Gitlab(f"https://{gitlab_url}", private_token=access_token)
     project = gl.projects.get(GITLAB_PROJECT_ID, lazy=True)
-    pipeline = project.trigger_pipeline(ref=ref, token=trigger_token, variables=pipeline_vars)
+    variables = [{"key": k, "value": v} for k, v in pipeline_vars.items()]
+    pipeline = project.pipelines.create({"ref": ref, "variables": variables})
     logger.info("Pipeline triggered: %s", pipeline.web_url)
 
 
@@ -112,9 +116,9 @@ def main():
         help="Name of the git remote pointing to the internal GitLab (e.g. gitlab)",
     )
     parser.add_argument(
-        "--trigger-token",
-        default=os.environ.get("GITLAB_TRIGGER_TOKEN"),
-        help="GitLab pipeline trigger token (or set GITLAB_TRIGGER_TOKEN env var)",
+        "--access-token",
+        default=os.environ.get("GITLAB_TOKEN"),
+        help="GitLab personal access token with 'api' scope (or set GITLAB_TOKEN env var)",
     )
     parser.add_argument(
         "--functional-test-scope",
@@ -140,8 +144,8 @@ def main():
     args = parser.parse_args()
     logging.basicConfig(level=logging.INFO, format="%(message)s")
 
-    if not args.trigger_token:
-        logger.error("--trigger-token or GITLAB_TRIGGER_TOKEN not set")
+    if not args.access_token:
+        logger.error("--access-token or GITLAB_TOKEN not set")
         sys.exit(1)
 
     branch = get_current_branch()
@@ -162,7 +166,7 @@ def main():
     }
 
     trigger_pipeline(
-        gitlab_hostname, args.trigger_token, target_branch, pipeline_vars, dry_run=args.dry_run
+        gitlab_hostname, args.access_token, target_branch, pipeline_vars, dry_run=args.dry_run
     )
 
 


megatron/core/transformer/moe/moe_utils.py
```diff
@@ -396,21 +362,23 @@ def permute(
             permuted_probs = probs_T_1D.index_select(0, indices_1D)
     else:
         # mask [num_tokens, num_experts] -> [num_experts, num_tokens]
-        routing_map = routing_map.bool().T.contiguous()
+        routing_map = routing_map.to(dtype=torch.int8).T.contiguous()
```
mathemakitten: booleans aren't orderable, hence this change allowing it to be passed to argsort with `descending=True`

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -327,47 +315,25 @@ def permute(
...
-        if fused_permute_and_pad_with_probs is not None and tokens_per_expert is not None:
-            return fused_permute_and_pad_with_probs(
-                tokens, probs, routing_map, tokens_per_expert, align_size
-            )
-        else:
-            output, permuted_probs, row_id_map = fused_permute_with_probs(
-                tokens, probs, routing_map, num_out_tokens=num_out_tokens
-            )
-            return output, permuted_probs, row_id_map, None, tokens_per_expert
+        return fused_permute_with_probs(tokens, probs, routing_map, num_out_tokens=num_out_tokens)
```
mathemakitten, line 369: `masked_select` is not cudagraphable as it produces variable-shaped outputs. We replace it with an equivalent graph-friendly op.

megatron/core/transformer/moe/moe_utils.py
mathemakitten: bools are not orderable, hence this change

siddharth9820: Just curious - what do your mean by orderable?

pretrain_mamba.py
```diff
@@ -246,6 +259,9 @@ def forward_step(data_iterator, model: MambaModel):
             max_seqlen_q=max_seqlen,
             max_seqlen_kv=max_seqlen,
         )
+        # Override __post_init__'s seq_idx (which used max_seqlen_q) with the
+        # correct value computed from the actual tensor sequence dimension.
+        packed_seq_params.seq_idx = seq_idx
```
tdene: I don't like how this computation is done twice and overriden. Can `PackedSeqParams` instead take a `total_tokens` optional parameter and use that instead of `self.max_seqlen_q` when it is provided? That way you can remove L239 -> L251 of this file and just pass in a not-None `total_tokens`.

megatron/rl/rl_utils.py
```diff
@@ -1763,6 +1803,21 @@ def megatron_rl_inference_mode(
...
+        # Reset drop_and_pad leaked from inference decode
+        set_decode_expert_padding(unwrap_model(model[0]), set_to=False)
+
+        # Change cudagraph scope for training
+        model[0].config.cuda_graph_scope = [
+            CudaGraphScope.mamba,
+            CudaGraphScope.attn,
+            CudaGraphScope.moe_router,
+            CudaGraphScope.moe_preprocess,
+        ]
```
tdene: Is there a way to avoid hardcoding this?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -963,6 +962,8 @@ def clone_ten(ten):
                     if self.is_last_layer:
                         gc.collect()
 
+        _set_warmup_end()
```
mathemakitten: Previously this was getting cleared inbetween layers and causing issues with graph capture and replay.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1128,7 +1129,11 @@ def replace_with_weak_ref(arg):
                 if not torch.is_tensor(arg):
                     return arg
 
-                ref = make_weak_ref(arg)
+                try:
+                    ref = make_weak_ref(arg)
+                except (RuntimeError, TypeError):
```
mathemakitten: TODO: narrow this or debug it

examples/rl/model_configs/nemotron6_3b_moe.sh
```diff
@@ -58,7 +58,6 @@ MODEL_OPTIONS="\
   --rl-skip-bos-token \
   --no-rl-use-sequence-packing \
   --rl-partial-rollouts \
-  --rl-offload-optimizer-during-inference \
```
mathemakitten, line 61: This flag is still supported but is not recommended to have on by default because of the large speed downside.

megatron/core/transformer/transformer_layer.py
```diff
@@ -1249,6 +1249,13 @@ def __init__(self, *args, **kwargs):
 
         super().__init__(*args, **kwargs)
 
+    def _should_call_local_cudagraph(self, *args, **kwargs):
```
mathemakitten, line 1269: This looks odd but it ensures that: 
* When partial cudagraphs are on for MoE training, the full-layer cudagraph manager does not execute because we must defer to the MoE router and preprocessor cudagraph managers to make sure the graph break works.
* When we explicitly disable cudagraphs via `disable_cuda_graphs` and the cuda_graph_impl is set to "none", this ensures it does not execute the local path by default. TODO: allow the [] case?

megatron/core/transformer/utils.py
```diff
@@ -448,6 +449,98 @@ def toggle_cuda_graphs(model, set_to="none", reset_cuda_graphs=True):
         delete_cuda_graphs()
 
 
+@contextmanager
+def disable_cuda_graphs(model, lang_module=None):
```
mathemakitten: TODO: check if this is still needed now that we run the partial cudagraph case correctly

megatron/rl/rl_utils.py
```diff
@@ -1678,9 +1710,17 @@ def megatron_rl_inference_mode(
...
 
+    # Change cudagraph scope for inference (empty list = full-layer capture)
+    model[0].config.cuda_graph_scope = []
+    model[0].config.cuda_graph_impl = "local"
+
     # If we get a lower precision wrapper, we go one object deeper.
     lang_module = model[0].module.module if hasattr(model[0].module, "module") else model[0].module
 
+    # Switch MoE layers to full CUDA graph capture for inference
+    if args.rl_training_cuda_graphs:
```
mathemakitten: TODO: add check for MoE

megatron/rl/rl_utils.py
```diff
@@ -1763,6 +1803,21 @@ def megatron_rl_inference_mode(
         if cuda_graph_impl != "none" and not args.rl_training_cuda_graphs:
             toggle_cuda_graphs(lang_module, 'none', reset_cuda_graphs=reset_cuda_graphs)
 
+        # Reset drop_and_pad leaked from inference decode
+        set_decode_expert_padding(unwrap_model(model[0]), set_to=False)
+
+        # Change cudagraph scope for training
```
mathemakitten: TODO here as well, add check for MoE

megatron/core/ssm/mamba_mixer.py
```diff
@@ -648,7 +648,7 @@ def _ssm_training(
                 _check_mamba_sequence_packing_support(for_inference_not_training=False)
             )
             assert sequence_packing_available, reason_for_no_sequence_packing
-            seq_idx = self._create_packed_seq_idx(packed_seq_params, zxBCdt.shape[1])
```
duncanriach, line 631: Should the function that is no longer being called be removed?

megatron/core/packed_seq_params.py
```diff
@@ -21,3 +22,26 @@ class PackedSeqParams:
...
+    def __post_init__(self):
+        # Pre-compute seq_idx for Mamba mixer CUDA graph compatibility.
+        # Stored as a non-field attribute so dataclasses.fields() won't include it, preventing it
+        # from being forwarded to TE attention.
+        cu_seqlens = (
+            self.cu_seqlens_q_padded if self.cu_seqlens_q_padded is not None else self.cu_seqlens_q
+        )
+        if isinstance(cu_seqlens, Tensor) and self.max_seqlen_q is not None:
+            total_tokens_tensor = torch.tensor(
+                [self.max_seqlen_q], dtype=cu_seqlens.dtype, device=cu_seqlens.device
```
duncanriach: Why is this being calculated with max_seqlen? It seems completely wrong, and then you have to override it? If this is for Mamba mixer, then it should be the total length of the pack

pretrain_mamba.py
duncanriach: I don't understand why the post_init code is calculating seq_idx using max_seqlen_q. That seems totally wrong. Can that code/version be completely removed?

I agree with Teo that it makes sense to encapsulate the creation of seq_idx inside the PackedSeqParams constructor. The way he suggests makes sense. Another option is to add a method such as set_seq_idx(total_tokens). Calling this here could cause seq_idx to be set properly for the Mamba mixer. If that function is not called, then seq_idx would just default to None (set via the post-init?)

pretrain_mamba.py
```diff
@@ -236,6 +236,19 @@ def forward_step(data_iterator, model: MambaModel):
     if cu_seqlens is None:
         packed_seq_params = None
     else:
+        # Pre-compute seq_idx for Mamba mixer CUDA graph compatibility.
+        # total_tokens must be the actual tensor sequence dimension (not max_seqlen,
+        # which is the max *individual* sequence length per TE convention).
```
duncanriach: This comment is so confusing. You're overriding something that doesn't make any sense and then explaining that you're overriding it.

megatron/core/ssm/mamba_mixer.py
duncanriach: I'm concerned about losing this docstring in the reshuffle

duncanriach: I think these example comments are also useful. It seems like this code has been re-written as this has been moved out of the mixer, and this documentation has been lost

pretrain_mamba.py
```diff
@@ -236,6 +236,19 @@ def forward_step(data_iterator, model: MambaModel):
...
+        total_tokens = tokens.size(1) if tokens is not None else labels.size(1)
```
duncanriach, line 239: In the mixer, at the point where total_tokens was being set, the sequence length was index 1 in the tensor. But that was after a transpose in mixer that moved it from index 0. Are we sure that the sequence length is index 1 at this point?

duncanriach: I just checked: index 1 is total_tokens here. No issue with that.

pretrain_mamba.py
mathemakitten: Yes, I'll separate this out and fix. Using `max_seqlen_q == max_seqlen_q == rl sequence packing bin size` is only correct in the RL path with sequence packing on and incorrect during regular training.

duncanriach: I don't understand how it could be correct in any context. It doesn't make any sense.

duncanriach: You have cu_seqlens and then why would you add the size of the biggest bin to the end of that? Would that not result in a non-monotonically-increasing sequence of numbers?

mathemakitten: I'm not sure why you would assert that it could not _possibly be correct in any context_ when this diff can be empirically verified.

The difference is that in the base Mamba code, `max_seqlen_q` is the longest _individual_ sequence. In that case, yes, you're right, it would be incorrectly non-monotonic. 

With RL sequence packing, max_seqlen_q = bin_size = max sequence length (total token count) because of the way we construct the padding based on the _number_ of bins, so you _do_ get a monotonically increasing sequence, which _does_ resolve to the exact same end product of a tensor of form `[0, 0, 0, 0, 0, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3, 3]`, which is [0]*(5-0) + [1]*(7-5) + [2]*(11-7) + [3]*(16-11) in the example. You can verify this for yourself by executing them side-by-side. 

That being said, I agree with you that it doesn't work for the base Mamba case and will sort this out accordingly with Teo's suggestion of taking a `total_tokens` parameter.

megatron/core/transformer/cuda_graphs.py
mathemakitten: Root caused: the router's float64 dtype cannot be mapped in transformer_engine/pytorch/utils.py. We fallback to a strong ref in the meantime.

megatron/core/transformer/utils.py
mathemakitten: Update: We do indeed still need to disable cudagraphs for the reference logprobs case when `torch.no_grad()` is on.

megatron/core/ssm/mamba_mixer.py
mathemakitten: Yes, this has now been removed.

megatron/core/ssm/mamba_mixer.py
mathemakitten: I've moved this docstring to `megatron/core/packed_seq_params.py` along with the examples.

mathemakitten: Have moved this to `megatron/core/packed_seq_params.py` as well.

pretrain_mamba.py
mathemakitten: Please see the rewrite, which reconciles both the RL sequence packing code and the regular hybrid mamba path through the post_init of `PackedSeqParams`.

megatron/core/packed_seq_params.py
mathemakitten: See comments in code tab re: reconciling the RL sequence pack + regular hybrid mamba paths. The rewrite now works for both.

shanmugamr1992 (approved):

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -396,14 +398,16 @@ def permute(
...
 
-        # Create a dense expert-to-token mapping from the sparse token-to-expert mapping
-        token_indices = (
-            torch.arange(num_tokens, device=routing_map.device).unsqueeze(0).expand(num_experts, -1)
-        )
-        sorted_indices = token_indices.masked_select(routing_map)
+        # Use argsort to get indices of non-zero entries in row-major order.
+        # This is equivalent to masked_select but produces fixed-shape output,
+        # making it compatible with CUDA graph capture.
+        flat_sorted = routing_map.reshape(-1).argsort(descending=True, stable=True)
```
buptzyb, line 406: `masked_select` is a data-dependent-shaped op, but here I think the number of 1s in `routing_map` is consistent, that is `tokens.shape[0]`. Isn't this graphable?

With argsort you put all the 1's indices at the front, and all the 0's indices at the end. So the length of `sorted_indices` will be num_tokens * num_experts. Is this valid?

Or you always pass in `num_out_tokens` to get the first num_out_tokens elements?

megatron/core/transformer/cuda_graphs.py
jiemingz: But this is suppose to only be toggled around the warmup? not warmup + capture

megatron/core/transformer/utils.py
```diff
@@ -425,14 +428,118 @@ def toggle_cuda_graphs(model, set_to="none"):
...
+        if isinstance(module, MoETransformerLayer):
+            module.use_partial_cudagraphs = False
+            # Reset fwd_execution_map from the string left by the last partial
+            # phase (e.g. "postprocess") back to the full list so that
+            # non-cudagraph forward calls run all phases of the MoE layer.
+            module.mlp.fwd_execution_map = ["route", "expert_compute", "postprocess"]
+            assert hasattr(module, 'cudagraph_manager'), (
+                "MoETransformerLayer missing full cudagraph_manager; "
+                "expected it to be created at __init__ with scope = [] "
+            )
```
jiemingz: I wonder for organization if its better move this logic into MoETransformer because its touching alot of the internals, ie giving it a `module.transition_cuda_graph_scope(mode='partial'/'full')`

and there maybe you could reuse the `create_mcore_cudagraph_manager` function? wdyt?

megatron/core/packed_seq_params.py
```diff
@@ -21,3 +22,43 @@ class PackedSeqParams:
...
+        In the above example, there are three sequences in the pack.
+        In general, the output has an additional sequence index (e.g. 0, 1, 2, 3) so that any tokens
+        beyond the last padded input sequence are accounted for as an extra sequence. However, If
+        cu_seqlens_q_padded[-1] == max_seqlen then this additional sequence index will not be
+        included.
+        """
+
+        # Pre-compute seq_idx for Mamba mixer CUDA graph compatibility.
+        # seq_idx is stored as a non-field attribute so dataclasses.fields() won't include it,
+        # preventing it from being forwarded to TransformerEngine's attention forward().
```
duncanriach: Does this concern also apply to total_tokens? Will that be forwarded to TransformerEngine's attention forward(), and, if so, does that matter?

megatron/core/packed_seq_params.py
```diff
@@ -21,3 +22,43 @@ class PackedSeqParams:
...
+
+        # Example: [0, 5, 7, 11] -> [0, 5, 7, 11, 16]
```
duncanriach: I think this comment line should be moved down to just before line 52 (assignment to cu_seqlens_with_max)

duncanriach (approved):
Some comments (please address) but also approving

megatron/core/transformer/moe/moe_utils.py
mathemakitten: `masked_select` cannot be graphed even if the number of 1s in `routing_map` is consistent because it internally allocates a dynamically-sized tensor and this part is rejected during capture. It seems that since the op is capable of operating with dynamic shapes, it cannot be cudagraphed at all. You can verify this by reverting this to the old logic and checking where cudagraph capture fails. 

Every place which calls this does pass `num_out_tokens`:

```
permuted_local_hidden_states, _, self.reversed_local_input_permutation_mapping, _, _ = (
            permute(
                hidden_states,
                self.local_map,
                num_out_tokens=tokens_per_expert.sum().item(),
                fused=self.config.moe_permute_fusion,
            )
        )
```

```
self.hidden_shape_before_permute = hidden_states.shape
        (
            permutated_local_input_tokens,
            permuted_probs,
            self.reversed_local_input_permutation_mapping,
            _,
            _,
        ) = permute(
            hidden_states,
            self.routing_map,
            probs=probs,
            num_out_tokens=self.num_out_tokens,
            fused=self.config.moe_permute_fusion,
            drop_and_pad=self.drop_and_pad,
        )
```

```
            hidden_states,
            permuted_probs,
            self.reversed_mapping_for_combine,
            self.pad_offsets,
            self.tokens_per_expert,
        ) = permute(
            hidden_states,
            self.dispatched_routing_map,
            probs=self.dispatched_probs,
            num_out_tokens=self.tokens_per_expert.sum().item(),
            fused=self.permute_fusion,
            tokens_per_expert=self.tokens_per_expert,
            align_size=get_align_size_for_quantization(self.config),
        )
```

I can add an assert that `num_out_tokens` is not None if you think it is necessary.

megatron/core/transformer/cuda_graphs.py
mathemakitten: Had this while debugging but just reverted and seems to run fine, so will revert.

megatron/core/packed_seq_params.py
mathemakitten: We discard it via `self.kept_packed_seq_params.discard("total_tokens")` which is in `megatron-rl/megatron/core/extensions/transformer_engine.py`

duncanriach: Sorry to be so picky, but this seems more messy than it could be. Please consider a mechanism for setting seq_idx inside the object that does not involve adding a parameter that needs to later be removed.

megatron/core/transformer/utils.py
mathemakitten: I actually like this a lot, especially since it keeps MoE-specific utilities isolated.

megatron/core/packed_seq_params.py
mathemakitten: The discard pattern is the canonical way to do this in `extensions/transformer_engine.py`, e.g. see  `self.kept_packed_seq_params.discard("cu_seqlens_kv_padded")`, same for `cu_seqlens_q_padded`. 

There are more complicated ways we could try to do this (e.g. using InitVar) but they have other ramifications (breaking cudagraph capture, which parses the PackedSeqParams object). I do think this is in line with established ways of handling args which should not be forwarded to TE.

duncanriach: Okay. Thank you

megatron/core/transformer/moe/moe_utils.py
buptzyb: I see. How about, if `num_out_tokens` is None, go to the `masked_select` path, otherwise go to the `argsort` path?

btw have you compared the performance of the two ways? If the `argsort` way saw some regressions, we'd better only use it in graph capturing.

jiemingz (approved):

megatron/core/transformer/moe/moe_utils.py
mathemakitten: Good question. I ran a standalone benchmark and found that the argsort path is actually faster than the masked_select path by about ~1.5x, so it is actually probably optimal to do this by default unless `num_out_tokens` is None. However, it seems like `num_out_tokens` is never actually None in practice. Is there a reason to support it?

I've added `assert num_out_tokens is not None` to the argsort permute path.

buptzyb: It was argsort, but we changed it to masked_select style in MR [1915](https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/merge_requests/1915). Three reasons are given. cc @yanring to comment whether we still need masked_select.

megatron/rl/rl_utils.py
```diff
@@ -1369,14 +1240,16 @@ def prepare_data_for_update(
         with torch.no_grad(), nvtx_range("compute_logprobs", time=True):
             # Before we can update the model, we need to get the logprobs for the \pi_{old} model.
 
-            # Wrap forward_backward_func for Full iteration CUDA graph
+            _logprobs_lang_module = (
+                (model.module.module if hasattr(model.module, "module") else model.module)
+                if args.rl_training_cuda_graphs else None
+            )
```
tdene: This does not get used.

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -393,17 +395,23 @@ def permute(
...
+        if num_out_tokens is not None:
```
tdene: This is already asserted above, so no need to check.

megatron/rl/rl_utils.py
```diff
@@ -1855,29 +1743,81 @@ def megatron_rl_inference_mode(
...
-        # don't consume GPU memory during training.
+        # TODO: Remove this if statement once a change to `toggle_cuda_graphs` makes it safe to.
+        if cuda_graph_impl != "none" and not args.rl_training_cuda_graphs:
+            toggle_cuda_graphs(lang_module, 'none', reset_cuda_graphs=reset_cuda_graphs)
+
+        # Reset drop_and_pad leaked from inference decode
+        set_decode_expert_padding(unwrap_model(model[0]), set_to=False)
+
+        # Restore partial capture cudagraph scope for training if this is MoE
+        if args.num_experts:
```
tdene: Please change this to `args.num_experts is not None` to be consistent with the other if statements.

megatron/rl/rl_utils.py
```diff
@@ -1855,29 +1743,81 @@ def megatron_rl_inference_mode(
...
+            model[0].config.cuda_graph_scope = [
+                CudaGraphScope.mamba,
+                CudaGraphScope.attn,
+                CudaGraphScope.moe_router,
+                CudaGraphScope.moe_preprocess,
+            ]
```
tdene, line 1919: Can you not save `model[0].config.cuda_graph_scope` earlier in this method, so that you can restore the same scope, instead of hardcoding a specific scope?

megatron/core/transformer/transformer_layer.py
```diff
@@ -1367,6 +1404,14 @@ def _forward_mlp_partial_cudagraphs(
             residual, hidden_states, probs, shared_expert_output = self._forward_mlp_router(
                 hidden_states, padding_mask=padding_mask
             )
+
+            # After the router graph replays, the captured .copy_() operations that update
+            # self.token_dispatcher_attrs via `_maybe_dtoh_and_synchronize` are queued on the
+            # current stream but may not have completed. Synchronize to ensure
+            # self.token_dispatcher_attrs contain the correct values before the eager
+            # expert_compute phase restores them onto the token dispatcher.
+            torch.cuda.current_stream().synchronize()
```
tdene: Is it possible to wait on a CUDA event specifically for the dtoh instead of synchronizing the whole stream? Just worried that having a synchronize in here will wait for more work than necessary - if not in today's code, than later on when more code is added.

megatron/core/packed_seq_params.py
```diff
@@ -21,3 +22,43 @@ class PackedSeqParams:
...
+            # Example: [5, 2, 4, 5] -> [0, 0, 0, 0, 0, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3, 3]
+            self.seq_idx = (
+                torch.repeat_interleave(
+                    torch.arange(seq_lengths.numel(), device=cu_seqlens.device), seq_lengths
+                )
+                .to(torch.int32)
+                .unsqueeze(0)  # Add a batch dimension
+            )
+        else:
+            self.seq_idx = None
```
tdene: Since this is a dataclass, can you please list `seq_idx` above `post_init`, and let it initialize to None there? That lets you get rid of the else clause (which means you can dedent this whole code block if you want), if makes type-hinting pick up on it, and it allows for `seq_idx` to be printed and copied alongside the rest of the dataclass (as is, a `copy` will drop `seq_idx` afaik).

megatron/core/transformer/cuda_graphs.py
```diff
@@ -112,6 +112,7 @@ def _set_warmup_start():
 def _set_warmup_end():
     """Set graph warmup has ended."""
     global _IS_GRAPH_WARMUP
+    _IS_GRAPH_WARMUP = False
```
tdene, line 115: I would like to make sure we fully understand the implications of this fix before it goes through. I'm afraid that some codepath somewhere may be relying on the fact that `_set_warmup_end` is broken. But this is more paranoia than anything; if you think it's safe to fix, I won't disagree.

megatron/rl/rl_utils.py
mathemakitten: `model[0].config.cuda_graph_scope` starts as the inference scope, so it is _never_ set to this scope.

megatron/core/transformer/moe/moe_utils.py
buptzyb: I'd like @yanring to review this part. I'm good with other changes. Approved in advance.

megatron/rl/rl_utils.py
tdene: Got it, resolving.

megatron/core/packed_seq_params.py
mathemakitten: Good call, done!

megatron/core/transformer/moe/moe_utils.py
yanring: The original switch to masked_select in !1915 was part of the broader routing representation refactor (index-based → mask-based), not a deliberate choice of masked_select over argsort. The argsort approach here LGTM, approved. Thanks!

megatron/core/transformer/cuda_graphs.py
mathemakitten: Sorry, I thought it didn't make a difference but now I think it does (and it's obviously correct to have it in the first place). Bringing it back.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1128,7 +1130,13 @@ def replace_with_weak_ref(arg):
                 if not torch.is_tensor(arg):
                     return arg
 
-                ref = make_weak_ref(arg)
+                try:
+                    ref = make_weak_ref(arg)
+                except RuntimeError:
```
tdene, line 1135: Please add a warning logger print here.

megatron/core/transformer/utils.py
```diff
@@ -425,14 +427,50 @@ def toggle_cuda_graphs(model, set_to="none"):
...
+def transition_moe_to_full_cudagraphs(model):
+    """Switch MoE layers to full-layer cudagraph capture for inference.
+
+    Delegates to MoETransformerLayer.transition_cudagraph_scope('full') on each layer.
+    """
+    from megatron.core.transformer.transformer_layer import MoETransformerLayer
+
+    for module in model.modules():
+        if isinstance(module, MoETransformerLayer):
+            module.transition_cudagraph_scope('full')
```
tdene: Should these be combined into a single `transition_moe_to_cudagraphs` method?

mathemakitten: Done!

PR #3377, opened 2026-02-11
Add flexible virtual pipeline parallel (fVPP) to hybrid model

duncanriach:
# What does this PR do ?

This PR introduces **flexible virtual pipeline parallel (fVPP)** for the hybrid model, allowing users to define explicit pipeline stage boundaries directly in the hybrid layer pattern. It also consolidates and simplifies the hybrid model configuration interface by replacing multiple arguments with a single `--hybrid-layer-pattern` flag and deprecating several legacy arguments. Pipelining for the hybrid model does not support MTP standalone mode, which is supported for the GPT model.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


yobibyte:
@jalbericiola @jon-barker when this is merged, we will need to update some of the configs in internal launching scripts.

ko3n1g:
/ok to test [415c973](https://github.com/NVIDIA/Megatron-LM/pull/3377/commits/415c973bacfb2ab6a76f7ea330f5824815240ce8)

jon-barker:
@duncanriach am I correct in understanding that the hybrid patterns that don't use the new `|` operator are unchanged so it's safe to continue to load them from checkpoints, e.g. nemotron6 nano?

duncanriach:
> @duncanriach am I correct in understanding that the hybrid patterns that don't use the new `|` operator are unchanged so it's safe to continue to load them from checkpoints, e.g. nemotron6 nano?

Yes. Old checkpoints, including v3 checkpoints (that use `--hybrid-override-pattern`) will still load properly. Layer patterns that do not use `|` are also valid and will still work. This has been verified by the dynamic inference functional tests. The only regression is that if you want to enable PP>1, then you need to specify in the layer pattern how you want the model partitioned.

jon-barker:
> > @duncanriach am I correct in understanding that the hybrid patterns that don't use the new `|` operator are unchanged so it's safe to continue to load them from checkpoints, e.g. nemotron6 nano?
> 
> Yes. Old checkpoints, including v3 checkpoints (that use --hybrid-override-pattern) will still load properly. Layer patterns that do not use `|` are also valid and will still work. This has been verified by the dynamic inference functional tests. The only regression is that if you want to enable PP>1, then you need to specify in the layer pattern how you want the model partitioned.

Great - LGTM from the RL side then

duncanriach:
@jon-barker, after thinking some more about our earlier interaction, I made [this change](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/1401de192ee237e3506d62c9bbf735041385911f).

megatron/core/ssm/mamba_block.py
```diff
@@ -147,34 +125,35 @@ def __init__(
...
-                    # Transformer layers apply their own pp_layer_offset
                     layer = build_module(
                         submodules.attention_layer,
                         config=self.config,
-                        layer_number=i + 1,
+                        layer_number=layer_number,
                         pg_collection=pg_collection,
                         is_mtp_layer=is_mtp_layer,
+                        add_layer_offset=False,
+                        pp_layer_offset=pp_layer_offset,
```
rogerwaleffe, line 135: We are passing pp_layer_offset to attention, but not to MoE/MLP. Is it not needed for those layers?

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -801,11 +801,13 @@ def __init__(
...
         if mtp_layer_pattern is not None and mamba_submodules is not None:
             from megatron.core.ssm.mamba_block import MambaStack
+            from megatron.core.ssm.mamba_hybrid_layer_allocation import validate_segment_layers
 
             self.mtp_model_layer = MambaStack(
                 config=self.config,
                 submodules=mamba_submodules,
-                hybrid_override_pattern=mtp_layer_pattern,
+                layer_type_list=validate_segment_layers(mtp_layer_pattern),
+                pp_layer_offset=0,
```
rogerwaleffe, line 823: Why is this set to 0 here? MTP happens on the last PP stage right, will this value get overwritten?

megatron/core/ssm/mamba_block.py
duncanriach: All the Transformer-based layers now get add_layer_offset, which is used to disable the addition of their layer-offset amount when calculating their layer number. We now pass in the correct layer number for all layers, and set add_layer_offset to false. The attention layers also need to know what the pp_layer_offset is. This is kv-related, for inference. In the hybrid model, we pass this in and it gets stored and used for inference. For the GPTModel, it gets this value through an internal function.

megatron/core/transformer/multi_token_prediction.py
duncanriach: At the moment, MTP is always on the last PP stage. pp_layer_offset usually refers to the layer offset in the decoder. It doesn't make sense for the layer stack in MTP. It makes sense to number the layers from zero inside MTP.

megatron/core/ssm/mamba_block.py
duncanriach: So, the non-attention layers don't need to know what pp_layer_offset is

rogerwaleffe: Thanks, yeah I see the use of that is only for inference.

rogerwaleffe (approved):
LGTM.

megatron/core/models/mamba/mamba_model.py
```diff
@@ -98,29 +110,72 @@ def __init__(
...
         self.share_embeddings_and_output_weights = share_embeddings_and_output_weights
         self.position_embedding_type = position_embedding_type
         self.vp_stage = vp_stage
 
-        # Parse unified pattern to extract main and MTP components
-        from megatron.core.ssm.mamba_hybrid_layer_allocation import parse_hybrid_pattern
+        # Backward compatibility for deprecated hybrid parameters
+        if hybrid_override_pattern is not None:
+            if hybrid_layer_pattern is None:
+                warnings.warn(
```
deepakn94: For these warnings, can you use `log_single_rank` with `WARNING` log level instead, to prevent log spam?

janEbert:
From a backward-compat perspective, doesn't it make sense to add the pipeline split symbols (`|`) in the layer patterns where they would have been previously, with the "automatic" pipeline split placement? Since otherwise, previous pipelining-dependent workflows are effectively disabled without further user interaction.

megatron/core/models/mamba/mamba_model.py
```diff
@@ -73,10 +84,11 @@ def __init__(
         mamba_stack_spec: ModuleSpec,
         vocab_size: int,
         max_sequence_length: int,
-        pre_process: bool = True,
-        hybrid_attention_ratio: float = 0.0,
-        hybrid_mlp_ratio: float = 0.0,
+        hybrid_layer_pattern: str = None,
+        hybrid_attention_ratio: float = None,
+        hybrid_mlp_ratio: float = None,
```
janEbert: Types need to become Optional (`| None`) here.

megatron/core/models/mamba/mamba_model.py
```diff
@@ -98,29 +110,72 @@ def __init__(
...
+                stacklevel=2,
+            )
+            if self.hybrid_layer_pattern is None:
+                from megatron.core.ssm.mamba_hybrid_layer_allocation import pattern_from_ratios
+
+                attn_ratio = hybrid_attention_ratio if hybrid_attention_ratio else 0.0
+                mlp_ratio = hybrid_mlp_ratio if hybrid_mlp_ratio else 0.0
+                self.hybrid_layer_pattern = pattern_from_ratios(
+                    config.num_layers, attn_ratio, mlp_ratio
+                )
```
janEbert, line 170: Could add a warning here in the `else`-branch that the `hybrid_layer_pattern` takes precedence, since it's specified (`not None`). (Or error out like above.)

megatron/core/models/mamba/mamba_model.py
```diff
@@ -98,29 +110,72 @@ def __init__(
...
         self.parallel_output = parallel_output
         self.share_embeddings_and_output_weights = share_embeddings_and_output_weights
         self.position_embedding_type = position_embedding_type
         self.vp_stage = vp_stage
 
-        # Parse unified pattern to extract main and MTP components
-        from megatron.core.ssm.mamba_hybrid_layer_allocation import parse_hybrid_pattern
+        # Backward compatibility for deprecated hybrid parameters
+        if hybrid_override_pattern is not None:
+            if hybrid_layer_pattern is None:
```
janEbert: Just for future safety and consistency, could we acess `self.hybrid_layer_pattern` here as well like below?

megatron/core/models/mamba/mamba_model.py
```diff
@@ -73,10 +84,11 @@ def __init__(
         mamba_stack_spec: ModuleSpec,
         vocab_size: int,
         max_sequence_length: int,
-        pre_process: bool = True,
-        hybrid_attention_ratio: float = 0.0,
-        hybrid_mlp_ratio: float = 0.0,
```
janEbert, line 82: These are still referenced in `examples/mamba`:
- [`examples/mamba/README.md`](https://github.com/duncanriach/Megatron-LM/blob/ca7454475b3720c275435f4a77d999bb2abd8d11/examples/mamba/README.md?plain=1#L43-L46) and [further below](https://github.com/duncanriach/Megatron-LM/blob/ca7454475b3720c275435f4a77d999bb2abd8d11/examples/mamba/README.md?plain=1#L70-L87)
- [`examples/mamba/train.sh`](https://github.com/duncanriach/Megatron-LM/blob/ca7454475b3720c275435f4a77d999bb2abd8d11/examples/mamba/train.sh#L67-L68)
- [`examples/mamba/run_text_gen_server_8b.sh`](https://github.com/duncanriach/Megatron-LM/blob/ca7454475b3720c275435f4a77d999bb2abd8d11/examples/mamba/run_text_gen_server_8b.sh#L33-L34)

megatron/core/ssm/mamba_block.py
```diff
@@ -79,9 +78,8 @@ def __init__(
         submodules: MambaStackSubmodules,
         residual_in_fp32=False,
         pre_process: bool = True,
-        hybrid_attention_ratio: float = 0.0,
-        hybrid_mlp_ratio: float = 0.0,
-        hybrid_override_pattern: str = None,
+        layer_type_list: list = None,
```
janEbert: Could you change this to `list[LayerSymbols] | None`, please?

megatron/core/ssm/mamba_hybrid_layer_allocation.py
```diff
@@ -62,12 +56,142 @@ class ParsedHybridPattern:
...
+    pipeline stage separators '|', and returns the character count.
+
+    Args:
+        pattern: Full hybrid layer pattern, possibly including MTP and pipe separators.
+
+    Returns:
+        Total number of layers in the main decoder pattern.
+    """
+    main_pattern = pattern.split(Symbols.MTP_SEPARATOR)[0]
+    return len(main_pattern.replace(Symbols.PIPE, ''))
```
janEbert, line 127: For future-proofing, maybe this should just replace _all_ non-`Symbols.VALID_LAYERS` symbols with `''`?

kevalmorabia97:
> From a backward-compat perspective, doesn't it make sense to add the pipeline split symbols (`|`) in the layer patterns where they would have been previously, with the "automatic" pipeline split placement? Since otherwise, previous pipelining-dependent workflows effectively disabled without further user interaction.

Agreed. It would be good to automatically add `|` if its not present and pp size > 1 (also factoring in uneven PP parameters) so we dont break bw compatibility. We can raise a deprecation warning if the automatic `|` placement is done to let user know to manually pass it next time

megatron/core/ssm/mamba_hybrid_layer_allocation.py
```diff
@@ -62,12 +56,142 @@ class ParsedHybridPattern:
...
+        for char in parsed.mtp_pattern:
+            if char in counts:
+                counts[char] += parsed.mtp_num_depths
+
+    return (
+        counts[Symbols.ATTENTION],
+        counts[Symbols.MAMBA],
+        counts[Symbols.MLP],
+        counts[Symbols.MOE],
+    )
```
janEbert: For the return values, I know the tuple style is already followed elsewhere, but could we move towards `namedtuple`s/`dict`s instead?

Then we could also set up `counts` for all `Symbolse.VALID_LAYERS`.

This pattern would enable destructuring into the tuple on the caller side (from https://stackoverflow.com/a/52083390):
```python
from operator import itemgetter

params = {'a': 1, 'b': 2}

a, b = itemgetter('a', 'b')(params)
```

megatron/core/ssm/mamba_hybrid_layer_allocation.py
```diff
@@ -135,162 +262,97 @@ def parse_hybrid_pattern(pattern: Optional[str]) -> ParsedHybridPattern:
...
 
-def _allocate_auto(
-    total_layers_count: int, target_attention_ratio: float, target_mlp_ratio: float
-) -> list:
-    # First, allocate attention (evenly spaced, starting and ending with mamba)
-    attention_layers_count: int = round(total_layers_count * target_attention_ratio)
-    mamba_layers_count: int = total_layers_count - attention_layers_count
-    mamba_sections_count: int = attention_layers_count + 1
-    mamba_section_length: float = mamba_layers_count / mamba_sections_count
+def validate_segment_layers(segment: str) -> list:
```
janEbert: More specifically, return type is `list[str]`, right? Could you please add that constraint?

megatron/training/arguments.py
```diff
@@ -2937,20 +3028,16 @@ def _add_experimental_args(parser):
                        'To use local spec specify local as the argument.'
                        'For more details, see the model class, '
                        '`transformer_block.py`, or `transformer_layer.py`')
-    group.add_argument('--hybrid-attention-ratio', type=float, default=0.0,
-                       help='Ratio of attention layers to total layers, in the '
-                       'range [0.0, 1.0].')
-    group.add_argument('--hybrid-mlp-ratio', type=float, default=0.0,
-                       help='Ratio of mlp layers to total layers, in the '
-                       'range [0.0, 1.0].')
```
janEbert, line 2967: Is it fine to remove these outright? They didn't seem to have a deprecation notice. If they're kept, should also add appropriate warnings when validating.

megatron/training/training.py
```diff
@@ -2032,8 +2010,9 @@ def training_log(
         if args.moe_z_loss_coeff is not None:
             track_names.append("z_loss")
 
-        if args.is_hybrid_model:
-            layers = args.hybrid_override_pattern.count('E')
+        if is_hybrid_model(args):
+            from megatron.core.ssm.mamba_hybrid_layer_allocation import get_hybrid_layer_counts
+            _, _, _, layers = get_hybrid_layer_counts(args.hybrid_layer_pattern)
         else:
             layers = args.num_layers
```
janEbert, line 2035: Unrelated to this PR, but isn't this wrong for Transformers with `--moe-layer-freq != 1`?

tests/functional_tests/shell_test_utils/_run_training.sh
```diff
@@ -124,8 +124,8 @@ else
                 value=$(echo "$value" | sed 's/^\[//;s/\]$//')
                 TRAINING_PARAMS_FROM_CONFIG+="$key $value "
 
-            # Case: contains spaces
-            elif [[ "$value" == *" "* ]]; then
+            # Case: contains spaces or shell metacharacters
+            elif [[ "$value" == *" "* || "$value" == *"|"* || "$value" == *"("* || "$value" == *")"* ]]; then
```
janEbert, line 128: Slightly unrelated to this PR since behavior seems to be fine; why didn't, e.g., asterisks have to be similarly escaped, or why don't they have to be handled here? Also, should other shell metachars be added here as well so they don't cause issues in the future?

megatron/core/models/mamba/mamba_model.py
duncanriach: Good point. Thanks. Fixed.

duncanriach: Fixed with [this commit](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/907543a9a3945b8f3308d7fd290a7cb1ea4c53e6)

megatron/core/models/mamba/mamba_model.py
duncanriach: Fixed by [this commit](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/a09844d9eb161c84a657c50764fc4039bec2b60a)

megatron/core/models/mamba/mamba_model.py
duncanriach: There's no need for warning. The else branch would be if the deprecated argument were not used. No need to warn or error about that. This would just be the normal use-case going forward.

megatron/core/models/mamba/mamba_model.py
duncanriach: Fixed with [this commit](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/b5c2ac8eeb867dafe8f2398ebe5e0e0ac1714f39)

megatron/core/ssm/mamba_block.py
duncanriach: Fixed with [this commit](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/c576075456a034eef0ebe7b1d3fe87e4f248e9b1)

megatron/core/ssm/mamba_hybrid_layer_allocation.py
duncanriach: Addressed by [this commit](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/1ce2783775ecba2e93458aaff0e671cbcb2fe8ae)

megatron/training/training.py
duncanriach: I think so.

tests/functional_tests/shell_test_utils/_run_training.sh
duncanriach: I don't know. They didn't need to be. I think things should be added in the future, if needed.

megatron/core/ssm/mamba_hybrid_layer_allocation.py
duncanriach: Fixed with [this commit](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/22e821a051f98e3fd6fd149076dd5b14e3b18393)

megatron/training/arguments.py
duncanriach: I think it's fine to remove them. I don't think anyone is using them. We don't use them internally.

megatron/core/ssm/mamba_hybrid_layer_allocation.py
duncanriach: Great suggested. Addressed by [this commit](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/77d32e8d0157a117a352fd8b8929530cb3530f51).

megatron/core/ssm/mamba_block.py
```diff
@@ -191,57 +170,6 @@ def __init__(
...
-    def _select_layers_for_pipeline_parallel(self, layer_type_list):
-        assert self.config.virtual_pipeline_model_parallel_size is None, (
-            "The Mamba hybrid model does not currently support "
-            "virtual/interleaved pipeline parallelism"
-        )
-
-        pp_rank = self.pp_group.rank()
-        pp_size = self.pp_group.size()
-
-        num_layers_in_first = self.config.num_layers_in_first_pipeline_stage
```
kevalmorabia97, line 198: Can we raise an assertion somewhere if `self.config.num_layers_in_first_pipeline_stage` or `self.config.num_layers_in_last_pipeline_stage` are set instead of silently ignoring them?

And if not already, also an assertion if `count of "|" != (pp_size - 1)` or `pp_size > 1 and "|" not in hybrid_layer_pattern`

megatron/core/models/mamba/mamba_model.py
duncanriach: Addressed by [this commit](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/8a6f950bbfff722cc855be374d81acbead8d1de2).

duncanriach:
@janEbert , @kevalmorabia97: addressing your comments above about backward compat for pipeline parallel, and also the error checks you asked about, Keval. All of that should now be covered by [this commit](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/d20d78e2e1cc3b95f48b2ac55e78b162285cb247). I intend to run a local manual functional test on the backwards compat mode to check that it basically works. CI functional tests should check the new/pipe-based mode.

megatron/core/models/mamba/mamba_model.py
janEbert: I meant the `else`-branch in the most-nested `if` here (the check `if self.hybrid_layer_pattern is None`).

Branching into the hypothetical `else`-branch to the `if self.hybrid_layer_pattern is None` check would imply that the deprecated arguments are set (which a warning is already emitted about above the `if`-query), but that they are ignored because `hybrid_layer_pattern` is also used. I do think a warning about that makes sense.

janEbert:
That was fast – thank you for all the quick changes! (I unresolved https://github.com/NVIDIA/Megatron-LM/pull/3377#discussion_r2857743406, but don't think that it should prevent the merge.)

megatron/training/training.py
janEbert: It's all fine actually:
The number of MoE layers is adjusted later in the `track_moe_metrics` function according to `moe_layer_freq` when `moe_layer_freq is not None`. However, the default of the argument is 1 and `args.moe_layer_freq` is passed to `track_moe_metrics`.
The end result for Mamba is again fine, though, because for `moe_layer_freq == 1`, it basically just uses the passed `num_layers` again (a modulo check used to determine whether a layer is MoE always matches).

megatron/core/models/mamba/mamba_model.py
duncanriach: I understand what you mean now. Thank you. Addressed by [this commit](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes/ec2956a496479520f26b3569d564e5f48cb8a592).

janEbert: Thank you!

PR #3381, opened 2026-02-12
HLModel Ablation [DO NOT SUBMIT]

Phlip79:
# Possible commands

## Default config

```
torchrun --nproc_per_node=8 pretrain_hl.py \
      --hlmodel-path megatron/core/models/hl/examples/ablations.py
```

## Override `hidden_size` and `num_experts`

```
torchrun --nproc_per_node=8 pretrain_hl.py \
      --hlmodel-path megatron/core/models/hl/examples/ablations.py \
      --override hidden_size=4096 num_experts=64
```

## List override options

```
python pretrain_hl.py --hlmodel-path megatron/core/models/hl/examples/ablations.py \
      --list-overrides
```

### DIFF
--- a/megatron/core/models/hl/examples/pretrain_hl.py
+++ b/megatron/core/models/hl/examples/pretrain_hl.py
@@ -1,0 +1,520 @@
+# Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+
+"""Pretrain using an HLModel config with optional CLI overrides.
+
+Usage:
+    # Use defaults
+    torchrun --nproc_per_node=8 pretrain_hl.py \
+        --hlmodel-path examples/ablations.py
+
+    # Override a single knob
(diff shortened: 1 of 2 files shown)


megatron/core/models/hl/examples/pretrain_hl.py
```diff
@@ -0,0 +1,520 @@
...
+        if "=" not in item:
+            raise ValueError(
+                f"Invalid override '{item}'. Expected format: key=value"
+            )
+        key, value_str = item.split("=", 1)
+        overrides[key] = _coerce_value(value_str)
+    return overrides
+
+
+def load_hlmodel_spec(config_path: str):
```
duncanriach: I think we need to not use the word spec for this, otherwise it will get very confusing

Phlip79: Removed.

PR #3391, opened 2026-02-12
Fix rollout logging in wandb.

yobibyte:
After #2966, we have several turns in the trajectory, this makes it hard to read wandb rollout tables. In this PR, I get back to the old way of showing the trajectories, though it is turn per row now instead of rollout per row.


yobibyte:
Closing this in favour of #3446 .

PR #3427, opened 2026-02-14
Fix LLaVA packed sequence cu_seqlens after image token expansion

aroshanghias-nvd:
## Summary

When sequence packing is used with LLaVA, `PackedSeqParams.cu_seqlens` boundaries arrive in **collapsed space** (1 token per image), but `_preprocess_data` expands each image placeholder into `img_seq_len` embedding tokens. The `cu_seqlens` were never updated to reflect this expansion, causing attention kernels to see incorrect sub-sequence boundaries — corrupting attention masking, RoPE positions, and CP/SP sharding for every packed sub-sequence.

### Changes

- Add `_remap_packed_seq_params_after_image_expansion()` to `LLaVAModel` which translates all `cu_seqlens` entries from collapsed to expanded space using the same cumulative-sum position mapping as `_preprocess_data`.
- Call it in `forward()` right after `_preprocess_data` and before the language model.
- Add `test_forward_packed_vs_batched` parameterized over `num_samples={1, 4, 16}` that compares batched-unpacked vs packed forward passes using `token_mult_prob_error`.

### Test results

| num_samples | num_images | total_expanded | Without fix | With fix |
|---|---|---|---|---|
| 1 | 1 | 597 | 1.1032 | 1.000000 |
| 4 | 3 | 1,889 | 1.1175 | 1.000000 |
| 16 | 5 | 3,565 | 1.1319 | 1.000000 |

Test uses a real `LLaVAModel` (CLIP vision encoder + MLP projection + 3-layer GPT) on GPU with TE attention kernels — no mocking.

## Test plan

- [x] New test `test_forward_packed_vs_batched` fails without the fix (token_mult_prob_error >> 1.02)
- [x] New test passes with the fix (token_mult_prob_error = 1.000000)
- [ ] Existing `test_forward`, `test_preprocess_data` still pass
- [ ] CI passes

### DIFF
--- a/megatron/core/models/multimodal/llava_model.py
+++ b/megatron/core/models/multimodal/llava_model.py
@@ -982,6 +982,10 @@ class LLaVAModel(MegatronModule):
         if num_image_tiles is None and images is not None:
             num_image_tiles = torch.ones(images.shape[0], dtype=torch.int, device=input_ids.device)
 
+        _image_token_index = (
+            image_token_index if image_token_index is not None else self.image_token_index
+        )
+
         combined_embeddings, new_labels, new_loss_mask = self._preprocess_data(
             image_embeddings,
             language_embeddings,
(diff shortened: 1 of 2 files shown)


aroshanghias-nvd:
Let's drop this idea for now. It turns out that Megatron-LM is handling CP and sequence packing alright, because it uses Energon, and Energon does image token handling at data preparation. So if we update MCore, it will break MLM SFT!

PR #3434, opened 2026-02-15
Use Protocols to type-check linear_proj submodules of Attention

nschank:
# What does this PR do ?
Defines Protocols representing `linear_proj` submodules, and uses them instead of ModuleSpec to enable typechecking of its construction in SelfAttention, CrossAttention, and MLA.

I also updated `Backend` to return `linear_proj` specifically, allowing type-checking of `RowParallelLinear` types as instances of `linear_proj` directly (otherwise `Backend` "hides" the type and makes no type-checking occur).

While I was in `attention`, I also updated the naming conventions of the existing interfaces to match what we've finalized on.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/core/models/gpt/gpt_layer_specs.py
```diff
@@ -119,7 +119,7 @@ def get_gpt_layer_with_inference_submodules(
                     linear_kv_down_proj=backend.linear(),
                     linear_kv_up_proj=linear_kv_up_proj,
                     core_attention=backend.core_attention(),
-                    linear_proj=backend.row_parallel_linear(),
+                    linear_proj=backend.linear_proj(),
```
yashaswikarnati: may be I'm missing something,, not sure if we would want to have the row_parallel in the naming here to easily idenity if row/column parallel is being used without having to look at whats being returned by the backend. just a taught.

nschank: Yeah that would be totally fine - like `row_parallel_linear_proj`?

The problem I'm trying to solve here is that `BackendSpecProvider`, by using inheritance, makes it super hard to define a return type that can actually be type checked. The return type of `row_parallel_linear` sort of needs to match a union of Protocols (the ones for `linear_fc2` and `linear_proj` at least), which Python doesn't support in a clean way right now ("intersection types" are the key phrase to look up; unfortunately existing workarounds won't work because the `Protocols` define the same method and it won't make sure that the class satisfies _both_ interfaces).

nschank:
Resynced after coming back from travel, sorry for delay!

megatron/core/models/gpt/gpt_layer_specs.py
nschank: Yash and I synced offline, I will update the name to `row_parallel_linear_proj`!

yashaswikarnati (approved):
synced offline, just had a minor comment,overall lgtm!

jaredcasper:
>I also updated Backend to return linear_proj specifically, allowing type-checking of RowParallelLinear types as instances of linear_proj directly (otherwise Backend "hides" the type and makes no type-checking occur).

Can you expand on this a bit? I'm guessing this is adding the row_parallel_linear_proj() function in addition to the "row_parallel_linear()" function? Don't those have the same inputs/outputs so same types? Why the need for a special one for "_proj"?

megatron/core/transformer/attention.py
```diff
@@ -205,10 +204,42 @@ def __call__(
         layer_number: int,
         attn_mask_type: AttnMaskType,
         attention_type: str,
-        cp_comm_type: Optional[str],
-        softmax_scale: Optional[float],
-        pg_collection: Optional[ProcessGroupCollection],
-    ) -> CoreAttention: ...
+        softmax_scale: float | None,
+        cp_comm_type: str | None,
```
santhnm2: Small nit: looks like these were reordered? It shouldn't matter since these have to be passed as kwargs but just pointing out in case you want invocation order to match parameter list

nschank:
@jaredcasper Sure! Fair criticism, this is sorta in a partial state so maybe I should update with a TODO for clarity or something. I'm trying to solve the following problem:

```python
backend: BackendSpecProvider = ...
submodules = SelfAttentionSubmodules(..., linear_proj=backend.get_type(), ...)
```

`SelfAttentionSubmodules.linear_proj` has a specific interface it _wants_ to require - it knows the exact signature that a `LinearProjBuilder` is supposed to satisfy, and same for the `LinearProjInterface` it must return. So whenever you provide something via `linear_proj=`, the type checker is given the opportunity to check that the interface actually matches.

It can only do so if the thing being _passed_ to `linear_proj=` actually has a type which can be tested against that interface. This is true of _specific_ classes (so if I pass something of type `type[RowParallelLinear]`), unions of classes, `Callable`s, `functools.partial`, etc.

But the return type of `BackendSpecProvider.row_parallel_linear()` is just `type`. `type` is basically equivalent to :shrug: as far as the type-checker is concerned, so doing `linear_proj=backend.row_parallel_linear()` will not catch a type error. Individual subclasses of `BackendSpecProvider` can provide a narrower return type for `row_parallel_linear`, which helps _somewhat_ (if callers are using a subclass directly), but any time a caller is using something which the type-checker only knows is a `BackendSpecProvider` (but not which kind) then it will not type-check `row_parallel_linear`.

I don't have a great Protocol to use here for what _generically_ a method named `row_parallel_linear()` should actually return - there are at least two distinct Protocols that `row_parallel_linear()` needs to satisfy (`LinearProjBuilder` and `LinearFc2Builder`), and it's not entirely obvious those two things are required to have identical interfaces. The ideal world would be if I could just say the return type is `LinearProjBuilder & LinearFc2Builder` (i.e. it must satisfy _both at once_) but Python doesn't support that.

Thus, my proposed solution here is effectively to have `BackendSpecProvider` offer individual methods for each particular `Builder` protocol that we end up introducing. If we later merge `LinearProjBuilder` and `LinearFc2Builder` into a single `LinearLayerBuilder` then both `column_parallel_linear` and `row_parallel_linear` could use it; but in the meantime I think we should have `row_parallel_linear_proj` (returning `LinearProjBuilder`) and I will rename `row_parallel_linear()` to `row_parallel_linear_fc2() -> LinearFc2Builder`. This basically means `BackendSpecProvider` might return the same class from multiple separate methods, but each one is enforcing that that class satisfies a different interface.

megatron/core/transformer/attention.py
nschank: Fixed - not sure why lol, must have copy-pasted or something?

shanmugamr1992 (approved):

santhnm2 (approved):

jaredcasper:
>it's not entirely obvious those two things are required to have identical interfaces.

Why not? Both are row parallel linear layers. Do we have cases where the fc2 row_parallel_linear has a different interface than the linear_proj row_parallel_linear? If so that should be fixed. I don't want a backend to say "this is what I want you to use for linear proj" and "this is what I want you to use for fc2"... That's not a backend, that's just an additional layer to specs (which is already too confusing as it is).  I want a backend to say "when you need a row_parallel_linear layer, wherever it is, this is the thing to use."

nschank:
> Why not? Both are row parallel linear layers. Do we have cases where the fc2 row_parallel_linear has a different interface than the linear_proj row_parallel_linear?

This is fair! I was trying to be less restrictive than that, and let the two APIs evolve independently. I don't really have an opinion between the two - if we want to basically merge the Linear protocols into one, and enforce that for all callers, that would be totally fine with me.

I would personally recommend letting me get this in first, and then I can merge them as an immediate followup - merging the interfaces for just row_linear will be a nontrivial task, and similarly it will be a bit of work to do the same for column_linear. So if you're fine with the somewhat gross intermediate step, I can follow up. But if not, I'm happy to spend some time getting together the full thing. Which makes more sense to you? @jaredcasper

jaredcasper:
I think it makes sense to have the backend define types for general layer types, not specific layers (i.e. define "row parallel linear" instead of "type specifically for the linear proj", otherwise, as I said, it's doing the same thing as the spec in general, just hidden behind yet another layer of abstraction. Putting this in the meantime adds this API to the backend that would then need to be changed again. Let's just go straight to row_parallel_linear().

nschank:
SG, will get back in a day or so with the update!

nschank:
@jaredcasper I realized the relevant work is actually somewhat independent so am opening a separate PR for it here: https://github.com/NVIDIA/Megatron-LM/pull/4087 - I simply reverted the Backend changes here so now we're just focusing on linear_proj, and the issue I noted with type-checking will be fixed by that PR.

gautham-kollu:
appears to be cluster related. Rerunning failed tests
2026-04-03T23:49:21.6764216Z 8-task-1-0/0 [default0]:        handles = [pynvml.nvmlDeviceGetHandleByIndex(i) for i in physical_device_indices]
2026-04-03T23:49:21.6764692Z 8-task-1-0/0 [default0]:        for i in range(len(handles)):
2026-04-03T23:49:21.6765036Z 8-task-1-0/0 [default0]:            for j in range(i + 1, len(handles)):
2026-04-03T23:49:21.6765587Z 8-task-1-0/0 [default0]:                status = pynvml.nvmlDeviceGetP2PStatus(handles[i], handles[j], pynvml.NVML_P2P_CAPS_INDEX_NVLINK)
2026-04-03T23:49:21.6766172Z 8-task-1-0/0 [default0]:>               assert status == pynvml.NVML_P2P_STATUS_OK, \
2026-04-03T23:49:21.6766766Z 8-task-1-0/0 [default0]:                    f'No NVLink connection between GPU {physical_device_indices[i]} and GPU {physical_device_indices[j]}, ' \
2026-04-03T23:49:21.6767468Z 8-task-1-0/0 [default0]:                    f'but allow_nvlink_for_{"low_latency" if low_latency_mode else "normal"}_mode=True'
2026-04-03T23:49:21.6768294Z 8-task-1-0/0 [default0]:E               AssertionError: No NVLink connection between GPU 0 and GPU 4, but allow_nvlink_for_normal_mode=True

jaredcasper (approved):

nschank:
Sorry for delayed response! Synced and fixed import issue.

yashaswikarnati:
/ok to test [bf377a2](https://github.com/NVIDIA/Megatron-LM/pull/3434/commits/bf377a2ee12ce8ab26cf38f36a3ef7e6bf19e58e)

mathemakitten (approved):
Re-approving on behalf of @santhnm2

PR #3446, opened 2026-02-16
Log RL metrics per environment

yobibyte:
This PR makes 2 main things:
- Adds per-environment wandb logging for rl runs, e.g. we can now track average rollout length per environment.
- Refactors the wandb logic. We now also do not store the mean/min/max, but compute them on the fly. I factor out prep_wandb_metrics() function and test some logic of it by a unit test.

### DIFF
--- a/megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
+++ b/megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
@@ -4,6 +4,7 @@ import asyncio
 import logging
 import time
 import traceback
+import uuid
 
 from megatron.core.inference.sampling_params import SamplingParams
 from megatron.core.tokenizers.text.parsers import PARSER_MAPPING
@@ -61,7 +62,7 @@ try:
             # Check for 'logprobs' (bool) and 'top_logprobs' (int)
             return_log_probs = bool(req.get("logprobs", False))
             top_n_logprobs = int(req.get("top_logprobs", 0)) if return_log_probs else 0
-            skip_prompt_log_probs = bool(req.get("skip_prompt_log_probs", False))
+            skip_prompt_log_probs = bool(req.get("skip_prompt_log_probs", True))
             add_BOS = bool(req.get("add_BOS", False))
 
             # The engine only handles add_BOS for string prompts, not pre-tokenized
@@ -96,17 +97,20 @@ try:
         for _ in range(n):
             tasks.append(client.add_request(prompt_tokens, sampling_params))
 
-        start_time = time.perf_counter()
+        if current_app.config['verbose']:
+            start_time = time.perf_counter()
+
         try:
             batch_results = await asyncio.gather(*tasks)
         except Exception as e:
             logger.error(f"Error during inference: {e}")
             return f"Error during inference: {e}", 500
 
-        logger.info(
-            f"Batch of {len(tasks)} requests (n={n}) processed in "
-            f"{time.perf_counter() - start_time:.2f}s"
-        )
+        if current_app.config['verbose']:
+            logging.info(
+                f"Batch of {len(tasks)} requests (n={n}) processed in "
+                f"{time.perf_counter() - start_time:.2f}s"
+            )
 
         # --- 4. Format OpenAI Response ---
         choices = []
@@ -177,6 +181,12 @@ try:
             if "reasoning" in metadata:
                 message["reasoning"] = metadata["reasoning"]
 
+            # Replicate data in the message field for compatibility.
+            message["prompt_token_ids"] = result["prompt_tokens"]
+            message["generation_token_ids"] = result["generated_tokens"]
+            message["generation_log_probs"] = result.get("generated_log_probs", None)
+            return_log_probs = sampling_params.return_log_probs
+
             choice_data = {
                 "index": request_idx,
                 "message": message,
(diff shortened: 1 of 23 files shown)


megatron/rl/rl_utils.py
```diff
@@ -790,95 +786,158 @@ def compute_group_stats(
...
+    advantages = group_stats.advantages
+    metrics = metrics | prep_wandb_metrics(wandb_writer=wandb_writer, 
+        traj_lens=traj_lens, turn_lens=turn_lens, rewards=rewards, num_turns=num_turns, advantages=advantages)  
+    env_stats = lambda cont, idx: [cont[i] for i in idx]
+    for env_id in set(group_stats.env_ids):
+        env_idx = [i for i, eidx in enumerate(group_stats.env_ids) if eidx == env_id]
+
+        # Advantages are flattened, we need to be more careful with those.
+        group_turn_counts = [sum(nt) for nt in num_turns]
+        env_advantages = []
```
tdene, line 924: This is never used.

megatron/rl/rl_utils.py
```diff
@@ -790,95 +786,158 @@ def compute_group_stats(
...
+    num_turns = group_stats.num_turns
+    advantages = group_stats.advantages
+    metrics = metrics | prep_wandb_metrics(wandb_writer=wandb_writer, 
+        traj_lens=traj_lens, turn_lens=turn_lens, rewards=rewards, num_turns=num_turns, advantages=advantages)  
+    env_stats = lambda cont, idx: [cont[i] for i in idx]
+    for env_id in set(group_stats.env_ids):
+        env_idx = [i for i, eidx in enumerate(group_stats.env_ids) if eidx == env_id]
+
+        # Advantages are flattened, we need to be more careful with those.
+        group_turn_counts = [sum(nt) for nt in num_turns]
```
tdene: This doesn't need to be inside the for loop

megatron/rl/rl_utils.py
```diff
@@ -260,25 +260,21 @@ def verify_model_weights_swap(
...
-    mean_reward: float
-    mean_length: float
-    mean_length_std: float
-    max_length: float
-    min_length: float
-    reward_means: list[float]
-    reward_stds: list[float]
-    rewards: list[float]
+    rewards: list[list[float]] # inner list is for a group
+    env_ids: list[int] # same length as len(rewards)
```
tdene: This is `list[str]` or `list[str | None]`

megatron/rl/rl_utils.py
```diff
@@ -260,25 +260,21 @@ def verify_model_weights_swap(
...
+    turn_lens: list[list[int]] #
```
tdene: Empty comment

megatron/rl/rl_utils.py
```diff
@@ -790,95 +786,158 @@ def compute_group_stats(
...
+
+    Args:
+        wandb_writer: Wandb run to log to.
+        traj_lens: Grouped list of trajectory lengths.
+        turn_lens: Grouped list of turn lengths.
+        rewards: Grouped list of rewards.
+        num_turns: Grouped list of number of turns in the trajectories.
+        advantages: Flattened list of advantages.
+        tokenizer: Tokenizer to untokenize trajectories for logging.
+        example_groups: A dict with values as list of rollouts of one group to log examples of trajectories. Keys are env names.
```
tdene: This is referring to `example_group: list[TokenRollout | Rollout] | None = None,`; doctstring is wrong.

megatron/rl/rl_utils.py
```diff
@@ -1135,13 +1192,17 @@ def prepare_data_for_update(
         # Note :- For EP, do not use the expert data parallel group here. Always 
         # use the regular data parallel group. 
 
-        # Use one group as an exampling for logging later.
-        example_group = rollouts[0]
+        # Get example group per environment to log their rollouts.
+        example_groups = {}
+        for g in rollouts:
+            if g[0].env_id not in example_groups:
+                example_groups[g[0].env_id] = g
```
tdene, line 1200: `RewardOnlyAgent` says `env_id: str | None = None`. But in practice, all of our environments do set an `env_id`, so in practice it's not a problem.

Still though, we should either `None`-check here, or we should enforce `env_id` as required in `RewardOnlyAgent`.

megatron/rl/rl_utils.py
yobibyte: Good catch, I'm removing None from the env_id type hints. @ArEsKay3 do you see any potential problems with it?

megatron/rl/rl_utils.py
yobibyte: Fixed.

megatron/rl/rl_utils.py
yobibyte: Thanks, fixed!

megatron/rl/rl_utils.py
yobibyte: Shoot, I forgot to replace advantages I sent to the function. This should be fixed now, please, have a look again.

megatron/rl/rl_utils.py
yobibyte: Moved out.

megatron/rl/rl_utils.py
yobibyte: I do not know why we had None there. I added env_id to Countdown which was the only None-id env probably because we forget. Maybe @ArEsKay3 remembers why we need None there.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
```diff
@@ -61,7 +61,7 @@ async def chat_completions():
             # Check for 'logprobs' (bool) and 'top_logprobs' (int)
             return_log_probs = bool(req.get("logprobs", False))
             top_n_logprobs = int(req.get("top_logprobs", 0)) if return_log_probs else 0
-            skip_prompt_log_probs = bool(req.get("skip_prompt_log_probs", False))
+            skip_prompt_log_probs = bool(req.get("skip_prompt_log_probs", True))
```
tdene: Please remove this diff; it's already been fixed in `main`, and having this diff adds extra reviewer burden on other teams.

yobibyte: Done

PR #3478, opened 2026-02-18
Automatically add review label

Phlip79:
# What does this PR do ?

Automatically add review label (either expert or final) when a PR is created.


Phlip79:
Review process is changing in #3659.

PR #3504, opened 2026-02-19
Fixed fp32 residuals

mkhona-nvidia:
# What does this PR do ?
Fix FP32 residual stream activations in Megatron which currently silently break. 

### Problem
The existing fp32_residual_connection config flag was broken, it never actually kept the residual stream in fp32. The core issue was in `_bias_dropout_add_func` (used by every layer type), which downcast the fp32 residual to match the computation dtype (bf16/fp16):

Previously: defeated fp32 residuals by casting residual DOWN to x's dtype:

`residual = residual if residual.dtype == x.dtype else residual.to(x.dtype)`

This meant:
The embedding layer correctly emitted fp32 tensors when `fp32_residual_connection=True`

But the very first `bias_dropout_add` call cast the residual back to bf16. The fp32 residual stream was lost from layer 0 onwards
Additionally, the transformer layer never cast its residuals to fp32 — only the Mamba layer had its own `residual_in_fp32` flag, which also got defeated by the same downcast.

### Solution

Reverse the casting direction in `_bias_dropout_add_func`: instead of downcasting the residual, upcast x (and bias) to match the residual's dtype. This preserves the fp32 residual stream through every bias-dropout-add call.

`megatron/core/fusions/fused_bias_dropout.py`	Upcast x/bias to residual dtype instead of downcasting residual to x dtype

`megatron/core/transformer/transformer_layer.py`	Cast residual to fp32 at all 4 residual sites (self-attn, cross-attn, MLP, MoE router) when `fp32_residual_connection=True`

`megatron/core/ssm/mamba_layer.py`	Respect config.fp32_residual_connection in addition to the existing `residual_in_fp32` flag

`megatron/core/transformer/transformer_config.py`	Auto-override pipeline_dtype to torch.float when `fp32_residual_connection=True` to prevent PP communication dtype mismatches

### Why the old approach was wrong

The original downcast was added to prevent pipeline parallel communication hangs (the residual would upcast the output to fp32, mismatching the bf16 PP recv buffers). The correct fix is to set `pipeline_dtype = torch.float` when fp32 residuals are enabled, which this PR does automatically in `TransformerConfig.__post_init__`. The inference wrapper (`abstract_model_inference_wrapper.py`) already handled this correctly.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/tests/unit_tests/fusions/test_bias_dropout_fusion.py
+++ b/tests/unit_tests/fusions/test_bias_dropout_fusion.py
@@ -5,7 +5,6 @@ import torch
 
 from megatron.core.fusions.fused_bias_dropout import _bias_dropout_add_func, get_bias_dropout_add
 
-
 # ---------------------------------------------------------------------------
 # Existing test: fused vs. unfused parity (same dtype)
 # ---------------------------------------------------------------------------


megatron/core/transformer/transformer_config.py
```diff
@@ -930,6 +930,12 @@ def __post_init__(self):
         details.
         """
         super().__post_init__()
+
+        # When fp32 residual connections are enabled, pipeline parallel communication must
+        # use fp32 to match the dtype of the residual stream between pipeline stages.
+        if self.fp32_residual_connection and self.pipeline_dtype is not None:
+            self.pipeline_dtype = torch.float
```
janEbert, line 949: Could you add a `warnings.warn` here so users aren't confused when their pipeline dtype changes?

mkhona-nvidia: done

janEbert: Thank you!

janEbert: I just realized the warning shouldn't be output when `self.pipeline_dtype` is already `torch.float`.

mkhona-nvidia: good point, fixed

deepakn94: Please only warn on a single rank if possible. We don't want output spam.

ananthsub (approved):

megatron/core/transformer/transformer_config.py
mkhona-nvidia: there is a warnings.warn all over that config file, not just the change I made. Shall we make another PR that fixes this for all the others?

(made a change, please let me know if that is ok, it needs `log_single_rank`)

PR #3509, opened 2026-02-20
Change the cudagraph distribution from linearly to exponentially-decreasing + grid for mixed prefill

mathemakitten:
# What does this PR do ?

This changes the distribution of cudagraphs from linearly-spaced in the top end to exponentially decreasing. In standalone inference, this is a >2x decrease in the number of graphs, a 15GB memory decrease, and slightly-better-throughput-than-before. 

These savings are possible because many of the existing graphs were redundant (at near-zero-padding token counts in the upper range where the next-largest graph would have done the job anyway), or not useful sizes. At max_tokens=16384 (RL use case) we were producing an enormous number of graphs, but only replaying at the max batch size since rollouts are sustained and the mixed/prefill graphs were all captured at a fixed request count of 16.

We also include a geometric distribution in the mixed loop: previously, every mixed CG used prefill request count (P) = cuda_graph_mixed_prefill_request_count (default 16). A real batch with P != 16 would slot-count-match a captured P=16 graph but the captured graph's prefill metadata laid out tokens assuming 16 prefill slots, which doesn't replay cleanly when real P differs. This means that mixed CGs were captured but were mostly unusable. Replaced with a grid e.g. `{1, 2, 4, 8, …, max_requests}` so real batches find a captured CG within a 2x factor of their actual P value.

breaking: --inference-dynamic-batching-cuda-graph-mixed-prefill-count (mapped to cuda_graph_mixed_prefill_request_count) is now an on/off toggle rather than a numeric specifier. > 0 enables mixed CGs across the full P-grid; <= 0 disables mixed CGs (decode-only path, as before).

Enable with `--inference-dynamic-batching-cuda-graph-sizing-distribution`.

main: 
```
9b | dynamic | graphs 123 | uvm 0 | requests: cli, n 1, g 10, dur 1.0e+01 r/sec 1.0e+02 | bf: 10 GB, 399 chunks [r 396, t 16384] … throughput: 135.500 tok/s …  total time: 0.074s … mem 33.2/40.2 GB … steps: 10 … capture --
```

now: 
```
9b | dynamic | graphs 60 | uvm 0 | requests: cli, n 1, g 10, dur 1.0e+01 r/sec 1.0e+02 | bf: 10 GB, 399 chunks [r 396, t 16384] … throughput: 137.488 tok/s …  total time: 0.073s … mem 29.8 allocated/36.3 reserved GB … steps: 10 … capture --
```

We also include bonus logging when cudagraphs are created to understand pool reuse efficiency. Logs now look like this, which tells us that while we allocated an extra 256kb for this graph, it did not increase the actual reserved mempool space: `INFO:root:  [graph 65/65] [1]: 0 P + 1 D | pool reserved=5.5 gb (Δiter=0 bytes) pool allocated=1.8 gb (Δiter=256.0 kb)`

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


janEbert:
Hey, are there empirics available to support the change? Should the old setting still be supported for cases where it may be better?

Also, do any tests need to be updated because of this?

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -232,38 +233,40 @@ def _calculate_cuda_graph_token_counts(
...
-                len(cuda_graph_token_counts) == 0
-                or cuda_graph_token_counts[-1] != cuda_graph_max_tokens
-            ):
-                cuda_graph_token_counts.append(cuda_graph_max_tokens)
-            cuda_graph_token_counts.reverse()
+            return [cuda_graph_max_tokens]
+
+        # Exponentially decreasing, stops after num_cuda_graphs entries
+        # or when below the minimum size.
+        cuda_graph_token_counts = []
```
yobibyte: Hey Helen! Should we test some edge cases for this one in unit tests?

mathemakitten:
> Hey, are there empirics available to support the change? Should the old setting still be supported for cases where it may be better?
> 
> Also, do any tests need to be updated because of this?

The empirics are the reinforcement learning runs. I can provide internal pointers if you need. I don't think anyone can presently make a strong case for the old setting.

I will update the values for `test_cuda_graph_token_counts`.

megatron/core/inference/batch_dimensions_utils.py
mathemakitten: The existing unit test is `test_cuda_graph_token_counts`. I will update it accordingly.

janEbert:
> The empirics are the reinforcement learning runs. I can provide internal pointers if you need. I don't think anyone can presently make a strong case for the old setting.

Awesome, thank you!

janEbert (approved):

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -232,38 +233,40 @@ def _calculate_cuda_graph_token_counts(
...
-            )
-            if (
-                len(cuda_graph_token_counts) == 0
-                or cuda_graph_token_counts[-1] != cuda_graph_max_tokens
-            ):
-                cuda_graph_token_counts.append(cuda_graph_max_tokens)
-            cuda_graph_token_counts.reverse()
+            return [cuda_graph_max_tokens]
+
+        # Exponentially decreasing, stops after num_cuda_graphs entries
```
lmcafee-nvidia: I also vote to leave the linear-spaced CGs as an option, there's no harm in doing so since the code is already setup, we can just default to exponential in the arguments.

One reason for keeping this setting & code is that vLLM uses linear spacing, they just create a ton more graphs than we do because they can create them so quickly and efficiently, and I think that just speaks to how unoptimized our CG system is. So I personally would keep the old option, and just plan to use it in the future.

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -232,38 +233,40 @@ def _calculate_cuda_graph_token_counts(
...
+            if rounded not in cuda_graph_token_counts:
+                cuda_graph_token_counts.append(rounded)
+            val //= 2
+            if val < rounder:
+                break
+
+        # Include a (possibly extra) size-1 graph
+        min_token_count = math.ceil(1 / tp_size) * tp_size
+        if cuda_graph_token_counts[-1] != min_token_count:
+            cuda_graph_token_counts.append(min_token_count)
```
lmcafee-nvidia: not sure how strict we want to be about sticking to `num_cuda_graphs`, but this line will generally cause use to have `num_cuda_graphs + 1` graphs. Probably a minor concern, but might want to consider if this could be confusing to anyone.

also, why are we adding this size-1 graph? we didn't have this before

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -232,38 +233,40 @@ def _calculate_cuda_graph_token_counts(
...
-            cuda_graph_token_counts.reverse()
+            return [cuda_graph_max_tokens]
+
+        # Exponentially decreasing, stops after num_cuda_graphs entries
+        # or when below the minimum size.
+        cuda_graph_token_counts = []
+        val = cuda_graph_max_tokens
+        for _ in range(num_cuda_graphs):
+            # Round down to multiple of rounder, then up to multiple of TP size
+            rounded = max(rounder, (val // rounder) * rounder)
```
lmcafee-nvidia, line 305: the old code guaranteed that `cuda_graph_max_tokens` is in the list, but now we don't have that guarantee anymore. Do we care about this, e.g., someone wants to very strictly set the max cuda graph size?

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -232,38 +233,40 @@ def _calculate_cuda_graph_token_counts(
...
+            rounded = math.ceil(rounded / tp_size) * tp_size
+            if rounded not in cuda_graph_token_counts:
+                cuda_graph_token_counts.append(rounded)
+            val //= 2
+            if val < rounder:
+                break
+
+        # Include a (possibly extra) size-1 graph
+        min_token_count = math.ceil(1 / tp_size) * tp_size
```
lmcafee-nvidia: this always evaluates to `tp_size`, so might as well simplify to `min_token_count = tp_size`

tests/unit_tests/inference/engines/test_dynamic_engine.py
```diff
@@ -694,11 +694,11 @@ def test_cuda_graph_token_counts(self) -> None:
         for num_cuda_graphs, expected_cuda_graph_token_counts in [
```
lmcafee-nvidia: since the cuda graph code is dependent on `tp_size`, I think it would be good to parameterize some `tp_size` tests into this test.

megatron/core/inference/batch_dimensions_utils.py
mathemakitten: I would be against adding a new flag to toggle the distribution of inference cudagraphs. Inference already has a lot of flags and users don't know how to combine them effectively, and there is no empirical case that makes the argument to keep linear distribution around at the moment. I will leave a TODO to re-enable when someone wants to take it on.

megatron/core/inference/batch_dimensions_utils.py
mathemakitten: Having one more size-1 graph is minorly confusing if you're counting graphs but not functionally breaking. 

I will replace a middle graph with the size-1 graph to adhere to `num_cuda_graphs`.

megatron/core/inference/batch_dimensions_utils.py
mathemakitten: I will bring this back and sub out a middle graph for it.

janEbert (approved):

megatron/core/inference/batch_dimensions_utils.py
mathemakitten: Also, #3527 already implements the vllm strategy orthogonal to this.

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -232,38 +233,48 @@ def _calculate_cuda_graph_token_counts(
...
+        if cuda_graph_token_counts[0] != cuda_graph_max_tokens:
+            cuda_graph_token_counts.insert(0, cuda_graph_max_tokens)
+
+        # Include a (possibly extra) size-1 graph
+        if cuda_graph_token_counts[-1] != tp_size:
+            cuda_graph_token_counts.append(tp_size)
+
+        # Trim from the middle if we exceed num_cuda_graphs requested by the user
+        while len(cuda_graph_token_counts) > num_cuda_graphs:
+            cuda_graph_token_counts.pop(-2)
```
santhnm2, line 321: Can you add these lines afterwards:
```
assert len(cuda_graph_token_counts) == num_cuda_graphs
assert cuda_graph_max_tokens in num_cuda_graphs
```

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -232,38 +233,48 @@ def _calculate_cuda_graph_token_counts(
...
+        # Ensure cuda_graph_max_tokens is always included
+        if cuda_graph_token_counts[0] != cuda_graph_max_tokens:
+            cuda_graph_token_counts.insert(0, cuda_graph_max_tokens)
+
+        # Include a (possibly extra) size-1 graph
+        if cuda_graph_token_counts[-1] != tp_size:
+            cuda_graph_token_counts.append(tp_size)
+
+        # Trim from the middle if we exceed num_cuda_graphs requested by the user
+        while len(cuda_graph_token_counts) > num_cuda_graphs:
```
santhnm2, line 320: Does this line need to be
```
while len(cuda_graph_token_counts) > num_cuda_graphs and len(cuda_graph_token_counts) >= 2
```
Otherwise `pop(-2)` might give an index error or (even worse) silently wrap around?

santhnm2 (commented):
Do any functional tests need to be updated as a result of this change?

megatron/core/inference/batch_dimensions_utils.py
mathemakitten: We have a guarantee that `num_cuda_graphs >= 1` at the top of the block and we also check `while len(cuda_graph_token_counts) > num_cuda_graphs`, so we're actually already guaranteed that `len(cuda_graph_token_counts) >= 2` when this block runs.

megatron/core/inference/batch_dimensions_utils.py
sidsingh-nvidia: @mathemakitten on second thought, I am also in favor of slowly phasing out the older strategy - mostly because it's the most stress tested one we have right now. We could make yours default, while having the option to fallback onto that.

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -232,38 +233,52 @@ def _calculate_cuda_graph_token_counts(
...
+        cuda_graph_token_counts = []
+        val = cuda_graph_max_tokens
+        for _ in range(num_cuda_graphs):
+            # Round down to multiple of rounder, then up to multiple of TP size
+            rounded = max(rounder, (val // rounder) * rounder)
+            rounded = math.ceil(rounded / tp_size) * tp_size
+            if rounded not in cuda_graph_token_counts:
+                cuda_graph_token_counts.append(rounded)
+            val //= 2
+            if val < rounder:
```
sidsingh-nvidia: why aren't we allowing a CG of size 1? 
Maybe we can make this check - `if val < 1 `or `if val`

sidsingh-nvidia (changes requested):
LGTM, except for the minor nit of allowing CG of size 1. 
Still unsure if we should simply chuck out the older CG strategy or keep it around as a fallback.

megatron/core/inference/batch_dimensions_utils.py
sidsingh-nvidia: re:3527 - it does use a linear function, but builds a lot more cudagraphs compared to our default strategy.

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -232,38 +233,52 @@ def _calculate_cuda_graph_token_counts(
...
+        # Include a (possibly extra) size-1 graph
+        if cuda_graph_token_counts[-1] != tp_size:
+            cuda_graph_token_counts.append(tp_size)
+
+        # Trim from the middle if we exceed num_cuda_graphs requested by the user
+        #  Since num_cuda_graphs >= 1, this only runs when we have at least 2 elements.
+        while len(cuda_graph_token_counts) > num_cuda_graphs:
+            cuda_graph_token_counts.pop(-2)
+
+        assert len(cuda_graph_token_counts) == num_cuda_graphs
```
tdene: This assert will fail when we request too few CUDA graphs for the size of the memory buffer. It looks like it would fail the unit-test as well.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -133,13 +133,42 @@ class EngineSuspendedError(Exception):
...
+    if mem_bytes < 0:
+        return "-" + format_mem_bytes(-mem_bytes)
     for power, suffix in [(4, "tb"), (3, "gb"), (2, "mb"), (1, "kb"), (0, "bytes")]:
         suffix_bytes = 1024**power
         if mem_bytes >= suffix_bytes:
             return "%.1f %s" % (mem_bytes / suffix_bytes, suffix)
     return "%d bytes" % mem_bytes
 
 
+def _cuda_graph_mempool_bytes():
```
santhnm2: Nit: can you make return type `-> Tuple[int, int]`?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -345,8 +374,16 @@ def create_cuda_graphs(self, reset_context: bool = True):
         controller = self.controller
 
         time_start = time.time()
+        torch.cuda.reset_peak_memory_stats()
```
santhnm2: Is it safe to reset the peak memory stats here for every request entry? Would this disrupt any existing memory recording?

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -274,22 +264,41 @@ def _calculate_cuda_graph_token_counts(
...
+        # Exponentially decreasing, stops after num_cuda_graphs entries
+        # or when below the minimum size.
+        cuda_graph_token_counts = []
+        val = cuda_graph_max_tokens
+        for _ in range(num_cuda_graphs):
+            # Round down to multiple of rounder, then up to multiple of TP size
+            rounded = max(rounder, (val // rounder) * rounder)
+            rounded = math.ceil(rounded / tp_size) * tp_size
+            if rounded not in cuda_graph_token_counts:
+                cuda_graph_token_counts.append(rounded)
```
santhnm2: Small nit: If we're doing existence checks before appending can we just make `cuda_graph_token_counts` a set and then sort after we finish the loop?

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -235,35 +236,24 @@ def _calculate_cuda_graph_token_counts(
...
+            [128, 64, 32, 16, 8, 4, 2, 1]
         """
         if num_cuda_graphs == -1:
-            # automatically determine the number of CUDA graphs to
-            # capture based on the `max_requests` value
-            cuda_graph_token_counts = (
-                [1, 2, 4] + list(range(8, 256, 8)) + list(range(256, cuda_graph_max_tokens + 1, 16))
+            # Each step in the exponential-decay loop below halves the cudagraph size, so we need
+            # ~log2(max_tokens) steps with an extra +2 to leave headroom for dedup/trim.
+            auto_n = max(4, int(math.log2(max(2, cuda_graph_max_tokens))) + 2)
```
santhnm2: This logic is pretty confusing to me, I don't quite get what the `+ 2` is for and also there's a lot of nested operations - Can we simplify this?

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -235,35 +236,24 @@ def _calculate_cuda_graph_token_counts(
 
         Example:
             >>> _calculate_cuda_graph_token_counts
-            (tp_size=2, num_cuda_graphs=4, cuda_graph_max_tokens=1000)
-            [1000, 752, 504, 256]
+            (tp_size=1, num_cuda_graphs=8, cuda_graph_max_tokens=128)
+            [128, 64, 32, 16, 8, 4, 2, 1]
         """
         if num_cuda_graphs == -1:
```
santhnm2, line 259: My understanding was the -1 path was meant to exactly replicate what vLLM was doing; if that's the case should we keep that behavior? cc @sidsingh-nvidia

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -402,6 +411,23 @@ def add_if_valid(token_count: int, prefill_req_count: int, decode_req_count: int
                 )
             )
 
+            # Include the smallest decode-only graphs when auto-sizing (num_cuda_graphs == -1).
+            # Without this, TP alignment (size 1 -> tp_size) and the num_speculative_tokens floor
+            # division may drop the size 1 and size 2 graph sizes.
```
santhnm2: Isn't the size 1 graph invalid with TP=2 + sequence parallelism though?

megatron/core/inference/batch_dimensions_utils.py
mathemakitten: The vLLM strategy is overall worse for the cases we care about. Is the argument to keep it around for benchmarking purposes? If so, do you have a design suggestion for how to gate this? We can add an additional '--cuda-graph-sizing-distribution' flag. We've been proliferating inference flags which then need to be tuned, but we can do this if it's the only way.

mathemakitten: Also, note that even if we add `--cuda-graph-sizing-distribution`, the -1 path will change a bit in that it will then signal "autotune the number of graphs according to the distribution in `--cuda-graph-sizing-distribution` as opposed to today's "match the vLLM strategy".

santhnm2: Yea I think the argument for keeping this is mostly for benchmarking / debugging purposes. Perhaps we can add a modular abstraction layer similar to what we do for the grouped gemm and sampling backends - so the user would provide `--cuda-graph-sizing-distribution-strategy [auto | vllm | linear]` or something like that and then we would just have a shared interface for each of these.

megatron/core/inference/batch_dimensions_utils.py
mathemakitten: Yes, you won't hit the size-1 graph with TP=2 because of `min_decode_tokens = math.lcm(spec_unit, tp_size)`. Will update the comment though.

santhnm2 (approved):
LGTM pending functional tests passing

mathemakitten:
/ok to test [60e7929](https://github.com/NVIDIA/Megatron-LM/pull/3509/commits/60e792916c9f739c35f3058545128d1e99794f21)

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -213,57 +213,72 @@ class CUDAGraphBatchDimensionBuilder:
...
     ) -> List[int]:
         """
         Calculate CUDA graph token counts for a given configuration.
 
-        This method computes evenly-spaced token counts from step_size up to
-        cuda_graph_max_tokens, ensuring proper rounding and TP alignment.
+        Dispatches on `sizing_distribution`:
+          - EXPONENTIAL (default): halves from cuda_graph_max_tokens down to tp_size, log-spaced,
+            creates log2(max_tokens) graphs.
+          - LINEAR: small graphs [1, 2, 4] + range(8, 256, 8) + range(256, max+1, 16);
```
sidsingh-nvidia, line 231: For good decode performance, we saw that these fine-grained linear graphs were really necessary. Could we have a hybrid strategy where we do linear till max_requests and then exponential from max_requests to max_tokens?

sidsingh-nvidia (changes requested):

megatron/core/inference/batch_dimensions_utils.py
mathemakitten: RL inference saturates the batch size by construction so active `D` oscillates very close to `max_requests` and these fine-grained linear graphs are not necessary for that workload. The `linear` distribution is already in this PR for standalone serving workloads that need the dense coverage, or benchmarking workloads. Do we have data that suggests a third distribution is necessary to get in right now for some workload as opposed to in a follow-up PR?

sidsingh-nvidia: In our NemoRL runs, 90% of our time was spent in very small decodes - so it does need the fine grained linear graphs for decodes. If we wanna cuda-graph prefills in the nemo-RL land as well, then a linear + exponential strategy makes the most sense to me.

It looks like megatron-RL's behavior is different and it does not have this "90% time spent in tiny decodes" behavior. In that case, yes I agree that you do not need this hybrid linear+exponential strategy.

sidsingh-nvidia (approved):

sidsingh-nvidia (approved):

kvareddy (approved):

ericharper (approved):

PR #3515, opened 2026-02-20
Track and plot per-token off-policy in RL

tdene:
# What does this PR do ?
`policy_staleness` describes how many checkpoints off-policy a request's tokens are. `kv_cache_staleness` describes how many checkpoints off-policy a request's KV$ is. `policy_staleness >= kv_cache_staleness`. Both are tracked on a per-token basis.

When not using recompute, the two metrics are expected to overlap - with one exception. If the KV cache buffer is full and needs to evict paused requests to make room, this will null `kv_cache_staleness` while not affecting `policy_staleness`. `num_evictions` is a per-request metric that tracks such "natural" prefill recompute.

Request staleness is tracked inside the engine. Once a request finishes, it leaves the engine. Since a request may not be consumed until much later than it finishes, RL has to manually increment staleness. This is represented by `completion_gap` and is graphed as `mean_completion_gap`.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/rl/rl_utils.py
+++ b/megatron/rl/rl_utils.py
@@ -814,6 +814,36 @@ def compute_group_stats(
     return stats
 
 
+def compute_true_staleness(
+    per_token_staleness: list[list[int]],
+    completed_at_steps: list[list[int]],
+    turn_lens: list[list[int]],
+    current_iteration: int,
+) -> list[int]:
+    """Compute true per-token staleness by adding the completion gap.
+
+    Args:
+        per_token_staleness: Grouped flat list of per-token raw staleness values.
+        completed_at_steps: Grouped list of per-turn completion steps.
+        turn_lens: Grouped list of per-turn token counts.
+        current_iteration: Current training iteration.
+
+    Returns:
+        Flat list of true staleness values (one per token across all groups).
+    """
+    result = []
+    for group_staleness, group_completed, group_turn_lens in zip(
+        per_token_staleness, completed_at_steps, turn_lens
+    ):
+        token_idx = 0
+        for completed_at, num_tokens in zip(group_completed, group_turn_lens):
+            gap = current_iteration - completed_at
+            for _ in range(num_tokens):
+                result.append(group_staleness[token_idx] + gap)
+                token_idx += 1
+    return result
+
+
 def prep_wandb_metrics(
         wandb_writer: wandb_run.Run,
         traj_lens: List[List[int]],
@@ -853,17 +883,10 @@ def prep_wandb_metrics(
         data=[[np.mean(g), np.std(g)] for g in rewards],
     )
 
-    # Expand per-turn completed_at_steps to per-token using turn_lens, then compute true staleness.
-    true_policy_staleness = [
-        s + current_iteration - c
-        for g_s, g_c, g_tl in zip(policy_staleness, completed_at_steps, turn_lens)
-        for s, c in zip(g_s, (c for c, n in zip(g_c, g_tl) for _ in range(n)))
-    ]
-    true_kv_staleness = [
-        s + current_iteration - c
-        for g_s, g_c, g_tl in zip(kv_cache_staleness, completed_at_steps, turn_lens)
-        for s, c in zip(g_s, (c for c, n in zip(g_c, g_tl) for _ in range(n)))
-    ]
+    true_policy_staleness = compute_true_staleness(
+        policy_staleness, completed_at_steps, turn_lens, current_iteration)
+    true_kv_staleness = compute_true_staleness(
+        kv_cache_staleness, completed_at_steps, turn_lens, current_iteration)
 
     metrics = {
             'group_means_hist': wandb_writer.plot.histogram(
--- a/tests/unit_tests/rl/test_rl_utils.py
+++ b/tests/unit_tests/rl/test_rl_utils.py
@@ -896,6 +896,30 @@ class TestRLUtils:
         CudaGraphManager.fwd_mempools = None
         CudaGraphManager.bwd_mempools = None
 
+    def test_compute_true_staleness(self):
+        # Single group, 2 turns: turn 1 has 1 token, turn 2 has 2 tokens
+        result = rl_utils.compute_true_staleness(
+            per_token_staleness=[[1, 2, 3]],
+            completed_at_steps=[[8, 7]],
+            turn_lens=[[1, 2]],
+            current_iteration=10,
+        )
+        # token 0: 1 + (10-8) = 3
+        # token 1: 2 + (10-7) = 5
+        # token 2: 3 + (10-7) = 6
+        assert result == [3, 5, 6]
+
+        # Multiple groups
+        result = rl_utils.compute_true_staleness(
+            per_token_staleness=[[0, 0], [1]],
+            completed_at_steps=[[5], [5]],
+            turn_lens=[[2], [1]],
+            current_iteration=6,
+        )
+        # group 1: [0+1, 0+1] = [1, 1]
+        # group 2: [1+1] = [2]
+        assert result == [1, 1, 2]
+
     @pytest.mark.parametrize(
         "initialize_model_parallel",
         [pytest.param((1, 1), id="tp1-pp1")],


megatron/rl/inference/megatron.py
```diff
@@ -70,6 +70,10 @@ async def base_generate(self, request: InferenceRequest) -> InferenceResponse:
             token_ids=choice.prompt_token_ids + choice.generation_token_ids,
             logprobs=choice.generation_log_probs,
             prompt_length=len(choice.prompt_token_ids),
+            policy_staleness=getattr(choice, 'policy_staleness', None),
```
yobibyte: Why are you using getattr instead of just using the attribute?

yobibyte: According to typing, this cannot be None.

megatron/rl/inference/megatron.py
```diff
@@ -70,6 +70,10 @@ async def base_generate(self, request: InferenceRequest) -> InferenceResponse:
...
+            kv_cache_staleness=getattr(choice, 'kv_cache_staleness', None),
```
yobibyte: According to typing, this cannot be None.

megatron/rl/inference/megatron.py
```diff
@@ -70,6 +70,10 @@ async def base_generate(self, request: InferenceRequest) -> InferenceResponse:
...
+            completed_at_step=args.curr_iteration,
+            num_evictions=getattr(choice, 'num_evictions', None),
```
yobibyte: According to typing, this cannot be None.

megatron/rl/server/api.py
```diff
@@ -21,6 +21,9 @@ async def suspend(self):
     def resume(self):
         pass
 
+    def increment_staleness(self):
```
yobibyte: Can you point me to an actual implementation of it? I could not find it in the PR

megatron/rl/rl_utils.py
```diff
@@ -837,6 +879,27 @@ def prep_wandb_metrics(
                 ),
                 'advantages', 'Advantages'
             ),
+            'policy_staleness_hist': wandb_writer.plot.histogram(
```
yobibyte: Should we just have 1 table with everything to easily compare?

megatron/rl/rl_utils.py
```diff
@@ -555,6 +560,16 @@ def get_environment_rollouts(
                     # Just set up space to collect the rollouts
                     rollouts = [[None for _ in range(samples_per_group)] for _ in range(n_prompts)]
 
+        # Catch-up staleness increment: rollouts need to have their staleness values adjusted
```
yobibyte: I think this bit of logic is wrong for GRPO_ITERATIONS > 1.

For instance, when GRPO_ITERATIONS = 2, we will run the code below once -> the gap calculation will be wrong for every odd update step (when we do not collect the data). I think, the proper place for adding the gap is prep_wandb_metrics function.

Also, while I was looking at the code, I found a bug in the way we plot stuff. We plot stuff in prepare_data_for_update which is called only when we send requests to collect the data! For situations when we collect more data than we consume, we will be plotting the data we do not do updates on yet. E.g. imagine you consume one rollout per step, and collect 2 rollouts every other iteration. At step 0, you collect r1 and r2 and will plot average stats on step 0. On step on you will not plot. 
This is wrong, we should only plot data we use for an update, e.g. r1 for step 0 and r2 for step 1.

I understand, this is probably beyond the scope of this PR. But as it touches the relevant code, and adds a bug related to this, I would appreciate if you move the `maybe_log_training_metrics` call from `prepare_data_for_update` to the outside of this if branch:

```python
    if (    
        buffered_rollouts is None or
    ): 
```

megatron/rl/rl_utils.py
```diff
@@ -555,6 +560,16 @@ def get_environment_rollouts(
...
+        # due to the gap between when they finished and the current step.
+        if rank == 0:
+            for group in rollouts:
+                for rollout in group:
+                    gap = args.curr_iteration - rollout.completed_at_step
+                    if gap > 0:
```
yobibyte: What if gap is 0?

megatron/rl/agent/api.py
```diff
@@ -59,6 +63,10 @@ class TokenRollout(AgentBaseModel):
     logprobs: list[list[float]] | None = None
     env_id: str = ''
     problem_id: str | None = None
+    policy_staleness: list[int]
```
yobibyte: Why is it list[int] and not list[list[int]]? What if the rollouts are multiturn?

megatron/rl/inference/megatron.py
tdene: My bad, I thought I had fixed this. Resolved now.

tdene: See above, resolved now.

megatron/rl/inference/megatron.py
tdene: See above, resolved now.

megatron/rl/inference/megatron.py
tdene: See above, resolved now.

megatron/rl/rl_utils.py
```diff
@@ -811,6 +851,8 @@ def prep_wandb_metrics(
         advantages: Flattened list of advantages.
         tokenizer: Tokenizer to untokenize trajectories for logging.
         example_groups: A list of rollouts of one group to log examples of trajectories.
+        policy_staleness: Grouped list of per-rollout max policy staleness.
+        kv_cache_staleness: Grouped list of per-rollout max KV cache staleness.
```
yobibyte: I think you added more arguments than docstring args ;)

megatron/rl/rl_utils.py
```diff
@@ -852,6 +915,15 @@ def prep_wandb_metrics(
             'mean_advantage': np.mean(advantages),
             'nonzero_groups_ratio': np.count_nonzero(advantages)
             / len(advantages),
+            'mean_policy_staleness': np.mean([np.mean(g) for g in policy_staleness]),
+            'max_policy_staleness': max([max(g) for g in policy_staleness]),
```
yobibyte: TODO for myself:re-review this after the multi-turn comment gets resolved.

megatron/rl/rl_utils.py
```diff
@@ -910,8 +982,15 @@ def maybe_log_training_metrics(
     rewards = group_stats.rewards
     num_turns = group_stats.num_turns
     advantages = group_stats.advantages
-    metrics = metrics | prep_wandb_metrics(wandb_writer=wandb_writer, 
-        traj_lens=traj_lens, turn_lens=turn_lens, rewards=rewards, num_turns=num_turns, advantages=advantages)  
+    policy_staleness = group_stats.policy_staleness
```
yobibyte, line 1000: Do we really need all these variables instead of just directly sending group_stats.attribute?

megatron/rl/server/api.py
tdene: `megatron/rl/rl_utils.py` L1856. But hm, you're right, it doesn't need to be defined all the way down here in the `Server`; it's not a control method.

Resolved!

yobibyte: I'm not sure I understand, you are still calling it, no?

megatron/rl/rl_utils.py
tdene: How do you feel about just sending `group_stats` directly?

yobibyte: nvm, I misread the code. My bad.

tdene: Ah, nevemind, that's a dumb question.

megatron/rl/rl_utils.py
tdene: Did it at the last moment and forgot my due diligence :P. Addressed.

megatron/rl/rl_utils.py
yobibyte: After this is resolved. We need to add a test to avoid this bug in the future. 

I suggest we simulate the situation with equal sampling/consuming (e.g. sample 2, consume 2), and non-equal (sample 4, consume 2 for 2 steps).

For the above, we run prepare_data_for_update(), and check that the staleness behaves as we want it to behave.

megatron/rl/rl_utils.py
tdene: Addressed.

megatron/rl/rl_utils.py
tdene: Addressed; does it seem better this way?

megatron/rl/rl_utils.py
tdene: Then we're doing a list comprehension with `[x+0 for x in l]`, and that seemed uglier to me.

megatron/rl/rl_utils.py
tdene: Reverted; GitHub didn't auto-refresh the comment chain.

megatron/rl/rl_utils.py
yobibyte: You still have a staleness table above. Do we want a separate staleness table at all?

megatron/rl/server/api.py
tdene:
```
    def increment_staleness(self):
        if dist.get_rank() == 0:
            self._client.increment_staleness()
```
in `megatron/rl/inference/megatron.py`.

megatron/rl/agent/api.py
tdene: I think I've fixed it, but I want to do a test run to make sure.

megatron/rl/rl_utils.py
tdene: My bad, fixed

megatron/rl/rl_utils.py
```diff
@@ -811,6 +851,11 @@ def prep_wandb_metrics(
         advantages: Flattened list of advantages.
         tokenizer: Tokenizer to untokenize trajectories for logging.
         example_groups: A list of rollouts of one group to log examples of trajectories.
```
janEbert: Could you please move these lines below the new arguments and also switch the ordering between `tokenizer` and `example_groups`? That way, the order between arguments and the docstring matches again.

megatron/rl/agent/api.py
yobibyte: We need to test this in a unit test to avoid issues like that. I think you just need to add some checks to existing unit tests that do small checks on multiturn/singleturn

megatron/rl/rl_utils.py
tdene: You got it! Resolved

megatron/rl/agent/api.py
tdene: Updated the unit test

megatron/rl/rl_utils.py
tdene: This makes sense.    
But eventually we will want to duplicate this calculation: once accurately in `prep_wandb_metrics`, and once less accurately inside `get_environment_rollouts`, because the day will come when the degree of off-policyness is a factor that determines whether we actually use the rollout as-is or throw it out / penalize it somehow.

That day is not today though.

Addressed!

yobibyte: Good point! Even though the filtering might be applied on the agent level who will decide whether to send more requests for rollouts or not.

megatron/rl/agent/reward_only_agent.py
```diff
@@ -104,13 +104,21 @@ async def rollout_from_response(
                 generation_mask=[generation_mask],
                 env_id=self.env_id,
                 problem_id=golden['problem_id'] if 'problem_id' in golden else None,
+                policy_staleness=[response.policy_staleness],
```
yobibyte, line 107: DynamicInferenceRequest has Optional[torch.Tensor] type for policy_staleness. Could you, please, point me to a place where we convert an Optional[torch.Tensor] to the list[int] we process here?

tdene: Of course! It gets serialized by [L112 here](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/inference/inference_request.py#L112) into a tuple of `("tensor", list[int])`, and then gets unwound into just the list [on L124 here](https://github.com/NVIDIA/Megatron-LM/blob/0dba6a00178d53fe35a3260989a1ade77f9218c5/megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py#L124).

megatron/rl/rl_utils.py
```diff
@@ -758,6 +771,14 @@ def compute_group_stats(
             roll_turn_lens = [len(t) for t in rollout.trajectory]
             group_turn_lengths.extend(roll_turn_lens)
             group_traj_lengths.append(sum(roll_turn_lens))
+            group_policy_staleness.append(max(s for turn in rollout.policy_staleness for s in turn))
+            group_kv_staleness.append(max(s for turn in rollout.kv_cache_staleness for s in turn))
```
yobibyte: This logic is different from other numbers we plot. Not saying that we should do it different way, but can you explain why we want max here?

yobibyte: This is my last q for this PR.

tdene: Because the staleness is tracked at a per-token level, we have an extra dimension to the data. On each step we have a distribution of staleness across tokens across rollouts.

So, lacking better ideas, I just max'd. That represents the worst staleness of any tokens in the rollouts. That way every rollout can be associated with a simple integer.

It loses valuable information. But we don't have any present ideas for what to do with that information right now. Once we brainstorm we can do a follow-up PR with different metrics.

yobibyte: I understand the extra dimension. If I understand it right, you are maxing across all turns in the rollout. For turn lens, for example, we extend, so that you get all turn lens in the aggregate stats. Does it mean that we get mean over max? Not, just mean staleness when we plot it?

@ArEsKay3 was interested in staleness metrics, maybe he has some ideas.

tdene: You're absolutely right: mean over max. I am not creative enough right now to think of a better way.

If we have a 3-turn rollout, where the first turn has staleness-per-token of `[10, 10, 9, 8]`, the second turn has staleness-per-token of `[6, 6, 5, 5]`, and the third turn has staleness-per-token of `[0, 0, 0, 0]`, we report that the rollout as a whole had a staleness of 10. That's very misleading, I agree with you.

My main logic is that this is an adequate starting point, due to the fact that we don't really know what we will want until we start doing serious runs. And it's better to implement what we actually want in practice when we get to it, than it is to sit down today and brainstorm what we may want in the future.

Development style, that's all.

yobibyte: I think we either need to convert to extend as we do with other metrics (or maybe we keep the average and call it like avg_turn_staleness). Or we need to replace the name of the metric (average max or smth).

I think the former is more useful to have in wandb. Interested in @ArEsKay3 thoughts.

yobibyte: re dev style, I think it is better we spend a couple of more hours on that to get it right/useful from the very beginning and iterate from there. Otherwise, we get a ton of code that will need to be updated/retested real soon.

jon-barker: I think the correct approach here is to understand the requirement and build for that in a way that minimizes re-writing required later. As @tdene points out, we don't know all the metrics we will want for off-policyness studies yet. So we are going to need to extend this later.

I believe the current requirement is that we can compare our benchmarking runs to other RL frameworks at "iso-off-policyness". For other frameworks we are comparing to, they enforce a single off-policyness value per rollout - no franken rollouts (for single or multi-turn). In that case I think @tdene's current solution does work. So I'd be happy with @yobibyte's suggestion that we just improve the naming (and/or add an explanatory comment)so it's clear what is actually being logged.

yobibyte: Happy to go with the latter approach, even though the former allows you to plot both, max/mean/etc and is more consistent with the rest of the code.

tdene: Sounds good! Let's have a call before we merge.

I did convert to extend btw, let me know if it looks good.

megatron/rl/rl_utils.py
```diff
@@ -809,15 +839,32 @@ def prep_wandb_metrics(
...
         tokenizer: Tokenizer to untokenize trajectories for logging.
-        example_groups: A list of rollouts of one group to log examples of trajectories.
     """
 
     group_table = wandb_writer.Table(
         columns=['group_means', 'group_stds'],
         data=[[np.mean(g), np.std(g)] for g in rewards],
     )
 
+    # Expand per-turn completed_at_steps to per-token using turn_lens, then compute true staleness.
```
yobibyte: Thanks for adjusting the metrics, I think it looks much more consistent and useful for the end user now. My last ask would be to simplify these list comprehensions. I would rather have a separate function that is less dense, but more readable. If we could add a unit test for a couple of cases for this, it will also be great.

tdene: @yobibyte did this last commit get it right? Or did you mean that you wanted ALL these list comprehensions, like also the `mean_reward` one for example?

PR #3517, opened 2026-02-20
Implement forced lag in RL

tdene:
# What does this PR do ?

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/rl/agent/api.py
```diff
@@ -180,17 +180,23 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
                 group = await self.group_rollout(request=request)
                 if (
                     not request.filter_groups_with_same_reward
                     or np.std([r.reward for r in group]) > 1e-6
                 ):
+                    # Tag each rollout with its submission order.
+                    group_index = yielded_groups
+                    yielded_groups += 1
+                    for rollout in group:
+                        rollout.submission_index = group_index
```
yobibyte: nit: remove group index and do rollout.submission_index = yielded_groups -1?

megatron/rl/rl_utils.py
```diff
@@ -442,11 +449,12 @@ def get_inference_interface(args, loop, model):
 
 def get_rollout_generator(args, inference_interface, n_prompts, samples_per_group):
     global _ROLLOUT_GENERATOR
-    if not args.rl_partial_rollouts or _ROLLOUT_GENERATOR is None:
+    oversubscribed = args.rl_partial_rollouts or args.rl_forced_lag > 0
```
yobibyte: Could you explain why you are calling this 'oversubscribed'?

megatron/training/arguments.py
```diff
@@ -440,6 +440,9 @@ def validate_args(args, defaults={}):
         if args.rl_use_sequence_packing:
             assert args.micro_batch_size == 1, \
                 "micro_batch_size must be 1 when using sequence packing. To increase compute per micro batch increase the sequence length."
+        assert rl.forced_lag > 0 or not args.rl_partial_rollouts, (
+            "--rl-forced-lag and --rl-partial-rollouts are incompatible."
```
yobibyte: nice!

megatron/rl/rl_utils.py
```diff
@@ -1450,44 +1491,72 @@ def get_grpo_data_iterator(
...
+                    )
+                )
 
     # We collect new rollouts when we've gone over the collected data 'grpo_iterations' times.
     global_batches_per_collection = (grpo_prompts_per_step * grpo_group_size) // global_batch_size 
     if (
-        buffered_rollouts is None or
-        iteration == runtime_state.last_collection_iteration + 
+        runtime_state.data_iterator is None or
+        iteration >= runtime_state.last_collection_iteration +
```
yobibyte: Why >=?

megatron/rl/rl_utils.py
```diff
@@ -1450,44 +1491,72 @@ def get_grpo_data_iterator(
...
         global_batch_size: Global batch size.
         sequence_packing: Use sequence packing if True.
         is_correction: Use IS correction if True.
-        buffered_rollouts: Previously collected rollouts (if any)
 
     Returns:
         RerunDataIterator for the current training step
     """
     runtime_state = get_rl_runtime_state()
+    args = get_args()
```
yobibyte: Please, send the argument as a function arg, do not call get_args() here.

megatron/rl/rl_utils.py
```diff
@@ -1450,44 +1491,72 @@ def get_grpo_data_iterator(
...
     if (
-        buffered_rollouts is None or
-        iteration == runtime_state.last_collection_iteration + 
+        runtime_state.data_iterator is None or
+        iteration >= runtime_state.last_collection_iteration +
         (grpo_iterations * global_batches_per_collection)
     ):
+        if forced_lag > 0:
+            runtime_state.lag_buffer.append(
+                get_environment_rollouts(
```
yobibyte: This can be above the if forced_lag branch.

megatron/rl/rl_utils.py
```diff
@@ -1450,44 +1491,72 @@ def get_grpo_data_iterator(
...
+                    model, inference_model, optimizer, grpo_prompts_per_step, grpo_group_size,
+                )
+            )
+            rollouts = runtime_state.lag_buffer.popleft()
```
yobibyte: Will be useful/interesting to track the length of the lag buffer. Is it gonna increase in time?

yobibyte: Also! Related to the staleness tracking, without the plotting fix, the staleness will be reported incorrectly even more.

megatron/rl/rl_utils.py
```diff
@@ -1450,44 +1491,72 @@ def get_grpo_data_iterator(
...
+            rollouts=rollouts,
             tokenizer=get_tokenizer(),
             sequence_packing=sequence_packing,
             is_correction=is_correction,
-            )
+        )
+
         runtime_state.reset_iteration_counters(iteration)
+        return runtime_state.data_iterator
+    return runtime_state.data_iterator
```
yobibyte: With those changes, we need to run tests for these scenarios.
- Normal, GRPO_ITERATIONS=1 sample X, consume X (batch size = prompts * group_size)
- sample 2X, consume X batch size = prompts * group_size / 2
- your newly added lag thing. Not sure if we need it in both scenarios above.

megatron/rl/rl_utils.py
```diff
@@ -1450,44 +1491,72 @@ def get_grpo_data_iterator(
...
         RerunDataIterator for the current training step
     """
     runtime_state = get_rl_runtime_state()
+    args = get_args()
+    forced_lag = args.rl_forced_lag
+
+    if runtime_state.start_iteration is None:
+        runtime_state.start_iteration = iteration
+        if forced_lag > 0:
+            runtime_state.lag_buffer = deque()
```
yobibyte: For me this is kind of stuff `__init()__` is for. Why do we need all this logic here?

megatron/rl/rl_utils.py
tdene: I could not think of a better word at the time. It's regarding the concept of submitting more requests than you actually need.

yobibyte: Yeah, that's why I asked! partial rollouts or forced lag are not the only cases when we sample more than we consume. Sending a request when batch_size < group_size * prompts will also do this.

megatron/rl/agent/api.py
tdene: Done!

megatron/rl/rl_utils.py
tdene: It's now called `streaming`.

megatron/rl/rl_utils.py
tdene: I think this is now entirely stale code.

megatron/rl/rl_utils.py
tdene: Addressed!

megatron/rl/rl_utils.py
tdene: I think this is now entirely stale code.

megatron/rl/rl_utils.py
tdene: I think this is now entirely stale code.

megatron/rl/agent/api.py
```diff
@@ -38,6 +38,8 @@ class GroupedRolloutRequest(Request):
     inference_interface: InferenceInterface
     validation: bool = False
     filter_groups_with_same_reward: bool = False
+    streaming: bool = False
```
tdene, line 43: New parameter that matches up to what `num_groups = -1` meant in the old code. It's bad practice to have a parameter that means one thing when it's positive, and an entirely different thing when it's set to -1.

megatron/rl/agent/api.py
```diff
@@ -38,6 +38,8 @@ class GroupedRolloutRequest(Request):
...
+    batch_results: bool = False
```
tdene: When `False`, this returns groups at a time like in the old flow. When `True`, this waits until `num_groups` are ready and then returns them all at once.

megatron/rl/agent/api.py
```diff
@@ -182,32 +184,75 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+            maxsize=self.buffer_size if request.streaming else 0
         )
         submitted_groups = 0
 
+        if request.batch_results:
+            # Each worker gets a permit for a batch of num_groups groups.
+            # The semaphore ensures that each batch only starts after the previous is consumed.
+            groups_per_worker = request.num_groups
+            num_workers = self.parallel_generation_tasks // groups_per_worker
+            submission_gate = asyncio.Semaphore(num_workers)
```
tdene: The whole point of this `submission_gate` is the following.

If we are doing forced lag, we can force the RL setup to only "generate" a certain number of groups per step. This is the flow without `submission_gate`. This is more performant, but does not guarantee consistent lag.

The flow with `submission_gate` forces the RL setup to only "consume" a certain number of groups per step. This is less performant, but guarantees consistent lag.

The difference between "generate N per step" and "consume N per step" is in the edge-case.

megatron/rl/server/agent/fastapi_env_server.py
```diff
@@ -127,7 +127,7 @@ async def get_grouped_rollouts(
         assert isinstance(
             request.inference_interface, InferenceServer
         ), "Rollout requests to remote server must contain an InferenceServer object"
-        assert request.num_groups != -1, "FastAPIEnvServer does not support group rollout streaming"
+        assert not request.streaming, "FastAPIEnvServer does not support group rollout streaming"
```
tdene, line 130: This is what I meant above. `num_groups != -1` was secretly hidden behavior. Now it's out in the open.

megatron/rl/rl_utils.py
```diff
@@ -450,11 +450,17 @@ def get_inference_interface(args, loop, model):
 
 def get_rollout_generator(args, inference_interface, n_prompts, samples_per_group):
     global _ROLLOUT_GENERATOR
-    if not args.rl_partial_rollouts or _ROLLOUT_GENERATOR is None:
-        agent = get_agent(args, parallel_generation_tasks=args.rl_parallel_generation_tasks)
-        # Collect Rollouts
+    streaming = args.rl_partial_rollouts or args.rl_forced_lag > 0
+    if not streaming or _ROLLOUT_GENERATOR is None:
+        if args.rl_forced_lag > 0:
+            pgt = (args.rl_forced_lag + 1) * n_prompts
```
tdene: We must control the inference batch size in forced lag mode.

megatron/training/arguments.py
```diff
@@ -2219,6 +2226,11 @@ def _add_rl_args(parser):
                             'If False, CUDA graphs are deleted on suspend and re-captured on resume.')
     group.add_argument('--rl-partial-rollouts', action=argparse.BooleanOptionalAction, default=False,
                        help='If set, use partial rollouts.')
+    group.add_argument('--rl-forced-lag', type=int, default=0,
+                       help='Forced rollout lag of L steps. After an initial warm-up of L steps, '
+                            'All steps N+L use only rollouts that were started on step N. '
+                            '0 (default) disabled this behavior.'
```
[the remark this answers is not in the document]
tdene: Resolved. Thank you Claude!

megatron/rl/agent/api.py
```diff
@@ -182,32 +184,75 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
-                if (
-                    not request.filter_groups_with_same_reward
-                    or np.std([r.reward for r in group]) > 1e-6
-                ):
-                    await grouped_rollouts.put(group)
+            while request.streaming or submitted_groups < self.parallel_generation_tasks:
+                if submission_gate is not None:
+                    await submission_gate.acquire()
+                batch_id = submitted_groups // groups_per_worker
+                submitted_groups += groups_per_worker
```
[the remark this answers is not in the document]
tdene: Resolved. Thank you Claude!

ArEsKay3:
I actually have pretty strong feeling about the arguments here.
If I want to change something I shouldn't have to adjust lot of others.
As it stands a change in say group size will change the number of active requests (even if the global batch size is adjusted accordingly), something I think we want to avoid.

So. My proposal is
`--rl-num-parallel-generations` OR `--rl-num-parallel-generation-batches`

They would be mutually exclusive. Both would require `--rl-partial-rollouts` to be set.

The first would imply no batching and would represents the number of parallel rollouts. This is what we really want to control. We would internally adjust for multiple tasks and divide it by grpo_group_size to ensure that changing the grpo_group_size or number of tasks doesn't require an associated change in this argument to keep the same amount of "active" rollouts.

The second would imply batching and would represent the number of batches active at a time. In this case we would set our controls so we have that batches in flight. This would be our "forced lag" or "N step off policy" setup.

tdene:
> I actually have pretty strong feeling about the arguments here. If I want to change something I shouldn't have to adjust lot of others. As it stands a change in say group size will change the number of active requests (even if the global batch size is adjusted accordingly), something I think we want to avoid.
> 
> So. My proposal is `--rl-num-parallel-generations` OR `--rl-num-parallel-generation-batches`
> 
> They would be mutually exclusive. Both would require `--rl-partial-rollouts` to be set.
> 
> The first would imply no batching and would represents the number of parallel rollouts. This is what we really want to control. We would internally adjust for multiple tasks and divide it by grpo_group_size to ensure that changing the grpo_group_size or number of tasks doesn't require an associated change in this argument to keep the same amount of "active" rollouts.
> 
> The second would imply batching and would represent the number of batches active at a time. In this case we would set our controls so we have that batches in flight. This would be our "forced lag" or "N step off policy" setup.

Hey Robert, you're right. I undid my experiment from this morning. Please see changes.
    
    

In summary, I'm aiming to be able to do the following. Let's call `--grpo-prompts-per-step` P and the desired lag L.

To turn on PPO-like generation, set `--grpo-group-size 1`. Any value greater than 1 is GRPO-like generation.

All of the below work for either PPO or GRPO:

- To keep things lag-free, set `--no-rl-partial-rollouts`. Otherwise, set `--rl-partial-rollouts`.
- For unbounded lag, set `--rl-num-parallel-generations math.inf`.
- For bounded, but not forced, lag, set `--rl-num-parallel-generations $(((L + 1) * P))`.
- For forced lag, set `--rl-num-parallel-generation-batches $((L + 1))`.

`--grpo-prompts-per-step` is a misnomer and should be called `--rl-prompts-per-step`, but I don't really want to pile that into this PR.

megatron/rl/agent/api.py
```diff
@@ -182,32 +184,67 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+        tasks = [asyncio.create_task(generate_task()) for _ in range(num_workers)]
 
         try:
+            next_batch_id = 0
+            pending: dict[int, list[list[Rollout]]] = {}
             while grouped_rollouts.qsize() > 0 or not all(task.done() for task in tasks):
-                yield await grouped_rollouts.get()
+                group = await grouped_rollouts.get()
+                pending.setdefault(group[0].batch_id, []).append(group)
+                if len(pending.get(next_batch_id, [])) >= groups_per_worker:
```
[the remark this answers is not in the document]
tdene: Thank you very much Claude, that was an important bug you found! Fixed.

tdene:
New summary after addressing reviewer comments:

Let P = `--grpo-prompts-per-step`, G = `--grpo-group-size`, L = desired lag.

To turn on PPO-like generation, set `--grpo-group-size 1`. Any value greater than 1 is GRPO-like generation.

All of the below work for either PPO or GRPO:

- Lag-free: `--no-rl-partial-rollouts` (default).
- Unbounded lag: `--rl-partial-rollouts` (default parallelism).
- Bounded, but not forced, lag: `--rl-num-parallel-generations $(((L + 1) * P * G))` (rollout units).
- Forced lag: `--rl-num-parallel-generation-batches $((L + 1))` (batch units).
- Arg only for power-users: `--rl-generation-batch-size N` (rollout units) overrides the default of `generation batch size == training batch size`.

megatron/training/arguments.py
```diff
@@ -406,14 +406,66 @@ def validate_args(args, defaults={}):
...
+                args.rl_num_parallel_generations // args.grpo_group_size)
+            if args.rl_generation_batch_size is None:
+                args.rl_generation_batch_size = 1
+        elif args.rl_num_parallel_generation_batches is not None:
+            assert args.rl_partial_rollouts, \
+                "--rl-num-parallel-generation-batches requires --rl-partial-rollouts."
+            if args.rl_generation_batch_size is None:
+                args.rl_generation_batch_size = args.grpo_prompts_per_step
+            args.rl_parallel_generation_tasks = (
+                args.rl_num_parallel_generation_batches * args.rl_generation_batch_size)
```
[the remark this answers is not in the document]
tdene: Thank you for catching that, Claude!

megatron/rl/agent/api.py
```diff
@@ -182,32 +185,76 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+            maxsize=self.buffer_size if request.streaming else 0
         )
         submitted_groups = 0
 
+        # generation_batch_size controls how many groups each worker generates and yields together.
+        # When it's set to 1, the semaphore is a no-op.
+        groups_per_worker = request.generation_batch_size
+        if groups_per_worker > 1:
+            assert not request.filter_groups_with_same_reward, \
+                "Filtering is not supported with generation_batch_size > 1"
```
yobibyte: Let's make this more actionable for the end user. Smth like, you cannot use request.filter_groups_with_same_reward with generation_batch_size > 1.

megatron/training/arguments.py
```diff
@@ -2190,9 +2233,24 @@ def _add_rl_args(parser):
                         'that evaluate pass@k so we default to a lower number.')
     # TODO(rkirby): allow for "complete" evaluation when --rl-prompts-per-eval is set to -1
     group.add_argument('--grpo-prompts-per-step', type=int, default=32,
-                       help="Number of GRPO groups (G in the paper).")
+                       help="Number of prompts per training data collection step.")
```
tdene: This is an artifact from a previous version; needs to be reverted.

megatron/rl/agent/api.py
```diff
@@ -182,32 +185,76 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+        assert self.parallel_generation_tasks >= groups_per_worker, \
+            f"parallel_generation_tasks ({self.parallel_generation_tasks}) must be >= " \
```
yobibyte: Nice one! No need to change, but might be easier to use debug expressions like 
```python
print(f"{var=}")
```

megatron/rl/agent/api.py
```diff
@@ -182,32 +185,76 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+            f"generation_batch_size ({groups_per_worker})"
+        num_workers = self.parallel_generation_tasks // groups_per_worker
```
yobibyte, line 225: What happens if they do not divide? Do we need an assert for that?

megatron/rl/agent/api.py
```diff
@@ -182,32 +185,76 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+        submission_gate = asyncio.Semaphore(num_workers)
+
+        async def generate_and_enqueue(batch_id, index_in_batch):
+            group = await self.group_rollout(request=request)
+            if (
+                not request.filter_groups_with_same_reward
+                or np.std([r.reward for r in group]) > 1e-6
```
yobibyte, line 239: Why do we do the standard deviation check here?

megatron/rl/agent/weighted_multi_task.py
```diff
@@ -182,33 +182,32 @@ async def get_reward_rollouts(self, request: RolloutRequest) -> list[Rollout]:
...
+        agent_pgts = self._distribute_counts(self.parallel_generation_tasks)
+        agent_slots = self._distribute_counts(request.num_groups, distribute_remainder=False)
         agent_slots = np.array(agent_slots) / np.gcd.reduce(agent_slots)
 
         # Create tasks for each agent with non-zero groups
-        generators: list[AsyncGenerator[list[Rollout], None]] = []
-        for agent, num_groups in zip(self.agents, agent_groups):
-            if num_groups != 0:
+        generators = []
+        for agent, num_groups, pgt in zip(self.agents, agent_groups, agent_pgts):
```
yobibyte: add strict=True to zip pls

megatron/rl/rl_utils.py
```diff
@@ -455,11 +450,13 @@ def get_inference_interface(args, loop, model):
 
 def get_rollout_generator(args, inference_interface, n_prompts, samples_per_group):
     global _ROLLOUT_GENERATOR
-    if not args.rl_partial_rollouts or _ROLLOUT_GENERATOR is None:
+    streaming = args.rl_partial_rollouts
+    if not streaming or _ROLLOUT_GENERATOR is None:
```
yobibyte: You missed such an opportunity to use the walrus operator!

https://peps.python.org/pep-0572/

megatron/rl/agent/api.py
yobibyte: Those will break nemo_gym integration. Please, adjust those scripts accordingly. ./nemo_gym_agent.py is the file you need.

megatron/training/arguments.py
```diff
@@ -2190,9 +2233,24 @@ def _add_rl_args(parser):
...
     group.add_argument('--grpo-group-size', type=int, default=2,
                        help="Number of samples per a GRPO group.")
+    group.add_argument('--rl-num-parallel-generations', type=int, default=None,
+                       help='Number of parallel rollouts in flight (no batching). '
```
yobibyte: I do not understand this help message. We have to make this more clear.

megatron/training/arguments.py
```diff
@@ -2190,9 +2233,24 @@ def _add_rl_args(parser):
...
+                            'Internally divided by grpo_group_size. '
+                            'Requires --rl-partial-rollouts. '
+                            'Mutually exclusive with --rl-num-parallel-generation-batches.')
+    group.add_argument('--rl-num-parallel-generation-batches', type=int, default=None,
+                       help='Number of generation batches in flight (lag = N - 1). '
```
yobibyte: lag = N - 1 is unclear, what is N?

megatron/training/arguments.py
```diff
@@ -2190,9 +2233,24 @@ def _add_rl_args(parser):
...
+                            'Each batch contains grpo_prompts_per_step groups by default. '
+                            'Requires --rl-partial-rollouts. '
+                            'Mutually exclusive with --rl-num-parallel-generations.')
+    group.add_argument('--rl-generation-batch-size', type=int, default=None,
+                       help='Override the number of rollouts per generation batch. '
+                            'Internally divided by grpo_group_size. '
```
yobibyte: Won't it be better to have a flag named smth like rl-generation-batch-size-per-group, and multiply by the grpo_group_size internally? 

I'm not asking to change this, just trying to think what will be easier to use for the end-user.

megatron/training/arguments.py
```diff
@@ -2227,7 +2285,7 @@ def _add_rl_args(parser):
                        help='Persist CUDA graphs when the inference engine is suspended. '
                             'If False, CUDA graphs are deleted on suspend and re-captured on resume.')
     group.add_argument('--rl-partial-rollouts', action=argparse.BooleanOptionalAction, default=False,
-                       help='If set, use partial rollouts.')
+                       help='If set, use partial rollouts (persistent streaming generator across iterations).')
```
yobibyte: I would have a paragraph in the docstring of partial rollouts (maybe it's already there?) and point the user to this docstring. 'Persistent streaming generator across iteration' is not very clear for the user who has no idea what the concept is. I can understand it, but I have more knowledge of the internals.

megatron/training/arguments.py
```diff
@@ -2300,8 +2358,8 @@ def _add_rl_args(parser):
                        help='If set, verify that the model weights were correctly transferred by comparing forward pass outputs on'
                        'the first swap of model weights.')
 
-    group.add_argument('--rl-parallel-generation-tasks', type=int, default=512,
-                        help='Number of parallel generation tasks for RL inference.')
+    group.add_argument('--rl-parallel-generation-tasks', type=int, default=None,
+                       help='Deprecated: use --rl-num-parallel-generations instead.')
```
yobibyte, line 2404: `megatron-rl/tests/functional_tests/test_cases/moe/gpt_grpo_tp8tp4_pp1_ep8ep2_dp8_throughputtest/model_config.yaml` test has this flag set to one. I do not think there is any other config in main using this flag. Should we just change it in the functional test and get rid of the arg now while nobody is using it?

megatron/rl/rl_utils.py
```diff
@@ -455,11 +450,13 @@ def get_inference_interface(args, loop, model):
...
         agent = get_agent(args, parallel_generation_tasks=args.rl_parallel_generation_tasks)
-        # Collect Rollouts
         request = GroupedRolloutRequest(
-            num_groups=-1 if args.rl_partial_rollouts else n_prompts,
+            num_groups=n_prompts,
```
yobibyte: You will need to adjust GroupedRolloutRequest in `mrl_extras/test/run_request.py` too.

yobibyte (changes requested):
Hey Teo, thanks for adding this!

My main concern with this PR is that it adds some logic / modifies the existing one, but it has zero unit tests. 

We will need to add some tests to make sure the system behaves as you expect it too before we merge this.

megatron/rl/agent/api.py
tdene: Will do this later today.

megatron/rl/agent/api.py
tdene: Resolved

megatron/rl/agent/api.py
tdene: Resolved

megatron/rl/agent/api.py
tdene: Resolved with a warning print

megatron/rl/agent/api.py
tdene: This standard deviation check is in `main` as well at basically the same spot. I don't agree that it should be here. But it's out-of-scope for this PR to fix that.

megatron/rl/agent/weighted_multi_task.py
tdene: Resolved

megatron/rl/rl_utils.py
tdene: Resolved :)

megatron/training/arguments.py
tdene: How does the new help string look?

megatron/training/arguments.py
tdene: How does the new help string look?

megatron/training/arguments.py
tdene: Done, except I didn't change the argument name because it's a new argument. I just specified in the help string what units this argument uses, and changed the logic elsewhere.

megatron/training/arguments.py
tdene: Hm. How do you feel about the change I just made, where I added a paragraph to the help string here? I'm not convinced this is that difficult of a concept. See https://xkcd.com/2501/

megatron/training/arguments.py
tdene: I changed it in the functional test, but I do not want to remove it altogether because I am pretty sure users are actively using this flag in their CLI arguments.

megatron/rl/rl_utils.py
tdene: Will do this later today.

tdene:
> Hey Teo, thanks for adding this!
> 
> My main concern with this PR is that it adds some logic / modifies the existing one, but it has zero unit tests.
> 
> We will need to add some tests to make sure the system behaves as you expect it too before we merge this.

Added unit test!

megatron/rl/agent/api.py
yobibyte: Let's get back to this after we deal with `completed_at_step` parameter.

tests/unit_tests/rl/test_grouped_rollouts.py
```diff
@@ -0,0 +1,111 @@
...
+from megatron.rl.inference import ReturnsRaw
+
+
+class MockGenerator(GroupedRolloutGenerator):
+    """Mock generator with configurable per-call delays."""
+
+    def __init__(self, env_id="test", slow_first=0, **kwargs):
+        super().__init__(**kwargs)
+        self.env_id = env_id
+        self.slow_first = slow_first
```
yobibyte: Does this mean first_fast_index? If yes, please rename. Otherwise, add a comment.

tests/unit_tests/rl/test_grouped_rollouts.py
```diff
@@ -0,0 +1,111 @@
...
+            num_groups=4,
+            rollouts_per_group=1,
+            inference_interface=MagicMock(spec=ReturnsRaw),
+            streaming=False,
+            generation_batch_size=1,
+            enforce_order=False,
+        )
+        groups = []
+        async for group in mt.get_grouped_rollouts(request):
+            groups.append(group)
```
yobibyte, line 110: Are we testing that the ordering of the envs is the same as the order we submitted here?

megatron/rl/agent/weighted_multi_task.py
```diff
@@ -199,7 +202,7 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
                 agent_request = GroupedRolloutRequest(
                     num_groups=num_groups,
                     streaming=request.streaming,
-                    generation_batch_size=request.generation_batch_size,
+                    generation_batch_size=num_groups,
```
yobibyte: It looks weird that we send the same value twice for different input arguments

megatron/training/arguments.py
tdene: Resolved

tests/unit_tests/rl/test_grouped_rollouts.py
tdene: Renamed

tests/unit_tests/rl/test_grouped_rollouts.py
tdene: Yes

megatron/rl/agent/api.py
```diff
@@ -182,33 +187,94 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+            ):
+                for rollout in group:
+                    rollout.submission_index = batch_id * groups_per_worker + index_in_batch
```
yobibyte: Had a slack discussion on this property. Trying to understand whether we need both the batch_id and submission_index or we can simplify. 

We later do:
```python
batch = pending.pop(next_batch_id)
batch.sort(key=lambda g: g[0].submission_index)
```
Which seems to prove my point. @tdene to think about it.

megatron/rl/agent/weighted_multi_task.py
yobibyte: bumping

megatron/training/arguments.py
yobibyte: I would say this is making our lives harder, but up to you.

megatron/rl/agent/api.py
```diff
@@ -182,33 +187,94 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+            await asyncio.gather(*tasks)
+            grouped_rollouts.shutdown()
+
+        shutdown_task = asyncio.create_task(shutdown_queue_when_done())
 
         try:
-            while grouped_rollouts.qsize() > 0 or not all(task.done() for task in tasks):
-                yield await grouped_rollouts.get()
+            next_batch_id = 0
+            pending: dict[int, list[list[Rollout]]] = {}
```
yobibyte: We should prob make a type alias for the group, will be much easier to read IMO

megatron/rl/agent/api.py
```diff
@@ -182,33 +187,94 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+            while True:
+                try:
+                    group = await grouped_rollouts.get()
+                except asyncio_QueueShutDown:
+                    break
+                if request.enforce_order:
+                    # Accumulate groups and enforce submission order across batches.
+                    pending.setdefault(group[0].batch_id, []).append(group)
```
yobibyte: I think it finally clicked for me. From here, it seems a group is not only a list of rollouts, but also some metadata on top. I'm not sure about perf implications, but readability-wise, it would be great to have a Group dataclass that will keep rollouts, as well as other metadata as batch_id, and submission_id. 

With that, we could clearly indicate what functions operate on groups, and it will be so much nicer without keeping extra fields in the Rollouts that do not belong there.

megatron/rl/agent/api.py
```diff
@@ -182,33 +187,94 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+                    while len(pending.get(next_batch_id, [])) >= groups_per_worker:
+                        batch = pending.pop(next_batch_id)
+                        batch.sort(key=lambda g: g[0].index_in_batch)
+                        next_batch_id += 1
+                        for g in batch:
+                            yield g
+                        submission_gate.release()
+                else:
+                    # Yield in completion order, no HOL blocking.
```
yobibyte: Replace the acronym with the actual name.

yobibyte (commented):
Yo. I think we need to refactor Rollout/Groups hierarchy. 

I have mixed feelings of whether it has to be done in this PR or not. The reason I think it should be here is that we add some stuff into Rollouts that do not belong there, and it is not nice. Adding metadata to a newly added Group dataclass would be much more readable and will make more sense.

megatron/rl/agent/weighted_multi_task.py
tdene: Thanks for the bump, I had missed this.

megatron/rl/agent/api.py
tdene: I agree! Done.

megatron/rl/agent/api.py
tdene: Addressed

megatron/rl/agent/weighted_multi_task.py
tdene: Resolved!

megatron/rl/agent/api.py
tdene: Addressed by the `RolloutGroup` refactor

megatron/rl/agent/api.py
tdene: Addressed by the `RolloutGroup` refactor

megatron/rl/agent/api.py
```diff
@@ -54,6 +59,24 @@ class Rollout(AgentBaseModel):
     num_evictions: list[int]
 
 
+@dataclass(slots=True)
+class RolloutGroup:
```
yobibyte: Should it inherit from the BaseModel?

megatron/rl/agent/api.py
```diff
@@ -182,33 +205,93 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+            await asyncio.gather(*tasks)
+            grouped_rollouts.shutdown()
+
+        shutdown_task = asyncio.create_task(shutdown_queue_when_done())
 
         try:
-            while grouped_rollouts.qsize() > 0 or not all(task.done() for task in tasks):
-                yield await grouped_rollouts.get()
+            next_batch_id = 0
+            pending: dict[int, list[RolloutGroup]] = {}
```
yobibyte: GroupedRollouts?

megatron/rl/agent/api.py
```diff
@@ -54,6 +59,24 @@ class Rollout(AgentBaseModel):
     num_evictions: list[int]
 
 
+@dataclass(slots=True)
+class RolloutGroup:
+    """A group of rollouts (e.g. multiple completions for one prompt) with batch metadata."""
+
+    rollouts: list[Rollout]
```
yobibyte: Rollout or Rolout|TokenRollout?

megatron/rl/agent/api.py
```diff
@@ -182,33 +205,93 @@ async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
...
+            pending: dict[int, list[RolloutGroup]] = {}
+            while True:
+                try:
+                    group = await grouped_rollouts.get()
+                except asyncio_QueueShutDown:
+                    break
+                if request.enforce_order:
+                    # Accumulate groups and enforce submission order across batches.
+                    pending.setdefault(group.batch_id, []).append(group)
+                    while len(pending.get(next_batch_id, [])) >= groups_per_worker:
```
yobibyte: Trying to understand this logic. Is there a chance we do not get to this loop? i.e. not get out of it because we have exhausted pending, but for some other reasons.

tdene: Certainly. The only way we can get stuck here is if we filter out rollouts (which we are asserting that we do not do), or if the worker tasks die. But if the worker tasks die, we shutdown the asyncio Queues (because I pack-ported Python 3.13 functionality into this repo, because asyncio Queues are broken before Python 3.13), and that breaks us out of a potentially infinite loop.

megatron/rl/agent/api.py
tdene: Done

megatron/rl/agent/api.py
tdene: Done

megatron/rl/agent/api.py
tdene: Done

yobibyte (approved):

jaredcasper (approved):

asolergi-nv (approved):

PR #3524, opened 2026-02-20
add --overlap-param-gather support for layer-wise optimizer. lots of unit tests.

mchrzanowski:
# What does this PR do ?

1) Adds --overlap-param-gather functionality to muon.
Builds on top of [@deyuf's PR](https://github.com/NVIDIA/Megatron-LM/pull/2787/changes?w=1)

2) Lots and lots of unit tests for synchronization and for the layer wise optimizer.

Passes functional tests here: https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/merge_requests/4555

### Pre-checks

### DIFF
--- a/megatron/core/distributed/param_and_grad_buffer.py
+++ b/megatron/core/distributed/param_and_grad_buffer.py
@@ -144,6 +144,7 @@ class _LayerwiseAllGatherHandle:
         self.handles = handles
 
     def wait(self):
+        """Wait on the last handle and clear all handles."""
         if self.handles:
             self.handles[-1].wait()
         self.handles = None


megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -236,7 +236,10 @@ def _allocate_buffers_for_parameters(
 
             # Set `next_param_gather_bucket_group` for different bucket groups by iterating through
             # buckets in reverse order (since all-gathers happen in reverse order of buckets).
-            if self.ddp_config.use_distributed_optimizer and self.ddp_config.overlap_param_gather:
```
mkhona-nvidia, line 239: How would we consider cases where users specify overlap-param-gather and AdamW, but no distributed optimizer?

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -344,9 +347,10 @@ def unmap_weight_tensor(m):
                     grad_acc.register_hook(self._make_backward_post_hook(param))
                     self.grad_accs.append(grad_acc)
 
-        self.use_forward_hook = (
```
mkhona-nvidia, line 347: same as above

megatron/core/distributed/distributed_data_parallel.py
mchrzanowski: it's not possible. you must use the dist opt with Adam to use this feature.

deepakn94: Without distributed optimizer, we do gradient all-reduces (there is no parameter all-gather).

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -317,7 +439,9 @@ def finish_param_sync(self, skip_next_bucket_dispatch: bool = False):
             skip_next_bucket_dispatch (bool, optional): if true, dispatch next
                 bucket's communication if available.
         """
-        assert self.ddp_config.use_distributed_optimizer
+        # overlap_param_gather covers the layer-wise optimizer case, which sets
+        # overlap_param_gather=True without use_distributed_optimizer.
+        assert self.ddp_config.use_distributed_optimizer or self.ddp_config.overlap_param_gather
         assert self.ddp_config.overlap_param_gather
```
janEbert, line 442: The check for `self.ddp_config.use_distributed_optimizer` is redundant here due to the existing assertion for `self.ddp_config.overlap_param_gather` below it. So the line added by this PR could be removed. (Or should the existing line checking just for `self.ddp_config.overlap_param_gather` be removed?)

mchrzanowski: muon doesnt use the (AdamW) distributed optimizer, so you still have to check for either here

megatron/training/arguments.py
```diff
@@ -1303,10 +1304,6 @@ def validate_args(args, defaults={}):
     # Muon optimizer check
     if 'muon' in args.optimizer:
 
-        # TODO: remove these checks once we support them
```
BoxiangW, line 1421: They are still not supported if using muon not dist_muon, maybe add a silent warning? Non-distributed is not being used much

mkhona-nvidia: Let's instead try to make `--use-distributed-optimizer` work with `--optimizer muon` and remove 2 `--optimizer {muon, dist-muon}` for Muon. This will likely involve some logic in `post_init` of arguments or `OptimizerConfig`

@skyw

mchrzanowski: sg I added them back

mchrzanowski: that can happen in another PR, this one is already so hairy

megatron/core/distributed/param_and_grad_buffer.py
janEbert: But you're doing `assert self.overlap_param_gather` right below the `assert self.ddp_config.use_distributed_optimizer or self.ddp_config.overlap_param_gather`.
So even if the first assertion passes with `self.ddp_config.use_distributed_optimizer = True` and `self.ddp_config.overlap_param_gather = False`, the second assertion fails anyway.

mchrzanowski: oh im sorry, I misunderstood, thank you. fixed

megatron/training/arguments.py
BoxiangW: Agree, we can actually build that on top of the kimi poc PR but that actually sounds more confusing. So after that change, there will be `local_muon`, `dist_opt_muon`, and `layer_wise_muon`

mkhona-nvidia: No, I mean remove all of them and have just 2 settings:

`--optimizer muon
`
`-use-distributed-optimizer`

If we don't use the second flag, we will have `local_muon`, if we do, we will have Mike's muon

BoxiangW: That makes more sense

mchrzanowski:
I ran nano v3 for 12B tokens on this branch and main; the relative lm loss differences look reasonable, as do the rest of the statistics.

Figure_1

Screenshot 2026-02-23 at 22 29 08

skyw:
> I ran nano v3 for 12B tokens on this branch and main; the relative lm loss differences look reasonable, as do the rest of the statistics.

Does the deterministic path work with Muon? If yes, under deterministic mode they should match exactly, only communication order change.

mchrzanowski:
I believe determinism still doesnt work with mamba models

mkhona-nvidia:
> I believe determinism still doesnt work with mamba models

@mchrzanowski it does, @paul-gibbons worked to enable it

megatron/training/arguments.py
FDecaYed: Agree with this, and this was the original design. decided to kept muon/layerwise separate since it is experimental and don't want to risk breaking dist-opt related code
and since we already have user now, need to document it well or have some back compat deprecation/

megatron/training/training.py
```diff
@@ -2257,6 +2257,13 @@ def save_checkpoint_and_time(
     one_logger_utils.track_e2e_metrics()
     if should_disable_forward_pre_hook(args):
         force_param_sync(model)
+    # Free overlap param-gather buffers and release cached GPU memory so
+    # that the async checkpoint worker process has enough GPU headroom for
+    # D2H tensor transfers.
+    for model_chunk in model:
+        if hasattr(model_chunk, 'free_overlap_buffers'):
+            model_chunk.free_overlap_buffers()
```
FDecaYed, line 2271: Added a separate comment about sorting out the memory issue and whether we need this explicit free. But assume we need this, i feel it is still best to make it called automatically when param sync finishes
(or at least stuck this one into force_param_sync)

megatron/core/optimizer/layer_wise_optimizer.py
```diff
@@ -143,34 +156,87 @@ def shard_params(self, optimizers):
...
+            # All ranks must participate in every broadcast call (collective).
+            for i in range(dp_size):
+                if flat_sizes[i] == 0:
+                    continue
+                src_global = torch.distributed.get_global_rank(group, i)
+                if i == rank:
+                    buf = src
+                else:
+                    buf = recv_buffers[i]
+                torch.distributed.broadcast(buf, src_global, group=group)
```
FDecaYed: I think for both synced(here) and async gather, we need to sort out what is the root cause of memory spike before just switch to broadcast. Here is my questions:
- I'm under impression that allgather_v internally handle uneven size with [broadcast](https://github.com/pytorch/pytorch/blob/1a4d3ed0f536993c22c1bf42115edb42cf947fc6/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L4890), so switch to bcast here might be only saving the local buffer?
- if this code does reduce spike, then there is chance coalescing makes difference(hard to believe since in/out are pre-allocated)? But in that case we also want to understanding the perf implications(esp for async case)
- in what case are we seeing OOM for async case? I feel per bucket extra usage in parallel to fprop should be relatively safe and we also need enough perf so overlap can hide it. I feel it is better to rule out the extra flatten buffer(and life scope of it) before commit to use bcast here

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -273,33 +314,123 @@ def start_param_sync(self, force_sync: bool = False):
...
+            group = self.intra_distributed_optimizer_instance_group
+            lw_work_handles = []
+            for bucket in self.buckets:
+                # Use param dtype (e.g., bf16), NOT grad dtype (which may be fp32
+                # when grad_reduce_in_fp32 is enabled). All ranks must use the same
+                # dtype for broadcast buffers, and ranks with local params will have
+                # bf16 src tensors from _flatten_dense_tensors.
+                param_dtype = bucket.params_list[0].dtype
+                src = (
+                    _flatten_dense_tensors(bucket.lw_params_list[local_rank])
```
FDecaYed: This is certainly an extra buffer. I left some comment on the broadcast change mentioning this, and we should definitely optimize this too.
If we adjust param order(or adjust param split base on their bucket order), then we can force params already flattened and contiguous by rank. This allow us not only remove this flatten(better speed and mem), but also enable potential reduce-scatter over current allreduce+drop.

Same tech is already mentioned in the new Qwen paper. we could do it in another PR

megatron/core/optimizer/layer_wise_optimizer.py
juntaowww: Hi, I am also analyzing the distributed muon, and maybe can provide some inputs. Based on profiling, the all gather in main branch indeed calls `ncclBroadcast` as the sizes are uneven and this becomes a bottleneck.

For the OOM issues, I would suspect that is due to the ping-pong sharding behavior. For real workload such as kimi-k2, we could observe the following sharding pattern:

For example with 4 ranks, and given sorted param lists with sizes `[10, 10, 100, 1000, 1000, 10000]` (as some params are just bias terms and some are huge matrices), the sharding after ping-pong assignment would be
`[[10], [10], [100, 10000], [1000, 1000]]`, leading to extremely uneven sizes. Especially in large scale where the number of parameters are not at least twice as large the ranks, the rank with small index number would only hold a single tiny parameter.

To mitigate the problem though may not completely solve the it, I want to propose doing a simple greedy load balance algorithm such as bin packing to make the sharded sizes more uniform and manually padding the buffer to allow more efficient `ncclAllGather` being called.

megatron/training/training.py
mchrzanowski:
```
● There's a catch — free_overlap_buffers needs to run regardless of whether force_param_sync is called. Look at lines 2258-2265:

  if should_disable_forward_pre_hook(args):    # conditional
      force_param_sync(model)
  # Free overlap buffers...                     # unconditional
  for model_chunk in model:
      if hasattr(model_chunk, 'free_overlap_buffers'):
          model_chunk.free_overlap_buffers()

  force_param_sync is gated behind should_disable_forward_pre_hook, but free_overlap_buffers runs unconditionally. So if you folded it into force_param_sync, you'd lose the buffer freeing in the case
  where the forward pre-hook isn't disabled.

  You could move it into force_param_sync and then also call force_param_sync unconditionally, but force_param_sync does start_param_sync(force_sync=True) which you presumably don't want to run
  unconditionally — it would change behavior when the forward pre-hook is active.

  A cleaner option might be to add a separate helper like free_overlap_buffers(model) that save_checkpoint_and_time calls, keeping the two concerns (param sync vs. buffer freeing) separate. But it's
  really just moving the loop into a function for aesthetics — the current code is already pretty clear.
```

mchrzanowski:
OK guys I have put the all gather back; thanks to @skyw for pointing out that claude was just confused and that all-gather absolutely should be able to run on variable sized inputs. It continues to not OOM when it used to.

mchrzanowski:
I couldn't get mamba to be deterministic so I ran with a transformer moe for 3.5B tokens. loss, grad norm, and num zeroes match perfectly

Figure_1

megatron/training/training.py
FDecaYed: From this change it is only called here and only with force sync. so I was thinking to put it there and not explicitly modify training loop code. The real fix is free as soon as buffer is not needed and automatically call somewhere, regardless whether there is force_sync and logging needs.

though I think the issue is minor and ok to merge as it is now. we can revisit this later

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -112,6 +113,38 @@ def __init__(
             global_start, global_end, _ = param_index_map[param]
             self.param_to_index[param] = (global_start - offset, global_end - offset)
 
+        # Layer-wise optimizer attributes for async param gather.
+        self.lw_params_list = None
```
deepakn94: Let's re-name `lw` to `layerwise`.

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -112,6 +113,38 @@ def __init__(
...
+        self.lw_param_flat_sizes = None
+        self.lw_gather_list = None
+        self._lw_src_buffer = None
+
+    def set_lw_params_list(self, lw_params_list: List[List[torch.nn.Parameter]]):
+        """Set per-rank parameter lists for layer-wise async all-gather.
```
deepakn94, line 123: Expand on what "parameter lists" means? List of parameters covered by a layerwise-sharded optimizer?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -112,6 +113,38 @@ def __init__(
...
+    NCCL guarantees in-order completion on the same communicator, so waiting
+    on only the last handle is sufficient.
+    """
+
+    def __init__(self, handles):
+        self.handles = handles
+
+    def wait(self):
+        if self.handles:
+            self.handles[-1].wait()
```
deepakn94, line 149: Clear `self.handles` afterwards?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -273,33 +310,112 @@ def start_param_sync(self, force_sync: bool = False):
...
-        # Coalesce communication kernels across buckets in the bucket group.
-        with _coalescing_manager(
-            self.intra_distributed_optimizer_instance_group, async_ops=async_op
-        ) as cm:
-            for idx, bucket in enumerate(self.buckets):
-                if self.cached_param_buffer_shard_list[idx] is None:
-                    self.cached_param_buffer_shard_list[idx] = shard_buffer(
-                        bucket.param_data, self.intra_distributed_optimizer_instance_size
+
+        if not self.ddp_config.use_distributed_optimizer:
```
deepakn94, line 318: I don't love `not self.ddp_config.use_distributed_optimizer` being re-interpreted as essentially `self.ddp_config.use_layerwise_distributed_optimizer`. Can we re-factor the config arguments here to make it more intuitive?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -273,33 +310,112 @@ def start_param_sync(self, force_sync: bool = False):
...
+                    bucket.lw_gather_list = None
+                    continue
+
+                # Flatten local params.  Detach from the autograd graph because
+                # start_param_sync can be called during the forward pass (where
+                # autograd is active) and all_gather will write into gather_list
+                # entries in-place.
+                local_size = bucket.lw_param_flat_sizes[local_rank]
+                if local_size > 0:
+                    src = _flatten_dense_tensors(bucket.lw_params_list[local_rank]).detach()
```
deepakn94: Better variable name?

megatron/core/distributed/param_and_grad_buffer.py
skyw: refactoring is coming https://github.com/NVIDIA/Megatron-LM/pull/3638.

deepakn94: Ok, will resolve this for now and look at Deyu's refactor.

deepakn94:
Throughputs look good on a contrived case with high DP communication overhead (8B Transformer with TP=8).

```
### Logfile: 8b_adam_nooverlap_101106_date_26-03-04_time_13-16-15.log
---------------------------------------------------------------------------
| Metric               | Throughput (TFLOP/s)   | Time per Iter (ms)     |
---------------------------------------------------------------------------
| Iterations Recorded  | 57                     | 57                     |
| Min                  | 183.60                 | 282.00                 |
| Max                  | 187.20                 | 287.40                 |
| Mean                 | 185.64                 | 284.30                 |
| Median               | 185.70                 | 284.20                 |
---------------------------------------------------------------------------


### Logfile: 8b_adam_overlap_101105_date_26-03-04_time_13-16-04.log
---------------------------------------------------------------------------
| Metric               | Throughput (TFLOP/s)   | Time per Iter (ms)     |
---------------------------------------------------------------------------
| Iterations Recorded  | 63                     | 63                     |
| Min                  | 199.60                 | 255.80                 |
| Max                  | 206.30                 | 264.40                 |
| Mean                 | 203.87                 | 258.89                 |
| Median               | 204.10                 | 258.60                 |
---------------------------------------------------------------------------


### Logfile: 8b_muon_nooverlap_101124_date_26-03-04_time_13-23-41.log
---------------------------------------------------------------------------
| Metric               | Throughput (TFLOP/s)   | Time per Iter (ms)     |
---------------------------------------------------------------------------
| Iterations Recorded  | 44                     | 44                     |
| Min                  | 142.70                 | 365.10                 |
| Max                  | 144.50                 | 369.80                 |
| Mean                 | 143.83                 | 366.95                 |
| Median               | 143.90                 | 366.75                 |
---------------------------------------------------------------------------


### Logfile: 8b_muon_overlap_101123_date_26-03-04_time_13-23-35.log
---------------------------------------------------------------------------
| Metric               | Throughput (TFLOP/s)   | Time per Iter (ms)     |
---------------------------------------------------------------------------
| Iterations Recorded  | 49                     | 49                     |
| Min                  | 158.50                 | 324.70                 |
| Max                  | 162.50                 | 333.00                 |
| Mean                 | 160.59                 | 328.66                 |
| Median               | 160.50                 | 328.90                 |
---------------------------------------------------------------------------
```

deepakn94:
/ok to test [079e2a1](https://github.com/NVIDIA/Megatron-LM/pull/3524/commits/079e2a1ed0b9044c5761e21d258b8f240df63a09)

deepakn94:
/ok to test [64dd7ea](https://github.com/NVIDIA/Megatron-LM/pull/3524/commits/64dd7eaa96fba0ecab4a1966036efbe63f64234f)

mchrzanowski:
/ok to test [62b5ae2](https://github.com/NVIDIA/Megatron-LM/pull/3524/commits/62b5ae24071f4048f399a390a41e1e271c275a42)

PR #3535, opened 2026-02-23
Introduce GDN to Mamba

Phlip79:
# What does this PR do ?

Adding Gated Delta Net (GDN) feature to MambaModel. GDN was added to GPTModel in #1989.

Adding unit tests. Also, tested end-to-end model using this Megatron-Bridge PR: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2520.

wandb: [Baseline](https://wandb.ai/nvidia-nemo-framework/ppetrakian/runs/idd0o5zi?nw=nwuserppetrakian) (w/ GPTModel) vs [MambaModel](https://wandb.ai/nvidia-nemo-framework/ppetrakian/runs/ba2i9pvk?nw=nwuserppetrakian)

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


yuzhongw-nvidia (approved):
LGTM. Thanks.

megatron/core/ssm/mamba_hybrid_layer_allocation.py
```diff
@@ -290,17 +292,24 @@ def allocate_layers(
...
+            logging.INFO,
+            f"{actual_gdn_layers_count} gdn layers in " f"{total_layers_count} total layers.",
+        )
     return layer_type_list
 
 
 def get_layer_maps_from_layer_type_list(
     layer_type_list: List[str],
-) -> Tuple[Dict[int, int], Dict[int, int], Dict[int, int]]:
+) -> Tuple[Dict[int, int], Dict[int, int], Dict[int, int], Dict[int, int], Dict[int, int]]:
```
janEbert, line 473: Hey, could we apply the same thing to this method [as what I mentioned here](https://github.com/NVIDIA/Megatron-LM/pull/3377#discussion_r2851920950)? Would be super nice to avoid having to unroll a giant tuple in the future. Also simplifies the typing here.

janEbert (approved):

megatron/core/ssm/mamba_hybrid_layer_allocation.py
```diff
@@ -30,8 +30,9 @@ class Symbols:
     ATTENTION = "*"
     MLP = "-"
     MOE = 'E'
+    GDN = 'G'
```
duncanriach: Wondering if GDN should be ordered after Mamba and before attention, to keep things that are similar together. Comment applies elsewhere in this file and others.

duncanriach:
Looking good.

Please merge after [3377](https://github.com/NVIDIA/Megatron-LM/pull/3377), which is very close to being merged.

megatron/core/ssm/mamba_block.py
```diff
@@ -150,6 +151,16 @@ def __init__(
...
                     )
+                elif layer_type == LayerSymbols.GDN:
+                    layer = build_module(
+                        submodules.gdn_layer,
+                        config=self.config,
+                        layer_number=layer_number,
+                        pg_collection=pg_collection,
+                        is_mtp_layer=is_mtp_layer,
+                        add_layer_offset=False,
+                        pp_layer_offset=pp_layer_offset,
```
duncanriach: Since this is not an attention layer, with no k-v cache, I don't think it's going to need pp_layer_offset to be passed in.

megatron/core/ssm/mamba_block.py
```diff
@@ -150,6 +151,16 @@ def __init__(
...
                         add_layer_offset=False,
                     )
+                elif layer_type == LayerSymbols.GDN:
+                    layer = build_module(
+                        submodules.gdn_layer,
+                        config=self.config,
+                        layer_number=layer_number,
+                        pg_collection=pg_collection,
+                        is_mtp_layer=is_mtp_layer,
+                        add_layer_offset=False,
```
duncanriach, line 161: Looking at the current GDN implementation in main, it doesn't seem to add its own pp_layer_offset like the attention layers do. I wonder if this is because it doesn't currently support pipeline parallel via the GPTModel. If it's not adding its own pp_layer_offset, then it doesn't need to be told not to when the HLModel passes a layer_number containing the pp_layer_offset it already calculated.

duncanriach (commented):
These are the things that stand-out from the review of your latest changes. I get the sense that there are other things that will need to be added to fully support GDN, such as the calculation of flops in training.py

Just realized I had some comments that were pending from yesterday that I forgot to submit. Submitted with this review cycle.

megatron/core/ssm/mamba_hybrid_layer_allocation.py
duncanriach, line 147: I think this function needs to be updated, and where it's called from

duncanriach:
Seems like there should be an update in arguments.py, if only to the help info for --hybrid-layer-pattern. Worried the scope of these changes is too narrow. Would be good to ask a coding agent to look for all the places in the code-base where the addition of this layer type might require changes.

megatron/core/ssm/mamba_block.py
yuzhongw-nvidia: pp_layer_offset is not needed for GDN (for now).

megatron/core/ssm/mamba_hybrid_layer_allocation.py
janEbert: FYI I've changed this in my fork; I will integrate your changes once they're merged, so you don't need to worry about it. :)

megatron/core/ssm/mamba_block.py
Phlip79: Removed.

megatron/core/ssm/mamba_hybrid_layer_allocation.py
Phlip79: Moved.

megatron/core/ssm/mamba_hybrid_layer_allocation.py
Phlip79: Why does it need to be updated? "G" is considered a layer currently. And what's wrong with where it's currently called from?

Phlip79:
> Seems like there should be an update in arguments.py, if only to the help info for --hybrid-layer-pattern. Worried the scope of these changes is too narrow. Would be good to ask a coding agent to look for all the places in the code-base where the addition of this layer type might require changes.

I've updated `arguments.py` to explicitly mention GDN in the `--hybrid-layer-pattern` definition.

megatron/core/models/mamba/mamba_layer_specs.py
```diff
@@ -117,6 +118,20 @@
                 pre_mlp_layernorm=TENorm, mlp=moe, mlp_bda=get_bias_dropout_add
             ),
         ),
+        gdn_layer=ModuleSpec(
```
duncanriach: Should move up under mamba_layer

megatron/core/ssm/mamba_block.py
duncanriach: This code should be moved up under mamba.

Sounds like add_layer_offset is not needed, because the GDN layer does not currently add its own pp_layer_offset. So perhaps add_layer_offset should not be included here.

megatron/core/ssm/mamba_hybrid_layer_allocation.py
duncanriach: Doesn't this function need to have 'G' added to its return dictionary?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -301,14 +301,14 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
             # For hybrid models, the layer map converts the global layer index to the
             # corresponding attention layer index or Mamba layer index depending on the
             # layer type.
-            attention_layer_map, mamba_layer_map, _, _ = get_layer_maps_from_layer_type_list(
-                mamba_inference_state_config.layer_type_list
+            attention_layer_map, mamba_layer_map, _, _, gdn_layer_map = (
+                get_layer_maps_from_layer_type_list(mamba_inference_state_config.layer_type_list)
```
duncanriach: Shouldn't the position of gdn_layer_map come after mamba_layer_map?

megatron/core/models/mamba/mamba_layer_specs.py
Phlip79: Done.

megatron/core/ssm/mamba_hybrid_layer_allocation.py
Phlip79: You're right, I completely missed that.

megatron/core/inference/contexts/dynamic_context.py
Phlip79: Updated here and in the function `get_layer_maps_from_layer_type_list`.

Phlip79:
> So perhaps add_layer_offset should not be included here.

I've removed `add_layer_offset` and `is_mtp_layer` as neither are used.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -312,14 +312,14 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
...
-                mamba_inference_state_config.layer_type_list
+            mamba_layer_map, gdn_layer_map, attention_layer_map, _, _ = (
+                get_layer_maps_from_layer_type_list(mamba_inference_state_config.layer_type_list)
             )
             self.num_attention_layers = len(attention_layer_map)
             self.num_mamba_layers = len(mamba_layer_map)
             self.mamba_conv_states_shape = mamba_conv_states_shape
             self.mamba_ssm_states_shape = mamba_ssm_states_shape
-            self.layer_map = attention_layer_map | mamba_layer_map
+            self.layer_map = attention_layer_map | mamba_layer_map | gdn_layer_map
```
santhnm2: If inference is not yet supported with GDN, can we add an explicit `NotImplementedError` if `len(gdn_layer_map) > 0`?

santhnm2 (changes requested):

megatron/core/inference/contexts/dynamic_context.py
Phlip79: Great point. PTAL.

megatron/core/ssm/mamba_block.py
Phlip79: I believe we should be setting `add_layer_offset=False` so that the `TransformerLayer` does not calculate additional layers (see [this code](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/transformer_layer.py#L296)).

santhnm2 (approved):

ericharper (approved):

megatron/core/models/mamba/mamba_layer_specs.py
```diff
@@ -143,6 +158,20 @@
                 mamba_bda=get_bias_dropout_add,
             ),
         ),
+        gdn_layer=ModuleSpec(
```
jaredcasper: In dynamic_context.py we assert that GDN isn't supported, so does it make sense to put a gdn_layer spec here?

Phlip79: That's a good catch, thank you. Removed.

maanug-nv (approved):

jaredcasper (approved):

jaredcasper:
Can we add a functional test or two for GDN? Or do we need more to do an end-to-end run?

PR #3537, opened 2026-02-23
fix: skip non-tensor optimizer state entries in distrib_optimizer sav…

ahmadki:
# What does this PR do ?
This is a fix for the TE compatibility issue with the precision-aware optimizer.

`_get_main_param_and_optimizer_states` (save path) iterates all keys in the optimizer state including non-tensor entries (like found_inf: bool). The save succeeds because `get_unscaled_state returns` the bool as-is. But on load, `_set_main_param_and_optimizer_states` passes that same bool to set_scaled_state, which assumes it's a tensor.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/optimizer/distrib_optimizer.py
+++ b/megatron/core/optimizer/distrib_optimizer.py
@@ -896,6 +896,8 @@ class DistributedOptimizer(MixedPrecisionOptimizer):
             sharded_model_param = self.optimizer.param_groups[group_index]["params"][group_order]
             tensors = {}
             for k in self.optimizer.state[sharded_model_param]:
+                if not isinstance(self.optimizer.state[sharded_model_param][k], torch.Tensor):
+                    continue
                 if isinstance(self.optimizer, HybridDeviceOptimizer):
                     tensors[k] = self.optimizer.state[sharded_model_param][k]
                     continue
@@ -905,7 +907,10 @@ class DistributedOptimizer(MixedPrecisionOptimizer):
         else:
             main_param = self.optimizer.param_groups[group_index]["params"][group_order]
             optim_state = self.optimizer.state[main_param]
-            tensors = {"param": main_param, **optim_state}
+            tensors = {"param": main_param}
+            for k, v in optim_state.items():
+                if isinstance(v, torch.Tensor):
+                    tensors[k] = v
         return tensors
 
     def _set_main_param_and_optimizer_states(self, model_param, tensors):
@@ -922,6 +927,8 @@ class DistributedOptimizer(MixedPrecisionOptimizer):
         if self.config.use_precision_aware_optimizer_no_fp8_or_ds_fp8:
             sharded_model_param = self.optimizer.param_groups[group_index]["params"][group_order]
             for k, v in tensors.items():
+                if not isinstance(v, torch.Tensor):
+                    continue
                 if isinstance(self.optimizer, HybridDeviceOptimizer):
                     if k == "param":
                         k = "master_param"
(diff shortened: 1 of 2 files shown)


megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -935,8 +942,13 @@ def _set_main_param_and_optimizer_states(self, model_param, tensors):
         else:
             main_param = self.optimizer.param_groups[group_index]["params"][group_order]
             optim_state = self.optimizer.state[main_param]
-            dst_tensors = {"param": main_param, **optim_state}
+            dst_tensors = {"param": main_param}
+            for k, v in optim_state.items():
+                if isinstance(v, torch.Tensor):
+                    dst_tensors[k] = v
             for key in dst_tensors:
+                if not isinstance(tensors.get(key), torch.Tensor):
```
janEbert: Why not index without a check here?

janEbert:
This semantically includes some of the changes in #3521, but handles more error cases and adds test. I'd thus add this instead of #3521 for the `distrib_optimizer` stuff, but keep the backward-compatibility tokenizer addition from #3521.

janEbert (approved):

megatron/core/optimizer/distrib_optimizer.py
ahmadki: It depends on the behavior we want to implement.

Is it legal for `optim_state` to include non-tensor keys ? Keeping in mind that the save path excludes them ?
We have three options:

1. Explicit failure in mcore (raise an exception instead of the current implementation).
2. mcore doesn't care, TE would fail or raise an exception in case of keys.
3. mcore doesn't care, and we don't want TE failure, so we exclude non-tensor keys in mcore

Thinking more about it, option 2 is probably the best here (removing the current check), let me know what you think.

janEbert: Sorry, I wasn't clear at all. I meant why don't you do `tensors[key]` instead of `tensors.get(key)`.

deepakn94 (approved):

megatron/core/optimizer/distrib_optimizer.py
ahmadki: My rational was that on missing keys, `tensor.get(key)` would return a None and the check will skip it. `tensor[key]` would raise a missing key exception. 
I looked at how pytorch optim handles similar cases and an error on missing keys is the safer option, I updated the PR to address this.

ahmadki:
/ok to test https://github.com/NVIDIA/Megatron-LM/commit/6e1d6fa643bc52a67fb94e8c57275ed81e63abb3

megatron/core/optimizer/distrib_optimizer.py
janEbert: Got it, thank you for the explanation and update!

ahmadki:
/ok to test https://github.com/NVIDIA/Megatron-LM/commit/9a0a3549831973627ab519ab98654903887ee0b8

ahmadki:
/ok to test https://github.com/NVIDIA/Megatron-LM/commit/f2d903ff714234b8aba73a0aa09e5d5c35c5bb4f

Issue #2294, opened 2025-11-19
IndexError in activation checkpointing: uniform split generates invalid layer indices


sdjasj:
**Describe the bug**
When enabling activation recomputation with the following configuration:
```
recompute_granularity="full",
recompute_method="uniform",
recompute_num_layers=7,
num_layers=24,
```
Megatron-core attempts to access transformer layer indices beyond the valid range (0–23), resulting in:
```
IndexError: index 24 is out of range
```
This indicates that the uniform splitting logic does not properly clamp the last block's layer indices when num_layers is not divisible by recompute_num_layers.

**Steps/Code to reproduce bug**

1. Build a GPT model (or any Transformer) with num_layers=24.
2. Enable activation recomputation using:
```
recompute_granularity="full"
recompute_method="uniform"
recompute_num_layers=7
```
3. Run a forward-backward pass.

Below is a minimal reproducible `main.py`:

```
import os
import torch
from megatron.core import parallel_state
from megatron.core.distributed.finalize_model_grads import finalize_model_grads


def initialize_distributed(tensor_model_parallel_size = 1, pipeline_model_parallel_size = 1):
    # Torch setup for distributed training
    rank = int(os.environ['LOCAL_RANK'])
    world_size = torch.cuda.device_count()
    torch.cuda.set_device(rank)
    torch.distributed.init_process_group(world_size=world_size, rank=rank)
    # Megatron core distributed training initialization
    parallel_state.initialize_model_parallel(tensor_model_parallel_size, pipeline_model_parallel_size)

from megatron.core.transformer.transformer_config import TransformerConfig
from megatron.core.models.gpt.gpt_model import GPTModel
from megatron.core.models.gpt.gpt_layer_specs import get_gpt_layer_local_spec
def model_provider():
    """Build the model."""

    transformer_config = TransformerConfig(
        recompute_granularity="full",
        recompute_method="uniform",
        recompute_num_layers=7,
        num_layers=24,
        hidden_size=12,
        num_attention_heads=4,
        use_cpu_initialization=True,
        pipeline_dtype=torch.float32)

    gpt_model = GPTModel(
        config=transformer_config,
        transformer_layer_spec=get_gpt_layer_local_spec(),
        vocab_size=100,
        max_sequence_length=64)

    return gpt_model


import torch
from torch.utils.data import DataLoader

from megatron.core.datasets.blended_megatron_dataset_builder import BlendedMegatronDatasetBuilder
[...]

PR #3562, opened 2026-02-24
Fix IndexError in uniform activation recompute when num_layers not divisible by recompute_num_layers

saakshigupta2002:
## Problem

When using `recompute_method="uniform"` with a `num_layers` value that is **not evenly divisible** by `recompute_num_layers`, the `_checkpointed_forward` method in `TransformerBlock` raises an `IndexError` due to out-of-bounds layer access.

### Reproduction

For example, with `num_layers=24` and `recompute_num_layers=7`:

- Chunk 1: layers 0–6 (7 layers)
- Chunk 2: layers 7–13 (7 layers)
- Chunk 3: layers 14–20 (7 layers)
- **Chunk 4: layers 21–27** — only indices 0–23 exist, so accessing layers 24–27 crashes with `IndexError`

```python
TransformerConfig(
    num_layers=24,
    recompute_granularity='full',
    recompute_method='uniform',
    recompute_num_layers=7,
    ...
)
```

```
  File ".../megatron/core/transformer/transformer_block.py", line 558, in _checkpointed_forward
    custom(layer_idx, layer_idx + self.config.recompute_num_layers)
  ...
IndexError: list index out of range
```

This happens because the `custom()` closure iterates `range(start, end)` where `end` exceeds the actual number of layers on the last chunk.

## Root Cause

In `_checkpointed_forward`, the uniform recompute path passes `layer_idx + self.config.recompute_num_layers` directly to the `custom()` function without clamping the end index to `self.num_layers_per_pipeline_rank`. Interestingly, the feature extraction loop just below the buggy line **already applies the correct clamping** via `min()`:

```python
chunk_end = min(
    layer_idx + self.config.recompute_num_layers, self.num_layers_per_pipeline_rank
)
```

The `custom()` call simply missed this same bounds check.

## Fix

Compute `chunk_end` **once** using `min()` before both the `custom()` call and the feature extraction loop, eliminating the code duplication and fixing the out-of-bounds access:

```python
while layer_idx < self.num_layers_per_pipeline_rank:
    chunk_end = min(
        layer_idx + self.config.recompute_num_layers,
        self.num_layers_per_pipeline_rank,
    )
    hidden_states, context = checkpoint_handler(
        custom(layer_idx, chunk_end)
    )
    # feature extraction loop already uses chunk_end correctly
    for idx in range(layer_idx, chunk_end):
        ...
```

The last chunk will simply have fewer layers than the others (e.g., 3 instead of 7), which is functionally correct and expected when the layer count is not evenly divisible.

## Changes

| File | Description |
|------|-------------|
| `megatron/core/transformer/transformer_block.py` | Clamp end index in uniform recompute `custom()` call using `min()` |
| `tests/unit_tests/transformer/test_transformer_block.py` | Add regression test with `num_layers=5`, `recompute_num_layers=3` |

## Testing

- Added `test_gpu_forward_uniform_checkpoint_non_divisible_layers` which creates a `TransformerBlock` with `num_layers=5` and `recompute_num_layers=3` (chunks: [0,1,2] and [3,4]), runs a forward pass, and verifies the output shape is correct. Without the fix, this test crashes with `IndexError`.
- Existing uniform recompute tests continue to pass (evenly divisible case is unchanged).
- The fix is minimal and surgical — only the `custom()` call's end index computation is affected. The `min()` pattern is the exact same idiom already used for the feature extraction loop in the same code block.

## Impact

- **Who is affected**: Anyone using `recompute_method="uniform"` where their layer count is not evenly divisible by `recompute_num_layers` (e.g., 24 layers with recompute_num_layers=7, or 32 layers with recompute_num_layers=5).
- **Severity**: This is a hard crash (`IndexError`) that prevents training from starting.
- **Risk**: Minimal. The change is a single `min()` clamp using the same pattern already present in the surrounding code. The last chunk having fewer layers is the expected and correct behavior.

Fixes https://github.com/NVIDIA/Megatron-LM/issues/2294


janEbert:
Please apply `tools/autoformat.sh`.

saakshigupta2002:
> Please apply `tools/autoformat.sh`.

Done! Thanks for the heads up! :)

PR #3563, opened 2026-02-24
Robust signaling for coordinator inference

tdene:
# What does this PR do ?
### State Diagram

```
          +------>  RUNNING  <------+
          |         +--+--+         | 
          |            |            |
       UNPAUSE      PAUSE          |
      (broadcast)  (idempotent)    |
          |            |            |
          |            v            |
          +-------  PAUSING        |
          |         +--+--+        |
          |            |           |
          |      EP all-reduce     |
          |       then world       |
          |       barrier          |
          |            |           |
          |            v           |
          +-------  PAUSED --------+
          |         +--+--+--+
          |            |     |     
          |        SUSPEND  STOP --------+
          |            |                 |
          |            v                 |
          |     SUSPENDING               |
          |         |                    |
          |    world barrier             |
          |         |                    |
          |         v                    v
          |     SUSPENDED ----STOP---> STOPPING
          |         |                    |
          |       RESUME            world barrier
          |         |                    |
          |         v                    v
          |      RESUMING              STOPPED
          |         |                  teardown, exit
          |    world barrier
          |         |
          +----  PAUSED
              (then UNPAUSE)  
```

### PAUSE Protocol (details)

```
Client           Coordinator         mp_src (engine)      mp_workers
  |--PAUSE---------->|                    |                   |
  |                  | state = PAUSED     |                   |
  |                  |---PAUSE----------->|                   |
  |                  |---PAUSE----------->| (all dp_ranks)    |
  |                  |                    |--PUB PAUSE------->|
  |                  |                    |                   |
  |                  |                    | state = PAUSING   |
  |                  |                    | EP all-reduce:    |
  |                  |                    |   report 0        |
  |                  |                    |   dummy forward   |
  |                  |                    |   until consensus |
  |                  |                    |                   |
  |                  |                    | world barrier     |
  |                  |                    | (all ranks sync)  |
  |                  |                    |                   |
  |                  |                    | state = PAUSED    |
  |                  |                    | paused.set()      |
```

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/inference/text_generation_server/dynamic_text_gen_server/flask_server.py
+++ b/megatron/core/inference/text_generation_server/dynamic_text_gen_server/flask_server.py
@@ -100,7 +100,7 @@ async def run_flask_server(
     """Initializes and runs the async Flask server
     starting an InferenceClient with the provided coordinator address."""
     inference_client = InferenceClient(coordinator_addr)
-    await inference_client.start()
+    inference_client.start()
     logger.info(f"Rank {rank}: InferenceClient connected.")
     try:
         await run_flask_server_on_client(inference_client, tokenizer, flask_port, parsers, verbose)
--- a/tests/unit_tests/inference/test_data_parallel_inference_coordinator.py
+++ b/tests/unit_tests/inference/test_data_parallel_inference_coordinator.py
@@ -72,6 +72,7 @@ class DummyContext:
 
     def __init__(self):
         self.active_cnt = 0
+        self.step_count = 0
 
     def get_active_request_count(self) -> int:
         return self.active_cnt
@@ -113,7 +114,6 @@ class DummyEngine(DynamicInferenceEngine):
 
         self.step_start_event = unittest.mock.MagicMock()
         self.step_end_event = unittest.mock.MagicMock()
-        self.step_count = 0
 
     async def run_engine_with_coordinator(self, *, loop=None):
         """Override to bypass @trace_async_exceptions for testability.
--- a/tests/unit_tests/inference/test_dynamic_prefix_caching_coordinator.py
+++ b/tests/unit_tests/inference/test_dynamic_prefix_caching_coordinator.py
@@ -443,7 +443,7 @@ class TestCoordinatorEndToEnd:
         try:
             if torch.distributed.get_rank() == 0:
                 client = InferenceClient(dp_addr)
-                await client.start()
+                client.start()
 
                 futures = [
                     client.add_request(prompt=prompt, sampling_params=params)


janEbert:
Hey, this still saw a lot of active work after opening it. Is it still in a WIP state? If yes, could you mark it as a draft, please? :)

megatron/core/inference/data_parallel_inference_coordinator.py
```diff
@@ -169,7 +177,39 @@ def get_next_data_parallel_rank(self):
         Returns:
             bytes: The ZMQ identity of the next data parallel rank to receive a request.
         """
-        return next(self.data_parallel_rank_iterator)
+        identities = self.identities_of_data_parallel_ranks
+        if not identities:
+            raise RuntimeError("No engines connected")
+        idx = self._round_robin_idx % len(identities)
+        self._round_robin_idx = idx + 1
+        return identities[idx]
```
tdene, line 212: Doing this because the list of engines connected to the coordinator is not static, so we can't build the iterator ahead of time.

megatron/core/inference/headers.py
```diff
@@ -13,17 +13,29 @@ class Headers(Enum):
     SUBMIT_REQUEST = auto()
     ENGINE_REPLY = auto()
     PAUSE = auto()
-    PAUSE_ACK = auto()
```
tdene, line 16: Remove the ACK logic altogether as it is very flaky.

megatron/core/inference/inference_client.py
```diff
@@ -73,8 +71,6 @@ def __init__(self, inference_coordinator_address: str):
 
         self._loop = None
         self.running = asyncio.Event()
-        self.paused = asyncio.Event()
-        self.stopped = asyncio.Event()
```
tdene, line 77: None of these are actually useful. They were added when the ACK logic would propagate readiness of engine up into the client. But that's very flaky.

megatron/core/inference/inference_client.py
```diff
@@ -165,19 +157,17 @@ def _connect_with_inference_coordinator(self):
         reply = msgpack.unpackb(self.socket.recv(), raw=False)[0]
         assert Headers(reply) == Headers.CONNECT_ACK
 
-    async def start(self, loop: Optional[asyncio.AbstractEventLoop] = None):
```
tdene, line 168: Making this async serves no purpose. `_connect_with_inference_coordinator` already blocks until the appropriate time.

tdene:
> Hey, this still saw a lot of active work after opening it. Is it still in a WIP state? If yes, could you mark it as a draft, please? :)

Hey @janEbert, sorry about the notification spam. I originally pushed a version that was good... and then I proceeded to mentally thrash over some race conditions, undoing and redoing the same code over and over.

It's all good now. I am running this in live prod code again, unit-tests pass locally and they should work through CI as well, and I'm running a pipeline so I can update the functional golden values. There should be no more changes other than addressing reviewer comments.

janEbert:
Haha got it, congrats on the resolution. :D

megatron/core/inference/inference_client.py
sidsingh-nvidia: ah nice, the inference client is simple again. Love this

examples/inference/gpt/gpt_dynamic_inference_with_coordinator.py
```diff
@@ -172,14 +172,18 @@ async def main(
...
 
-        # kill the engines and suspend the client
-        # Right now, we can only call stop when all requests are done.
-        # Todo: Make this explicit in the Client class....
-        await client.stop_engines()
-        client.stop()
+        # Pause before stopping: STOP requires PAUSED or SUSPENDED state.
+        client.pause_engines()
+
+    await engine.paused.wait()
```
sidsingh-nvidia: Did we change from stop_engines() to pause()=>stop() because it's mandatory to do this now? Or because you want these signals being tested in CI?

sidsingh-nvidia: `await engine.paused.wait()` - why are we directly querying the engine on the client side? For non-colocated inference, we will not have access to the engine.

megatron/core/inference/engines/async_zmq_communicator.py
```diff
@@ -102,9 +102,45 @@ async def all_reduce_max(self, local_val: int) -> int:
                 except zmq.Again:
                     await asyncio.sleep(0.001)  # Yield to event loop
 
+    async def all_reduce_max_pair(self, val_a: int, val_b: int) -> tuple[int, int]:
```
sidsingh-nvidia: Thanks for this! I plan to use this to all-reduce multiple integers in the future, by folding in the allreduce in  batch_dimension_utils into this call. If you have the bandwidth, can you generalize this to multiple integers please? Totally optional, but it would help me a lot.

megatron/core/inference/engines/async_zmq_communicator.py
```diff
@@ -102,9 +102,45 @@ async def all_reduce_max(self, local_val: int) -> int:
...
+                except zmq.Again:
+                    await asyncio.sleep(0.001)
+
     def close(self):
         """
         Close the ZMQ sockets.
         """
-        self.gather_sock.close()
-        self.bcast_sock.close()
+        self.gather_sock.close(linger=0)
```
sidsingh-nvidia, line 117: what is linger? Can you add a comment explaining this?

examples/inference/gpt/gpt_dynamic_inference_with_coordinator.py
sidsingh-nvidia: Ah actually nevermind. Even in non-colocated inference, the inference world of GPUs will 1. ofcourse have access to the engine and 2. will need to explicitly check for change of state before they update weights / resume inference etc. So this is cool.

sidsingh-nvidia: Okay, how about we create a method in the engine called wait_until_paused, which calls await self.paused.wait? Could be generalize to engine.wait_until(signal_enum) as well.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -248,12 +262,13 @@ def reset(self) -> None:
         self._loop = get_asyncio_loop(getattr(self, "_loop", None))
         self._cond = asyncio.Condition()
         self.running = asyncio.Event()
+        self.running.set()
         self.paused = asyncio.Event()
+        self.suspended = asyncio.Event()
         self.stopped = asyncio.Event()
```
sidsingh-nvidia: Are these events there to detect change of state? If so, can we add a comment here explaining that. Otherwise, it's easy to confuse them with self.state

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -399,7 +414,7 @@ async def start_listening_to_data_parallel_coordinator(
             "pip install msgpack"
         )
 
-        self.zmq_context = zmq.Context().instance()
```
sidsingh-nvidia, line 424: reason for this change?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -505,6 +520,10 @@ async def start_listening_to_data_parallel_coordinator(
         self.model_parallel_num_msgs_subscriber_socket = self.zmq_context.socket(zmq.SUB)
         self.model_parallel_num_msgs_subscriber_socket.connect(mp_len_addr)
         self.model_parallel_num_msgs_subscriber_socket.setsockopt_string(zmq.SUBSCRIBE, "")
+        # Set a receive timeout so that blocking recv() calls are interruptible.
+        # Without this, non-MP-coordinator ranks block forever in schedule_requests()
+        # and their engine_loop_task can never be cancelled by asyncio.
```
sidsingh-nvidia: I am confused by this comment. Shouldn't the engine_loop_task eventually end itself gracefully after receiving STOP? Why do we need timeouts?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -521,6 +540,14 @@ async def start_listening_to_data_parallel_coordinator(
                 self.zmq_context, process_group=self.pg_collection.ep
             )
 
+        # initialize zmq-based world communicator for consensus barriers
+        total_world_size = torch.distributed.get_world_size()
+        if total_world_size > 1:
+            world_group = torch.distributed.new_group()
+            self.world_zmq_communicator = AsyncZMQCommunicator(
+                self.zmq_context, process_group=world_group
```
sidsingh-nvidia: Nit: you can pass None as the process_group

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1583,77 +1621,140 @@ def schedule_requests(self) -> int:
                 all_messages = []
 
         range_pop()
-        for message in all_messages:
+        # We need to put the messages in a queue so that we can break after control signals.
+        self._pending_signals.extend(all_messages)
```
sidsingh-nvidia: The variable is titled pending_signals, but we seem to be pushing in requests as well?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1583,77 +1621,140 @@ def schedule_requests(self) -> int:
...
-                self.paused.set()
-                self.received_pause = False
-            elif header == Headers.STOP_ACK:
-                self.stopped.set()
-                self.received_stop = False
+                if self.state == EngineState.RUNNING:
+                    self.state = EngineState.PAUSING
+                    self.running.clear()
+                # Any other state can safely ignore PAUSE.
+                break
```
sidsingh-nvidia: Behavior on main - even if a pause is received, we still add all of the submitted requests to the waiting queue. 

Behavior on this branch - we do not add these requests, and they now sit in a different queue? 
So now we have two waiting queues - seems unnecessary, unless I am missing something. I feel we should still call add_requests on all of the received requests. This way we can continue doing more work until all ranks have achieved consensus on the signal.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1583,77 +1621,140 @@ def schedule_requests(self) -> int:
...
+
             elif header == Headers.INCREMENT_STALENESS:
                 waiting = set(self.waiting_request_ids)
                 for request_id, entry in self.requests.items():
                     entry.record.increment_staleness(policy_only=request_id in waiting)
+
             elif header == Headers.STOP:
-                self.received_stop = True
+                assert self.state in (
```
sidsingh-nvidia, line 1815: So we now require, either pause or suspended to be issues before we call stop?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1583,77 +1621,140 @@ def schedule_requests(self) -> int:
...
-        if it exists, and destroys the model parallel state.
-        This method ensures that any running inference coordinator subprocess
-        is properly terminated, and cleans up resources associated with
-        model parallelism.
+    async def shutdown(self, immediate: bool = False):
+        """Shut down the engine and clean up ZMQ resources.
+
+        When immediate=False (called from the engine loop's finally block), just does ZMQ cleanup.
+
+        When immediate=True (external emergency shutdown), cancels the engine loop task directly.
```
sidsingh-nvidia: When does this case arise? Is this why we are inserting timeouts in the tensor parallel sockets?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1677,118 +1778,126 @@ async def run_engine(self, *, loop: Optional[asyncio.AbstractEventLoop] = None):
...
-            # it will pass a non-zero value to the all-reduce, and hence the global work will be non-zero,
-            # and we will defer processing the signal.
-            # When all ranks receive the signal, global work will be zero, and we can process the signal safely.
-            local_work = 0
+        range_push("_ep_establish_consensus")
+
+        consensus_val = -1 if signal_consensus else 0
 
         if self.ep_world_size > 1:
-            # Perform all-reduce to get max global work across EP group.
```
sidsingh-nvidia, line 1836: Can we retain this comment?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1677,118 +1778,126 @@ async def run_engine(self, *, loop: Optional[asyncio.AbstractEventLoop] = None):
...
+                        else:
+                            # Dummy forward to participate in the EP collective.
+                            self.step_start_event.record()
+                            self.controller.dummy_forward()
+                            self.step_end_event.record()
+                            self.step_end_event.synchronize()
+                            self.step_count += 1
+                    elif all_pausing:
+                        # All EP peers are PAUSING with no work — consensus.
+                        await self._world_barrier()
```
sidsingh-nvidia: Do we need this? _ep_establish_consensus is already a barrier on the signal? But perhaps you want the entire world (and not just the ep group) to pause together.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -683,6 +710,10 @@ def resume(self):
             )
         )
 
+        # In the coordinator path, the state is already handled.
+        if not self.use_coordinator:
```
sidsingh-nvidia, line 753: This doesn't make sense to me. All ranks are running the `run_engine_with_coordinator` coroutine, so all of them have self.use_coordinator = True.

megatron/core/inference/engines/dynamic_engine.py
tdene: This is mainly to prevent flakiness in CI unit-tests.

If we believe the engine_loop_task is stuck and will never end gracefully, we can run `shutdown` with `immediate=True` inside `dynamic_engine.py`. This will attempt to `task.cancel` the engine_loop_task.

But `task.cancel` does not work if we're stuck on a synchronous receive. Hence the attempt to fix it with a timeout.

I also dislike this timeout. I'll try to find a better solution, but it's slow and difficult since the unit-test only hangs when run through CI, and passes just fine when run in an interactive job.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -634,14 +659,16 @@ def suspend(self):
         for request_id in recompute_active_ids:
             self.requests[request_id].record.checkpoint()
 
+        # In the coordinator path, the state is already handled.
+        if not self.use_coordinator:
```
sidsingh-nvidia, line 702: This doesn't make sense to me. All ranks are running the run_engine_with_coordinator coroutine, so all of them have self.use_coordinator = True.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1583,77 +1621,140 @@ def schedule_requests(self) -> int:
...
+            immediate: Cancel the engine task instead of draining gracefully.
         """
+        task = getattr(self, 'engine_loop_task', None)
 
-        if hasattr(self, "inference_coordinator_process"):
-            self.inference_coordinator_process.join()
-        for socket in self.zmq_sockets:
-            socket.close()
+        if task is not None and not task.done() and task is not asyncio.current_task():
+            if immediate:
```
sidsingh-nvidia: why can't we do the signal injection for immediate as well?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1583,77 +1621,140 @@ def schedule_requests(self) -> int:
...
+                assert self.state == EngineState.PAUSED, f"Received SUSPEND in state {self.state}"
+                self.suspend()
+                self.state = EngineState.SUSPENDING
+                break
+
             elif header == Headers.RESUME:
-                self.suspend_signal = False
+                assert self.state == EngineState.SUSPENDED, f"Received RESUME in state {self.state}"
+                self.suspended.clear()
+                self.resume()
```
sidsingh-nvidia, line 1811: Is the resume still asynchronous?

megatron/core/inference/engines/dynamic_engine.py
sidsingh-nvidia: `If we believe the engine_loop_task is stuck ` - do you know why this would happen? I think we should try to root cause this first.

megatron/core/inference/engines/dynamic_engine.py
tdene: Yeah, exactly. I want the whole EP group to first establish consensus among itself so it can stop submitting dummy work, and then I want the whole world to establish consensus and pause together.

megatron/core/inference/engines/dynamic_engine.py
tdene: Yeah, that's why we're inserting timeouts.

I will keep working on this part of the code. I also do not like having a non-graceful shutdown, but I'm trying to get the CI runner to handle unit tests properly.

megatron/core/inference/engines/dynamic_engine.py
tdene: We can still run inference without the coordinator. It's probably not advised, but it's not officially deprecated.

megatron/core/inference/engines/dynamic_engine.py
tdene: We can still run inference without the coordinator. It's probably not advised, but it's not officially deprecated.

megatron/core/inference/engines/dynamic_engine.py
tdene: I tried that, and it's because `immediate` is meant to handle the case of an external environment (like pytest) killing processes. If we do the signal injection, we need to guarantee that all other ranks are still alive. The point of `immediate` is to have a fall-back option when we lack exactly that guarantee.

examples/inference/gpt/gpt_dynamic_inference_with_coordinator.py
tdene: > Did we change from stop_engines() to pause()=>stop() because it's mandatory to do this now? Or because you want these signals being tested in CI?

Because it's mandatory to do this now. At a high-level, I'm enforcing that only a PAUSED engine can STOP, because it eliminates a lot of possible race conditions. And it also makes sense.

> Okay, how about we create a method in the engine called wait_until_paused, which calls await self.paused.wait? Could be generalize to engine.wait_until(signal_enum) as well.

Addressed!

megatron/core/inference/engines/async_zmq_communicator.py
tdene: Addressed!

megatron/core/inference/engines/async_zmq_communicator.py
tdene: Addressed!

megatron/core/inference/engines/dynamic_engine.py
tdene: Addressed!

megatron/core/inference/engines/dynamic_engine.py
tdene: Reverted to original. Addressed.

megatron/core/inference/engines/dynamic_engine.py
tdene: Done!

megatron/core/inference/engines/dynamic_engine.py
tdene: Addressed!

megatron/core/inference/engines/dynamic_engine.py
tdene: Addressed!

megatron/core/inference/engines/dynamic_engine.py
tdene: Addressed

megatron/core/inference/engines/dynamic_engine.py
tdene: Yeah, I agree. I have gone ahead and removed all this logic until I can root-cause why unit tests fail in CI but not when run locally.

megatron/core/inference/engines/dynamic_engine.py
tdene: We want to have a barrier that achieves consensus that everyone has resumed, so that an external waiter can `wait_for` until we are in the `RESUMED` state, if that's what you are asking.

Other than that, in the engine, "resume" just means bring back the KV$ that was previously suspended.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -247,18 +269,34 @@ def reset(self) -> None:
...
-        self.received_stop: bool = False
-        self.suspend_signal = False
-        self.is_suspended = False
+        self.state = EngineState.RUNNING
+        # Awaitable events for external callers to detect state transitions.
+        # These mirror self.state for the four stable states: the engine sets/clears them when it
+        # transitions, and external callers can await them to block until the target state.
+        for state, attr in self._STATE_EVENTS.items():
+            event = asyncio.Event()
+            if state == self.state:
```
sidsingh-nvidia: Is this always setting the running event to True? Perhaps we could just write it as that?

megatron/core/inference/engines/dynamic_engine.py
sidsingh-nvidia: Got it, can you add this as a comment please?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1677,118 +1753,127 @@ async def run_engine(self, *, loop: Optional[asyncio.AbstractEventLoop] = None):
...
+            (global_work, all_pausing): max work across EP, and whether
+            all peers signaled consensus.
         """
-        range_push("_ep_group_has_work")
-
-        is_stopped = self.stopped.is_set() or self.received_stop
-        is_paused = self.paused.is_set() or self.received_pause
-        is_suspended = self.suspend_signal
-        if is_stopped or is_paused or is_suspended:
-            # Signals can be received asynchronously on EP ranks.
```
sidsingh-nvidia, line 1825: Can we keep both of these comments around?

megatron/core/inference/engines/dynamic_engine.py
sidsingh-nvidia: I am thinking if we should have a world barrier before resume. Not sure.

megatron/core/inference/engines/dynamic_engine.py
sidsingh-nvidia: Thanks!

megatron/core/inference/engines/dynamic_engine.py
sidsingh-nvidia: resolving because its outdated.

megatron/core/inference/engines/dynamic_engine.py
sidsingh-nvidia: thanks!

megatron/core/inference/engines/dynamic_engine.py
tdene: I thought it was fine, because the state-transition logic and asserts guarantees we will always be in `SUSPENDED` before we run resume. And `SUSPENDED` has a world barrier to make sure all ranks suspend together. And then there's a world barrier after `RESUME` to make sure all ranks finish resuming together.

So because of the state-transition guarantee (we're always resuming out of suspend), we do have a world barrier before and one after.

megatron/core/inference/engines/dynamic_engine.py
tdene: Addressed

megatron/core/inference/engines/dynamic_engine.py
tdene: Addressed

megatron/core/inference/engines/dynamic_engine.py
tdene: Addressed.

megatron/core/inference/engines/dynamic_engine.py
tdene: I have fixed the need for all this logic by making `InferenceClient.stop_engines` no longer broadcast a signal that kills BOTH the engine AND the coordinator. Instead, `InferenceClient` now has separate methods: `stop_engines` and `shutdown_coordinator`.

ko3n1g (approved):

tdene:
> @tdene how do we test that this:
> 
> * Does not cause any slowdown on top of what we have.
> * Does not lead to the accuracy regressions

In terms of accuracy regression, I don't think we currently have the ability to test this for any PR. That said, this does not affect rollout generation, so I do not see a way in which it could cause accuracy regression.

In terms of slowdown, there's functional tests in CI to check against slowdown. Also this PR actives a unit-test to verify the coordinator does not slow down (and it does not slow down; this unit test was written but inactive for a few months now).

yobibyte:
> > @tdene how do we test that this:
> > 
> > * Does not cause any slowdown on top of what we have.
> > * Does not lead to the accuracy regressions
> 
> In terms of accuracy regression, I don't think we currently have the ability to test this for any PR. That said, this does not affect rollout generation, so I do not see a way in which it could cause accuracy regression.
> 
> In terms of slowdown, there's functional tests in CI to check against slowdown. Also this PR actives a unit-test to verify the coordinator does not slow down (and it does not slow down; this unit test was written but inactive for a few months now).

Could you, please, run `just functional` with this before/after and compare the time it takes to finish? You can prob use my test harness for that.

tdene:
> > > @tdene how do we test that this:
> > > 
> > > * Does not cause any slowdown on top of what we have.
> > > * Does not lead to the accuracy regressions
> > 
> > 
> > In terms of accuracy regression, I don't think we currently have the ability to test this for any PR. That said, this does not affect rollout generation, so I do not see a way in which it could cause accuracy regression.
> > In terms of slowdown, there's functional tests in CI to check against slowdown. Also this PR actives a unit-test to verify the coordinator does not slow down (and it does not slow down; this unit test was written but inactive for a few months now).
> 
> Could you, please, run `just functional` with this before/after and compare the time it takes to finish? You can prob use my test harness for that.

Done! They both took about 45 minutes, with this branch finishing slightly early at `02:15:16.136000`, and `main` finishing at `02:15:29.290000`.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -251,13 +276,12 @@ def reset(self) -> None:
...
         self._loop = get_asyncio_loop(getattr(self, "_loop", None))
         self._cond = asyncio.Condition()
-        self.running = asyncio.Event()
-        self.paused = asyncio.Event()
-        self.stopped = asyncio.Event()
-        self.received_pause: bool = False
-        self.received_stop: bool = False
-        self.suspend_signal = False
-        self.is_suspended = False
+        for attr in self._STATE_EVENTS.values():
```
yobibyte: I personally dislike setattr() as they make code less readable and less observable even in the debugger. IMO, it'll be much nicer to have a self._state_events as:

```python
self._state_events = {k: asyncio.Event(),for k in _STATE_EVENTS}
```

With this, you do not need the string representation, and you can easily print the state by just printing the state_events dict.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -626,11 +667,9 @@ def suspend_resume_ctx(key: str, *, unified_memory_level: int) -> None:
     def suspend(self):
         """Suspend engine by deallocating context's GPU state."""
 
-        # Skip if already suspended, which can happen when using the inference
-        # coordinator.
-        if self.is_suspended:
+        # Skip if already suspended or in the process of suspending.
+        if self.state in (EngineState.SUSPENDED, EngineState.SUSPENDING):
```
yobibyte, line 670: I will either make a method that checks this or make `SUS_STATES=(EngineState.SUSPENDED, EngineState.SUSPENDING)` and check `self.state in SUS_STATES`.

This will be easier to maintain in case we want to add another state here. Otherwise, we will have to modify each `if self.state in ...` which you duplicate.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1807,118 +1881,144 @@ async def run_engine(self, *, loop: Optional[asyncio.AbstractEventLoop] = None):
...
-                try:
-                    await self.async_step()
-                except EngineSuspendedError:
+                elif self.state == EngineState.PAUSED:
                     await asyncio.sleep(0.02)
-                    continue
 
-        except asyncio.CancelledError:
-            pass
+                elif self.state == EngineState.UNPAUSING:
```
yobibyte, line 1997: I think, we should move your diagram from the PR description to code.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/flask_server.py
```diff
@@ -70,6 +76,8 @@ def health_check():
         return "Megatron Dynamic Inference Server is running."
 
     loop = asyncio.get_event_loop()
+    executor = ThreadPoolExecutor(max_workers=8192)
```
yobibyte: There should be a constant for this.

megatron/core/inference/engines/dynamic_engine.py
tdene: Addresed!

megatron/core/inference/engines/dynamic_engine.py
tdene: I disagree with this one for two reasons:

- This call is in the `suspend` method, and this method can be assumed to take responsibility.
- The list of states being checked here should never changed. It will always be `SUSPENDED` and `SUSPENDING`.

megatron/core/inference/text_generation_server/dynamic_text_gen_server/flask_server.py
tdene: Addressed by reverting all changes to this file; [#3648](https://github.com/NVIDIA/Megatron-LM/pull/3648/) will soon rewrite it anyway.

megatron/core/inference/engines/dynamic_engine.py
tdene: I need to check with other reviewers on this, and will act based on their feedback.

tdene: I do not think adding the current figure is a good idea since it is not straightforward to understand. It needs simplification. After the PR is merged, I will review the design in detail.

megatron/core/inference/engines/dynamic_engine.py
yobibyte: I got triggered because you have self.state in (SUSPENDING, SUSPEND) 3 or 4 times in this PR.

megatron/core/inference/engines/dynamic_engine.py
yobibyte: Up to you!

PR #3580, opened 2026-02-25
Reverse polarity of the off-policy measurement

tdene:
# What does this PR do ?

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -262,6 +262,7 @@ def reset(self) -> None:
         self.requests: Dict[int, RequestEntry] = {}
         self.waiting_request_ids = deque()
         self.failed_request_ids = []
+        self._generation_epoch: Optional[int] = None
```
yobibyte, line 278: In what situation are we okay with this being None?

tdene: When we're not doing RL.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1132,6 +1133,34 @@ def post_process_requests(
...
+                                    torch.full(
+                                        (total - len(tensor),),
+                                        it,
+                                        dtype=tensor.dtype,
+                                        device=tensor.device,
+                                    ),
+                                ),
+                                dim=0,
+                            ),
+                        )
```
santhnm2: Can you move this to a helper function (e..g., `_set_generation_epoch`)?

megatron/core/inference/data_parallel_inference_coordinator.py
```diff
@@ -447,7 +447,9 @@ def start(self):
                     self.state = self.CoordinatorState.STOPPING
 
                 # Broadcast the control signal if we're in a good state.
-                broadcast_payload = msgpack.packb([header.value], use_bin_type=True)
+                # Forward the full deserialized payload so that data-bearing
+                # signals (e.g. SET_GENERATION_EPOCH) retain their arguments.
+                broadcast_payload = msgpack.packb(deserialized_payload, use_bin_type=True)
```
santhnm2, line 450: Does this add overhead?

tdene: This was my main concern. I timed it. It does not add any measurable overhead, because it is still a single packet that's being sent over the network (just a slightly larger packet).

megatron/core/inference/engines/dynamic_engine.py
tdene: Addressed! I rewrote the logic itself so that it's not awful code. Do you think this new version is readable enough that we can get rid of the `_set_generation_epoch` helper and just in-line this?

megatron/rl/rl_utils.py
```diff
@@ -1895,15 +1862,14 @@ def megatron_rl_inference_mode(
             toggle_cuda_graphs(lang_module, cuda_graph_impl)
 
         inference_interface = get_inference_interface(args, loop, model)
+        inference_interface.set_generation_epoch(get_args().curr_iteration)
```
lmcafee-nvidia, line 1867: if I'm understanding this correctly, the high level goal is update each token's 'staleness tracking' value only once on its first iteration, rather than updating/incrementing it for each iteration that the token is live, is that correct? are there any other goals here, such as knowing the training iteration (I don't know if this was tracked before)?

megatron/rl/rl_utils.py
```diff
@@ -795,13 +794,13 @@ def compute_group_stats(
             roll_turn_lens = [len(t) for t in rollout.trajectory]
             group_turn_lengths.extend(roll_turn_lens)
             group_traj_lengths.append(sum(roll_turn_lens))
-            group_policy_staleness.extend(s for turn in rollout.policy_staleness for s in turn)
-            group_kv_staleness.extend(s for turn in rollout.kv_cache_staleness for s in turn)
-            group_completed_at_steps.extend(rollout.completed_at_step)
+            group_policy_epoch.append(min(s for turn in rollout.policy_epoch for s in turn))
```
lmcafee-nvidia: is it possible for these lists to ever be empty? if so, the min/max ops on these lines need to be handled properly

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1160,6 +1165,25 @@ def post_process_requests(
 
         return active_request_ids, finished_request_records
 
+    @staticmethod
+    def _stamp_generation_epoch(request: DynamicInferenceRequest, epoch: int):
+        """Extend policy_epoch and kv_cache_epoch tensors to cover all current tokens."""
+        total = len(request.prompt_tokens) + len(request.generated_tokens)
+
+        if (t := request.policy_epoch) is not None:
+            request.policy_epoch = torch.cat(
```
lmcafee-nvidia: may be a small performance optimization to skip calling `torch.cat` if `total - len(t) == 0`. i.e., guard with a `if total - len(t) > 0:`

megatron/rl/rl_utils.py
lmcafee-nvidia, line 1867: can there be a race condition at all between the epoch signal and the resume signal on the next line. For example, can the resume signal arrive first, and tokens continue generating with the old epoch value, and then the updated epoch value arrives, in which case some tokens are stamped with the wrong epoch?

megatron/core/inference/engines/dynamic_engine.py
lmcafee-nvidia: just an observation, each `torch.cat()` is an O(N) copy operation where N==generation length, and this `torch.cat()` is called O(N) times since it's called during `_post_process_requests()` for each generated token. so while `torch.cat()` is relatively cheap, we are doing O(N^2) operations for tracking staleness. are there any scalability implications to worry about?

megatron/rl/rl_utils.py
tdene: > if I'm understanding this correctly, the high level goal is update each token's 'staleness tracking' value only once on its first iteration, rather than updating/incrementing it for each iteration that the token is live, is that correct?

Correct. When the token is generated, it gets stamped with a label. That label never changes.

> are there any other goals here, such as knowing the training iteration (I don't know if this was tracked before)?

Right now, in GRPO, we are stamping the token with the training iteration.    
We have been talking about other algorithms that will require other kinds of stamps... but this PR is future-facing and would let us implement such "other algorithms", because we're just stamping the token with a value.

tdene: There cannot be, specifically because both the epoch signal and resume signal are coded as "signal"s. That makes them enter a specific "signal" queue in the engine, which is guaranteed to arrive and be processed in order.

lmcafee-nvidia: got it, thanks

lmcafee-nvidia: makes sense!

megatron/rl/rl_utils.py
tdene: Handled with an assert.

megatron/core/inference/engines/dynamic_engine.py
tdene: See the thread below about performance optimizations.

tdene: I will actually do better than fix this.

A different reviewer's feedback has led me to redesign the system, so that instead of storing massive tensors of repeated numbers, we just compress them from `[3, 3, 3, 4, 4, 4, 5, 5, 6, 6, 6, 6]` to `[ (3,0), (4, 3), (5, 6), (6, 8) ]`.

lmcafee-nvidia: great approach!

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -828,6 +829,12 @@ def _add_request(
             )
             request.add_event_add_engine()  # Record when request enters engine
 
+            # Stamp new request with the current generation epoch.
+            if self._generation_epoch is not None:
+                epoch = self._generation_epoch
+                request.policy_epoch = [(0, epoch)]
```
yobibyte, line 835: Can generation_epoch be 0?

tdene: Yes, it can be. It could be any value inference's user (i.e. RL) specifies. Does that cause any issues here, I don't see.

yobibyte: NVM, I misread the meaning of the tuple. I first thought we will have two zeros there.

PR #3584, opened 2026-02-25
Fix default cuda graph persist arg. Add persist to rl common.sh.

yobibyte:
#3029 made the default arguments invalid: it made `--rl-persist-cuda-graphs` True by default with the `--cuda-graph-impl=none` from before that fired the assert in `validate_args`. 

This PR makes persist False by default to make things consistent and default arg configuration valid.

### DIFF
--- a/megatron/core/enums.py
+++ b/megatron/core/enums.py
@@ -8,6 +8,14 @@ class ModelType(enum.Enum):
 
     encoder_or_decoder = 1
 
+    @property
+    def encoder_and_decoder(self):
+        """Deprecated property - use encoder_or_decoder instead."""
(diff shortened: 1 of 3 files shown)


jalbericiola (approved):
LGTM

yobibyte:
/ok to test [d373519](https://github.com/NVIDIA/Megatron-LM/pull/3584/commits/d3735191c764e1aa6e1625860df2b53e5fa08aad)[d373519](https://github.com/NVIDIA/Megatron-LM/pull/3584/commits/d3735191c764e1aa6e1625860df2b53e5fa08aad)

PR #3594, opened 2026-02-25
Add speculative decoding support with MTP layers

santhnm2:
# What does this PR do ?

This PR adds speculative decoding support for inference.

### How it works
Each generation step proceeds in phases:

1. Input construction — For each active decode request, 1 + `K` tokens are fed into the model: the previously sampled token plus K speculative tokens from the MTP heads of the previous step. These are interleaved across requests with matching position IDs.
(`dynamic_context.py`: `update_requests`)

2. Forward pass — The model processes all tokens in one pass. The base decoder produces logits at every position (note: `materialize_only_last_token_logits` must be off). The MTP heads produce `K` additional sets of logits from their lightweight transformer/Mamba layers, cached in `model._mtp_logits_cache`. These are concatenated to produce a `[1+K, seq_len, vocab_size]` logit tensor.
(`text_generation_controller.py`: `_dynamic_step_forward_logits`)

3. Sampling & verification — Both base and MTP logits are sampled (grouped into temperature/top_k/top_p buckets for efficiency). Then a greedy token-matching verification determines how many speculative tokens to accept: a speculative token at position `t+k` is accepted iff the base model's output at position `t+k-1` equals it. Acceptance is consecutive — once a mismatch occurs, all subsequent speculative tokens for that request are rejected (enforced via `cummin`).
(`text_generation_controller.py`: `_dynamic_step_sample_logits_and_verify_tokens`)

4. KV cache rewind — For rejected tokens, the KV cache is rolled back: block offsets are decremented, and if the rewind crosses a block boundary, the block is released back to the allocator. For Mamba/hybrid models, SSM recurrent state is restored from intermediate snapshots captured during the Triton kernel execution.
(`text_generation_controller.py`: `_rewind_kv_cache`)

5. Bookkeeping — Sequence lengths advance by `accepted_count + 1` (not just 1). Finish conditions (EOS, max length) are checked, and the accepted + sampled tokens are appended to each request's output. New MTP-sampled tokens are staged for the next step.

### MTP head architecture

Each `MultiTokenPredictionLayer` (`multi_token_prediction.py`) takes the hidden states from the previous depth, concatenates them with shifted input embeddings via a learned projection (`eh_proj`), and runs the result through a transformer block (or Mamba stack). This is repeated `K` times to produce predictions at positions `t+1` through `t+K`. A `mtp_use_repeated_layer` option shares weights  across all K layers.

### Mamba/hybrid SSM support

SSM models require special handling because they carry recurrent state that must be rollback-able:
  - Conv state: The circular buffer is enlarged by `K` slots, so no explicit save/restore is needed.
  - SSM state: The Triton SSM kernel (`mamba_ssm.py`) dumps intermediate state snapshots at every speculative step into a pre-allocated `[num_layers, max_requests, K+1, ...]` buffer — zero extra GPU-CPU sync. On rewind, the correct snapshot is indexed by accepted token count.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/examples/post_training/modelopt/README.md
+++ b/examples/post_training/modelopt/README.md
@@ -34,6 +34,7 @@ knowledge distillation, pruning, speculative decoding, and more.
 | `moonshotai/Kimi-K2-Instruct` | ✅ | ✅ | - | - |
 | `nvidia/NVIDIA-Nemotron-Nano-9B-v2` | ✅ | - | ✅ | ✅ |
 | `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16` | ✅ | - | ✅ | ✅ |
+| `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16` | ✅ | - | ✅ | ✅ |
 | `openai/gpt-oss-{20b, 120b}` | ✅ | **Online** | ✅ | ✅ |
 | `Qwen/Qwen3-{0.6B, 8B}` | ✅ | ✅ | ✅ | ✅ |
 | `Qwen/Qwen3-{30B-A3B, 235B-A22B}` | **WAR** | ✅ | ✅ | ✅ |
--- a/examples/post_training/modelopt/conf/nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16.sh
+++ b/examples/post_training/modelopt/conf/nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+if [ -z ${HF_MODEL_CKPT} ]; then
+    HF_MODEL_CKPT=nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16
+    TOKENIZER_MODEL=nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16
+else
+    TOKENIZER_MODEL=${HF_MODEL_CKPT}
+fi
+
+
+
+MODEL_ARGS=" \
+    --trust-remote-code \
+    --save-interval 100000 \
+    --micro-batch-size 1 \
+    --enable-experimental \
+    --use-fused-weighted-squared-relu \
+    --cross-entropy-loss-fusion \
+    --cross-entropy-fusion-impl native \
+    --num-experts 512 \
+    --moe-router-score-function sigmoid \
+    --moe-grouped-gemm \
+    --moe-aux-loss-coeff 1e-4 \
+    --moe-router-topk 22 \
+    --moe-permute-fusion \
+    --moe-router-topk-scaling-factor 5.0 \
+    --moe-router-enable-expert-bias \
+    --moe-router-dtype fp32 \
+    --moe-router-load-balancing-type seq_aux_loss \
+    --moe-shared-expert-intermediate-size 5376 \
+    --moe-token-dispatcher-type allgather \
+    --moe-latent-size 1024 \
+    \
+    --attention-backend flash \
+    --disable-gloo-process-groups \
+    --is-hybrid-model \
+    --mamba-num-heads 128 \
+    --mamba-head-dim 64 \
+    --hybrid-layer-pattern MEMEMEM*EMEMEMEM*EMEMEMEM*EMEMEMEMEM*EMEMEMEMEM*EMEMEMEMEM*EMEMEMEMEM*EMEMEMEM*EMEMEMEME \
+    \
+    --use-mcore-models \
+    --untie-embeddings-and-output-weights \
+    --disable-bias-linear \
+    --init-method-std 0.014 \
+    --position-embedding-type none \
+    --squared-relu \
+    --hidden-size 4096 \
+    --num-attention-heads 32 \
+    --group-query-attention \
+    --num-query-groups 2 \
+    --ffn-hidden-size 2688 \
+    --kv-channels 128 \
+    --normalization RMSNorm \
+    --attention-dropout 0.0 \
+    --hidden-dropout 0.0 \
+    \
+    --tokenizer-type HuggingFaceTokenizer \
+    --bf16 \
+    --seq-length 8192 \
+    --max-position-embeddings 8192 \
+    --export-model-type MambaModel \
+    "


megatron/inference/utils.py
```diff
@@ -341,7 +341,7 @@ def get_inference_config_from_model_and_args(model: MegatronModule, args):
         mamba_inference_state_config=mamba_inference_state_config,
         pg_collection=pg_collection,
         use_flashinfer_fused_rope=args.use_flashinfer_fused_rope,
-        materialize_only_last_token_logits=not args.return_log_probs,
+        materialize_only_last_token_logits=(not args.return_log_probs and not args.num_speculative_tokens > 0),
```
mathemakitten: Any reason not to do `args.num_speculative_tokens == 0`?

megatron/core/inference/batch_dimensions_utils.py
```diff
@@ -372,7 +382,7 @@ def add_if_valid(token_count: int, prefill_req_count: int, decode_req_count: int
             ):
                 cuda_graph_max_tokens = max_tokens
 
-            assert cuda_graph_max_tokens == max_requests, (
+            assert cuda_graph_max_tokens == max_requests * (num_speculative_tokens + 1), (
                 f"cuda_graph_max_tokens ({cuda_graph_max_tokens}) must equal max_requests "
                 f"({max_requests}). This is required for correctly syncing EP ranks: "
```
mathemakitten: Update this comment to note speculative tokens

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -2024,55 +2080,59 @@ def resume_paused_requests(
         resume_request_count = 0
         if self.paused_request_count > 0:
             active_block_count_avail = self.block_allocator.get_active_avail()
-            paused_block_counts = self.request_kv_block_counts[: self.paused_request_count]
+            paused_block_counts = self.request_kv_block_counts[: self.paused_request_count].clone()
```
mathemakitten: Why is this needed?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -360,6 +362,7 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
                 math.prod(self.mamba_ssm_states_shape) * self.mamba_ssm_states_dtype.itemsize
             )
             mamba_states_memory_per_request *= self.num_mamba_layers
+            mamba_states_memory_per_request *= self.num_speculative_tokens + 1
```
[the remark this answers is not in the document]
santhnm2: This is to account for the intermediate mamba states that are allocated.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1200,8 +1243,15 @@ def _check_stop_words_for_request_post_append(self, request: DynamicInferenceReq
             stop_len = len(stop_word_ids)
             if len(generated_tokens) >= stop_len:
                 # Check if the last stop_len tokens match the stop word
-                if generated_tokens[-stop_len:] == stop_word_ids:
-                    return True
+                if stop_len > self.num_speculative_tokens:
```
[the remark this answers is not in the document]
santhnm2: Can you add a corresponding unit test for this change?

santhnm2:
> general question, have you thoroughly tested speculative decoding + prefix caching? I heard that it took the Nemo group 4 months to get this working together correctly, but I'm not sure why that is.

I've been testing on a proxy model, haven't seen any obvious issues yet but will continue looking at it.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1199,10 +1242,13 @@ def _check_stop_words_for_request_post_append(self, request: DynamicInferenceReq
...
-                # Check if the last stop_len tokens match the stop word
-                if generated_tokens[-stop_len:] == stop_word_ids:
-                    return True
-
+                # Check the last stop_len tokens shifting by 1 up to num_speculative_tokens.
+                # We do this regardless of stop_len because speculative decoding can append
+                # multiple tokens at once, meaning the stop word might end at any of those positions.
+                for i in range(self.num_speculative_tokens + 1):
+                    end_idx = -i if i > 0 else None
+                    if list(generated_tokens[-stop_len - i : end_idx]) == stop_word_ids:
```
lmcafee-nvidia, line 1312: `generated_tokens[-stop_len - i : end_idx]` can potentially be shorter than `stop_word_ids` ... can that cause any issues?

megatron/inference/utils.py
lmcafee-nvidia: order of precedence is not clear here, this would be clearer (I think) as `... and args.num_speculative_tokens == 0`

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: is this correct? this multiplies the already large mamba memory by a factor, say 4-5x or so? are we actually permanently storing all of these states, or are they only used temporarily in between steps? this seems like it'll make the mamba buffer very large.

megatron/core/models/gpt/gpt_model.py
```diff
@@ -613,20 +613,41 @@ def _postprocess(
...
-                compute_language_model_loss=self.compute_language_model_loss,
-                config=self.config,
-                cp_group=self.pg_collection.cp,
-                packed_seq_params=packed_seq_params,
-                scale_logits_fn=self._scale_logits if self.config.use_mup else None,
-            )
+            assert self.config.mtp_num_layers > 0
+            # The new process_mtp_loss function doesn't handle mtp_logits_cache,
+            # so we manually generate and cache MTP logits when in inference mode.
+            if in_inference_mode:
```
lmcafee-nvidia, line 636: this code is shared between `GPTModel` and `MambaModel` as far as I can tell. I suggest we share this code in a utility method.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -618,25 +662,380 @@ def _dynamic_step_sample_bookkeeping(self):
...
+            accepted_tokens_per_decode_request = accepted_tokens_per_request[is_decode_mask]
+
+            if decode_mamba_indices.numel() > 0:
+                context.mamba_ssm_states[:, decode_mamba_indices] = (
+                    context.mamba_intermediate_ssm_states[
+                        :, decode_mamba_indices, accepted_tokens_per_decode_request
+                    ]
+                )
+
+    def _dynamic_step_sample_logits_and_verify_tokens(
```
lmcafee-nvidia: this method is quite long now. you might consider breaking it up into separate steps.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -202,6 +202,21 @@ def __init__(self, controller: TextGenerationController, context: DynamicInferen
...
+
+        if self.num_speculative_tokens > 0:
+            assert (
+                not inference_config.materialize_only_last_token_logits
+            ), "Speculative decoding requires materialize_only_last_token_logits to be False"
+            assert (
+                self.num_speculative_tokens <= self.controller.num_mtp_heads
+            ), f"Number of speculative tokens {self.num_speculative_tokens} must be less than or equal to number of MTP heads {self.controller.num_mtp_heads}"
+
+        self.controller._init_mtp_sampling_tensor()
```
lmcafee-nvidia: is it possible to initialize this tensor during controller construction? it's always a bit awkward to me when things are initialized after the fact.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -2380,6 +2476,14 @@ def update_requests(self, active_requests_mask: Tensor, new_tokens: Tensor) -> T
         # 6. Now that we have the requests in following order [Paused, Active, Finished]
         # We determine how many requests we can resume and resume them
 
+        # For multi-token generation: store previous block IDs BEFORE resume allocates new blocks.
+        # This allows us to know which block tokens should go to if they don't cross the boundary.
+        # After resume_paused_requests, request_last_kv_block_id will be updated to the NEW block
+        # for resumed requests, but we need the OLD block for tokens that don't cross.
+        prev_last_block_ids = None
+        if self.num_speculative_tokens > 0:
+            prev_last_block_ids = self.request_last_kv_block_id.clone()
```
lmcafee-nvidia, line 2538: this clone is unnecessary as far as I can tell

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -2407,45 +2511,144 @@ def update_requests(self, active_requests_mask: Tensor, new_tokens: Tensor) -> T
...
-        self.token_to_input_ids[: self.active_token_count] = next_tokens[
-            self.paused_request_count : self.total_request_count
-        ]
-
         if self.paused_request_count > 0:
-            self.paused_tokens = next_tokens[: self.paused_request_count]
-
-        # add_ and fill_ calls seems to work as intended with sliced indexing (i.e. x[3:5].add(...) or x[3:5].fill_)
-        # but when another tensor is used for indexing, it does not work as expected (i.e. x[y] if x and y are torch tensors)
+            self.paused_tokens = next_tokens[: self.paused_request_count].clone()
```
lmcafee-nvidia, line 2573: why did we add this clone?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -2276,6 +2357,10 @@ def update_requests(self, active_requests_mask: Tensor, new_tokens: Tensor) -> T
         if self.paused_request_count != 0:
             assert self.paused_tokens is not None
             next_tokens = torch.cat((self.paused_tokens, new_tokens))
+            if new_speculative_tokens is not None:
+                new_speculative_tokens = torch.cat(
+                    (self.paused_speculative_tokens, new_speculative_tokens), dim=1
+                )
```
tdene: If I'm reading this correctly, `self.paused_speculative_tokens` is used here but only initialized later in the same function. Both places check `if self.paused_request_count > 0`. So it seems like we have an error after the first paused request?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -618,25 +662,380 @@ def _dynamic_step_sample_bookkeeping(self):
...
+            decode_outputs_shifted = decode_outputs.roll(1, dims=1)
+
+            decode_mask_2d = decode_inputs == decode_outputs_shifted
+
+            # The first token (base token) is always accepted
+            decode_mask_2d[:, 0] = True
+
+            # ENFORCE CONSECUTIVE ACCEPTANCE:
+            # cummin() on booleans propagates False (0) to the right, invalidating all subsequent mismatches
+            decode_mask_2d = decode_mask_2d.cummin(dim=1).values
```
lmcafee-nvidia, line 1068: out of curiosity, why do we use `cummin` here instead of `argmin`?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -618,25 +662,380 @@ def _dynamic_step_sample_bookkeeping(self):
...
+            is_decode_mask = context.request_in_prefill_status_tensor[active_request_slice] == 0
+            decode_mamba_indices = active_mamba_indices[is_decode_mask]
+            accepted_tokens_per_decode_request = accepted_tokens_per_request[is_decode_mask]
+
+            if decode_mamba_indices.numel() > 0:
+                context.mamba_ssm_states[:, decode_mamba_indices] = (
+                    context.mamba_intermediate_ssm_states[
+                        :, decode_mamba_indices, accepted_tokens_per_decode_request
+                    ]
+                )
```
tdene, line 786: Do we not need to rewind the conv state as well?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -652,24 +1051,33 @@ def _dynamic_step_sample_logits(self, logits: Tensor):
         if context.config.materialize_only_last_token_logits:
             # When materialize_only_last_token_logits is true, last_token_logits is
             # already called in the forward pass of GPT.
-            last_token_logits = logits.squeeze(0)
+            required_token_indices = logits.squeeze(0)
```
tdene: nit: I don't think these are "indices". This is actual logits data, just filtered.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -618,25 +662,380 @@ def _dynamic_step_sample_bookkeeping(self):
...
+            f"but got {len(required_logit_indices)} for num_decode_requests {num_decode_requests} "
+            f"and num_prefill_requests {num_prefill_requests}"
+        )
+
+        required_logits = logits.squeeze(0)[required_logit_indices, :]  # Shape [1, 11, vocab_size]
+        required_mtp_logits = mtp_logits[
+            :, required_logit_indices, :
+        ]  # Shape [num_speculative_tokens, 11, vocab_size]
+
+        # ================ PART 1 The following part of the code is to sample the logits and mtp logits based on the sampling parameters =========
```
tdene: nit: You have two PART 1's. This one should be PART 2.

megatron/core/inference/contexts/dynamic_context.py
santhnm2: These are only needed for the duration of a single step, but we would have to keep allocating and de-allocating the buffer for every step otherwise. Agree that it's a large memory overhead though. For prefix caching do we not have to pay the same memory penalty (that is, to materialize intermediate states)?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: I'm using this circular buffer implementation so it basically just rolls the window depending on how many tokens are accepted (this is controlled by `cache_seqlens` which reflects the total sequence length so far, including whichever tokens were accepted).

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1177,6 +1219,7 @@ def _get_and_clear_stop_word_finished_ids(self, active_request_ids: list[int]) -
         self.stop_word_finished_request_ids -= result
         return result
 
+    # TODO : We also might have to delete some tokens, if stop word hit in the middle (speculative case)
```
mathemakitten: So I'm clear — this means right now we might have a few extra tokens past the stopword and we just leave them in there? Would this be problematic to use as-is?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -618,25 +662,380 @@ def _dynamic_step_sample_bookkeeping(self):
...
+
+        # TODO : Maybe its okay to have a loop with num spec tokens ? (Since it will only be max 3 , so might be faster)
+        for request_indices, temp, top_k, top_p in self._torch_sampling_buckets:
+            request_indices_tensor = torch.tensor(
+                request_indices, device=token_to_request_index.device
             )
+            required_indices = torch.where(
+                torch.isin(token_to_request_index, request_indices_tensor)
+            )[0]
+            # TODO : Can maybe club the following two and then split later ?
```
mathemakitten: What does this mean?

megatron/core/inference/data_parallel_inference_coordinator.py
```diff
@@ -389,14 +389,12 @@ def start(self):
                     return
 
                 if request_hashes:
-                    self._update_rank_hashes(next_data_parallel_rank_identity, request_hashes)
```
kvareddy, line 392: What is this change about? Was it a bug fix?

megatron/core/inference/contexts/dynamic_context.py
santhnm2: I think clone is necessary because `request_last_kv_block_id` is updated in `resume_paused_requests`

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Added clone because `self.paused_tokens` is persisted across iterations

megatron/core/inference/engines/dynamic_engine.py
santhnm2: Added this line:
`max_shift = min(self.num_speculative_tokens, len(generated_tokens) - stop_len)`

megatron/core/inference/data_parallel_inference_coordinator.py
santhnm2: Yea I don't think `next_data_parallel_rank_identity` existed

megatron/core/models/gpt/gpt_model.py
santhnm2: Done

megatron/inference/utils.py
santhnm2: Fixed

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: Done

megatron/core/inference/engines/dynamic_engine.py
santhnm2: Done

megatron/core/inference/engines/dynamic_engine.py
santhnm2: Fixed

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Fixed

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: We need cummin to handle the case of [0, 1, 1] (i.e., first token is rejected so all subsequent tokens need to be rejected)
I think argmin would give a false negative on [1, 1, 0] right?

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Removed clone, the subsequent `flip` call makes a copy

megatron/core/inference/batch_dimensions_utils.py
santhnm2: Done

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: Fixed

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -970,16 +982,41 @@ def post_process_requests(
...
+            zip(request_ids.tolist(), sample.tolist(), accepted_tokens_iter, log_probs_iter)
         ):
+
+            # Ensure tokens is always a list for consistent handling
+            if not isinstance(tokens, list):
+                tokens = [tokens]
+
+            if self.num_speculative_tokens > 0:
+                accepted_tokens = list(filter(lambda tok: tok != -1, accepted_tokens_list))
+                tokens = accepted_tokens + tokens
```
kvareddy, line 1033: Should this be tokens = tokens + accepted_tokens?

megatron/core/models/gpt/gpt_model.py
```diff
@@ -613,20 +614,33 @@ def _postprocess(
...
+                hidden_states, self._mtp_logits_cache = compute_mtp_inference_logits(
+                    hidden_states=hidden_states,
```
kvareddy: I guess we need to filter out hidden_states based on inference_context.config.materialize_only_last_token_logits

megatron/core/inference/engines/dynamic_engine.py
santhnm2: Updated with comment - the current order is correct because `accepted_tokens` is the set of accepted tokens from the previous iteration, while `tokens` is the new token conditioned on those accepted tokens.

megatron/core/models/gpt/gpt_model.py
santhnm2: Reverted optimization for supporting `materialize_only_last_token_logits = True`, will address in a follow up PR

PR #3596, opened 2026-02-26
Update README Quick Start 

ilml:
## Summary

- Replace the `pip install` Quick Start instructions with a Docker-based workflow that works out of the box
- Steps: clone repo, launch NVIDIA PyTorch container, install from source (`pip install -e .`), run example
- The previous `pip install --no-build-isolation megatron-core[mlm,dev]` does not work OOTB

## Test plan

- [x] Verified the Docker workflow end-to-end on `nvcr.io/nvidia/pytorch:26.01-py3`
- [x] Confirmed `torchrun --nproc_per_node=2 examples/run_simple_mcore_train_loop.py` runs successfully

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/claude-complexity-label.yml
+++ b/.github/workflows/claude-complexity-label.yml
@@ -1,7 +1,7 @@
 name: Claude Complexity Label
 
 on:
-  pull_request:
+  pull_request_target:
     types: [ready_for_review]
 
 jobs:
--- a/.github/workflows/claude_review.yml
+++ b/.github/workflows/claude_review.yml
@@ -3,7 +3,7 @@ name: Claude Code Review
 on:
   issue_comment:
     types: [created]
-  pull_request:
+  pull_request_target:
     types: [ready_for_review]
 
 jobs:
@@ -70,7 +70,7 @@ jobs:
 
   review-on-ready:
     name: Claude Review (ready for review trigger)
-    if: github.event_name == 'pull_request' && github.event.action == 'ready_for_review'
+    if: github.event_name == 'pull_request_target' && github.event.action == 'ready_for_review'
     runs-on: ubuntu-latest
     permissions:
       contents: read


README.md
```diff
@@ -24,20 +24,34 @@ This repository contains two components: **Megatron-LM** and **Megatron Core**.
...
+      -v /path/to/Megatron-LM:/workspace/megatron-lm \
+      -e PIP_CONSTRAINT= \
+      nvcr.io/nvidia/pytorch:26.01-py3
+    ```
+
+3. Install from source inside the container:
+
+    ```bash
+    cd /workspace/megatron-lm
+    pip install -e .
```
ko3n1g:
```suggestion
    pip install -e --no-build-isolation .
```

ko3n1g (commented):
We should recommend no build isolation so that extensions are built against the host environment. But that's just a nit: Thanks a lot I think it makes sense to change the README!

README.md
ilml: fixed, thanks!

docs/get-started/install.md
```diff
@@ -28,69 +26,91 @@ Installation is supported using Docker and pip.
...
 
-- `dev`: Moving head that supports the most recent upstream dependencies
-- `lts`: Long-term support of NGC PyTorch 24.01
+### Install Megatron Core Only
 
-Both containers can be combined with `mlm`, which adds package dependencies for Megatron-LM on top of Megatron Core.
+For a minimal install with only PyTorch as a dependency:
 
+```bash
+pip install megatron-core
```
janEbert: This should also have `--no-build-isolation`. :)

docs/get-started/install.md
```diff
@@ -28,69 +26,91 @@ Installation is supported using Docker and pip.
...
 
-* For running an Megatron LM application
+### Install with LTS Dependencies
 
-        ```bash
-        pip install "setuptools<80.0.0,>=77.0.0" "packaging>=24.2"
-        pip install --no-build-isolation megatron-core[mlm,dev]
-        ```
-* Install packages for LTS support NGC PyTorch 24.01
+Long-term support pinned to NGC PyTorch 24.01:
```
janEbert: Actually current [LTS is based on NGC PyTorch 25.09](https://github.com/NVIDIA/Megatron-LM/blob/310082a6d257d25aa8da5127de63cd51705f2acf/docker/.ngc_version.lts). We should update this as well.

docs/get-started/install.md
janEbert: @ko3n1g or does it? The previous instruction for this also didn't have it.

docs/get-started/quickstart.md
```diff
@@ -7,54 +7,40 @@
    license agreement from NVIDIA CORPORATION is strictly prohibited.
 -->
 
-# Quick Start
+# Your First Training Run
 
-## Quick Installation
+This guide walks you through running your first training jobs with Megatron Core. Make sure you have completed [installation](install.md) before proceeding.
```
janEbert, line 12: "make sure you have completed **the** installation"

janEbert:
Looks beautiful!

Phlip79 (approved):

docs/get-started/install.md
ilml: fixed, thanks!

README.md
```diff
@@ -22,23 +22,12 @@ This repository contains two components: **Megatron-LM** and **Megatron Core**.
 **[Megatron Bridge](https://github.com/NVIDIA-NeMo/Megatron-Bridge)** provides bidirectional Hugging Face ↔ Megatron checkpoint conversion with production-ready recipes.
 
 
-## Quick Start
```
sbhavani, line 25: I think we need to keep some pip install instructions in the README. We can mention prereqs if it's incomplete on its own. 

A common feedback for megatron / TE is that installation is difficult and pushing users towards NGC containers just shows our process is brittle.

docs/get-started/install.md
```diff
@@ -7,90 +7,67 @@
...
 ```bash
-# Run container with mounted directories
-docker run --runtime --nvidia --gpus all -it --rm \
-  -v /path/to/megatron:/workspace/megatron \
+docker run --gpus all -it --rm \
   -v /path/to/dataset:/workspace/dataset \
   -v /path/to/checkpoints:/workspace/checkpoints \
   -e PIP_CONSTRAINT= \
-  nvcr.io/nvidia/pytorch:25.04-py3
+  nvcr.io/nvidia/pytorch:25.09-py3
```
sbhavani: I think we need to update the container version. If the TE version in the container is outdated then it will try to install newer TE (e.g. TE >=2.9)

docs/get-started/install.md
```diff
@@ -7,90 +7,67 @@
...
-1. Install the latest release dependencies
 
-    ```bash
-    pip install "setuptools<80.0.0,>=77.0.0" "packaging>=24.2"
-    pip install --no-build-isolation megatron-core[dev]
-    ```
+## Step 2: Install Megatron Core
 
-2. Next choose one of the following options:
+### Option A: Pip Install (Recommended)
```
sbhavani: should we mention somewhere that nv-grouped-gemm and mamba-ssm take time to compile?

docs/get-started/install.md
```diff
@@ -7,90 +7,67 @@
...
-* For a version of Megatron Core with only Torch, run
+For development or to run the latest unreleased code:
 
-        ```bash
-        pip install megatron-core
-        ```
+```bash
+git clone https://github.com/NVIDIA/Megatron-LM.git
+cd Megatron-LM
+pip install --no-build-isolation -e .[mlm,dev]
```
sbhavani: This failed when I tested with pip because it looks for emerging_optimizers on package indexes. It might work with uv since there's a mapping to the git repo there.

docs/get-started/install.md
janEbert: Also warn about OOMs that can be fixed with a lower `MAX_JOBS` envvar!

megatron/core/datasets/utils.py
```diff
@@ -18,15 +18,32 @@ class Split(Enum):
...
+
+    If the extension was already built (e.g. via pip wheel or setup.py), this is a no-op.
+    Falls back to compiling via ``make`` for editable / source installs that include the Makefile.
+    """
+    try:
+        import megatron.core.datasets.helpers_cpp  # noqa: F401
+
+        return
+    except ImportError:
+        pass
```
janEbert: It's good to still execute the `make` even if the module is available, so that potential updates to the C++ files get incorporated.

docs/get-started/install.md
janEbert: I agree, now that all mentions of `lts` are removed (but we should at least mention the LTS container and corresponding pip feature, no?), this should rather point to the `dev` container ([version 25.11](https://github.com/NVIDIA/Megatron-LM/blob/53c5973c07a56802d3cb0edf9396e1901eec8926/docker/.ngc_version.dev)).

README.md
```diff
@@ -22,23 +22,17 @@ This repository contains two components: **Megatron-LM** and **Megatron Core**.
...
 
-Install Megatron Core with pip:
-
-1. Install Megatron Core with required dependencies:
-
-    ```bash
-    pip install --no-build-isolation megatron-core[mlm,dev]
-    ```
+```bash
+pip install megatron-core
```
jaredcasper: Why do we need to pip install megatron-core to get started? You can just clone the repo and run.

ilml: yes this goes without saying , we want to keep the main readme simple and clean and showcase the ease of use of megatron per other's feedback.

deepakn94: Maybe include both options?

ilml: added. thanks

docs/get-started/install.md
```diff
@@ -7,90 +7,89 @@
...
-  -v /path/to/checkpoints:/workspace/checkpoints \
-  -e PIP_CONSTRAINT= \
-  nvcr.io/nvidia/pytorch:25.04-py3
+pip install "setuptools<80.0.0,>=77.0.0" "packaging>=24.2" pybind11 "torch>=2.6.0"
+pip install --no-build-isolation megatron-core[mlm,dev]
 ```
 
-## Pip Installation
+```{note}
+`--no-build-isolation` requires build dependencies to be pre-installed in the environment. `torch` is needed because several `[dev]` packages (`mamba-ssm`, `nv-grouped-gemm`, `transformer-engine`) import it at build time to compile CUDA kernels. Expect this step to take **20+ minutes** depending on your hardware. If you prefer pre-built binaries, the [NGC Container](#option-c-ngc-container) ships with these pre-compiled.
```
janEbert, line 57: I still think we should warn about memory use during build when the `MAX_JOBS` envvar is not set. People complained about this in the FlashAttention repo for a long time, so now [it's in the README](https://github.com/Dao-AILab/flash-attention/blob/ceb1099ee0b9d6b8ca9426d54d0b6d8598a48e19/README.md#L90-L96).

janEbert:
I'm still not sure – should every mention of LTS really be removed?

docs/get-started/install.md
janEbert: I suppose the same comment could be added to the README from-source install, too. :)
Or it could be linked as a "when you encounter issues" reference.

README.md
ilml: added, thanks

docs/get-started/install.md
ilml: fixed. thanks!

docs/get-started/install.md
```diff
@@ -7,90 +7,97 @@
...
+Building from source can consume a large amount of memory. By default the build runs one compiler job per CPU core, which may cause out-of-memory failures on machines with many cores. To limit parallel compilation jobs, set the `MAX_JOBS` environment variable before installing (e.g. `MAX_JOBS=4`).
+```
 
-- `dev`: Moving head that supports the most recent upstream dependencies
-- `lts`: Long-term support of NGC PyTorch 24.01
+To clone the repository for examples:
 
-Both containers can be combined with `mlm`, which adds package dependencies for Megatron-LM on top of Megatron Core.
+```bash
+git clone https://github.com/NVIDIA/Megatron-LM.git
```
deepakn94, line 71: This seems clunky? You pip install and then still clone?

docs/get-started/install.md
```diff
@@ -7,90 +7,97 @@
...
-        pip install "setuptools<80.0.0,>=77.0.0" "packaging>=24.2"
-        pip install --no-build-isolation megatron-core[mlm,dev]
-        ```
-* Install packages for LTS support NGC PyTorch 24.01
 
-        ```bash
-        pip install "setuptools<80.0.0,>=77.0.0" "packaging>=24.2"
-        pip install --no-build-isolation megatron-core[lts]
-        ```
+## Option C: NGC Container
```
deepakn94, line 97: Why is this option C? Isn't this the easiest way to get the required dependencies?

README.md
jaredcasper: Added how?

docs/get-started/install.md
ilml: per previous discussion with @sbhavani , we want to put the pip installation first to showcase the ease of use of megatron

docs/get-started/install.md
ilml: we have the quickstart part as the following, so this is just prepare the user for the examples.

README.md
ilml: see this commit 
https://github.com/NVIDIA/Megatron-LM/pull/3596/changes/13f3007544596a67c84eea3d8c4c265fc193e00a

ilml: @jaredcasper hey jared,  is this okay ?

ilml:
> I'm still not sure – should every mention of LTS really be removed?

@ko3n1g do you have any context ?

docs/get-started/install.md
```diff
@@ -7,90 +7,97 @@
...
+```
 
-This container comes with all dependencies pre-installed with compatible versions and optimized configurations for NVIDIA GPUs:
+To include optional training dependencies (Weights & Biases, SentencePiece, HF Transformers):
 
-- PyTorch (latest stable version)
-- CUDA, cuDNN, NCCL (latest stable versions)
-- Support for FP8 on NVIDIA Hopper, Ada, and Blackwell GPUs
-- For best performance, use NVIDIA Turing GPU architecture generations and later
+```bash
```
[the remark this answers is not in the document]
maanug-nv: agree with this, but i think the change should be on the line below this one

docs/get-started/install.md
```diff
@@ -7,90 +7,117 @@
...
+uv pip install "megatron-core[training]"
+```
 
-Megatron Core installation offers support for two NGC PyTorch containers:
+For all extras including [Transformer Engine](https://github.com/NVIDIA/TransformerEngine):
 
-- `dev`: Moving head that supports the most recent upstream dependencies
-- `lts`: Long-term support of NGC PyTorch 24.01
+```bash
+uv pip install "setuptools<80.0.0,>=77.0.0" "packaging>=24.2" pybind11 "torch>=2.6.0"
```
maanug-nv: i think probably makes sense to use the build group in case anything gets added in the future.
```suggestion
uv pip install --group "build"
```

ilml: fixed, thanks

megatron/core/QuickStart.md
```diff
@@ -21,8 +21,7 @@ This guide for Megatron Core walks you through the following tasks:
 
     git clone https://github.com/NVIDIA/Megatron-LM.git
     cd Megatron-LM
-    pip install -U "setuptools<80.0.0,>=77.0.0" packaging
-    pip install --no-build-isolation .[dev]
+    uv pip install -e .
```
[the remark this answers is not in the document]
ilml: fixed.

ko3n1g (approved):

Issue #1877, opened 2025-10-20
moe-token-dispatcher-type alltoall error


gaoshangle:
**Describe the bug**

Got runtime error when running moe megatron training on Ampere with:
--tensor-model-parallel-size 4
--expert-model-parallel-size 8
--expert-tensor-parallel-size 1
--moe-token-dispatcher-type alltoall

**Steps/Code to reproduce bug**

`[rank11]:     output = output.view(self.hidden_shape)
[rank11]: RuntimeError: shape '[2048, 1, 4096]' is invalid for input of size 0`

https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/moe/token_dispatcher.py#L803

BambiSheng:
Hi @gaoshangle I’m running into the same issue. Did you manage to resolve it? Thanks a lot!

Connor-XY:
Hello @gaoshangle and @BambiSheng, thanks for reporting the alltoall token-dispatcher shape failure.

The failing line suggests the dispatcher is trying to reshape an empty output buffer back to the original hidden shape. With `TP=4`, `EP=8`, `ETP=1`, and `moe-token-dispatcher-type=alltoall`, this likely happens when a rank/expert receives zero tokens for a local dispatch slice but the code still assumes a non-empty tensor for the final view.

This is already assigned to @yashaswikarnati. The maintainer follow-up should confirm whether the alltoall dispatcher needs an explicit zero-token fast path for this parallelism combination, or whether this argument combination is unsupported and should fail earlier with a clearer error.

PR #3626, opened 2026-02-26
fix: handle zero-size tensors in MoE token dispatchers

callum-ward-inflection:
# What does this PR do ?

Fixes MoE token dispatcher crash when an Expert Parallelism rank receives zero tokens from the router.

When an EP rank receives zero tokens from the router, the fused permute/unpermute autograd pair breaks and .view() crashes on a zero-size tensor. Switch to unfused path symmetrically, reconnect gradient graph for backward collectives, and guard .view() in all three dispatcher classes.

Fixes: NVIDIA/Megatron-LM#1877

## Changes

Three changes to `megatron/core/transformer/moe/token_dispatcher.py`:

1. **Symmetric unfused permute/unpermute for empty EP ranks** — TE's `fused_permute` saves state that `fused_unpermute` reads during backward. With zero tokens this state is invalid. Both `dispatch_postprocess` and `combine_preprocess` detect zero-token ranks and fall back to the unfused PyTorch path together (they must match — mixing fused permute with unfused unpermute crashes due to incompatible index formats).

2. **Gradient connectivity for backward collectives** — Unfused `unpermute` with zero tokens returns a tensor disconnected from the autograd graph. During backward, distributed collectives (AllGather/ReduceScatter) need every rank to participate. A detached tensor means one rank never triggers its collective and all others hang (NCCL timeout). Reconnected with `unpermuted_local_hidden + hidden_states.sum() * 0`.

3. **Safety guard on `.view()` for all three dispatcher classes** — `MoEAllGatherTokenDispatcher`, `MoEAlltoAllTokenDispatcher`, and `MoEFlexTokenDispatcher` all have `.view(self.hidden_shape)` that crashes on zero-size input. Guarded with a zero tensor of the correct shape.

## Reproduction

Reproducible with `openai/gpt-oss-20b` (32 experts, hidden_size=2880) SFT with TP=4, EP=2 on 16 GPUs. The pretrained router's weight distribution combined with reduced tokens-per-rank from TP=4 sequence parallelism causes certain EP ranks to consistently receive zero tokens.

## Testing

| Config | Result |
|--------|--------|
| TP=4, EP=2 (previously crashing) | **PASS** — 20 steps |
| TP=2, EP=2 | PASS |
| TP=1, EP=2 | PASS |
| TP=4, EP=4 | PASS |

### Pre-checks

### DIFF
--- a/megatron/core/transformer/moe/token_dispatcher.py
+++ b/megatron/core/transformer/moe/token_dispatcher.py
@@ -294,12 +294,17 @@ class MoEAllGatherTokenDispatcher(MoETokenDispatcher):
 
         tokens_per_expert = self.local_map.sum(dim=0).long().cpu()
 
+        # TE's fused_permute saves autograd state that fused_unpermute reads during backward.
+        # With zero tokens this state is invalid — fall back to unfused path symmetrically.
+        self._empty_local_dispatch = tokens_per_expert.sum().item() == 0
+        use_fused = self.config.moe_permute_fusion and not self._empty_local_dispatch
+
         permuted_local_hidden_states, _, self.reversed_local_input_permutation_mapping, _, _ = (
             permute(
                 hidden_states,
                 self.local_map,
                 num_out_tokens=tokens_per_expert.sum().item(),
-                fused=self.config.moe_permute_fusion,
+                fused=use_fused,
             )
         )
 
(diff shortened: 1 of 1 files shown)


Victarry:
Hi @callum-ward-inflection , I think the ideal way to resolve this issue is to fix the fused permute kernel. Adding more conditions in the token dispatcher will make it more complex.

Could you post the detailed error you facing with zero-tokens? IIRC, the TE fused permute kernel should support the zero-token case. cc @hxbai

hxbai:
I agree with @Victarry , it is better to create a fix to TE's permute function rather than to MCore.

Phlip79:
@callum-ward-inflection any update here? Otherwise, I will close this PR.

PR #3634, opened 2026-02-27
[GDN] Keep `A_log` and `out_norm` in FP32 for numerical stability

cuichenx:
# What does this PR do ?
### Summary

Qwen3.5 (the successor to Qwen3-Next) stores the GDN `A_log` and `out_norm` (output layernorm) parameters in FP32 in its released checkpoints, while the rest of the model is BF16. The current GatedDeltaNet implementation allocates these in `config.params_dtype`, causing dtype mismatches during checkpoint conversion. This PR aligns the Megatron implementation with the upstream precision convention.

- Store `A_log` parameter in FP32 instead of `config.params_dtype`
- Cast `out_norm` (output layernorm) to FP32 after construction
- Keep `A_log` initialization in FP32 in `reset_parameters`
- No changes to `dt_bias` (remains at `config.params_dtype`)

### Motivation

The Qwen3.5 family (both the 27B dense and three MoE variants) releases `A_log` and `linear_attn.norm.weight` in FP32, even though the rest of the checkpoint is BF16. This is a deliberate precision choice by the model authors for numerical stability.

The GDN forward pass already computes the decay gate `g` in FP32:

```python
g = -self.A_log.exp() * F.softplus(alpha.float() + self.dt_bias)  # In fp32
```

However, the `A_log` parameter and `out_norm` weights were previously allocated in `config.params_dtype` (typically BF16). This creates two problems:

1. **Precision loss in `A_log`**: `A_log` stores `log(A)` where `A` is a per-head decay rate. The exponential `A_log.exp()` amplifies quantization error from BF16 storage, even though the compute is done in FP32.

2. **Precision loss in `out_norm`**: The output RMSNorm applied after the recurrence operates on per-head-dim scale factors. Storing these in BF16 introduces unnecessary rounding.

Both parameters are small (per-head or per-head-dim vectors), so the FP32 memory overhead is negligible.

### Changes

**`megatron/core/ssm/gated_delta_net.py`**:

- `__init__`: `A_log` allocated with `dtype=torch.float32` (was `config.params_dtype`)
- `__init__`: `out_norm` cast to FP32 via `self.out_norm.to(torch.float32)` after `build_module`
- `reset_parameters`: `A_log` initialization uses `dtype=torch.float32` (was `config.params_dtype`)

### Precedent

This follows the existing convention in `MambaMixer`, where `A_log` is already unconditionally stored in FP32:

```python
# mamba_mixer.py, line 336-341
A = torch.empty(self.nheads_local_tp, dtype=torch.float32, ...)
A_log = torch.log(A)  # Keep A_log in fp32
self.A_log = nn.Parameter(A_log)
```

### Alternative designs considered

**Option A: Configurable via TransformerConfig flags**

Add two boolean fields (`linear_A_log_in_fp32`, `linear_out_norm_in_fp32`) to `TransformerConfig`, defaulting to `True`, and have `GatedDeltaNet` read them.

- Pro: Models that need BF16 for these params (e.g., Qwen3-Next) can set the flags to `False` for bitwise-identical checkpoint roundtrips.
- Con: `TransformerConfig` already has ~197 fields. Adding per-parameter precision knobs for tiny vectors (one scalar per head) is hard to justify given the negligible memory difference. Every GDN-based model bridge must also remember to set these flags correctly.

**Why we chose hardcoded FP32 instead:**

- The memory overhead is negligible (these are per-head vectors, not weight matrices).
- The forward pass already computes in FP32 regardless.
- `MambaMixer` already hardcodes `A_log` to FP32 — this aligns the two SSM-family modules.
- For models that previously stored these in BF16 (e.g., Qwen3-Next), the BF16-to-FP32 upcast is lossless (all BF16 values are exactly representable in FP32), so checkpoint loading is unaffected. Export will produce FP32 values that are mathematically identical to the original BF16 values.
- Qwen3-Next is superseded by Qwen3.5, which already uses FP32 for these params.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/ssm/gated_delta_net.py
+++ b/megatron/core/ssm/gated_delta_net.py
@@ -188,23 +188,22 @@ class GatedDeltaNet(MegatronModule):
             )
         )
         setattr(self.dt_bias, "tensor_model_parallel", True)
-        # A_log parameter
+        # A_log is kept in FP32 for numerical stability
         self.A_log = nn.Parameter(
             torch.empty(
-                self.num_v_heads_local_tp,
-                dtype=config.params_dtype,
-                device=torch.cuda.current_device(),
+                self.num_v_heads_local_tp, dtype=torch.float32, device=torch.cuda.current_device()
             )
         )
         setattr(self.A_log, "tensor_model_parallel", True)
 
-        # Output layernorm before projection
+        # Output layernorm before projection — kept in FP32 to match HF checkpoint precision
         self.out_norm = build_module(
             submodules.out_norm,
             config=self.config,
             hidden_size=self.value_head_dim,
             eps=self.config.layernorm_epsilon,
         )
+        self.out_norm.to(torch.float32)
 
         self.out_proj = build_module(
             submodules.out_proj,
@@ -238,10 +237,10 @@ class GatedDeltaNet(MegatronModule):
                     dtype=self.config.params_dtype,
                     device=torch.cuda.current_device(),
                 )
-                # A_log
+                # A_log (FP32)
                 A = torch.empty(
                     self.num_v_heads_local_tp,
-                    dtype=self.config.params_dtype,
+                    dtype=torch.float32,
                     device=torch.cuda.current_device(),
                 ).uniform_(*self.A_init_range)
                 self.A_log.data.copy_(torch.log(A))


yuzhongw-nvidia:
Hi @cuichenx , thanks for your work, it is a very good catch. But I'm afraid that your change may not work because the model will be wrapped with `Float16Module` [here](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/training.py#L1306-L1309) so the params will be converted back to bfloat16.

Please refer to moe expert bias ([impl](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/moe/router.py#L218-L227), [usage](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/moe/router.py#L680-L712)) to refine this PR. And please also help double-check whether the two dtypes are correct in an E2E training loop. Thanks!

yuzhongw-nvidia:
Also, please help double-check whether the two changes are correct. My observation is that, although the tensor dtype of GDN out norm and A_log are fp32 in the checkpoint file, they are still constructed and maintained as bf16 tensors in the code of HF ([code](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py#L490)).

I run an example to verify it as below.

```
>>> model = transformers.AutoModelForCausalLM.from_pretrained("Qwen/Qwen3.5-35B-A3B")
>>> model.model.layers[0].linear_attn.A_log.dtype
torch.bfloat16
>>> model.model.layers[0].linear_attn.norm.weight.dtype
torch.bfloat16
```

I'm not sure which one is correct. Could you please help find out what data type does qwen3.5 use here on earth? Thanks!

cuichenx:
Thanks @yuzhongw-nvidia, marking this as draft for now

Issue #3644, opened 2026-02-28
Enable chunked MLP during training


pengdurice:
**Is your feature request related to a problem? Please describe.**
During training, especially when the context window is long, we usually run into OOM issues. The OOM may happen during the mlp layer (dense or MOE). 

Tag the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) 
to get oncall's attention to this issue.

**Describe the solution you'd like**
It is noticed that there is a chunked MLP solution in prefill phase under inference. However the same trick can be applied to the training.
```
        should_chunk_mlp_for_prefill = (
            self.config.mlp_chunks_for_prefill > 1
            and inference_context is not None
            and not inference_context.is_decode_only()
            and not isinstance(self.mlp, IdentityOp)
            and not self.config.transformer_impl == "inference_optimized"
        )
        ....
        elif should_chunk_mlp_for_prefill:
            # Chunk input along sequence dimension
            num_chunks = min(self.config.mlp_chunks_for_prefill, pre_mlp_layernorm_output.shape[0])
            chunks = pre_mlp_layernorm_output.chunk(num_chunks, dim=0)
            ....
            mlp_output_with_bias = (mlp_output, bias_output)
```

**Describe alternatives you've considered**
I tried directly handle the efficiency issues inside of the MLP layer. However this simple chunking trick can altogether reduce the memory peak for activations. 

**Additional context**

Phlip79:
Have you taken a look at our [parallelism guide](https://docs.nvidia.com/megatron-core/developer-guide/latest/user-guide/parallelism-guide.html)? Specifically, the context parallel section may be helpful for you.

pengdurice:
Hi @Phlip79 , thank you for chiming in! However, enabling CP = N will require N times more GPUs. I have tested different tp, cp, pp, dp settings and nothing worked. Given a fixed number of GPUs, we may just shift the memory pressure from one place to another. Enabling this chunked MLP during training will help reduce the activation peak memory usage with the downside of being a bit slower. WDYT? thanks!

Phlip79:
Thanks for creating the PR. I have provided comments there.

ericharper:
https://github.com/NVIDIA/Megatron-LM/pull/3656 merged

PR #3656, opened 2026-03-02
Add logic to enable chunked MLP during training

pengdurice:
# What does this PR do ?

This PR enables chunked MLP during training phases (on top of prefilling of inference phase)
## related issue:
https://github.com/NVIDIA/Megatron-LM/issues/3644
## Tests:
Added unit tests and passed.
Applied to a multi node GLM4.7 model training and unblocked OOM with context window > 30K
## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR

### DIFF
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -972,7 +972,7 @@ class TransformerConfig(ModelParallelConfig):
     mlp_chunks_for_training: int = 1
     """The number of chunks along the sequence dimension to use for MLP computation
     during training."""
-    
+
     heterogeneous_block_specs: bool = False
     """Whether to use heterogeneous block specs (nemotron-nas architecture)."""
 
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -969,7 +969,10 @@ class TransformerConfig(ModelParallelConfig):
     mlp_chunks_for_prefill: int = 1
     """The number of chunks along the sequence dimension to use for MLP computation
     during prefill."""
-
+    mlp_chunks_for_training: int = 1
+    """The number of chunks along the sequence dimension to use for MLP computation
+    during training."""
+    
     heterogeneous_block_specs: bool = False
     """Whether to use heterogeneous block specs (nemotron-nas architecture)."""
 
--- a/megatron/core/transformer/transformer_layer.py
+++ b/megatron/core/transformer/transformer_layer.py
@@ -822,7 +822,12 @@ class TransformerLayer(GraphableMegatronModule, BaseTransformerLayer):
             # Aggregate chunk outputs
             mlp_output = torch.cat([out for out, _ in outputs], dim=0)
             bias_chunks = [bias for _, bias in outputs if bias is not None]
-            bias_output = bias_chunks[0] if bias_chunks else None
+            bias_output = torch.stack(bias_chunks, dim=0).sum(dim=0) if bias_chunks else None
+            if bias_output is not None:
+                for i in range(1, len(bias_chunks)):
+                    assert torch.allclose(bias_chunks[i], bias_chunks[0])
+            if self.config.cuda_graph_impl == 'none':
+                bias_output = bias_chunks[0] if bias_chunks else None
             mlp_output_with_bias = (mlp_output, bias_output)
         else:
             if using_fused_tp_inference_kernel:
--- a/megatron/core/transformer/transformer_layer.py
+++ b/megatron/core/transformer/transformer_layer.py
@@ -823,10 +823,8 @@ class TransformerLayer(GraphableMegatronModule, BaseTransformerLayer):
             mlp_output = torch.cat([out for out, _ in outputs], dim=0)
             bias_chunks = [bias for _, bias in outputs if bias is not None]
             bias_output = torch.stack(bias_chunks, dim=0).sum(dim=0) if bias_chunks else None
-            if bias_output is not None:
-                for i in range(1, len(bias_chunks)):
-                    assert torch.allclose(bias_chunks[i], bias_chunks[0])
             if self.config.cuda_graph_impl == 'none':
+                # elements in bias_chunks are the same for all chunks, so we can just use the first one
                 bias_output = bias_chunks[0] if bias_chunks else None
             mlp_output_with_bias = (mlp_output, bias_output)
         else:
--- a/megatron/core/transformer/transformer_layer.py
+++ b/megatron/core/transformer/transformer_layer.py
@@ -822,10 +822,8 @@ class TransformerLayer(GraphableMegatronModule, BaseTransformerLayer):
             # Aggregate chunk outputs
             mlp_output = torch.cat([out for out, _ in outputs], dim=0)
             bias_chunks = [bias for _, bias in outputs if bias is not None]
-            bias_output = torch.stack(bias_chunks, dim=0).sum(dim=0) if bias_chunks else None
-            if self.config.cuda_graph_impl == 'none':
-                # elements in bias_chunks are the same for all chunks, so we can just use the first one
-                bias_output = bias_chunks[0] if bias_chunks else None
+            # elements in bias_chunks are the same for all chunks, so we can just use the first one
+            bias_output = bias_chunks[0] if bias_chunks else None
             mlp_output_with_bias = (mlp_output, bias_output)
         else:
             if using_fused_tp_inference_kernel:
--- a/megatron/core/transformer/transformer_layer.py
+++ b/megatron/core/transformer/transformer_layer.py
@@ -773,6 +773,11 @@ class TransformerLayer(GraphableMegatronModule, BaseTransformerLayer):
             and not isinstance(self.mlp, IdentityOp)
             and not self.config.transformer_impl == "inference_optimized"
         )
+        should_chunk_mlp_for_training = (
+            self.config.mlp_chunks_for_training > 1
+            and inference_context is None
+            and not isinstance(self.mlp, IdentityOp)
+        )
 
         using_fused_tp_inference_kernel = (not self.training) and (
             self.config.inference_fuse_tp_communication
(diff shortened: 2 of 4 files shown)


pengdurice:
@Phlip79 , Hi, would you mind reviewing this PR? thanks!

Phlip79 (changes requested):
Could you update the `test()` helper to pass `mlp_chunks_for_training` into `TransformerConfig` instead of hardcoding it to 4, so the loop actually verifies that chunked and non-chunked MLP produce identical outputs?

pengdurice:
> Could you update the `test()` helper to pass `mlp_chunks_for_training` into `TransformerConfig` instead of hardcoding it to 4, so the loop actually verifies that chunked and non-chunked MLP produce identical outputs?

Hi @Phlip79 , thank you for the review, fixed it;-)

megatron/core/transformer/transformer_layer.py
```diff
@@ -742,6 +742,11 @@ def _forward_mlp(
             and not isinstance(self.mlp, IdentityOp)
             and not self.config.transformer_impl == "inference_optimized"
         )
+        should_chunk_mlp_for_training = (
+            self.config.mlp_chunks_for_training > 1
+            and inference_context is None
```
Phlip79, line 778: Does this need `and self.training`?

tests/unit_tests/transformer/test_transformer_layer.py
```diff
@@ -130,7 +133,25 @@ def test(
...
+                    input_hidden_states,
+                    inference_context,
+                )
+                assert hidden_states.shape[0] == sequence_length
+                assert hidden_states.shape[1] == micro_batch_size
+                assert hidden_states.shape[2] == hidden_size
+
+                outputs[mlp_chunks_for_training] = (hidden_states, context)
+
+            assert torch.equal(outputs[1][0], outputs[4][0])
```
Phlip79, line 131: Each call to `test()` creates a new `TransformerLayer` with fresh random weights. I think the better approach is to create it before the call to `test()` and re-use it.

Phlip79 (changes requested):

megatron/core/transformer/transformer_layer.py
pengdurice: I thought inference_context and self.training was mutually exclusive, yes, sure, let me add it.

tests/unit_tests/transformer/test_transformer_layer.py
pengdurice: Thanks! I brought the code from test outside and then directly call the created transformer layer. I also locally tested if `transformer_config.mlp_chunks_for_training = mlp_chunks_for_training` works by printing the num_chunk (not shown here.) and verified it works.  At the same time, I changed inputs from torch.ones to torch.randn with a fixed seed to further verify that it works. LMK if you have any more questions, thanks!

Phlip79 (approved):

pengdurice:
@Phlip79  Hi, thanks again for reviewing this PR. I’ve fixed a lint issue—would appreciate another set of eyes when you have time!

pengdurice:
@Phlip79 I saw several errors from the CICD run I downloaded the errors and checked and believe they fail on other places and probably due to that this PR is too far behind main. I rebased and fixed the lint errors again. Once you've got a chance, would you please trigger the CI/CD again? thank you!

pengdurice:
@Phlip79 Hi, sorry for the back and forth! The lint check failed on the last CI test due to that I merged the main instead of rebasing on main, and I also tried to fix the lint errors causing more files to be changed which then trigger even more lint errors. The clean solution is to cherry pick only my commits (my 4 files' changes) and rebase on main. Exactly what I did now. 
I tested locally confirmed that the tests still passed and also the following passed:
Now the PR contains only the 4 file changes.  
> pre-commit run --files \
>   megatron/core/transformer/transformer_config.py \
>   megatron/core/transformer/transformer_layer.py \
>   tests/unit_tests/models/test_mamba_moe_model.py \
>   tests/unit_tests/transformer/test_transformer_layer.py
>   black....................................................................Passed
>   pylint...................................................................Passed
>   isort....................................................................Passed

Would you please trigger CI again? thank you so much!

pengdurice:
> /ok to test [2cb97c5](https://github.com/NVIDIA/Megatron-LM/commit/2cb97c5bbeb2fd7f7beb13b8fd8e138e28886bd5)

@Phlip79 thank you for triggering the test. Would you mind take another look at it as most tests passed (other than approval workflow)? Thank you so much!

Phlip79:
All good now. Need review groups to take a look now.

pengdurice:
> All good now. Need review groups to take a look now.

Great, thank you, let me know if the team needs anything from me;-)

megatron/core/transformer/transformer_config.py
```diff
@@ -969,6 +969,9 @@ class TransformerConfig(ModelParallelConfig):
     mlp_chunks_for_prefill: int = 1
     """The number of chunks along the sequence dimension to use for MLP computation
     during prefill."""
+    mlp_chunks_for_training: int = 1
+    """The number of chunks along the sequence dimension to use for MLP computation
+    during training."""
```
santhnm2, line 974: Rather than having a separate flag, perhaps we should just consolidate into a single `mlp_chunks` flag and then enable it either when we are in training mode (i.e., `inference_context` is `None`) or we are in prefill mode (`inference_context` is not `None` and `inference_context.is_decode_only() == False`). The only scenario where we would hypothetically need two separate flags is if we were doing RL and wanted the training to chunk MLP but the inference to not chunk MLP (or vice versa). This seems like a fairly contrived setting though. @jon-barker any thoughts here?

jon-barker: I have never thought about this in the RL context and don't really have an opinion how likely we are to want to do this. But it seems like maybe the proposed interface is cleaner anyway and leaves open the option of setting it separately for inference and training.

santhnm2: Sure, we can keep it as is for now. If we decide to consolidate later we can do that in a backwards-compatible way.

tests/unit_tests/transformer/test_transformer_layer.py
```diff
@@ -64,73 +64,71 @@ def test_gpu_forward(self):
...
+                    hidden_states=input_hidden_states,
+                    attention_mask=attention_mask,
+                    inference_context=None,
+                )
+                assert hidden_states.shape[0] == sequence_length
+                assert hidden_states.shape[1] == micro_batch_size
+                assert hidden_states.shape[2] == hidden_size
+                outputs[mlp_chunks_for_training] = (hidden_states, context)
+
+            assert torch.equal(outputs[1][0], outputs[4][0])
```
santhnm2, line 131: For training would it also make sense to add gradient equivalence tests?

pengdurice: Thank you for the great suggestion! Just added. Actually, this helped identified a bug in the existing profiling chunking logic in this line 
`bias_output = torch.stack(bias_chunks, dim=0).sum(dim=0) if bias_chunks else None`
It may not manifest for inference, but will impact training since bias_chunks is the chunks of identical bias parameter and we should just take one of them rather than stacking them.

megatron/core/transformer/transformer_layer.py
```diff
@@ -808,7 +822,7 @@ def _forward_mlp(
             # Aggregate chunk outputs
             mlp_output = torch.cat([out for out, _ in outputs], dim=0)
             bias_chunks = [bias for _, bias in outputs if bias is not None]
-            bias_output = torch.stack(bias_chunks, dim=0).sum(dim=0) if bias_chunks else None
+            bias_output = bias_chunks[0] if bias_chunks else None
```
deepakn94: Can you add an assertion before this line checking that they're equal?

deepakn94: Let's add this behind a `if self.config.cuda_graph_impl == 'none'` guard to make sure we don't interfere with CG capture / replay!

pengdurice: I feel adding an assertion is causing additional overheads and actually these biases are the same bias instance (if any) from the same mlp layer. To be on the safe side , I reverted to the original code and do: 
```
if self.config.cuda_graph_impl == 'none': 
     bias_output = bias_chunks[0] if bias_chunks else None
```
Does this make sense? thank you!

pengdurice: @deepakn94 , would you mind take a look at this when you have time? thank you!

megatron/core/transformer/transformer_layer.py
```diff
@@ -809,6 +823,11 @@ def _forward_mlp(
             mlp_output = torch.cat([out for out, _ in outputs], dim=0)
             bias_chunks = [bias for _, bias in outputs if bias is not None]
             bias_output = torch.stack(bias_chunks, dim=0).sum(dim=0) if bias_chunks else None
+            if bias_output is not None:
+                for i in range(1, len(bias_chunks)):
+                    assert torch.allclose(bias_chunks[i], bias_chunks[0])
+            if self.config.cuda_graph_impl == 'none':
+                bias_output = bias_chunks[0] if bias_chunks else None
```
deepakn94: Shouldn't this always be set?

I meant the assertion can be in a `if self.config.cuda_graph_impl == 'none':` block. But on second thought, let's just add a comment saying they should all be identical and remove the assertion.

pengdurice: Hi @deepakn94 thank you for the review, just did that!

deepakn94: Let's get rid of `if self.config.cuda_graph_impl == 'none':` too.

pengdurice: SG! Now it defaults to instead of doing the sum, it uses the first element with a comment, thank you!

deepakn94: Approving from my side.

deepakn94 (approved):

santhnm2 (approved):

pengdurice:
Happy Monday! @deepakn94  and @santhnm2 thank you guys so much for reviewing and approving! @Phlip79 would you mind kicking off another CICD test? thank you!

pengdurice:
@santhnm2 , thank you for kicking off the CICD, it still failed approval workflow from more-oncall and core-nemo. Anything that I can do? Do we need to involve additional reviewers? thank you!

ericharper (approved):

ericharper:
> @santhnm2 , thank you for kicking off the CICD, it still failed approval workflow from more-oncall and core-nemo. Anything that I can do? Do we need to involve additional reviewers? thank you!

Just approved. mcore-oncall is not required and that's just for our internal systems to triage.

PR #3680, opened 2026-03-03
Improve error logging when invalid number of tokens is requested.

yobibyte:
Log actual error that happens on the inference server to help with the debugging. Right now it is impossible to change your hyperparameters / understand the actual reason of the failure.


yobibyte:
> Can you add the request ID to the logging statements?

Good idea, how does this look?

The outputs look like this:

```bash
ERROR:root:request_id=12 Invalid number of tokens to generate. Prompt len: 84, tokens to generate: -44, max seq len: 40.                                                                                               
ERROR:root:request_id=7 Invalid number of tokens to generate. Prompt len: 84, tokens to generate: -44, max seq len: 40.                                                                                                
ERROR:root:request_id=11 Invalid number of tokens to generate. Prompt len: 87, tokens to generate: -47, max seq len: 40.    
```

PR #3689, opened 2026-03-04
Add Engram model structure integration (v1)

ilml:
> **For reviewers:**
> 1. This PR focuses on model implementation and inference correctness only. Training support (builder, entry point, CLI args) will follow in a subsequent PR.
> 2. We have verified numerical correctness by comparing our `EngramModule` against the [official reference implementation](https://github.com/deepseek-ai/Engram) with identical weights and inputs — outputs match exactly via `torch.testing.assert_close`.

## Summary

- Integrate DeepSeek's [Engram](https://github.com/deepseek-ai/Engram) n-gram hash embedding module into Megatron-LM as a new model type under `megatron/core/models/engram/`
- Extends GPTModel with Engram-augmented transformer layers that inject gated n-gram embeddings before self-attention at configurable layer positions
- **Numerical match verified**: unit test copies the reference `engram_demo_v1.py` verbatim, loads identical weights into our `EngramModule`, and confirms output matches via `torch.testing.assert_close`

## Components

| File | Purpose |
|---|---|
| `megatron/core/models/engram/engram_module.py` | Core: CompressedTokenizer, NgramHashMapping, MultiHeadEmbedding, ShortConv, EngramModule |
| `megatron/core/models/engram/engram_model.py` | EngramTransformerLayer + EngramGPTModel (thin wrappers over TransformerLayer/GPTModel) |
| `megatron/core/models/engram/test_engram.py` | Numerical comparison test against reference implementation |

## Design decisions

- **Extends GPTModel** (not LanguageModule) since Engram is fundamentally GPT + extra module in specific layers
- **Two-phase forward**: hash IDs pre-computed at model level (CPU/numpy), embeddings cached in each EngramModule, consumed during layer forward — avoids modifying TransformerBlock's interface
- **HC_MULT** (hyper-connection multiplier) kept from reference to match the architecture, though without real mHC the HC slots receive replicated hidden states
- **Uses sympy.isprime** directly (matching reference) instead of custom primality test
- **No core Megatron files modified** — pure inheritance + ModuleSpec system

## Known limitations (v1)

- HC dimension is faked (replicated from single hidden state) since Megatron lacks native hyper-connections — gates still differ per slot due to separate key projections
- Engram embedding tables are not tensor-parallel-sharded
- Hash computation runs on CPU (numpy)
- Builder and layer specs not yet wired into training entry point

## Test plan

- [x] Numerical comparison: reference `Engram` and our `EngramModule` loaded with identical weights produce identical output for the same input (`torch.testing.assert_close` passes)
- [x] Test requires DeepSeek-V3 tokenizer (skipped gracefully if unavailable)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/megatron/core/models/engram/run_gpt_mock_baseline.sh
+++ b/megatron/core/models/engram/run_gpt_mock_baseline.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+MEGATRON="${MEGATRON:-$(cd -- "${SCRIPT_DIR}/../../../.." && pwd)}"
+HF_TOKENIZER="${HF_TOKENIZER:-openai-community/gpt2}"
+TB_ROOT="${TB_ROOT:-${MEGATRON}/tensorboard_runs}"
+RUN_GROUP="${RUN_GROUP:-gpt_mock_baseline}"
+RUN_STAMP="${RUN_STAMP:-$(date +%Y%m%d_%H%M%S)}"
+RUN_NAME="${RUN_NAME:-${RUN_GROUP}_${RUN_STAMP}}"
+
+mkdir -p "${TB_ROOT}"
+cd "${MEGATRON}"
+
+export PYTHONPATH="${MEGATRON}:${PYTHONPATH:-}"
+export TORCH_COMPILE_DISABLE="${TORCH_COMPILE_DISABLE:-1}"
+
+echo "Writing TensorBoard logs to ${TB_ROOT}/${RUN_NAME}"
+
+COMMON_ARGS=(
+  --transformer-impl local
+  --tensor-model-parallel-size 1
+  --pipeline-model-parallel-size 1
+  --num-layers 2
+  --hidden-size 256
+  --ffn-hidden-size 1024
+  --num-attention-heads 4
+  --seq-length 128
+  --max-position-embeddings 128
+  --micro-batch-size 1
+  --global-batch-size 1
+  --train-iters 100
+  --lr 1e-4
+  --min-lr 1e-5
+  --lr-decay-style cosine
+  --weight-decay 0.01
+  --clip-grad 1.0
+  --mock-data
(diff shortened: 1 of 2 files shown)


megatron/core/models/engram/test_engram.py
yanring, line 1: This should be better in the unit test folder.

yanring:
I'd like to defer to @Victarry to help take a deeper look.

ilml:
sync with @Victarry
1. inheritance from gpt model (pp?)
2.e2e runnable: loss verification. 
    2.1 TODO: reuse pretrain_gpt.py

ilml:
Summary from local E2E checks on this branch:

- Verified the `pretrain_gpt.py` integration is runnable end-to-end through the builder + CLI path (tested in a TP=1 / PP=1 smoke setup). The Engram path is genuinely active: during model construction / parameter listing we see `decoder.layers.*.engram.*` parameters, so this is not a dead code path.
- Added simple baseline vs Engram launch scripts under `megatron/core/models/engram/` to make the smoke comparison reproducible.
- Important caveat: Megatron `--mock-data` is useful for wiring / smoke validation only. It generates deterministic synthetic counting-style token streams, not natural text / knowledge-heavy data, so it is not a reliable signal for whether Engram should lower LM loss.
- We did align the baseline and Engram smoke scripts on tokenizer / seed / data args, so they consume the same mock-data stream. However, because Engram adds extra modules, `same seed` does not imply identical backbone initialization; mock-loss deltas should therefore not be over-interpreted as a model-quality result.
- The Engram paper's training results are on real-language pretraining with an iso-parameter / iso-FLOPs MoE comparison and extra recipe details (for example conv-zero-init / Engram-specific optimizer treatment), so the local mock-data test here should not be treated as evidence for or against the paper claim: [Conditional Memory via Scalable Lookup: A New Axis of Sparsity for Large Language Models](https://arxiv.org/pdf/2601.07372v1).

TL;DR: this branch now has a working E2E / smoke-training path for the Engram integration, but a meaningful effectiveness claim still needs a real-data experiment that is closer to the paper recipe.

ilml:
loss curve:
image

ilml:
wait for @Victarry to reivew

Issue #3670, opened 2026-03-03
Gradient synchronization incorrect when `--overlap-grad-reduce` and `--num-distributed-optimizer-instances > 1` due to autograd hook stream affinity


zyeric:
### Describe the bug

In Megatron-LM v0.15.2 (and likely the affected commit range), enabling both `--overlap-grad-reduce` and `--num-distributed-optimizer-instances > 1` results in incorrectly synchronized gradients. The root cause is a stream synchronization issue in the distributed data parallel (DDP) gradient post‑hook.

When `DistributedDataParallel` is constructed, it is wrapped in a custom CUDA stream context (introduced in PR #e7c55de). Inside that context, a temporary tensor is created via `expand_as` to retrieve the gradient accumulator (`grad_acc`). Because `expand_as` is executed on a non‑default stream, the `grad_acc` object inherits that stream’s affinity. Consequently, the post‑hook registered on `grad_acc` runs on the same side stream during backward.

The hook contains gradient checks and scaling that must complete **before** the reduce‑scatter operation on the communication stream starts. However, because the hook runs on a side stream and no explicit synchronization is performed, the communication stream may launch reduce‑scatter before the hook finishes, leading to incorrect gradient values.

This problem only manifests when `--num-distributed-optimizer-instances > 1` because that setting changes the way gradient communication is orchestrated, making it sensitive to the stream ordering.

---

### Steps/Code to reproduce bug

A minimal unit test that demonstrates the underlying stream affinity issue (without requiring a full training run) is provided below. Save it as `test_stream_affinity.py` and run with `pytest -s` or `python -m unittest test_stream_affinity.py`.

```python
import torch
import unittest

class TestAutogradStreamAffinity(unittest.TestCase):
    def setUp(self):
        if not torch.cuda.is_available():
            self.skipTest("CUDA not available")
        self.device = 'cuda'

    def _make_hook(self, label):
        def hook(grad_out, grad_in):
            current_id = torch.cuda.current_stream().stream_id
            print(f"\n[{label}] Hook running on stream {current_id}")
            return grad_out
        return hook

    def test_case_1_pure_registration(self):
        """Node created in default stream, hook registered in side stream."""
        param = torch.randn(10, requires_grad=True, device=self.device)
        param_tmp = param.expand_as(param)          # default stream
        grad_acc = param_tmp.grad_fn.next_functions[0][0]

        custom_stream = torch.cuda.Stream()
        with torch.cuda.stream(custom_stream):
[...]

PR #3693, opened 2026-03-04
Fix DDP bug with --overlap-grad-reduce and --num-distributed-optimizer-instances > 1

wplf:
Thanks for [zyeric](https://github.com/zyeric)'s issue, you can see more details in https://github.com/NVIDIA/Megatron-LM/issues/3670.

## Bug reproduce  and fix 
image


wplf:
Hi @zyeric,  Please review this PR and thank you for your contribution again.

yaox12 (approved):

shjwudp (approved):
LGTM

wplf:
/ok to test [6a4bec3](https://github.com/NVIDIA/Megatron-LM/pull/3693/commits/6a4bec3f6f8fb4eadc928e3bbb971249dd7a8bce)

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -567,10 +567,10 @@ def start_grad_sync(self, force_all_reduce: Optional[bool] = False):
             # need to overlap communication.
             stream_context = torch.cuda.stream(self.communication_stream)
 
-            # The RS/AR communication stream needs to wait for the default stream
+            # The RS/AR communication stream needs to wait for the current stream
             # to complete its gradient computation before launching the next
             # gradient reduction collective.
-            self.communication_stream.wait_stream(torch.cuda.default_stream())
+            self.communication_stream.wait_stream(torch.cuda.current_stream())
```
sanandaraj5597, line 573: Why do we need this? I don't understand the difference as we submit all work on default stream. In the training case default stream and current stream are the same right?

sanandaraj5597 (approved):
LGTM.

deepakn94 (approved):

yaoyu-33 (approved):

megatron/core/distributed/param_and_grad_buffer.py
wplf: Default stream and current stream is not different when running.
I print these two stream and get default stream id=0, current stream id=91.

The problematic change was introduced in PR: [e7c55de](https://github.com/NVIDIA/Megatron-LM/commit/e7c55de99ed24278dcc2ed44455bf61ac2d95383) (wrapping DDP construction in a CUDA stream context).

The root cause is that expand_as inside that context makes the gradient accumulator inherit the side stream, causing the post‑hook to run on that stream.
- A fix has been implemented (or is proposed) that replaces the synchronisation from waiting on the - default stream to waiting on the current stream inside the hook. This ensures the communication stream waits for the hook’s stream before launching reduce‑scatter.
- The unit test above isolates the stream‑affinity behaviour and can be used to verify the fix.

self.communication_stream.wait_stream(torch.cuda.current_stream())
The first wait is ok. Because the grad acc's side stream (current stream) is absolutely after default stream.

torch.cuda.current_stream().wait_stream(self.communication_stream)
The second wait is ok, because default stream and current stream are the same.
image


Before fix, the grad acc and comm is kind of overlap.
image


After Fix.
image

ericharper:
Fixes #3670

PR #3698, opened 2026-03-04
Fix `config.softmax_scale` not being considered

janEbert:
Previously, `config.softmax_scale` was sometimes ignored and the default used unconditionally. This is fine in most cases, but when `softmax_scale` is set, we do – of course – want to use it.

### DIFF
--- a/tests/unit_tests/transformer/test_softmax_scale.py
+++ b/tests/unit_tests/transformer/test_softmax_scale.py
@@ -1,0 +1,201 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+"""Tests for softmax_scale config propagation.
+
+Verifies that TransformerConfig.softmax_scale is properly considered
+in both Attention.flash_decode_and_prefill and MLASelfAttention.__init__.
+"""
+
+import math
+from unittest.mock import patch
+
+import pytest
+import torch
+
+from megatron.core.extensions.transformer_engine import TEDotProductAttention
+from megatron.core.models.common.embeddings import _yarn_get_mscale
+from megatron.core.models.gpt.gpt_layer_specs import (
+    get_gpt_layer_with_transformer_engine_submodules,
+)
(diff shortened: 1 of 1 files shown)


megatron/core/transformer/multi_latent_attention.py
```diff
@@ -125,7 +125,15 @@ def __init__(
...
 
         mscale = _yarn_get_mscale(self.config.rotary_scaling_factor, self.config.mscale_all_dim)
-        self.softmax_scale = mscale * mscale / math.sqrt(self.q_head_dim)
+        self.softmax_scale = (
+            mscale
+            * mscale
+            * (
+                (1 / math.sqrt(self.q_head_dim))
+                if self.config.softmax_scale is None
+                else self.config.softmax_scale
```
jaredcasper, line 134: I'm not super familiar with the details of MLA... does it even make sense to have a configured softmax_scale for MLA? I'm wondering if we should instead fail in this case?

janEbert: We could ask the same about standard Attention, right? :)
One argument that objectively speaks for allowing this would be a hypothetical μP implementation for MLA, which would likely make use of `softmax_scale` similar to [how it's done for standard Attention](https://github.com/NVIDIA/Megatron-LM/blob/f5b2ec0130ff55f5f6d16d41f62aed1119868a3d/megatron/core/transformer/transformer_config.py#L1690-L1695).

janEbert: For comparing this to standard attention, you can basically ignore the `mscale` stuff, that's just relevant for YaRN.

MLA defines its softmax_scale as `(qk_head_dim + qk_pos_emb_head_dim)^-0.5`. The first term in the sum is the same as in standard attention; the second term is the per-head dimensionality of the decoupled queries, which are part of the modified, decoupled RoPE for MLA.

janEbert:
Fixed other places where it was missing and added non-functional unit tests.

ericharper (commented):
#

Issue #3177, opened 2026-01-30
Add muP-Muon scaling


BoxiangW:
**Is your feature request related to a problem? Please describe.**

Tag the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) 
to get oncall's attention to this issue.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

plugyawn:
I'll monitor this as #3058 progresses; I suppose we'll need #3175 first.

BoxiangW:
I think #3175 is not related to this because it is only for performance. This is related to #3058 though. **But**, as I checked here https://arxiv.org/pdf/2308.01814, Section 2.1, it did have a _entrywise_ assumption. It would be great if you want to go through the math discovery but be aware it is not being covered by TP series paper.

plugyawn:
Seems like it has since been worked out? 
https://arxiv.org/pdf/2505.02222
https://www.essential.ai/research/optimizer

plugyawn:
Looks doable. I could look into implementing their telescoping algorithm post #3058 (although I'm not sure it's anything relevant). They mention in Sec. 2.4 that the same muP scaling that is used for AdamW works for Muon (without any justification for the non-entrywise problem). 

Edit: https://arxiv.org/abs/2312.12226v2
This looks relevant; it's on Shampoo, but Muon should be similar.

plugyawn:
@BoxiangW now that #3058 is awaiting final review, perhaps could get started here?

BoxiangW:
@plugyawn Please feel free to move forward on this issue, let me actually add you into the assignees as well, thanks!

plugyawn:
@BoxiangW Sounds good! Thanks! I'll try to take a look at #3175 too in the process of reading up on Megatron's Muon implementation in more detail. I suppose Dion is also planned along the same lines?

plugyawn:
#3058 is merged! I'll get started working on this.

plugyawn:
@BoxiangW, added a PR for this. Took a while to be confident that the math is proper, would be great if you could take a look too. Seems like a simple change, and the transfer works as we expect.

plugyawn:
I think we can close this!

BoxiangW:
Thanks @plugyawn, closing since #3715 merged

PR #3715, opened 2026-03-05
Scaling for MuP over Muon optimizer.

plugyawn:
# What does this PR do ?
Addresses #3177, make default Muon scaling when conducting MuP transfer behave as it should. 
By default, Megatron scales Muon's matrix sign directions by $\sqrt{\max(\text{fan}\_{in}, \text{fan}\_{out})}$ (`spectral`). An alternate option is to use `unit_rms_norm`, or $\sqrt{\frac{\text{fan}\_{out}}{\text{fan}\_{in}}}$.

- Earlier, as a bug, turning MuP on would also apply Adam-wise LR/eps updates on the Muon-managed matrices, making MuP unprincipled on Muon (the matrices would be scaled both by Muon and by MuP). This has been fixed.

- In addition, even without double scaling, Muon-managed matrices scaled with the default `spectral`, loses width-invariance due to the `max()`. For square-matrices, for example, it grows as $\sqrt{n}$

The principled way to go forward seems to be to set Muon scaling to `unit_rms_norm`, i.e, $\sqrt{\frac{\text{fan}\_{out}}{\text{fan}\_{in}}}$, and so, we default to Muon with `unit_rms_norm` when MuP is active.

L to R: SP, MuP+`spectral`, MuP+`unit_rms_norm` (new default), old MuP double-scaling + `spectral` (old default)
image

L: SP, R: MuP+`spectral` vs MuP + `unit_rms_norm`. As expected, `unit_rms_norm` shows MuP transfer better than `spectral` on these sweeps.
image

Experiments on a slice of `wikitext-103`, 12 layer transformer, 100 iterations, if not stated otherwise. Confidence bounds from 5 runs with random seeds.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [x] I have added relevant documentation (bugfix, assuming further documentation not necessary?)

### Code review

<details>

#### (Step 3): Final Review

1. Add `Final Review` label

</details>

<details>

</details>

### Merging your PR


megatron/training/arguments.py
```diff
@@ -2148,9 +2159,11 @@ def _add_regularization_args(parser):
                        help='Whether to split QKV parameters for Muon optimizer')
     group.add_argument('--muon-use-nesterov', action='store_true',
                        help='Whether to use Nesterov-style momentum in the internal SGD')
-    group.add_argument('--muon-scale-mode', type=str, default='spectral',
+    group.add_argument('--muon-scale-mode', type=str, default=None,
```
BoxiangW: Can we leave this default unchanged and then add an assertion for args.muon_scale_mode == 'unit_rms_norm' when args.use_mup?

BoxiangW: cc @skyw

BoxiangW:
Thanks for your PR! The loss v. lr graph looks great!

megatron/training/arguments.py
skyw: I don't understand why it needs to change anything other than error out or maybe warns when incompatible scaling mode is set?

plugyawn: Thanks for review! Changed and set it to warn instead. Didn't throw an error since there should technically a way to also support spectral scaling by adding an explicit width rescale to the muon-managed matrices... in case that is to be used.

BoxiangW: Okay, can you also add a comment in the argument.py on how to turn on spectral scaling + mup? Other than that, looks good to me

BoxiangW (approved):
LGTM

megatron/training/arguments.py
plugyawn: Done!

plugyawn:
@BoxiangW thank you for the review! Added the arguments.py help string. Should be good for merge now, I think.

plugyawn:
@skyw, I've updated the warn behavior on `spectral` vs `unit_rms_norm`. It now explicitly throws a warning whenever MuP is used without `unit_rms_norm`, but doesn't change the defaults.

ericharper (approved):

megatron/training/arguments.py
```diff
@@ -1430,6 +1430,9 @@ def validate_args(args, defaults={}):
         assert not args.use_megatron_fsdp, "Muon optimizer does not support Megatron-FSDP for now."
         assert args.ckpt_format in ["torch", "torch_dist"], "Muon optimizer supports torch and torch_dist checkpoint format."
 
+        if args.muon_scale_mode is None:
```
skyw: this can be removed as well.

skyw (approved):
some left over needs to be cleaned. otherwise LGTM

megatron/training/arguments.py
plugyawn: My bad. Done!

plugyawn:
@skyw thanks! Made the changes.

megatron/core/optimizer/__init__.py
```diff
@@ -152,6 +155,16 @@ def get_mup_config_overrides(
             message += " MuP Adam epsilon scaling remains applied to hidden matrix-like parameters."
         log_single_rank(logger, logging.WARNING, message)
 
+    if is_muon_optimizer:
+        muon_scale_mode = getattr(config, 'muon_scale_mode', 'spectral')
+        if muon_scale_mode == 'spectral':
+            log_single_rank(
+                logger,
+                logging.WARNING,
+                "Muon + MuP is configured with muon_scale_mode=spectral. "
```
jaredcasper: I'm confused what this warning means. Is this an invalid configuration? What happens when muon_scale_mode=spectral in this case? It feels like this should either be an explicit error or nothing.

plugyawn: This should be more verbose? The notion is that MuP with `spectral` is not principled (but it's also the default behavior on Muon), as it doesn't maintain the required width invariance when scaling the 2D matrices. The right way would be `unit_rms_norm`. However, one could want to set a custom scale on the matrices to get invariance back, while retaining the spectral scaling... so I thought we'd want to not directly error it out.

plugyawn: On second thoughts, it should be more verbose.

plugyawn: Changing it to say "Both MuP and muon_scale_mode=spectral are enabled. Muon-managed matrix parameters will continue using spectral Muon scaling. Set --muon-scale-mode unit_rms_norm if you want
  unit_rms_norm scaling for Muon-managed matrices with MuP."

plugyawn: @jaredcasper does that sound fine? Or should we explicitly error it out?

plugyawn:
Ran a few more exps to see how compensatory scaling on the Muon (2D) matrices on `spectral` behaves.
image

 
Left is MuP while letting Muon handle the 2D scaling in `spectral`. As we see, it still does fine.

Right is MuP + Muon with 2D scaling in `spectral`, with an extra multiplicative compensation term of `1/sqrt(width_mult)` on the 2D matrices, to cancel the width growth of spectral scaling and restore width-invariant update magnitude across widths.

The best achieved loss seems to be lower, and the clustering of the minima appears to be tighter, but that might be an artifact of the steps to train. 

I've kept it out of the PR for now since I've not found it used in published literature (so far) and it seems to make sense to have a standing warning (considering it _does_ still work rather fine).

plugyawn:
@BoxiangW @jaredcasper could you help with what the direction should be?

BoxiangW:
We need Jared's approval to merge this PR, I'll ping him again, thanks

megatron/core/optimizer/__init__.py
jaredcasper: I see, makes more sense now, thanks!

jaredcasper (approved):

plugyawn:
@BoxiangW sorry, I think the force-pushing removed the Approved tag. Could you please push it along again?

asolergi-nv (approved):

asolergi-nv:
/ok to test [c8c9078](https://github.com/NVIDIA/Megatron-LM/pull/3715/commits/c8c9078937c12f25ca285a3053157fb33cbd7e18)

BoxiangW:
@plugyawn Can you take a look at the CI fail?

plugyawn:
@BoxiangW sorry, there was a misbehaving test, fixed and updated.

plugyawn:
@BoxiangW not sure why the CI is still failing... I actually ran out of compute 😭, but I'll try to get some more and update this. 

Update: I asked Codex, and it suggested that the newer GB200 workflow is getting skipped, since this is a non-maintainer PR, but I'm unsure if that's the case. 

---
Edit: the exact diagnosis is:
```
  - The internal CI branch aac1e1b14 merged a newer main.
  - That newer workflow added GB200 integration jobs:
      - cicd-parse-integration-tests-gb200
      - cicd-integration-tests-latest-gb200
  - Those jobs are maintainer-only in .github/workflows/cicd-main.yml:
      - needs.is-not-external-contributor.outputs.is_maintainer == 'true'
  - For this PR they get skipped.
  - Nemo_CICD_Test then counts skipped jobs globally and treats them as fatal because SKIPPING_IS_ALLOWED=false.
```

plugyawn:
The MLA-Yarn unit test fail looks unrelated to the MuP changes. Perhaps a transient error?

plugyawn:
Thanks everyone!

PR #3730, opened 2026-03-06
[FSDP] don't sync param when use no_shard

wplf:
# What does this PR do ?
Don't sync param when use no_shard.

## bug root cause
When use no_shard and overlap_param_gather is False, we will step into https://github.com/NVIDIA/Megatron-LM/blob/41daf81772fd6120174d34e9f62c0c2e9e017678/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L1189-L1203.
And the function here use mbuf.

https://github.com/NVIDIA/Megatron-LM/blob/41daf81772fd6120174d34e9f62c0c2e9e017678/megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py#L4032

**But mbuf is None when use no_shard, the bug shows up.**

## fix
When using no_shard strategy, grad will be allreduce and not reduce_scatter, so the whole optim step and whole weight update will be done. And param sync is not needed anymore.
Similiar logic is here https://github.com/NVIDIA/Megatron-LM/blob/41daf81772fd6120174d34e9f62c0c2e9e017678/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L434

 

<details>

</details>


wplf:
Hi, @shjwudp  @cspades
Could you please look at this PR?

PR #3741, opened 2026-03-06
RL support for nanov3 sft checkpoint

jon-barker:
# What does this PR do ?
Small changes to support correct loading and tokenization when using the nanov3 SFT checkpoint.

For reviewers of `arguments.py`: it is necessary to weaken the assertion that prevents setting both `--hybrid-layer-pattern` and `--num-layers`. Some checkpoints include both of these in their arguments, e.g. nanov3 sft checkpoint, but we still want to be able to use `--use-checkpoint-args`. As such, we make it a warning that in these cases the hybrid layer pattern will always override the num-layers.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/.github/workflows/claude-complexity-label.yml
+++ b/.github/workflows/claude-complexity-label.yml
@@ -27,6 +27,7 @@ jobs:
         uses: anthropics/claude-code-action@v1
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          github_token: ${{ secrets.PAT }}
           prompt: |
             REPO: ${{ env.REPO }}
             PR NUMBER: ${{ env.PR_NUMBER }}
--- a/.github/workflows/claude_review.yml
+++ b/.github/workflows/claude_review.yml
@@ -3,8 +3,6 @@ name: Claude Code Review
 on:
   issue_comment:
     types: [created]
-  pull_request_target:
-    types: [ready_for_review]
 
 jobs:
   review-on-comment:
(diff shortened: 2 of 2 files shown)


examples/rl/model_configs/nemotron6_3b_moe.sh
```diff
@@ -55,9 +55,11 @@ ENV_DEPENDENT="\
   --langrl-env-config $ENV_CONFIG "
 
 MODEL_OPTIONS="\
+  --no-use-tokenizer-model-from-checkpoint-args \
   --rl-skip-bos-token \
   --no-rl-use-sequence-packing \
   --rl-partial-rollouts \
+  --rl-offload-optimizer-during-inference \
```
tdene: This was intentionally removed by a previous PR, with the reason given being

> This flag is still supported but is not recommended to have on by default because of the large speed downside.

Are we sure we want to re-add it as a default flag?

jon-barker: Good catch - i'll remove it as a default. I do know that for some workloads it is necessary to prevent UVM thrash on the inference side, but it's probably still better to make it an opt-in thing.

jon-barker:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3741/commits/ba73d11fc531d169765bb2b46e753b611a69b82e

megatron/training/arguments.py
```diff
@@ -2304,6 +2299,9 @@ def _add_rl_args(parser):
                         help='Number of parallel generation tasks for RL inference.')
     group.add_argument('--rl-skip-bos-token', action=argparse.BooleanOptionalAction, type=bool, default=False,
                         help='Skip BOS token at the beginning of the sequences. Default is False.')
+    group.add_argument('--rl-inference-parsers', nargs='*', default=[],
```
yobibyte, line 2302: This one is a bit weird. We add the argument, but do not change any usage of it in the PR.

examples/rl/model_configs/nemotron6_3b_moe.sh
```diff
@@ -55,6 +55,7 @@ ENV_DEPENDENT="\
   --langrl-env-config $ENV_CONFIG "
 
 MODEL_OPTIONS="\
+  --no-use-tokenizer-model-from-checkpoint-args \
```
tdene, line 58: Does this break the current, non-SFT non-HF, nanov3 model flow?

jon-barker: No, because we specify tokenizer-model in run_job. What does break that workflow though is that I’m changing the tokenizer-type from TikTokenizer to HuggingFaceTokenizer. I think we should do this though since it makes sense to have an SFT’d model as the default since we’ve largely removed the ability to do chat prompting for base models with the move to OpenAI chat API

megatron/training/arguments.py
jon-barker: This came from @ArEsKay3 - I believe it’s required to use functionality introduced in other PRs for supporting tool calling etc in inference rollouts

jon-barker: I should have said: right now the usage of this arg will be in `mrl_internal` as part of `run_job` or the testing infra. We could add it to the nemotron6_3b_moe config though if we want to make it default

yobibyte: Got it! Sounds good.

PR #3763, opened 2026-03-10
Fix incorrect HAVE_TE detection in multiple modules

### LINKED ISSUE #3764: Incorrect HAVE_TE detection when transformer_engine is not installed
**Describe the bug**

Several modules may incorrectly detect TransformerEngine as available when the `transformer_engine` package is not installed.

The issue is caused by importing from `megatron.core.extensions.transformer_engine` before explicitly checking whether `transformer_engine` is installed.

Since `megatron.core.extensions.transformer_engine` falls back to `MagicMock`, the import can still succeed even when TransformerEngine is unavailable. As a result, `HAVE_TE` may be incorrectly set to `True`, which can enable TE-dependent logic in a no-TE environment.

Affected files include:

- `megatron/core/transformer/multi_latent_attention.py`
- `megatron/core/transformer/moe/shared_experts.py`
- `examples/multimodal/layer_specs.py`
- `examples/multimodal/radio/radio_g.py`

1. Prepare an environment where `transformer_engine` is not installed.

2. Run one of the following examples:

   ```python
   from megatron.core.transformer.multi_latent_attention import HAVE_TE
   print(HAVE_TE)
   ````

3. Observe that `HAVE_TE` may evaluate to `True` even though the `transformer_engine` package is not installed.

**Expected behavior**

`HAVE_TE` should be `False` when the `transformer_engine` package is not installed.

TE-dependent logic should only be enabled after Megatron-LM confirms that TransformerEngine is actually available.

**Additional context**

I have already opened a PR for this issue: 
[https://github.com/NVIDIA/Megatron-LM/pull/3763](https://github.com/NVIDIA/Megatron-LM/pull/3763)

returnL:
# What does this PR do?

Fixes #3764

Fix incorrect TransformerEngine availability detection in several modules and tests by ensuring TE-dependent logic is gated by the shared `HAVE_TE` flag from `megatron.core.extensions.transformer_engine`.

## Problem

Some files inferred TransformerEngine availability from whether imports from
`megatron.core.extensions.transformer_engine` succeeded.

Because that module falls back to `MagicMock` when `transformer_engine` is not
installed, those imports could still succeed even in environments without
TransformerEngine, which could incorrectly set `HAVE_TE = True` locally.

This could enable TE-dependent logic in environments where TransformerEngine is unavailable.

## Fix

Stop inferring TE availability from successful imports of TE-backed helpers.

Instead, use `megatron.core.extensions.transformer_engine.HAVE_TE` as the
single source of truth, and keep TE-only imports and behavior behind that gate.

Also update related tests to align with the corrected TE availability behavior.

## Scope of changes

This PR applies the same fix pattern across multiple TE-dependent code paths.

The changes fall into three categories:
- Core Megatron modules that conditionally enable TE-backed code paths
- Example multimodal/model-spec files that expose TE-backed layer definitions
- Unit tests that previously inferred TE availability from successful extension imports

The change pattern is intentionally narrow:
- use the shared `HAVE_TE` signal from `megatron.core.extensions.transformer_engine`
- keep TE-only imports behind that gate
- avoid enabling TE-dependent logic when `transformer_engine` is unavailable

## Validation

- `HAVE_TE = False` when `transformer_engine` is not installed
- `HAVE_TE = True` when `transformer_engine` is installed
- TE-dependent imports are no longer used to infer TE availability
- Related tests are updated to match the corrected behavior

### Pre-checks

### Code review

3. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/.github/workflows/build-test-publish-wheel.yml
+++ b/.github/workflows/build-test-publish-wheel.yml
@@ -66,14 +66,14 @@ jobs:
           GITHUB_RUN_ID: ${{ github.run_id }}
           SKIPPING_IS_ALLOWED: false
         run: |
-          FAILED_JOBS=$(gh run view $GITHUB_RUN_ID --json jobs --jq '[.jobs[] | select(.status == "completed" and .conclusion != "success")] | length') || echo 0
+          FAILED_JOBS=$(gh run view $GITHUB_RUN_ID --json jobs --jq '[.jobs[] | select(.status == "completed" and .conclusion != "success" and (.name | test("build-and-test-wheels")))] | length') || echo 0
 
           if [ "${FAILED_JOBS:-0}" -eq 0 ] || [ "$SKIPPING_IS_ALLOWED" == "true" ]; then
-              echo "✅ All previous jobs completed successfully"
+              echo "✅ All build-and-test-wheels jobs completed successfully"
               exit 0
           else
-              echo "❌ Found $FAILED_JOBS failed job(s)"
+              echo "❌ Found $FAILED_JOBS failed build-and-test-wheels job(s)"
               # Show which jobs failed
-              gh run view $GITHUB_RUN_ID --json jobs --jq '.jobs[] | select(.status == "completed" and .conclusion != "success") | .name'
+              gh run view $GITHUB_RUN_ID --json jobs --jq '.jobs[] | select(.status == "completed" and .conclusion != "success" and (.name | test("build-and-test-wheels"))) | .name'
               exit 1
           fi
--- a/megatron/core/inference/contexts/attention_context/triton/tensor_ops.py
+++ b/megatron/core/inference/contexts/attention_context/triton/tensor_ops.py
@@ -3,8 +3,21 @@
 from typing import Optional
 
 import torch
-import triton  # type: ignore
-import triton.language as tl  # type: ignore
+
+try:
+    import triton
+    import triton.language as tl
+
+    HAVE_TRITON = True
+except ImportError:
+    from unittest.mock import MagicMock
+
+    from megatron.core.utils import null_decorator
+
+    triton = MagicMock()
+    triton.jit = null_decorator
+    tl = MagicMock()
+    HAVE_TRITON = False
 
 
 @triton.jit
--- a/megatron/core/inference/contexts/fused_kv_append_kernel.py
+++ b/megatron/core/inference/contexts/fused_kv_append_kernel.py
@@ -1,8 +1,21 @@
 # Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
 
-import triton
-import triton.language as tl
 from torch import Tensor
+
+try:
+    import triton
+    import triton.language as tl
+
+    HAVE_TRITON = True
+except ImportError:
+    from unittest.mock import MagicMock
+
+    from megatron.core.utils import null_decorator
+
+    triton = MagicMock()
+    triton.jit = null_decorator
+    tl = MagicMock()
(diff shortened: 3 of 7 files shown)


returnL:
Hi, I’ve updated this PR to include the additional affected files and corresponding test updates. Could someone please take a look when available? Thanks!

examples/multimodal/radio/radio_g.py
```diff
@@ -18,24 +18,25 @@
 from megatron.core.typed_torch import not_none
 
 try:
+    import transformer_engine  # pylint: disable=unused-import
```
yaox12: I'm wondering if we could directly import the `HAVE_TE` flag from `megatron.core.extensions.transformer_engine`, instead of checking it in every file separately. How do you think?

returnL: Thanks, I think that’s a good suggestion. Importing `HAVE_TE` directly from `megatron.core.extensions.transformer_engine` would make the logic cleaner and more consistent across files.

I’ll update the PR to follow this approach.

returnL: Implemented this. I switched the updated files to import `HAVE_TE` directly from `megatron.core.extensions.transformer_engine` instead of checking it locally in each file, and I also fixed the remaining files in this PR that followed the same pattern.

returnL:
Hi, I've fixed the CI failures (f6dbf87) and synced with the latest main branch. The PR is ready for re-review. Thanks!

returnL:
Hi maintainers,

The previous CI issues have been addressed and all checks are now passing after rerunning CI.
Could this PR move forward to final review?

Thanks!

megatron/core/post_training/modelopt/layers.py
```diff
@@ -16,18 +16,13 @@
 
 logger = logging.getLogger(__name__)
 
-try:
-    import transformer_engine as te
+if HAVE_TE or TYPE_CHECKING:
```
yashaswikarnati, line 19: does this break when TE is not available, do we need the try except on ImportError here?
CC: @nschank wdyt?

maanug-nv: I think the intent is to defer breakages to specific modules that use TE. If you are not importing `Norm` from this file, no need to break.

Breakage should be deferred to here: https://github.com/returnL/Megatron-LM/blob/6887d71c96280021d0826416f558227389a2485c/megatron/core/post_training/modelopt/layers.py#L61-L73

> do we need the try except on ImportError here?
If someone wants to use `Norm` and doesn't have TE installed, how would you want to handle the ImportError? Use a torch-only fallback impl?

returnL: Thanks for the careful review!

In the original code on `main`, line 9 already has an unguarded top-level import:
```python
from megatron.core.extensions.transformer_engine import _get_extra_te_kwargs
```
This never raises `ImportError` without TE, because `megatron.core.extensions.transformer_engine` falls back to `MagicMock`. This PR moves `_get_extra_te_kwargs` behind the `if HAVE_TE` guard, so the behavior is unchanged — if anything, slightly **more** guarded than before.

As @maanug-nv mentioned, the `if not HAVE_TE: raise ImportError` at L61-65 is pre-existing and not part of this change.

That said, adding a torch-only fallback for `Norm` when TE is unavailable sounds like a valid improvement — happy to help with that in a follow-up PR if needed!

yashaswikarnati (approved):

PR #3766, opened 2026-03-10
[Dev][feat] Support overlapping A2A Combine backprop with wgrad GEMM

Wohox:
# What does this PR do ?

PR for main: https://github.com/NVIDIA/Megatron-LM/pull/3795
### Problem

In MoE models, the expert weight gradient (wgrad) computation during backward is serialized on the main CUDA stream. This blocks the data gradient (dgrad) from flowing to earlier layers until the expert wgrad finishes, even though there is no data dependency between them. The result is wasted GPU cycles — earlier layers' backward pass sits idle waiting for expert wgrad to complete.

With FSDP, this is further compounded because the gradient reduce-scatter for expert parameters is also blocked on the same critical path.

### Solution

This PR introduces a new flag `--overlap-dispatch-backward-with-experts-wgrad` that separates the expert wgrad computation from the main backward stream:

1. **Two autograd functions** are inserted into the MoE layer's forward graph:
   - `_RecordExpertDgradCompletion` — placed before the expert computation; during backward, it records a CUDA event once the expert dgrad is done.
   - `_RegisterDelayedWgradForExperts` — placed at the dispatch boundary; during backward, it waits on the dgrad event, then launches `backward_dw()` on a **dedicated CUDA stream**, and synchronizes back to the main stream before proceeding.

2. **FSDP integration** — When used with MegatronFSDP, expert parameters are marked with `_fsdp_delay_grad_reduce = True` so the normal post-accumulate-grad hook skips them. A callback is registered via `register_process_expert_grads_fn()` that triggers the FSDP reduce-scatter for expert parameters only after the delayed wgrad computation completes.

3. **TE GroupedLinear** is configured with `delay_wgrad_compute=True`, which tells Transformer Engine to skip wgrad during the normal autograd backward and instead wait for an explicit `backward_dw()` call.

### How to enable

```bash
--overlap-dispatch-backward-with-experts-wgrad
```

**Requirements:**
- Transformer Engine >= 2.3.0
- `moe_grouped_gemm` enabled (not legacy grouped gemm)
- Mutually exclusive with `--delay-wgrad-compute` (the existing A2A-overlap-based delay)
- Mutually exclusive with `--overlap-moe-expert-parallel-comm`

Works with both **FSDP** and **3-D parallelism** (TP/EP/PP).

### What is achieved

The expert wgrad computation runs on a separate CUDA stream, overlapping with the EP communication within the same transformer layer. This reduces the wall-clock time of the backward pass without changing numerical results — the feature is **bit-exact** with the non-delayed baseline (verified by unit tests comparing per-step losses and final weights over multiple optimizer steps).

### Changes

| File | Description |
|------|-------------|
| `megatron/core/model_parallel_config.py` | New config flag `delay_wgrad_compute_for_te_grouped_gemm` |
| `megatron/core/transformer/transformer_config.py` | Validation assertions for the new flag |
| `megatron/core/transformer/moe/moe_layer.py` | Autograd functions for delayed wgrad + dedicated CUDA stream/event + `register_process_expert_grads_fn` callback |
| `megatron/core/extensions/transformer_engine.py` | Pass `delay_wgrad_compute=True` to TE GroupedLinear when the new flag is set |
| `megatron/core/distributed/fsdp/.../megatron_fsdp.py` | FSDP hook to defer reduce-scatter for expert params and trigger it after delayed wgrad |
| `tests/unit_tests/a2a_overlap/test_delay_wgrad_compute.py` | Unit tests covering basic, shared-expert, multi-layer, and FSDP scenarios |

### Test plan

- [x] Unit test: `test_delay_wgrad_compute_for_te_grouped_gemm` — full-model training loop (forward → backward → optimizer) comparing delayed vs. non-delayed across `num_layers × shared_experts × dispatcher_type × fp8_flag`
- [x] Unit test: `test_delay_wgrad_compute_for_te_grouped_gemm_with_fsdp` — same comparison with MegatronFSDP wrapping (`fully_shard_model` + `fully_shard_optimizer`), verifying the deferred reduce-scatter path

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


Wohox:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/48b1c2984ed8bcca85d45f8c977932005d1929d1

Wohox:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/adb70af91225337b8af6f0257ccdc922c37470ed

Wohox:
/ok to test [6d5d7be](https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/6d5d7bea11c120b651fa28c3bc58fdca44effa29)

Wohox:
/ok to test [a7d2b2b](https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/a7d2b2b7df9b082b85259c5768343a8f59bf22dd)

Wohox:
/ok to test [3466356](https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/34663563cc31115a5afafc64eb112255ba8a1c30)

Wohox:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/9d33e8f82fc485ddcd38eb54dabf1e8518af4316

Wohox:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/00bd62eb845a750f3e249ff49afa51c901d15272

Wohox:
/ok to test [ab7f8e4](https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/ab7f8e47c8d7e6301c72a772725041e78a9488ba)

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -1014,7 +1045,11 @@ def _register_pre_backward_param_unshard_hook(module):
             for param in grad_acc_param_list:
                 self.grad_acc_hooks[f"grad_acc and reduce for {self.param_to_name[param]}"] = (
                     param.register_post_accumulate_grad_hook(
-                        lambda p: _process_post_backward_gradients([p])
+                        lambda p: (
```
Victarry: **[CRITICAL Correctness]** When `skip_backward_post_hook=True`, this branch suppresses immediate `_process_post_backward_gradients([p])`. The delayed fallback is only installed in `setup_delayed_wgrad_acc_hook()` when `need_backward_dw()` is true. If that fallback is absent for any skipped parameter, gradient post-processing can be silently dropped.

**Suggestion:** Add a guard here: if `skip_backward_post_hook` is true but `post_wgrad_grad_acc_hook` is missing, either (1) call `_process_post_backward_gradients([p])` immediately as fallback, or (2) raise an explicit error to fail fast.

tests/unit_tests/a2a_overlap/test_delay_wgrad_compute.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+class TestDelayWgradCompute:
+    """Verify that delay_wgrad_compute_for_te_grouped_gemm produces identical
+    training behaviour (per-step loss and final weights) as the non-delayed baseline
+    across multiple forward-backward-optimizer steps on the full GPTModel.
+    """
+
+    def setup_method(self, method):
+        Utils.initialize_model_parallel(
```
Victarry: **[IMPORTANT Test Coverage]** This FSDP test validates deferred expert-gradient processing, but it does not assert a data-parallel world size greater than 1. On DP=1, the cross-rank reduce-scatter path can be bypassed while the test still passes.

**Suggestion:** Add a precondition/assertion that DP world size is `> 1` for this case (or run a dedicated CI variant with fixed DP>1), so the deferred reduce-scatter path is guaranteed to be exercised.

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -250,6 +250,20 @@ def __init__(
...
+        # Setup events and streams for delayed wgrad computation.
+        self.setup_delayed_wgrad_for_dispatch_backward_overlap()
+
+    def setup_delayed_wgrad_for_dispatch_backward_overlap(self):
+        """Initializes CUDA events and streams for overlapping expert
+        weight gradient computation with dispatch backward.
+        """
+        self._delayed_wgrad_event: Optional[torch.cuda.Event] = None
+        self._delayed_wgrad_stream: Optional[torch.cuda.Stream] = None
+        self._process_expert_grads_fn = None
```
Victarry: **[SUGGESTION]** `_process_expert_grads_fn` is introduced here but currently not read or invoked in this PR path.

**Suggestion:** Remove the unused member to reduce API surface, or wire it into the actual deferred-grad callback flow and document the intended ownership relative to `post_wgrad_grad_acc_hook`.

megatron/core/model_parallel_config.py
```diff
@@ -278,6 +278,15 @@ class ModelParallelConfig:
...
     """Delay the weight gradient computation to improve batch-level communication overlapping"""
 
+    delay_wgrad_compute_for_te_grouped_gemm: bool = False
+    """Delay the weight gradient computation for TE Grouped GEMM MoE experts.
+    When enabled with FSDP, the expert weight gradients are computed on a separate
+    CUDA stream after the data gradients finish, allowing overlap of wgrad compute
+    with EP A2A communication. The FSDP gradient reduce-scatter for
+    expert parameters is deferred until the delayed wgrad computation completes.
+    This requires transformer_engine with GroupedLinear support (TE >= 2.3.0).
+    """
```
Victarry, line 288: This `delay_wgrad_compute_for_te_grouped_gemm` name seems not very representative. Could you help clarify the following questions?
1. Is there any specific reason we can't make `delay_wgrad_compute_for_te_grouped_gemm` as the default behavior when we are not using 1F1B overlap? Will this cause perf regression?
3. Could we also make wgrad(qkv proj, output proj) in attention module overlap with combine backward?

One of my proposal is to make this PR as one of the implementations in `overlap_moe_expert_parallel_comm`. We could add a new argument like `overlap_moe_expert_parallel_comm_pattern` with `inter-batch` and `intra-batch`. This PR corresponds to the second case.

Wohox: 1. This feature won't hurt performance, we can make it the default option.
2. Could we also make wgrad(qkv proj, output proj) in attention module overlap with combine backward?
    * Within the same layer: No, because attn dgrad computation happens after combine backward and wgrad can only be delayed after attn dgrad.
    * Cross layers: Yes, this requires cross-layer event handling - need `layers[i]` to access `layers[i+1]`'s event and record before combine backward begins. However, from the influence to code design, this may complicates the code for `transformer_layer` and `next_layer`/`prev_layer` is needed, otherwise hacks in the training loop is necessary.

Victarry: 1. BTW, will this feature increase the peak memory? For non-FSDP case, I guess the memory overhead is negligible. Not sure about that with FSDP. Do you have any data? 
2. I see. Maybe we need a more general design for transformer_layer for these manually controlled fine-grained overlap. Let's continue with this PR first.
Thanks for the great work!

Wohox: * No, if the FSDP unit is neither `TransformerLayer` nor `MLP`, but a even smaller module like `TEGroupedGemm`, then wgrad overlap require extra unsharding hook and resharding hook at the begining and end of the delayed backawrd_dw computation. Even under this case, the unsharded expert weights will be released before the backward of next layer. However, I think we are missing an assertion on the FSDP unit being `TransformerLayer` or containing `MLP`, and what do you think @shjwudp, is this assertion necessary, will the customers use super fine grained FSDP units?
* No, if the FSDP unit is `TransformerLayer` or contains `MLP`. The unsharded expert weights is available throughout the backward pass of `TransformerLayer`/`MLP` and no extra sharding&resharding is required. 

I will run with a DSV3 proxy to verify the effect on memory.

Wohox:
/ok to test [4981158](https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/4981158d645d49dc569d03b77582da7e14bc9a04)

tests/unit_tests/a2a_overlap/test_schedule_chunk_1f1b.py
```diff
@@ -103,7 +103,6 @@ def test_1f1b_schedule_model_chunk(self, mtp_layers, dispatcher_type, fp8_flag,
         extra_kwargs = {"moe_token_dispatcher_type": dispatcher_type}
         if dispatcher_type == "flex":
             extra_kwargs["moe_flex_dispatcher_backend"] = "deepep"
-            extra_kwargs["moe_router_dtype"] = "fp32"
```
lhb8125, line 106: Is it intended?

Wohox: Yes, because we now need the `moe_router_dtype = fp32` option for all token dispatchers. This change is intended to resolve the accuracy issue.

I can include this update in the current PR for submission, or I can create a separate new PR if preferred.

Wohox:
/ok to test [6e26e97](https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/6e26e97589a99a5c009968f81af79fafec7707bc)

Wohox:
/ok to test [38368a2](https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/38368a2676982b2859ac31ef3c2a138b5cd61612)

megatron/core/model_parallel_config.py
Wohox: @Victarry  This feature won't include extra memory usage, but may increase the CPU overhead and leads to worse perf, keeping it as a standalone flag is better.
Screenshot 2026-03-31 at 15 50 07
Screenshot 2026-03-31 at 15 51 14

Wohox:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/625d76b98050b58fd01f3c754d18c0f03f8cd7d9

Wohox:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/4e58bcbb41658846852ba684df2f653be8425429

Victarry:
Current dev branch is not healthy, trying to fix it with https://github.com/NVIDIA/Megatron-LM/pull/4123

Wohox:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3766/commits/7b978c42f812e0981c2b96016a5333e073d823f6

PR #2217, opened 2025-11-12
[DEV] Fix aux loss scale when cp enabled

Victarry:
# What does this PR do ?
Fix aux loss scale when CP enabled. PR to main branch https://github.com/NVIDIA/Megatron-LM/pull/2237

image

## Explanation of this Fix
Ideally, we want to make sure the accumulated aux loss to be scaled by the total number of batches.

Current, there are 2 kinds of scaling for aux loss(when calculate_per_token_loss is False):
1. Gradient scale in DDP with DP size. https://github.com/NVIDIA/Megatron-LM/blob/bb5a0fd0e376604fd426d2e5da79f7d7a009ea34/megatron/core/distributed/distributed_data_parallel.py#L308
2. scale by number of micro-batches(GA) in schdules.py https://github.com/NVIDIA/Megatron-LM/blob/bb5a0fd0e376604fd426d2e5da79f7d7a009ea34/megatron/core/pipeline_parallel/schedules.py#L271

With TP or CP enabled, each rank only calculate partial aux loss. TP will SUM reduce the router gradient with TP_group(https://github.com/NVIDIA/Megatron-LM/blob/bb5a0fd0e376604fd426d2e5da79f7d7a009ea34/megatron/core/distributed/finalize_model_grads.py#L346), and CP will SUM reduce the gradient in the DDP part with CP_DP_Group.
* When TP enabled, the DP size is halved but GA number is doubled, so the actual loss scale is consistent.
* **When CP enabled, DP size is the same but the GA number if doubled. So the actual aux loss scale will be halved, which make the aux loss becomes larger with CP=2** To fix this, the aux loss scale is multiplied by `cp_size` in this PR.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
        C1[Expert Review] --> C2[Final Review]
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [ ] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

[...]

Issue #1652, opened 2025-06-26
[BUG] aux_loss and z_loss is incorrect when use calculate_per_token_loss and cp


OutisKwak:
**Describe the bug**

1. If using calculate_per_token_loss and cp > 1, 
   firstly, aux_loss is divided by the square of full num_tokens (considered cp）
  https://github.com/NVIDIA/Megatron-LM/blob/a845aa7e12b3a117e24c2352b9e3e60bad2e3a17/megatron/core/transformer/moe/moe_utils.py#L60)
  
secondly， aux_loss is scaled  by num_local_tokens here.
  https://github.com/NVIDIA/Megatron-LM/blob/a845aa7e12b3a117e24c2352b9e3e60bad2e3a17/megatron/core/transformer/moe/router.py#L312
 
finally, scale both the main_loss gradient and aux_loss gradient by  1/(num_local_tokens * dp_size *  num_micro_batches) in finalize_model_grads function. 
 however, the num_local_tokens is not local but full.
  https://github.com/NVIDIA/Megatron-LM/blob/a845aa7e12b3a117e24c2352b9e3e60bad2e3a17/pretrain_gpt.py#L179
 
so we should scale aux_loss by full num_tokens (considered cp and sp）not num_local_tokens

2. If not use calculate_per_token_loss but use cp, gradient is divided  by dp*cp in finalize_model_grads function. lm_loss is scaled by cp in advance, but aux_loss is not scaled by cp, so should we multiply aux_loss by cp?

yanring:
Adding @shifangx

Victarry:
One of our customers reported the same issue. Fixed in https://github.com/NVIDIA/Megatron-LM/pull/3105

PR #3768, opened 2026-03-10
[Dev] fix(moe): align MoE router loss scale semantics

### LINKED ISSUE #3672: Global auxiliary loss gradient incorrectly scaled due to averaging over global token count
### Describe the bug

In the MoE (Mixture of Experts) auxiliary loss functions—`_apply_aux_loss`, `_apply_seq_aux_loss`, and `_apply_global_aux_loss`—the function [`switch_load_balancing_loss_func`](https://github.com/NVIDIA/Megatron-LM/blob/bbbedbb9f53343762e4dc70abc771b813a83d817/megatron/core/transformer/moe/moe_utils.py#L44) is called. This loss function computes a mean loss by dividing by an input argument `total_num_tokens`.

When `calculate_per_token_loss` is disabled (the common setting), the mean loss for the first two auxiliary losses (`_apply_aux_loss` and `_apply_seq_aux_loss`) is correctly computed based on the number of tokens in the current training step (micro-batch). However, **`_apply_global_aux_loss` averages its loss over the *global* total number of tokens** (i.e., across the entire data parallel group).

Because the global total number of tokens is larger by a factor of `self.tp_dp_cp_group.size()` (the world size), the gradient back-propagated from the global aux loss is proportionally smaller. Concretely, the gradient is scaled down by this factor compared to what the `aux_loss_coeff` would imply, making the coefficient's effect on the global loss inconsistent with its effect on the other two aux losses.

---

### Steps/Code to reproduce bug

1.  Enable MoE with auxiliary losses in a training setup that uses tensor/model parallelism (e.g., `--num-experts 8 --moe-aux-loss-coeff 0.01`).
2.  Ensure `--calculate-per-token-loss` is **not** set (the default behavior).
3.  Run a few iterations and compare the gradient magnitudes contributed by the global aux loss versus the per-token or sequence-level aux losses.
4.  Observe that the global aux loss gradients are significantly smaller (by a factor approximately equal to the DP group size) for the same coefficient value.

A minimal code inspection can also confirm the issue:
- In `megatron/core/transformer/moe/moe_utils.py`, examine `switch_load_balancing_loss_func`. Note it returns `loss / total_num_tokens`.
- Trace the calls to this function in the three auxiliary loss methods mentioned above. You will see that `_apply_global_aux_loss` passes a `total_num_tokens` that is aggregated across the entire parallel group, while the other two use the local micro-batch token count.

---

### Expected behavior

The global auxiliary loss should be averaged over the same token count as the other auxiliary losses when `calculate_per_token_loss` is disabled. That is, it should be averaged over the number of tokens in the **current training step** (micro-batch), not the global aggregated token count. This would ensure that the `aux_loss_coeff` has a consistent scaling effect across all auxiliary loss components.

Alternatively, if the global averaging is intentional, its behavior should be clearly documented and the coefficient scaling should be adjusted accordingly. However, the current implementation appears to be an unintended inconsistency.

---

### Additional context

- The root cause is that `total_num_tokens` in `_apply_global_aux_loss` is the sum of tokens across the entire parallel group, while in the other loss functions it is the local token count.
- This leads to a practical effect where the global aux loss contributes much less to the overall gradient than expected for a given coefficient. For example, with a world size of 8, the effective coefficient for the global loss becomes `aux_loss_coeff / 8`.
- Fixing this would involve either:
    1.  Using the local token count for the global loss as well, or
    2.  Explicitly scaling the loss or coefficient to maintain the intended influence.
- This issue is particularly relevant when tuning MoE auxiliary loss coefficients, as the same coefficient value will have different impacts depending on which aux loss function it applies to.

Tagging @mcore-oncall for visibility.

Victarry:
## Summary

This PR aligns the effective scale of MoE router auxiliary losses across aux-loss variants, `--calculate-per-token-loss`, and TP/CP/DP parallel strategies.

The conservative semantic baseline is kept unchanged: `load_balancing_type=aux_loss` with `--calculate-per-token-loss=false`. Other paths are aligned to that baseline.

Fixes #3672.

## Problem Manifestation

Before this fix, equivalent MoE training configurations could apply different effective router regularization strengths even when using the same `moe_aux_loss_coeff`.

The most visible symptoms were:

- `global_aux_loss` did not have the same effective scale as the standard `aux_loss`. In practice, the active load-balancing loss and router gradient strength could drift when switching only the aux-loss type from `aux_loss` to `global_aux_loss`.
- With `aux_loss` and `--calculate-per-token-loss` enabled, changing TP/CP sizes changed the effective aux-loss scale. The same model and data could therefore show different router-loss/router-gradient behavior under different TP/CP parallel strategies.
- The same per-token normalization issue also affected `seq_aux_loss`/`z_loss` style router-loss attachment paths, so aux type, per-token mode, and TP/CP/DP strategy were not cleanly comparable.

The expected behavior is that these configurations should preserve the same coefficient semantics: changing aux-loss type or parallel decomposition should not systematically rescale router gradients, except where the loss definition itself intentionally differs.

## Root Cause

There were two scale mismatches in the router-loss attachment path:

- `global_aux_loss` builds `tokens_per_expert` from a DP-aggregated `tp_dp_cp_group`, while router probabilities remain local. Since the final gradient is averaged across DP ranks, the attached gradient was diluted by `dp_size`.
- Per-token aux/z-loss paths attached loss scaled by the local valid-token count, but `finalize_model_grads` normalizes by the global token count. With TP/CP enabled, this missed the `tp_cp_group.size()` factor.

As a result, otherwise equivalent configurations could show systematic router-loss and router-gradient scale drift when switching aux-loss type, per-token-loss mode, or TP/CP/DP strategy.

## Fix

- Scale `global_aux_loss` by `dp_size = tp_dp_cp_group.size() // tp_cp_group.size()` so its backward strength matches `aux_loss`.
- Keep the logged global aux loss normalized through the same scale factor, so reported metrics stay comparable.
- Scale per-token aux-loss and z-loss attachment by `local_valid_tokens * tp_cp_group.size()` to match global-token-count gradient normalization.
- Preserve the non-per-token `aux_loss` baseline behavior.

## Validation

Public W&B report: https://api.wandb.ai/links/megatron-core-moe-dev/xafmkree

Original W&B report: https://wandb.ai/megatron-core-moe-dev/moe-aux-loss-scale-alignment/reports/MoE-Aux-Loss-Scale-Alignment-E2E-Ablation---2026-05-25--VmlldzoxNzAwODkxMw==

E2E validation uses the same model/config family and compares loss ratios against the baseline instead of forcing load balancing. Horizontal aux-type comparisons use a deterministic `mock_repeat` dataset so each global batch is identical across aux types. Vertical parallel-strategy comparisons use the real SlimPajama data path and load all compared TP/CP/DP jobs from the same one-iteration checkpoint.

Key validation results:

| Comparison | Before fix active LB ratio | After fix active LB ratio | After fix LM ratio |
| --- | ---: | ---: | ---: |
| `aux_loss` vertical | 0.9994 - 1.0455 | 0.9952 - 1.0063 | 0.9985 - 1.0053 |
| `seq_aux_loss` vertical | 0.9929 - 1.0419 | 0.9948 - 1.0053 | 0.9997 - 1.0057 |
| `global_aux_loss` vertical | 0.9483 - 1.0080 | 0.9981 - 1.0086 | 0.9986 - 1.0035 |
| `z_loss` vertical | 0.9844 - 2.6181 | 0.9254 - 1.0281 | 0.9968 - 1.0067 |

Additional checks:

- Horizontal deterministic repeat-data jobs: 12/12 completed.
- `aux_loss + --calculate-per-token-loss=false` is bitwise identical before/after fix, confirming the baseline is unchanged.
- `aux_loss` and `seq_aux_loss` fixed TP1/CP1 horizontal runs are bitwise identical before/after fix where the scale correction should be neutral.
- `global_aux_loss` changes after the first step as expected because the fix intentionally changes its backward scale.
- Added unit coverage for `global_aux_loss` gradient scaling.

Local sanity checks:

```bash
git diff --check
python3 -m py_compile megatron/core/transformer/moe/router.py megatron/core/transformer/moe/moe_utils.py tests/unit_tests/transformer/moe/test_aux_loss.py
```

### DIFF
--- a/megatron/core/transformer/moe/router.py
+++ b/megatron/core/transformer/moe/router.py
@@ -443,9 +443,14 @@ class TopKRouter(Router):
             fused=self.config.moe_router_fusion,
         )
 
+        # `global_aux_loss` uses DP-aggregated tokens_per_expert, while router probs remain
+        # local to each rank. Scale only by DP so the coefficient matches aux_loss semantics.
+        dp_size = self.tp_dp_cp_group.size() // self.tp_cp_group.size()
+        global_aux_loss = global_aux_loss * dp_size
+
         probs = self.attach_and_log_load_balancing_loss(
             probs,
-            global_aux_loss_coeff,
+            global_aux_loss_coeff * dp_size,
             global_aux_loss,
             "global_load_balancing_loss",
             self.tp_dp_cp_group,
@@ -457,7 +462,7 @@ class TopKRouter(Router):
     def attach_and_log_load_balancing_loss(
         self,
         activation: torch.Tensor,
-        aux_loss_coeff: float,
+        normalize_scale: float,
         aux_loss: torch.Tensor,
         aux_loss_name: str,
         reduce_group: torch.distributed.ProcessGroup,
@@ -468,7 +473,8 @@ class TopKRouter(Router):
 
         Args:
             activation (torch.Tensor): Activation tensor to attach the aux loss to.
-            aux_loss_coeff (float): Coefficient for the aux loss.
+            normalize_scale (float): Scale factor to normalize aux_loss for logging.
+                The logged value is `aux_loss / normalize_scale`.
             aux_loss (torch.Tensor): Computed aux loss.
             aux_loss_name (str): Name of the aux loss for logging.
             reduce_group (torch.distributed.ProcessGroup): Process group for reduction.
@@ -502,22 +508,17 @@ class TopKRouter(Router):
 
         get_moe_metrics_tracker().record(
             aux_loss_name,
-            aux_loss / aux_loss_coeff,
+            aux_loss / normalize_scale,
             layer_number,
             num_layers,
             reduce_group=reduce_group,
             needs_dp_avg=needs_dp_avg,
         )
         if self.calculate_per_token_loss:
-            # Scale the aux_loss by the number of tokens.
-            # The expected final scaling for aux_loss gradients is 1/(num_micro_batches * dp_size).
-            # After commit 02648000, Megatron started using the number of total tokens to scale
-            # gradients under the argument of calculate_per_token_loss,
-            # which scales both the main_loss gradient and aux_loss gradient by
-            # 1/(num_local_tokens * dp_size * num_micro_batches) in finalize_model_grads function.
-            # To correct this scaling, we need to scale the aux_loss by num_local_tokens here.
-            # Use valid_token_count (excluding padding) if provided, otherwise use total tokens.
+            # Match the non-per-token aux_loss baseline by converting the local mean-style
+            # aux loss to the TP/CP-wide token sum before final per-token normalization.
             num_tokens = valid_token_count if valid_token_count is not None else activation.shape[0]
+            num_tokens = num_tokens * self.tp_cp_group.size()
             activation = MoEAuxLossAutoScaler.apply(activation, aux_loss * num_tokens)
         else:
             activation = MoEAuxLossAutoScaler.apply(activation, aux_loss)
@@ -541,15 +542,9 @@ class TopKRouter(Router):
             moe_z_loss_coeff = self.config.moe_z_loss_coeff / self.tp_cp_group.size()
             z_loss = z_loss_func(logits, moe_z_loss_coeff, padding_mask=padding_mask)
             if self.calculate_per_token_loss:
-                # The expected final scaling for z_loss gradients is
-                # 1/(num_micro_batches * dp_size).
-                # After commit 02648000, Megatron started using the number of total tokens
-                # to scale gradients under the argument of calculate_per_token_loss,
-                # which scales both the main_loss gradient and z_loss gradient by
-                # 1/(num_local_tokens * dp_size * num_micro_batches) in finalize_model_grads().
-                # To correct this scaling, we need to scale the z_loss by num_local_tokens here.
-                # Count valid tokens: sum of inverted mask (False -> True = valid)
+                # Keep z_loss on the same scale as the non-per-token baseline.
                 num_tokens = (~padding_mask).sum() if padding_mask is not None else logits.shape[0]
+                num_tokens = num_tokens * self.tp_cp_group.size()
                 logits = MoEAuxLossAutoScaler.apply(logits, z_loss * num_tokens)
             else:
                 logits = MoEAuxLossAutoScaler.apply(logits, z_loss)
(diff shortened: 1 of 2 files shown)


yaox12 (approved):

zyeric:
why use `dp_size` instead of `self.tp_dp_cp_group.size()`. Since `global_tokens_per_expert` is reduced at `tp_dp_cp_group`.

Victarry:
> why use `dp_size` instead of `self.tp_dp_cp_group.size()`. Since `global_tokens_per_expert` is reduced at `tp_dp_cp_group`.

It's to keep consistent with the standard aux loss, otherwise the effective scale of global aux loss will be increased.

zyeric:
> > why use `dp_size` instead of `self.tp_dp_cp_group.size()`. Since `global_tokens_per_expert` is reduced at `tp_dp_cp_group`.
> 
> It's to keep consistent with the standard aux loss, otherwise the effective scale of global aux loss will be increased.

I think all aux loss should align with `--calculate-per-token-loss`'s behavior: `activation = MoEAuxLossAutoScaler.apply(activation, aux_loss * activation.shape[0])` (mean loss -> sum loss). If `dp_size` is used when tp * cp > 1, I think the semantic is confusing.

Victarry:
> > > why use `dp_size` instead of `self.tp_dp_cp_group.size()`. Since `global_tokens_per_expert` is reduced at `tp_dp_cp_group`.
> > 
> > 
> > It's to keep consistent with the standard aux loss, otherwise the effective scale of global aux loss will be increased.
> 
> I think all aux loss should align with `--calculate-per-token-loss`'s behavior: `activation = MoEAuxLossAutoScaler.apply(activation, aux_loss * activation.shape[0])` (mean loss -> sum loss). If `dp_size` is used when tp * cp > 1, I think the semantic is confusing.

@zyeric  Thanks for your comments.
I agreed aux loss with/without `--calculate-per-token-loss` should have the same bahavior.
I think the aux loss without calculate per token loss is the most commonly used loss, so we don't want change its semantics for now, which will influence the optimal hyper parameters. So I prefer to keep the current formula. There are some historical issues for the current "confusing" implementation, like https://github.com/NVIDIA/Megatron-LM/pull/2217/changes#r2521887449

For the `--calculate-per-token-loss`, there are still some issues which may cause loss diff with different parallel sizes, see https://github.com/NVIDIA/Megatron-LM/pull/3105 and https://github.com/NVIDIA/Megatron-LM/issues/1652 for details. I'll revisit the implementations to make sure all of different combinations(parallel size, per-token-loss, different aux loss types) follow the same semantics.



Marked this PR as draft until resolving all aux-loss related issues.

Victarry:
Closed this in favor of #5047

PR #3771, opened 2026-03-10
fix: correct HAVE_TE detection in remaining files

liuyun7345:
# What does this PR do?

PR #3763 fixed incorrect `HAVE_TE` detection in `multi_latent_attention.py`
and `shared_experts.py`. This PR applies the same fix to 4 additional files
that have the identical bug.

## Root Cause

Importing from `megatron.core.extensions.transformer_engine` never raises
`ImportError` because the module falls back to `MagicMock` when
TransformerEngine is unavailable. This causes `HAVE_TE` to be incorrectly
set to `True`.

## Fix

Add `import transformer_engine` before the extension imports to ensure
`ImportError` is properly raised when TE is not installed.

## Files Fixed

| File | Context |
|------|---------|
| `examples/multimodal/layer_specs.py` | Multimodal example layer specs |
| `examples/multimodal/radio/radio_g.py` | RADIO-G vision encoder specs |
| `tests/unit_tests/transformer/test_multi_token_prediction.py` | MTP unit tests |
| `tests/unit_tests/transformer/moe/test_upcycling.py` | MoE upcycling unit tests |

## Methodology

I audited all 45 files containing `HAVE_TE` in the codebase. The 30+
files in `megatron/core/` already use the correct pattern (importing
`transformer_engine` first). Only these 4 files in `examples/` and
`tests/` were missed.

Related to #3764

### DIFF
--- a/examples/multimodal/layer_specs.py
+++ b/examples/multimodal/layer_specs.py
@@ -17,6 +17,8 @@ from megatron.core.transformer.transformer_layer import TransformerLayer, Transf
 from megatron.core.typed_torch import not_none
 
 try:
+    import transformer_engine  # noqa: F401
+
     from megatron.core.extensions.transformer_engine import (
         TEColumnParallelLinear,
         TEDotProductAttention,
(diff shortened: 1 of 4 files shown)


returnL:
Hi, these changes are already covered in my PR #3763, so it might make sense to close this PR to avoid duplication.

PR #3772, opened 2026-03-10
Make nemotron6 config take --expert-tensor-parallel-size from ETP env.

yobibyte:
As in the title, ETP now makes nemotron6 model to set up the --expert-tensor-parallel-size.


Phlip79:
@yobibyte FYI we don't use the "Expert Review" label anymore

PR #3779, opened 2026-03-10
Replace rampup batch size scheduler with custom step batch size schedules

mkhona-nvidia:
## What does this PR do?

Replaces Megatron's linear rampup batch size scheduler (`--rampup-batch-size`) with a more flexible step-wise batch size schedule (`--step-batch-size-schedule`).

### Step batch size schedule

The new `--step-batch-size-schedule` argument accepts a string of `THRESHOLD:BATCH_SIZE` pairs that define arbitrary step-wise batch size changes during training. Thresholds support K/M/B/T suffixes and are interpreted as tokens when `--seq-length` is provided, otherwise as samples.

Example:
```
--step-batch-size-schedule "0:768 250B:1536 500B:3072 750B:6144"
```

This is implemented via a new `StepBatchsizeNumMicroBatchesCalculator` class.

### Rampup batch size removal

The old `RampupBatchsizeNumMicroBatchesCalculator` and `--rampup-batch-size` argument are removed. Existing example configs are converted to equivalent step batch size schedules.

### Files changed

- **`megatron/core/num_microbatches_calculator.py`**: Added `StepBatchsizeNumMicroBatchesCalculator`, removed `RampupBatchsizeNumMicroBatchesCalculator`, removed `rampup_batch_size` from all public API functions.
- **`megatron/training/config/training_config.py`**: Replaced `rampup_batch_size` field with `step_batch_size_schedule`.
- **`megatron/training/arguments.py`**: Removed rampup-related assertions and help text.
- **`megatron/training/training.py`**: Updated `update_train_iters` to handle step schedule for sample-based training.
- **`megatron/training/global_vars.py`**: Updated `init_num_microbatches_calculator` call.
- **`examples/`**, **`tests/functional_tests/`**: Converted rampup configs to step batch size schedules.
- **`tests/unit_tests/test_num_microbatches_calculator.py`**: Replaced rampup tests with step schedule tests.

### DIFF
--- a/megatron/core/num_microbatches_calculator.py
+++ b/megatron/core/num_microbatches_calculator.py
@@ -4,14 +4,16 @@
 
 import logging
 from abc import ABC, abstractmethod
-from typing import List, Optional, Union
+from typing import List, Optional, Tuple, Union
 
 logger = logging.getLogger(__name__)
 
 # TODO: global_var merge into mcore?
 _GLOBAL_NUM_MICROBATCHES_CALCULATOR: Union[
-    'ConstantNumMicroBatchesCalculator', 'RampupBatchsizeNumMicroBatchesCalculator'
-] = None
+    'ConstantNumMicroBatchesCalculator',
+    'RampupBatchsizeNumMicroBatchesCalculator',
+    'StepBatchsizeNumMicroBatchesCalculator',
+] = None  # type: ignore[assignment]
 
 
 def get_num_microbatches() -> int:
@@ -68,6 +70,8 @@ def init_num_microbatches_calculator(
     micro_batch_size: int,
     data_parallel_size: int,
     decrease_batch_size_if_needed: bool = False,
+    step_batch_size_schedule: Optional[str] = None,
+    seq_length: Optional[int] = None,
 ) -> None:
     """Initialize number of microbatches calculator. Supporting backward compatibility.
 
@@ -86,6 +90,15 @@ def init_num_microbatches_calculator(
         decrease_batch_size_if_needed (bool, optional):
             If true, scale down batch size to ensure divisibility by DP size * microbatch size.
             Defaults to False.
+        step_batch_size_schedule (Optional[str]):
+            Step batch size schedule string in format "THRESHOLD:BS THRESHOLD:BS ...".
+            Thresholds are interpreted as samples unless seq_length is provided, in which case
+            thresholds are interpreted as tokens and converted to samples.
+            Thresholds support suffixes: K (1e3), M (1e6), B (1e9), T (1e12).
+            Example: "0:768 250B:1536 500B:3072 750B:6144"
+        seq_length (Optional[int]):
+            Sequence length for token-to-sample conversion when using step_batch_size_schedule.
+            If provided, thresholds are interpreted as tokens. If None, thresholds are samples.
     """
     _configure_global_num_microbatches_calculator(
         rank,
@@ -94,6 +107,8 @@ def init_num_microbatches_calculator(
         micro_batch_size,
         data_parallel_size,
         decrease_batch_size_if_needed,
+        step_batch_size_schedule,
+        seq_length,
         init=True,
     )
 
@@ -111,6 +126,8 @@ def reconfigure_num_microbatches_calculator(
     micro_batch_size: int,
     data_parallel_size: int,
     decrease_batch_size_if_needed: bool = False,
+    step_batch_size_schedule: Optional[str] = None,
+    seq_length: Optional[int] = None,
 ) -> None:
     """Reconfigure number of microbatches calculator. Supporting backward compatibility.
 
@@ -129,6 +146,13 @@ def reconfigure_num_microbatches_calculator(
         decrease_batch_size_if_needed (bool, optional):
             If true, scale down batch size to ensure divisibility by DP size * microbatch size.
             Defaults to False.
+        step_batch_size_schedule (Optional[str]):
+            Step batch size schedule string in format "THRESHOLD:BS THRESHOLD:BS ...".
+            Thresholds support suffixes: K (1e3), M (1e6), B (1e9), T (1e12).
+            Example: "0:768 250B:1536 500B:3072 750B:6144"
+        seq_length (Optional[int]):
+            Sequence length for token-to-sample conversion when using step_batch_size_schedule.
+            If provided, thresholds are interpreted as tokens. If None, thresholds are samples.
     """
     _configure_global_num_microbatches_calculator(
         rank,
@@ -137,6 +161,8 @@ def reconfigure_num_microbatches_calculator(
         micro_batch_size,
         data_parallel_size,
         decrease_batch_size_if_needed,
+        step_batch_size_schedule,
+        seq_length,
         init=False,
     )
 
@@ -148,6 +174,8 @@ def _configure_global_num_microbatches_calculator(
     micro_batch_size: int,
     data_parallel_size: int,
     decrease_batch_size_if_needed: bool = False,
+    step_batch_size_schedule: Optional[str] = None,
+    seq_length: Optional[int] = None,
     init: bool = False,
 ) -> None:
     """Configure number of microbatches calculator. Can be used for initialization and
@@ -168,6 +196,13 @@ def _configure_global_num_microbatches_calculator(
         decrease_batch_size_if_needed (bool, optional):
             If true, scale down batch size to ensure divisibility by DP size * microbatch size.
             Defaults to False.
+        step_batch_size_schedule (Optional[str]):
+            Step batch size schedule string in format "THRESHOLD:BS THRESHOLD:BS ...".
+            Thresholds support suffixes: K (1e3), M (1e6), B (1e9), T (1e12).
+            Example: "0:768 250B:1536 500B:3072 750B:6144"
+        seq_length (Optional[int]):
+            Sequence length for token-to-sample conversion when using step_batch_size_schedule.
+            If provided, thresholds are interpreted as tokens. If None, thresholds are samples.
         init (bool, optional):
             If true, initialize the calculator. Defaults to False.
     """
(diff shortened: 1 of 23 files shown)


megatron/core/num_microbatches_calculator.py
```diff
@@ -506,3 +576,213 @@ def update(self, consumed_samples: int, consistency_check: bool, verbose: bool =
...
+        batch_size = self.schedule[0][1]
+        for threshold, bs in self.schedule:
+            if consumed_samples >= threshold:
+                batch_size = bs
+            else:
+                break
+        return batch_size
+
+    @staticmethod
+    def _round(batch_size: int, divisor: int) -> int:
```
mchrzanowski: I dont think we need --decrease-batch-size-if-needed; it makes this code way more complicated than it has to be. just remove it and dont let the two arguments work together

megatron/training/arguments.py
```diff
@@ -2319,6 +2319,39 @@ def _add_training_args(parser):
     group.add_argument('--batch-size', type=int, default=None,
                        help='Old batch size parameter, do not use. '
                        'Use --micro-batch-size instead')
+    group.add_argument('--global-batch-size', type=int, default=None,
```
mchrzanowski: we moved to training_config.py (thankfully); you'll need to refactor

mchrzanowski: nvm, I didnt see that most of the arguments are still present here

mchrzanowski (approved):
lgtm modulo comments

megatron/training/arguments.py
maanug-nv: '--global-batch-size' and '--rampup-batch-size' are already being generated by TrainingConfig, let's not add them back here. 

'--step-batch-size-schedule' is the only new one, and can be added to TrainingConfig and the args will be automatically generated.

maanug-nv: Regarding the args still present, not all of them fit into the existing configs and I haven't had a chance to decide where they should go.

maanug-nv (changes requested):
please add the new argument to TrainingConfig instead of 'arguments.py' . will help to more easily keep this feature in sync in Megatron Bridge.

megatron/training/arguments.py
mkhona-nvidia: Removed from `arguments.py` and added to `training_config`

maanug-nv (approved):
lgtm from training side

megatron/core/num_microbatches_calculator.py
```diff
@@ -83,6 +85,13 @@ def init_num_microbatches_calculator(
             Micro batch size at initialization.
         data_parallel_size (int):
             Data parallel size.
+        step_batch_size_schedule (Optional[str]):
+            Step batch size schedule string in format "THRESHOLD:BS THRESHOLD:BS ...".
+            Thresholds support suffixes: K (1e3), M (1e6), B (1e9), T (1e12).
+            Example: "0:768 250B:1536 500B:3072 750B:6144"
```
deepakn94, line 92: This is not that intuitive. Is it better to specify the number of tokens at each step? And then the last step does not need to be specified here because `--global-batch-size` already includes this?

mkhona-nvidia: We skip using --global-batch-size when using this step batch size. It is computed on the fly based on iteration number (that's why logic for checking world size divisibility was moved to training.py)

is that bad?

deepakn94: Hmm, keep that maybe?

deepakn94: Because I think of this as specifying a warmup schedule.

mchrzanowski: i think we should make the two options mututally exclusive, to avoid weird mixups from happening

mchrzanowski: i do think deepak is right that it's weird to mix and match tokens and samples but i for one think this approach is fine, because gbs is always denominated in terms of samples in megatron, but users think in terms of tokens. i like it as-is

megatron/core/num_microbatches_calculator.py
```diff
@@ -204,32 +245,57 @@ def _build_num_microbatches_calculator(
...
+    if step_batch_size_schedule is not None:
+        if decrease_batch_size_if_needed:
+            raise ValueError(
+                'Cannot specify both --step-batch-size-schedule and '
+                '--decrease-batch-size-if-needed'
             )
-    # Batch size ramp up.
-    else:
+        if global_batch_size is not None and rank == 0:
+            logger.warning('--global-batch-size is ignored when using --step-batch-size-schedule')
```
deepakn94: Warnings are useless, exception.

mkhona-nvidia: Should i change the warning to an exception? global_batch_size is populated not here, but in the arguments file

mkhona-nvidia: Changed to raising error in both `arguments.py `and `yaml_arguments.py`

megatron/core/num_microbatches_calculator.py
```diff
@@ -204,32 +245,55 @@ def _build_num_microbatches_calculator(
...
+            Ignored when step_batch_size_schedule is provided.
         micro_batch_size (int):
             Micro batch size at initialization.
         data_parallel_size (int):
             Data parallel size.
         decrease_batch_size_if_needed (bool):
             If true, scale down batch size to ensure divisibility by DP size * microbatch size.
-
+        step_batch_size_schedule (Optional[str]):
+            Step batch size schedule string in format "THRESHOLD:BS THRESHOLD:BS ...".
```
jaredcasper, line 235: Maybe include in this docstring as well that thresholds are in samples unless seq_length is supplied, in which case they are in tokens. People might not noticed the seq_length docstring.

megatron/training/arguments.py
```diff
@@ -562,6 +564,13 @@ def validate_args(args, defaults={}):
     # Batch size.
     assert args.micro_batch_size is not None
     assert args.micro_batch_size > 0
+    explicit_global_batch_size = getattr(
```
jaredcasper: Put this check into `TrainingConfig.__post_init__()` if possible.

megatron/core/num_microbatches_calculator.py
mkhona-nvidia: Done

megatron/training/arguments.py
mkhona-nvidia: Here's what Claude says (I am not as familiar with the new TrainingConfig structure):

`TrainingConfig.__post_init__()` won’t help much here because TrainingConfig is not actually instantiated in the current CLI/YAML validation flow. Right now it is used as a schema for argparse generation via ArgumentGroupFactory, while the real validation happens later on an argparse.Namespace / SimpleNamespace in `validate_args()` and `validate_yaml()`. I also checked the repo and there are no current `TrainingConfig(...) ` instantiations, so moving this check only into `__post_init__()` would make it effectively dead for the normal path.

mkhona-nvidia: (the check exists in validate_yaml: https://github.com/NVIDIA/Megatron-LM/pull/3779/changes#diff-7575a9e65a444850ccde2a28412be75339115166804fd678aa50842977253b1cR106)

megatron/core/num_microbatches_calculator.py
```diff
@@ -168,6 +194,13 @@ def _configure_global_num_microbatches_calculator(
         decrease_batch_size_if_needed (bool, optional):
             If true, scale down batch size to ensure divisibility by DP size * microbatch size.
             Defaults to False.
+        step_batch_size_schedule (Optional[str]):
```
deepakn94, line 199: Does it make sense to name this `rampup_batch_size_schedule` to match the existing argument?

megatron/training/arguments.py
```diff
@@ -102,6 +102,8 @@ def parse_args(extra_args_provider=None, ignore_unknown_args=False):
     else:
         args = parser.parse_args()
 
+    args._explicit_global_batch_size = args.global_batch_size is not None
```
deepakn94: Why do we need this in args?

megatron/core/num_microbatches_calculator.py
mkhona-nvidia: they are 2 different schedules; `rampup_batch_size_schedule` should only exist for legacy reasons, the new schedule is more flexible

megatron/training/arguments.py
mkhona-nvidia: We need to preserve whether global_batch_size was explicitly provided by the user, because later validation needs to distinguish that from the case where global_batch_size is auto-derived from `micro_batch_size * data_parallel_size`. A local variable in `parse_args() `would be lost once that function returns, while `validate_args()` only receives args. Storing the flag on args is the least invasive way to carry that parser-time information into the later validation step without widening function signatures or trying to recompute it after `args.global_batch_size` may already have been mutated.

megatron/core/num_microbatches_calculator.py
deepakn94: I agree that this is more flexible, but it is still a batch size rampup schedule? I like the parallelism in argument naming because they're doing similar things, this is just more expressive.

megatron/training/arguments.py
deepakn94: Ok, can we make it clearer that this is a bool? E.g., `args._is_global_batch_size_explicitly_specified`. Or something less verbose.

megatron/core/num_microbatches_calculator.py
mkhona-nvidia: The form of the argument is different for both though,

`rampup: [start_global_batch_size, batch_size_increment, ramup_samples]`

`step: "THRESHOLD:BS THRESHOLD:BS ..."`

megatron/training/arguments.py
mkhona-nvidia: Done

megatron/core/num_microbatches_calculator.py
deepakn94: Sure, but they're doing similar things (specifying how batch size needs to be increased to get to steady-state batch size).

deepakn94: Basically I am being a bit pushy here because a) we have too many Megatron-LM arguments and it's easy to get confused, b) these two arguments are doing the same high-level thing, they're just specified in different formats and what you've implemented is more expressive.

mkhona-nvidia: Can we remove the older batch size rampup? It is a really bad schedule and is not useful

deepakn94: Ok, please remove it then.

deepakn94 (approved):

megatron/core/num_microbatches_calculator.py
deepakn94: You can say it's being deprecated by this new schedule that is more expressive.

tdene (approved):

jaredcasper (approved):

yaoyu-33 (approved):

PR #3787, opened 2026-03-10
Add parallel MLA inference tests, fix corresponding issues

janEbert:
MLA used to not be tested for parallel dynamic inference. This PR implements appropriate test parameters into `TestDynamicInferenceEngine.test_parallel_inference` to test MLA with and without latent caching (although MLA currently only works with latent caching).

The test itself would fail without further changes, so the PR contributes several fixes to the code and improvements like making FlashMLA optional or making the local Transformer implementation work with MLA. For example, FlashMLA expects a block size of exactly 64 while FlashAttention expects a block size divisible by 256. In the generative phase, FlashMLA is not used, only FlashAttention, so any inference work other than decode-only tasks would not work with MLA without the changes in this PR.

Finally, tests for Mamba with the inference-optimized Transformer implementation are enabled since they pass without issues.


janEbert:
Added support for local Transformer implementation, added functional tests.

PR #3795, opened 2026-03-11
[Main][feat] Support overlapping A2A Combine backprop with wgrad GEMM

Wohox:
# What does this PR do ?

PR for dev: https://github.com/NVIDIA/Megatron-LM/pull/3766
### Problem

In MoE models, the expert weight gradient (wgrad) computation during backward is serialized on the main CUDA stream. This blocks the data gradient (dgrad) from flowing to earlier layers until the expert wgrad finishes, even though there is no data dependency between them. The result is wasted GPU cycles — earlier layers' backward pass sits idle waiting for expert wgrad to complete.

With FSDP, this is further compounded because the gradient reduce-scatter for expert parameters is also blocked on the same critical path.

### Solution

This PR introduces a new flag `--delay-wgrad-compute-for-te-grouped-gemm` that separates the expert wgrad computation from the main backward stream:

1. **Two autograd functions** are inserted into the MoE layer's forward graph:
   - `_RecordExpertDgradCompletion` — placed before the expert computation; during backward, it records a CUDA event once the expert dgrad is done.
   - `_RegisterDelayedWgradForExperts` — placed at the dispatch boundary; during backward, it waits on the dgrad event, then launches `backward_dw()` on a **dedicated CUDA stream**, and synchronizes back to the main stream before proceeding.

2. **FSDP integration** — When used with MegatronFSDP, expert parameters are marked with `_fsdp_delay_grad_reduce = True` so the normal post-accumulate-grad hook skips them. A callback is registered via `register_process_expert_grads_fn()` that triggers the FSDP reduce-scatter for expert parameters only after the delayed wgrad computation completes.

3. **TE GroupedLinear** is configured with `delay_wgrad_compute=True`, which tells Transformer Engine to skip wgrad during the normal autograd backward and instead wait for an explicit `backward_dw()` call.

### How to enable

```bash
--delay-wgrad-compute-for-te-grouped-gemm
```

**Requirements:**
- Transformer Engine >= 2.3.0
- `moe_grouped_gemm` enabled (not legacy grouped gemm)
- Mutually exclusive with `--delay-wgrad-compute` (the existing A2A-overlap-based delay)
- Mutually exclusive with `--overlap-moe-expert-parallel-comm`

Works with both **FSDP** and **3-D parallelism** (TP/EP/PP).

### What is achieved

The expert wgrad computation runs on a separate CUDA stream, overlapping with the EP communication within the same transformer layer. This reduces the wall-clock time of the backward pass without changing numerical results — the feature is **bit-exact** with the non-delayed baseline (verified by unit tests comparing per-step losses and final weights over multiple optimizer steps).

### Changes

| File | Description |
|------|-------------|
| `megatron/core/model_parallel_config.py` | New config flag `delay_wgrad_compute_for_te_grouped_gemm` |
| `megatron/core/transformer/transformer_config.py` | Validation assertions for the new flag |
| `megatron/core/transformer/moe/moe_layer.py` | Autograd functions for delayed wgrad + dedicated CUDA stream/event + `register_process_expert_grads_fn` callback |
| `megatron/core/extensions/transformer_engine.py` | Pass `delay_wgrad_compute=True` to TE GroupedLinear when the new flag is set |
| `megatron/core/distributed/fsdp/.../megatron_fsdp.py` | FSDP hook to defer reduce-scatter for expert params and trigger it after delayed wgrad |
| `tests/unit_tests/a2a_overlap/test_delay_wgrad_compute.py` | Unit tests covering basic, shared-expert, multi-layer, and FSDP scenarios |

### Test plan

- [x] Unit test: `test_delay_wgrad_compute_for_te_grouped_gemm` — full-model training loop (forward → backward → optimizer) comparing delayed vs. non-delayed across `num_layers × shared_experts × dispatcher_type × fp8_flag`
- [x] Unit test: `test_delay_wgrad_compute_for_te_grouped_gemm_with_fsdp` — same comparison with MegatronFSDP wrapping (`fully_shard_model` + `fully_shard_optimizer`), verifying the deferred reduce-scatter path

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/core/model_parallel_config.py
+++ b/megatron/core/model_parallel_config.py
@@ -261,7 +261,7 @@ class ModelParallelConfig:
     delay_wgrad_compute: bool = False
     """Delay the weight gradient computation to improve batch-level communication overlapping"""
 
-    delay_wgrad_compute_for_te_grouped_gemm: bool = False
+    overlap_dispatch_backward_with_experts_wgrad: bool = False
     """Delay the weight gradient computation for TE Grouped GEMM MoE experts.
     When enabled with FSDP, the expert weight gradients are computed on a separate
     CUDA stream after the data gradients finish, allowing overlap of wgrad compute
--- a/megatron/core/model_parallel_config.py
+++ b/megatron/core/model_parallel_config.py
@@ -261,6 +261,15 @@ class ModelParallelConfig:
     delay_wgrad_compute: bool = False
     """Delay the weight gradient computation to improve batch-level communication overlapping"""
 
+    delay_wgrad_compute_for_te_grouped_gemm: bool = False
+    """Delay the weight gradient computation for TE Grouped GEMM MoE experts.
+    When enabled with FSDP, the expert weight gradients are computed on a separate
+    CUDA stream after the data gradients finish, allowing overlap of wgrad compute
+    with EP A2A communication. The FSDP gradient reduce-scatter for
+    expert parameters is deferred until the delayed wgrad computation completes.
+    This requires transformer_engine with GroupedLinear support (TE >= 2.3.0).
+    """
+
     ep_overlap_early_attn_memory_release: bool = False
     """Enable early memory release of attention activations during EP overlap.
     EP overlap can increase peak memory usage when the overlapped forward module allocates 
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
@@ -696,7 +696,8 @@ class MegatronFSDP(torch.nn.Module):
             # is set but the delayed hook was never installed, process the parameter
             # immediately as a safety fallback to avoid silently dropping gradients.
             param_list = [
-                p for p in param_list
+                p
+                for p in param_list
                 if not getattr(p, 'skip_backward_post_hook', False)
                 or not hasattr(p, 'post_wgrad_grad_acc_hook')
             ]
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
@@ -76,7 +76,7 @@ class TrainingState(Enum):
 def setup_delayed_wgrad_acc_hook(module, grad_acc_func):
     """Configure delayed wgrad gradient processing for MoE expert parameters.
 
-    When ``delay_wgrad_compute_for_te_grouped_gemm`` is enabled on a TransformerLayer,
+    When ``overlap_dispatch_backward_with_experts_wgrad`` is enabled on a TransformerLayer,
     this function:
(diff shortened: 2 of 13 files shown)


shjwudp (approved):

cspades (approved):

Wohox:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/3795/commits/da0f997cbfd5bc21385e0ad29f1c670224a57b3f

zhongbozhu:
Maybe dumb question, but why do we need a new toggle for this?

Victarry:
Comments posted on https://github.com/NVIDIA/Megatron-LM/pull/3766, please take a look. Thanks!

Wohox:
> Maybe dumb question, but why do we need a new toggle for this?

As discussed in #3766, we can make this the default option when `overlap_moe_expert_parallel_comm=False`

yaoyu-33 (approved):

kvareddy:
@fanshiqing  can you please take a look at this MR?

fanshiqing (approved):

megatron/core/model_parallel_config.py
```diff
@@ -261,6 +261,15 @@ class ModelParallelConfig:
     delay_wgrad_compute: bool = False
     """Delay the weight gradient computation to improve batch-level communication overlapping"""
 
+    delay_wgrad_compute_for_te_grouped_gemm: bool = False
```
jaredcasper: I don't think "te" should be in the name of the option. The fact that it is implemented in TE is an implementation detail, it should not be part of the interface. This should just be `delay_wgrad_compute_for_grouped_gemm`, or better I think would be `delay_grouped_gemm_wgrad_compute`.

Also, why is this different from `delay_wgrad_compute`? Can't you just do this optimization when `delay_wgrad_compute` is True? At the least the documentation should explain these two options.

Wohox: > I don't think "te" should be in the name of the option
* I agree, the name has been changed into `overlap_dispatch_backward_with_experts_wgrad`, this should be more informative and can be easily distinguished with `delay_wgrad_compute`.

> why is this different from delay_wgrad_compute? Can't you just do this optimization when delay_wgrad_compute is True? At the least the documentation should explain these two options.

* `delay_wgrad_compute` affects all TE modules, whereas `overlap_dispatch_backward_with_experts_wgrad` only affects `TEGroupedGemm` module, there is a discussion why we cannot extend this feature to overlap `combine_backward`: [link](https://github.com/NVIDIA/Megatron-LM/pull/3766#discussion_r2950529603). 
* We can switch from using new flag to setting `delay_wgrad_compute` to true and `1f1b_overlap` to false. However, the semantics of `delay_wgrad_compute` would become somewhat confusing. This is because it wouldn't actually be performing delay wgrad for all available TE modules.

jaredcasper (approved):

ilml:
/ok to test [81c8e38](https://github.com/NVIDIA/Megatron-LM/pull/3795/commits/81c8e3891d8633382dd22fc9e0a35c2a72f90566)

PR #3797, opened 2026-03-11
[Main][feat] Support A2A Overlap for Megatron-FSDP

Wohox:
# What does this PR do ?
PR for dev: https://github.com/NVIDIA/Megatron-LM/pull/3796
Based on: https://github.com/NVIDIA/Megatron-LM/pull/3795

### Supported Feature Combinations
```
# without param sharding
--use-megatron-fsdp --data-parallel-sharding-strategy "optim_grads" --overlap-moe-expert-parallel-comm
--use-megatron-fsdp --data-parallel-sharding-strategy "optim_grads" --overlap-moe-expert-parallel-comm --delay-wgrad-compute


# param sharding
--use-megatron-fsdp --data-parallel-sharding-strategy "optim_grads_params" --overlap-moe-expert-parallel-comm
--use-megatron-fsdp --data-parallel-sharding-strategy "optim_grads_params" --overlap-moe-expert-parallel-comm --delay-wgrad-compute
```

### Problem

The EP overlap schedule (`overlap_moe_expert_parallel_comm`) interleaves forward and backward passes at the sub-module level (attn, mlp, moe_dispatch, moe_combine) to overlap Expert Parallel A2A communication with compute. This schedule bypasses `TransformerLayer.forward()` entirely, calling sub-modules directly. As a result, FSDP hooks registered on `TransformerLayer` — including parameter all-gather, parameter release, gradient reduction, and root pre/post backward — are never triggered, making FSDP with `optim_grads_params` incompatible with this schedule.

### FSDP Hook Triggering: Standard vs 1F1B Overlap

| Hook | Standard Path (non-1F1B) | 1F1B Overlap Path | Reason for Difference |
|---|---|---|---|
| **`_replace_param_with_raw_if_needed`** | Called inside `MegatronFSDP.forward()` at the start of every forward call. | Called **once** at the top of `combined_1f1b_schedule_for_no_pipelining`, before any schedule phase begins. | The overlap schedule bypasses `MegatronFSDP.forward()` entirely — it calls `GPTModel.build_schedule_plan()` on the unwrapped module directly. The param swap must happen explicitly before the schedule accesses layers. |
| **`_pre_forward_param_unshard`** (per-layer all-gather) | Registered as `nn.Module.register_forward_pre_hook` on each FSDP unit. Triggered **automatically** by PyTorch when `TransformerLayer.forward()` is called. | **Same mechanism** — still triggered by PyTorch forward pre-hooks. With `enable_fine_grained_param_gather=True`, hooks are registered on **every sub-module** (not just FSDP units), giving finer-grained all-gather. | The schedule still calls each layer's `forward()` via `ScheduleNode`, which triggers the registered forward pre-hooks normally. No manual intervention needed. |
| **`_post_forward` (post_forward_release_module)** — per-layer param release after forward | Registered as `nn.Module.register_forward_hook` on each FSDP unit. Triggered **automatically** by PyTorch after `TransformerLayer.forward()` returns. | Called **manually** by the schedule plan via `set_fsdp_reshard_hooks`. Wired in `combined_forward_backward_step` onto each `TransformerLayerSchedulePlan`, and invoked after the last forward node (mlp/moe_combine) of each layer completes. | The schedule decomposes `TransformerLayer.forward()` into sub-ops (attn, mlp, moe_dispatch, moe_combine). The registered `register_forward_hook` on `TransformerLayer` is **never fired** because the schedule never calls `TransformerLayer.forward()` as a whole — it calls sub-modules directly. So the release must be done explicitly. |
| **`_post_backward_release_module`** — per-layer param release after backward | Registered via `_register_post_backward_hook` (a `register_forward_pre_hook` that inserts a `RegisterFSDPBackwardFunction` into the autograd graph). Triggered **automatically** by autograd during backward. | Called **manually** by the schedule plan via `set_fsdp_reshard_hooks`. Wired onto `TransformerLayerSchedulePlan.attn` (the last backward node of each layer). | Same reason as `_post_forward`: the schedule runs backward per-sub-op, not per-`TransformerLayer`. The autograd-based post-backward hook on the FSDP unit output is never reached because the layer-level forward was never called as a whole. |
| **`_pre_backward_param_unshard`** (per-layer backward all-gather) | Attached via `create_custom_backward_hook` (a `register_forward_hook` that registers `register_multi_grad_hook` on the output tensors). Triggered **automatically** by autograd when gradients arrive at the layer's output. | With `enable_fine_grained_param_gather=True`, hooks are registered on **every sub-module** via `create_custom_backward_hook`. Triggered **automatically** by autograd when gradients flow through each sub-module. | When `enable_fine_grained_param_gather=False`, hooks only exist on FSDP units; since the layer-level forward is skipped, those hooks are never hit. With `=True`, hooks exist on every sub-module, so autograd still triggers them during the schedule's backward. |
| **`_root_pre_backward`** | Attached via `create_custom_backward_hook` on the root module's output. Triggered **automatically** by autograd (gradient flows back through root output). Also queues `_root_post_backward` via `queue_callback`. | Called **manually** via `fsdp_wrapper.pre_backward()` at the start of each `combined_forward_backward_step` that has a backward model (`b_model is not None`). Called with `skip_backward_hook=True` so it does **not** queue `_root_post_backward`. | The overlap schedule does not call `MegatronFSDP.forward()`, so no root-level output tensor exists in the autograd graph to trigger the hook. It must be called manually. `skip_backward_hook=True` because `_root_post_backward` is also called manually. |
| **`_root_post_backward`** (grad accumulation + reduce-scatter) | Queued via `torch.autograd.Variable._execution_engine.queue_callback()` inside `_root_pre_backward`. Triggered **automatically** by the autograd engine after the entire backward pass finishes. | Called **manually** via `fsdp_wrapper.post_backward()` at the end of each `combined_forward_backward_step` that has a backward model. | Since `_root_pre_backward` is called with `skip_backward_hook=True`, the autograd `queue_callback` is skipped. The schedule must call `post_backward()` explicitly after all layer backward ops complete to trigger gradient accumulation and reduce-scatter. |
| **`_process_post_backward_gradients`** (per-param grad accumulation) | Registered via `param.register_post_accumulate_grad_hook`. Triggered **automatically** by autograd when `.grad` is accumulated on each parameter. | **Same mechanism** w.o. `delay_wgrad_compute` — still triggered automatically by autograd. Parameters are the same leaf tensors regardless of which code path invokes the computation. **Manual Triggered** with `delay_wgrad_compute` since grad acc hook needs to be delayed as well. | This is a parameter-level hook (not module-level), so it fires whenever autograd writes to `.grad`, independent of whether the compute was driven by `module.forward()` or the schedule. |

**Summary**: The root cause of all differences is that the 1F1B overlap schedule **decomposes** `TransformerLayer.forward()` into sub-operations (attn, mlp, moe_dispatch, moe_combine) and interleaves them across microbatches. This means:

1. `MegatronFSDP.forward()` is never called → `_replace_param_with_raw_if_needed` must be explicit.
2. `TransformerLayer.forward()` is never called as a whole → module-level `register_forward_hook` / `register_forward_pre_hook` on the FSDP unit are not fired → `_post_forward` and `_post_backward_release_module` must be wired manually via `set_fsdp_reshard_hooks`.
3. No root-level output tensor in autograd graph → `_root_pre_backward` / `_root_post_backward` must be called manually.
4. Hooks that operate at the **parameter level** (`_process_post_backward_gradients`) or **sub-module level** (`_pre_forward_param_unshard`, `_pre_backward_param_unshard` with `enable_fine_grained=True`) continue to work automatically because the schedule still calls sub-module forwards, and autograd still writes to parameter `.grad`.

### Solution

Bridge the gap between Megatron-FSDP's hook-based lifecycle and the EP overlap schedule's direct sub-module invocation by:
1. Enabling fine-grained per-sub-module all-gather hooks so each sub-module can unshard its own parameters.
2. Exposing manual handles for FSDP lifecycle events that the schedule must drive explicitly.

### Code changes

**`combined_1f1b.py`** — Schedule-level FSDP integration
- Add `find_megatron_fsdp()` helper to locate the FSDP wrapper from the model chain.
- Call `fsdp_wrapper._replace_param_with_raw_if_needed()` before the schedule starts (normally done inside `MegatronFSDP.forward()`).
- Manually invoke `fsdp_wrapper.pre_backward()` / `post_backward()` around the combined forward+backward step since the schedule bypasses the autograd hooks that normally trigger them.
- Wire per-layer FSDP release callbacks via `set_fsdp_reshard_hooks()`.

**`megatron_fsdp.py`** — Expose manual hook handles & support delayed wgrad
- Add `skip_backward_hook` parameter to `_root_pre_backward` so the manual caller can prevent the redundant `queue_callback(_root_post_backward)`.
- Expose `pre_backward`, `post_backward`, `post_forward_release_module`, and `post_backward_release_module` as public attributes for external schedule use.
- Add `is_delayed` parameter to `_process_post_backward_gradients` to distinguish delayed wgrad gradient processing from the standard path, preventing deferred parameters from being filtered out by the `skip_backward_post_hook` guard.
- Register fine-grained backward all-gather hooks (`_register_pre_backward_param_unshard_hook`) on every sub-module when `overlap_moe_expert_parallel_comm` + `optim_grads_params`, so backward all-gather works at the sub-module level.
- In per-parameter `register_post_accumulate_grad_hook`, skip parameters with `skip_backward_post_hook=True` so their gradient processing is deferred to the delayed wgrad path.

**`mcore_fsdp_adapter.py`** — Auto-configure FSDP for EP overlap
- Enable `enable_fine_grained_param_gather_hook` when `overlap_moe_expert_parallel_comm` is set (for all sharding strategies), so every sub-module gets its own forward all-gather/unshard hooks.
- Assert that `fsdp_double_buffer` and partial CUDA graph scopes are disabled when `overlap_moe_expert_parallel_comm` is set.
- Assert that `fsdp_unit_modules` must be `[TransformerLayer]` when combined with `optim_grads_params`.

**`utils.py`** — Config attribute helper
- Add `get_unwrapped_transformer_config_attribute()` to safely access `TransformerConfig` attributes from within FSDP internals, where the config type may not always be available.

**`model_chunk_schedule_plan.py`** — Per-layer FSDP hook wiring
- Add `set_fsdp_reshard_hooks()` to `TransformerLayerSchedulePlan` that attaches post-forward and post-backward release callbacks to the correct schedule nodes (last forward node, first backward node).

**`fine_grained_callables.py`** — Node-level FSDP hooks
- Override `forward()` / `backward()` in `TransformerLayerNode` to invoke FSDP reshard hooks at layer boundaries.
- In `backward_dw()`, collect and execute delayed `post_wgrad_grad_acc_hook` for FSDP gradient processing, then invoke the reshard hook.
- Add `set_fsdp_post_forward_reshard_hook()` / `set_fsdp_post_backward_reshard_hook()` registration methods.
- Add `parameters()` iterator to `_BackwardDWWrapper` so FSDP can discover parameters owned by the attn and shared-expert backward-dw callables.
### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -697,14 +701,15 @@ def _process_post_backward_gradients(param_list):
...
-                    getattr(p, 'skip_backward_post_hook', False)
-                    and hasattr(p, 'post_wgrad_grad_acc_hook')
-                )
-            ]
+            if not is_delayed:
+                param_list = [
+                    p
+                    for p in param_list
+                    if not (
+                        getattr(p, 'skip_backward_post_hook', False)
```
shjwudp: Does this line fall under the condition checked at line 1093? If so, it could be removed to simplify the logic. Also, could you point out where `is_delayed` is being called?

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -1007,12 +1014,26 @@ def _register_pre_backward_param_unshard_hook(module):
                 create_custom_backward_hook(module, _pre_backward_param_unshard)
             )
 
+        # Saved function handle for manual hook management.
+        self.post_forward_release_module = partial(_post_forward, input=None, output=None)
+        self.post_backward_release_module = _post_backward_release_module
+        self.pre_backward = partial(_root_pre_backward, module=None, skip_backward_hook=True)
+        self.post_backward = _root_post_backward
```
shjwudp, line 1015: Can we add the following functions — `module_unshard`, `module_reshard`, `root_module_pre_backward`, `root_module_post_backward`, and `post_wgrad_grad_acc` and skip register hooks when "overlap_moe_expert_parallel_comm=True"? These are all FSDP hooks, and keeping them together here would make the code easier to read and maintain.

e.g.,
```python
self.module_unshard_pre_forward = _pre_forward_param_unshard
self.module_unshard_pre_backward = _pre_backward_param_unshard
self.module_reshard = release_module_parameters
self.module_grad_acc = _grad_acc # need implement per-module grad acc function
self.root_module_pre_backward = _root_pre_backward
self.root_module_post_backward = _root_post_backward
```

shjwudp (commented):
Review of the M-FSDP-related changes: This PR unblocks the overlap feature between M-FSDP and A2A by exposing key M-FSDP hooks, which are manually triggered within the A2A pipeline.

Great that we can now have this feature! There are still a few minor points we need to discuss.

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
shjwudp: Discussed with Pingtian — we can’t remove all FSDP hook registrations in the 1F1B (A2A) overlap, since some modules (e.g., the embedding layer) are not managed by the 1F1B overlap pipeline.

Should add comments to clarify the purpose of these functions.

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
Wohox: For params with valid `post_wgrad_grad_acc_hook` and identified as delayed(with `skip_backward_post_hook`), they should not be executed right now but delayed, o.w. they are kept in `param_list`. This safety guard is redundant due to another safety guard at L1090. Will delete this filter and change to assertion instead.

shjwudp (approved):

ko3n1g (approved):

megatron/core/pipeline_parallel/combined_1f1b.py
```diff
@@ -20,6 +20,20 @@
 Shape = Union[List[int], torch.Size]
 
 
+def find_megatron_fsdp(model):
```
jaredcasper, line 23: Should this live in the the fsdp code?

megatron/core/models/gpt/fine_grained_callables.py
```diff
@@ -348,6 +397,16 @@ def backward_dw(self):
 
         self.bwd_dw_callables = None
 
+    def set_fsdp_post_forward_reshard_hook(self, hook):
```
jaredcasper, line 400: Are these hooks specific to FSDP? Maybe better just to have generic names? (i.e. just `post_forward_hook` not `fsdp_post_forward_reshard_hook`)

Wohox: These hooks are specific for FSDP, that's why it's called `fsdp_*`, will change the name to be more general.

megatron/core/pipeline_parallel/combined_1f1b.py
Wohox: Maybe the utils.py under pipeline_parallel folder? There is a similar method to get unwrapped GPTModel from DDP instance.

lhb8125 (approved):
LGTM, thanks!

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -157,6 +158,31 @@ def __init__(
...
+            )
+            partial_cuda_graph_scopes = [
+                scope
+                for scope in config.cuda_graph_scope
+                if scope
+                not in (CudaGraphScope.full_iteration, CudaGraphScope.full_iteration_inference)
+            ]
+            assert not partial_cuda_graph_scopes, (
+                "1F1B overlap with FSDP does not support partial CUDA graph scopes "
+                f"({partial_cuda_graph_scopes}). "
```
fanshiqing, line 174: QQ: what's blocking issue for supporting with partial cuda graphs here?

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -1008,12 +1005,30 @@ def _register_pre_backward_param_unshard_hook(module):
...
+
+        config = getattr(root_module, "config", None)
+        enable_fine_grained_param_gather_backward_hook = (
+            get_unwrapped_transformer_config_attribute(config, "overlap_moe_expert_parallel_comm")
+            and self.ddp_config.data_parallel_sharding_strategy == "optim_grads_params"
+        )
```
cspades: Nit: Can this be in the DDPConfig like `enable_fine_grained_param_gather_hook` so we don't hide anymore logic inside MFSDP source code?

This is a MCore toggle, so I think it would be best to be in the MCore adapter right? (We had `use_decoupled_grad` hidden in MFSDP for a very long time....)

(If it goes into DDPConfig, make sure to name it with a `megatron_fsdp_...` prefix.)

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -169,8 +195,16 @@ def __init__(
...
+                # parameters individually.  This applies to all sharding
+                # strategies (not only optim_grads_params) because the hooks
+                # also maintain per-module training-state bookkeeping that the
+                # gradient-reduction pipeline relies on.
                 enable_fine_grained_param_gather_hook=(
-                    config.fp8_recipe == "mxfp8" and ddp_config.fp8_param_gather
+                    (config.fp8_recipe == "mxfp8" and ddp_config.fp8_param_gather)
+                    or config.overlap_moe_expert_parallel_comm
                 ),
             ),
```
cspades, line 213: If we could set `enable_fine_grained_param_gather_backward_hook` here that would be great!

cspades (approved):
I have some nits, but I guess this can go in immediately.

wujingyue (commented):
Thanks for the PR! 

Regarding the FSDP changes, this PR introduces knobs to skip certain hooks so they can be triggered manually.

I'd suggest a slightly different approach:
1. Expose the hooks as public methods. For example, instead of assigning `self.post_backward_release_module = _post_backward_release_module`, rename `_post_backward_release_module` to `def post_backward_hook(...)`.
2. Expose the corresponding RemovableHandles as well, so callers can explicitly remove them (e.g., `self.post_backward_hook_handle.remove()`).

This would make the control more explicit and the code easier to reason about.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
Wohox: partial cuda graph is not supported for FSDP yet.

Wohox:
> Thanks for the PR!
> 
> Regarding the FSDP changes, this PR introduces knobs to skip certain hooks so they can be triggered manually.
> 
> I'd suggest a slightly different approach:
> 
> 1. Expose the hooks as public methods. For example, instead of assigning `self.post_backward_release_module = _post_backward_release_module`, rename `_post_backward_release_module` to `def post_backward_hook(...)`.
> 2. Expose the corresponding RemovableHandles as well, so callers can explicitly remove them (e.g., `self.post_backward_hook_handle.remove()`).
> 
> This would make the control more explicit and the code easier to reason about.

I agree, but there is refactor work from @shjwudp, so I will leave it out of the scope for this PR~

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
Wohox: Moved to mcore_fsdp_adapter.py~

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
Wohox: Done!

wujingyue:
> I agree, but there is refactor work from @shjwudp, so I will leave it out of the scope for this PR~

That's fair!

ericharper (approved):

megatron/core/pipeline_parallel/combined_1f1b.py
```diff
@@ -356,6 +383,25 @@ def forward_backward_step():
                 f_schedule_plan, AbstractSchedulePlan
             ), "first output of forward_step_func must be one instance of AbstractSchedulePlan"
 
+        # Wire per-layer FSDP parameter release callbacks.  The EP overlap
+        # schedule bypasses normal FSDP forward/backward hooks, so we release
+        # each layer's all-gathered parameters explicitly after its compute.
+        # Only needed for optim_grads_params strategy (where params are sharded).
+        forward_fsdp_wrapper = find_megatron_fsdp(f_model)
```
jaredcasper, line 390: Can these be implemented as a generic hook instead of FSDP specific code? In general want to avoid code like "if FSDP: do special FSDP stuff" that I'm seeing more and more.

wujingyue: > "if FSDP: do special FSDP stuff" that I'm seeing more and more.

Great point. 

To help us brainstorm, @Wohox, what overlap schedule are you aiming to achieve? The PR description didn’t make that fully clear. Users can already register hooks on more [fine-grained](https://github.com/NVIDIA/Megatron-LM/blob/5fe3f06658b393283eafc32f039c3d0c2468a69f/megatron/core/distributed/fsdp/mcore_fsdp_adapter.py#L98)
FSDP units, so I’m curious what additional capability or flexibility you’re looking for. Understanding that would help us think about how Megatron FSDP could evolve to better support these use cases in the future.

cspades: Can answer @wujingyue 's question to my best knowledge, but 1F1B doesn't follow the usual FSDP FWD-BWD flow, it likely interleaves and overlaps them not only at a module level, but also at a operational level for MoE EP:
```
Layer 1 Attention FWD <> Layer N Combine BWD
Layer 1 Dispatch FWD <> Layer N MLP DGrad + WGrad BWD
Layer 1 MLP FWD <> Layer N Dispatch BWD
Layer 1 Combine FWD <> Layer N Attention DGrad + WGrad BWD

Layer 2 Attention FWD <> Layer N-1 Combine BWD
...
```
(more or less @Wohox correct me if I'm wrong)

FSDP hooks need to respect all of this^ mess, unsharding parameters needed for MLP and Attention FWD and DGrad. Because these operations are done inside of an FSDP unit, you also cannot re-shard at the wrong time either, i.e. after DGrad MLP, you shouldn't re-shard yet since you're gonna use those weights almost immediately for the FWD MLP.

I'm not 100% sure if this implementation is the best approach to enable the above though. It's a bit confusing to me still. I might read through the schedule and draw an updated picture with FSDP AG and RS alongside the 1F1B MoE A2A timeline.

cspades: @jaredcasper @wujingyue I think even from XTP (extended TP) it's becoming quite clear that if you want model-sharded data parallelism, fine-grained or customizable hooks (or straight up model code in the XTP case) embedded in model code that doesn't fit the FSDP paradigm but re-cycles a ton of FSDP code and offers significant performance improvements, will likely lead to something that looks like this.

We'd best start writing these patterns down to make things as clean and robust (i.e. not affecting other paths if FSDP is not used) as possible if we want to make FSDP reach its potential.

Wohox: Thanks @cspades for the clear explanation, and yes, that matches my understanding.

The root cause of the FSDP-specific branch here is that EP overlap drives forward/backward at the submodule/op level through fine-grained callables, while the current FSDP unit lifecycle is still organized around coarser module forward/backward boundaries. Even with `enable_fine_grained_param_gather_hook` and `enable_fine_grained_param_gather_backward_hook` to bypass the usual FSDP unit boundary for param unshard, some FSDP lifecycle hooks still cannot be reached at the correct points in this schedule. The table in the PR description lists the missing/shifted hook points.

So the intent of this PR is to bridge that lifecycle gap by invoking the relevant hooks at the schedule-visible boundaries where the parameters are actually safe to gather/release.

I agree with the concern about avoiding more `if FSDP` branches in schedule/model code. I have not yet assessed the full effort needed to make this completely backend/generic-hook based, but I think the direction would be to expose/register these lifecycle hooks at the same fine-grained submodule boundaries as `enable_fine_grained_param_gather_hook`.

wujingyue: > EP overlap drives forward/backward at the submodule/op level through fine-grained callables

Understood. That's why I asked earlier whether we can let MFSDP operate directly on fine-grained “submodules”. It looks like that API [exists](https://github.com/NVIDIA/Megatron-LM/blob/5fe3f06658b393283eafc32f039c3d0c2468a69f/megatron/core/distributed/fsdp/mcore_fsdp_adapter.py#L98), and that [MoELayer](https://github.com/NVIDIA/Megatron-LM/blob/932d9eeee6b4ac42731001ae788f6339224a29e4/megatron/core/transformer/moe/moe_layer.py#L208) is well structured into submodules. I was mainly trying to understand other flexibility gaps for this particular PR. 

> after DGrad MLP, you shouldn't re-shard yet since you're gonna use those weights almost immediately for the FWD MLP.

Thanks — that’s an interesting capability for the rewrite. Adding a per-module preserve_unsharded knob shouldn’t be too difficult.

Wohox: Yes, but using `fsdp_unit_modules` to support EP overlap requires adding all modules with direct params into the list, which is a disaster for design, and also hooks registered on output tensor still won't be fired.

The ideal design would be: all hooks are reigstered at fine grained module level, if we have that, then most of the "if else" will be unnecessary. 

cc @shjwudp : wondering if this is possible for the refactoring?

Wohox:
/ok to test [e2a77b9](https://github.com/NVIDIA/Megatron-LM/pull/3797/commits/e2a77b984d3fcb16d542943e607399e15ef02ec6)

megatron/core/models/gpt/fine_grained_callables.py
```diff
@@ -350,6 +399,16 @@ def backward_dw(self):
 
         self.bwd_dw_callables = None
 
+    def set_post_forward_hook(self, hook):
+        """Register a hook to release FSDP params after this node's forward pass."""
```
jaredcasper: FSDP is still in docstring here... this should be generic.

megatron/core/models/gpt/fine_grained_callables.py
```diff
@@ -341,6 +360,36 @@ def backward_dw(self):
                 module.backward_dw()
             nvtx_range_pop(nvtx_msg)
 
+        # Setup delayed wgrad for Megatron FSDP by collecting gradient acc
+        # hooks if there is `.grad` attribute attached to param.
+        if self.fsdp_post_backward_gradient_hooks is None:
```
jaredcasper: Can these be part of the "post_backward_hook" called later? This all seems convoluted and very FSDP specific, can this be cleaned up? In general, where possible, for shared modules generic functionality should be added that will enable FSDP to use that generic functionality to achieve the things they need to do.

megatron/core/pipeline_parallel/combined_1f1b.py
jaredcasper: So the conclusion from this discussion isn't clear to me. Are we going to rework the fine grained hooks to support the FSDP use case?

jaredcasper (approved):

megatron/core/models/gpt/fine_grained_callables.py
cspades: Agreed, commenting so I can find this later.

Wohox:
/ok to test [1382745](https://github.com/NVIDIA/Megatron-LM/pull/3797/commits/1382745582efaed1654c29c90092e1b19675e93f)

PR #3810, opened 2026-03-11
[Megatron-FSDP] Support 'auto' argument which defaults to pre-MixedPrecisionPolicy be…

cspades:
…havior for supporting per-parameter grad dtypes.

# What does this PR do ?

- Support the `auto` (i.e. `None`) argument for Megatron-FSDP MixedPrecisionPolicy. Addresses the possible case where model gradients are mixed precision and static gradient buffers are used (as in the case of `--gradient-accumulation-fusion`), in which case we fall-back to the original logic of using the parameter data-type for the gradient buffer data-type, and always use BF16 for quantized parameters.

# Details

- The default memory utilization has changed to FP32 gradients instead of BF16 gradients. `grad_reduce_in_fp32 = False` does not set gradients to BF16, it just does not set gradients to FP32 and lets the new `--megatron-fsdp-*-dtype` arguments determine the precision, are which default to FP32 until this PR.
- `--gradient-accumulation-fusion` is not compatible with the default `torch.float32` value for gradient communication dtype when using FP8 parameters, because `get_main_grad()` is called during [Megatron-LM's backward implementation](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/tensor_parallel/layers.py#L556) and produces an FP32 gradient buffer for BF16 gradients if `--megatron-fsdp-grad-comm-dtype bf16` is not set. [Megatron-Bridge](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/models/common/unimodal.py#L240) currently does not support setting this argument, so we have to hard-code it to unblock MLPerf. This PR generalizes this compatibility in case models actually have mixed precision gradients, we need the `None` option in Megatron-LM.
- Mixed-precision gradients per-parameter or per-module isn't supported yet, if we hypothetically wanted to pre-allocate gradient buffers for fused gradient accumulation.
- While we recommend users use `torch.float32` for the main gradient data-type, it causes quite a lot of inflexibilities when using mixed-precision models, i.e. it is the fall-back data-type when the communication data-type is not specified (and customizing the communication data-type costs memory), and does not match the BF16 data-type for quantized parameters. Thus, we will instead default to the pre-mixed-precision behavior with `auto` for gradient data-type, which derives the data-type from the parameter's data-type or characteristics.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/training/arguments.py
cspades, line 873: Should I guard the args that don't support `None`? The argparser should already take care of it...

cspades:
Also configured in MBridge here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2784

cspades:
Almost forgot, updating documentation...

cspades:
`README.md` updated with my newest learnings of my own feature.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -93,7 +93,7 @@ def __init__(
             main_params_dtype=main_params_dtype,
             # Grandfathered Argument: grad_reduce_in_fp32
             main_grads_dtype=torch.float32 if ddp_config.grad_reduce_in_fp32 else main_grads_dtype,
-            grad_comm_dtype=grad_comm_dtype,
+            grad_comm_dtype=torch.float32 if ddp_config.grad_reduce_in_fp32 else grad_comm_dtype,
```
cspades, line 96: Enforces the legacy memory constraint - for HSDP, we need to reduce the main buffer, so adding a custom communication data-type here will allocate an extra communication buffer for the cast.

Turn off `grad_reduce_in_fp32` and the default is `None` for gradients, which is also legacy Megatron-FSDP behavior where there is no difference between main and communication buffers and they are equivalent to each other. (Many MBridge recipes do this already even though it's set to FP32 in `bf16_mixed`, which is a whole other can of worms to unpack...)

😵 on the argument parsing...

cspades: Will be happy to deprecate `grad_reduce_in_fp32` usage (as well as its primitives in `arguments.py`) specifically for Megatron-FSDP with a planned migration with MBridge.

PR #3812, opened 2026-03-11
DDP refactoring: Extract parameter layout computation into optimizer classmethod

deepakn94:
At a high level, this PR refactors `DistributedDataParallel` (DDP). DDP's constructor now accepts a `full_param_layout` argument that describes how parameters and gradients should be mapped in the underlying `_ParamAndGradBuffer`. Distributed optimizers compute this mapping via a static `compute_full_param_layout` method. This allows different optimizers to control gradient buffer layout — e.g., the existing `DistributedOptimizer` uses a sequential layout with padding to ensure buckets are multiples of the data-parallel size, and in the future, `LayerwiseOptimizer` for Muon could specify a mapping such that each gradient bucket shard contains a full layer's gradients. This ensures that: a) gradient reduction can use reduce-scatter instead of all-reduce (gradient and parameter sharding follows optimizer sharding exactly), and b) no extra parameter copies are needed after the parameter all-gather (results go directly into the right place for the next forward pass).

**Key changes:**
- New `param_layout.py` module with `BufferKey`, `PerBufferParamLayout`, and `FullParamLayout` dataclasses, plus shared padding utilities (`pad_param_start`, `pad_bucket_end`).
- `DistributedOptimizer.compute_full_param_layout()` and `_compute_per_buffer_param_layout()` static methods that pre-compute padded layouts.
- `DistributedDataParallel` accepts an optional `full_param_layout` and passes per-buffer layouts to `_ParamAndGradBuffer`.
- `_ParamAndGradBuffer` consumes pre-computed layouts when provided, falling back to `_compute_default_per_buffer_param_layout()` (no-padding layout for non-distributed-optimizer cases).
- `training.py` pre-computes layouts via `DistributedOptimizer.compute_full_param_layout()` when using the distributed optimizer with DDP.
- New unit tests in `test_param_layout.py` for layout computation functions.

In the future, `training.py` can assemble layouts from multiple distributed optimizers (this should be straightforward since each optimizer will have disjoint buffers).

### DIFF
--- a/megatron/core/distributed/distributed_data_parallel.py
+++ b/megatron/core/distributed/distributed_data_parallel.py
@@ -131,6 +131,14 @@ class DistributedDataParallel(_BaseDataParallel):
         # This maintains backward compatibility for callers that create DDP directly
         # without pre-computing layouts (e.g., tests, external code).
         if full_param_layout is None and self.ddp_config.use_distributed_optimizer:
+            log_single_rank(
+                logger,
+                logging.WARNING,
+                "DistributedDataParallel: full_param_layout not provided with "
+                "use_distributed_optimizer=True. Auto-computing layout inside DDP. "
+                "Callers should pre-compute layouts via "
+                "DistributedOptimizer.compute_full_param_layout() and pass them in.",
+            )
             from ..optimizer.distrib_optimizer import DistributedOptimizer
 
             full_param_layout = DistributedOptimizer.compute_full_param_layout(
--- a/megatron/core/distributed/distributed_data_parallel.py
+++ b/megatron/core/distributed/distributed_data_parallel.py
@@ -7,15 +7,15 @@ from typing import Optional
 import torch
 
 from ..config_logger import has_config_logger_enabled, log_config_to_disk
-from ..fp4_utils import is_nvfp4tensor
 from ..fp8_utils import is_float8tensor, post_all_gather_processing
+from ..optimizer.param_layout import FullParamLayout
 from ..process_groups_config import ProcessGroupCollection
 from ..transformer.cuda_graphs import is_graph_capturing
 from ..transformer.transformer_config import TransformerConfig
 from ..utils import log_single_rank
 from .data_parallel_base import _BaseDataParallel
 from .distributed_data_parallel_config import DistributedDataParallelConfig
-from .param_and_grad_buffer import _ParamAndGradBuffer, partition_buckets
+from .param_and_grad_buffer import _ParamAndGradBuffer, group_params_for_buffers, partition_buckets
 
 logger = logging.getLogger(__name__)
 
@@ -36,6 +36,9 @@ class DistributedDataParallel(_BaseDataParallel):
             use standard bucketing policy: assign parameters to smaller buckets and all-reduce
             per bucket _if_ overlap_grad_reduce is True and pp_rank is 0.
         pg_collection: Optional unified process group for distributed training.
+        full_param_layout: Optional FullParamLayout providing pre-computed layouts for all
+            dtype groups. When provided, each buffer uses the corresponding PerBufferParamLayout
+            instead of computing a default one.
 
     """
 
@@ -46,6 +49,7 @@ class DistributedDataParallel(_BaseDataParallel):
         module: torch.nn.Module,
         disable_bucketing: bool = False,
         pg_collection: Optional[ProcessGroupCollection] = None,
+        full_param_layout: Optional[FullParamLayout] = None,
     ):
         super().__init__(config=config, module=module)
         if has_config_logger_enabled(config):
@@ -104,11 +108,10 @@ class DistributedDataParallel(_BaseDataParallel):
 
         self.param_to_bucket_group = {}
 
-        # Group parameters by their gradient type.
+        # Collect all trainable parameters.
         param_to_name = {}
-        dense_params = []
-        expert_parallel_params = []
         self.params_with_grad = []
+        all_params = []
         for name, param in self.module.named_parameters():
             if not param.requires_grad:
                 continue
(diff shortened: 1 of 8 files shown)


megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -806,103 +889,15 @@ def _pad_end_of_bucket_if_needed(bucket_end_index: int) -> int:
             Pads end index of bucket if using distributed optimizer (to ensure uniform sharding).
             """
             if self.ddp_config.use_distributed_optimizer:
-                # Workaround for TE bug causing cuBLAS to pick an incompatible algorithm.
```
deepakn94, line 835: Don't delete this.

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -796,7 +868,18 @@ def __init__(
...
+            layout = optimizer_class.compute_param_layout(
+                params, bucket_size, self.data_parallel_world_size, ddp_config
+            )
+        else:
+            layout = _default_param_layout(params, bucket_size)
+        self.param_index_map = layout.param_index_map
+        self.bucket_indices = layout.bucket_indices
+        per_bucket_numel_unpadded = layout.per_bucket_numel_unpadded
 
         def _pad(number_to_be_padded: int, divisor: int) -> int:
```
deepakn94: Is this method used at all?

kvareddy (approved):
LGTM.

jiemingz (approved):

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -290,20 +194,111 @@ def _allocate_buffers_for_parameters(
...
                 expert_gradient_scaling_factor = 1.0 / data_parallel_world_size
 
-        # Allocate the param+grad buffers for dense params' grads.
-        self.buffers, self.bucket_groups = _allocate_buffers_for_parameters(
-            dense_params, self.intra_dp_cp_group, gradient_scaling_factor=gradient_scaling_factor
-        )
+        # Allocate buffers for each group.
+        self.buffers = []
+        self.expert_parallel_buffers = []
+        all_buffers = []
```
deepakn94: Why do I need this variable?

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -290,20 +194,111 @@ def _allocate_buffers_for_parameters(
...
+        pg_collection = ProcessGroupCollection()
```
skyw: Suggestion: Favor setting groups at contructing time, e.g. ProcessGroupCollection(tp=xxx, ...)

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -758,6 +758,116 @@ def register_grad_ready(
...
 
+from ..optimizer.param_layout import BufferKey, PerBufferParamLayout
+
+
+def _compute_default_per_buffer_param_layout(
+    params: List[torch.nn.Parameter], bucket_size: Optional[int]
+) -> PerBufferParamLayout:
+    """Compute parameter layout for the non-distributed-optimizer case.
+
+    No padding is applied. Parameters are iterated in reverse order (backprop order)
```
skyw, line 819: Nit: explain why iterating in reversed order.

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -758,6 +758,116 @@ def register_grad_ready(
...
+        this_numel = param.data.nelement()
+        param_end_index = param_start_index + this_numel
+        param_index_map[param] = (param_start_index, param_end_index, bucket_id)
+        bucket_params.add(param)
+
+        if bucket_size is not None and (param_end_index - bucket_start_index) >= bucket_size:
+            per_bucket_numel_unpadded.append(param_end_index - bucket_start_index)
+            bucket_indices.append((bucket_start_index, param_end_index))
+            bucket_start_index = param_end_index
+            bucket_params = set()
```
skyw, line 850: Confusing code. Not clear why it is needed.
If param shouldn't be added when this condition met, bucket_params.add(param) should be moved under this condition.

megatron/core/optimizer/param_layout.py
```diff
@@ -0,0 +1,70 @@
...
+    Attributes:
+        param_index_map: Mapping from parameter to (start_index, end_index, bucket_id) in buffer.
+        bucket_indices: List of (start_index, end_index) for each bucket.
+        per_bucket_numel_unpadded: Number of unpadded elements per bucket.
+        param_indices: The index of each param among same-dtype params (using the "fake"
+            high-precision dtype for FP8 params). Needed for loading non-native-fp8
+            checkpoints in native-fp8 mode. Order matches param_index_map iteration order.
+    """
+
+    param_index_map: Dict[torch.nn.Parameter, Tuple[int, int, int]] = field(default_factory=dict)
```
skyw, line 74: Q: Is there any performance concern with large object as dict key? the time spend on hashing.

deepakn94: PyTorch just looks at the memory location of the underlying weights, so should be cheap (`O(1)`).

deepakn94:
Note to self: move the assertion on [this line](https://github.com/NVIDIA/Megatron-LM/pull/4328/changes#diff-c4ada2b3f7563ba8629e9bb5617ade92aa63194d8490e1fbc17642b26dc66f74R1058) outside the `if self.ddp_config.use_distributed_optimizer` if condition.

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -470,6 +475,134 @@ def _build_model_and_main_param_groups(
...
+            params: List of parameters to lay out.
+            bucket_size: Approximate number of elements per bucket, or None for single bucket.
+            data_parallel_world_size: Size of the data-parallel group.
+            ddp_config: DistributedDataParallel config object.
+
+        Returns:
+            PerBufferParamLayout with the computed mapping.
+        """
+
+        def _pad_start_of_param(param_start_index: int) -> int:
```
deepakn94: Why do I need these helper methods if they're just calling the imported methods?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -753,6 +754,116 @@ def register_grad_ready(
...
+            bucket_id += 1
+            param_start_index = param_end_index
```
deepakn94: Duplicated across if and else blocks. Move out of if-else to de-duplicate.

yaoyu-33 (approved):

PR #3822, opened 2026-03-12
feat(fsdp): use TE general_gemm for mixed-precision wgrad in FSDP path

Victarry:
## Summary

- Replace `torch.matmul` with TE `general_gemm` in the FSDP gradient accumulation fusion path to support mixed-precision output (e.g. bf16 input → fp32 `main_grad` buffer)
- `torch.matmul` requires the `out` tensor dtype to match the compute dtype, which forces FSDP to use `grad_reduce_in_bf16` when `gradient_accumulation_fusion` is enabled
- TE's `general_gemm` supports arbitrary `out_dtype` via cuBLAS, removing this hard dependency
- Uses inline lazy import to avoid circular dependency with `megatron.core.extensions.transformer_engine`

## Test plan

- [x] Verified on DeepSeek-V3-proxy (MLA+MoE+MTP) with FSDP (`fsdp_double_buffer=True`, `gradient_accumulation_fusion=True`, fp32 grad buffer, TP1/PP1/EP4/DP2, 8×H100)
- [x] Training converges normally (~103 TFLOP/s/GPU, lm_loss decreases from 10.35 to 0.30)


cspades (approved):
Yup, this will mean that users can use `(megatron_fsdp_)main_grads_dtype=torch.float32` and `(megatron_fsdp_)grad_comm_dtype=torch.float32` even with FP8 params / BF16 gradients, thanks to `out_dtype`, thanks @Victarry !

cspades:
Rebasing on main to hopefully get rid of the CodeCov failure.

PR #3824, opened 2026-03-12
[Dev] feat(fsdp): use TE general_gemm for mixed-precision wgrad in FSDP path

Victarry:
## Summary

PR to main https://github.com/NVIDIA/Megatron-LM/pull/3822
- Replace `torch.matmul` with TE `general_gemm` in the FSDP gradient accumulation fusion path to support mixed-precision output (e.g. bf16 input → fp32 `main_grad` buffer)
- `torch.matmul` requires the `out` tensor dtype to match the compute dtype, which forces FSDP to use `grad_reduce_in_bf16` when `gradient_accumulation_fusion` is enabled
- TE's `general_gemm` supports arbitrary `out_dtype` via cuBLAS, removing this hard dependency
- Uses inline lazy import to avoid circular dependency with `megatron.core.extensions.transformer_engine`

## Test plan

- [x] Verified on DeepSeek-V3-proxy (MLA+MoE+MTP) with FSDP (`fsdp_double_buffer=True`, `gradient_accumulation_fusion=True`, fp32 grad buffer, TP1/PP1/EP4/DP2, 8×H100)
- [ ] Unit tests

### DIFF
--- a/megatron/core/tensor_parallel/layers.py
+++ b/megatron/core/tensor_parallel/layers.py
@@ -561,7 +561,23 @@ class LinearWithGradAccumulationAndAsyncCommunication(torch.autograd.Function):
                 # In case of Megatron-FSDP, need to create main grad buffers in-place
                 if hasattr(weight, "__fsdp_param__"):
                     weight.main_grad = weight.get_main_grad()
-                    torch.matmul(grad_output.t(), total_input, out=weight.main_grad)
+                    # Import here to avoid circular import
+                    from megatron.core.extensions.transformer_engine import te_general_gemm
+
+                    if te_general_gemm is not None:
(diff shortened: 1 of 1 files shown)


yaox12:
> torch.matmul requires the out tensor dtype to match the compute dtype, which forces FSDP to use grad_reduce_in_bf16 when gradient_accumulation_fusion is enabled

Was there a check or assertion before? If yes, we should remove it.

Victarry:
> > torch.matmul requires the out tensor dtype to match the compute dtype, which forces FSDP to use grad_reduce_in_bf16 when gradient_accumulation_fusion is enabled
> 
> Was there a check or assertion before? If yes, we should remove it.

No check before, just raise error if you hit it. 🤣

PR #3754, opened 2026-03-09
fix no_shard training convergency and add unittest for no_shard

wplf:
# What does this PR do ?
Fixes no_shard convergence by correcting grad_norm and all_reduce logic.

1.  fix no_shard grad_norm calculation 
2. fix no_shard all_reduce grad continue bug
3. Make no_shard to be compatiable without overlap_param_gather and overlap_grad_reduce
4. add  unittest for no_shard
5. add assert False for no_shard and `init_model_with_meta_device` follow [fully_shard.py:326](https://github.com/NVIDIA/Megatron-LM/blob/597721a17891669c6ebcb098beb74cd83087205d/megatron/core/distributed/fsdp/src/megatron_fsdp/fully_shard.py#L326)

Grad_norm and loss convergency and unittest result is below.

image
image
image

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

### Pre-checks

- [x] I have added relevant unit tests
- [x] I have added relevant functional tests
- [x] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [x] I have added relevant documentation
- [x] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
6. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

Once all required reviewers have approved, the `Approved` label is applied **automatically**.

### Merge

[...]

PR #3835, opened 2026-03-12
[dev] fix no_shard training convergency and add unittest for no_shard

wplf:
# What does this PR do ?
**main PR** https://github.com/NVIDIA/Megatron-LM/pull/3754

- Fixes no_shard convergence by correcting grad_norm and all_reduce logic.
- fix no_shard grad_norm calculation
- fix no_shard all_reduce grad continue bug
- Make no_shard to be compatiable without overlap_param_gather and overlap_grad_reduce
- add unittest for no_shard
- add assert False for no_shard and init_model_with_meta_device follow [fully_shard.py:326]

### DIFF
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -675,7 +675,18 @@ def _get_megatron_optimizer_based_on_param_groups(
             # This is needed for case where num_distributed_optimizer_instances > 1. In this case,
             # weight gradients are all-reduced across optimizer instances, so each instance has
             # the duplicated weight gradients, need to reduce gradient stats inside each instance.
-            setattr(optimizer, 'grad_stats_parallel_group', intra_dist_opt_group)
+            # Besides, for Megatron-FSDP with no_shard, gradients are replicated across DP ranks (after
+            # all-reduce), so grad stats should only be reduced across model-parallel ranks
+            # (TP*PP) to avoid inflating the grad norm by sqrt(DP_world_size).
+            ddp_config = getattr(model_chunks[0], 'ddp_config', None)
+            if (
+                ddp_config is not None
+                and getattr(ddp_config, 'use_megatron_fsdp', False)
+                and ddp_config.data_parallel_sharding_strategy == 'no_shard'
+            ):
+                setattr(optimizer, 'grad_stats_parallel_group', model_parallel_group)
+            else:
+                setattr(optimizer, 'grad_stats_parallel_group', intra_dist_opt_group)
         else:
             optimizer = Float16OptimizerWithFloat16Params(*optimizer_args)
             setattr(optimizer, 'grad_stats_parallel_group', model_parallel_group)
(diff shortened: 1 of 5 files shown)


megatron/core/optimizer/__init__.py
```diff
@@ -473,7 +473,18 @@ def init_state_fn(opt, config=None):
...
             # weight gradients are all-reduced across optimizer instances, so each instance has
             # the duplicated weight gradients, need to reduce gradient stats inside each instance.
-            setattr(optimizer, 'grad_stats_parallel_group', intra_dist_opt_group)
+            # Besides, for Megatron-FSDP with no_shard, gradients are replicated across DP ranks (after
+            # all-reduce), so grad stats should only be reduced across model-parallel ranks
+            # (TP*PP) to avoid inflating the grad norm by sqrt(DP_world_size).
+            ddp_config = getattr(model_chunks[0], 'ddp_config', None)
+            if (
+                ddp_config is not None
+                and getattr(ddp_config, 'use_megatron_fsdp', False)
```
FDecaYed: use_megatron_fsdp is already used to guard calling of this function. so I feel a better place to do this is when use_megatron_fsdp is used, check no_shard and pass in intra_dist_opt_group=model_parallel_group from caller side. It is less duplicated checking and less change to shared code path

something like
```
     ddp_config = model_chunks[0].ddp_config  # Use the first model chunk's DDP config
     if ddp_config.use_megatron_fsdp:
+        # For no_shard, gradients are replicated across DP ranks after all-reduce, so grad stats
+        # should only be reduced over TP/PP (model_parallel_group) to avoid inflating the norm.
+        effective_intra_dist_opt_group = (
+            mp_group
+            if ddp_config.data_parallel_sharding_strategy == 'no_shard'
+            else intra_dist_opt_group
+        )
         for model_chunk, overlap_param_gather_with_optimizer_step in zip(
             all_dense_model_chunks, overlap_param_gather_with_optimizer_step_flags
         ):
@@ -900,7 +907,7 @@ def get_megatron_optimizer(
                     data_parallel_group=dp_cp_group,
                     data_parallel_group_gloo=intra_dp_cp_group_gloo,
                     data_parallel_group_idx=model_parallel_rank,
-                    intra_dist_opt_group=intra_dist_opt_group,
+                    intra_dist_opt_group=effective_intra_dist_opt_group,
                     distributed_optimizer_instance_id=distributed_optimizer_instance_id,
                     pg_collection=pg_collection,
                 )
```

FDecaYed (dismissed):
left comments

megatron/core/optimizer/__init__.py
wplf: This is indeed better.  Thank you very much for reviewing.

I'll fix this tonight.

wplf: Done

shjwudp (approved):

wplf:
/ok to test [c4a4463](https://github.com/NVIDIA/Megatron-LM/pull/3835/commits/c4a446351e6b1879895fc4a88645c154aa843c21)

Victarry (approved):

cspades (approved):

PR #3842, opened 2026-03-12
Add activation logging and tokens per expert logging

Mellonta:
Adding activation logging and tokens per expert logging,
activation logging can be enabled through --save-activations-interval,
tokens per expert logging can be enabled through --save-tokens-per-experts-interval,
when enabled, they will be dumped at checkpoint folder. tokens per expert will be logged in args.save/tokens_per_expert/rank3.jsonl.

### DIFF
--- a/megatron/training/activation_logging.py
+++ b/megatron/training/activation_logging.py
@@ -1,6 +1,6 @@
 # Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
 
-"""dgrad logging using backward hooks."""
+"""Forward activation logging using forward hooks."""
 
 from collections import defaultdict
 import torch
@@ -14,7 +14,7 @@ from .utils import unwrap_model
 
 
 def _get_linear_types():
-    """Build tuple of linear layer types to capture gradients from."""
+    """Build tuple of linear layer types to capture activations from."""
     types = [nn.Linear, nn.Embedding, ColumnParallelLinear, RowParallelLinear, Router]
 
     # Add Transformer Engine layers if available.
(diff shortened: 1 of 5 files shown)


megatron/training/activation_logging.py
```diff
@@ -0,0 +1,245 @@
...
+    # ------------------------------------------------------------------
+
+    def _make_tpe_hook(self, _model_chunk_name: str, module_name: str) -> Callable:
+        """Forward hook that captures only the non-Tensor ``input1`` (tokens_per_expert).
+
+        The layer number is extracted from *module_name*
+        (e.g. ``decoder.layers.3.mlp.experts.linear_fc1`` → ``3``).
+        """
+        parts = module_name.split(".")
+        layer = parts[2] if len(parts) > 2 else module_name
```
deepakn94: Extracting `layer` using a regex might be a bit more failsafe.

megatron/training/activation_logging.py
```diff
@@ -0,0 +1,245 @@
...
+
+        def hook(_, args, kwargs, output):
+            input_tuple = args if isinstance(args, tuple) else (args,)
+            if len(input_tuple) > 1 and input_tuple[1] is not None:
+                inp = input_tuple[1]
+                if not isinstance(inp, torch.Tensor):
+                    self._tpe_records.append((layer, inp))
```
deepakn94: Should we make this a dictionary? `inp` will be a tuple with length equal to number of experts on this GPU?

megatron/training/activation_logging.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def save_tpe(self, iteration: int):
+        """Append captured tokens-per-expert records to a shared plain text file.
+        All ranks append to a single file ``{save_dir}/tokens_per_expert.log``.
+        Lines are formatted as::
+            Rank={rank} iter={iter} layer={layer} tokens_per_expert={list}
+        """
+        if not self._tpe_records:
+            return
+        rank = torch.distributed.get_rank() if torch.distributed.is_initialized() else 0
```
deepakn94, line 207: Should we try doing the aggregation / accumulation here instead of doing it posthoc?

megatron/training/activation_logging.py
```diff
@@ -0,0 +1,245 @@
...
+        if not self._activations_state_dict:
+            return
+        save_grads(self._save_dir, self._activations_state_dict, iteration, "activations")
+        self._activations_state_dict.clear()
+
+    # ------------------------------------------------------------------
+    # Tokens-per-expert hooks
+    # ------------------------------------------------------------------
+
+    def _make_tpe_hook(self, _model_chunk_name: str, module_name: str) -> Callable:
```
deepakn94, line 161: Would adding this logic to the regular activation hook make things more complicated?

megatron/training/activation_logging.py
Mellonta: We only dump the results once per iteration, which may span multiple microbatches, so using a list lets us capture all of them.

megatron/training/activation_logging.py
Mellonta: Do you mean aggregate across ranks or across microbatches? For ranks, I wanted to avoid introducing explicit comm in the training loop as it could consume SMs or interfere with overlapping. For microbatches, I need finer granularity here to analyze load balancing within each microbatch.

megatron/training/activation_logging.py
Mellonta: It may become complicated when we have both activation and tokens per expert logging enabled but with different intervals.

deepakn94: Good point, let's keep them separate!

megatron/training/activation_logging.py
deepakn94: I think it's a good idea still to make `self._tpe_records` a dictionary, but agreed that the value of the dictionary should be a list of tuples.

megatron/training/activation_logging.py
deepakn94: Good points.

megatron/training/activation_logging.py
```diff
@@ -0,0 +1,245 @@
...
+        All ranks append to a single file ``{save_dir}/tokens_per_expert.log``.
+        Lines are formatted as::
+            Rank={rank} iter={iter} layer={layer} tokens_per_expert={list}
+        """
+        if not self._tpe_records:
+            return
+        rank = torch.distributed.get_rank() if torch.distributed.is_initialized() else 0
+        os.makedirs(self._save_dir, exist_ok=True)
+        filepath = os.path.join(self._save_dir, "tokens_per_expert.log")
+        lines = "".join(
```
deepakn94, line 211: I think it's good if we make this file format easily machine-ingestible too. Maybe CSV is better (can directly load this into a Pandas dataframe for example)?

megatron/training/activation_logging.py
```diff
@@ -0,0 +1,245 @@
...
+            f"Rank={rank} iter={iteration} layer={layer} tokens_per_expert={inp}\n"
+            for layer, inp in self._tpe_records
+        )
+        with open(filepath, "a") as f:
```
deepakn94, line 216: Multiple processes are trying to write here, right? So does this need to be protected by a locking mechanism?

Mellonta: It's an O_APPEND so it should be atomic. I'm not so sure about lustre, but I haven't seen any interleaving in my test runs so far.

deepakn94 (approved):

yaoyu-33 (approved):

deepakn94:
/ok to test [f36c030](https://github.com/NVIDIA/Megatron-LM/pull/3842/commits/f36c030d8380633e97057c82c84ad686c8aa9cfa)

deepakn94:
/ok to test [ccf1f02](https://github.com/NVIDIA/Megatron-LM/pull/3842/commits/ccf1f028669497b3c9b5f6a6b8007011d2d27d33)

PR #3890, opened 2026-03-16
Support GEMM + Swiglu fused MLP

ksivaman:
# What does this PR do ?

This PR supports GEMM + Swiglu fused MLP via Transformer Engine sequential ops.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/core/transformer/moe/experts.py
+++ b/megatron/core/transformer/moe/experts.py
@@ -97,30 +97,12 @@ class GroupedMLP(MegatronModule):
         ), "MoE latent projection not supported in GroupedMLP yet."
 
         self.expert_parallel = config.expert_model_parallel_size > 1
-        glu_interleave_size = self.config.moe_mlp_glu_interleave_size
         if self.config.gated_linear_unit:
             if self.config.activation_func not in (F.silu, F.gelu):
                 raise ValueError("Activation function must be silu or gelu when using GroupedMLP.")
-            if glu_interleave_size is not None and (
-                not isinstance(glu_interleave_size, int) or glu_interleave_size <= 0
-            ):
-                raise ValueError(
-                    "`moe_mlp_glu_interleave_size` must be a positive integer or None, "
-                    f"got {glu_interleave_size!r}."
-                )
 
             @jit_fuser
             def glu(x):
-                # Undo interleaving if needed
-                if glu_interleave_size is not None:
-                    shape = x.size()
-                    x = x.reshape(
-                        -1, shape[-1] // (2 * glu_interleave_size), 2, glu_interleave_size
-                    )
-                    x = x.transpose(1, 2)
-                    x = x.reshape(shape)
-
-                # Compute gated activation
                 x = torch.chunk(x, 2, dim=-1)
                 return self.config.activation_func(x[0]) * x[1]
 
@@ -152,19 +134,6 @@ class GroupedMLP(MegatronModule):
         # How many feature each rank holds for fc1 and fc2, respectively.
         tp_size = self.tp_group.size()
         tp_rank = self.tp_group.rank()
-        if (
-            self.config.gated_linear_unit
-            and glu_interleave_size is not None
-            and self.config.moe_ffn_hidden_size % (tp_size * glu_interleave_size) != 0
-        ):
-            raise ValueError(
-                "Unsupported `moe_mlp_glu_interleave_size`: expected "
-                "`moe_ffn_hidden_size` to be divisible by "
-                "`expert_tensor_parallel_size * moe_mlp_glu_interleave_size`. "
-                f"Got moe_ffn_hidden_size={self.config.moe_ffn_hidden_size}, "
-                f"expert_tensor_parallel_size={tp_size}, "
-                f"moe_mlp_glu_interleave_size={glu_interleave_size}."
-            )
 
         fc1_output_size = self.config.moe_ffn_hidden_size * self.num_local_experts
         if config.gated_linear_unit:


megatron/core/transformer/moe/experts.py
```diff
@@ -1,12 +1,14 @@
 # Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 from __future__ import annotations
 
+from collections.abc import Callable
```
QiZhangNV: Duplicate import. Please remove one.

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -1286,9 +1286,11 @@ def get_align_size_for_quantization(config: TransformerConfig) -> int:
         int: The alignment size for quantization.
     """
     if config.fp8:
-        return get_fp8_align_size(config.fp8_recipe)
+        # CUTLASS kernel for MXFP8 grouped GEMM assumes 256 alignment
+        return 256
```
QiZhangNV: Hardcoding alignment to 256 here affects all callers , not just the fused grouped GEMM path.  Please keep the original recipe-based alignment as default and only apply 256 when config.use_transformer_engine_op_fuser is enabled.

megatron/core/transformer/moe/experts.py
```diff
@@ -737,6 +756,191 @@ def _apply_bias(intermediate_parallel, bias_parallel, tokens_per_expert, permute
...
+        # Apply padding if needed
+        unpadded_tokens_per_expert = None
+        padding_align_size = None
+        if self.config.moe_router_padding_for_quantization:
+            # Padding has already been applied in router
+            pass
+        elif self.config.fp8 or self.config.fp4:
+            tokens_per_expert = tokens_per_expert.tolist()
+            unpadded_tokens_per_expert = tokens_per_expert
+            padding_align_size = 256  # Requirement for grouped GEMM
```
QiZhangNV: Since Fp8Padding.__init__ already accepts align_size, it's better to set it at construction (line 732) instead:

```
align_size = 256 if self._with_fused_impl else None
self.quantization_padding = Fp8Padding(self.num_local_experts, align_size=align_size)
self.quantization_unpadding = Fp8Unpadding(self.num_local_experts, align_size=align_size)
```

megatron/core/transformer/moe/experts.py
```diff
@@ -737,6 +756,191 @@ def _apply_bias(intermediate_parallel, bias_parallel, tokens_per_expert, permute
...
+
+
+        # Copy the weights from GroupedLinear module to GroupedLinear op.
+        if fc2_single_grouped_parameter:
+            setattr(op, "weight", getattr(self.linear_fc2, "weight"))
+
+        for idx in range(self.linear_fc2.num_gemms):
+            if not fc2_single_grouped_parameter:
+                setattr(op, f"weight{idx}", getattr(self.linear_fc2, f"weight{idx}"))
+            if self.linear_fc1.use_bias:
```
QiZhangNV:
```suggestion
            if self.linear_fc2.use_bias:
```

megatron/core/transformer/moe/experts.py
```diff
@@ -737,6 +756,191 @@ def _apply_bias(intermediate_parallel, bias_parallel, tokens_per_expert, permute
             .to(intermediate_parallel.dtype)
         )
 
+    def _is_fused_impl_supported(self) -> bool:
```
QiZhangNV, line 763: `use_transformer_engine_op_fuser`  is defined but never checked. The fused path activates automatically whenever conditions are met. Please add:

```
if not self.config.use_transformer_engine_op_fuser:
    return False
```

megatron/core/transformer/transformer_config.py
```diff
@@ -761,6 +765,15 @@ class TransformerConfig(ModelParallelConfig):
     """Number of SMs to use for HybridEP. In pure NVL scenarios,
     16 SMs can generally achieve good bandwidth."""
 
+    moe_mlp_glu_interleave_size: Optional[int] = None
```
QiZhangNV, line 768: This value is not validated. Please add an assert to reject invalid values in `TEGroupedMLP`.

megatron/core/transformer/moe/experts.py
```diff
@@ -790,15 +1009,38 @@ def forward(
...
 
         def bias_act_func(intermediate_parallel, bias_parallel, permuted_probs):
+
+            # Whether activation function is interleaved GLU
+            with_glu_interleaving = (
+                self.config.gated_linear_unit
+                and self.config.moe_mlp_glu_interleave_size is not None
+            )
+
+            def remove_glu_interleaving(x: torch.Tensor) -> torch.Tensor:
```
QiZhangNV, line 1022: `remove_glu_interleaving`  runs a `contiguous()` copy every forward pass in the non-fused path. If the interleaved format is designed for the fused kernel, using it without the fused path is likely overhead with no benefit. Could we consider warning when moe_mlp_glu_interleave_size is set but _is_fused_impl_supported() returns False?

megatron/core/transformer/moe/experts.py
```diff
@@ -737,6 +756,191 @@ def _apply_bias(intermediate_parallel, bias_parallel, tokens_per_expert, permute
...
+        # Note: The fused impl is stored in a tuple to avoid
+        # registering submodules.
+        if self._fused_ops is None:
+            self._fused_ops = (self._make_fused_ops(),)
+        ops, = self._fused_ops
+
+        # Apply padding if needed
+        unpadded_tokens_per_expert = None
+        padding_align_size = None
+        if self.config.moe_router_padding_for_quantization:
```
QiZhangNV, line 918: A quick question: for sync-free / full-iteration CUDA graph scenarios, is `moe_router_padding_for_quantization=True `always expected? In other words, the elif fallback exists because the fused ops are device-initiated, so padding must go through `D2H -> pad -> H2D `when not pre-padded in the router, making it not sync-free / CUDA-graphable?

megatron/core/transformer/transformer_config.py
```diff
@@ -395,6 +395,10 @@ class TransformerConfig(ModelParallelConfig):
     fused_single_qkv_rope: bool = False
     """If set, avoid splitting QKV before ROPE forward and avoid concatenating ROPE dgrads."""
 
+    use_transformer_engine_op_fuser: bool = False
```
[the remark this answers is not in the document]
ksivaman: Done

megatron/core/transformer/moe/experts.py
```diff
@@ -737,6 +756,191 @@ def _apply_bias(intermediate_parallel, bias_parallel, tokens_per_expert, permute
...
+            permuted_local_hidden_states,
+            tokens_per_expert,  # FC1
+            permuted_probs,  # Scaled SwiGLU
+            tokens_per_expert,  # FC2
+        )
+
+        # Remove padding if needed
+        if unpadded_tokens_per_expert is not None:
+            self.quantization_unpadding.align_size = padding_align_size
+            output = self.quantization_unpadding(output, unpadded_tokens_per_expert)
```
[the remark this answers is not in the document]
ksivaman: Addressed in above comment from @QiZhangNV

megatron/core/transformer/moe/experts.py
```diff
@@ -1,12 +1,14 @@
 # Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 from __future__ import annotations
 
+from collections.abc import Callable
 import copy
 import logging
 from collections.abc import Callable
```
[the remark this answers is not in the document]
ksivaman: Addressed

megatron/core/transformer/moe/experts.py
ksivaman: Good point, I have kept the function the same but instead added this check in the init as that made more sense

QiZhangNV: Thanks, looks good! A suggestion: since the assert above already guards _is_fused_impl_supported(), maybe line 753 could be simplified to self._with_fused_impl = self.config.use_transformer_engine_op_fuser?

megatron/core/transformer/moe/experts.py
ksivaman: Yes, that's the expect path

megatron/core/transformer/moe/experts.py
```diff
@@ -712,10 +747,24 @@ def __init__(
 
             set_save_original_input(self.linear_fc1)
 
+        # Fused implementation with Transformer Engine op fuser API
+        if self.config.use_transformer_engine_op_fuser:
+            assert self._is_fused_impl_supported(), "Fused GroupedMLP is not supported for this configuration."
+        self._with_fused_impl: bool = self._is_fused_impl_supported() and self.config.use_transformer_engine_op_fuser
```
ksivaman:
```suggestion
        self._with_fused_impl: bool = self.config.use_transformer_engine_op_fuser
```

megatron/core/transformer/moe/experts.py
ksivaman: Done

yaox12:
Can you add unit tests for the numerics of the fusion, and the remapping of parameter keys?

ksivaman:
@yaox12 The unit test for both the checkpoint loading as well as specifically the fusion numerics are included in https://github.com/NVIDIA/TransformerEngine/pull/2769. I think adding them separately would be a duplicate. Unless you mean an e2e test?

megatron/core/transformer/moe/experts.py
```diff
@@ -133,6 +152,19 @@ def activation_func_with_probs(x, probs):
         # How many feature each rank holds for fc1 and fc2, respectively.
         tp_size = self.tp_group.size()
         tp_rank = self.tp_group.rank()
+        if (
```
zhongbozhu: Why do we need to edit class GroupedMLP when we are actually using TEGroupedMLP?

megatron/core/transformer/moe/experts.py
```diff
@@ -712,10 +744,32 @@ def __init__(
...
+                "tensor reordering/copy overhead each forward pass.",
+                self.config.moe_mlp_glu_interleave_size,
+            )
+
         if self.config.fp8 or self.config.fp4:
             assert HAVE_TE, "FP8 and FP4 requires TE."
-            self.quantization_padding = Fp8Padding(self.num_local_experts)
-            self.quantization_unpadding = Fp8Unpadding(self.num_local_experts)
+            align_size = 256 if self._with_fused_impl else None
+            self.quantization_padding = Fp8Padding(self.num_local_experts, align_size=align_size)
```
zhongbozhu, line 738: So the Fp8Padding module itself is not graph safe, when we are running full CG, this module should be disabled and let permute kernel do the padding, will this be address in a followup PR?

megatron/core/transformer/moe/experts.py
```diff
@@ -737,6 +791,192 @@ def _apply_bias(intermediate_parallel, bias_parallel, tokens_per_expert, permute
...
+        return True
+
+    def _make_fused_ops(self) -> torch.nn.Module:
+        """Construct fused module for FC1, activation, and FC2."""
+
+        # Container for fusible ops
+        ops = te.pytorch.ops.Sequential()
+
+        # Check if there are 1 or "num_gemms" params in the GroupedLinear module.
+        fc1_single_grouped_parameter = self.linear_fc1.single_grouped_parameter
```
zhongbozhu, line 803: Is single grouped weight still being controlled by a TE env var?

megatron/core/transformer/moe/experts.py
yaox12: You're right. In fact the whole `GroupedMLP` class has been removed on the main branch, which has not been synced to the dev branch yet.

megatron/core/transformer/moe/experts.py
```diff
@@ -737,6 +791,192 @@ def _apply_bias(intermediate_parallel, bias_parallel, tokens_per_expert, permute
...
+        # the params are modified after the constructor.
+        # Note: The fused impl is stored in a tuple to avoid
+        # registering submodules.
+        if self._fused_ops is None:
+            self._fused_ops = (self._make_fused_ops(),)
+        (ops,) = self._fused_ops
+
+        # Apply padding if needed
+        unpadded_tokens_per_expert = None
+        if self.config.moe_router_padding_for_quantization:
```
zhongbozhu, line 918: both router padding & explicit padding are not graph safe, only padding with permute kernels are graph safe, will this be addressed in the followup PR?

megatron/core/transformer/moe/experts.py
```diff
@@ -737,6 +791,192 @@ def _apply_bias(intermediate_parallel, bias_parallel, tokens_per_expert, permute
...
+            # Padding has already been applied in router
+            pass
+        elif self.config.fp8 or self.config.fp4:
+            tokens_per_expert = tokens_per_expert.tolist()
```
zhongbozhu, line 922: Same^, this will be fixed in the next PR?

PR #3781, opened 2026-03-10
Fix PersistentAsyncCaller.__del__ crash during Python shutdown

cluster2600:
## Summary

Fixes #3775.

During interpreter shutdown or garbage collection the distributed process group
is already destroyed by the time `PersistentAsyncCaller.__del__` invokes
`close()`.  The unconditional `torch.distributed.get_rank()` call inside
`close()` then raises `ValueError`, producing a noisy traceback on every rank
at the end of every training run when async checkpointing is enabled.

### What changed

- Introduced a module-level `_get_rank_or_unknown()` helper that checks
  `torch.distributed.is_initialized()` before calling `get_rank()` and wraps
  everything in a broad `try/except` so that even if `is_initialized()` itself
  fails during late teardown the helper returns `"?"` instead of propagating.
- Replaced all bare `get_rank()` calls inside the `close()` methods of
  `AsyncCaller`, `TemporalAsyncCaller`, and `PersistentAsyncCaller` with the
  safe helper.  This covers both the specific crash site recieved in the issue
  traceback and the same pattern in the other two callers.
- Added unit tests (`test_async_utils_shutdown.py`) that verify the helper's
  behaviour when dist is initialised, when it is not, and when
  `is_initialized()` itself raises.  Also verifies that `close()` / `__del__`
  on both `PersistentAsyncCaller` and `TemporalAsyncCaller` do not raise when
  the process group is unavailable.

### Why a helper rather than inline guards

Multiple `close()` methods across the class hierarchy had the same unguarded
pattern.  Centralising teh logic in one helper keeps the call sites readable
and makes it harder to reintroduce the bug when new log statements are added.

## Test plan

- [x] New unit tests in `tests/unit_tests/dist_checkpointing/test_async_utils_shutdown.py`
- [ ] Existing `test_async_save.py` suite still passes (async save + abort paths)
- [ ] Manual verification: run a training job with `async_save = True`, confirm
      clean shutdown with no `ValueError` traceback on any rank

PR #3892, opened 2026-03-16
Fix PersistentAsyncCaller.__del__ crash during shutdown

### LINKED ISSUE #3775: PersistentAsyncCaller.__del__ crashes during shutdown: get_rank() after process group destroyed
## Bug Description

`PersistentAsyncCaller.__del__()` in `megatron/core/dist_checkpointing/strategies/async_utils.py` calls `torch.distributed.get_rank()` in its `close()` method (line 429), but during Python shutdown/garbage collection the distributed process group has already been destroyed, causing a `ValueError`.

This happens on every rank at the end of every training run when async checkpointing is enabled.

## Traceback

```
Exception ignored in: <function PersistentAsyncCaller.__del__ at 0x746f0cbc8d60>
  File ".../megatron/core/dist_checkpointing/strategies/async_utils.py", line 446, in __del__
    self.close()
  File ".../megatron/core/dist_checkpointing/strategies/async_utils.py", line 429, in close
    f"PersistentAsyncCaller: {torch.distributed.get_rank()}, Destroying Async Caller"
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../torch/distributed/distributed_c10d.py", line 2354, in get_rank
    default_pg = _get_default_group()
  File ".../torch/distributed/distributed_c10d.py", line 1306, in _get_default_group
    raise ValueError(
ValueError: Default process group has not been initialized, please make sure to call init_process_group.
```

## Root Cause

`async_utils.py:429` unconditionally calls `torch.distributed.get_rank()` inside a debug log message in `close()`. When `close()` is called from `__del__` during garbage collection, the process group is already torn down.

## Suggested Fix

Guard the `get_rank()` call in `close()`:

```python
def close(self):
    if torch.distributed.is_initialized():
        logger.debug(
            f"PersistentAsyncCaller: {torch.distributed.get_rank()}, Destroying Async Caller"
        )
    # ... rest of cleanup
```

## Environment

- Megatron-Core (via Megatron-Bridge 0.3.0, NGC nemo:base-26.02 image)
- PyTorch 2.x
- Python 3.12
- Async checkpoint saving enabled (`config.checkpoint.async_save = True`)

cluster2600:
## Summary

- Guard `torch.distributed.get_rank()` calls in `PersistentAsyncCaller.close()` with `torch.distributed.is_initialized()` to prevent `ValueError` when `close()` is invoked from `__del__` during Python garbage collection, after the distributed process group has already been destroyed.

Fixes #3775

## Test plan

- [ ] Enable async checkpointing (`config.checkpoint.async_save = True`) and run a training job to completion — verify no `ValueError` traceback on shutdown
- [ ] Verify that debug log messages still appear during normal (non-shutdown) `close()` calls
- [ ] Test abort path (`close(abort=True)`) both with and without initialized process group

### DIFF
--- a/megatron/core/dist_checkpointing/strategies/async_utils.py
+++ b/megatron/core/dist_checkpointing/strategies/async_utils.py
@@ -493,16 +493,21 @@ class PersistentAsyncCaller(AsyncCaller):
             abort (bool, optional): Default to False. Needs to be manually set to true when
                 the checkpoint async process needs to be aborted.
         """
-        logger.debug(
-            f"PersistentAsyncCaller: {torch.distributed.get_rank()}, Destroying Async Caller"
-        )
+        if torch.distributed.is_initialized():
+            logger.debug(
+                f"PersistentAsyncCaller: {torch.distributed.get_rank()}, Destroying Async Caller"
+            )
         if self.process:
             if abort:
-                log_single_rank(
-                    logger,
-                    logging.WARNING,
-                    f"Persistent worker aborted in rank {torch.distributed.get_rank()}",
-                )
+                if torch.distributed.is_initialized():
+                    log_single_rank(
+                        logger,
+                        logging.WARNING,
+                        f"Persistent worker aborted in rank {torch.distributed.get_rank()}",
+                    )
+                else:
+                    logger.warning("Persistent worker aborted (rank unavailable)")
+
                 self.process.kill()
             else:
                 self._persistent_queue.put('DONE')


cluster2600:
### Tested on NVIDIA GB10 (DGX Spark), PyTorch 2.10, Python 3.12

**Before fix** — bug reproduced with a minimal script that inits dist, creates a `PersistentAsyncCaller`, destroys the process group, then calls `close()`:

```
Exception ignored in: <function PersistentAsyncCaller.__del__ at 0xe15f200a6480>
  File ".../async_utils.py", line 497, in close
    f"PersistentAsyncCaller: {torch.distributed.get_rank()}, Destroying Async Caller"
ValueError: Default process group has not been initialized
```

Both explicit `close()` and `__del__` via GC crash.

**After fix** — clean exit, no errors:

```
process group destroyed
calling close() after dist teardown...
SUCCESS: close() completed without error
letting __del__ fire via del...
done
```

cluster2600:
Closing in favour of #3781, which takes a broader approach — it introduces a centralised `_get_rank_or_unknown()` helper and covers `AsyncCaller`, `TemporalAsyncCaller`, and `PersistentAsyncCaller` rather than just the last one.

The DGX Spark test results posted here are still valid and I've added them to #3781 as well. Cheers!

PR #3899, opened 2026-03-17
Adding NVRx as a dependency and keeping the current code base optionally

dimapihtar:
# What does this PR do ?

1. Adds NVRx async save routines and deprecates mcore's.
2. Introduces `async_strategy` param to make it configurable from the user perspective.
3. NVRx strategy is set as default async strategy. MCore's async strategy will be used if NVRx is not installed.
4. Still possible to use MCore's async strategy by setting `--async-strategy mcore`.
5. Keeps mcore's async strategy for backward compatibility but will be fully removed and subsituted with related nvrx solution in 1-2 releases.

### Pre-checks

### Code review

6. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/training/async_utils.py
```diff
@@ -101,4 +111,5 @@ def reset_persistent_async_worker():
         del _results_queue
     _results_queue = None
```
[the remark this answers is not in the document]
sbak5: you need to import _results_queue from NVRx or megatron.core.dist_checkpointing.filesystem_async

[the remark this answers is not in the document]
sbak5: @dimapihtar

tests/functional_tests/test_cases/gpt/gpt3_mcore_te_tp4_pp1_resume_torch_dist_dist_optimizer_overlap_grad_reduce_param_gather/model_config.yaml
```diff
@@ -54,5 +54,6 @@ MODEL_ARGS:
   --bf16: true
   --log-memory-to-tensorboard: true
   --async-save: true
```
sbak5, line 56: Only a few functional tests run with mcore and others run with nvrx?

sbak5:
https://github.com/NVIDIA/Megatron-LM/actions/runs/23266155298/job/67649083947?pr=3899
It's supposed to raise error during writing but this injection may not work with some issue. @dimapihtar

megatron/core/dist_checkpointing/strategies/fully_parallel.py
```diff
@@ -185,7 +193,12 @@ def __init__(
         self.cached_global_metadata: Optional[Metadata] = None
 
     @debug_time("FullyParallelLoadStrategyWrapper.load", logger)
-    def load(self, sharded_state_dict: ShardedStateDict, checkpoint_dir: Path) -> StateDict:
+    def load(
```
sbak5, line 196: Let's remove any load related changes. We don't have anything yet for loading.

dimapihtar: @sbak5 load calss `_get_filesystem_reader`: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/dist_checkpointing/strategies/torch.py#L807

which uses `CachedMetadataFileSystemReader`: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/dist_checkpointing/strategies/torch.py#L766 so we need to import it properly in respect to `async_strategy`.

dimapihtar: hmm but I guess we can take it from common state dict
let me see

sbak5:
@dimapihtar please resolve the conflicts.

megatron/training/config/training_config.py
```diff
@@ -441,6 +441,9 @@ class CheckpointConfig:
     async_save: bool = False
     """Apply async checkpointing save. Currently works only with `torch_dist` distributed checkpoint format."""
 
+    async_strategy: Literal["nvrx", "mcore"] = "nvrx"
```
asolergi-nv, line 447: Do we really want `nvrx` to be the default backend from now on?

sbak5 (approved):

megatron/training/arguments.py
```diff
@@ -1434,6 +1434,8 @@ def validate_args(args, defaults={}):
                 'Disabling --async-save.'
             )
             args.async_save = False
+        assert args.async_strategy in ["nvrx", "mcore"], \
```
jaredcasper: This should be in the post_init for CheckpointConfig, not in arguments.py.

tests/unit_tests/dist_checkpointing/test_local.py
```diff
@@ -151,7 +151,7 @@ def test_sharded_tensors(self, tp, pp, use_torch_fsdp2):
     @pytest.mark.internal
     @pytest.mark.parametrize(('tp,pp'), [(2, 4), (1, 1)])
     @pytest.mark.parametrize(('use_ramdisk'), [True, False])
-    @pytest.mark.parametrize(('async_save'), [True, False])
+    @pytest.mark.parametrize(('async_save'), [False])
```
jaredcasper: So we are no longer testing async_save here (and many other places with this change?)

uv.lock
```diff
@@ -43,9 +43,11 @@ version = "1.13.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "huggingface-hub" },
-    { name = "numpy", version = "2.2.6", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.11' or (extra == 'extra-13-megatron-core-dev' and extra == 'extra-13-megatron-core-lts')" },
-    { name = "numpy", version = "2.4.3", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.11' or (extra == 'extra-13-megatron-core-dev' and extra == 'extra-13-megatron-core-lts')" },
-    { name = "packaging" },
+    { name = "numpy", version = "2.0.2", source = { registry = "https://pypi.org/simple" }, marker = "extra == 'extra-13-megatron-core-dev'" },
```
jaredcasper, line 78: What's going on with these changes? These seem unrelated?

tests/unit_tests/dist_checkpointing/test_local.py
dimapihtar: > So we are no longer testing async_save here (and many other places with this change?)

yes, nvrx has it's own unit tests for async save. We just enabled `async_save` for our functional tests so we can make sure it's covered.

uv.lock
sbak5: We've run a suggested command from @ko3n1g to update uv.lock. 
We've updated the commit for NVRx and run `uv lock` with the ci container.

tests/unit_tests/dist_checkpointing/test_local.py
sbak5: this test is made for the previous local ckpt work. A similar test is in NVRx and we'll redesign this local ckpt. As far as I know, this is not used by anyone.

megatron/training/arguments.py
dimapihtar: fixed

megatron/core/dist_checkpointing/strategies/fully_parallel.py
```diff
@@ -185,7 +193,12 @@ def __init__(
...
+        self,
+        sharded_state_dict: ShardedStateDict,
+        checkpoint_dir: Path,
+        async_strategy: str = "nvrx",
```
asolergi-nv, line 200: Same, are we sure we want nvrx to be the default strategy from now on?

megatron/core/dist_checkpointing/strategies/torch.py
```diff
@@ -711,62 +745,80 @@ def async_save(
...
+                        self.cached_global_metadata = save_state_dict_ret[1]  # Cache Metadata
+                    # Only Coordinator rank holds cached global_metadata
+                    # (None is returned for global_metadata)
+                    elif coordinator == rank:
+                        logger.debug(
+                            f"rank: {rank}, reuse global metadata cached from previous"
+                            f" save iteration, {save_state_dict_ret[1]}"
+                        )
+                        save_state_dict_ret = list(save_state_dict_ret)
+                        save_state_dict_ret[1] = self.cached_global_metadata
```
asolergi-nv, line 788: Should we put all this code into a new method e.g. _validate_checkpoint_save_metadata or smth?

megatron/core/dist_checkpointing/strategies/torch.py
```diff
@@ -961,3 +1020,87 @@ def check_backend_compatibility(self, loaded_version):
...
+                "save_state_dict_async_plan": save_state_dict_async_plan,
+            }
+            async_strategy = "nvrx"
+        except (ImportError, ModuleNotFoundError):
+            # do mcore async imports
+            imports = _import_mcore_async()
+            async_strategy = "mcore"
+    elif async_strategy == "mcore":
+        # do mcore async imports
+        imports = _import_mcore_async()
```
asolergi-nv, line 1068: Why do we have `_import_mcore_async` for mcore imports but not `_import_nvrx_async` for nvrx imports? Should we harmonize this, WDYT?

asolergi-nv (commented):
What is stoping us from having a megatron/core/dist_checkpointing/strategies/nvrx.py strategy?

megatron/training/config/training_config.py
sbak5: Yes. The code has migrated already and NVRx side has more chnages than the codebase in MCore.

megatron/core/dist_checkpointing/strategies/fully_parallel.py
sbak5: Yes.

megatron/core/dist_checkpointing/strategies/torch.py
sbak5: We temporarily keep this code until we remove the code in MCore. Further refactoring doesn't seem necessary. (This code was in `async_save` in the same method)

megatron/training/config/training_config.py
asolergi-nv: And what will happen if the user doesn't have NVRx installed? Are we guarding against this? Should we include a informative message stating that the `mcore` backend is still available, but they should move to NVRx?

sbak5: We can do some warning and fall back to mcore when NVRx is not installed. but the message would be like mcore code base will be deprecated and frozen. latest perf fixes will be only available in NVRx.

megatron/core/dist_checkpointing/strategies/torch.py
dimapihtar: I've added `_import_mcore_async` to avoid code duplication because we need to import mcore-related modules twice in this funciton but nvrx only once so I don't think this func is really necessary for nvrx imports.

megatron/training/config/training_config.py
dimapihtar: @asolergi-nv @sbak5 

1. we do that in `get_async_strategy` func. Mcore async implemnetation will be automatically used if nvrx is not installed. Maybe we just need to add additional warning.
2. We have a deprecation message for mcore async.

asolergi-nv: I see, in the except block.

jaredcasper: > we do that in get_async_strategy func. Mcore async implemnetation will be automatically used if nvrx is not installed

We should not do that. You shouldn't silently use different code than someone requested (a warning in megatron logs is "silent"). If they request nvrx and it isn't available, we should error out and make them change the backend.

uv.lock
jaredcasper: @ko3n1g can you verify this is what you intended? Seems weird to be changing numpy versions, etc.?

tests/unit_tests/dist_checkpointing/test_local.py
jaredcasper: So you're saying that we are just testing this in NVRx and this is just removing the tests of the lame-duck/dead code in mcore? How are we going to test the interaction between the two libraries?

uv.lock
ko3n1g: Thanks @jaredcasper. Yeah, this on me - the initial tooling i provided causes this.

@sbak5 Can you fetch the uv.lock off main and re-run the [CLI](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/upgrade_dependencies.sh) with the updated [instructions](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/upgrade_dependencies.md)? It will lead to a cleaner minimal diff

yaoyu-33 (approved):

megatron/core/dist_checkpointing/strategies/torch.py
jaredcasper: now that we don't fall back to mcore when nvrx isn't available we can just bring them up into here? Make it a bit cleaner?

sbak5: https://github.com/sbak5/Megatron-LM/blob/sbak/ckpt_migrate/megatron/core/dist_checkpointing/strategies/torch.py#L1028-L1065
We don't fall back to mcore but raising an issue. Not sure why this update doesn't show up here.

tests/unit_tests/dist_checkpointing/test_local.py
sbak5: I'll handle this because local ckpt works with temporal Async Caller. This PR sets `async_calls_queue` only when use_persistent_ckpt_worker is set. Let me make a fix.

sbak5: we've reverted these changes because this unit test fails with a change to set async_calls_queue to None at `training/async_utils.py`. 
Added a commit to fix this.

megatron/core/dist_checkpointing/strategies/torch.py
jaredcasper: right, but now we don't need a _import_mcore_Async() function, since it is only called once. so should pull that up inline here. But not super important I guess since it'll go away at some point.

jaredcasper (approved):

PR #3913, opened 2026-03-18
Add --step-batch-size-schedule for step-wise batch size control

mchrzanowski:
Adds a new --step-batch-size-schedule argument that allows specifying arbitrary step-wise batch size schedules (e.g. "0:768 250B:1536 500B:3072"). Thresholds support K/M/B/T suffixes and are interpreted as tokens when --seq-length is provided, otherwise as samples. Mutually exclusive with --rampup-batch-size.

Includes deferred divisibility checking so early schedule entries smaller than the current GPU configuration don't crash at init, a runtime check in setup_model_and_optimizer that validates world size can accommodate the current batch size after checkpoint loading, and sample-based training iteration calculation via update_train_iters.

### DIFF
--- a/megatron/core/num_microbatches_calculator.py
+++ b/megatron/core/num_microbatches_calculator.py
@@ -262,6 +262,10 @@ def _build_num_microbatches_calculator(
         raise ValueError(
             'Cannot specify both --step-batch-size-schedule and --rampup-batch-size'
         )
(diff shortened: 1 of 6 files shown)


mchrzanowski:
closing because @mkhona-nvidia already has a PR for this: https://github.com/NVIDIA/Megatron-LM/pull/3779/

PR #3558, opened 2026-02-24
[Draft] Fix excess CUDA memory reservation on resume of training

Birch-san:
# What does this PR do ?

Fixes excess CUDA reserved memory when resuming training from a torch_dist checkpoint with optim states (https://github.com/NVIDIA/Megatron-LM/issues/2339, https://github.com/NVIDIA/Megatron-LM/issues/1380, https://github.com/NVIDIA/Megatron-LM/issues/1746)

I will not able to test this on main Megatron; we have forked pretty heavily. But thought I'd at least show an (untested) draft of what this patch would look like on main, in case anybody else is struggling with this and wants a lead on how to fix it.

### The problem

If I train Qwen3 8B on 2 GPUs, my CUDA memory usage is different depending on whether I start training anew (via `--pretrained-checkpoint`) or resume training via `--load` (with optimizer weights).

If I add this logging in `train()`:

```diff
  while iteration < args.train_iters:
+     print_rank_0(f"Beginning iteration {iteration}... Memory: {torch.cuda.memory_reserved()/1024**3:.1f}GiB")
```

Training anew, I get:

```
Beginning iteration 0... Memory: 30.0GiB
Beginning iteration 1... Memory: 43.0GiB
Beginning iteration 2... Memory: 43.7GiB
Beginning iteration 3... Memory: 43.7GiB
```

Resuming from a torch_dist checkpoint saved (with optimizer weights) on step 100, I get more memory usage:

```
Beginning iteration 100... Memory: 63.3GiB
Beginning iteration 101... Memory: 64.1GiB
Beginning iteration 102... Memory: 64.1GiB
Beginning iteration 103... Memory: 64.1GiB
```

The main allocations are:

get_model(): 0->30.0GiB  
_load_base_checkpoint: 30.0 -> 60.0GiB (we can get this down to 41.2GiB if we empty_cache()
optimizer.load_state_dict(state_dict['optimizer']): 60.0 -> 63.3GiB

I believe the problem is that `load_kwargs["sharded_state_dict"]`, `state_dict['optimizer']`, and possibly (I didn't check this one) `state_dict['model']` retain python references to CUDA tensors. we need to not only free the references, but also GC collect them, and then we can empty_cache() to get our CUDA memory back.

For me, this improved my situation on resumes to:

```
Beginning iteration 101... Memory: 43.7GiB
Beginning iteration 102... Memory: 43.7GiB
Beginning iteration 103... Memory: 43.7GiB
```

which is parity with training anew.

[...]

PR #3920, opened 2026-03-18
Free memory to avoid OOM when resume optimizer states

BestJuly:
# What does this PR do ?
## Summary

When resuming training **with optimizer states loaded** (the default behavior), the first training iteration can OOM during the logging phase. After `train_step()` completes, PyTorch's CUDA caching allocator still holds freed activation memory. When `training_log()` subsequently triggers NCCL collectives (aux-loss all-reduce across PP/DP groups, timer barrier, etc.), the NCCL internal buffer allocations push GPU memory over the limit.

The fix calls `torch.cuda.empty_cache()` once on the first iteration, right before `training_log()`, to return cached-but-unused memory to the CUDA driver and make room for NCCL buffers.

- Only runs on the first iteration — no steady-state performance impact.
- `empty_cache()` does not affect any live tensors or training correctness.
- Adding `--no-load-optim` avoids the OOM (optimizer states are zero-initialized instead of loaded), confirming that the extra memory from resumed optimizer states is the root cause.

## Reproduction

The issue reproduces with a MoE proxy model using the distributed optimizer and expert parallelism. Reproduction:

```bash
torchrun --nproc_per_node=8 pretrain_gpt.py --distributed-timeout-minutes 220 --tensor-model-parallel-size 1 --pipeline-model-parallel-size 1 --expert-model-parallel-size 8 --context-parallel-size 1 --expert-tensor-parallel-size 1 --use-distributed-optimizer --no-create-attention-mask-in-dataloader --cross-entropy-loss-fusion --cross-entropy-fusion-impl te --enable-experimental --use-mcore-models --sequence-parallel --use-flash-attn --disable-bias-linear --micro-batch-size 1 --global-batch-size 32 --train-samples 268554688 --exit-duration-in-mins 230 --manual-gc --manual-gc-interval 5 --transformer-impl transformer_engine --tokenizer-type HuggingFaceTokenizer --tokenizer-model Qwen/Qwen3-235B-A22B --mock-data --split 99,1,0 --no-mmap-bin-files --num-workers 6 --untie-embeddings-and-output-weights --position-embedding-type rope --rotary-percent 1.0 --rotary-base 1000000 --rotary-seq-len-interpolation-factor 1 --normalization RMSNorm --swiglu --norm-epsilon 1e-06 --num-layers 10 --hidden-size 8192 --ffn-hidden-size 12288 --num-attention-heads 64 --kv-channels 128 --group-query-attention --num-query-groups 4 --qk-layernorm --seq-length 4096 --max-position-embeddings 4096 --make-vocab-size-divisible-by 1187 --attention-dropout 0.0 --hidden-dropout 0.0 --clip-grad 1.0 --weight-decay 0.1 --lr-decay-samples 584765624 --lr-warmup-samples 1536000 --lr-warmup-init 3.9e-7 --lr 3.9e-6 --min-lr 3.9e-7 --lr-decay-style cosine --adam-beta1 0.9 --adam-beta2 0.95 --num-experts 16 --moe-ffn-hidden-size 1536 --moe-router-load-balancing-type aux_loss --moe-router-topk 8 --moe-grouped-gemm --moe-aux-loss-coeff 1e-3 --moe-permute-fusion --moe-router-dtype fp32 --moe-router-fusion --eval-iters 32 --eval-interval 500 --auto-detect-ckpt-format --load ./results/debug_optim_load/checkpoints --save ./results/debug_optim_load/checkpoints --save-interval 10 --dist-ckpt-strictness log_all --init-method-std 0.02 --log-timers-to-tensorboard --log-memory-to-tensorboard --log-num-zeros-in-grad --log-params-norm --log-validation-ppl-to-tensorboard --log-throughput --log-interval 1 --logging-level 40 --tensorboard-dir ./results/debug_optim_load/tensorboard --bf16 --moe-token-dispatcher-type alltoal
```
After checkpointing saving and re-launch the script, without the PR it will OOM but works well by adding `--no-load-optim`. This PR helps to resume optimizer states without OOM error.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


asolergi-nv:
We have #3558 which presents a similar issue. Should we investigate this further? In #3558 they are calling `torch.cuda.empty_cache()` earlier during checkpoint load.

jaredcasper:
> We have #3558 which presents a similar issue. Should we investigate this further? In #3558 they are calling `torch.cuda.empty_cache()` earlier during checkpoint load.

It seems the excess memory does come from checkpoint load? (otherwise why would we only need to do this on the first iteration?). If that's the case I think #3558 makes more sense, making it clear it's releasing memory from checkpoint load. As is here it's a bit confusing why its only done on the first iteration.

BestJuly:
Thanks @asolergi-nv and @jaredcasper for the info. I think the issue should be the same and I have verified the [PR#3558](https://github.com/NVIDIA/Megatron-LM/pull/3558https://github.com/NVIDIA/Megatron-LM/pull/3558) can also help with my case. Close this PR as that PR make it more clear for memory releasing.

PR #3922, opened 2026-03-18
CUDA graph support for prefix caching on hybrid models

lmcafee-nvidia:
## Summary
- Make Mamba state extraction compatible with CUDA graphs during prefix caching
- Enable CUDA graphs for chunked prefill on hybrid (transformer + Mamba) models
- Add comprehensive correctness tests for prefix caching + CUDA graph interactions

## Test plan
- [ ] `test_prefix_caching_cuda_graphs.py` passes
- [ ] Existing `test_dynamic_prefix_caching.py` tests still pass
- [ ] `test_dynamic_engine.py` chunked prefill + CUDA graph tests pass


megatron/core/inference/contexts/attention_context/mamba_metadata.py
```diff
@@ -82,6 +87,23 @@ def __init__(self, max_requests: int, max_tokens: int, mamba_chunk_size: int = 1
         )
         self.mamba_state_free_slot_count = self.max_requests
 
+        # Intermediate state extraction buffers (CUDA graph compatible)
+        # Each prefill request can produce up to 3 intermediate offsets
+        self.max_intermediate_count = 3 * max_requests
```
santhnm2: Why is this hardcoded to 3?

megatron/core/inference/contexts/attention_context/mamba_metadata.py
```diff
@@ -284,13 +313,91 @@ def update(
...
+                offsets = intermediate_token_offsets[i]
+                count = 0
+                for offset in offsets:
+                    chunk_idx = first_chunk_idx + (offset // chunk_size) - 1
+                    abs_pos = seq_start + offset
+                    chunk_indices_list.append(chunk_idx)
+                    abs_positions_list.append(abs_pos)
+                    count += 1
+                per_request_counts.append(count)
+                cumulative_chunks += num_chunks
```
santhnm2: As @kvareddy alluded to, can we vectorize this / directly construct the tensors rather than using lists and loops?

megatron/core/inference/contexts/attention_context/mamba_metadata.py
```diff
@@ -284,13 +313,91 @@ def update(
...
+
+            # Fill buffers: real entries + padding
+            if real_count > 0:
+                self._intermediate_chunk_indices_buffer[:real_count].copy_(
+                    torch.tensor(chunk_indices_list, dtype=torch.int64)
+                )
+                self._intermediate_abs_positions_buffer[:real_count].copy_(
+                    torch.tensor(abs_positions_list, dtype=torch.int32)
+                )
+            # Pad remaining with safe defaults
```
santhnm2: Could you elaborate on this comment as to why these are the safe defaults?

megatron/core/inference/contexts/mamba_slot_allocator.py
```diff
@@ -70,7 +70,19 @@ def __init__(
         self._intermediate_offsets: list = [None] * context.max_requests
         self._intermediate_block_ids: list = [None] * context.max_requests
         self._eos_cache_block_id: list = [None] * context.max_requests
-        self._intermediate_buffer: dict = {}
+
+        # Pre-allocated output buffers for CUDA graph compatible extraction
+        self.max_intermediate_count = 3 * context.max_requests
```
santhnm2: Same comment as above about hardcoding this to 3

megatron/core/ssm/mamba_mixer.py
```diff
@@ -583,17 +570,28 @@ def _dynamic_inference_prefill(
...
-        result = self._ssm_prefill(
+        # Prepare intermediate extraction buffers (always passed, CUDA graph compat)
+        slot_allocator = context.mamba_slot_allocator
+        intermediate_chunk_indices = metadata.intermediate_chunk_indices
+        intermediate_abs_positions = metadata.intermediate_abs_positions
+        intermediate_ssm_out = None
+        intermediate_conv_out = None
+        if slot_allocator is not None and mamba_layer_idx is not None:
+            intermediate_ssm_out = slot_allocator._intermediate_ssm_out[mamba_layer_idx]
+            intermediate_conv_out = slot_allocator._intermediate_conv_out[mamba_layer_idx]
```
santhnm2: If these attributes are accessed publicly should we remove the leading underscore? Same thing with `_conv_gather_offsets`

megatron/core/inference/contexts/mamba_slot_allocator.py
```diff
@@ -387,32 +392,42 @@ def commit_intermediate_states(self) -> None:
...
+                                self._intermediate_ssm_out[layer_idx, ssm_offset + j]
+                            )
+                            self.conv_states[layer_idx, slot].copy_(
+                                self._intermediate_conv_out[layer_idx, ssm_offset + j]
+                            )
+
+                        # Register in mamba hash map
+                        block_hash = ctx.kv_block_allocator.block_hashes[bid].item()
+                        if block_hash > 0:
+                            self.register_block_hash(bid, block_hash)
```
santhnm2: Same comment about potential vectorization

megatron/core/inference/contexts/attention_context/mamba_metadata.py
lmcafee-nvidia: extracted to `MAX_INTERMEDIATE_OFFSETS_PER_REQUEST = 3` in `mamba_slot_allocator.py` with a comment explaining the three candidates:
- sequence divergence point
- last-aligned block
- last-aligned block - 1 (the 'penultimate' block; this is for an edge case where we need to recompute mamba state, and we save this block so that we don't need to recompute from the start of the sequence)

the three potential extraction points are computed in `compute_and_store_offsets()`. and both `mamba_metadata.py` and `mamba_slot_allocator.py` now reference this constant.

megatron/core/inference/contexts/mamba_slot_allocator.py
lmcafee-nvidia: see my other comment about this, but (up to) three intermediate indices are: 1) divergence point, 2) penultimate last-aligned block, 2) last-aligned block. and this constant value of 3 has been abstracted into a variable and used where it's needed

megatron/core/inference/contexts/attention_context/mamba_metadata.py
lmcafee-nvidia: done, we now use vectorized GPU ops here. please take another look

megatron/core/inference/contexts/attention_context/mamba_metadata.py
lmcafee-nvidia: expanded the comment, let me know if it's clear

megatron/core/inference/contexts/mamba_slot_allocator.py
lmcafee-nvidia: this method has been partially vectorized, in the sense that block IDs and hashes are batch-transferred from the GPU via a single `.tolist()`. the loop isn't easily parallelizable though because it may trigger LRU eviction with side effects like deregistration and free-pool mutation

megatron/core/ssm/mamba_mixer.py
lmcafee-nvidia: good catch, I removed the leading underscores from these vars

megatron/core/inference/contexts/mamba_slot_allocator.py
santhnm2: Would it be possible to do the copy into `conv_states` and `ssm_states` as a first step and then do the block hash registration as a second step? And if so, maybe we could more effectively parallelize at least the copy into `conv_states` and `ssm_states` (maybe with a Triton kernel?)

megatron/core/inference/contexts/attention_context/mamba_metadata.py
santhnm2: Sounds good, thanks

megatron/core/inference/contexts/attention_context/mamba_metadata.py
santhnm2: Yup this is good, thanks

megatron/core/inference/contexts/mamba_slot_allocator.py
lmcafee-nvidia: this method has been completely rewritten into a pipeline of batch-operation sub-methods. I think the biggest improvement here comes from batching the slot allocation (`allocate_slots_batch()`), since this potentially results in eviction which is costly. please take another look.

PR #3813, opened 2026-03-11
Add Lion optimizer support 

mchrzanowski:
Import Lion optimizer from emerging_optimizers and integrate it as a standalone optimizer (--optimizer lion) and as a Muon nonlinear parameter optimizer (--muon-nonlinear-optimizer lion).

PR #3927, opened 2026-03-18
Add --muon-coefficient-type argument for Muon optimizer

mchrzanowski:
Allow users to select the Newton-Schulz polynomial coefficient set (e.g. simple, quintic, polar_express, aol) via the new --muon-coefficient-type CLI flag. Supported types are discovered dynamically from the installed emerging_optimizers package so that upstream additions are picked up automatically without code changes.


megatron/core/optimizer/optimizer_config.py
```diff
@@ -272,6 +272,10 @@ class OptimizerConfig:
     muon_fp32_matmul_prec: str = "medium"
     """The precision to use for the fp32 matmul. Defaults to "medium"."""
 
+    muon_coefficient_type: str = "quintic"
```
mkhona-nvidia, line 275: Again, would recommend pinging downstream users (MBridge, etc) to add this to the list of Muon args

mkhona-nvidia (approved):

chtruong814 (approved):

megatron/core/optimizer/muon.py
```diff
@@ -288,6 +311,7 @@ def lion_init_state_fn(opt, config=None):
         "use_nesterov": config.muon_use_nesterov,
```
deepakn94, line 307: BTW, should we check for certain versions of `emerging_optimizers`? I believe this got re-named to `nesterov` in the latest release.

FDecaYed: I feel it is better to keep the code only support one version. And I think having that version be the tagged in pyproject is reasonable. Optionally we can add a global check.
We will bump main to support v0.2.0 soon(after dev refactor+bump and main2dev sync, both are finalizing).

cuichenx (approved):

megatron/core/optimizer/muon.py
```diff
@@ -48,6 +48,28 @@
 logger = logging.getLogger(__name__)
 
 
+def get_supported_coefficient_types() -> tuple[str, ...]:
+    """Return the coefficient types supported by the installed emerging_optimizers.
+
+    Reads the members of the ``NSCoeffT`` Literal type so that new types
```
skyw, line 45: Suggestion: Don't make this dependency. Couple of reasons:
* It usually needs a good reason for runtime to depend on static type checking code. NSCoeffT is not supposed to be used in runtime by design.
* `NSCoeffT` is not guarantee to provide all the supported coefficient type. It is used for static type checking. and there are options to expand like use union, e.g. `Union[NSCoeffT, ExpCoeffT, "blah"]` as long as it can be justified instead of adding every new type into the type.

Although I don't incline to change, it is not beneficial for megatron to make this dependency.

mchrzanowski: i mean do you have a better idea other than hard-coding the acceptable strings right now? that seems so hacky; you cant even see the list of supported coefficient types because it's in another repo

skyw: It would depend on the goal, automatically picking up new types? or verify megatron use the right coefficient type string to call emerging_optimizers, or both.

For the former, it needs emerging_optimizer to commit to interface stability (which it hasn't although we try to avoid breaking changes) and provide a blessed list for public consumption. For the latter, Megatron can also use type check in theory (although reality is hard). There are many alternatives, depend on the level of protection you want to achieve.

mchrzanowski: i was thinking more of the former; agreed that type checking sounds hard and not worth the effort

mchrzanowski: i think im going to leave this in there for now, but we can refactor once emerging optimizers decided what it wants to do

megatron/core/optimizer/muon.py
```diff
@@ -28,7 +28,7 @@
         OrthogonalizedOptimizer,
         get_muon_scale_factor,
     )
-    from emerging_optimizers.orthogonalized_optimizers.muon_utils import newton_schulz_tp
+    from emerging_optimizers.orthogonalized_optimizers.muon_utils import NSCoeffT, newton_schulz_tp
```
ericharper: Have we updated the requirements to use v0.2.0? If so can we also update the Lion import that  we moved previously?

mchrzanowski: done

deepakn94 (approved):

megatron/core/optimizer/optimizer_config.py
ericharper: @janEbert , should we add this to mbridge?

janEbert: I really only see [`optimizer_utils.py`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/c15475efef0275e946cc797c2b3c4c08910b6897/src/megatron/bridge/recipes/utils/optimizer_utils.py#L28-L42) as a place with existing Muon config values where the new arguments could meaningfully be added. However, even that function is only used by the Kimi-K2 recipe and gets hardcoded arguments passed in. Is that the place you're looking to have modified? That's also [where I added](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2959) the [previous PR's](https://github.com/NVIDIA/Megatron-LM/pull/3813) changes.

Another option would be [`argument_parser.py`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/f386fd68c9646358e128ce56cae332aa585263ba/scripts/performance/argument_parser.py#L553), but that file doesn't contain any other Muon settings yet (not even Adam's betas).

I guess it speaks for MBridge's design; the underlying MCore implementation is just used and doesn't really need to be updated as much. :)

ericharper (approved):

mchrzanowski:
/ok to test [0d613ec](https://github.com/NVIDIA/Megatron-LM/pull/3927/commits/0d613ec896750ce32618fc090866bca97a9db62c)

megatron/core/optimizer/muon.py
skyw: sounds fine.

mchrzanowski:
/ok to test [10bf5eb](https://github.com/NVIDIA/Megatron-LM/pull/3927/commits/10bf5eb9285246b51e3b4a53d724c69cc7063a03)

PR #3928, opened 2026-03-18
Inference | Per-block MoE routing storage for prefix caching compatibility

lmcafee-nvidia:
## Summary
- Moves MoE routing indices from per-request step-by-step accumulation to per-block storage on `KVBlockAllocator`, fixing the serialization assertion failure when both `enable_prefix_caching` and `moe_enable_routing_replay` are enabled.
- After each forward pass, routing is scattered into per-block tensors using the context's token-to-block mapping. At request completion, full routing is reconstructed by concatenating blocks in order.
- Matched (prefix-cached) blocks retain routing from the original request, so reconstruction naturally covers all tokens including skipped prefixes.

## Test plan
- [x] All 16 existing prefix caching tests pass
- [x] 7 new `TestPerBlockRouting` tests: store/get round-trip, clear-on-allocate, clear-on-reset, persist-through-deregister, full reconstruction with partial last block, graceful None on missing data, LRU matched blocks retain routing
- [ ] End-to-end: routing replay + prefix caching with MoE model (serialization assertion passes)


megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -1268,6 +1268,74 @@ def _router_record_bookkeeping(self) -> Optional[Dict[int, Tensor]]:
...
+        active_request_ids = context.request_ids[active_request_slice].tolist()
+        routing_parts = [
+            routing_indices_per_request[rid]
+            for rid in active_request_ids
+            if rid in routing_indices_per_request
+        ]
+        if not routing_parts:
+            return
+        flat_routing = torch.cat(routing_parts, dim=0)  # [token_count, num_layers, topk]
+        if flat_routing.shape[0] != token_count:
```
kvareddy: Should we assert here?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -1268,6 +1268,74 @@ def _router_record_bookkeeping(self) -> Optional[Dict[int, Tensor]]:
...
+
+        # Move to CPU for dict-based storage
+        flat_routing_cpu = flat_routing.cpu()
+        block_ids_cpu = block_ids.cpu()
+        positions_cpu = positions.cpu()
+
+        block_size = context.block_size_tokens
+        dummy = allocator.dummy_block_idx
+
+        # Group tokens by block_id using sort for efficient scatter
```
kvareddy: We might need a kernel for this computation.

PR #3776, opened 2026-03-10
Track errors through the inference return path

tdene:
# What does this PR do ?

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

Once all required reviewers have approved, the `Approved` label is applied **automatically**.

### Merge

Any member of [mcore-engineers](https://github.com/orgs/NVIDIA/teams/mcore-engineers) will be able to merge your PR.

<details>
<summary>For MRs into `dev` branch</summary>
The proposed review process for `dev` branch is under active discussion.

MRs are mergable after one approval by either `eharper@nvidia.com` or `zijiey@nvidia.com`.
</details>

PR #3930, opened 2026-03-18
Improve load balancing behavior for prefix cache-aware routing

santhnm2:
# What does this PR do ?
Improves the load balancing behavior for prefix cache-aware routing by prioritizing ranks with less load in the event of ties on the longest prefix match.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


kvareddy (approved):

kvareddy (approved):

megatron/core/inference/data_parallel_inference_coordinator.py
```diff
@@ -253,22 +255,46 @@ def compute_request_hashes(self, prompt):
...
+        rank (fewest pending requests) is preferred. Ties on load are broken by
+        most-recent timestamp. Falls back to round-robin when no rank matches.
 
         Args:
             request_hashes: List of block hashes for the request.
 
         Returns:
             bytes: The ZMQ identity of the selected data parallel rank.
         """
+        # Always prioritize completely idle ranks regardless of routing policy.
```
lmcafee-nvidia: I don't know if I like the idea of prioritizing idle ranks over the routing policy itself. let's say we have 8 DP ranks and 7 unique prompts, each with several duplicates. I don't think I agree that a prompt with a prefix match on a worker rank should pick an idle rank over the rank with the match. if this is causing us to suffer, this tells me something else might be wrong, or maybe the dataset simply doesn't have the diversity necessary to highlight prefix caching

there's also a tradeoff of sending a request to a rank with a match vs. an idle rank. the idle rank will initially have less competition for resources, but the matching rank will allow skipping prefill computation of the matching part of the prompt. I think it's hard to say what's the generally optimal solution here, it's data dependent. we might also consider adding a flag to control happens in this case.

also, 'idle' is very binary. what if a rank has 1 request instead of 0 requests? it's close to idle but not precisely idle. should we score the load of each rank?

this also brings into question other potential optimizations along this line. for example let's say a request matches on multiple ranks. should we consider the total load of that rank? several things we can consider here.

megatron/core/inference/data_parallel_inference_coordinator.py
```diff
@@ -388,6 +417,10 @@ def start(self):
                     del self.request_id_to_client_request_id[request_id]
                     return
 
+                self.request_id_to_rank[request_id] = next_identity
+                self._rank_pending_count[next_identity] = (
+                    self._rank_pending_count.get(next_identity, 0) + 1
```
lmcafee-nvidia: can we rely on this per-engine request count always being in sync with the engine's actual number of requests? is it possible at all for an engine to drop a request or crash or raise an exception in any kind of way that would lead to the coordinator having an incorrect count here?

megatron/core/inference/data_parallel_inference_coordinator.py
```diff
@@ -467,6 +500,11 @@ def start(self):
                     client_request_identity = self.request_id_to_client_request_id[fid]
                     del self.request_id_to_client_id[fid]
                     del self.request_id_to_client_request_id[fid]
+                    assigned_rank = self.request_id_to_rank.pop(fid, None)
+                    if assigned_rank is not None and assigned_rank in self._rank_pending_count:
+                        self._rank_pending_count[assigned_rank] = max(
```
lmcafee-nvidia: might be related to my previous comment, but why `max(0, ...)` instead of raising an error? something's wrong if this drops below zero, and we're catching it after the fact.

megatron/core/inference/data_parallel_inference_coordinator.py
```diff
@@ -253,22 +255,46 @@ def compute_request_hashes(self, prompt):
         token_tensor = torch.tensor(tokens, dtype=torch.int64)
         return compute_block_hashes_batched(token_tensor, self.block_size_tokens)
 
+    def _get_idle_rank(self):
+        """Return the idle rank with the lowest rank index, or None if all ranks are busy.
+
+        An idle rank is one with zero pending requests. Ties are broken by
+        deterministic rank index (lowest wins) so behavior is predictable.
+        """
+        idle_ranks = [
```
lmcafee-nvidia: can we implement this without a loop? maybe by keeping track of idle ranks in a set? this can get expensive for high-DP counts

megatron/core/inference/data_parallel_inference_coordinator.py
santhnm2: I believe errors should be propagated back to the server now that [!3776](https://github.com/NVIDIA/Megatron-LM/pull/3776) has been merged, so the count should be in sync.

megatron/core/inference/data_parallel_inference_coordinator.py
santhnm2: Changed this to assert that the pending count is >= 1 before decrementing.

megatron/core/inference/data_parallel_inference_coordinator.py
santhnm2: Added vectorization with numpy

megatron/core/inference/data_parallel_inference_coordinator.py
santhnm2: Added a scoring function for each rank as 
```
score = alpha * match + (1 - alpha) * normalized_load
```

megatron/core/inference/hash_rank_table.py
```diff
@@ -0,0 +1,126 @@
...
+        Primarily useful in tests; production code should prefer :meth:`record`.
+        """
+        row = self._get_or_create_row(h)
+        self._timestamps[row, rank_idx] = timestamp
+
+    # ------------------------------------------------------------------
+    # Queries
+    # ------------------------------------------------------------------
+
+    def match_vector(self, hashes: Sequence[int]) -> np.ndarray:
```
lmcafee-nvidia: I think this method needs to be policy aware, and be more of a score than a binary match. for example, if the policy is:
- `first_block_hash`: return 0 or 1
- `longest_prefix`: return match length normalized to [0,1]

the current implementation is what we want for `first_block_hash` (binary), but I think we should also support `longest_prefix` here also. and we should merge the functionality for the  reverse hash scan of for `longest_prefix` so that we only do it once, and it's not duplicated between this method and `get_best_data_parallel_rank()` in `data_parallel_inference_coordinator.py`

megatron/core/inference/hash_rank_table.py
```diff
@@ -0,0 +1,126 @@
...
+        """Return a float64 vector of shape ``(n_ranks,)`` with 1.0 where *any*
+        of the given *hashes* are present for that rank, 0.0 otherwise.
+
+        This is fully vectorized: the only Python-level iteration is over the
+        hash→row dict lookups; the per-rank reduction is done with numpy.
+        """
+        row_indices = [self._hash_to_row[h] for h in set(hashes) if h in self._hash_to_row]
```
lmcafee-nvidia: for each policy, I think we should:
- `first_block_hash`: only check `hashes[:1]`
- `longest_prefix`: do a reverse scan over the hashes and break as soon as we find a match, since finding the single longest match is all we need to know that there's a match

megatron/core/inference/config.py
```diff
@@ -260,6 +260,12 @@ class InferenceConfig:
     Only applies when enable_prefix_caching is True and using a coordinator.
     """
 
+    prefix_caching_routing_alpha: float = 0.5
```
lmcafee-nvidia, line 263: I think inside `__post_init__` we should validate that `0 <= prefix_caching_routing_alpha <= 1`

megatron/core/inference/hash_rank_table.py
```diff
@@ -0,0 +1,126 @@
...
+            self._hash_to_row[h] = row
+        return row
+
+    def _ensure_capacity(self, needed_rows: int) -> None:
+        current = self._timestamps.shape[0]
+        if needed_rows <= current:
+            return
+        new_capacity = current
+        while new_capacity < needed_rows:
+            new_capacity *= 2
```
lmcafee-nvidia: the table grows, but never shrinks. we probably want to shrink it if, say, it's at half capacity

tests/unit_tests/inference/test_data_parallel_inference_coordinator.py
```diff
@@ -670,3 +673,161 @@ async def test_throughput(self, initialize_model_parallel, coordinator, test_cas
...
             await cleanup_engine(engine, client, timeout=60.0)
+
+
+def _set_hash_rank(coord, h, rank_identity, timestamp):
+    """Test helper: set a hash→rank timestamp via HashRankTable."""
+    rank_idx = coord.identity_to_rank_index[rank_identity]
+    coord.hash_table.set(h, rank_idx, timestamp)
+
+
+def _make_routing_coordinator(
```
lmcafee-nvidia, line 691: I think this is mostly duplicated with `make_coordinator_direct` in `test_dynamic_prefix_caching_coordinator.py`, so it would be good to merge this code if possible.

megatron/core/inference/hash_rank_table.py
santhnm2: Added support for `longest_prefix`

megatron/core/inference/hash_rank_table.py
santhnm2: Done

megatron/core/inference/config.py
santhnm2: Done

megatron/core/inference/hash_rank_table.py
santhnm2: Added compaction

tests/unit_tests/inference/test_data_parallel_inference_coordinator.py
santhnm2: Added `cooordinator_test_utils.py` to consolidate

megatron/core/inference/hash_rank_table.py
lmcafee-nvidia: where is `compact()` called? I only see it referenced in tests. can we add compaction directly in this method using a something like:

```
while needed_rows < 0.5 * new_capacity:
    new_capacity /= 2
```

megatron/core/inference/data_parallel_inference_coordinator.py
```diff
@@ -257,36 +278,46 @@ def compute_request_hashes(self, prompt):
...
+        # Build policy-aware match vector (binary for first_prefix_block,
+        # normalized prefix depth for longest_prefix).
+        match = self.hash_table.match_vector(
+            request_hashes, policy=self.prefix_caching_coordinator_policy
+        )
 
+        alpha = self.prefix_caching_routing_alpha
+
+        # Vectorized score: alpha * match + (1-alpha) * free_capacity_fraction.
+        free_slots = np.maximum(0, self.max_requests - self._pending_counts).astype(np.float64)
```
lmcafee-nvidia, line 326: what happens if `max_requests` is `None` or `0`?

santhnm2: Added assertion to guard against this

megatron/core/inference/hash_rank_table.py
santhnm2: Good point, added this logic in the compaction flow

shanmugamr1992 (approved):

PR #3956, opened 2026-03-19
Add moe loss normalization for RL SFT

pthombre:
# What does this PR do ?
This change adds a separate loss scaling function specifically for MoE (Mixture of Experts) auxiliary losses during RL SFT training.                                                                                                                        
                                                                                                                                                                                                                                                              
  Problem: Previously, MoE auxiliary losses used the same grad_scale_func as the main model loss. In RL SFT scenarios, you may need to scale the MoE auxiliary loss differently from the primary loss (e.g., the main loss uses a dynamic scaler tied to RL   
  rewards, but MoE load-balancing losses need their own scaling).
                                                                                                                                                                                                                                                              
  What changed:   
  - A new optional config field moe_grad_scale_func is added to ModelParallelConfig
  - In the forward step, MoE loss scaling now checks for moe_grad_scale_func first, falls back to grad_scale_func if not set, and defaults to 1 otherwise                                                                                                     
  - Note that moe_grad_scale_func takes no arguments (returns the scale directly), unlike grad_scale_func which takes a tensor input — this allows a simpler interface where the caller pre-computes the scale
                                                                                                                                                                                                                                                              
  Impact: Existing code is unaffected since moe_grad_scale_func defaults to None, preserving the current fallback behavior. 

## Contribution process

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


yaoyu-33 (approved):

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -289,13 +289,16 @@ def forward_step_calc_loss(
...
+        # grad_scale_func (fallback), or default to 1.
         device = get_tensor_device(output_tensor)
-        loss_scale = (
-            config.grad_scale_func(torch.ones(1, device=device))
-            if config.grad_scale_func is not None
-            else torch.ones(1, device=device)
-        )
+        moe_grad_scale_func = getattr(config, 'moe_grad_scale_func', None)
+        if moe_grad_scale_func is not None:
+            loss_scale = moe_grad_scale_func()
```
yashaswikarnati, line 323: Would this return a constant value or is this dependent on the microbatch ? I wonder if this can cause issues with scale thats globally stored in  MoEAuxLossAutoScaler.

yashaswikarnati: Like in interleaved 1F1B, the scale computed in one forward pass for a microbatch could be accidentally used for next immediate backward, which might be for a different backward

pthombre (commented):
This is a valid callout. In this particular case moe_grad_scale_func returns a constant per global batch, not a per-microbatch value, so we should be okay even in the interleaved 1F1B case

yashaswikarnati (approved):

yashaswikarnati:
> This is a valid callout. In this particular case moe_grad_scale_func returns a constant per global batch, not a per-microbatch value, so we should be okay even in the interleaved 1F1B case

thanks! Just a nit; would be still great to add a gaurd this incase someone passes a function that returns a fixed scalar and not silently cause wrong behaviour.

jaredcasper (approved):

PR #3970, opened 2026-03-20
Cudagraphs: Remove fwd_graph_input_surface weakref

mathemakitten:
# What does this PR do ?
Tensors in `fwd_graph_input_surface` which are not owned by the cudagraph memory pool should not be weakref'd. `_WeakRefTensor` takes a `data_ptr()` and constructs a weakref, then wraps that pointer back into a torch tensor based on a raw integer address, not the original tensor's storage object. Pytorch's reference counting hinges on the storage object refcount, and the caching allocator frees the block when refcount goes to 0.

The weakref created has a different storage object, so you have two tensors pointing to the same CUDA address, but only the original keeps the caching allocator from freeing it.`tensor_strong_refs` should keep this live in theory, but if the caching allocator is under pressure, or the cache is emptied, it might move the original tensor's `data_ptr()`. Since this tensor is in `tensor_strong_refs` and not the graph pool, it is not guaranteed a fixed address, leaving the weakref holding a stale data pointer has been freed, resulting in a segfault.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/core/transformer/cuda_graphs.py
+++ b/megatron/core/transformer/cuda_graphs.py
@@ -1149,13 +1149,18 @@ class _CudaGraphRunner(torch.nn.Module):
                 return ref
 
             # Weak refs replace tensors with raw-pointer wrappers that do not hold a storage
-            # reference. This is safe for surfaces whose memory is managed by
-            # the CUDA graph pool (driver-pinned, stable addresses) but not safe for
-            # tensors allocated by the caching allocator, whose data_ptr() may be invalidated by
-            # block coalescing or empty_cache().
-            # We do not weakref tensors in fwd_graph_input_surface. These may be allocated from
-            # tensor_reuse_pool which is outside the scope of the graph pool. Weakref'ing them
-            # creates a dangling pointer if the caching allocator reclaims the block.
+            # reference. This is safe for surfaces whose memory is managed by the CUDA graph pool
+            # (driver-pinned, stable addresses) but not safe for tensors allocated by the caching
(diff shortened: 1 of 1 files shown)


megatron/core/transformer/cuda_graphs.py
```diff
@@ -1148,13 +1148,19 @@ def replace_with_weak_ref(arg):
                     ref.can_skip_replay_copy = arg.can_skip_replay_copy
                 return ref
 
-            self.fwd_graph_input_surface = tree_map(
```
jiemingz, line 1151: I think the fwd input surface might hold references to tensors in the graph pool, for instance if an fwd output is directly consumed as a fwd input?
Since `_resolve_input_buffer` will only allocate from the reuse pool, why not filter out only the problematic tensors?

megatron/core/transformer/cuda_graphs.py
jiemingz, line 1131: for instance what if you were to also filter  by `if _CudagraphGlobalRecord.tensor_reuse_pool.owns(arg): return`

megatron/core/transformer/cuda_graphs.py
mathemakitten: Tested this out and it works to filter out only the tensors owned by the reuse pool. See `replace_with_weak_ref_for_input_surface`. I made a new function instead of adding the check to the existing `replace_with_weak_ref` to avoid the extra check on surfaces where it isn't needed.

PR #3981, opened 2026-03-21
Fix `RerunStateMachine` crash (`TypeError: 'NoneType' object is not subscriptable`) by not saving a checkpoint after a transient NaN / Inf

### LINKED ISSUE #3884: RerunStateMachine crashes on resume from RERUNNING_IN_PLACE checkpoint (exit code 17)
RerunStateMachine crashes with TypeError: 'NoneType' object is not subscriptable at rerun_state_machine.py:1007 when a job is resumed from a checkpoint that was saved with RERUNNING_IN_PLACE state (exit code 17).

## Steps/Code to reproduce bug

  1. Enable RerunStateMachine with fatal=True and comparison > tolerance (e.g. spiky loss detection)
  2. A validation failure triggers an in-place rerun; the rerun also fails (transient error detected)
  3. Job exits with EXIT_CODE_FAILED_ON_RESULT_VALIDATION (17) and saves a checkpoint with state RERUNNING_IN_PLACE
  4. Scheduler resumes the job from that checkpoint (ignoring exit code)
  5. load_state_dict() restores self.state = RERUNNING_IN_PLACE but self.saved_state remains None (not persisted in checkpoint by design)
  6. First call to should_run_forward_backward() enters the RERUNNING_IN_PLACE branch and calls _restore_state() → crash

## Expected behavior

  Resuming from a RERUNNING_IN_PLACE checkpoint should not crash. Since exit code 17 semantically means "transient error detected, checkpoint saved for forensic purposes only", the state machine should reset to NOT_RUNNING_YET on resume and continue training normally.

## Additional context

A simpler fix of adding a None guard in _restore_state() is not reliable when --no-load-rng is set: in that case the RNG state is intentionally not restored from checkpoint, so skipping _restore_state() silently leaves the RNG in an incorrect state, making any subsequent rerun result untrustworthy. The fix in load_state_dict() avoids the issue entirely regardless of --no-load-rng.

yezhengmao1:
# What does this PR do ?

Fix `RerunStateMachine` crash (`TypeError: 'NoneType' object is not subscriptable`) when a job is resumed from a checkpoint saved with `RERUNNING_IN_PLACE` state (exit code 17). Fixes #3884.

## Root cause

The path that produces a `RERUNNING_IN_PLACE` checkpoint is a **transient NaN/Inf**:

1. `INITIAL_RUN`: NaN/Inf detected (`fatal=True`) → `rerun_requested = True`
2. `RERUNNING_IN_PLACE`: re-run passes (transient) → `comparison > tolerance`, but `fatal=True` means neither `continue_requested` nor `checkpoint_requested` is set → state stays `RERUNNING_IN_PLACE`
3. `should_checkpoint_and_exit()` saves a forensic checkpoint and exits with code 17
4. Scheduler resumes from that checkpoint

On resume, `load_state_dict` restores `self.state = RERUNNING_IN_PLACE`, but `saved_state` (an in-memory RNG snapshot that is intentionally not persisted to disk) is `None`. The first call to `should_run_forward_backward()` enters the `RERUNNING_IN_PLACE` branch and calls `_restore_state()`, which crashes trying to subscript `None`.

A simple `None` guard in `_restore_state()` is not sufficient: if `--no-load-rng` is set, the RNG would not be restored from checkpoint, making any subsequent rerun result untrustworthy even if the crash were suppressed.

## Fix

**`megatron/training/checkpointing.py`** — move rerun state loading to *after* RNG state restoration. This ensures the checkpoint RNG is already live when `load_state_dict` is called. The reordering is safe because the two operations are independent.

**`megatron/core/rerun_state_machine.py`** — in `load_state_dict`, when the restored state is `RERUNNING_IN_PLACE`, immediately call `_save_state()` to capture the just-restored checkpoint RNG into `saved_state`. This makes `_restore_state()` correct when `should_run_forward_backward()` is next called.

### Pre-checks


dimapihtar (approved):
LGTM. Thank you!

cyme:
I think the safest and simplest fix is to simply skip the creation of a forensic checkpoint in case of transient errors. It seemed like a good idea when writing the original code. But the actual value of these checkpoints is unclear to me, particularly given how infrequent these errors are. Additionally the state RERUNNING_IN_PLACE is not designed to be resumable, so we would need to verify the state machine does what we want.

megatron/core/rerun_state_machine.py
cyme: That's the fix I'm suggesting to skip the creation of a forensic checkpoint in the `RERUNNING_IN_PLACE` state.

```suggestion
            return False, True, EXIT_CODE_FAILED_ON_RESULT_VALIDATION
```

cyme: Update the log message to reflect that there is no forensic checkpoint anymore

```suggestion
                "Exiting now. The job can be resumed from a previous checkpoint"
```

yezhengmao1:
> I think the safest and simplest fix is to simply skip the creation of a forensic checkpoint in case of transient errors. It seemed like a good idea when writing the original code. But the actual value of these checkpoints is unclear to me, particularly given how infrequent these errors are. Additionally the state RERUNNING_IN_PLACE is not designed to be resumable, so we would need to verify the state machine does what we want.

 Thanks for the review! I think the forensic checkpoint in the RERUNNING_IN_PLACE state is still worth keeping.

  The transient error (not reproducible on rerun) actually confirms that the checkpoint state itself is correct — the NaN was caused by an SDC during computation, not by a corrupted checkpoint. Saving it allows the scheduler to resume training from this point rather
  than falling back to an earlier checkpoint and redoing the work. On large-scale clusters, transient SDC errors are quite common, so avoiding unnecessary recomputation matters.

  Also, I don't think the current behavior of saving the checkpoint here is unintentional — the RERUNNING_IN_PLACE case in should_checkpoint_and_exit() explicitly returns (True, True, ...). The issue is just that the resume path wasn't handled, which is what this PR
  fixes.

  Regarding saved_state not being persisted (line 927) — that's not because RERUNNING_IN_PLACE is designed to be non-resumable. The RNG state is already captured in the checkpoint itself during save_checkpoint(), so there's no need to persist saved_state redundantly.
  The fix in this PR simply recaptures it from the checkpoint RNG on resume via _save_state() in load_state_dict.

cyme:
Hi @yezhengmao1 , thanks for your feedback.

In the original code (which I wrote) that saves a forensic checkpoint after a transient error, not much thought was given to the usefulness or even integrity of that forensic checkpoint. The state `RERUNNING_IN_PLACE` after a transient error is supposed to be a terminal state and was not designed to be resumable. Looking more closely, I see multiple issues:
- the forensic checkpoint effectively skips the iteration, causing the data stream and iteration count to be out of sync (data iterators are not rewinded before taking the checkpoint).
- upon resuming the checkpoint in the RERUNNING_IN_PLACE, I believe it will attempt to revalidate the results.

All these issues can be fixed, at the cost of making the code more complex. I do understand the incremental value of resuming from a more recent checkpoint. But my assessment is that these events are not frequent enough, even in large clusters, to justify the additional complexity. Hence my suggestion that the preferred path is to simply skip the forensic checkpoint.

Do you agree?

yezhengmao1:
Hi, thanks for the detailed explanation.

I didn't notice the issue about data stream and iteration count being out of sync (I modified the dataloader to have a separate state dict). I agree with your suggestion to skip the forensic checkpoint and not save it.

Additionally, I want to mention that in our large cluster, these transient errors occur quite frequently. The current issue is that after removing the faulty nodes, the system cannot successfully recover automatically and requires manual intervention, which is quite troublesome for us.

cyme:
Thanks! Are you planning to modify your MR to skip the checkpoint? The changes I suggested should do just that.

Regarding the need in your environment for manual intervention in case of transient errors, does the mitigation proposed address the problem (skip the forensic checkpoint and resume instead from an earlier, regular periodic checkpoint)?

yezhengmao1:
@cyme Yes, I've modified it to skip saving the checkpoint. The proposed mitigation does address the issue, but it seems that resuming from an earlier point will waste some time. Perhaps we need a better fault tolerance mechanism?

cyme:
> Perhaps we need a better fault tolerance mechanism?

Do you have something specific in mind?

deepakn94 (approved):

ericharper (approved):

PR #3995, opened 2026-03-23
Fix config compatibility with Megatron-Core

maanug-nv:
# What does this PR do ?

Match config attribute names with Megatron-Bridge.

Creates a mismatch between some argument names and the corresponding dataclasse attributes, through overriding `dest`. However, this mismatch can be resolved when configs are being used in the training loop code. 
Mismatch shouldn't affect usage of args. 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/training/config/common_config.py
+++ b/megatron/training/config/common_config.py
@@ -85,7 +85,7 @@ class DistributedInitConfig:
     local_rank: int = field(default_factory=lambda: int(os.getenv("LOCAL_RANK", "0")))
     """local rank passed from distributed launcher."""
 
-    lazy_init: bool = field(default=False, metadata={"argparse_meta": {"arg_names": ["--lazy-mpu-init"], "dest": "lazy_mpu_init"}})
+    lazy_mpu_init: bool = False
     """If set to True, initialize_megatron() skips DDP initialization and returns function to complete it instead.
     Also turns on --use-cpu-initialization flag. This is for external DDP manager."""
 
--- a/megatron/training/config/common_config.py
+++ b/megatron/training/config/common_config.py
@@ -106,7 +106,7 @@ class DistributedInitConfig:
     """If set, distributed ranks initialize order is changed from tp-cp-ep-dp-pp to tp-cp-ep-pp-dp.
     """
 
-    use_gloo_process_groups: bool = field(default=True, metadata={"argparse_meta": {"arg_names": ["--disable-gloo-process-groups"], "dest": "enable_gloo_process_groups"}})
+    use_gloo_process_groups: bool = field(default=True, metadata={"argparse_meta": {"arg_names": ["--disable-gloo-process-groups"]}})
     """If enabled, create Gloo process groups for communications."""
 
     use_sharp: bool = False
--- a/megatron/training/config/common_config.py
+++ b/megatron/training/config/common_config.py
@@ -85,7 +85,7 @@ class DistributedInitConfig:
     local_rank: int = field(default_factory=lambda: int(os.getenv("LOCAL_RANK", "0")))
     """local rank passed from distributed launcher."""
 
-    lazy_mpu_init: bool = False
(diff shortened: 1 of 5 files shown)


asolergi-nv (commented):
Shouldn't we make these changes to Megatron-Bridge to adapt it to Megatron-LM, not the other way around?

maanug-nv:
> Shouldn't we make these changes to Megatron-Bridge to adapt it to Megatron-LM, not the other way around?

These config classes are not really being used yet in MLM, just to generate the arguments. Whereas in Megatron Bridge, they are already in use and would require changes in several places to support the rename (eg 'lazy_init' -> 'lazy_mpu_init'). It could also affect customer's scripts. This is why I'm making the change here.

megatron/training/config/common_config.py
```diff
@@ -85,7 +85,7 @@ class DistributedInitConfig:
     local_rank: int = field(default_factory=lambda: int(os.getenv("LOCAL_RANK", "0")))
     """local rank passed from distributed launcher."""
 
-    lazy_mpu_init: bool = False
+    lazy_init: bool = field(default=False, metadata={"argparse_meta": {"arg_names": ["--lazy-mpu-init"], "dest": "lazy_mpu_init"}})
```
jaredcasper: IIRC this was added in for the very old initial Nemo integration, maybe we should just deprecate/remove it? Do we know if its still needed?

If we keep it, lazy_init is a bit misleading... lazy_mpu_init is an old name, but at least conveys what is being initialized lazily... maybe `lazy_distributed_init` or something?

megatron/training/config/common_config.py
```diff
@@ -106,7 +106,7 @@ class DistributedInitConfig:
     """If set, distributed ranks initialize order is changed from tp-cp-ep-dp-pp to tp-cp-ep-pp-dp.
     """
 
-    enable_gloo_process_groups: bool = field(default=True, metadata={"argparse_meta": {"arg_names": ["--disable-gloo-process-groups"]}})
+    use_gloo_process_groups: bool = field(default=True, metadata={"argparse_meta": {"arg_names": ["--disable-gloo-process-groups"], "dest": "enable_gloo_process_groups"}})
```
jaredcasper: shouldn't "dest" be `use_gloo_process_groups`?

megatron/training/config/training_config.py
```diff
@@ -454,7 +454,7 @@ class CheckpointConfig:
     async_ckpt_io_priority: Optional[int] = 3
     """I/O scheduling class (0-3, 3=idle) for the async checkpoint writer process."""
 
-    ckpt_fully_parallel_load: bool = False
+    fully_parallel_load: bool = field(default=False, metadata={"argparse_meta": {"arg_names": ["--ckpt-fully-parallel-load"], "dest": "ckpt_fully_parallel_load"}})
```
jaredcasper, line 457: Almost all of the other checkpoint options have the ckpt_ prefix... and this makes this ambiguous if its talking about checkpoint load or data load. Why are we renaming this?

maanug-nv: All of these renames are to match up with Megatron Bridge. I'm trying to import these dataclasses in Megatron Bridge so that we don't define them twice, but the naming mismatches are causing issues. Not sure how they got mismatched in the first place, but my reasoning for renaming things here was that renaming in Bridge would require changing a lot of lines in recipes and tests, whereas here the config attributes aren't being used yet.
But I agree, we should stick with the most clear name.

maanug-nv: however this setting is in CheckpointConfig, I think that should make it clear that this is referring to checkpoint load and not data load or something else right? The argparse arg will still be '--ckpt-fully-parallel-save'. and for now, it will still be referenced with 'args.ckpt_fully_parallel_save'.

jaredcasper: ah right, forgot these were already scoped in a config class.

jaredcasper (approved):

yaoyu-33 (approved):

PR #4005, opened 2026-03-23
NVFP4 native weights for DDP

WanZzzzzz:
# What does this PR do ?

Add support for NVFP4 native (packed) weights in Megatron's Distributed Data Parallel (DDP) pipeline. This allows parameters to be stored, communicated, and gathered in NVFP4 format, reducing memory footprint and communication volume during training.
## Problem
Previously, NVFP4 quantization was only applied at the compute level via Transformer Engine. The DDP param buffers and all-gather operations still used full-precision (BF16) representations, missing the memory and bandwidth savings that native NVFP4 storage provides — similar to what FP8 native weights already support.
## Solution
Extend the existing FP8 native weight infrastructure to support NVFP4, accounting for NVFP4's unique dual-buffer layout: the param buffer stores packed bytes (half numel) while the grad buffer uses full numel.
Important notice: This PR only support 2D quantized nvfp4 weights. With 2D quantization, rowwise and colwise tensors share the same values. Given this, we only need to all-gather the rowwise data and transpose to get colwise data later.
### Core DDP infrastructure (`param_and_grad_buffer.py`)
- NVFP4 params use packed storage (`numel // 2`) in the param buffer but full numel in the grad buffer
- Added `nvfp4_grad_index_map` for separate grad buffer offset tracking
- Added `get_grad_index_map()` so the distributed optimizer uses correct (full-numel) indices for grad/param range calculations
- Bucket creation passes separate grad offsets for NVFP4 buffers
- `partition_buckets` updated to support `reduce_scatter_with_fp32_accumulation` (one bucket per group)
### Distributed optimizer (`distrib_optimizer.py`)
- Extended FP8 model param / main param initialization paths to also handle NVFP4
- Added `_get_nvfp4_params_and_shard_fp32_from_nvfp4()` to collect NVFP4 params with their FP32 master weight shards
- `_copy_main_params_to_model_params` calls `quantize_nvfp4_param_shard` to cast FP32 shards back to NVFP4 after optimizer step
### FP4 utilities (`fp4_utils.py`)
- `get_nvfp4_rowwise_packed_shape`: compute packed byte shape for param buffer
- `modify_nvfp4_rowwise_storage`: swap NVFP4 tensor's rowwise data pointer into the DDP param buffer
- `quantize_nvfp4_param_shard`: wraps TE's `quantize_master_weights` for shard-level FP32→NVFP4 casting
- `fp4_model_init` now passes `preserve_high_precision_init_val=True` during training to retain BF16 init values for accurate FP32 main param initialization
### Arguments & config
- Added `--fp4-param-gather` CLI flag (consistent with existing `--fp8-param-gather`)
- Added `fp4_param_gather` to `DistributedDataParallelConfig`
- Fixed `TransformerConfig.fp4_param` field definition to be consistent with `fp8_param`
- Added `fp4_param` to `ArgumentGroupFactory` exclude list to prevent duplicate argparse registration
### Tests (`test_fp4_param.py`)
- Training correctness: verified NVFP4 vs BF16 loss bitwise match (with/without DP overlap)
- Inference mode
- First/last layers in BF16
- Determinism helper for debugging
### Temporary aids for deterministic-mode debugging (to be removed before merge)
- Stochastic rounding disabled in NVFP4 recipe (`disable_stochastic_rounding=True`)
- `grad_norm` rounding in `ChainedOptimizer` (`round(grad_norm, 4)`)

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/core/distributed/param_and_grad_buffer.py
+++ b/megatron/core/distributed/param_and_grad_buffer.py
@@ -941,9 +941,9 @@ class _ParamAndGradBuffer:
             # NVFP4 params are packed (2 values per byte), so the param buffer uses
             # half the logical numel. Non-NVFP4 params use the full numel for both.
             if self.has_nvfp4_params and is_nvfp4tensor(param):
-                assert full_numel % 2 == 0, (
-                    f"NVFP4 requires even numel for packing, got {full_numel}"
-                )
+                assert (
+                    full_numel % 2 == 0
+                ), f"NVFP4 requires even numel for packing, got {full_numel}"
                 param_numel = full_numel // 2
             else:
                 param_numel = full_numel
@@ -964,7 +964,11 @@ class _ParamAndGradBuffer:
             if self.has_nvfp4_params:
                 grad_start_index = _pad_start_of_param_if_needed(grad_start_index)
                 grad_end_index = grad_start_index + full_numel
-                self.nvfp4_unpacked_param_index_map[param] = (grad_start_index, grad_end_index, bucket_id)
+                self.nvfp4_unpacked_param_index_map[param] = (
+                    grad_start_index,
+                    grad_end_index,
+                    bucket_id,
+                )
                 grad_start_index = grad_end_index
 
             # If we have enough elements already or the current param is part of the shared


megatron/training/arguments.py
```diff
@@ -889,8 +889,13 @@ def validate_args(args, defaults={}):
                 'all_gather instead, turning off fp8_param_gather',
                 args.rank,
             )
-        if args.fp4_param and not is_te_min_version("2.7.0.dev0"):
-            raise ValueError("--fp4-param requires Transformer Engine >= 2.7.0.dev0.")
+        if args.fp4_param_gather and not is_te_min_version("2.7.0.dev0"):
+            args.fp4_param_gather = False
```
gautham-kollu, line 893: Can we make this the default behavior whenever fp4 is enabled ?
Asking if this argument is necessary ?  (may be debug purposes)

gautham-kollu:
Thanks for the detailed description

megatron/training/arguments.py
WanZzzzzz: We should make fp4 and fp4_param_gather separate (like what we did for fp8 and fp8_param_gather.) Reasons are:
1. They do different things: fp4 reduces precision in compute, while fp4_param_gather reduces precision in memory and communication.
2. When we do fp8/fp4 gather, the quantization happens in the optimizer rather than fwd/bwd. Without CG optimizer, performance overheads are expected. So we should let users decide whether to turn it on.

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -987,33 +1023,70 @@ def _does_param_require_new_bucket(param):
...
                 bucket_end_index = _pad_end_of_bucket_if_needed(param_start_index)
-                self.buckets.append(
-                    self._new_bucket(
-                        bucket_params=bucket_params,
-                        start_index=bucket_start_index,
-                        end_index=bucket_end_index,
-                        numel_unpadded=per_bucket_numel_unpadded[cur_bucket_id],
-                        bucket_id=cur_bucket_id,
+                # For NVFP4, compute grad bucket boundaries separately
+                if self.has_nvfp4_params:
```
jaredcasper, line 1149: I think this would be cleaner to set grad_{start,end}_index in the if/else then just have one append/new call. (same for below)

WanZzzzzz: Thanks. Simplified.

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -905,10 +906,24 @@ def _does_param_require_new_bucket(param):
                 and self.ddp_config.use_distributed_optimizer
             )
 
+        # Check if this buffer contains NVFP4 params
+        self.has_nvfp4_params = any(is_nvfp4tensor(p) for p in self.params)
+        # For NVFP4, we need separate tracking for param buffer (packed) and grad buffer (full)
```
deepakn94: Can you add some ASCII art describing this in more detail?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -905,10 +906,24 @@ def _does_param_require_new_bucket(param):
...
+
         for param, _ in params_with_names[::-1]:
             # Iterate through parameters in reverse order to roughly follow backprop order.
 
-            this_numel = param.data.nelement()
+            # Use full numel for grad buffer allocation
+            full_numel = param.data.nelement()
+            # For NVFP4 params, use packed numel for param buffer to avoid zeros.
+            if self.has_nvfp4_params and is_nvfp4tensor(param):
+                # NVFP4: use packed numel (half of logical numel) for param buffer
```
deepakn94: Is `this_numel` a good variable name for this?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -921,6 +936,13 @@ def _does_param_require_new_bucket(param):
             self.param_index_map[param] = (param_start_index, param_end_index, bucket_id)
             bucket_params.add(param)
 
+            # For NVFP4, track separate grad buffer offsets (using full numel)
```
deepakn94: Why?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -939,6 +961,18 @@ def _does_param_require_new_bucket(param):
         # padding as necessary).
         self.numel = bucket_end_index
         self.numel_unpadded = sum(per_bucket_numel_unpadded)
+
+        # For NVFP4, grad buffer needs full size (2x packed size)
+        if self.has_nvfp4_params:
+            # Pad grad_numel to be divisible by dp_size
+            self.grad_numel = grad_start_index
+            if self.ddp_config.use_distributed_optimizer:
+                remainder = self.grad_numel % self.data_parallel_world_size
```
deepakn94: Use existing padding helper functions?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1023,23 +1059,41 @@ def _does_param_require_new_bucket(param):
...
-                    if is_float8tensor(param):
+                    if is_nvfp4tensor(param):
+                        # NVFP4 params: map rowwise packed bytes (uint8) into the param buffer.
+                        from ..fp4_utils import modify_nvfp4_rowwise_storage
+
+                        packed_shape = get_nvfp4_rowwise_packed_shape(param.data.shape)
+                        rowwise_bytes_view = self._get(
+                            packed_shape, param_start_index, buffer_type=BufferType.PARAM
+                        )
+                        modify_nvfp4_rowwise_storage(param, rowwise_bytes_view)
```
deepakn94, line 1097: What is this doing?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1326,8 +1431,32 @@ def partition_buckets(
         fp8_buffer = dtype_to_buffer_map[torch.uint8]
         for bucket in fp8_buffer.buckets:
             if len(bucket_groups) == len(fp8_buffer.buckets) - 1:
-                # The last bucket group.
-                group_buckets = [bucket] + non_fp8_buckets
+                # When using reduce-scatter with FP32 accumulation,
+                # we can only have one bucket per group.
+                if reduce_scatter_with_fp32_accumulation:
```
deepakn94, line 1460: Why do we need this?

deepakn94 (commented):
More comments needed to describe tricky code. Some meta comment somewhere describing how this works (and how it differs from the FP8 version) also useful.

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Done.

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Changed to param_numel

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Added a comment explaining: the grad buffer is full-sized, so we maintain a parallel index walk using full_numel for every param to keep the grad offsets correct.

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Replaced the manual one with the existing helper function.

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Added a comment explaining: this remaps the NVFP4 tensor's internal rowwise uint8 storage pointer to point into the contiguous DDP param buffer, enabling all-gather to communicate packed NVFP4 bytes directly (analogous to how FP8 params use modify_underlying_storage).

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Added a comment explaining: reduce_scatter_with_fp32_accumulation requires exactly one bucket per group (enforced by an assert in _ParamAndGradBucketGroup.reduce_scatter). Without this branch, the non-FP8 buckets would be merged into the last FP8 group, violating that single-bucket constraint. So they're split into individual groups instead.

gautham-kollu:
@WanZzzzzz  I could not find any fp4 functional tests in MLM.  My request is add at least fp4 functional test that runs as part of each PR.  (essentially to protect against regressions). 

We can just use something like
Megatron-LM/tests/functional_tests/test_cases/moe/gpt3_mcore_te_tp2_pp1_te_8experts2parallel_top2router/model_config.yaml. as the reference and just update the precision and we can use that to update all fp4 related perf configs.

WanZzzzzz:
/ok to test [628a9fe](https://github.com/NVIDIA/Megatron-LM/pull/4005/commits/628a9fed767fb8849d2674fe3837bf232b9ff795)

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -909,10 +910,39 @@ def _does_param_require_new_bucket(param):
...
             )
 
+        # Check if this buffer contains NVFP4 params.
+        #
+        # NVFP4 uses a dual-buffer layout: the param buffer stores packed bytes (half the
+        # logical numel) while the grad buffer uses the full numel. This is because NVFP4
+        # packs two FP4 values into a single uint8 byte for storage/communication, but
+        # gradients are computed and reduced in BF16 at full element count.
+        #
+        #   Logical param (BF16 view):  [v0, v1, v2, v3, ...]   numel = N
```
deepakn94: Is "(BF16 view)" important here?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -909,10 +910,39 @@ def _does_param_require_new_bucket(param):
...
+        #
+        #   Param buffer  (uint8):      [byte0, byte1, ...]      numel = N // 2
+        #                                 ^^^^^ packs v0+v1
+        #
+        #   Grad buffer   (BF16):       [g0, g1, g2, g3, ...]   numel = N
```
deepakn94: Why BF16? This is often in FP32. Maybe remove this?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -909,10 +910,39 @@ def _does_param_require_new_bucket(param):
...
+
         for param, _ in params_with_names[::-1]:
             # Iterate through parameters in reverse order to roughly follow backprop order.
 
-            this_numel = param.data.nelement()
+            full_numel = param.data.nelement()
+            # NVFP4 params are packed (2 values per byte), so the param buffer uses
+            # half the logical numel. Non-NVFP4 params use the full numel for both.
+            if self.has_nvfp4_params and is_nvfp4tensor(param):
+                param_numel = full_numel // 2
```
deepakn94, line 947: Do we have to be careful about integer rounding here? What if `full_numel` is odd?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1069,16 +1138,26 @@ def _does_param_require_new_bucket(param):
 
             if bucket_id != cur_bucket_id:
                 bucket_end_index = _pad_end_of_bucket_if_needed(param_start_index)
+                # For NVFP4, compute grad bucket boundaries separately
+                if self.has_nvfp4_params:
+                    grad_bucket_end_index = _pad_end_of_bucket_if_needed(grad_start)
+                else:
+                    grad_bucket_start_index = None
```
deepakn94: Do we need this else condition? `grad_bucket_start_index` and `grad_bucket_end_index` should not be used if we don't have native NVFP4 parameters?

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -2412,6 +2417,46 @@ def get_shard_fp32_from_fp8(shard_main_groups, model_groups):
...
+        for buffer in buffers:
+            for param in buffer.params:
+                if is_nvfp4tensor(param):
+                    nvfp4_params.append(param)
+                    shard_fp32_from_nvfp4.append(None)
+                    shard_offsets_in_nvfp4.append(None)
+                    nvfp4_param_to_idx_map[param] = idx
+                    idx += 1
+
+        def get_shard_fp32_from_nvfp4(shard_main_groups, model_groups):
```
deepakn94: Docstring? Add underscore at the start since it's an internal method?

megatron/core/distributed/param_and_grad_buffer.py
deepakn94: If not, delete? And make "Logical param" "Logical view"?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1134,17 +1219,33 @@ def scale_gradients(self, scaling_factor: float) -> None:
         for grad in self.extra_main_grads:
             grad *= scaling_factor
 
+    def get_grad_index_map(self) -> Dict[torch.nn.Parameter, tuple[int, int, int]]:
```
deepakn94: Should we make this more generic? Like "get_unpacked_index_map"? Gradient doesn't seem important here.

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Makes sense. Done.

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Removed

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: For nvfp4 params, we can only pack even number of elements. Added an assertion to catch odd-sized buffer with error message.

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Makes sense. Deleted them and also initialized grad_bucket_end_index to be none at the beginning otherwise _new_bucket will raise a NameError.

megatron/core/optimizer/distrib_optimizer.py
WanZzzzzz: Done.

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Makes sense. Done.

WanZzzzzz:
/ok to test [7d5aa89](https://github.com/NVIDIA/Megatron-LM/pull/4005/commits/7d5aa899deb66842b85d63764bca9474a79b494b)

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1134,17 +1216,34 @@ def scale_gradients(self, scaling_factor: float) -> None:
...
+        For NVFP4 buffers, param_index_map uses packed numel (half the logical size),
+        so this returns nvfp4_grad_index_map which has full-numel indices instead.
+        For other buffers, param and grad indices are identical, so param_index_map
+        is returned directly.
+
+        The distributed optimizer uses this to determine which rank owns which portion
+        of each parameter's data.
+        """
+        if self.has_nvfp4_params:
+            return self.nvfp4_grad_index_map
```
deepakn94: Let' re-name `nvfp4_grad_index_map` to `nvfp4_unpacked_index_map`?

WanZzzzzz: sounds good. Done.

WanZzzzzz:
/ok to test [51ad2bf](https://github.com/NVIDIA/Megatron-LM/pull/4005/commits/51ad2bf9c9ba33374e6000a7bee3051bce11562f)

Phlip79:
The label "Approved" is being added by mistake due to a bug. This will be fixed in https://github.com/NVIDIA/Megatron-LM/pull/4280.

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -909,10 +910,43 @@ def _does_param_require_new_bucket(param):
...
+        #
+        #   Logical view:  [v0, v1, v2, v3, ...]   numel = N
+        #
+        #   Param buffer  (uint8):      [byte0, byte1, ...]      numel = N // 2
+        #                                 ^^^^^ packs v0+v1
+        #
+        #   Grad buffer:  [g0, g1, g2, g3, ...]   numel = N
+        #
+        # We therefore maintain two index maps:
+        #   - param_index_map:        offsets into the packed param buffer  (numel // 2)
```
deepakn94, line 928: Thinking about this more, I'm wondering if we should name the two variables `param_index_map` and `packed_param_index_map`. `param_index_map` has the vanilla mapping, and `packed_param_index_map` has the offsets in the packed param buffer.

megatron/core/distributed/param_and_grad_buffer.py
deepakn94: I just left another comment about variable naming. If you agree with that comment, maybe this method should be re-named to `get_packed_param_index_map` to unify.

megatron/core/distributed/param_and_grad_buffer.py
WanZzzzzz: Do we have to make this so general? Packing two values into one byte is specific to nvfp4 case, not fp6 or fp2 even if we have it in the future. If we do the renaming, the current param_index_map will become packed_param_index_map and nvfp4_grad_index_map will become param_index_map, which requires a big refactoring of this PR.

deepakn94: Well, the other way to think about this is we are keeping `param_index_map` un-changed from before. And adding a new `packed_param_index_map` or `nvfp4_packed_param_index_map` (or something else) to handle the specific NVFP4 case.

deepakn94: Basically, I don't like the concept of a `grad_index_map`. Everything is just a `param_index_map`. And if a parameter takes less than the smallest data type we have available, then we have to do something special (and that's the special case).

WanZzzzzz: I renamed nvfp4_grad_index_map to nvfp4_unpacked_param_index_map, which serves the purpose here and avoids changing the logic.

WanZzzzzz:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4005/commits/4cebfe8e85ce1d5f804d3476c570586667a87762

WanZzzzzz:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4005/commits/2b97a3404488d538b368b8a646a9231afd0db5f4

megatron/core/distributed/param_and_grad_buffer.py
deepakn94: Sorry, I don't love this...you are changing what `param_index_map` means when NVFP4 is used. `nvfp4_unpacked_param_index_map` now does what `param_index_map` does for BF16 or MXFP8 training, and `param_index_map` stores different things based on the precision.

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -909,10 +910,43 @@ def _does_param_require_new_bucket(param):
...
+        #   - nvfp4_unpacked_param_index_map: offsets using full (unpacked) numel
+        #
+        self.has_nvfp4_params = any(is_nvfp4tensor(p) for p in self.params)
+        self.nvfp4_unpacked_param_index_map = {}
+        grad_start_index = 0 if self.has_nvfp4_params else None
```
deepakn94, line 933: There is no `grad_index_map` now. Update these variables to mirror the name of the index map (`self.nvfp4_unpacked_param_index_map`).

WanZzzzzz: grad_start_index and grad_end_index are used to get the index of grad data in a grad bucket. I don't think we should rename them to unpacked_param_start_index or something.

Phlip79:
This PR is actually ready to merge now (i.e. the "Approved" label is accurate).

PR #2812, opened 2026-01-05
Nccl gloo refit for RL

wdykas:
# What does this PR do ?
This is separating out just the GLOO and NCCL refit methods from https://github.com/NVIDIA/Megatron-LM/pull/2696 for easier merging.

This allows rl training with different inference and training parallelisms. The fast path NVshmem backend will be added after this goes in.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @megatron-oncall.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
        C1[Expert Review] --> C2[Final Review]
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [ ] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

The following process is enforced via the CODEOWNERS file for changes into `megatron/core`. For changes outside of `megatron/core`, it is up to the PR author whether or not to tag the Final Reviewer team.

<details>
<summary>For MRs into `main` branch</summary>

Feel free to message or comment the @megatron-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

#### (Step 2): Collect the expert reviewers reviews

1. Attach the `Expert Review` label when your PR is ready for review.
2. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

:warning: Only proceed to the next step once all reviewers have approved, merge-conflict are resolved and the CI is passing.  
Final Review might get declined if these requirements are not fulfilled.

#### (Step 3): Final Review

1. Add `Final Review` label
2. GitHub auto-assigns final reviewers based on your changes. They will get notified and pick up your PR soon.

#### (Optional Step 4): Cherry-pick into release branch

If this PR also needs to be merged into `core_r*` release branches, after this PR has been merged, select `Cherry-pick` to open a new PR into the release branch.

</details>

<details>
<summary>For MRs into `dev` branch</summary>
[...]

PR #4036, opened 2026-03-26
Add stacklevel to warnings.warn() calls across megatron/core

Bias92:
### Pre-checks

added `stacklevel` parameters to `warning.warn()` calls across `megatron/core/`.
without stacklevel, warnings point to the internal warn() call site instaed of the caller's code, making it difficult for users to trace where they used a deprecated API.
spotted from @Skylion007 's review comment on #2812  only that one call was fixed. this pr addresses the remaning calls across the codebase.

### Summary
- 31 files, ~60 calls updated
- `stacklevel=2` for calls inside functions/methods so the warning points at the caller
- `stacklevel=3` kept for `unified_memory.compile_allocator` (warns through one wrapper)
- added `DeprecationWarning` category to 22 deprecation messages missing it
- left at the default on purpose:
  - module-level import fallbacks: stacklevel=2 there blames whichever module imports the file first, not useful
  - dataclass `__post_init__`: it runs under the generated `__init__` whose frame shows up as `<string>`, so stacklevel=2 actually erases the location. subclass `super().__post_init__()` chains make =3 wrong too. this also drops the recently added stacklevel=2 from the TE cross entropy warn in `ModelParallelConfig` for the same reason
- unit test: `tests/unit_tests/test_warnings_stacklevel.py`, ast based, checks that function-level warns pass stacklevel. pre-existing explicit levels in files this pr doesn't touch are left alone

### Before / After
warning when user code passes the deprecated `inference_parameters=` to `Scheduler.add_request()`:

before (points at megatron internals):
```
.../megatron/core/inference/scheduler.py:81: DeprecationWarning: `inference_parameters` has been renamed to `sampling_params`, and the previous name will be removed in `megatron-core` 0.13.
```

after (points at the caller):
```
/workspace/my_inference_script.py:42: DeprecationWarning: `inference_parameters` has been renamed to `sampling_params`, and the previous name will be removed in `megatron-core` 0.13.
```


yaox12 (approved):

megatron/core/models/common/embeddings/rope_utils.py
```diff
@@ -273,13 +273,15 @@ def apply_rotary_pos_emb(
                 # TODO: Add a check in TransformerConfig and remove this unfused implementation.
                 warnings.warn(
                     "apply_rope_fusion does not support mRoPE in bshd format when bs > 1. "
-                    "Please set apply_rope_fusion to false. This will become an error in v0.16."
+                    "Please set apply_rope_fusion to false. This will become an error in v0.16.",
+                    stacklevel=3,
```
gautham-kollu: why a stacklevel 3 here ?

gautham-kollu:
@Bias92  is there  a way to make the default level to "2".  If not specified, it would just use that ?

Also can you add the log lines before/after this PR for easy testing? and a unit test would be great.

megatron/core/models/common/embeddings/rope_utils.py
Bias92: ahh sorry. it was wrong. the warn() is directly inside `apply_rotary_pos_emb`, not in a nested helper. should be `stacklevel=2`. i will fix this.

Bias92:
> @Bias92 is there a way to make the default level to "2". If not specified, it would just use that ?
> 
> Also can you add the log lines before/after this PR for easy testing? and a unit test would be great.

python's warnings.warn() defaults stacklevel to 1, no way to change that globally. a wrapper that defaults to 2 is possible but would be a separate change.
also fixed rope_utils.py stacklevel=3 → stacklevel=2, thanks for catching that.
will look into adding before/after output and a unit test.

megatron/core/models/common/embeddings/rope_utils.py
Bias92: fixed, changed to stacklevel=2. the warn() is directly inside apply_rotary_pos_emb, not a nested helper.

ericharper:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4036/commits/cdabde521a578e720e9e89b531acdbbc46a7553c

Bias92:
updated the branch:
- merged latest main, fixed the black failures from the last run
- moved the unit test under `tests/unit_tests/` (a root-level file is picked up by neither lint nor unit test ci) and rewrote it with ast
- revisited stacklevel placement: module-level and dataclass `__post_init__` warns now keep the default, reasoning is in the updated description
- force-pushed once to add the missing DCO sign-offs to the three older commits (the DCO check is new since this pr was opened), content unchanged

@ericharper could you re-run `/ok to test` for the new head (01c6c397f9328026ec4714a734faecf42da3d17e) and re-enable auto-merge when you get a chance? the branch update invalidated the previous vetting. thanks!

dimapihtar (approved):

cspades (approved):

Skylion007:
Enable ruff linter for this in the codebase: https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/

PR #4046, opened 2026-03-28
Fix HybridDeviceOptimizer KeyError after mixed-precision param replacement

### LINKED ISSUE #4042: [Bug] HybridDeviceOptimizer KeyError with MoE models when using CPU offload
## Problem

When training MoE models (e.g., Qwen3.5-35B-A3B) with optimizer_cpu_offload=True, HybridDeviceOptimizer raises a KeyError at the first optimizer step.

## Error

```
File "hybrid_optimizer.py", line 160, in step
    self._sync_hdo_param_groups_to_sub_optimizers()
File "hybrid_optimizer.py", line 339, in _sync_hdo_param_groups_to_sub_optimizers
    inner_param = self.param_to_inner_param[param]
KeyError: tensor([...], device='cuda:6')
```

## Cause

param_to_inner_param is built in _init_sub_optimizers using tensor object identity as dict keys. After initialization, the outer optimizer wrapper (e.g., MixedPrecisionOptimizer) may replace tensor objects in self.param_groups, making the original mapping stale.

## Config

- Model: Qwen3.5-35B-A3B (MoE, 64 experts)
- TP=2, PP=4, EP=2, bf16
- optimizer_cpu_offload=True, optimizer_offload_fraction=1.0

## Suggested Fix

Use a stable key (e.g., parameter index) instead of tensor identity, or rebuild the mapping before each step().

ma-ben:
## Summary

This fixes the `HybridDeviceOptimizer` `KeyError` that happens on the first `optimizer.step()` when CPU optimizer offload is used together with bf16/fp16 mixed precision.

## Root Cause

`HybridDeviceOptimizer` built its internal mapping using tensor object identity from the original `param_groups`.

Later, `Float16OptimizerWithFloat16Params` replaces entries in `optimizer.param_groups` with new FP32 main parameters. On the first step, HDO still looks up the old tensor objects and fails with:

```python
inner_param = self.param_to_inner_param[param]
KeyError: tensor(..., device='cuda:X')
```

## Fix

Use a stable logical mapping based on parameter position in `param_groups` instead of relying on tensor object identity remaining unchanged.

This also updates:
- grad sync from current params to inner/sub-optimizer params
- copy-back from sub-optimizers to current params
- HDO state synchronization to follow the current wrapped params

## Tests

Added a single-GPU regression test covering:
- `HybridDeviceOptimizer + Float16OptimizerWithFloat16Params`
- `bf16`
- `fp16`
- first `optimizer.step()` no longer raises `KeyError`

Validation run:
```bash
OMP_NUM_THREADS=1 .venv/bin/python -m pytest tests/unit_tests/test_optimizer_cpu_offloading.py -vv
```

Fixes #4042

### DIFF
--- a/megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py
+++ b/megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py
@@ -80,26 +80,35 @@ class HybridDeviceOptimizer(torch.optim.Optimizer):
         self._init_sub_optimizers()
         self._register_load_state_dict_hooks()
 
+    def _iter_current_param_group_entries(self):
+        for group_id, group in enumerate(self.param_groups):
+            for param_id, param in enumerate(group["params"]):
+                yield (group_id, param_id), param
+
+    def _get_current_param(self, param_group_index):
+        group_id, param_id = param_group_index
+        return self.param_groups[group_id]["params"][param_id]
+
     def _set_sub_optimizer_grads(self):
         if self.param_update_in_fp32:
-            for param in self.param_to_fp32_param:
-                if param in self.gpu_params_map_cpu_copy:
-                    # Skip if the param is offloaded to CPU, it should be handled
-                    # in the following part.
+            for param_group_index, param in self._iter_current_param_group_entries():
+                inner_param = self.param_group_index_to_inner_param[param_group_index]
+                if not inner_param.is_cuda:
+                    # Offloaded params are handled in the following part.
                     continue
-                fp32_param = self.param_to_fp32_param[param]
                 grad = getattr(param, "decoupled_grad", param.grad)
                 if grad is not None:
-                    fp32_param.grad = grad.to(fp32_param.dtype)
-                    fp32_param.requires_grad = True
+                    inner_param.grad = grad.to(inner_param.dtype)
+                    inner_param.requires_grad = True
                 else:
-                    fp32_param.requires_grad = False
+                    inner_param.requires_grad = False
 
         # Sync the grads from GPU to CPU.
         for optimizer in self.cpu_optimizers:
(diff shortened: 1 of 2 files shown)


shjwudp (commented):
Hi @ma-ben , thanks a lot for catching this issue and for the careful investigation and fix. 

I agree that there are currently composability issues between `HybridDeviceOptimizer` and other optimizers. I’ve been thinking about how to address this in a more systematic way.

First, I don’t think `param_update_in_fp32` is a good option in HDO, as it essentially introduces hard-coded behavior.

From a design perspective:

- `HybridDeviceOptimizer` (optimizer offloading) is responsible for offloading optimizer states and gradients to CPU to reduce GPU memory usage, while leveraging a side CPU optimizer to perform the optimizer step.
- `Float16OptimizerWithFloat16Params` is responsible for maintaining FP32 copies of parameters outside the optimizer to improve update precision.

If we want to make optimizer offloading more general and composable with different optimizers, I think we may need to:

1. Apply optimizer offloading at the final stage, so that as many optimizer states as possible (e.g., FP32 parameters introduced by `Float16OptimizerWithFloat16Params`) can be offloaded to CPU.
2. Remove `param_update_in_fp32` from HDO, and make the CPU optimizer class optional, so that HDO can still provide value (i.e., optimizer state offloading) even when a CPU optimizer is not available.

Based on these considerations, I feel that instead of adding special-case compatibility in `HybridDeviceOptimizer` for `Float16OptimizerWithFloat16Params`, it may be better to rethink the relative positioning of `HybridDeviceOptimizer` in the optimizer stack.

This likely requires some refactoring of HDO and may not be trivial to implement immediately. I don’t yet have a complete design in mind, but I’d be happy to discuss this further if you’re interested.

PR #4062, opened 2026-03-30
Preserve type of decorated methods/classes

nschank:
# What does this PR do ?

Use a TypeVar to ensure that decorators don't obfuscate the structure of the thing they're decorating.

Before this change, the type-checker and IDEs would treat any functions/classes decorated by these decorators as completely opaque (no known signature/methods/anything); now, these decorators are transparent and the type-checker knows the underlying class/type is unchanged.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/core/utils.py
+++ b/megatron/core/utils.py
@@ -83,6 +83,10 @@ _mamba_ssm_version = None
 _causal_conv1d_version = None
 
 
+_Wrapped = TypeVar('_Wrapped', bound=Callable)
+"""A function or class which has been wrapped by a decorator."""
+
+
 @contextmanager
 def null_decorator(*args, **kwargs):
     """
@@ -120,7 +124,7 @@ def experimental_fn(introduced_with_version: str):
     """
     logged_functions = set()
 
-    def validator(func: Callable, max_lifetime: int = 3) -> Callable:
+    def validator(func: _Wrapped, max_lifetime: int = 3) -> _Wrapped:
         """Validates the request to the experimental function.
 
         Args:
@@ -186,7 +190,7 @@ def experimental_cls(introduced_with_version: str):
     """
     logged_classes = set()
 
-    def validator(cls: Callable, max_lifetime: int = 3) -> Callable:
+    def validator(cls: _Wrapped, max_lifetime: int = 3) -> _Wrapped:
         """Validates the request to the experimental function.
 
         Args:
@@ -2209,7 +2213,9 @@ def _nvtx_decorator_get_func_path(func):
     return f"{module.__name__}.{caller_func}"
 
 
-def nvtx_decorator(message: Optional[str] = None, color: Optional[str] = None):
+def nvtx_decorator(
+    message: Optional[str] = None, color: Optional[str] = None
+) -> Callable[[_Wrapped], _Wrapped]:
     """Decorator to add NVTX range to a function.
 
     Args:
@@ -2229,7 +2235,7 @@ def nvtx_decorator(message: Optional[str] = None, color: Optional[str] = None):
             pass
     """
 
-    def decorator(func: Callable) -> Callable:
+    def decorator(func: _Wrapped) -> _Wrapped:
         if _nvtx_enabled:
             return nvtx.annotate(
                 message=message or _nvtx_decorator_get_func_path(func), color=color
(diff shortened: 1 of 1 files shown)


megatron/core/utils.py
```diff
@@ -2366,13 +2366,15 @@ async def wrapper(*args, **kwargs):
 # Backward Compatibility Decorators
 # ============================================================================
 
+_Wrapped = TypeVar('_Wrapped', bound=Union[type, Callable])
```
janEbert: Why is just `Callable` not enough here?

megatron/core/utils.py
```diff
@@ -2366,13 +2366,15 @@ async def wrapper(*args, **kwargs):
...
+
 
 def deprecated(
     version: str,
     removal_version: Optional[str] = None,
     alternative: Optional[str] = None,
     reason: Optional[str] = None,
-) -> Callable:
+):
```
janEbert: Shouldn't this still return a `Callable`, or more specifically, a `Callable[[_Wrapped], _Wrapped]`?

megatron/core/utils.py
janEbert: Same as above for `deprecated`, shouldn't this return a `Callable`, or more specifically, a `Callable[[_Wrapped], _Wrapped]`?

megatron/core/utils.py
nschank: It is! But I've found people are often confused by the fact that `type`s are also `Callable`s so I did it to be a bit more self-documenting. But a docstring probably works better

megatron/core/utils.py
nschank: Yes, but I was leaning on type inference - for decorators you often need to, unfortunately. I updated these two as requested, but `experimental_fn` is an example where actually marking the return type would obfuscate the optional additional parameter in the returned function - in this case we could technically still get around it with a `Protocol` (although very excessive lol), but for complex decorators you can end up not being able to specify the type hint correctly at all!

nschank:
Haha mostly because this file is 2500 lines so I didn't really want to go yak shaving too hard, but fair. I updated a few other ones where it was immediately obvious how to, but there's some decorators which would require more work so I skipped them.

megatron/core/utils.py
janEbert: Oh curious, okay. So in the case of `experimental_fn`, it's better to have `Any`/`Unknown`, as the type checker figures things out better by itself? I don't really understand why it works for other places but not there.

Do you have an easy test setup that you're using for this? You made me also want to play around to get a better grasp. :)

janEbert:
Thank you!

jaredcasper (approved):

megatron/core/utils.py
nschank: I use pyright locally! Specifically I have `pip install pyright==1.1.408` in my devcontainer setup, then the Pylance extension should start giving you the red squigglies. For this particular PR, I mostly ended up comparing what VSCode autocomplete/mouse-over showed me for decorated functions - before, it would have no idea what the arguments were, and now it does.

In the case of `experimental_fn`, we actually want to not specify a return type at all. The type checker will infer the return type from the contents of the function, and the "name" of the type is too hard/annoying to spell out. 

Here's an example of a function which cannot have its return type decorated, because python's type system won't let you "write" the type correctly (without https://peps.python.org/pep-0827/): 

```python
def without_parameters[T, R](decorator: Callable[[T], R], /):  # noqa: ANN201
  """Converts a method which directly accepts an object into a decorator.

  The provided decorator should accept a decorated object as its first argument.
  The returned decorator can then be used either directly on an object (no
  parentheses) or with parentheses.

  Example:
    ```
    @decorator_builder.without_parameters
    def my_decorator(func: Callable[[int], int], /) -> Callable[[int], int]:
       ...

    @my_decorator
    def func1(x: int) -> int:
      return x + 1

    @my_decorator()
    def func2(x: int) -> int:
      return x + 2
    ```
  """

  @overload
  def wrapper(decorated: None = None, /) -> Callable[[T], R]: ...

  @overload
  def wrapper(decorated: T, /) -> R: ...

  @functools.wraps(decorator)
  def wrapper(decorated: T | None = None, /) -> R | Callable[[T], R]:
    if decorated is None:
      return decorator
    else:
      return decorator(decorated)

  return wrapper
  ```

If you just leave the annotation unset, the type-checker infers the correct type for you. `Any` would be bad because it would just tell the type checker it shouldn't attempt to infer the type!

janEbert: Thanks a lot! That gives me a much better sense of the issues, although I still don't understand the difference between the functions, where it works for some, but not others. I'll play around myself a bit, thanks again for the detailed summary!

PR #4063, opened 2026-03-30
Align chat completions endpoint with vLLM

santhnm2:
# What does this PR do ?
Aligns chat completions endpoint with vLLM.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py
@@ -476,7 +476,7 @@ def split_dtensor(
 
         new_dtensor = DTensor.from_local(
             sliced_tensor,
-            shape=out_shape,
+            shape=tuple(out_shape),
             stride=sliced_tensor.stride(),
             placements=dtensor.placements,
             device_mesh=dtensor.device_mesh,
--- a/megatron/core/resharding/nvshmem_copy_service/core/gpu_resource_manager.py
+++ b/megatron/core/resharding/nvshmem_copy_service/core/gpu_resource_manager.py
@@ -8,6 +8,8 @@ and event lifecycle.
 """
 
 import logging
+import os
+from importlib import metadata
 from typing import Dict, Optional
 
 from ..compat import ensure_nvshmem_compat, get_cuda_core_device_class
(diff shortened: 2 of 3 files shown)


megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
```diff
@@ -446,7 +409,9 @@ async def chat_completions():
 
         req = await request.get_json()
         tools = req.get("tools", None)
-        tools_requested = bool(tools)
+        tool_choice = req.get("tool_choice", None)
+        parallel_tool_calls = req.get("parallel_tool_calls", True)
```
santhnm2, line 413: @ArEsKay3 @i-riyad should this default to False?

megatron/core/tokenizers/text/text_tokenizer.py
```diff
@@ -75,18 +75,20 @@ def tokenize(self, text: str) -> List[int]:
 
         return self._tokenizer.text_to_ids(text)
 
-    def detokenize(self, ids: List[int]) -> str:
+    def detokenize(self, ids: List[int], skip_special_tokens: bool = True) -> str:
```
asolergi-nv: TikToken tokenizer is defaulting to `False`, HuggingFace tokenizer to `None` & this to `True`. Let's default all 3 values to `False` and set it to `True` in your required parts of the code, WDYT?

asolergi-nv: Can we also rename this argument to `remove_special_tokens`?

megatron/core/inference/text_generation_server/dynamic_text_gen_server/endpoints/chat_completions.py
santhnm2: from @sancha:
> vllm defaults https://docs.vllm.ai/en/stable/serving/openai_compatible_server/ parallel_tool_calls = True as default

Therefore will leave as is

ericharper:
Do we need to add any tests for this PR?

megatron/core/tokenizers/text/text_tokenizer.py
```diff
@@ -75,18 +75,20 @@ def tokenize(self, text: str) -> List[int]:
...
+            ids (list): token IDs to be detokenized.
+            skip_special_tokens (bool): Whether to strip special tokens
+                (e.g. <|im_end|>) from the output. Defaults to True.
 
         Returns:
             text: detokenized text.
         """
 
-        return self._tokenizer.ids_to_text(ids)
+        return self._tokenizer.ids_to_text(ids, remove_special_tokens=skip_special_tokens)
```
ericharper: Will this break with some tokenizers like NullTokenizer? 

https://github.com/NVIDIA/Megatron-LM/blob/97e36aaba37245e2fa23d4fda0e956ece21e6c9e/megatron/core/tokenizers/text/libraries/null_tokenizer.py#L23

megatron/core/tokenizers/text/text_tokenizer.py
santhnm2: Thanks for the suggestion! I looked into this but I think changing the default values isn't safe without also updating all the backend `ids_to_text` implementations to accept `skip_special_tokens`, which is perhaps too invasive for this PR. I can add a TODO to revisit this in a follow-up PR if that works?

santhnm2: So I think Hugging Face actually uses `skip_special_tokens` in their API: https://huggingface.co/docs/transformers/main_classes/tokenizer#transformers.PythonBackend.decode.skip_special_tokens
But we're using `remove_special_tokens` just in the Megatron tokenizer library. Perhaps the better change is to update `remove_special_tokens` to `skip_special_tokens`?

santhnm2:
> Do we need to add any tests for this PR?

Added a unit test `test_hf_detokenize_skip_special_tokens` to `tests/unit_tests/tokenizers/test_tokenizer.py`

megatron/core/tokenizers/text/text_tokenizer.py
santhnm2: Yes thanks for catching this - there was code in other places to gate this argument on whether the function actually has the parameter which I've now refactored into the `accepts_parameter` function. So now the code should be safe for all tokenizers (the `remove_special_tokens` argument will just be ignored if the parameter does not exist); I added a unit test to verify this also.

santhnm2: I also changed the default value for the `skip_special_tokens` argument to `None` so that we only pass it if it's explicitly set.

megatron/core/tokenizers/text/text_tokenizer.py
santhnm2: Actually I ended up making the default `None` here as well, so at least that aligns with the Hugging Face tokenizer now.

PR #4067, opened 2026-03-31
Add QK layernorm support for dot-product attention in MambaModel

Phlip79:
## Summary

- Adds config-driven QK norm support to `SelfAttention.__init__` so models with static specs (e.g. MambaModel) can enable QK layernorm via `--qk-layernorm` without modifying their specs
- Config selects the default norm class (`TENorm` or `L2Norm`); spec overrides if `q_layernorm`/`k_layernorm` are explicitly set
- GPTModel behavior is unchanged — its specs always provide `q_layernorm`/`k_layernorm`, which take precedence

## Test plan

- [x] Default config (no norm) — verify `q_layernorm` and `k_layernorm` are `None`
- [x] `qk_layernorm=True` — verify norms are created via `TENorm`
- [x] `qk_l2_norm=True` — verify norms are `L2Norm` instances
- [x] Spec-provided norm takes precedence over config default
- [x] Forward pass with `qk_layernorm=True` produces correct output shape
- [x] [Functional tests](https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/pipelines/47669127)

Closes MCORE-20.

### DIFF
--- a/tests/unit_tests/transformer/test_attention.py
+++ b/tests/unit_tests/transformer/test_attention.py
@@ -728,10 +728,8 @@ def test_parallel_attention_correctness_num_query_groups_less_than_tp_size(
 
 def test_qk_layernorm_from_config_fallback():
     """config.qk_layernorm=True with spec q/k_layernorm=None builds TENorm."""
-    pytest.importorskip("transformer_engine")
+    te_pytorch = pytest.importorskip("transformer_engine.pytorch")
     from dataclasses import replace
-
-    from megatron.core.extensions.transformer_engine import TENorm
 
     Utils.initialize_model_parallel(1, 1)
     model_parallel_cuda_manual_seed(123)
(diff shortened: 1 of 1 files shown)


janEbert:
I think this is exactly what we want to avoid; as far as I understand, we do not want to start to make the spec dynamic in code. :)
Isn't this solvable by dynamically passing the arguments in the `MambaStack` or `MambaModel` constructors?

Phlip79:
@claude review

megatron/core/transformer/attention.py
```diff
@@ -1288,23 +1289,41 @@ def __init__(
...
 
-        if submodules.q_layernorm is not None:
-            self.q_layernorm = submodules.q_layernorm(
+        # Resolve which norm class to use for Q and K.
+        # Config selects the default norm class; spec overrides if set.
+        default_norm_cls = None
+        if self.config.qk_layernorm and TENorm is not None:
+            default_norm_cls = TENorm
+        elif self.config.qk_l2_norm:
+            default_norm_cls = L2Norm
```
janEbert: (Very) minor nit: GPT prioritizes `qk_l2_norm`, this prioritizes `qk_layernorm`.

megatron/core/transformer/attention.py
```diff
@@ -1288,23 +1289,41 @@ def __init__(
...
+
+        q_norm_cls = submodules.q_layernorm or default_norm_cls
+        k_norm_cls = submodules.k_layernorm or default_norm_cls
+
+        if self.config.qk_layernorm and (q_norm_cls is None or k_norm_cls is None):
```
janEbert: To also consider `qk_l2_norm`, this should be
```python
if (self.config.qk_layernorm or self.config.qk_l2_norm) and (q_norm_cls is None or k_norm_cls is None):
```
, right?

megatron/core/transformer/attention.py
Phlip79: Good catch! Updated.

megatron/core/transformer/attention.py
Phlip79: Yes you're right.

megatron/core/transformer/attention.py
```diff
@@ -1288,23 +1289,43 @@ def __init__(
...
 
-        if submodules.q_layernorm is not None:
-            self.q_layernorm = submodules.q_layernorm(
+        # Resolve which norm class to use for Q and K.
+        # Config selects the default norm class; spec overrides if set.
+        default_norm_cls = None
+        if self.config.qk_l2_norm:
+            default_norm_cls = L2Norm
+        elif self.config.qk_layernorm and TENorm is not None:
+            default_norm_cls = TENorm
```
yuzhongw-nvidia: A quick question: if we do not use spec, how can users specify local norm instead of TE norm? Do we still use `config.transformer_impl` or something new?

janEbert: Mamba itself doesn't allow changing the backend yet, TE layers are hardcoded in [`mamba_layer_specs.py`](https://github.com/NVIDIA/Megatron-LM/blob/2368a2e3a71a98cd724effbf137785dc6eb7771f/megatron/core/models/mamba/mamba_layer_specs.py). I don't know if there are plans to change this in the future.

megatron/core/transformer/attention.py
```diff
@@ -1288,23 +1289,43 @@ def __init__(
             tp_group=self.pg_collection.tp,
         )
 
-        if submodules.q_layernorm is not None:
-            self.q_layernorm = submodules.q_layernorm(
+        # Resolve which norm class to use for Q and K.
+        # Config selects the default norm class; spec overrides if set.
+        default_norm_cls = None
```
jaredcasper: Why have a default norm class?

Phlip79: It's default because we still want to support override using the spec system.

megatron/core/transformer/attention.py
yuzhongw-nvidia: Well, let's mark it as a TODO. It is not a very high priority case, but we still need it sometimes.
```suggestion
            # TODO(yuzhongw, janpabloe): Support local backend.
            default_norm_cls = TENorm
```

PR #4073, opened 2026-03-31
[Dev] Add permute/unpermute fusion with dispatch/combine in Hybrid-EP

Autumn1998:
# What does this PR do ?

This PR introduce the new feature: fuse the permute/unpermute with the dispatch/combine into 1 kernel
This feature is provided by the hybrid-ep
related PR in main: #4089 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/tests/unit_tests/models/test_mamba_moe_model.py
+++ b/tests/unit_tests/models/test_mamba_moe_model.py
@@ -167,7 +167,7 @@ GOLDEN_CONFIG: Dict[str, Any] = {
     "moe_ffn_hidden_size": 1856,
     "moe_flex_dispatcher_backend": "deepep",
     "moe_grouped_gemm": True,
-    "moe_hybridep_num_sms": 16,
+    "moe_hybridep_num_sms": None,
     "moe_hybridep_num_blocks_permute": None,
     "moe_hybridep_num_blocks_unpermute": None,
     "moe_input_jitter_eps": None,


megatron/core/transformer/transformer_config.py
```diff
@@ -772,6 +709,9 @@ class TransformerConfig(ModelParallelConfig):
     Options are "deepep" and "hybridep". Currently only "hybridep" backend supports 
     the MNNVL case."""
 
+    moe_hybridep_permute_fusion: bool = False
```
Victarry: Suggest to use name like `moe_permute_fusion_into_hybriep` to indicate the fusion/overlap between permute and comm.

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -400,7 +403,11 @@ def backward(ctx, grad_x, grad_probs, grad_scaling_factor, grad_tokens_per_exper
         '''
         handle = ctx.handle
         combined_hidden, combined_probs = _hybrid_ep_buffer.combine_with_unpermute(
-            hidden=grad_x, probs=grad_probs, handle=handle, pad_multiple=ctx.pad_multiple
+            hidden=grad_x,
+            probs=grad_probs,
+            handle=handle,
+            pad_multiple=ctx.pad_multiple,
+            fuse_unpermute_combine=ctx.fused,
```
Victarry: Could we add a version guard to this new API key?

Autumn1998: Added a version guard to the buffer initialization function

megatron/core/transformer/transformer_config.py
```diff
@@ -820,14 +760,11 @@ class TransformerConfig(ModelParallelConfig):
     """Number of SMs to use for HybridEP. In pure NVL scenarios,
     16 SMs can generally achieve good bandwidth."""
 
-    moe_mlp_glu_interleave_size: Optional[int] = None
-    """When set, GLU activations in the MoE grouped MLP layer will use a
-    block interleaved format. Instead of interpreting the input tensor
-    as a concatenation of gates and linear units, it will be
-    interpreted as alternating blocks of gates and linear units.
+    moe_hybridep_num_blocks_permute: int = 96
```
gautham-kollu: I am trying to understand why restrict this to some blocks and not all the blocks ? Is block an moe block here ?   Also why 96 ?

gautham-kollu:
@Autumn1998  Can we add this to an existing functional test in MLM that is marked "mr-github-slim" so that we protect this against regressions ?

@ko3n1g  is "mr-github-slim" the right place to so we run as a part of each PR ?

Consider Megatron-LM/tests/test_utils/recipes/h100/moe.yaml 
cc: @ko3n1g   for any other thoughts on testing

megatron/core/transformer/transformer_config.py
yashaswikarnati: may be a nit; just a little unclear to me. Could we add a short info what exactly these blocks are how to select a value for it

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -451,6 +495,9 @@ def hybrid_ep_dispatch(
         num_local_experts,
         num_sms_dispatch_api=24,
         num_sms_combine_api=24,
+        num_blocks_permute=96,
```
yashaswikarnati: should the default value be None or 0, as by default fusion is not enabled and we have single source of truth to control this

megatron/core/transformer/transformer_config.py
Autumn1998: The 'block' here refers to a CUDA thread block. In fuse mode, it can be thought of as how many SMs are used by the permute operation. 96 is a setting we tested that delivers reasonably good performance. However, after thinking about it more, it would be better to set this to `None`, because Hybrid EP will automatically choose a good value when None is passed in.

Autumn1998: This is actually the number of CUDA thread blocks used by the permute operation. In the fuse path, it corresponds to the number of SMs used. In the non-fuse path, there can be cases where multiple blocks are scheduled on the same SM.

megatron/core/transformer/moe/token_dispatcher.py
vasunvidia: Can this assertion be removed with this PR?

Autumn1998: sure

Issue #3999, opened 2026-03-23
[Bug] HybridEP dispatcher passes incorrect max_num_of_tokens_per_rank to DeepEP, causing RDMA QP assertion failure


tingmingzhong:
## Summary

When using the HybridEP backend (`--moe_flex_dispatcher_backend hybridep`) for MoE expert-parallel training across multiple nodes, `HybridEPDispatch.forward()` in `fused_a2a.py` incorrectly uses the **total number of tokens in a micro-batch** (`seq_length × micro_batch_size`) as the `max_num_of_tokens_per_rank` parameter for `HybridEPBuffer`. This causes the RDMA Queue Pair send-queue depth (`tx_depth`) to exceed the hardware limit of 65535, triggering an assertion failure in DeepEP's internode communication initialization.

## Environment

- **Megatron-LM**: latest (hybrid-ep branch of DeepEP integrated)
- **DeepEP**: v1.2.1 (hybrid-ep branch)
- **Model**: Qwen3-30B-A3B (MoE, 128 experts)
- **Hardware**: 2 nodes × 8 GPUs, InfiniBand RDMA interconnect
- **Training config**: `--max_length 8192 --micro_batch_size 8 --packing true --expert_model_parallel_size 16 --moe_flex_dispatcher_backend hybridep`

## Error

```
python: /path/to/DeepEP/csrc/hybrid_ep/buffer/internode.cu:167:
void setup_qp_init_attr(..., int): Assertion `tx_depth > 0 && tx_depth < 65536' failed.
```

All ranks crash with `SIGABRT (signal 6)` during HybridEP buffer initialization.

### The call chain

1. **`MoEFlexTokenDispatcher.dispatch_preprocess()`** (`token_dispatcher.py:1438`) reshapes `hidden_states` from `[seq_length, batch_size, hidden_size]` to `[seq_length * batch_size, hidden_size]`:

   ```python
   hidden_states = hidden_states.view(-1, self.hidden_shape[-1])
   ```

2. **`HybridEPDispatch.forward()`** (`fused_a2a.py:354-359`) extracts the first dimension of the already-flattened tensor and uses it as `seq_len`:

   ```python
   if _hybrid_ep_buffer is None:
       seq_len, hidden_dim = x.shape[-2:]  # x is [seq_len * batch_size, hidden_dim]
       init_hybrid_ep_buffer(group, hidden_dim, seq_len, ...)  # seq_len is actually num_total_tokens
   ```

[...]

PR #4094, opened 2026-04-01
Add Python-side guardrail for HybridEP InfiniBand limit and rename seq_len

Shreyas-S-809:
**Description** 

Currently, when running multi-node MoE training with the HybridEP backend, passing a total token count (`seq_length * micro_batch_size`) that results in a DeepEP Queue Pair depth (`tx_depth = 3 * num_tokens + 1`) greater than 65535 causes an immediate and ungraceful C++ `SIGABRT` across all ranks due to InfiniBand hardware limits.

Following the architectural discussion in the issue thread, this PR improves the UX around this hardware limitation by catching the overflow in Python and raising a clean, actionable error message.

**Changes**

* **Reverted** the previous attempt to alter the tensor shape logic in token_dispatcher.py, as DeepEP intentionally expects the fully flattened batch.

* **Renamed** `seq_len` to `num_tokens` in `fused_a2a.py` (`init_hybrid_ep_buffer` and `HybridEPDispatch.forward`) to accurately reflect that the variable holds the folded `seq_length * batch_size`.

* **Added** a `ValueError` guardrail in `HybridEPDispatch.forward`. If the calculated `tx_depth` breaches the InfiniBand limit, it now raises a clear error advising the user to reduce sequence length/batch size or increase their TP/CP degrees, rather than dumping core.

**Testing**

Tested locally to ensure Python syntax and logic are sound. Relying on CI to verify no regressions in the standard dispatch workflow.

### DIFF
--- a/megatron/core/transformer/moe/fused_a2a.py
+++ b/megatron/core/transformer/moe/fused_a2a.py
@@ -278,7 +278,7 @@ _hybrid_ep_buffer = None
 def init_hybrid_ep_buffer(
     group: torch.distributed.ProcessGroup,
     hidden_dim: int,
-    seq_len: int,
+    num_tokens: int,
     num_local_experts: int,
     num_sms_dispatch_api: int,
     num_sms_combine_api: int,
@@ -313,7 +313,7 @@ def init_hybrid_ep_buffer(
     _hybrid_ep_buffer = HybridEPBuffer(
         group=group,
         hidden_dim=hidden_dim,
-        max_num_of_tokens_per_rank=seq_len,
+        max_num_of_tokens_per_rank=num_tokens,
         num_local_experts=num_local_experts,
         use_fp8=fp8_dispatch,
         num_sms_dispatch_api=num_sms_dispatch_api,
--- a/megatron/core/transformer/moe/fused_a2a.py
+++ b/megatron/core/transformer/moe/fused_a2a.py
@@ -351,12 +351,26 @@ class HybridEPDispatch(torch.autograd.Function):
         Forward pass of fused dispatch of the HybridEP backend
         '''
         if _hybrid_ep_buffer is None:
-            seq_len, hidden_dim = x.shape[-2:]
+            num_tokens, hidden_dim = x.shape[-2:]
+
+            # --- Hardware Limit Gaurdrail ---
+            # DeepEP calculates tx_depth = 3 * num_tokens + 1.
+            # # InfiniBand strictly asserts tx_depth < 65536.
+            tx_depth = 3 * num_tokens + 1
+            if tx_depth >= 65536:
+                raise ValueError(
+                    f"HybridEP RDMA Queue Pair depth ({tx_depth}) exceeds the InfiniBand "
+                    f"hardware limit of 65535. This occurs because the total tokens per rank "
+                    f"({num_tokens}) is too high. Please reduce sequence length or micro-batch size, "
+                    f"or increase Tensor Parallelism (TP) / Context Parallelism (CP) to reduce "
+                    f"the number of tokens processed per rank."
+                )
+            
             fp8_dispatch = False  # Currently, we do not support fp8 dispatch
             init_hybrid_ep_buffer(
                 group,
                 hidden_dim,
-                seq_len,
+                num_tokens,
                 num_local_experts,
                 num_sms_dispatch_api,
                 num_sms_combine_api,
--- a/megatron/core/transformer/moe/fused_a2a.py
+++ b/megatron/core/transformer/moe/fused_a2a.py
@@ -353,19 +353,18 @@ class HybridEPDispatch(torch.autograd.Function):
         if _hybrid_ep_buffer is None:
             num_tokens, hidden_dim = x.shape[-2:]
 
-            # --- Hardware Limit Gaurdrail ---
+            # --- Hardware Limit Guardrail ---
             # DeepEP calculates tx_depth = 3 * num_tokens + 1.
-            # # InfiniBand strictly asserts tx_depth < 65536.
+            # InfiniBand strictly asserts tx_depth < 65536.
             tx_depth = 3 * num_tokens + 1
             if tx_depth >= 65536:
                 raise ValueError(
                     f"HybridEP RDMA Queue Pair depth ({tx_depth}) exceeds the InfiniBand "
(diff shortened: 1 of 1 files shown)


megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -351,11 +352,12 @@ def forward(
         Forward pass of fused dispatch of the HybridEP backend
         '''
         if _hybrid_ep_buffer is None:
-            seq_len, hidden_dim = x.shape[-2:]
+            seq_len, hidden_dim = x.shape[-1]
```
janEbert: This change should probably be reverted or you just want to assign to `hidden_dim`; you can't expect to get a two-tuple when indexing into a single value in the `shape` tuple.

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -351,11 +352,12 @@ def forward(
...
             fp8_dispatch = False  # Currently, we do not support fp8 dispatch
             init_hybrid_ep_buffer(
                 group,
                 hidden_dim,
+                max_tokens_per_rank,
                 seq_len,
```
janEbert: `seq_len` should be removed, since `max_tokens_per_rank` superceeds its value, right?

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1023,6 +1023,15 @@ def setup_metadata(self, routing_map: torch.Tensor, probs: torch.Tensor):
...
             )
+            # Set max tokens per rank based on strict capacity.
+            self.max_tokens_per_rank = self.capacity * self.num_local_experts
+        else:
+            # Calculate expected average tokens per EP rank + padding factor.
+            avg_tokens_per_rank = (num_tokens * slef.config.moe_router_topk) // self.group.size()
+            safe_limit = int(avg_tokens_per_rank * (self.capacity_factor or 2.0))
+            # Hardware limit guard : DeepEP QP depth - 3*N + 1 < 65536.
+            # N <= 21845.
+            self.max_tokens_per_rank = min(safe_limit, 21845)
```
janEbert: `capacity` is not actually used to change the shape of the input tensor for this dispatcher, so I don't think this is correct, sadly. In general, I'm unsure if the issue is correctly flagged. Let's discuss in the issue before continuing work on the PR.

janEbert (changes requested):

janEbert:
[See the comment in the issue](https://github.com/NVIDIA/Megatron-LM/issues/3999#issuecomment-4177847325); I'm not sure whether we need to fix anything here, actually.

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -351,12 +351,26 @@ def forward(
         Forward pass of fused dispatch of the HybridEP backend
         '''
         if _hybrid_ep_buffer is None:
-            seq_len, hidden_dim = x.shape[-2:]
+            num_tokens, hidden_dim = x.shape[-2:]
+
+            # --- Hardware Limit Gaurdrail ---
```
janEbert: Gaurdrail → Guardrail

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -351,12 +351,26 @@ def forward(
...
+            # DeepEP calculates tx_depth = 3 * num_tokens + 1.
```
janEbert, line 357: I see that they also have a `tx_depth = 2 * num_tokens + 1` calculation (in `RDMACoordinator::allocate_combine_buffers`), giving us an even lower bound.

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -351,12 +351,26 @@ def forward(
...
+            # # InfiniBand strictly asserts tx_depth < 65536.
```
janEbert: Please change `# # ` to `# `.

janEbert (changes requested):
I think we should check for the lower bound, i.e., `tx_depth = 2 * num_tokens + 1`!
Thanks a lot!

Shreyas-S-809:
> I think we should check for the lower bound, i.e., `tx_depth = 2 * num_tokens + 1`! Thanks a lot!

Hey @janEbert , great catch finding the `allocate_combine_buffers` logic!

Just a quick mathematical double-check before I push the update: Since both the dispatch (`3x + 1`) and combine (`2x + 1`) queues must stay under `65536`, doesn't the `3 * num_tokens + 1` calculation actually hit the hardware ceiling first?

* `3 * num_tokens + 1` >= 65536 triggers when tokens exceed 21,845

* `2 * num_tokens + 1` >= 65536 triggers when tokens exceed 32,767

If we update the guardrail to only check `2x + 1`, a config with e.g., 25,000 tokens will pass our Python check, but still crash the C++ backend with a `SIGABRT` during the dispatch allocation.

Should we keep the stricter `3 * num_tokens + 1` check to safely cover both allocations, or would you prefer I check the `max()` of both explicitly? Happy to push whichever you think is best!

janEbert:
Oops, you're totally right, forget about that part!

megatron/core/transformer/moe/fused_a2a.py
janEbert: Ignore; we want to check for the higher bound as @Shreyas-S-809 correctly stated.

janEbert (approved):

janEbert (approved):

janEbert:
Hey, please run `tools/autoformat.sh` (or pre-commit hooks) to fix the linting errors.

janEbert:
Afterwards, we still need reviews from @NVIDIA/core-adlr, @NVIDIA/core-nemo, @NVIDIA/mixture-of-experts-adlr, and @NVIDIA/mixture-of-experts-devtech.

Shreyas-S-809:
> Hey, please run `tools/autoformat.sh` (or pre-commit hooks) to fix the linting errors.

Fixed lint issues, sorry for that.

yashaswikarnati (approved):

deepakn94 (approved):

yaox12 (approved):

fanshiqing (approved):

shanmugamr1992 (approved):

PR #4096, opened 2026-04-01
Fix RL reward due to stop token

tdene:
# What does this PR do ?

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/core/dist_checkpointing/strategies/torch.py
+++ b/megatron/core/dist_checkpointing/strategies/torch.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
 """ Strategies using PyTorch distributed.checkpoint as an underlying format. """
+import inspect
 import io
 import os
 import pickle
@@ -600,6 +601,7 @@ class TorchDistSaveShardedStrategy:
         thread_count: int = 1,
         cached_metadata: bool = False,
         separation_hint: Optional[str] = None,
+        cpu_shm_mode: bool = False,
     ):
         """Adds parameters specific to PyT Distributed format
         Args:
@@ -614,6 +616,10 @@ class TorchDistSaveShardedStrategy:
                 gathering local metadata every checkpointing invocation
             separation_hint(str, optional): If provided, all tensors whose keys have this
                 prefix will be saved to a separate file.
+            cpu_shm_mode (bool, optional): Copy GPU tensors to CPU shared-memory in the
+                training process before handing off to the async worker. Avoids CUDA IPC /
+                NVLink fabric handles in the worker subprocess. Only applies with nvrx async
+                strategy.
         """
         self.backend = backend
         self.version = version
@@ -640,6 +646,7 @@ class TorchDistSaveShardedStrategy:
         self.cached_global_metadata: Optional[Metadata] = None
 
         self.separation_hint = separation_hint
+        self.cpu_shm_mode = cpu_shm_mode
 
         self.validated_loaded_metadata_reuse = False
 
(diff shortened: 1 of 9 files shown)


examples/rl/environments/countdown/countdown_agent.py
```diff
@@ -41,4 +41,5 @@ async def get_prompt(self, validation=False) -> tuple[str, dict]:
         return self.make_prefix(**golden), golden
 
     async def get_reward(self, response, golden: dict, finish_reason: str | None = None) -> float:
+        # NOTE: finish_reason is intentionally unused.
```
yobibyte: This is a very confusing comment that only makes it more mysterious for a future dev.

examples/rl/environments/math/math_agent.py
```diff
@@ -60,6 +60,8 @@ def compute_score(
         """
         # Generation that stopped cleanly (EOD or stop word) rather than
         # hitting the token limit is eligible for full reward.
+
+        # NOTE: finish_reason=None is treated as "stopped" here for backwards compatibility.
```
yobibyte: Do we actually need to keep backwards compatibility here? Can we update all of the call sites that use that and avoid supporting this?

examples/rl/environments/math/math_agent.py
```diff
@@ -106,7 +108,7 @@ def compute_score(
                 if correct_answer:
                     if stopped and not after:
                         return 1.0
-                    if not after:
+                    if stopped:
```
yobibyte, line 109: I think this is a bug. You assign to 'after' above, but check 'stopped'.

tests/unit_tests/rl/test_rl_agents.py
```diff
@@ -52,9 +52,9 @@ class TestRLAgentRewards:
             # Correct answer, stopped cleanly, nothing trailing.
             pytest.param(lambda c, w: c, "stop", 1.0, id="correct-stop-clean"),
             # Correct answer, hit token limit, nothing trailing.
-            pytest.param(lambda c, w: c, "length", 0.1, id="correct-length-clean"),
+            pytest.param(lambda c, w: c, "length", 0.05, id="correct-length-clean"),
             # Correct answer, stopped cleanly, but trailing text.
-            pytest.param(lambda c, w: c + " extra", "stop", 0.05, id="correct-stop-trailing"),
+            pytest.param(lambda c, w: c + " extra", "stop", 0.1, id="correct-stop-trailing"),
```
yobibyte, line 57: If the issue I found above is a bug, add a test case covering it.

tdene: I wrote `pytest.param(lambda c, w: c, "length", 0.05, id="correct-length-clean"),` and `pytest.param(lambda c, w: c + " extra", "stop", 0.1, id="correct-stop-trailing"),` to cover those cases. I'm pretty sure I have full coverage.

examples/rl/environments/math/math_agent.py
tdene: It took me a while to understand and get this right. Original code was:
```
  if correct_answer:
      # 1. End token immediately after answer tag -> full reward
      for token in end_tokens:
          if after.startswith(token):
              return 1.0
      # 2. End token somewhere later (trailing text before it) -> partial_end_reward
      for token in end_tokens:
          if token in after:
              return self.partial_end_reward
      # 3. No end token at all (hit length limit) -> format_reward
      return self.format_reward
```

In the old code, the EOD text was part of the actual text. This means that if `after` was empty, this always meant there was no EOD, which always meant `format_reward`.     
In the new code, an empty `after` doesn't necessarily mean that; `stopped` needs to be checked instead.

So the new code is:
```
  if stopped and not after:    # 1. EOD generated, no text after answer
      return 1.0
  if stopped:                  # 2. EOD generated, text between answer and EOD
      return self.partial_end_reward
  return self.format_reward    # 3. EOD not generated
```

examples/rl/environments/math/math_agent.py
tdene: Yes please! I wanted to do the same.

Done.

examples/rl/environments/countdown/countdown_agent.py
tdene: Resolved due to the other comment!

examples/rl/environments/math/math_agent.py
```diff
@@ -46,23 +47,28 @@ def __init__(self,
...
         """Take a response and a golden answer and return a score. Supports tagged or boxed answers.
 
         Uses the final answer in the response string to compute the score.
         """
-        # Allow <answer> tags or \boxed{} tags (this is a bit of cheating in favor of deepseek distilled models I think)
-        matched_format = None
-        end_tokens = ["<|end_of_text|>", "<|endoftext|>", "</s>", "<|eot_id|>", "<|im_end|>"]
+        # Generation that stopped cleanly (EOD or stop word) rather than
+        # hitting the token limit is eligible for full reward.
+        stopped = finish_reason != "length"
```
yobibyte, line 63: Not sure how stable the API is, but is there a chance "length" can be changed for smth else in the future?

examples/rl/environments/math/math_agent.py
yobibyte: Okay, I see it now. My bad!

examples/rl/environments/math/math_agent.py
tdene: I do not believe so. This follows the OpenAI API standard, which everyone tries to ascribe to.

PR #4101, opened 2026-04-01
Make MTP / prefix cache stats persist for engine lifetime

santhnm2:
# What does this PR do ?
Makes MTP acceptance rate and prefix cache hit rate stats persist for the engine lifetime instead of resetting on every logging statement.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


ArEsKay3 (approved):

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1218,13 +1221,16 @@ def post_process_requests(
                     request
                 )
 
-                # Track acceptance statistics for logging.
-                if len(request.generated_tokens) > 0 and self.num_speculative_tokens > 0:
+                # Track per-position acceptance statistics for logging.
+                # Skip prefill requests: MTP heads only propose speculative tokens
+                # for decode requests, so counting prefill requests would inflate
+                # the denominator and artificially deflate the acceptance rate.
+                if not is_prefill and len(request.generated_tokens) > 0 and self.num_speculative_tokens > 0:
```
lmcafee-nvidia: my understanding is that `_spec_steps` still increments even for prefill-only steps, as long as `accepted_tokens` is returned. should `_spec_steps` be updated to only increment if there is at least one decode request?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -121,11 +121,21 @@ def __init__(self, inference_wrapped_model: AbstractModelInferenceWrapper, token
...
+        configured ``num_speculative_tokens`` rather than the physical
+        layer count stored in ``mtp_num_layers``.
+        """
         model = self.inference_wrapped_model.model
-        if hasattr(model, 'config') and hasattr(model.config, 'mtp_num_layers'):
-            return model.config.mtp_num_layers or 0
-        return 0
+        if not (hasattr(model, 'config') and hasattr(model.config, 'mtp_num_layers')):
+            return 0
+        if not model.config.mtp_num_layers:
```
lmcafee-nvidia: my understanding is that there are two MTP modes: 1) separate layers for each token, and 2) one MTP layer used multiple times. it seems like we might be returning `0` as a result of the multi-layer case, without considering whether `mtp_use_repeated_layer` is true, which means we are potentially skipping MTP even though it should be active.

megatron/core/inference/engines/dynamic_engine.py
santhnm2: Good catch, thanks - fixed

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: I cleaned up this logic so that we don't have redundant `mtp_num_heads` and `mtp_num_depths` - now we only have `mtp_num_depths` which should be computed correctly

megatron/core/inference/engines/dynamic_engine.py
lmcafee-nvidia: it looks like `_spec_steps` still updates above for non-decode batches, line ~1130 inside the code block:

```
        if self.num_speculative_tokens > 0 and accepted_tokens is not None:
            self._spec_steps += 1
```

Phlip79:
Will you add `HybridModel` support in a follow-up PR?

santhnm2:
> Will you add `HybridModel` support in a follow-up PR?

Yea I was experimenting with a change to `hybrid_model.py` but removed it from this PR, will defer to a different PR

lmcafee-nvidia (approved):

kvareddy (approved):

shanmugamr1992 (approved):

PR #4104, opened 2026-04-01
fix: remove weights_only=False for multimodal example

faradawn:
## Summary
Fix security issue: https://nvbugspro.nvidia.com/bug/5996966

- Removes hardcoded `weights_only=False` from `torch.load` in `examples/multimodal/combine_state_dicts.py` (introduced in #3452)
- Users who need legacy checkpoint loading should opt in via `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1`
- Adds a one-line note to the multimodal README explaining the env var

### DIFF
--- a/examples/multimodal/README.md
+++ b/examples/multimodal/README.md
@@ -39,7 +39,7 @@ Update the paths to point to the mcore converted CLIP and Mistral models and run
 examples/multimodal/combine_lm_vision_checkpoints.sh /path/to/mistral/model /path/to/clip/model /output/dir
 ```
 
-> **Note:** If you encounter a loading error, try setting `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1`.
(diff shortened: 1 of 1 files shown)


examples/multimodal/README.md
```diff
@@ -39,6 +39,8 @@ Update the paths to point to the mcore converted CLIP and Mistral models and run
 examples/multimodal/combine_lm_vision_checkpoints.sh /path/to/mistral/model /path/to/clip/model /output/dir
 ```
 
+> **Note:** If you encounter a loading error, try setting `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1`.
```
chtruong814: Thanks for this.  Could you add the caveat about only doing with trusted checkpoints?

faradawn: Added the notes for trusted endpoint!

PR #4110, opened 2026-04-02
[Dev] overload factor logging

nanz-nv:
# What does this PR do ?

This PR introduces a utility to log overload factor through `log_overload_factor`.

## MoE overload factor

**Overload factor** is the ratio of the token count on the **most loaded** rank in a TP-EP group to the **balanced** token count per rank—the count each rank would see with perfectly balanced routing. It measures workload imbalance.

### `moe/avg_overload_factor`

Arithmetic mean of `overload factor` over all **layer × microbatch x DP slice** entries recorded on this rank in the step.

### `moe/max_overload_factor`

Maximum of `overload factor` over all **layer × microbatch x DP slice** entries. Useful for estimating **peak buffer size** for intermediate activations in the forward path.

### `moe/max_cum_overload_factor`

Reflects the **cumulative** fwd/bwd tokensL ratio of peak cumulative actual tokens to peak cumulative balanced count. Useful for estimating **how much activation-related memory** may need to be retained through backward.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -743,7 +743,7 @@ class TransformerConfig(ModelParallelConfig):
     ForwardGroupedMLP_CuTeGEMMSwiGLU_MXFP8 fusion on SM100+ with MXFP8 recipe.
     Requires ``use_te_op_fuser=True`` and SwiGLU activation.
     """
-    log_overload_factor: bool = False
+
     log_moe_overload_factor: bool = False
     """When True, log MoE overload metrics (avg/max vs balanced token count per step; max cum
     overload = peak cumulative actual tokens / peak cumulative balanced count over interleaved


megatron/core/transformer/moe/moe_layer.py
```diff
@@ -467,9 +484,37 @@ def routed_experts_compute(self, hidden_states: torch.Tensor, probs: torch.Tenso
...
+                tensor=dispatched_input,
+                tokens_per_expert=tokens_per_expert,
+                local_balanced_token_count=local_balanced,
+                layer_number=self.layer_number,
+                tp_ep_group=self.tp_ep_group,
+                dp_group=self.dp_group,
+            )
         if (
             hasattr(self, "_inference_token_dispatcher")
             and self.is_inference_cuda_graphed_iteration
```
Victarry, line 554: **[SUGGESTION]** The `log_overload_factor` block inside `experts_compute_dispatch` (balanced count retrieval, dispatcher type check, tensor construction, hook registration) should be extracted into a private method such as `_record_overload_factor(self, dispatched_input, tokens_per_expert)`.

Mixing this logic into `experts_compute_dispatch` hurts readability. A single call site keeps the dispatch method focused on dispatch logic.

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -431,6 +434,20 @@ def dispatch(self, hidden_states: torch.Tensor, probs: torch.Tensor):
...
+        """Routing map still held after ``token_dispatch`` (cleared in AllGather ``dispatch_postprocess``).
+
+        Flex/HybridEP keep the map on ``_comm_manager``.
+        """
+        td = self.token_dispatcher
+        rm = getattr(td, "routing_map", None)
+        if rm is not None:
+            return rm
+        cm = getattr(td, "_comm_manager", None)
+        if cm is not None:
```
Victarry: **[SUGGESTION]** Avoid abbreviated variable names in parallel/distributed code where clarity is critical:
- `td` → `token_dispatcher`
- `rm` → `routing_map`
- `cm` → `comm_manager`

Same applies to `td` (L501) and `ws` (L495) in `experts_compute_dispatch`. Use `tp_ep_world_size` for `ws`.

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -431,6 +434,20 @@ def dispatch(self, hidden_states: torch.Tensor, probs: torch.Tensor):
         """
         return self.token_dispatcher.token_dispatch(hidden_states, probs)
 
+    def _routing_map_after_token_dispatch(self) -> Optional[torch.Tensor]:
+        """Routing map still held after ``token_dispatch`` (cleared in AllGather ``dispatch_postprocess``).
+
+        Flex/HybridEP keep the map on ``_comm_manager``.
+        """
+        td = self.token_dispatcher
+        rm = getattr(td, "routing_map", None)
```
Victarry: **[SUGGESTION]** The balanced token count (`routing_map.shape[0] * topk`) could be computed earlier in `MoELayer.forward()` directly from `hidden_states` before any dispatch, rather than reading `routing_map` in this post-`token_dispatch` window.

Current approach has two fragilities:
1. Requires `routing_map` to still be alive after `token_dispatch` but before `dispatch_postprocess` clears it — a timing assumption tied to dispatcher internals.
2. Needs separate handling for AllGather (`routing_map` attr) vs Flex/HybridEP (`_comm_manager.routing_map`) dispatchers — coupling to internal implementation details.

Computing from `hidden_states.shape[0]` in `MoELayer.forward()` would remove `_routing_map_after_token_dispatch` entirely.

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -467,9 +484,37 @@ def routed_experts_compute(self, hidden_states: torch.Tensor, probs: torch.Tenso
...
+            # fair share so report()'s SUM over tp_ep matches one global balanced count.
+            td = self.token_dispatcher
+            if isinstance(td, MoEAllGatherTokenDispatcher) and (
+                td.tp_size > 1 or td.ep_size > 1
+            ):
+                base = base / ws
+            local_balanced = torch.empty(
+                (), device=dispatched_input.device, dtype=torch.float32
+            )
+            local_balanced.fill_(base)
```
Victarry: **[SUGGESTION]** `torch.empty(()) + fill_()` is unnecessarily verbose. Use `torch.tensor` directly:

```python
local_balanced = torch.tensor(base, device=dispatched_input.device, dtype=torch.float32)
```

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -978,6 +981,87 @@ def clear_aux_losses_tracker() -> None:
...
+        Returns:
+            ``tensor`` unchanged.
+        """
+        if layer_number is None:
+            return tensor
+
+        tokens_on_rank = tokens_per_expert.detach().sum()
+        if not tokens_on_rank.is_floating_point():
+            tokens_on_rank = tokens_on_rank.float()
+        tokens_on_rank = tokens_on_rank.to(device=tensor.device, dtype=torch.float32).reshape(())
```
Victarry: **[SUGGESTION]** `.reshape(())` is a no-op here — `tokens_per_expert.detach().sum()` already returns a 0-dim tensor. Same applies to `balanced` on L1016. Both `.reshape(())` calls can be removed.

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -978,6 +981,87 @@ def clear_aux_losses_tracker() -> None:
...
+        layer_number: Layer index (1-based).
+        tp_ep_group: TP × EP process group.
+        dp_group: Data-parallel group for cross-replica stats.
+
+    Returns:
+        ``tensor`` unchanged.
+    """
+    get_moe_overload_factor_tracker().set_process_groups(
+        tp_ep_group=tp_ep_group, dp_group=dp_group
+    )
```
Victarry: **[SUGGESTION]** `set_process_groups` is called on every forward pass (once per layer per microbatch per step), but process groups never change during training — this is pure redundant overhead.

Move the call to `MoELayer.__init__` so it runs once at model construction time.

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+            local_actual_peak = cum_actual.max()
+            local_balanced_peak = cum_balanced.max()
+            ratio_t = torch.tensor(
+                [
+                    (local_actual_peak / (local_balanced_peak + 1e-8)).item()
+                    if local_balanced_peak.item() > 0
+                    else 0.0
+                ],
+                device=fwd_bwd_stacked.device,
+                dtype=torch.float32,
```
Victarry: **[SUGGESTION]** `ratio_t` is constructed by calling `.item()` twice (two CPU-GPU syncs), then placing the Python scalar back into a GPU tensor. This can be avoided entirely by staying on GPU:

```python
cum_overload_ratio = torch.where(
    local_balanced_peak > 0,
    local_actual_peak / (local_balanced_peak + 1e-8),
    local_actual_peak.new_zeros(()),
).unsqueeze(0)
```

Also, the name `ratio_t` is non-idiomatic — `_t` suffix is not a PyTorch convention. Suggest `cum_overload_ratio`.

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+        self, tokens_on_rank: torch.Tensor, local_balanced_token_count: torch.Tensor
+    ) -> None:
+        """Record backward-pass (negated actual and balanced count) for paired cumsums."""
+        self._flush_pending_clear()
+        self._fwd_bwd.append(-tokens_on_rank.detach())
+        self._fwd_bwd_balanced.append(-local_balanced_token_count.detach())
+
+    def _pp_group_for_overload(self) -> Optional[torch.distributed.ProcessGroup]:
+        if not torch.distributed.is_initialized():
+            return None
```
Victarry: **[SUGGESTION]** `_pp_group_for_overload` is a two-line method that adds an indirection layer without any abstraction value. Inline it directly in `report()`:

```python
pp_group = parallel_state.get_pipeline_model_parallel_group(check_initialized=False) \
    if torch.distributed.is_initialized() else None
```

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+    ``clear()`` only marks storage for reset on the next ``record_fwd`` or
+    ``record_bwd`` so tensor handles stay valid until Python runs a recording
+    hook again (e.g. across CUDA graph replay windows that skip those hooks).
+    """
+
+    def __init__(self) -> None:
+        self._fwd: Dict[int, List[torch.Tensor]] = {}  # layer_idx -> list of 0-dim float (tokens on rank)
+        self._fwd_balanced: Dict[int, List[torch.Tensor]] = {}  # same keys as _fwd, balanced token count
+        self._fwd_bwd: List[torch.Tensor] = []
+        # Mirrors _fwd_bwd: +balanced count on forward, - on backward.
```
Victarry: **[SUGGESTION]** Several field names are unclear — suggestions for the four instance variables declared here and on L123-124:

| Current | Suggested | Reason |
|---|---|---|
| `_fwd` | `_layer_fwd_tokens` | Makes clear it's a per-layer dict of forward token counts |
| `_fwd_balanced` | `_layer_fwd_balanced` | Symmetric with above |
| `_fwd_bwd` | `_cumulative_timeline` | This is a time-ordered sequence of +fwd / -bwd events, not a "fwd_bwd" hybrid |
| `_fwd_bwd_balanced` | `_cumulative_balanced_timeline` | Same reasoning |

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+        if self._fwd_bwd:
+            if len(self._fwd_bwd_balanced) != len(self._fwd_bwd):
+                raise ValueError(
+                    f"Overload tracker: _fwd_bwd ({len(self._fwd_bwd)}) and "
+                    f"_fwd_bwd_balanced ({len(self._fwd_bwd_balanced)}) length mismatch."
+                )
+            fwd_bwd_stacked = torch.stack(
+                [t.float() for t in self._fwd_bwd], dim=0
+            )  # [num_events]
+            balanced_bwd_stacked = torch.stack(
```
Victarry: **[SUGGESTION]** `balanced_bwd_stacked` is asymmetric with `fwd_bwd_stacked` — both cover the full forward+backward timeline, not just backward. Rename to `balanced_fwd_bwd_stacked` for consistency.

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+
+        avg_overload_factors = overload_avg
+        max_overload_factors = overload_max
+
+        avg_overload_factor = avg_overload_factors.mean().item()
+        max_overload_factor = max_overload_factors.max().item()
+
+        if use_pp_reduce:
+            cum_in = (
+                float(max_cum_overload_factor)
```
Victarry: **[SUGGESTION]** `cum_in` is cryptic. Rename to `max_cum_value` to make clear it's the `max_cum_overload_factor` value being packed into the PP reduce buffer.

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -467,9 +484,37 @@ def routed_experts_compute(self, hidden_states: torch.Tensor, probs: torch.Tenso
...
+        routing_map_for_balanced_count = None
+        if self.config.log_overload_factor:
+            routing_map_for_balanced_count = self._routing_map_after_token_dispatch()
+
         dispatched_input, tokens_per_expert, permuted_probs = (
             self.token_dispatcher.dispatch_postprocess(hidden_states, probs)
         )
+        if self.config.log_overload_factor and routing_map_for_balanced_count is not None:
+            ws = float(self.tp_ep_group.size())
+            base = float(routing_map_for_balanced_count.shape[0]) * float(
```
Victarry: **[SUGGESTION]** `base` is too generic in this context. Rename to `local_balanced_count` to make clear it represents the balanced token count for this rank before the `tp_ep` SUM reduction.

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+                    ratio_t, group=tp_ep_group, op=torch.distributed.ReduceOp.MAX
+                )
+            if dp_group is not None:
+                torch.distributed.all_reduce(
+                    ratio_t, group=dp_group, op=torch.distributed.ReduceOp.MAX
+                )
+            max_cum_overload_factor = ratio_t.item()
+
+        # tp_ep: max token count per entry across the group, then overload = max / balanced_per_rank.
+        stacked = torch.stack([t.float() for t in fwd_tensors], dim=0)
```
Victarry: **[SUGGESTION]** `stacked` should be `actual_tokens_stacked` to distinguish it from `balanced_stacked` (L291) and `fwd_bwd_stacked` (L248), which follow a more descriptive naming pattern.

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+            overload_max = tp_ep_overload.clone()
+            torch.distributed.all_reduce(
+                overload_max, group=dp_group, op=torch.distributed.ReduceOp.MAX
+            )
+        else:
+            overload_avg = tp_ep_overload
+            overload_max = tp_ep_overload
+
+        avg_overload_factors = overload_avg
+        max_overload_factors = overload_max
```
Victarry: **[SUGGESTION]** `avg_overload_factors` and `max_overload_factors` are unnecessary aliases for `overload_avg` and `overload_max` — they add two lines of indirection with no abstraction benefit. Use `overload_avg` / `overload_max` directly in all subsequent references.

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+                float(max_cum_overload_factor)
+                if max_cum_overload_factor is not None
+                else 0.0
+            )
+            pp_buf = torch.tensor(
+                [max_overload_factor, cum_in],
+                device=device,
+                dtype=torch.float32,
+            )
+            torch.distributed.all_reduce(
```
Victarry: **[SUGGESTION]** `report()` contains 6–7 sequential `all_reduce` calls across the same groups. Some can be fused to reduce collective launch overhead:

- **tp_ep**: `all_reduce(max_actual, MAX)` (L283) and `all_reduce(balanced_stacked, SUM)` (L295) can be packed into one call by stacking both tensors.
- **dp**: `all_reduce(overload_avg, AVG)` (L303) and `all_reduce(overload_max, MAX)` (L307) operate on the same data — consider a single fused reduce.
- **`ratio_t`**: two sequential all_reduces (tp_ep MAX L268, dp MAX L272) could be deferred and folded into the tp_ep and dp passes above.

At scale (hundreds of MoE layers, large DP), reducing the number of collectives per logging step is meaningful.

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+        self._fwd.clear()
+        self._fwd_balanced.clear()
+        self._fwd_bwd.clear()
+        self._fwd_bwd_balanced.clear()
+
+    def _flush_pending_clear(self) -> None:
+        if self._pending_clear:
+            self._pending_clear = False
+            self._clear_storage()
+
```
Victarry, line 142: **[SUGGESTION]** The `_pending_clear` deferred-clear mechanism adds complexity that does not deliver its intended benefit.

The rationale (from the class docstring) is to keep tensor handles valid during CUDA graph replay windows. However, CUDA graph replay does not re-execute Python-side autograd functions — `record_fwd` and `record_bwd` are never called during replay. So the tracker never receives new data during replay regardless of whether storage has been cleared, making the deferred-clear protection moot.

For non-CUDA-graph training the extra state (`_pending_clear` flag + `_flush_pending_clear()` call in every `record_fwd`/`record_bwd`) is pure overhead with no benefit.

**Suggestion**: Have `clear()` call `_clear_storage()` directly and remove `_pending_clear` and `_flush_pending_clear()`.

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+        self._fwd_bwd.append(-tokens_on_rank.detach())
+        self._fwd_bwd_balanced.append(-local_balanced_token_count.detach())
+
+    def _pp_group_for_overload(self) -> Optional[torch.distributed.ProcessGroup]:
+        if not torch.distributed.is_initialized():
+            return None
+        g = parallel_state.get_pipeline_model_parallel_group(check_initialized=False)
+        return g
+
+    def report(
```
Victarry: **[SUGGESTION]** `report()` is ~200 lines and mixes several independent responsibilities: tp_ep reduction, dp reduction, cumsum peak computation, pp reduction, TensorBoard/W&B logging, and log-string assembly. This makes it hard to test, read, and debug any single concern.

Suggest splitting into focused private methods:
- `_reduce_tp_ep(fwd_tensors, balanced_tensors)` → returns `tp_ep_overload`
- `_reduce_dp(tp_ep_overload)` → returns `(overload_avg, overload_max)`
- `_compute_max_cum_overload()` → returns `max_cum_overload_factor`
- `_reduce_pp(max_overload, max_cum)` → returns pp-reduced scalars
- `_log_to_writers(writer, wandb_writer, scalars, iteration)` → writes TB/W&B

`report()` then becomes a thin orchestrator responsible only for call ordering.

megatron/core/transformer/moe/moe_logging.py
```diff
@@ -66,6 +66,334 @@ def destroy_moe_metrics_tracker() -> None:
...
+    ``_fwd_bwd`` / ``_fwd_bwd_balanced`` mirror interleaved fwd/bwd so cumulative
+    peaks of actual vs balanced token counts can be compared.
+
+    Lifecycle: set_process_groups() and record_fwd/record_bwd during forward
+    (SaveOverloadFactorFunction in MoELayer) → report() at step end
+    (sync, aggregate, log, deferred clear) → repeat.
+
+    ``clear()`` only marks storage for reset on the next ``record_fwd`` or
+    ``record_bwd`` so tensor handles stay valid until Python runs a recording
+    hook again (e.g. across CUDA graph replay windows that skip those hooks).
```
Victarry: Comments could be refactored with AI for better readability

Example comment refactored by GPT
```
    """Track MoE overload-factor metrics.

    Recorded values
    ---------------
    - Per-layer actual tokens on this rank after dispatch:
      tokens_per_expert.sum().
    - Per-layer balanced token count before dispatch:
      routing_map.shape[0] * moe_router_topk (read after token_dispatch).
    - Both values are captured by an autograd hook on dispatched_input.
    - _fwd_bwd and _fwd_bwd_balanced mirror interleaved fwd/bwd events so
      cumulative peaks of actual vs balanced token counts can be compared.

    How report() aggregates
    -----------------------
    1. In tp_ep_group, run all_reduce(MAX) on per-rank actual totals.
    2. Divide by balanced tokens per rank (summed local balanced counts / group size)
       to get per-entry tp_ep overload.
    3. In dp_group, run all_reduce(AVG) and all_reduce(MAX) on overload
       before scalar summaries.
    4. In the pipeline-parallel group, max and max_cum use all_reduce(MAX) so
       every stage agrees on the worst overload. Ranks without MoE layers
       contribute 0.
    5. Mean overload is not reduced across PP; each rank logs its local mean
       (0 if nothing was recorded).

    Lifecycle
    ---------
    set_process_groups() and record_fwd()/record_bwd() are called during forward
    (SaveOverloadFactorFunction in MoELayer). report() runs at step end
    (sync, aggregate, log, deferred clear), then the cycle repeats.

    clear() behavior
    ----------------
    clear() does not immediately reset storage. It marks storage for reset on
    the next record_fwd() or record_bwd() so tensor handles stay valid until
    Python executes a recording hook again (for example across CUDA graph replay
    windows that skip those hooks).
```

megatron/core/transformer/moe/moe_utils.py
```diff
@@ -978,6 +981,87 @@ def clear_aux_losses_tracker() -> None:
...
+    @staticmethod
+    def backward(ctx, grad_output):
+        """Backward pass: append negated actual and balanced count for paired cumsums."""
+        if ctx.saved_tensors:
+            tokens_on_rank, balanced = ctx.saved_tensors
+            get_moe_overload_factor_tracker().record_bwd(tokens_on_rank, balanced)
+        return grad_output, None, None, None
+
+
+def save_overload_factor_to_tracker(
```
Victarry: **[SUGGESTION]** `save_overload_factor_to_tracker` and `SaveOverloadFactorFunction` are misleadingly named — neither computes nor saves an overload factor. What they actually do is **record post-dispatch token counts** (actual tokens on this rank + balanced token count) into the tracker. The overload factor itself is only computed later in `report()`.

Suggested renames:
- `SaveOverloadFactorFunction` → `RecordDispatchTokenCountsFunction`
- `save_overload_factor_to_tracker` → `record_dispatch_token_counts`

Victarry (commented):
Hi @nanz-nv , thanks for this PR, it would be very useful for tracing MoE routing information. 

Above are mixed comments by me and Claude Code, please take a look~

megatron/core/transformer/moe/moe_logging.py
nanz-nv: >> However, CUDA graph replay does not re-execute Python-side autograd functions — record_fwd and record_bwd are never called during replay. 

during `record_fwd` and `record_bwd`, the tensors for holding token counts are being recorded and tracked in a python list. Once a CUDA graph is captured, the kernels in the graph continues to write to these tensors so the overload tracker outside can compute statistics on these tensors to get overload factor. If `clear()` is called outside the CUDA graph, we lose track of these tensors.

megatron/core/transformer/moe/moe_layer.py
nanz-nv: This way of creating tensor is intentional to make it CUDA graph compatible. torch.tensor is not graph-safe

megatron/core/transformer/moe/moe_logging.py
nanz-nv: >> tp_ep: all_reduce(max_actual, MAX) (L283) and all_reduce(balanced_stacked, SUM) (L295) can be packed into one call by stacking both tensors.
These two cannot be fused because one is max and one is sum. Similarly for the dp allreduce.
For ratio_t, all_reduce is conditional while the other is unconditional. So for modularity, I'd prefer to separate them for now.

megatron/core/transformer/moe/moe_layer.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_layer.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_layer.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_utils.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_utils.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_logging.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_logging.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_logging.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_logging.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_logging.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_layer.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_logging.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_logging.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_logging.py
nanz-nv: Thanks for the suggestion. Suggestions taken. I refactored report a bit and separate them into smaller internal functions.

megatron/core/transformer/moe/moe_logging.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_utils.py
nanz-nv: Thanks for the suggestion. Suggestions taken.

megatron/core/transformer/moe/moe_logging.py
Victarry: Got it! Thanks for the explanation.

Victarry:
Hi, @nanz-nv, could you please fix the linting error and make sure the CI is passed?

Should also submit a PR to the main branch, please~

Victarry:
Please also add the `--log-overload-factor` to one of the functional test configs in `tests/functional_tests/test_cases/moe` to make sure this feature is tested.

Victarry (approved):

PR #4125, opened 2026-04-03
RL: prevent unnecessary entry into inference mode

tdene:
# What does this PR do ?

During the inference stage of RL, rollouts are generated.

We then switch to training to train.

We then switch back to inference to get more rollouts.

What if we already have enough rollouts though? Why would we switch back to inference? We don't need to!

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/rl/agent/api.py
+++ b/megatron/rl/agent/api.py
@@ -1,15 +1,9 @@
 # Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
-import asyncio
-import time
 from abc import ABC, abstractmethod
-from typing import AsyncIterator, Awaitable, Callable, Generic, NamedTuple, TypeVar
+from typing import Awaitable, Callable, Generic, NamedTuple, TypeVar
 
-import numpy as np
 from pydantic import BaseModel
-
-from megatron.core.inference.utils import asyncio_Queue, asyncio_QueueShutDown
-from megatron.core.utils import trace_async_exceptions
 
 from ..__init__ import Request, TypeLookupable
 from ..inference import (
@@ -17,14 +11,8 @@ from ..inference import (
     InferenceRequest,
     InferenceResponse,
     LLMChatMessage,
-    ReturnsRaw,
 )
-from ..rollout_granularity import (
-    RELEASE_STATE_BY_SUBMISSION,
-    ConsumptionGranularity,
-    ReleaseState,
-    SubmissionGranularity,
-)
+from ..rollout_granularity import ConsumptionGranularity, SubmissionGranularity
 
 
 class AgentBaseModel(BaseModel, extra='allow'):
@@ -110,6 +98,7 @@ class GroupRolloutParams(NamedTuple):
 
     inference_request: InferenceRequest
     build_rollout: Callable[[InferenceResponse], Awaitable[Rollout]]
+    agent: "Agent"
 
 
 class ContrastiveRollout(AgentBaseModel):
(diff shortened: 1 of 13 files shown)


megatron/rl/agent/api.py
```diff
@@ -299,6 +299,49 @@ async def shutdown_queue_when_done():
                 task.cancel()
 
 
+class RolloutStream(AsyncIterator):
```
yobibyte: Since it's an iterator now, can we just send this or a sync version of this as a data iterator to training.py and do not have rl-specific code in training.py?

megatron/rl/rl_utils.py
```diff
@@ -584,6 +563,67 @@ def get_environment_rollouts(
...
+    # NOTE(jbarker): we need to double check this when using PP>1
+    rank = torch.distributed.get_rank()
+
+    inference_pg_collection = get_attr_wrapped_model(model[0], "pg_collection")
+    pg_size = get_pg_size(inference_pg_collection.ep)
+    assert (n_prompts % pg_size == 0), f"{n_prompts=} must be divisible by {pg_size=}"
+
+    fast_drained = False
+
+    # Fast path: when partial rollouts are enabled and the generator already has
```
yobibyte: I do not like that there are two places that decide whether we should sample or not.

First place is the call site (like 1611), which checks that we need to sample due to `pseudo async` logic.  If that logic decides that we *should*, we have another case for partial rollouts here. 

I think, a good design will be to move all the 'should we actually sample now?' logic to a separate function and will call this before entering this function. WDYT?

This is probably a good time for us to double check the consistency of `pseudo async` with partial rollouts on and decide what behaviour we want to allow, and what not. E.g. do we want to allow / maintain sampling more than we consume with partial rollouts? If no, we should prob add an assert at the arg validation if it does not yet exist.

megatron/rl/agent/api.py
yobibyte: This is a big ask, and probably should be done in a separate PR, I was just pointing out what I have always wanted to do since I implemente the grpo loop, but could not find the time to.

tdene: Been looking and thinking about this.

I think it is doable, but it will still be too complicated and is ultimately out-of-scope for this PR.

megatron/rl/rl_utils.py
tdene: I also had the same dislike, and that's what I tried to solve with the factoring out of the `colocated_inference` method.

Just writing this next part out for myself. The key issue here is that we need a `torch.distributed.broadcast_object_list` under `sync-rollouts`, inside of `get_environment_rollouts`.     
In `main`, `get_environment_rollouts` always does `do the colocated inference stuff` -> `broadcast_object_list`. In this branch, `get_environment_rollouts` sometimes skips the `do the colocated inference stuff` part, but still needs to run `broadcast_object_list` in the end.

Now, to address your point. Yes, we have two different checks. But at the time of the first check, ALL ranks have the requisite data to do the check. At the time of the second check, only rank 0 has the requisite data, and no matter what the second check does, rank 0 then has to communicate with the rest of the ranks.

I haven't been able to see how they can be merged. Going to post this reply here and then try to tweak the code some more...

tdene: I do not see how it can be done better than this. If we move the 2nd check out of `get_environment_rollouts`, too much code has to come with it.

I think this is the best we're going to get: `get_environment_rollouts` is the method that we call when non-0 ranks do not have sufficient rollouts. The method then asks rank 0 if it personally has enough rollouts; if rank 0 does not, we do inference; if rank 0 does, we skip to the part where we share rank 0's rollouts with the non-0 ranks.

I have, however, added this explanation to the docstring of `get_environment_rollouts`. Please take a look at the latest commit

tdene: On your second point, of double-checking the consistency, I would strongly prefer to wait until this entire batch of PRs is merged.

#4127 is a MASSIVE quality-of-life improvement to the args. I want to see it plus #4125 in `main`, and brainstorm a bit, before we make any judgments. Especially because inconsistent behavior is not going to be inherently incorrect, it'll just be undesirable.

yobibyte: I think, the main problem in there is that checking 'should we run inference' logic for partial rollouts is 'should we run inference? *get rollouts if not right now*' logic. This makes it harder to disentangle.

If we could peek at the RolloutStream to get its length, we could make the logic smth like that:
```
def should_we_sample():
    if (
        buffered_rollouts is None or
    ):
         if partial_rollouts:
                # all broadcasting logic to get the flag goes here too.
                # check if the ROLLOUT_GENERATOR len is large enough
          else:
               return True
    else:
         return False
```

Then, in get_grpo_data_iterator we will run:
```
if should_we_sample():
     if partial_rollouts:
          sample partial_logic
     else:
          sample non_partial
else:
     if partial:
           return drain()
    else: 
          return buffered_rollouts
```

How does this look?

tdene: Sorry for not understanding properly earlier. I was hung up on the broadcast having to go into `should_we_sample`.

Take a look at the newest state of the PR; I think I've addressed what you described. Let me know what you think.

megatron/rl/rl_utils.py
```diff
@@ -473,23 +474,61 @@ def get_rollout_generator(args, inference_interface, n_prompts, samples_per_grou
...
+    The verdict of whether or not inference is needed must be broadcast from rank 0.
 
     Returns:
-        GroupedRollouts object which is a nested list with each element being a list of rollouts of a group.
+        (need_new_rollouts, need_inference):
+            need_new_rollouts denotes whether we should enter get_environment_rollouts.
+            need_inference denotes whether we should enter colocated_inference to collect rollouts.
+    """
+    global_batches_per_collection = (grpo_prompts_per_step * grpo_group_size) // global_batch_size
+    if not (
```
yobibyte: I would swap this to if buf is not None and ... The negation of or is harder to parse IMO.

tdene: Done!

jalbericiola:
I am happy we are doing this!

megatron/rl/rl_utils.py
tdene: Fixed!

tdene: We can peek at RolloutStream now because the entire codebase has been refactored and the new version of "RolloutStream" is inherently peekable!

lauradang (approved):
LGTM

PR #4133, opened 2026-04-03
Fix incorrectly set decoupled_grad and DistOpt mechanics for MFSDP.

cspades:
# What does this PR do ?

- Update `DistributedOptimizer` to correctly use `decoupled_grad` with Megatron-FSDP. (FSDP will always use decoupled gradients when using FusedAdam, which simplifies the logic a lot.)

TL;DR
```
use_decoupled_grad=self.config.use_precision_aware_optimizer_no_fp8_or_ds_fp8
or (
    # Megatron-FSDP always uses decoupled_grad with FusedAdam.
    self.config.use_precision_aware_optimizer
    and getattr(params[0], "__fsdp_param__", False)  # == ddp_config.use_megatron_fsdp
),
```

## Details

- Megatron-FSDP does not use FusedAdam's master weights, but Megatron-LM hard-codes `master_weights=True` if `OptimizerConfig.use_precision_aware_optimizer_no_fp8_or_ds_fp8` / `use_decoupled_grad` are True. This PR turns off FusedAdam master weights when using Megatron-FSDP, as FusedAdam should only provide an `optimizer.step()` to Megatron-FSDP's `DTensor(FP32/BF16)` main weights.
- Improved Megatron-FSDP argument validation and added `NotImplementedError` for completely unexpected paths in the distributed optimizer.

## Testing

- Adding E2E `--use-precision-aware-optimizer` unit test to temporarily guarantee functionality. Cross-tested with this TE E2E test: https://github.com/NVIDIA/TransformerEngine/pull/2845
- Using HFSDP + FP8 delayed scaling, we get a lot of memory back by turning off `master_weights`:

```
# No FusedAdam Master Weights (HFSDP + FP8 Delayed Scaling)
[Rank 0] (after 2 iterations) memory (MB) | allocated: 19595.55 | max allocated: 27022.17 | reserved: 22504.00 | max reserved: 30774.00
[2026-04-03 12:37:38.220805] iteration      100/15258789 | consumed samples:        12800 | elapsed time per iteration (ms): 3263.1 | throughput per GPU (TFLOP/s/GPU): 230.1 | learning rate: 4.915198E-07 | global batch size:   128 | lm loss: 5.473558E+00 | loss scale: 1.0 | grad norm: 9.413 | num zeros: 0 | number of skipped iterations:   0 | number of nan iterations:   0 |

# FusedAdam Master Weights (HFSDP + FP8 Delayed Scaling)
[Rank 0] (after 2 iterations) memory (MB) | allocated: 23425.69 | max allocated: 30852.31 | reserved: 26344.00 | max reserved: 34788.00
[2026-04-03 12:44:12.848377] iteration      100/15258789 | consumed samples:        12800 | elapsed time per iteration (ms): 3149.4 | throughput per GPU (TFLOP/s/GPU): 238.4 | learning rate: 4.915198E-07 | global batch size:   128 | lm loss: 5.476564E+00 | loss scale: 1.0 | grad norm: 9.424 | num zeros: 0 | number of skipped iterations:   0 | number of nan iterations:   0 |
```

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/training/training.py
```diff
@@ -1467,7 +1467,17 @@ def build_model():
...
+            kwargs['megatron_fsdp_use_decoupled_grad'] = args.use_precision_aware_optimizer and (
+                # NOTE(@cspades): Follow the same __post_init__ as in OptimizerConfig:
+                #   OptimizerConfig.use_precision_aware_optimizer_no_fp8_or_ds_fp8
+                # Megatron-FSDP only needs to be consistent with FusedAdam(use_decoupled_grad=?),
+                # and because the OR logic of use_precision_aware_optimizer_no_fp8_or_ds_fp8
+                # isn't particularly restrictive, Megatron-FSDP uses decoupled gradients under
+                # the same conditions that the distributed optimizer uses decoupled gradients.
+                args.main_params_dtype != torch.float32
+                or (args.fp8_recipe is None or args.fp8_recipe == "delayed")
+                or args.optimizer_cpu_offload
```
shjwudp: Do we really need to follow this combination of conditions? 🤔
Unless there are some specific constraints, I think we’d better avoid using such a complicated setup — it makes maintenance harder.

Theoretically, M-FSDP shouldn’t be restricted by these conditions and can freely use `decoupled_grad`.

shjwudp (approved):
Overall LGTM, just have a little concern about `megatron_fsdp_use_decoupled_grad`.

This PR avoids FusedAdam from redundantly creating master weights when M-FSDP already maintains FP32 main weights, which is the correct way to use FusedAdam under M-FSDP (in terms of memory usage).

megatron/training/training.py
cspades: Not sure if I can change the current logic for non-MFSDP, so if there are no constraints on the Megatron-FSDP side, I think matching the behavior should be inconsequential, right?

Otherwise, I will likely need to do something like this:

- If Megatron-FSDP, use a separate `decoupled_grad` argument to determine if it is used.
  - `if use_megatron_fsdp and use_precision_aware_optimizer: ...`
- If not Megatron-FSDP, follow the MLM logic where only BF16 and FP8 DS use it.
  - `if use_precision_aware_optimizer_no_fp8_or_ds_fp8: ...`

Another thing I can do is to move the `OptimizerConfig.__post_init__` logic outside so we can use the same variable for the DP wrapper and the optimizer, but it won't make it significantly easier to maintain.

cspades: Maybe this is the cleanest solution?
```
# Optimizer Functions
if use_precision_aware_optimizer_no_fp8_or_ds_fp8 or (
    use_megatron_fsdp and use_precision_aware_optimizer
):
    # Make sure no use of FusedAdam master weights when using Megatron-FSDP.
```
And Megatron-FSDP can just always use `decoupled_grad` when using FusedAdam, or it can be controlled using a new argument but in my opinion users do not need to care about this.

shjwudp: I think this approach is better — it avoids complex condition checks later that might make us wonder why M-FSDP doesn’t support using `decoupled_grad` in certain cases.

yaoyu-33 (approved):

megatron/training/arguments.py
```diff
@@ -961,17 +977,14 @@ def validate_args(args, defaults={}):
...
         assert args.ckpt_format == "fsdp_dtensor", \
-            "Megatron FSDP only supports fsdp_dtensor checkpoint format"
+            "Megatron-FSDP requires using the `fsdp_dtensor` checkpointing format."
         
     if args.fsdp_manual_registration:
-        assert args.use_megatron_fsdp, "FSDP manual registration is only supported with Megatron FSDP"
-        assert args.nccl_ub, "FSDP manual registration is only supported with nccl-ub option"
-
-        if args.use_megatron_fsdp:
-            args.reuse_grad_buf_for_mxfp8_param_ag = False
```
cspades, line 974: This looks like a rebase error, we should NEVER have this be True, not just when using NCCL UB... It should be in the above set of argument validations.

megatron/training/arguments.py
```diff
@@ -950,9 +950,25 @@ def validate_args(args, defaults={}):
...
 
-        if args.data_parallel_sharding_strategy in ["optim_grads_params", "optim_grads"]:
+        # Megatron-FSDP requires the DistributedOptimizer.
+        if not args.use_distributed_optimizer:
             warn_rank_0(
-                'Please make sure your TransformerEngine support FSDP + gradient accumulation fusion',
+                'Megatron-FSDP is only compatible with --use-distributed-optimizer. Using DistributedOptimizer...',
+                args.rank,
+            )
+        args.use_distributed_optimizer = True
```
cspades, line 959: We should automatically set this for users? We don't support other optimizers, neither layer-wise nor non-distributed.

cspades: I'm 99% sure we should. We don't support any other optimizer with FSDP.

megatron/training/training.py
cspades: @shjwudp LGTM'd offline, resolving this.

jaredcasper (approved):

cspades:
Random unit test failure in merge queue: https://github.com/NVIDIA/Megatron-LM/actions/runs/24220276331/job/70712019031

🤷🏻 Sending it back in.

cspades:
Previous unit test passed: https://github.com/NVIDIA/Megatron-LM/actions/runs/24222278675/job/70717988646

Then something else randomly failed: https://github.com/NVIDIA/Megatron-LM/actions/runs/24222278675/job/70717988616

Rerunning this test to be sure: https://github.com/NVIDIA/Megatron-LM/actions/runs/24222278675/job/70801031886 (✅ 🤷🏻‍♂️)

Issue #4124, opened 2026-04-03
[QUESTION] forward_step output destructor performance issues


Baibaifan:
The `forward_step` return value has performance issues in destructuring procedures. The performance loss is significant during multiple gradient accumulations.

### Description

For example: `forward_backward_no_pipelining`

https://github.com/NVIDIA/Megatron-LM/blob/dcc6d62103c8cab5f41e7e1d23d2a43361001282/megatron/core/pipeline_parallel/schedules.py#L686

The perf function revealed significant scheduling gaps and graph destruction time consumption during microbatches iterations.

The `output_tensor` variable generated in each iteration is not immediately released after `backward_step` is completed. Due to the local variable overwriting mechanism, the reference count of the `output_tensor` from the previous iteration is implicitly decremented by 1 only when `forward_step` is called and the value is reassigned in the next iteration.

This causes the cleanup and memory destruction operations of the previous iteration's autograd graph/nodes to be delayed until the beginning of the next microbatch, leading to race conditions and observable time consumption (such as several milliseconds of CPU scheduling downtime), thus slowing down the overall timeline parallel efficiency.

<img width="1542" height="1070" alt="Image" src="https://github.com/user-attachments/assets/894fa690-9c1e-4dfb-9e8d-4c7d4e7e7a73" />

### Impact
Qwen3.5 w35a3

<img width="2398" height="306" alt="Image" src="https://github.com/user-attachments/assets/e32ce8fe-3ad4-408b-8b2d-5ca0603a84f7" />

Qwen3 w30a3

reference: https://github.com/yanring/Megatron-MoE-ModelZoo/blob/main/model_configs/benchmarking/Qwen3-30B-A3B.yaml 

layers 48 -> 24

<img width="429" height="180" alt="Image" src="https://github.com/user-attachments/assets/eb32993c-e1b0-4bab-949c-d53551de4ad3" />

Phlip79:
Can you please provide a reproducer? We have recipes for Qwen3 defined in [Megatron-Bridge](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/recipes/qwen/qwen3_moe.py) that you can use.

Baibaifan:
[...]

PR #4153, opened 2026-04-06
[opt] Optimize the output destructor problem.

Baibaifan:
# What does this PR do ?
Optimize the output destructor problem.

Reference Questions: https://github.com/NVIDIA/Megatron-LM/issues/4124

Before optimization:
image

After optimization:
image


ilml:
hi @Baibaifan , thanks for the fix, this is indeed a performance issue.
but i do have concern on your solution:
1. we don't need to keep the reference through out the for loop right
2. due to GIL, thread is not real async, it will compete with the main thread. and you need to join the thread.

I think we only need to add `del output_tensor` after backward_step, wdyt

ilml (changes requested):
left some comments

Baibaifan:
> del output_tensor

hi, @ilml. 
Deleting `output_tensor` will trigger a destruction issue on the next forward pass, resulting in a performance penalty. Our tests showed that accumulating multiple times and triggering the destruction in an asynchronous thread at once is the method with the least performance penalty.


image

image

**The time added before the forward metering was 860.773 us -> 1.487 ms.**

image

**Multiple destructions take only 1.355ms.**

We also tried using `jemalloc`, which resulted in some improvements without changing the code.

ilml:
> > del output_tensor
> 
> hi, @ilml. Deleting `output_tensor` will trigger a destruction issue on the next forward pass, resulting in a performance penalty. Our tests showed that accumulating multiple times and triggering the destruction in an asynchronous thread at once is the method with the least performance penalty.
> 
> image image
> **The time added before the forward metering was 860.773 us -> 1.487 ms.**
> 
> image
> **Multiple destructions take only 1.355ms.**
> 
> We also tried using `jemalloc`, which resulted in some improvements without changing the code.

sg, thanks for the great work!

ilml (approved):
lgtm

Baibaifan:
> lgtm

hi, @ilml @gautham-kollu 

I modified the pylint format. Please help me rerun the ci/cd test.

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -687,6 +689,15 @@ def forward_backward_no_pipelining(
                 total_num_tokens += num_tokens
                 if not forward_only:
                     backward_step(input_tensor, output_tensor, output_tensor_grad, config)
+                    # Free the output tensor from the forward pass since it's no longer needed.
```
yashaswikarnati, line 692: sorry if I'm simplifying things here or missing something.

1) I'm curious if this increases peak or mem usage - as we keep some ref to output tensor in the list, do we know the autograd behaviour? 

2) also what about just trying the existing deallocate_output_tensor

3) also wonder if just del output_tensor would be enough here

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -687,6 +689,15 @@ def forward_backward_no_pipelining(
...
+                    output_tensors_to_free.append(output_tensor)
+                    output_tensor = None
+
+            def _async_free(tensors: List[torch.Tensor]):
+                tensors.clear()
+
+            threading.Thread(target=_async_free, args=(output_tensors_to_free,)).start()
```
yashaswikarnati, line 699: seems like we spawn this thread even in case of just forward pass, no bacvkward needed ?

megatron/core/pipeline_parallel/schedules.py
Baibaifan: hi, @yashaswikarnati 
1. Our actual measurements show that the `output_tensor` is not very large; it is just a `loss tensor`.
2. `deallocate_output_tensor` might also work, but we haven't tried it.
3. Deleting `output_tensor `will trigger a destruction issue on the next forward pass, resulting in a performance penalty. Our tests showed that accumulating multiple times and triggering the destruction in an asynchronous thread at once is the method with the least performance penalty. https://github.com/NVIDIA/Megatron-LM/pull/4153#issuecomment-4219573433

megatron/core/pipeline_parallel/schedules.py
Baibaifan: > spawn

This asynchronous thread is only used to destruct the `output_tensor`. If you want to reproduce this problem, you can refer to the question I raised. https://github.com/NVIDIA/Megatron-LM/issues/4124#issuecomment-4196937769

PR #4154, opened 2026-04-06
fix(megatron-fsdp): preserve non-meta tensors during meta device materialization

xuwchen:
# What does this PR do ?
During meta device initialization, the materialization path calls `m.to_empty(device, recurse=False)` before `reset_parameters()`. `to_empty()` unconditionally replaces all tensors (both parameters and buffers) in the module with uninitialized memory, regardless of whether they are actually on the meta device.

This becomes a problem for MoE when `--moe-router-enable-expert-bias` is enabled: `expert_bias` is registered with an explicit `device=torch.cuda.current_device()` [[link](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/moe/router.py#L183-L189)], so it bypasses the meta context and is correctly initialized to zeros on GPU. But `to_empty()` clobbers it with uninitialized memory (NaN), and `reset_parameters()` only reinitializes parameters, not buffers, so `expert_bias` stays NaN for the entire training run. 

The fix here replaces `m.to_empty()` with a targeted helper that only materializes `is_meta=True` tensors, leaving non-meta buffers (like `expert_bias`) untouched.

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

[...]

PR #4155, opened 2026-04-06
[DEV] fix(megatron-fsdp): preserve non-meta tensors during meta device materialization

xuwchen:
main PR: #4154 

# What does this PR do ?
During meta device initialization, the materialization path calls `m.to_empty(device, recurse=False)` before `reset_parameters()`. `to_empty()` unconditionally replaces all tensors (both parameters and buffers) in the module with uninitialized memory, regardless of whether they are actually on the meta device.

This becomes a problem for MoE when `--moe-router-enable-expert-bias` is enabled: `expert_bias` is registered with an explicit `device=torch.cuda.current_device()` [[link](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/moe/router.py#L183-L189)], so it bypasses the meta context and is correctly initialized to zeros on GPU. But `to_empty()` clobbers it with uninitialized memory (NaN), and `reset_parameters()` only reinitializes parameters, not buffers, so `expert_bias` stays NaN for the entire training run. 

The fix here replaces `m.to_empty()` with a targeted helper that only materializes `is_meta=True` tensors, leaving non-meta buffers (like `expert_bias`) untouched.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -104,6 +104,19 @@
...
 
 
+def _materialize_meta_tensors(module: torch.nn.Module, device: torch.device) -> None:
+    """
+    Materialize only meta tensors in a module, preserving non-meta tensors that are already initialized on device.
+    """
+    for name, param in module.named_parameters(recurse=False):
+        if param.is_meta:
+            new = torch.empty_like(param, device=device)
+            setattr(module, name, torch.nn.Parameter(new, requires_grad=param.requires_grad))
```
shjwudp: Please do not reconfigures module parameters, which might cause loss of existing attributes on those parameters or invalidate maps that used these parameters as keys.

shjwudp (commented):
Good catch! Some bias parameters are stored as `nn.Module` buffers (though that’s not a standard PyTorch practice). These buffers weren’t initialized on the meta device and shouldn’t be converted to empty tensors. This PR fixes that issue and also resolves the unexpected NaNs observed in the functional tests with the MoE layer.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
xuwchen: Updated the implementation to use `Module._apply()` with an `is_meta` guard, which is the same mechanism that `to_empty()` uses internally. For meta→cuda conversion, `_apply` always creates new Parameter objects; this is a PyTorch behavior, not something we introduced. The existing `_reset_parameters` already handles dict remapping and attribute copying for this case.

shjwudp (approved):

yaox12 (approved):

PR #4160, opened 2026-04-06
Add /split-pr Claude Code command for splitting PRs by CODEOWNERS

Phlip79:
## Summary

Adds a `/split-pr` Claude Code slash command that helps split large PRs into smaller, logically separate PRs to reduce the number of required CODEOWNERS reviewer groups.

The command:
1. Analyzes a PR's changed files against `.github/CODEOWNERS`
2. Proposes a split strategy that minimizes expert groups per PR
3. Executes the split (creating new branches and draft PRs)


.claude/skills/split-pr/SKILL.md
```diff
@@ -0,0 +1,61 @@
...
+4. Build a summary table grouped by CODEOWNERS group, showing which files pull in which groups.
+5. Count the total number of distinct reviewer groups the PR currently requires.
+
+### 2. Propose a split that minimizes reviewer groups per PR
+
+The primary optimization goal: **minimize the number of CODEOWNERS reviewer groups required for each resulting PR**.
+
+Strategy:
+1. Cluster files by their CODEOWNERS groups. Files owned by the same set of groups naturally belong together.
+2. Identify the largest cluster — this becomes the first (and usually largest) PR.
```
maanug-nv, line 32: i think this will work for PRs that make multiple independent changes, eg a new inference feature and new mamba perf feature in the same PR. 

but there are often large PRs focused on a single feature/change that have to touch various parts of the codebase, eg core, training, megatron rl, etc. I think this is the more common case right? these can still be split up, despite the dependencies. claude would need to understand the dependencies, group, and order the changes. so following the previous example, the goal could be changes to core in the first pr, training in the next, rl in the last. 
wdyt? are the current instructions enough to handle cases like that? i think it might be necessary to point claude to identifying dependencies earlier on to determine the best way to split things up rather than just choosing the largest group first?

.claude/skills/split-pr/SKILL.md
```diff
@@ -0,0 +1,61 @@
...
+- Dependencies on other PRs (if any)
+
+Wait for user approval before proceeding.
+
+### 3. Execute the split (after user approval)
+
+For each new PR:
+1. Create a new branch from the appropriate base (`main`, or a dependency PR's branch).
+2. Extract the relevant changes: `git diff upstream/main..<source-branch> -- <file paths> | git apply`.
+3. Stage, commit with a clear message, and push to the user's fork.
```
maanug-nv, line 50: is there any way to preserve the original author's commits or github id in the split PRs?

Phlip79: Good point. I updated the prompt to mention the original author if the user is not the same person.

.claude/skills/split-pr/SKILL.md
Phlip79: Current instructions do this because it tries breaking up PRs based off the CODEOWNERS file. So for example for https://github.com/NVIDIA/Megatron-LM/pull/3430, here is what the skill suggests: 
Screenshot 2026-04-09 at 5 35 29 PM

maanug-nv: ok this seems reasonable. i think we can merge for now.

PR #4222, opened 2026-04-09
fix: store_param_remainders int16 dtype assertion in dp_zero checkpoint load

pandengyao:
## Summary

- When `store_param_remainders=True` (TE >= 2.1.0 default), the dp_zero checkpoint load path passes float32 tensors to TE's `set_scaled_state("master_param", v)`, which asserts `v.dtype == torch.int16`, causing a crash
- Add float32-to-int16 conversion in `_set_main_param_and_optimizer_states` with a range check to catch mismatched `store_param_remainders` configurations
- Add two unit tests: dp_zero roundtrip test and range check error path test

## Root Cause

The dp_zero save path calls `get_unscaled_state("master_param")` which returns the raw int16 remainder tensor, then implicitly upcasts to float32 via `.copy_()` into float32 buffers. On load, these float32 tensors are passed directly to `set_scaled_state()` which asserts int16 dtype.

## Changes

1. **`megatron/core/optimizer/distrib_optimizer.py`**: In `_set_main_param_and_optimizer_states`, convert float32→int16 before calling `set_scaled_state` when `store_param_remainders=True` and `param.dtype == bfloat16`. Includes a range check to detect mismatched checkpoint configurations (e.g., loading a checkpoint saved with `store_param_remainders=False`).

2. **`tests/unit_tests/test_optimizer.py`**: Add two new tests:
   - `test_distrib_optimizer_store_param_remainders_dp_zero_roundtrip`: Verifies dp_zero save/load roundtrip preserves int16 remainders
   - `test_distrib_optimizer_store_param_remainders_range_check`: Verifies ValueError is raised when out-of-range values are detected

## Test plan

- [ ] `torchrun --nproc_per_node=1 -m pytest tests/unit_tests/test_optimizer.py::test_distrib_optimizer_store_param_remainders_dp_zero_roundtrip -v`
- [ ] `torchrun --nproc_per_node=1 -m pytest tests/unit_tests/test_optimizer.py::test_distrib_optimizer_store_param_remainders_range_check -v`
- [ ] `torchrun --nproc_per_node=1 -m pytest tests/unit_tests/test_optimizer.py::test_distrib_optimizer_save_load_with_non_tensor_state -v` (regression)


asolergi-nv:
Hey @pandengyao, how are you triggering this error? Is it by changing the value of `store_param_remainders` across jobs?

pandengyao:
> Hey @pandengyao, how are you triggering this error? Is it by changing the value of `store_param_remainders` across jobs?

I encountered this issue while resuming the training of the Qwen3-VL model with use_precision_aware_optimizer=True enabled. The **root cause** of the issue is consistent with that described in the PR.

PR #4224, opened 2026-04-09
fix: pre-stage CUDA tensors in training process for persistent ckpt worker

ko3n1g:
## Summary

Fixes `RuntimeError: pidfd_getfd: Bad file descriptor` that crashes the NVRx persistent checkpoint worker on configs with virtual pipeline parallelism (VPP > 1), e.g. `gpt3_15b_8t_release_gb200`.

**Root cause**: when `--use-persistent-ckpt-worker` is set, `schedule_async_save` puts an `AsyncRequest` containing CUDA tensors (`write_buckets`) into a `multiprocessing.Queue` backed by a spawn/forkserver process. In that mode PyTorch transfers CUDA tensors via `pidfd_getfd`-based IPC handles. If the parent's file descriptors backing those handles are closed (e.g. by GC) before the worker's `queue.get()` reconstructs them, you get:

```
RuntimeError: pidfd_getfd: Bad file descriptor
```

**Why VPP > 1 specifically**: with `--num-layers-per-virtual-pipeline-stage N`, `write_buckets` grows proportionally with the number of virtual stages (8× for VPP=8). This means more CUDA IPC file descriptors to keep alive and a longer queue pipe transit time, making the race condition reliably reproducible.

**Fix**: call `preload_fn()` (GPU→CPU D2H) in the training process before the request enters the queue. The worker then receives only CPU tensors — no CUDA IPC is required. Training was already blocking on D2H completion (via the `persistent_preload_q.join()` rendezvous in the mcore path), so there is no change in effective timing.

```python
# Before: CUDA tensors sent into queue → pidfd_getfd EBADF race
queue.schedule_async_request(async_request)

# After: D2H done in training process, only CPU tensors reach the queue
if use_persistent_ckpt_worker and async_request.preload_fn is not None:
    preloaded_args[1] = async_request.preload_fn()   # GPU → CPU
    async_request = async_request._replace(async_fn_args=..., preload_fn=None)
queue.schedule_async_request(async_request)
```

## Test plan

- [ ] Verify `gpt3_15b_8t_release_gb200` no longer crashes with `pidfd_getfd: Bad file descriptor`
- [ ] Verify `gpt3_15b_8t_release_sm` (same flags, no VPP) continues to pass
- [ ] Confirm other configs using `--async-save --use-persistent-ckpt-worker` are unaffected

### DIFF
--- a/megatron/training/async_utils.py
+++ b/megatron/training/async_utils.py
@@ -77,10 +77,42 @@ def init_persistent_async_worker(rank: int, mp_mode: str = 'spawn'):
 def schedule_async_save(async_request: AsyncRequest | NVRxAsyncRequest):
     """Schedule the async save request.
 
+    When using a persistent checkpoint worker the AsyncRequest is serialized into a
+    multiprocessing queue.  Any CUDA tensors inside it are transferred between processes
+    via ``pidfd_getfd``-based IPC handles.  If the parent-process file descriptors backing
+    those handles are closed (e.g. by GC) before the worker calls ``queue.get()``, the
+    reconstruction fails with ``RuntimeError: pidfd_getfd: Bad file descriptor``.  This is
+    more likely on large models that use virtual pipeline parallelism (VPP > 1) because
+    ``write_buckets`` is proportionally larger, leading to more IPC handles and a longer
+    pipe transit time.
+
+    To avoid this, pre-stage CUDA tensors to CPU (call ``preload_fn``) in the training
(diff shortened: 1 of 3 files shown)


dimapihtar (approved):
LGTM. Thank you!

asolergi-nv (approved):

deepakn94 (approved):

ko3n1g:
close in favor of https://github.com/NVIDIA/Megatron-LM/pull/4178

PR #4229, opened 2026-04-09
feat(optimizer): add FlashAdamW optimizer integration

meinie0826:
# What does this PR do ?

Integrates [FlashAdamW](https://github.com/databricks/flashoptim) from the `flashoptim` library (>= 0.1.3) into Megatron-Core's optimizer infrastructure.

FlashAdamW reduces optimizer memory from **16 bytes/param** (BF16 model + fp32 master weight + fp32 exp_avg + fp32 exp_avg_sq) to **~7 bytes/param** via:
- **Master weight splitting**: BF16 param + INT8/INT16 ECC correction term (24-bit or 32-bit effective precision)
- **Companded state quantization**: softsign-transformed INT8 `exp_avg`, sqrt-transformed INT8 `exp_avg_sq`

Closes #4171

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


dimapihtar:
Hi @meinie0826 

thanks for the contribution. @gautham-kollu could you take a look, please?

guihong-nv:
Thanks for the contribution. I don't think this is mergeable yet; there are several blocking integration issues:

1. **The BF16/FP16 path defeats FlashAdamW's master-weight design.** `get_megatron_optimizer` still wraps the optimizer with `Float16OptimizerWithFloat16Params`, which replaces the low-precision parameters in `param_groups` with FP32 clones. This retains Megatron's external FP32 master weights and prevents FlashAdamW's BF16+ECC representation from providing the advertised memory savings. FlashAdamW needs a wrapper/path that keeps the original low-precision parameters, plus a test that constructs it through `get_megatron_optimizer` rather than directly.

2. **The declared `flashoptim>=0.1.3` range is incompatible with current 0.1.4.** The new `init_state_fn` calls `step_param` while gradients are absent. In FlashOptim 0.1.4, `step_param` returns before `_ensure_state_initialized` when `p.grad is None`. Megatron calls this initializer while constructing the expected optimizer state for distributed-checkpoint loading, before gradients exist, so the state remains empty and the optimizer moments cannot be restored correctly. The added `test_init_state_fn_populates_all_params` also fails with 0.1.4. Please either use a supported explicit initialization API or constrain the dependency until the integration supports newer releases.

3. **`--flashadamw-compress-state-dict` is incompatible with Megatron's default `torch_dist` checkpoint path.** Compressed FlashOptim state contains per-group scale tensors whose shapes differ from the corresponding parameters, while `optim_state_to_sharding_state` asserts that every optimizer-state tensor matches the parameter shape. Enabling this option therefore breaks default checkpoint creation. The current tests only call raw `torch.save(opt.state_dict())`, bypassing Megatron's sharded checkpoint code. This option needs custom sharding support or validation restricting it to compatible checkpoint formats.

4. **The default FP32 configuration crashes.** Selecting `--optimizer flashadamw` without `--bf16`/`--fp16` leaves `master_weight_bits=24`, but FlashOptim rejects non-`None` master-weight bits when all parameters are FP32. Validation should either require a supported low-precision mode or set the master-weight option appropriately for FP32. Fixing this must also account for the wrapper issue above.

5. **The checkpoint round-trip tests do not verify optimizer restoration.** They compare losses computed before either restored optimizer update. Because both models start from the same saved weights and input, the assertion passes even if optimizer state loading is completely broken. Please compare post-step parameters or a subsequent forward loss after both updates.

Finally, the PR currently rejects `--use-distributed-optimizer`, while #4171 explicitly requests distributed-optimizer and distributed-checkpoint compatibility. As submitted, this does not satisfy the scope needed to close that issue.

PR #4238, opened 2026-04-09
fix: enforce correct pass thresholds for deterministic and approximate tests

ko3n1g:
### Fix: test pass thresholds for deterministic and approximate tests

- **Deterministic tests** now require 100% of steps to match (previously the bar was ~2% minimum passing, making the test nearly meaningless)
- **Approximate tests** now require at most 1% of steps to fall outside the tolerance band (previously required at least 1% to pass — the inverse of the intended semantics)

### Fix: BERT MLM non-determinism

The HF fast tokenizer (`use_fast=True`) is backed by a Rust `HashMap` whose iteration order is non-deterministic across process restarts. `token_lookup` was built from `list(inv_vocab.keys())`, so the 10%-random-replacement branch of `_get_token_mask` produced different results each run even with the same seed.

Sorting the keys guarantees a stable `[0, 1, ..., vocab_size-1]` order regardless of the underlying hash-map's iteration order.

### Fix: NCCL collective algorithm

Added `NCCL_ALGO=Ring` to the env vars in `launch_nemo_run_workload.py`. The goal is to make the AWS GitHub Actions infrastructure behave more like the internal Slurm infrastructure, where Ring is already selected by the cluster tuner. Without it, NCCL auto-selects between Ring and Tree based on message size and topology. Tree has a different floating-point reduction order, which produces slightly different loss values and causes golden value mismatches on deterministic tests. `NCCL_ALGO=Ring` joins the existing set of vars that serve the same alignment purpose: `NCCL_PROTO=simple`, `NCCL_NVLS_ENABLE=0`, `NCCL_TUNER_PLUGIN`, `NCCL_BUFFSIZE`, and `NCCL_P2P_NET_CHUNKSIZE`.

### Golden value updates

Golden values regenerated for all affected test cases after the threshold, BERT, and NCCL fixes.

## Background

Two commits introduced the bugs this PR corrects:

**[`696f164`](https://github.com/ko3n1g/Megatron-LM/commit/696f164de1076f46c87c904e58ca293108459572) (Oct 2025) — "ci: Fix thresholds for iteration-time"**

Replaced `np.allclose` with a per-step check intended to mean "at most 1% of steps may deviate from golden values". The comparison was written backwards:

```python
# Intended: passes if ≥99% of steps match
# Actual:   passes if ≥1% of steps match (allows 99% to fail)
passing = np.mean(is_close) >= (num_failing_steps_allowed / total_steps_evaluated)
#                                ≈ 0.01
```

Any training run — even a fully diverged one — would pass this check as long as a single step happened to be within tolerance.

**[`46a90b0`](https://github.com/ko3n1g/Megatron-LM/commit/46a90b0b6c61b9c5b0ef310ebb567aa6fe1f5f91) (Oct 2025) — "ci(fix): Iteration testing"**

Simplified iteration-time evaluation to compare medians instead of per-step values, and moved `total_steps_evaluated` into a metric-specific branch. This preserved the inverted threshold without correcting it, and also routed deterministic tests through the same broken approximate path rather than requiring exact equality.

The result: for ~6 months, the CI was silently accepting results it should have rejected, and rejecting results it should have accepted.

## Root cause

**Threshold bug** — In `common.py::pipeline`, the variable `num_failing_steps_allowed` was used as a *minimum passing* threshold instead of a *maximum failing* threshold:

```python
# Before (wrong): passes if ≥1% of steps are within tolerance
passing = np.mean(is_close) >= (num_failing_steps_allowed / total_steps_evaluated)

# After (correct): passes if ≥99% of steps are within tolerance
passing = np.mean(is_close) >= 1 - (num_failing_steps_allowed / total_steps_evaluated)
```

Deterministic tests were also going through the same approximate path instead of requiring exact equality:

```python
# After: deterministic requires all steps to match
if test.type_of_test_result == TypeOfTestResult.DETERMINISTIC:
    passing = bool(np.all(is_close))
```

**BERT bug** — `token_lookup` iteration order depended on Rust's `HashMap`, which randomises bucket layout per process:

```python
# Before (non-deterministic): iteration order varies across runs
token_lookup = list(inv_vocab.keys())

# After (deterministic): stable ordering guaranteed
token_lookup = sorted(inv_vocab.keys())
```

## Test plan

- [ ] Verify a run with matching golden/actual values passes both deterministic and approximate checks
- [ ] Verify a run with diverging values (e.g. `deepseek_proxy_fsdp_ep2_fsdp2`) now correctly fails the deterministic check
- [ ] Verify BERT training loss is reproducible across runs with identical seeds


tests/functional_tests/python_test_utils/test_pretraining_regular_pipeline.py
```diff
@@ -10,13 +10,14 @@
...
 
+
 CHECK_THRESHOLDS = {
     "iteration-time": [common.ApproximateTest(atol=0, rtol=0.05)],
     "mem-allocated-bytes": [common.ApproximateTest(atol=0, rtol=0.05)],
     "mem-max-allocated-bytes": [common.ApproximateTest(atol=0, rtol=0.05)],
     "lm loss": [common.DeterministicTest(), common.ApproximateTest(atol=0, rtol=0.05)],
     "mtp_1 loss": [common.DeterministicTest(), common.ApproximateTest(atol=0, rtol=0.05)],
-    "num-zeros": [common.DeterministicTest(), common.ApproximateTest(atol=0, rtol=0.05)],
+    "num-zeros": [common.DeterministicTest(), common.ApproximateTest(atol=0, rtol=0.20)],
```
chtruong814, line 20: Why do we need to increase this?

tests/test_utils/recipes/h100/moe.yaml
```diff
@@ -168,16 +168,6 @@ products:
...
-  - test_case: [gpt3_moe_mcore_te_tp2_pp2_ep4_etp1_fine_grained_offloading]
-    products:
-      - environment: [dev]
-        scope: [mr]
-        platforms: [dgx_h100]
-  - test_case: [gpt3_moe_mcore_te_tp2_pp2_ep4_etp1_no_mtp_no_a2a_ovlp_fine_grained_offloading]
-    products:
-      - environment: [dev]
-        scope: [mr]
-        platforms: [dgx_h100]
```
chtruong814, line 180: Why are these being removed?

tests/test_utils/recipes/h100/unit-tests.yaml
```diff
@@ -185,3 +185,17 @@ products:
...
+        scope: [unit-tests]
+        n_repeat: [1]
+        time_limit: [1800]
+  - test_case: [tests/unit_tests/transformer/test_multi_token_prediction.py]
+    products:
+      - environment: [lts, dev]
+        tag: [latest, legacy]
+        scope: [unit-tests]
+        n_repeat: [1]
+        time_limit: [1800]
```
chtruong814, line 201: Why split these out?

bisect.sh
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env bash
```
chtruong814: This will be great

tests/test_utils/recipes/h100/moe-dynamic-inference-with-coordinator.yaml
```diff
@@ -68,5 +68,5 @@ products:
   - test_case: [gpt_dynamic_inference_tp4_etp1_pp1_ep8_16B_logitsmatch_zmq_suspend_resume]
     products:
       - environment: [dev]
-        scope: [mr, mr-github]
+        scope: [mr-broken, mr-github-broken]
```
chtruong814, line 71: Is this broken for different reasons?  Or we just can't make them deterministic?

chtruong814 (commented):
Can we have Claude add a unit test that's ran for the golden value comparisons?  I think it'll be good to have something to clarify the intent there.

Just had questions to understand better some of the other chagnes.

tests/functional_tests/python_test_utils/test_pretraining_regular_pipeline.py
ko3n1g: After the first bugs in October, we later tightened the thresholds to 0.05: https://github.com/ko3n1g/Megatron-LM/commit/adc69dba2dec6c9014d51b6cad3edf1020731786#diff-98a6573afbafb748a7739f6c370d567b40cc1260df4f7831b30a654262614898 - after fixing to check for all steps, these bounds are too tight. So setting it back to 0.20 is only restoring the thresholds prior to the original bug

tests/test_utils/recipes/h100/moe-dynamic-inference-with-coordinator.yaml
ko3n1g: these are the tests that suddenly have very large golden values files

tests/test_utils/recipes/h100/moe.yaml
ko3n1g: These are duplicates of https://github.com/ko3n1g/Megatron-LM/blob/3e1fd5e9cd00668e22df40d749d7100a7e836db3/tests/test_utils/recipes/h100/moe.yaml#L203 and https://github.com/ko3n1g/Megatron-LM/blob/3e1fd5e9cd00668e22df40d749d7100a7e836db3/tests/test_utils/recipes/h100/moe.yaml#L198 - they've been broken for long

tests/test_utils/recipes/h100/unit-tests.yaml
ko3n1g: the transformer unit tests have grown to >15min, so splitting this out for faster turnaround

chtruong814:
Sorry, were you able to add a unit test for the golden values comparison utilities?

jaredcasper (approved):

ko3n1g:
need to follow up on https://github.com/NVIDIA/Megatron-LM/actions/runs/24390893975/job/71247139916

PR #4260, opened 2026-04-11
Enable CUDA graphs for MTP inference

santhnm2:
# What does this PR do ?
Enables CUDA graphs for MTP inference.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


examples/inference/gpt/gpt_dynamic_inference.py
```diff
@@ -286,10 +286,6 @@ def main():
     )
     initialize_megatron()
 
-    # Start Nsight profiler.
-    if os.environ.get("NSIGHT_PREFIX"):
```
sidsingh-nvidia, line 290: How does this example work with MoEs, given that it does not have the coordinator?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -390,12 +406,39 @@ def create_cuda_graphs(self, reset_context: bool = True):
...
+                n = cuda_graph_batch_dimension.req_count
+                if sp_enabled:
+                    n += (tp_size - n % tp_size) % tp_size
+                if n > 0 and n not in mtp_seen_batch_sizes:
+                    mtp_seen_batch_sizes.add(n)
+                    device = torch.cuda.current_device()
+                    batch_dim = n // tp_size if sp_enabled else n
+                    # Use zeros (not empty) — garbage token IDs cause OOB embedding lookups during graph capture/replay.
+                    for depth in mtp_warmup_depths:
+                        with graph_capture():
```
sidsingh-nvidia, line 421: why do we use the graph capture decorator here, but not in the regular cudagraph capture?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -174,6 +182,27 @@ def _init_mtp_sampling_tensor(self):
                 )
                 * -1
             )
+            self._accepted_token_counts_per_request = torch.zeros(
```
sidsingh-nvidia: are all of these new buffers for cuda-graphing MTP? If so, a comment would be helpful

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -581,6 +610,24 @@ def _dynamic_step_context_init(
             is_expert_parallel_dummy_cuda_graph_step=is_dummy_forward,
         )
 
+        # Derive the MTP padded batch size from the EP-synced graph dimensions.
+        # In eager mode MTP uses locally SP-aligned batch size instead.
+        if getattr(self, '_has_mtp_cuda_graphs', False) and context.using_cuda_graph_this_step():
+            self._mtp_resolved_padded_count = context.padded_batch_dimensions.req_count
```
sidsingh-nvidia, line 619: This has to happen for cuda-graphed MTP w/o MoEs as well, right? The above comment makes it sound like this is MoE specific?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1428,6 +1449,7 @@ def __init__(
         self.pg_collection = pg_collection
         rng_tracker = get_cuda_rng_tracker()
         self.need_backward = need_backward
+        self.is_mtp = function_name == "compute_mtp_single_step"
```
lmcafee-nvidia: this looks fragile. can we use either an enum, or use a variable `func` assigned to the actual function, and do something like `self.mtp = func == self.compute_mtp_single_step`? I'd personally go with an enum, or maybe find a different way of setting `self.is_mtp`

megatron/core/transformer/cuda_graphs.py
```diff
@@ -506,6 +517,16 @@ def delete_cuda_graphs():
         runner.bwd_graph = None
         runner.mempool = None
 
+    # Reset MTP runners (excluded from the global inference record).
+    for mgr in _CudagraphGlobalRecord.mtp_cudagraph_managers:
```
mathemakitten, line 524: Since `_CudagraphGlobalRecord` is a persistent global, do you actually want to call `_CudagraphGlobalRecord.mtp_cudagraph_managers.clear()` at the end here? Otherwise you would accumulate entries into the `mtp_cudagraph_managers` list forever (which is usually not a real problem, unless you're doing e.g. test teardown-and-rebuild).

mathemakitten: I have some reservations about the asymmetry of adding `mtp_cudagraph_managers` to the `_CudagraphGlobalRecord` since it is a list of managers as opposed to a list of runners, but we can table a discussion about splitting up the runner replay ledge vs. a manager registry at some point. Maybe we can just leave a comment to look at a refactor later?

megatron/core/models/common/language_module/language_module.py
```diff
@@ -63,6 +63,25 @@ def __init__(
         self.vp_stage = None
         self.vp_size = self.config.virtual_pipeline_model_parallel_size
 
+    def _setup_mtp_cuda_graphs(self):
+        """Wrap ``compute_mtp_single_step`` with a CudaGraphManager.
+
+        Must be called by subclasses after ``self.mtp`` is created.
```
mathemakitten: style nit: can we get rid of the claude `` double-brackets here :')

megatron/core/models/gpt/gpt_model.py
```diff
@@ -241,7 +243,15 @@ def __init__(
                 self.embedding_activation_buffer = None
                 self.grad_output_buffer = None
 
-            self.output_layer = tensor_parallel.ColumnParallelLinear(
+            if config.transformer_impl == "inference_optimized":
+                from megatron.core.tensor_parallel.inference_layers import (
```
sidsingh-nvidia: Can we move this import up?

megatron/core/models/common/language_module/language_module.py
```diff
@@ -63,6 +63,25 @@ def __init__(
...
+        """
+        if self.config.cuda_graph_impl == "local":
+            from megatron.core.transformer.cuda_graphs import (
```
sidsingh-nvidia: move this import up

mathemakitten (approved):

examples/inference/gpt/gpt_dynamic_inference.py
santhnm2: Good point, `gpt_dynamic_inference_with_coordinator.py` should be preferred for any MoE workload. But the base script still works, just without data parallelism (I pass in a prompt file so all ranks run all prompts). We should probably consolidate these two scripts in the future though.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: Yea I think the comment was trying to reflect that there's no additional sync that has to happen for MTP if we've already synced for the forward pass; I agree that it's confusing - will fix.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: Will add a comment

megatron/core/inference/engines/dynamic_engine.py
santhnm2: The `create_cudagraphs` function in `cuda_graphs.py` manually calls `_set_capture_start()` and `_set_capture_end()`, we could theoretically use the wrapper there as well but I mostly just added this for syntactic sugar in the MTP code. Open to removing the wrapper if it's too messy

megatron/core/tensor_parallel/inference_layers.py
```diff
@@ -312,20 +326,51 @@ def _all_gather(self, x: torch.Tensor, symm_mem_buffer: dict) -> None:
             x, _ = gather_along_first_dim(x, process_group=self.tp_group)
             return x
 
-    def forward(self, x: torch.Tensor) -> Tuple[torch.Tensor, None]:
-        """
-        Forward pass.
-        """
-        if self.training:
-            return super().forward(x)
+    def _nvls_gather_last_dim(self, x: torch.Tensor) -> torch.Tensor:
```
lmcafee-nvidia: what's the difference between this method and `inference_all_gather_last_dim` below? they seem to do the same thing. if so, we should get rid of one.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -390,12 +406,39 @@ def create_cuda_graphs(self, reset_context: bool = True):
...
+                            )
+
             context.reset()
 
         # Disable inference dispatcher after graph capture
         if is_inference_optimized_ep:
             unset_inference_cuda_graphed_iteration_for_ep_inference(unwrapped_model)
 
+        if mtp_warmup_enabled and mtp_seen_batch_sizes:
+            controller._has_mtp_cuda_graphs = True
```
lmcafee-nvidia: nit: `_has_mtp_cuda_graphs` is private but we are setting it directly. might consider making this attribute non-private, or access it through a helper

megatron/core/models/common/language_module/language_module.py
```diff
@@ -325,24 +344,27 @@ def shared_embedding_or_output_weight(self) -> Tensor:
...
-        self, hidden_states: Tensor, next_token_ids: Tensor, position_ids: Tensor, depth: int
+        self,
+        hidden_states: Tensor,
+        next_token_ids: Tensor,
+        position_ids: Tensor,
+        depth: Optional[int] = None,
     ) -> tuple:
         """Compute a single MTP depth for speculative decoding.
 
-        This is called after speculative token verification to compute MTP
```
lmcafee-nvidia, line 332: why are we removing this docstring?

megatron/core/models/hybrid/hybrid_model.py
```diff
@@ -258,10 +258,19 @@ def __init__(
                 mtp_num_depths=self.mtp_num_depths,
                 hybrid_submodules=hybrid_submodules,
             )
+            self._setup_mtp_cuda_graphs()
 
         # Output
         if post_process or self.mtp_process:
-            self.output_layer = tensor_parallel.ColumnParallelLinear(
+            if config.transformer_impl == "inference_optimized":
```
lmcafee-nvidia: this same code block is used in both `gpt_model.py` and here. I suggest abstracting this into a reusable method

examples/inference/gpt/gpt_dynamic_inference.py
```diff
@@ -505,19 +526,17 @@ def escape_str(s):
         #     f"count [ p {p_count}, d {d_count} ]."
         # )
         capture_str = f"{engine.capture_stats['time']:.2f} sec" if engine.capture_stats else "--"
+        cuda_throughput = total_output_tokens / (cuda_elapsed_ms / 1000.0)
         print(
             f"{setup_prefix} … " f"throughput: {throughput:.3f} tok/s … ",
             f"total time: {total_time:.3f}s … "
+            f"cuda time: {cuda_elapsed_ms:.1f}ms ({cuda_throughput:.3f} tok/s) … "
```
lmcafee-nvidia: should we just use these CUDA numbers, and get rid of the CPU throughput and time?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -581,6 +610,24 @@ def _dynamic_step_context_init(
...
+            if self._sp_enabled:
+                self._mtp_resolved_padded_count += (
```
sidsingh-nvidia: if this another way of saying mtp_resolved_padded_count = ceil_div(mtp_resolved_padded_count / tp_size) * tp_size? If so, then is the way I am framing it much easier to parse?

examples/inference/gpt/gpt_dynamic_inference.py
sidsingh-nvidia: At some point we should get rid of this example. Perhaps move it inside the tests folder, since some tests still rely on it.

megatron/core/transformer/cuda_graphs.py
sidsingh-nvidia: Is there a fundamental reason why we need a separate list for mtp cuda graph managers?

mathemakitten: There is logic which uses `_CudagraphGlobalRecord.cudagraph_inference_record` to wire the output buffer of one layer's graph to be the input buffer of the next layer's graph, and adding MTP runners to this would pollute the traversal. 

That being said, I think the only reason it needs to exist at all is to give `delete_cuda_graphs` a way to delete the graphs without dropping all strong refs to the model, which is scoped to basically just testing at this point.

megatron/core/inference/engines/dynamic_engine.py
santhnm2: Made this a public attribute

megatron/core/tensor_parallel/inference_layers.py
santhnm2: Got rid of `_nvls_gather_last_dim`

megatron/core/transformer/cuda_graphs.py
santhnm2: Now I'm just directly adding an `is_mtp_inference` parameter

megatron/core/models/common/language_module/language_module.py
santhnm2: Done

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: Updated the comment

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: Added comment

examples/inference/gpt/gpt_dynamic_inference.py
santhnm2: Reverted changes to `gpt_dynamic_inference.py`

megatron/core/inference/text_generation_controllers/text_generation_controller.py
santhnm2: Added a `round_to_nearest_multiple` utility function for clarity

megatron/core/models/common/language_module/language_module.py
santhnm2: Done

megatron/core/models/gpt/gpt_model.py
santhnm2: Done

examples/inference/gpt/gpt_dynamic_inference.py
santhnm2: Reverted changes to `gpt_dynamic_inference.py`

megatron/core/models/hybrid/hybrid_model.py
santhnm2: I'm now keeping `tensor_parallel.ColumnParallelLinear` but just substituting the all gather op internally for the NVLS version

megatron/core/models/common/language_module/language_module.py
santhnm2: Restored

megatron/core/inference/text_generation_controllers/text_generation_controller.py
sidsingh-nvidia: Remind me to update this comment after the nvls dispatcher goes in.

sidsingh-nvidia (approved):

lmcafee-nvidia (approved):

megatron/core/tensor_parallel/inference_layers.py
```diff
@@ -473,3 +477,73 @@ def forward(
         else:
             x = self._matmul_reduce_scatter(x)
             return x, None
+
+
+def inference_all_gather_last_dim(
```
santhnm2: Change to `inference_all_gather_from_tensor_model_parallel_region`

megatron/core/tensor_parallel/inference_layers.py
```diff
@@ -473,3 +477,73 @@ def forward(
...
+
+        if are_tensors_nvls_eligible(x) and symm_mem_buffer["handle"] is not None:
+            multimem_all_gather(symm_mem_buffer["tensor"], x, symm_mem_buffer["handle"])
+            tensor_list = symm_mem_buffer["tensor"].chunk(tp_size, dim=0)
+            return torch.cat(tensor_list, dim=-1).contiguous()
+
+    return gather_from_tensor_model_parallel_region(x, group=tp_group)
+
+
+def inference_reduce_scatter_first_dim(
```
santhnm2: `inference_reduce_scatter_to_sequence_parallel_region`

megatron/core/inference/text_generation_controllers/mtp_utils_triton.py
santhnm2, line 1: Make this `mtp_utils_triton.py`, then restore pytorch functions to `mtp_utils_pytorch.py` and ensure 1:1 API match between the two implementations

kvareddy (approved):

mathemakitten (approved):

jaredcasper (approved):

shanmugamr1992 (approved):

PR #4292, opened 2026-04-14
Refactor CUDA graph API: decompose cuda_graph_scope into full_iteration impl, inference scope, and per-layer capture modules

buptzyb:
# What does this PR do ?
Dev PR #4293 

This PR decomposes the overloaded `cuda_graph_scope` field into three dedicated, semantically distinct concepts and cleans up naming throughout.

### Problem

The old API overloaded `--cuda-graph-scope` with three unrelated concerns:
- `--cuda-graph-scope full_iteration` → full-iteration training graphs (a capture *strategy*, not a module)
- `--cuda-graph-scope full_iteration_inference` → block-owned inference graphs (inference *ownership*, not a module)
- `--cuda-graph-scope attn mlp` → per-layer capture regions (the actual intended use)

`CudaGraphScope` mixed iteration-level control flow with per-layer module selection. The three concepts have nothing in common and cannot be meaningfully combined in one field.

### Solution

**Four concrete changes:**

1. **`full_iteration` becomes its own `cuda_graph_impl` value.**
   `--cuda-graph-impl full_iteration` replaces `--cuda-graph-impl local --cuda-graph-scope full_iteration`.

2. **`--inference-cuda-graph-scope` is a new dedicated field.**
   `InferenceCudaGraphScope` (`none` / `layer` / `block`) replaces `full_iteration_inference` in `cuda_graph_scope`. The default for `--cuda-graph-impl local` is `layer` (preserving prior behaviour).

3. **`CudaGraphScope` is renamed to `CudaGraphModule`; `cuda_graph_scope` to `cuda_graph_modules`.**
   With the two non-module values removed, the enum and field names now accurately reflect their purpose: selecting which per-layer modules to capture.

4. **Normalization is centralized in `cuda_graph_config.py`.**
   `normalize_cuda_graph_modules`, `normalize_inference_cuda_graph_scope`, and `validate_deprecated_cuda_graph_modules_migration_inputs` are shared between `TransformerConfig.__post_init__` and `validate_args`, removing duplication and making the migration logic testable in isolation.

### Backward compatibility

All deprecated inputs are still accepted and silently migrated at startup:

| Old input | New equivalent |
|---|---|
| `--enable-cuda-graph` | `--cuda-graph-impl local` |
| `--external-cuda-graph` | `--cuda-graph-impl transformer_engine` |
| `--cuda-graph-scope full_iteration` | `--cuda-graph-impl full_iteration` |
| `--cuda-graph-scope full_iteration_inference` | `--cuda-graph-impl local --inference-cuda-graph-scope block` |
| `--cuda-graph-scope attn mlp` | `--cuda-graph-modules attn mlp` |
| `from megatron.core.transformer.enums import CudaGraphScope` | `CudaGraphScope` alias kept; use `CudaGraphModule` |
| `TransformerConfig(cuda_graph_scope=...)` | `cuda_graph_scope` field kept, deprecated; use `cuda_graph_modules` |

Conflicting combinations (e.g. passing both a deprecated value and the new flag with an incompatible value) are rejected with a clear assertion.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


mathemakitten:
Can we please make sure that the existing `transition_cudagraph_scope` function which is used to transition between full-layer and partial cudagraphs reflects these changes?

buptzyb:
> Can we please make sure that the existing `transition_cudagraph_scope` function which is used to transition between full-layer and partial cudagraphs reflects these changes?

@mathemakitten  I looked at `transition_cudagraph_scope`, and if I understand it correctly, there is nothing need to change inside the function. The existing `transition_cudagraph_scope` helper is not using the old `cuda_graph_scope` API to select graph modules; it only switches the MoE layer runtime state between `partial` and `full` modes. The actual API transition to `cuda_graph_modules` is already handled in `create_mcore_cudagraph_manager`. Right?

mathemakitten:
> > Can we please make sure that the existing `transition_cudagraph_scope` function which is used to transition between full-layer and partial cudagraphs reflects these changes?
> 
> @mathemakitten I looked at `transition_cudagraph_scope`, and if I understand it correctly, there is nothing need to change inside the function. The existing `transition_cudagraph_scope` helper is not using the old `cuda_graph_scope` API to select graph modules; it only switches the MoE layer runtime state between `partial` and `full` modes. The actual API transition to `cuda_graph_modules` is already handled in `create_mcore_cudagraph_manager`. Right?

Ah you're right! Thanks for checking.

docs/user-guide/features/cuda_graph.md
```diff
@@ -0,0 +1,207 @@
...
+| Implementation | Flag | Granularity | Use case |
+|---|---|---|---|
+| Per-layer (local) | `--cuda-graph-impl local` | Per-layer | Automatic MCore `CudaGraphManager` backend |
+| Per-layer (TE) | `--cuda-graph-impl transformer_engine` | Per-layer | Per-layer TE-backed implementation using `make_graphed_callables()` |
+| Full-iteration | `--cuda-graph-impl full_iteration` | Entire training iteration (excluding optimizer) | Training / validation; maximum overhead reduction |
+
+Think of the configuration in two layers:
+
+- `--cuda-graph-impl` selects the implementation family.
+- `--cuda-graph-modules` refines coverage within the per-layer families (`local` and `transformer_engine`).
```
mathemakitten:
```suggestion
- `--cuda-graph-modules` defines exactly which modules should be graphed within the per-layer families (`local` and `transformer_engine`).
```

docs/user-guide/features/cuda_graph.md
```diff
@@ -0,0 +1,207 @@
...
+not in the user-facing per-layer capture model. The only mode with a fundamentally different
+granularity is `full_iteration`.
+
+---
+
+## Per-Layer CUDA Graph — Local Implementation (`--cuda-graph-impl local`)
+
+Uses MCore's built-in `CudaGraphManager`. Like `transformer_engine`, this is a per-layer mode:
+leaving `--cuda-graph-modules` unset captures the whole Transformer layer, while specifying
+modules restricts capture to selected sub-regions.
```
mathemakitten: This is true for training but not inference. For inference, local + `inference-cuda-graph-scope=block` attaches the CudaGraphManager at the block level and skips per-layer managers entirely. We should change both the title of this section and this description to reflect.

docs/user-guide/features/cuda_graph.md
```diff
@@ -0,0 +1,207 @@
...
+
+For implementation background and design details, see NVIDIA's
+[Transformer Engine and Megatron-LM CUDA Graph Support](https://docs.nvidia.com/dl-cuda-graph/torch-cuda-graph/te-megatron-cuda-graphs.html).
+That article is a useful conceptual reference, but some examples there still use older flags such as
+`--enable-cuda-graph` or `--cuda-graph-modules full_iteration`; in this repository, prefer
+`--cuda-graph-impl local|transformer_engine|full_iteration` as documented below.
+
+## Overview
+
+| Implementation | Flag | Granularity | Use case |
```
mathemakitten: Can we split this overview table into training and inference? This is currently correct for training but in inference we support `--inference-cuda-graph-scope=block` as well which is not per-layer as stated here.

docs/user-guide/features/cuda_graph.md
```diff
@@ -0,0 +1,207 @@
...
+Uses MCore's built-in `CudaGraphManager`. Like `transformer_engine`, this is a per-layer mode:
+leaving `--cuda-graph-modules` unset captures the whole Transformer layer, while specifying
+modules restricts capture to selected sub-regions.
+
+Operationally, this path is tightly integrated into MCore itself:
+
+- graphable modules create and own their `CudaGraphManager` instances automatically
+- the stock training schedules drive warmup/capture/replay automatically
+- users typically select the mode through config flags only; there is no separate helper API to
+  wire into a custom training loop
```
mathemakitten:
```suggestion
Operationally, this path is tightly integrated into MCore training and inference:

- graphable modules create and own their `CudaGraphManager` instances automatically
- the existing training schedules drive warmup/capture/replay automatically
- users select the mode through config flags only; there is no separate helper API to
  wire into a custom training loop or a separate need to handle static input buffers
```

docs/user-guide/features/cuda_graph.md
```diff
@@ -0,0 +1,207 @@
...
+-->
+
+# CUDA Graph
+
+CUDA Graphs reduce kernel-launch overhead by recording GPU operations once and replaying the recording on subsequent iterations. Megatron-LM provides three CUDA graph implementations controlled by `--cuda-graph-impl`.
+
+For implementation background and design details, see NVIDIA's
+[Transformer Engine and Megatron-LM CUDA Graph Support](https://docs.nvidia.com/dl-cuda-graph/torch-cuda-graph/te-megatron-cuda-graphs.html).
+That article is a useful conceptual reference, but some examples there still use older flags such as
+`--enable-cuda-graph` or `--cuda-graph-modules full_iteration`; in this repository, prefer
```
mathemakitten: nit, if we're talking about backward compatibility, the old flag was `--cuda-graph-scope`
```suggestion
That article is a useful conceptual reference, but some examples there still use older flags such as
`--enable-cuda-graph` or `--cuda-graph-scope full_iteration`; in this repository, prefer
```

docs/user-guide/features/cuda_graph.md
mathemakitten: e.g. here's an example generated version which is clearer / more correct imo, and also puts all the options for the flags upfront: 

```
  ## Overview

  CUDA graph behavior is set by three orthogonal flags:

  | Flag | Values | Purpose |
  |---|---|---|
  | `--cuda-graph-impl` | `none` / `local` / `transformer_engine` / `full_iteration` | Which capture backend / strategy |
  | `--cuda-graph-modules` | `attn` / `mlp` / `moe` / `moe_router` / `moe_preprocess` / `mamba` (multi-valued) | Per-layer capture coverage; only meaningful for `local` and `transformer_engine` during training |
  | `--inference-cuda-graph-scope` | `none` / `layer` / `block` | Granularity of CUDA graphs *during inference*; only `local` supports non-`none` values |

  Supported combinations:

  | `--cuda-graph-impl` | Backend | Training capture | Inference capture |
  |---|---|---|---|
  | `none` | — | off | off |
  | `local` | MCore `CudaGraphManager` | per-layer (controlled by `--cuda-graph-modules`) | `layer` (default) or `block` |
  | `transformer_engine` | TE `make_graphed_callables()` | per-layer (controlled by `--cuda-graph-modules`) | not supported (`none` only) |
  | `full_iteration` | MCore `FullCudaGraphWrapper` | one graph per training iteration; `--cuda-graph-modules` must be empty | not supported (`none` only) |
```

docs/user-guide/features/cuda_graph.md
```diff
@@ -0,0 +1,207 @@
...
+Transformer layer, while specifying modules restricts capture to selected sub-regions. The main difference from
+`local` is the backend implementation and feature compatibility, not the capture granularity.
+
+Compared to `local`, this path exposes a more general and self-contained API via TE's
+`make_graphed_callables()`, giving users greater flexibility and control over how CUDA graphs are
+wired into custom training loops. The trade-off is that it requires more manual setup:
+
+- the training loop must instantiate `TECudaGraphHelper`
+- the training loop must call helper methods such as `create_cudagraphs()` and
+  `cuda_graph_set_manual_hooks()` at the correct points
```
mathemakitten, line 97: I'd also note that this path does not do the static input buffer checking that the local path does to ensure that inputs get fixed memory addresses out of the box.

docs/user-guide/features/cuda_graph.md
```diff
@@ -0,0 +1,207 @@
...
+--cuda-graph-impl transformer_engine \
+--cuda-graph-modules attn moe_router moe_preprocess
+```
+
+The same `--cuda-graph-modules` options apply as for `local`, and the default is likewise
+whole-layer capture when the flag is omitted.
+
+---
+
+## Full-Iteration CUDA Graph (`--cuda-graph-impl full_iteration`)
```
mathemakitten: Specify that this is training, and/or bring the equivalent inference flow into this section

megatron/core/transformer/enums.py
```diff
@@ -58,14 +58,57 @@ class AttnBackend(enum.Enum):
...
+
+class InferenceCudaGraphScope(enum.Enum):
+    """Inference CUDA graph scope.
+
+    This controls the ownership boundary for inference CUDA graphs:
+    - none: inference runs without CUDA graphs.
+    - layer: graphs are owned at the module/layer boundary. This name
+      does not by itself imply any finer-grained replay contract within the layer.
+    - block: graphs are owned by the enclosing block rather than
+      individual modules.
```
mathemakitten:
```suggestion
    This controls the ownership boundary for inference CUDA graphs:
    - none:  inference runs in eager mode (no CUDA graphs).
    - layer: graphs are owned at the module/layer boundary, e.g. TransformerLayer or MambaLayer.
    - block: graphs are owned by the enclosing block, e.g. TransformerBlock or HybridBlock.
```

megatron/core/transformer/transformer_config.py
```diff
@@ -867,22 +879,64 @@ class TransformerConfig(ModelParallelConfig):
...
-
-    cuda_graph_scope: Union[str, CudaGraphScope, List[str], List[CudaGraphScope]] = "full"
-    """Determines the CUDA graphs capturing scope.
-    When cuda_graph_impl is set to "transformer_engine", valid values are "attn", "mlp", "moe",
-    "moe_router", "moe_preprocess", "mamba". "full" or an empty list means the full layer. "full"
-    is actually deprecated, but for backward compatibility, we still use "full" as the default
-    value. It will be transformed to an empty list in __post_init__.
-    When cuda_graph_impl is set to "local", "full_iteration" can be specified as cuda_graph_scope
-    to enable whole iteration CUDA graph. All other values enable layerwise CUDA graph."""
+    "local": per-layer CUDA graph using MCore's implementation.
```
mathemakitten: same comment here re: per-layer and inference's "block" scope

megatron/core/transformer/transformer_config.py
```diff
@@ -867,22 +879,64 @@ class TransformerConfig(ModelParallelConfig):
...
+        },
+    )
+    """Controls the CUDA graph scope during inference.
+    When unset, the effective default is derived from cuda_graph_impl:
+    "local" -> "layer", all other impls -> "none".
+    "none": inference runs without CUDA graphs.
+    "layer": inference graphs are owned at the layer boundary. This name does not
+    by itself imply any finer-grained replay contract within the layer.
+    "block": inference graphs are owned by the enclosing block rather than
+    individual layers.
```
mathemakitten: same comment as above

megatron/inference/utils.py
```diff
@@ -347,7 +347,7 @@ def get_inference_config_from_model_and_args(model: MegatronModule, args):
         mamba_memory_ratio=args.inference_dynamic_batching_mamba_memory_ratio,
         num_cuda_graphs=(
             args.inference_dynamic_batching_num_cuda_graphs
-            if args.cuda_graph_impl == "local"
+            if args.cuda_graph_impl in ("local", "full_iteration")
```
mathemakitten: I don't think we need to keep the full_iteration impl + dynamic inference path here? We only want to setup inference graphs if `inference_cuda_graph_scope` is used, right?

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -335,16 +334,16 @@ def create_cuda_graphs(self, reset_context: bool = True):
...
-        if (
-            CudaGraphScope.full_iteration in self.cuda_graph_scope
-            and CudaGraphScope.full_iteration_inference not in self.cuda_graph_scope
-        ):
+        if self.cuda_graph_impl == "full_iteration":
             warnings.warn(
-                "\n\n*** WARNING: 'full_iteration' CUDA graph scope used during inference! "
-                "This will not create inference CUDA graphs. Use '--cuda-graph-scope=full_iteration_inference' instead. ***\n"
+                "\n\n*** WARNING: '--cuda-graph-impl=full_iteration' used during inference! "
+                "For compatibility, this preserves the legacy '--cuda-graph-modules=full_iteration' "
```
mathemakitten:
```suggestion
                "For compatibility, this preserves the legacy '--cuda-graph-scope=full_iteration' "
```

mathemakitten (commented):
Thank you for the thorough refactor! I've tried my best to catch the outstanding inconsistencies in the docs, mostly related to inference.

docs/user-guide/features/cuda_graph.md
buptzyb: I think `TECudaGraphHelper` makes this checking too?

megatron/inference/utils.py
buptzyb: Okay, in this way the old and new behavior won't align in the full_iter path. But I'm good if you think this is safe.

I modified it to `if args.inference_cuda_graph_scope != InferenceCudaGraphScope.none` here. Also in `DynamicInferenceEngine.create_cuda_graphs()`.

megatron/core/inference/engines/dynamic_engine.py
buptzyb: Following the above idea, I directly judge `inference_cuda_graph_scope != InferenceCudaGraphScope.none` and `cuda_graph_impl == "local"` here, so this warning message is no longer needed.

mathemakitten (approved):

megatron/core/transformer/cuda_graph_config.py
```diff
@@ -0,0 +1,147 @@
...
+# Shared by transformer_config.__post_init__ and validate_args to avoid duplication.
+ALLOWED_INFERENCE_SCOPES: dict[str, Set[InferenceCudaGraphScope]] = {
+    "none": {InferenceCudaGraphScope.none},
+    "local": {InferenceCudaGraphScope.layer, InferenceCudaGraphScope.block},
+    "transformer_engine": {InferenceCudaGraphScope.none},
+    "full_iteration": {InferenceCudaGraphScope.none},
+}
+
+
+class NormalizedCudaGraphModules(NamedTuple):
```
jiemingz: Why do we need a class here? This is only ever returned upon `normalize_cuda_graph_modules` right? Do we actually take advantage of the `NamedTuple` functionality?

megatron/core/transformer/cuda_graph_config.py
```diff
@@ -0,0 +1,147 @@
...
+            f"cuda_graph_impl={cuda_graph_impl!r}."
+        )
+        assert inference_cuda_graph_scope in (None, InferenceCudaGraphScope.block), (
+            "cuda_graph_modules='full_iteration_inference' cannot be combined with "
+            "inference_cuda_graph_scope="
+            f"{getattr(inference_cuda_graph_scope, 'name', inference_cuda_graph_scope)!r}."
+        )
+
+
+def get_deprecated_cuda_graph_modules_migration(
```
jiemingz, line 129: Since this function is only called once and is so short, does it really need a whole function?

megatron/core/transformer/cuda_graph_config.py
```diff
@@ -0,0 +1,147 @@
...
+            if scope in CUDA_GRAPH_MODULES_DEPRECATIONS:
+                attr, value = CUDA_GRAPH_MODULES_DEPRECATIONS[scope]
+                deprecated_scopes.append((scope, attr, value))
+            else:
+                normalized_scopes.append(CudaGraphModule[scope])
+
+    return NormalizedCudaGraphModules(normalized_scopes, deprecated_scopes, False)
+
+
+def normalize_inference_cuda_graph_scope(
```
jiemingz, line 62: Same comment here, why does this need to be put into a function? Does carving this out into a function really improve readability?

megatron/core/transformer/cuda_graph_config.py
buptzyb: Make sense. Looking into the calling site, we don't actually need a NamedTuple. Returning a tuple with three elements is good enough. Fixed.

megatron/core/transformer/cuda_graph_config.py
buptzyb: No, it's called in two places: arguments.py `validate_args()`, and transformer_config.py `TransformerConfig.__post_init__()` in case the user directly uses `core/`.

The reason I normalize in arguments.py as well is that the CLI args path does not immediately hand everything off to TransformerConfig. `parse_and_validate_args()` performs additional validation directly on args before `core_transformer_config_from_args()` builds the TransformerConfig, and some later training/inference setup also continues to use args as the source of truth. For example, arguments.py checks inference_cuda_graph_scope against InferenceCudaGraphScope.block and checks membership of CudaGraphModule values in args.cuda_graph_modules.

  If we only normalized inside `TransformerConfig.__post_init__`, those args-level checks would still have to handle raw CLI strings and deprecated spellings separately, or they could behave differently from the dataclass config path. Keeping the shared normalization helpers lets both entry points canonicalize to the same enum/list representation and use the same deprecated-value migration rules.

megatron/core/transformer/cuda_graph_config.py
buptzyb: Same as above. This method is used in two places.

megatron/core/transformer/cuda_graph_config.py
jiemingz: Make sense, I missed the `validate_args`

jiemingz (approved):

yashaswikarnati:
this PR is too big to do a careful review :) . LGTM for schedules and moe changes.

yashaswikarnati (approved):

sidsingh-nvidia (approved):

Victarry (approved):

santhnm2 (approved):

tdene (approved):

yaoyu-33 (approved):

kvareddy (approved):

shjwudp (approved):

PR #4301, opened 2026-04-14
Inference | Per-block MoE routing storage for prefix caching

lmcafee-nvidia:
## Summary
- Convert MoE routing indices from GPU tensors to CPU numpy arrays after the forward pass, reducing GPU memory pressure at long sequences
- Add per-block routing storage on `KVBlockAllocator` so routing indices are scattered into KV cache blocks and reconstructed at request completion — prefix-cached blocks retain routing from the original request
- Add chunk-based routing accumulation (`add_routing_indices` / `finalize_routing_chunks`) and numpy serialization support on `DynamicInferenceRequest`

## Test plan
- [x] 7 new `TestPerBlockRouting` tests (store/get round-trip, cleared on allocate/reset, persists through deregister, reconstruct from blocks, missing block returns None, survives LRU prefix match)
- [x] All 25 prefix caching tests pass with no regressions
- [ ] CI


shanmugamr1992 (approved):

megatron/core/inference/contexts/kv_block_allocator.py
```diff
@@ -358,3 +369,139 @@ def evict_lru_blocks(self, num_blocks_needed: int) -> bool:
...
+        active_request_slice = slice(context.paused_request_count, context.total_request_count)
+        active_request_ids = context.request_ids[active_request_slice].tolist()
+        routing_parts = [
+            routing_indices_per_request[rid]
+            for rid in active_request_ids
+            if rid in routing_indices_per_request
+        ]
+        if not routing_parts:
+            return
+        flat_routing = np.concatenate(routing_parts, axis=0)  # [token_count, num_layers, topk]
```
sidsingh-nvidia, line 414: It seems like this is the reverse of what we are doing in the text_generation_controller's router_record_bookkeeping function. Over there we get the map as [token_count, num_layers, topk] and convert it into a per request map. Here we are doing the opposite conversion. Do you think we could simply return the per token map from router_record_bookeeping and simplify things?

sidsingh-nvidia:
We should not store the routing indices in both the block store as well inside each request, given how much memory they can use. I suggest we only use the block store to store them, and once a request is finished, we reconstruct the entire data from it's blocks and send it back to the coordinator.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1293,21 +1320,16 @@ def post_process_requests(
...
                 step_routing = routing_indices_per_request[
                     request_id
                 ]  # [num_tokens, num_layers, topk]
-                if request.routing_indices is None:
-                    request.routing_indices = step_routing.clone()
-                else:
-                    request.routing_indices = torch.cat(
-                        [request.routing_indices, step_routing], dim=0
-                    )
+                request.add_routing_indices(step_routing)
```
sidsingh-nvidia, line 1332: Can't we get rid of this storage now? Let the block store be the only place where we store the routing indices.

sidsingh-nvidia (changes requested):

shanmugamr1992:
@lmcafee-nvidia  this wont work with sequence parallel
You need to add the fix I suggested

megatron/core/inference/engines/dynamic_engine.py
lmcafee-nvidia: great point, I got rid of `routing_indices_chunks`. Now, the only thing left on the request object is `routing_indices` that we use for unrolling the block routing indices right before serializing a finished request

megatron/core/inference/contexts/kv_block_allocator.py
lmcafee-nvidia: good point, I got right of the per-request dict and now I just pass the flat token-dimension indices

sidsingh-nvidia (approved):
lgtm

kvareddy (approved):

PR #4309, opened 2026-04-15
[training migration] Add serialization features to config container

maanug-nv:
# What does this PR do ?
Enables serializing+deserializing config containers to/from yaml or dictionary.

Also includes an allowlist for instantiating from yaml/dict for security purposes.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/training/argument_utils.py
```diff
@@ -248,3 +261,58 @@ def _get_field_docstrings(self, src_cfg_class: type) -> dict[str, str]:
...
+    for f in fields(cls):
+        if hasattr(args, f.name):
+            kwargs[f.name] = getattr(args, f.name)
+
+    if return_instance:
+        return cls(**kwargs)
+    else:
+        return kwargs
+
+def pretrain_container_from_args(args: Namespace) -> PretrainConfigContainer:
```
yobibyte: I would rename this to pretrain_cfg_container_from_args. Alternatively, we can make this a class method of PretrainConfigContainer and just name it `from_args`

maanug-nv: I'd like to move all arguments related stuff outside of `megatron/training` eventually, including everything in `argument_utils.py` . The training loop is packaged with MCore right now, and ideally that code should be interface agnostic (i.e. offers dataclasses and users can decide how to fill those out).

I'll rename it to `pretrain_cfg_container_from_args`.

megatron/training/config/container.py
```diff
@@ -1,13 +1,218 @@
...
+        Create a config container from a YAML file.
+
+        Args:
+            yaml_path: Path to the YAML file
+            mode: Serialization mode (strict or lenient)
+
+        Returns:
+            A new instance of this class initialized with the YAML file values
+        """
+        if MultiStorageClientFeature.is_enabled():
```
jaredcasper, line 83: This pattern keeps showing up, here and other places... any way to clean this up? I thought MSC was supposed to be mostly transparent in the code?

megatron/training/config/instantiate_utils.py
```diff
@@ -0,0 +1,552 @@
+# Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
```
jaredcasper: Wrong year.

megatron/training/config/instantiate_utils.py
```diff
@@ -0,0 +1,552 @@
...
+    """
+
+    def __init__(self) -> None:
+        self._allowed_prefixes: list[str] = list(_DEFAULT_ALLOWED_PREFIXES)
+        self._allowed_exact: set[str] = set(_DEFAULT_ALLOWED_EXACT)
+        self._enabled: bool = True
+        self._check_env_override()
+
+    def _check_env_override(self) -> None:
+        if os.environ.get("MEGATRON_ALLOW_ALL_TARGETS", "").strip() == "1":
```
jaredcasper: Is there any other way than an environment variable?? Why do we even need this option?

maanug-nv: i'm fine to drop this method of control completely. i guess the env var allows you to disable the allowlist checking without modifying your python script, thought some user doing heavy customization might want this when loading their own  classes from a yaml.
but looking at it again, it's just as easy to import this and do `target_allowlist.disable()`.

megatron/training/config/container.py
maanug-nv: naive solution would be to create a bunch of helper functions that create handle different file operations just like is being done here. eg:
```py
def msc_exists(*args, **kwargs):
    if MultiStorageClientFeature.is_enabled():
        msc = MultiStorageClientFeature.import_package()
        return msc.os.path.exists(*args, **kwargs)
    else:
        return os.path.exists(*args, **kwargs)
```

but that could be a lot of functions and would require updating.

it looks this is all the modules that we use from msc:
```
msc.open
msc.os
msc.Path
msc.torch
```

so another solution could be like this:
```py
class MaybeMSC:
    def __init__(self):
        if MultiStorageClientFeature.is_enabled():
            msc = MultiStorageClientFeature.import_package()
            self.open = msc.open
            self.os = msc.os
            self.Path = msc.Path
            self.torch = msc.torch
        else:
            import os, torch
            from pathlib import Path
            self.open = open
            self.os = os
            self.Path = Path
            self.torch = torch
maybe_msc = MaybeMSC()
...
maybe_msc.os.path.exists(yaml_path) # will automatically use the correct module        
```

maanug-nv: would prefer to do that in a separate pr. if you are ok with one of these approaches, ill create an issue

jaredcasper (approved):

megatron/training/config/container.py
maanug-nv: created #4507 for this

cuichenx (approved):
LGTM

PR #2256, opened 2025-11-14
[Dev] Feature: linear cross entropy fusion

Jianbing-D:
# What does this PR do ?

This PR introduces an implementation that fuses Linear Layer of lm_head and Cross-Entropy, in order to avoid materializing the intermediate logits tensor, helping reducing memory footprint.

PR to the main branch: https://github.com/NVIDIA/Megatron-LM/pull/2206

:warning: For major changes (either in lines of code or in its impact), please make sure to first share discuss a design-doc with the team.  

# Details about this feature
Training LLM typically involves a two-stage pipeline at the output layer: hidden states are projected into vocabulary `logits` via a linear transformation (lm_head Layer), followed by Cross-Entropy loss computation against target tokens. While conceptually simple, such workflow incurs substantial overhead. The intermediate logits tensor, with dimension proportional to `batch size, sequence length, and vocabulary size`, must be fully materialized in GPU memory, even though only one target token per position is ultimately used. This leads to significant memory footprint and bandwidth consumption, limiting scalability and slowing training throughput. The following code snippet might better illustrate that workflow:
```python
hidden_state = xxx # shape = [batch, seqlen, dim]
weight = lm_head.weight # shape = [vocabsize, dim]
labels = xxx # shape = [batch, seqlen]

logits = hidden_state @ weight.T
loss = torch.nn.functional.cross_entropy(logits, labels, reduction="none")
```

On top of the `local logit` tensor, other techniques might need some other intermediate buffers for collecting full information across all GPUs. For example, the following snippet is a TP compatible layer, comprised of torch native ops:
```python
tp_rank = xxx
tp_world_size = xxx

logits = hidden @ weight.T

whole_logits = torch.empty(
    (logits.shape[0], logits.shape[-1] * tp_world_size),
     dtype=logits.dtype,
     device=logits.device,
)
whole_logits_ref = [
     whole_logits[..., i * logits.shape[-1] : (i + 1) * logits.shape[-1]]
     for i in range(tp_world_size)
]
dist.all_gather(whole_logits_ref, logits, group=tp_group)

logprobs = torch.nn.functional.cross_entropy(
    whole_logits.view(-1, whole_logits.shape[-1]), labels.view(-1), reduction=reduction
)
```

By fusing `Linear` and `Cross-Entropy` into one single operation, this PR could help avoid materializing the intermediate `logit` tensor.
```python
hidden_state = xxx # shape = [batch, seqlen, dim]
weight = lm_head.weight # shape = [vocabsize, dim]
[...]

PR #2206, opened 2025-11-11
Feature: linear cross entropy fusion

Jianbing-D:
# What does this PR do ?

This PR introduces an implementation that fuses Linear Layer of `lm_head` and Cross-Entropy, in order to avoid materializing the intermediate `logits` tensor, helping reducing memory footprint.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share discuss a design-doc with the team.  

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
        C1[Expert Review] --> C2[Final Review]
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [ ] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [x] I have added relevant unit tests
- [x] I have added relevant functional tests
- [x] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [x] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

The following process is enforced via the CODEOWNERS file for changes into `megatron/core`. For changes outside of `megatron/core`, it is up to the PR author whether or not to tag the Final Reviewer team.

<details>
<summary>For MRs into `main` branch</summary>

#### (Step 2): Collect the expert reviewers reviews

1. Attach the `Expert Review` label when your PR is ready for review.
2. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

:warning: Only proceed to the next step once all reviewers have approved, merge-conflict are resolved and the CI is passing.  
Final Review might get declined if these requirements are not fulfilled.

#### (Step 3): Final Review

1. Add `Final Review` label
2. GitHub auto-assigns final reviewers based on your changes. They will get notified and pick up your PR soon.

#### (Optional Step 4): Cherry-pick into release branch

If this PR also needs to be merged into `core_r*` release branches, after this PR has been merged, select `Cherry-pick` to open a new PR into the release branch.

</details>

<details>
<summary>For MRs into `dev` branch</summary>
The proposed review process for `dev` branch is under active discussion.

MRs are mergable after one approval by either `eharper@nvidia.com` or `zijiey@nvidia.com`.
</details>

### Merging your PR

[...]

PR #4312, opened 2026-04-15
feat: Optimize memory footprint of long-context training via fused kernel and chunking

terminator123:
# What does this PR do ?
Introduces a fused CrossEntropy kernel and output chunking strategy to reduce the peak memory consumption of logits during long-context training.

# Technical Details
This PR addresses the high VRAM usage bottleneck in large-scale training by targeting the logits tensor memory footprint.

- Fused Kernel: Utilizes the Liger-Kernel's fused CrossEntropy implementation to reduce intermediate memory overhead.

- Output Chunking: Implements an output chunking mechanism where the model's output is processed in blocks.

- Memory-Specific Optimization: The peak memory is reduced by a factor proportional to the number of chunks (1/N ). The more chunks the output is divided into, the lower the peak memory .


Phlip79:
We are in the process of developing this same feature: https://github.com/NVIDIA/Megatron-LM/pull/2206. @Jianbing-D  can you please take a look at this?

Jianbing-D:
Hi @terminator123,

We have similar feature already merged to `dev` branch. https://github.com/NVIDIA/Megatron-LM/pull/2256

And regarding your PR, here are some questions:
1. Is there any measurement numbers regarding your feature? like latency of fwd pass and bwd pass, as well as storage. Like what we did here: https://github.com/NVIDIA/Megatron-LM/pull/2256
2. Your kernels are written with OAI triton, but that library failed to achieve good performance on Blackwell GPUs. If you could provide any perf numbers, that would be great for us to determine whether your kernels are good enough.
3. Your feature seems not support `reduction=none`. Please correct me if I understand it wrong. If this feature doesn't support `reduction=none`, how could the users handle `token masking and padding`, where `invalid tokens` shall have `zero` as grad, whilst `valid tokens` shall have valid grad values.

terminator123:
> Hi @terminator123,
> 
> We have similar feature already merged to `dev` branch. #2256
> 
> And regarding your PR, here are some questions:
> 
> 1. Is there any measurement numbers regarding your feature? like latency of fwd pass and bwd pass, as well as storage. Like what we did here: [[Dev] Feature: linear cross entropy fusion #2256](https://github.com/NVIDIA/Megatron-LM/pull/2256)
> 2. Your kernels are written with OAI triton, but that library failed to achieve good performance on Blackwell GPUs. If you could provide any perf numbers, that would be great for us to determine whether your kernels are good enough.
> 3. Your feature seems not support `reduction=none`. Please correct me if I understand it wrong. If this feature doesn't support `reduction=none`, how could the users handle `token masking and padding`, where `invalid tokens` shall have `zero` as grad, whilst `valid tokens` shall have valid grad values.

I took a look at PR #2256. It seems that the --cross-entropy-loss-fusion flag currently only supports Blackwell GPUs ？ I‘m using Hopper now

PR #4329, opened 2026-04-15
Add hybrid FSDP unit module support

Phlip79:
Adds hybrid-model FSDP unit coverage by registering the non-standard hybrid layer types in the default FSDP wrapping policy and by running the existing Megatron-FSDP E2E path through a hybrid Mamba+MoE model.

Changes:
- Register `MambaLayer` and `MoETransformerLayer` in the default Megatron-FSDP unit modules for `optim_grads_params`.
- Register `MambaLayer` and `MoETransformerLayer` in the Torch FSDP2 default wrapping set.
- Allow the MoE expert-parallel overlap guard to accept the expanded default unit-module list.
- Switch the Megatron-FSDP E2E helper from the GPT builder to the hybrid builder with a `MEME/ME` Mamba+MoE pattern.

Testing:
- `uv run isort megatron/core/distributed/fsdp/mcore_fsdp_adapter.py megatron/core/distributed/torch_fully_sharded_data_parallel.py tests/unit_tests/distributed/megatron_fsdp/utils.py`
- `git diff --check`
- `uv run pytest -q tests/unit_tests/distributed/megatron_fsdp/test_mcore_tensor_parallelism_detect.py` could not run in this local environment because `torch` is not installed in the local `uv` environment.


ilml (approved):

shjwudp (approved):
LGTM, thanks @Phlip79 !

1. This PR sets MambaLayer as the default sharding layer.
2. IIt adds MFSDP-TP compatibility for MambaLayer — the layer now sets `partition_dim` on parameters directly, rather than on modules.

It would be great to make Mamba + MoE the default model for E2E tests if you're up for it — this would help us catch and track Mamba-related regressions more reliably. Reference: [tests/unit_tests/distributed/megatron_fsdp/utils.py#L55](https://github.com/NVIDIA/Megatron-LM/blob/main/tests/unit_tests/distributed/megatron_fsdp/utils.py#L55)

Phlip79:
Closing this PR in favor of https://github.com/NVIDIA/Megatron-LM/pull/4467 which solves the same issue.

cspades:
@Phlip79 Let's just keep the stuff we need:

1. MambaLayer and MoE layers into FSDP unit modules.
2. Tests for hybrid models.
3. Rebase on main branch if some things are already implemented, remove all the other code that may not be relevant to what we already have working on main branch.

Phlip79:
@cspades Updated this PR to match the requested scope:

- rebased on current main and squashed to one scoped commit
- final diff only registers Mamba/MoE hybrid layer units for FSDP and switches the FSDP E2E helper to a hybrid Mamba+MoE model
- removed the TP annotation fallback/tests from the PR diff

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -173,9 +174,13 @@ def __init__(
...
-            assert self.fsdp_unit_modules == [TransformerLayer], (
+            assert self.fsdp_unit_modules in (
+                [TransformerLayer],
+                [TransformerLayer, MoETransformerLayer, MambaLayer],
+            ), (
                 "EP overlap with FSDP currently requires fsdp_unit_modules "
-                f"to be [TransformerLayer], got {self.fsdp_unit_modules}."
+                "to be [TransformerLayer] or the default FSDP unit modules, "
+                f"got {self.fsdp_unit_modules}."
             )
```
cspades, line 185: Should be:
```
# Supported MCore Models
assert all(x in [TransformerLayer, MoETransformerLayer, MambaLayer] for x in self.fsdp_unit_modules)
```

megatron/core/distributed/torch_fully_sharded_data_parallel.py
```diff
@@ -59,6 +60,8 @@ def __init__(
         module: torch.nn.Module,
         sub_modules_to_wrap: Set[torch.nn.Module] = {
             TransformerLayer,
+            MoETransformerLayer,
+            MambaLayer,
```
cspades: Did you test FSDP2? Do we need it in this PR?

Phlip79:
> 2. Tests for hybrid models.

@cspades anything else aside from the changes I made to utils.py?

megatron/core/distributed/torch_fully_sharded_data_parallel.py
Phlip79: This was a mistake and should not have been added.

cspades (approved):
LGTM, I want to test it one more time since I was using Xiaowei's branch that was like 3 commits ahead of main, but I think it should be fine.

ericharper (approved):

jaredcasper (approved):

PR #4370, opened 2026-04-17
Add fault injection support via nvidia_resiliency_ext.

hexinw-nvidia:
Add fault injection support via nvidia_resiliency_ext
                                                                                                                                   
Adds fault injection capability for resilience testing, driven by CLI args.
Fault injection is a no-op unless --fault-injector-ranks or --fault-injector-num-ranks is explicitly passed.                                                                                                                               
                                                                                                                                        
  - **megatron/core/fault_injector.py**: new module wrapping                                                                                
    nvidia_resiliency_ext.shared_utils.inject_fault; provides
    setup_fault_injection() (broadcasts fault plan across ranks) and                                                                    
    re-exports maybe_raise_workload_exception()                                                                                         
  - **megatron/training/config/resilience_config.py**: FaultInjectorConfig
    dataclass (ranks, fault type/probabilities, MTTI-based delay, seed)                                                                 
    following the RerunStateMachineConfig pattern                                                                                       
  - **megatron/training/arguments.py**: registers fault injector args via                                                                   
    ArgumentGroupFactory(FaultInjectorConfig); all default to None                                                                        
  - **megatron/training/training.py**: lazily imports and calls                                                                             
    setup_fault_injection() at train() entry when args are set;
    calls maybe_raise_workload_exception() each step after the first


megatron/training/config/resilience_config.py
```diff
@@ -40,3 +40,32 @@ class StragglerDetectionConfig:
...
+    fault_injector_num_ranks: Optional[int] = None
+    """Number of ranks to inject faults on (random selection)."""
+
+    fault_injector_fault_types: Optional[str] = None
+    """Comma-separated list of fault types to inject (e.g. 'hang,crash')."""
+
+    fault_injector_fault_probabilities: Optional[str] = None
+    """Comma-separated list of fault probabilities (normalized at runtime)."""
+
+    fault_injector_fault_delay: Optional[float] = None
```
sbak5, line 60: Along with delay, you need an option to trigger fault by the iteration count. 
Because, this delay may fire too early if the init / iteration time is too long. 

I'd use the iteration number as the major variable to control when a fault is injected.

jaredcasper (approved):

megatron/core/fault_injector.py
maanug-nv, line 1: Why do we take an arbitrary `cli_args` object for all these functions instead of the FaultInjectorConfig? Seems like that config should just live in this file right?

maanug-nv: I can see that currently, `args` is provided in `training.py` to these functions for convenience. but once the [ConfigContainer](https://github.com/NVIDIA/Megatron-LM/pull/4227) is in place, I can rewrite the calls in training.py to pass in the explicit config instead of the entire `args` object.

hexinw-nvidia: Refactor the code to pass FaultInjectorConfig directly. Please take a look.

megatron/core/fault_injector.py
```diff
@@ -39,19 +40,46 @@ def _require_nvidia_resiliency_ext():
...
+
+    def random(self) -> float:
+        """Return a floating-point value in the half-open interval [0.0, 1.0)."""
+        ...
+
+
+rng: _FaultInjectorRNG | None = None
+
+if TYPE_CHECKING:
+    from megatron.training.config import FaultInjectorConfig
```
maanug-nv: importing from `megatron.training` might trigger import of optional dependencies that are not installed if mcore was not installed with '--extra training' .
can we just copy the config here and reverse the direction of import?

hexinw-nvidia: Thanks. Made the change to remove the megatron.training dependency.

maanug-nv (approved):
lgtm, thanks for addressing feedback

ericharper (approved):

ericharper:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4370/commits/61791092f775985d02e91715155ab2896d8ee5a9

PR #4384, opened 2026-04-20
moe: align HybridEP buffer to combine-API chunk size to fix JIT failure

neiblegy:
# What does this PR do ?

Fix a JIT compile-time `static_assert` failure in deep_ep's HybridEP `combine_kernel`, triggered when the per-rank token count is aligned to 16 (deep_ep's TMA alignment) but not to the combine-API chunk size (default 64).

## Contribution process

This issue was originally reported and patched downstream in [NVIDIA-NeMo/Megatron-Bridge#3294](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3294). 

## Root cause
- `deep_ep/hybrid_ep_buffer.py:141-142` rounds `num_of_tokens_per_rank` up to a multiple of **16** for TMA load/store alignment. (hybrid-ep branch)
- `deep_ep/csrc/hybrid_ep/backend/hybrid_ep_backend.cuh:4552`, inside `combine_kernel`, carries:
  ```cpp
  static_assert(MAX_NUM_OF_TOKENS_PER_RANK % NUM_OF_TOKENS_PER_CHUNK == 0,
                "MAX_NUM_OF_TOKENS_PER_RANK must be multiple of NUM_OF_TOKENS_PER_CHUNK.");
  ```
  where `NUM_OF_TOKENS_PER_CHUNK` defaults to 64 (configurable via `NUM_OF_TOKENS_PER_CHUNK_COMBINE_API`).
A value that is a multiple of 16 is **not** automatically a multiple of 64. For example, on H100 with TP=1, EP=8, PP=1, mbs=2, cur_len=1776, the per-rank token count is 3552, and `3552 % 64 = 32`. The Python layer happily lets it through, and the failure surfaces only at JIT compile time inside the combine kernel:

31d11c1cd6f5a1ef3ed61d9a8cb5ad03

4dfcaf80384cb14e7811042f121f661f

The relevant tail is:
```
RuntimeError: Failed to compile the code, compile command: nvcc ...
.../hybrid_ep/jit/.../2048-3552-32-...-64-...-rank-0-...so
```
where `3552` is `MAX_NUM_OF_TOKENS_PER_RANK` and `64` is `NUM_OF_TOKENS_PER_CHUNK` — exactly the values that violate the `static_assert`.

## Fix
In `megatron/core/transformer/moe/fused_a2a.py`:
- Add `_get_hybrid_ep_buffer_alignment()`, which **probes** the actual combine-API chunk size by instantiating a throw-away `hybrid_ep_cpp.Configurer` and reading `buffer_config.num_of_tokens_per_chunk_combine_api`. The `Configurer` constructor only queries CUDA device properties and fills a host-side `BufferConfig` struct — no communication-buffer allocation, no JIT compilation, no NCCL/group interaction — so this is a microsecond-scale probe.
- `init_hybrid_ep_buffer()` rounds `seq_len` up to that chunk size before constructing `HybridEPBuffer`.
Probing rather than reading the env var directly was chosen so that mcore:
- tracks deep_ep's built-in default across versions (e.g. 64 today, 128 in older builds) without having to hard-code anything;
- automatically honors any user-set `NUM_OF_TOKENS_PER_CHUNK_COMBINE_API` override, since deep_ep itself reads the same env var inside `Configurer`.
A best-effort fallback path (env var → hard-coded 64) is kept for the unlikely case that the probe fails (e.g. future deep_ep API change). It is not expected to be reachable on any currently-supported deep_ep release.
Only the **combine-API** chunk size needs alignment. The `dispatch_kernel` has no equivalent `static_assert` on `MAX_NUM_OF_TOKENS_PER_RANK`; its `num_of_dispatch_chunks` is sized via host-side ceil division and tolerates a non-aligned `max_num_of_tokens_per_rank`. Aligning to the combine chunk is therefore both necessary and sufficient.

### DIFF
--- a/megatron/core/transformer/moe/fused_a2a.py
+++ b/megatron/core/transformer/moe/fused_a2a.py
@@ -13,6 +13,9 @@ try:
 except ImportError:
     HAVE_DEEP_EP = False
 
+import math
+import os
+
 import torch
 
 _buffer = None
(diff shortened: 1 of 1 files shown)


maanug-nv:
Hi @neiblegy , working on finding an available reviewer. in the meantime, can you run the formatter: `uv run --only-group linting bash tools/autoformat.sh`

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -309,11 +355,15 @@ def init_hybrid_ep_buffer(
...
+
+    chunk_size = _get_hybrid_ep_buffer_alignment()
+    aligned_seq_len = math.ceil(seq_len / chunk_size) * chunk_size
+
     global _hybrid_ep_buffer
     _hybrid_ep_buffer = HybridEPBuffer(
         group=group,
         hidden_dim=hidden_dim,
-        max_num_of_tokens_per_rank=seq_len,
+        max_num_of_tokens_per_rank=aligned_seq_len,
```
Autumn1998, line 366: Thank you. I understand this is a hybrid EP issue — I can add the corresponding checks and calculations in hybrid EP to ensure the buffer is an integer multiple of the chunk size during initialization. This way we can avoid making the fuse a2a part more complex. Does that seem reasonable to you? Or do you think the changes in the mcore part of this PR are still necessary?

neiblegy: Your approach sounds more reasonable. If hybrid EP guarantees this alignment internally, then this PR is no longer necessary, and the overall design is cleaner and more appropriate this way.

Autumn1998: https://github.com/deepseek-ai/DeepEP/pull/638/commits
The PR in hybrid-ep

yaoyu-33:
deep EP side PR merged.

PR #4399, opened 2026-04-20
Megatron-FSDP: extract several allocator classes into allocators.py

wujingyue:
## Summary
- Move `Bucket`, `TemporaryBucketAllocator`, `StorageResizeBasedBucketAllocator`, `RotaryBucketAllocator`, and `MultiGroupUBRAllocator` (plus the private `_p_assert`/`_alloc_storage`/`_free_storage` helpers) out of `param_and_grad_buffer.py` into a new `allocators.py`.
- Move `GlobalMemoryBuffer` and `get_global_memory_buffer` from `utils.py` into `allocators.py` — they are allocation primitives (`RotaryBucketAllocator` routes every tensor request through them), and nothing outside the `megatron_fsdp` subpackage imports them from `utils`.
- Move the `nccl_allocator` try/except shim and the `NCCL_ALLOCATOR` constant into `allocators.py`, since `MultiGroupUBRAllocator.__init__` needs `nccl_allocator` at construction time.
- `param_and_grad_buffer.py` re-imports all allocator names from `.allocators` so the public surface is unchanged; no other file needed to be touched.
- Pure code move — no behavior change.

## What stays in `param_and_grad_buffer.py`
- **`FixedPoolAllocator`** is tightly coupled to `ParameterGroup` (it reads `.fsdp_unit_id`, `.dtype`, and `.params` on each group), so it stays colocated with `ParameterGroup` in `param_and_grad_buffer.py`. An earlier version of this PR moved it with a `TYPE_CHECKING` forward-reference hack; that was reverted in favor of leaving it in place until the coupling to `ParameterGroup` can be refactored separately.
- `NCCL_MEMORY_POOL` stays here as well — it's mutated as module-global state by `ParamAndGradBuffer`, not an allocator.

## Notes on the boundary
- The `nccl_allocator` shim uses the `from megatron.core import nccl_allocator` form (equivalent to `import … as …`, satisfies ruff PLR0402).
- Now-unused imports (`traceback`, `typing.Any`) have been removed from `param_and_grad_buffer.py`; `defaultdict` stays (still used by `FixedPoolAllocator`). `operator`/`reduce`/`nullcontext` have been removed from `utils.py` now that `GlobalMemoryBuffer` no longer lives there.

## Test plan
- [x] `isort` on all touched files
- [x] `python -m py_compile` on all touched files
- [x] Import-smoke: re-exported names (`Bucket`, four allocators, `nccl_allocator`, `NCCL_ALLOCATOR`, `get_global_memory_buffer`) are identity-equal to the originals from `allocators.py`; `FixedPoolAllocator` is in `param_and_grad_buffer.py` only; `utils.py` no longer exposes `GlobalMemoryBuffer`/`get_global_memory_buffer`
- [x] `pytest tests/unit_tests/distributed/megatron_fsdp/` locally on a 2-GPU box — all world-size-1 tests pass (40 passed, 145 skipped); 198 failures and 6 xfails are topology preconditions (tests require 8 ranks and fail in `DeviceMesh` init before reaching our code)
- [x] Basic allocate/free round-trip on CPU for `TemporaryBucketAllocator` and `StorageResizeBasedBucketAllocator`
- [x] CI (`Run tests` label — non-numerical refactor, no golden-value movement expected)


megatron/core/distributed/fsdp/src/megatron_fsdp/allocators.py
```diff
@@ -0,0 +1,353 @@
...
+# limitations under the License.
+
+import dataclasses
+import logging
+import traceback
+from typing import Any, Callable, Optional
+
+import torch
+
+from .utils import get_global_memory_buffer, log_single_rank
```
cspades: We could also move the GlobalMemoryBuffer here, it's super relevant to allocation and not relevant to much else, IIRC!

(Also, I believe that the GlobalMemoryBuffer import logic is currently broken, and I haven't told anyone about it. It _**always**_ uses the Megatron-FSDP implementation, when it should be using the Megatron-Core version if installed. Can you fix it? 👀)

wujingyue: Done.

PR #4403, opened 2026-04-21
fix(checkpoint_inspector): allow empty --param-to-param-group-map-json

DAISY-gh:
## Fix

`tools/checkpoint/checkpoint_inspector.py convert-torch-dist-to-fsdp-dtensor` crashes with `FileNotFoundError: [Errno 2] No such file or directory: '{}'` when `--param-to-param-group-map-json` is not provided — which is the case in every example shipped in the module docstring.

Root cause: the option's default is the literal string `"{}"` while the implementation calls `open(param_to_param_group_map_json, "r")`, i.e. treats the value as a file path.

Changes:
- `default="{}"` → `default=None`; help text updated to reflect that the value is a file path.
- Skip `open()` when the flag is unset; fall back to an empty dict. All downstream uses already short-circuit on an empty map.

After this fix, the existing Qwen3.5-VL / generic SWiGLU examples in the docstring run without specifying the flag.


wplf (approved):

asolergi-nv:
/ok to test [47eb225](https://github.com/NVIDIA/Megatron-LM/pull/4403/commits/47eb225b282f5a23ce8ca16d35b7ea1a813bf36e)

PR #4417, opened 2026-04-21
style: modernize typing annotations in param_and_grad_buffer.py

wujingyue:
## Summary
- Replace deprecated `typing.{List,Dict,Tuple,Optional}` aliases in `megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py` with the PEP 585 / PEP 604 built-in generics and union syntax (`list[...]`, `dict[...]`, `tuple[...]`, `X | None`). The project requires Python >= 3.12, so these aliases are no longer needed.
- Move `Callable` from `typing` to `collections.abc` (its non-deprecated home).
- Update matching references in docstrings so they stay in sync with the annotations.

## Test plan
- [x] CI `Run tests` passes.

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
@@ -652,53 +652,35 @@ class FixedPoolAllocator(TemporaryBucketAllocator):
     def __init__(
         self,
         name: str,
-        fsdp_param_groups: List["ParameterGroup"],
+        fsdp_double_buffer_units: List[int],
+        bucket_to_unit: List[int],
         size: int = 2,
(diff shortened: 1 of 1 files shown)


cspades (commented):
We need to support older Python versions (3.9 or below) ideally (even if MCore has moved on to 3.12). @shjwudp should we wait until our customers are all okay with it?

wujingyue:
> We need to support older Python versions (3.9 or below) ideally (even if MCore has moved on to 3.12). @shjwudp should we wait until our customers are all okay with it?

Good question--I'm curious as well what Python versions our customers are on. It'll impact what language features we can use for other PRs. 

FYI, standard-collection type hints are supported in Python >= 3.9 (https://peps.python.org/pep-0585/) and 3.9 is about five years old.

Phlip79 (approved):

PR #4419, opened 2026-04-21
[Inference] Protocol Robustness Improvements in Coordinator (Follow-up to #4176)

DhineshPonnarasan:
## Summary
This PR implements protocol robustness improvements for the inference coordinator as part of follow-up work for #4176.

The focus is to ensure coordinator communication is resilient to malformed or unexpected inputs without impacting stability or existing routing behavior.

---

## Scope
This PR is intentionally scoped only to protocol robustness to keep changes focused and easy to review.

**Planned follow-up PRs:**
- PR 2: Observability enhancements (metrics, routing quality, latency)  
- PR 3: Metadata policy extensions (TTL-based expiration)  

---

### 1. Input Validation
- Added strict validation for inbound messages (structure, required fields, types)  
- Enforced type checks (e.g., `SET_GENERATION_EPOCH` requires integer epoch)  
- Added validation for prompt payload and list contents  

---

### 2. Crash Safety
- Guarded multipart frame handling in:
  - Startup registration path  
  - Runtime receive loop  

- Removed runtime assertions from protocol paths  
- Ensured malformed inputs never crash the coordinator  

---

### 3. Safe Handling of Invalid Inputs
Gracefully handles:
- Unknown message types  
- Malformed payloads  
- Unknown senders  
- Duplicate engine identities  

Invalid inputs are safely ignored with structured logging  

---

### 4. Error Classification
- Clear separation between:
  - `client_error` → malformed or invalid client inputs  
  - `internal_error` → unexpected system failures  

- Fixed classification boundary in `SUBMIT_REQUEST` to avoid misclassifying internal exceptions  

---

### 5. State Integrity
- Ensured request-id mappings are always cleaned up on failure paths  
- Prevents state leaks in long-running coordinator execution  

---

### 6. Logging Consistency
- Standardized invalid-input handling using `_log_protocol_error`  
- Removed inconsistent `logging.warning` / `logging.info` usage in protocol paths  

---

## Testing

Added focused protocol-robustness tests covering:
- Malformed and invalid messages  
- Multipart frame edge cases (0 / 1 / >2 frames)  
- Invalid `SUBMIT_REQUEST` payloads  
- Error classification (`client_error` vs `internal_error`)  
- Duplicate engine identity handling  
- Unknown sender handling  
- State cleanup (no request-id leaks)  
- Stress-style invalid input sequences  

---

## Results
```bash
pytest test_data_parallel_inference_coordinator.py -k ProtocolRobustness -q
```

**Result:**
- 22 passed, 0 failed  

**Note:**  
Full test suite includes GPU/distributed tests. Any failures observed locally are due to lack of CUDA devices (`torch.cuda.device_count() == 0`) and are unrelated to this PR.

---

## Out of Scope (Explicit)
This PR does **not** modify:
- Routing logic  
- Prefix cache behavior or eviction policies  
- Observability / metrics systems  
- Performance optimizations beyond safety  

---

## Result
The coordinator is now:

- Resilient to malformed or adversarial inputs  
- Protected against crash scenarios  
- Consistent in error classification  
- Safe for long-running distributed inference workloads


santhnm2:
@DhineshPonnarasan what is the intended use case for these changes? The data parallel inference coordinator is meant to be used for evaluations in a controlled environment where we would not expect malformed inputs.

DhineshPonnarasan:
> @DhineshPonnarasan what is the intended use case for these changes? The data parallel inference coordinator is meant to be used for evaluations in a controlled environment where we would not expect malformed inputs.


Thanks @santhnm2 , that is a fair point.

The intent here is not to handle adversarial traffic, but to provide defense-in-depth for long-running distributed evaluation jobs.

Even in controlled environments, malformed protocol inputs can still occur due to non-malicious causes, for example:
- Version skew between components during rolling updates  
- Partial or corrupted messages from restarts or interrupted sockets  
- Duplicate or stale identities during reconnect races  
- Bugs in test harnesses or orchestration scripts  
- Gradual payload drift as new control signals are introduced  

Before this change, some of these cases could terminate the coordinator or leave inconsistent state, which in practice leads to failed long-running eval jobs and difficult-to-debug issues.

This PR makes those failure modes safe by:
- Rejecting invalid inputs early  
- Ensuring the coordinator remains alive  
- Clearly classifying errors (client vs internal)  
- Cleaning up request state on failure paths  

Valid message handling and existing behavior remain unchanged.

So the goal is improved reliability and debuggability within controlled evaluation workflows, where a single malformed message should not bring down the entire run.

Please let me know if you’d like any of these checks scoped down or adjusted — happy to refine.

To keep reviews focused, I’m planning to follow up with two separate PRs:
- Observability enhancements (metrics, routing quality, latency)  
- Metadata policy extensions (TTL-based expiration)  

Thank you for your guidance.

PR #4423, opened 2026-04-22
fix(megatron-fsdp): compute SWiGLU/GDN split in item coordinates for non-DTensor optimizer states

xuwchen:
# What does this PR do ?
`split_swiglu_linear_fc1` and `split_gdn_fused` originally used `data.numel() // tp_mesh.mesh.numel()` to compute component slices (`w_slice`/`v_slice` for SWiGLU, `comp_slice` for GDN), then intersected them with `fsdp_slice` to determine each rank's portion. When `data` is a DTensor, `data.numel()` returns the global element count (=`param.numel()`), so component slices and `fsdp_slice` are both in item coordinates and the intersection is valid. However, when `data` is a plain Tensor (optimizer states created by FusedAdam), `data.numel()` returns the local shard size on this rank, making the component slices's range far smaller than the item coordinate space and resulting in an intersection across incompatible coordinate systems.

The fix replaces `data.numel()` with `dist_param.numel()` (`dist_param` is always a DTensor and therefore reflects the global element count), ensuring `fsdp_slice` and all component slices are in item coordinates.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/core/transformer/fsdp_dtensor_checkpoint.py
```diff
@@ -273,10 +273,26 @@ def is_swiglu_key(key):
...
+        global_shape = dist_param.shape
+        if isinstance(data, DTensor):
+            assert data.shape == global_shape, (
+                f"DTensor shape mismatch: data.shape={data.shape} vs "
+                f"dist_param.shape={global_shape}"
+            )
+        assert global_shape[swiglu_shard_axis] % 2 == 0, (
+            f"SWiGLU FC1 must have even global size along axis {swiglu_shard_axis}, "
+            f"got {global_shape[swiglu_shard_axis]} (global_shape={list(global_shape)})"
         )
```
cspades, line 296: So basically, if data is a DTensor, we should ensure it matches the state DTensor. And then, we always use the global shape. (I've hit this error before when `FusedAdam` was broken, which I fixed here: https://github.com/NVIDIA/TransformerEngine/pull/2795)

When would `data` not be a DTensor in the context of this function? 👀 Is it a bug or just so this function can be used outside of MLM / MBridge checkpointing?

cspades (approved):
I think this is a versatile / good change, just curious about why you came across this error as I did! (See my comment about the TransformerEngine bug.)

megatron/core/transformer/fsdp_dtensor_checkpoint.py
xuwchen: Good point! Your fix in https://github.com/NVIDIA/TransformerEngine/pull/2795 would indeed resolve this issue by making `FusedAdam` optimizer states (`exp_avg`, `exp_avg_sq`) DTensors, so `data` would always be a DTensor and the original code would work correctly. 

The case we've hit so far is specifically `FusedAdam` producing plain Tensor optimizer states. I'm not 100% sure whether other code paths could also pass plain Tensors to this function, but since `split_swiglu_linear_fc1` doesn't enforce any type constraint on `data`, I think it's safer to keep this fix on the MCore side as defensive programming. The DTensor case is the "easy" case, we just assert shape consistency there. The real fix ensures that even if `data` is a plain Tensor, the W/V split boundaries are computed in the correct (global) coordinate system.

I've attached a visualization based on a deepseek proxy model experiment to help illustrate this. Hope this clarifies the motivation behind the fix.

swiglu_checkpoint_bug

PR #4433, opened 2026-04-22
Important bugfixes in local CG implementation that were leading to loss curve gaps for latent MoE models

jiemingz:
# What does this PR do ?
The PR specifically is addressing the following issues that were seen to affect convergence:
- the input hidden states tensor to the first layer was improperly being freed before bwd graph read it, which would cause incorrect results. This was due to an invalid optimization being made in `cuda_graphs.py` that is removed.
- [The expert bias](https://github.com/NVIDIA/Megatron-LM/blob/90e09b68c7ddbf36f168b2bdc123ed0f77fc39cd/megatron/core/transformer/moe/router.py#L183), which is persistent across forward passes, was being accumulated on graph warmup, corrupting its value.
- The expert bias is cleared depending on the `.training()` state of the router. During graph capture we set the module to `.eval()` in the case of its recomputed in the backward pass, causing the expert bias to be corrupted. 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/core/transformer/cuda_graphs.py
```diff
@@ -797,6 +797,10 @@ def create_fwd_graph(self, args, kwargs, outputs=None, clone_inputs=True):
         self.outputs = outputs
 
         # save grads and other variables that may be affected by graph warmup
+        buffer_backup = []
+        for buf in self.base_module.buffers():
+            buffer_backup.append(buf.clone())
```
mathemakitten, line 805: The corresponding restoration code below is gated on `if self.training and torch.is_grad_enabled():`, so in inference the clone here would be wasted overhead (though I don't think we have any buffers) — should we gate it?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1643,10 +1637,6 @@ def __call__(self, megatron_module, args, kwargs):
                 runner = self.get_cudagraph_runner(
                     megatron_module, args, kwargs, self.reuse_cudagraphs
                 )
-                # check if a layer is frozen during training.
```
mathemakitten, line 1646: Is this related to the convergence issue?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -913,7 +916,7 @@ def _resolve_input_buffer(ten):
                     def clone_ten(ten):
                         if not torch.is_tensor(ten):
                             return ten
-                        return torch.zeros_like(ten).requires_grad_(ten.requires_grad)
+                        return torch.clone(ten).detach().requires_grad_(ten.requires_grad)
```
mathemakitten, line 922: Why are we using real values now?

megatron/core/transformer/cuda_graphs.py
jiemingz: I think buffers are meant to be persistent with or without autograd, so its possible that an fwd-only graph would depend on buffers and not want warmup to overwrite them?
If inference isnt registering any buffers, then this would just be a noop?

megatron/core/transformer/cuda_graphs.py
jiemingz: This is trying to minimize any unintended changes warmup may have on the module being graphed.
Specifically the token dispatcher sets several `self` attributes during its forward pass. When we graph with the real values it prevents it from zeroing out these attributes, which fixes some logging issues. 
We might want think of a more robust solution forward but I think its a sufficient fix for now.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -986,17 +989,6 @@ def clone_ten(ten):
                 o.cg_buffer_metadata.fwd_cudagraph_buffer = fwd_graph_out
                 fwd_buffer_reuse_ref_count += 1
 
-        # if an input buffer requires a copy, and does not have metadata attached to it at this
```
deepakn94, line 989: Why is this being deleted?

deepakn94:
Can you list the bugfixes in the PR description?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -797,6 +797,10 @@ def create_fwd_graph(self, args, kwargs, outputs=None, clone_inputs=True):
         self.outputs = outputs
 
         # save grads and other variables that may be affected by graph warmup
```
deepakn94: Can you add an example here?

megatron/core/transformer/cuda_graphs.py
jiemingz: Yes; runner.eval() will also recurse into the base module and set it to `.eval()` as well.
however in megatron there is some code [that relies on reading the `.training() ` state](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/distributed/finalize_model_grads.py#L301) of the module.
Originally this was used to ensure the frozen encoder of a VLM would be graphed without autograd, but since then we've added better no_grad support for forward graphs so we can just remove this code.

megatron/core/transformer/cuda_graphs.py
jiemingz: This was causing a convergence issue, due to buffers being reused too early and affecting the correctness of the backward pass.
This optimization was not actually valid since the tensors this optimization was targeting were not actually able to be freed, so I'm just deleting it here. Because it only targets tensors that require an explicit copy into the graph buffer, its also not used for inference.

mathemakitten (approved):

deepakn94 (approved):

ericharper (approved):

ericharper:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4433/commits/db9e078e8c5553e85345c56714f802003f8ca287

PR #4435, opened 2026-04-22
[1/N] Megatron FSDP: Introduce Megatron-FSDP2 with per-module fully_shard() API

shjwudp:
# What does this PR do ?

## Summary
Introduce **Megatron-FSDP2** — a refactored per-module sharding implementation (`fully_shard_v2`) with FSDP2-compatible API, communication-compute overlap, activation recompute support, and a pooled memory allocator for Megatron Core.
## Key Features
- **`fully_shard()` API**: FSDP2-compatible wrapping interface that converts modules to `FSDPModule` dynamically, groups parameters into `ParameterGroup`s with flat `DataParallelBuffer`s, and installs forward/backward hooks for the unshard → forward → reshard → backward → reduce lifecycle.
- **Communication-compute overlap (two streams)**:
  - *Unshard prefetch* (`ag_stream`): all-gathers parameters for the next module while the current module computes forward.
  - *Async reduce-grad* (`rs_stream`): reduce-scatters gradients while later modules compute backward. Sliding drain rule keeps ≤ 2 gradient buffers live at any time.
- **Activation recompute support**: Derived `backward_module` tracking + persistent `unshard_done_events` prevent redundant all-gathers and premature resharding during gradient checkpointing recompute passes.
- **`TracePoolAllocator`**: Three-phase (trace → plan → optimized) pooled memory allocator using greedy left-edge interval coloring. Eliminates per-call `torch.empty` overhead with a static pool replayed across micro-batches.
- **Four sharding strategies**: `no_shard`, `optim` (ZeRO-1), `optim_grads` (ZeRO-2), `optim_grads_params` (ZeRO-3), with uneven DTensor and distributed checkpoint support.
## Files Changed (27 files, +5,014 / −9)
| Area | Files | Description |
|------|-------|-------------|
| `v2/` package | `__init__.py`, `fully_shard.py`, `fsdp_module.py`, `hooks.py`, `param_group.py`, `dp_buffer.py`, `allocator.py`, `utils.py`, `mixed_precision.py`, `design.md`, `README.md` | Core implementation and design docs |
| Integration | `mcore_fsdp_adapter.py`, `src/megatron_fsdp/__init__.py`, `megatron_fsdp.py`, `distributed_data_parallel_config.py` | Adapter supporting legacy and v2 paths |
| Example | `examples/megatron_fsdp/fsdp_toy.py` | Standalone training example |
| Tests | `tests/.../v2/test_allocator.py`, `test_param_group.py`, `test_mcore_fully_shard_api.py`, `test_checkpoint_online_convert.py` | Unit and integration tests |
## Experimental Results — Per-Module Sharding Rewrite
Verification was performed comparing the refactored implementation against the baseline.
| Metric | Baseline | Megatron-FSDP2 | Change |
|--------|----------|-----------------|--------|
| **Performance** (TFLOPs) | 496.78 | 513.90 | Slight improvement, within expected fluctuation range |
| **Memory** (GB) | 48.30 | 47.15 | Memory usage reduction |
| **Convergence** | — | — | Comparable to baseline; observed difference ≈ 0.01, likely attributable to changes in parameter grouping |
The refactored implementation achieves performance parity or better, modest memory savings, and convergence comparable to the baseline.

W&B link: https://wandb.ai/nvidia/megatron-fsdp/reports/M-FSDP-Rewrite-Convergence-Test-llama3-8b---VmlldzoxNjgzNTg3OQ

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/core/distributed/fsdp/src/megatron_fsdp/fully_shard_rewrite/fully_shard.py
```diff
@@ -0,0 +1,778 @@
...
+        functools.partial(_register_post_backward_hook, post_backward), with_kwargs=True
+    )
+
+
+def _register_post_backward_final_callback(state: _FSDPState, module: nn.Module) -> None:
+    """
+    Register the final callback that runs after all backward passes complete.
+
+    This is only registered by the root FSDP module to avoid duplicate
+    callbacks. It reshards all modules and reduces gradients at the end
```
wujingyue: @shjwudp Isn't a gradient tensor reduced after its containing FSDP module for overlapping? (I might have misunderstood your definition of "root".)

shjwudp: Per-module post_backward hooks are registered via autograd on grad-carrying inputs, but in several situations those hooks may be skipped or never fire (e.g., traceable/compiled modes that explicitly skip FSDP hooks, no-grad forwards or detached inputs, conditional/MoE usage, or nonstandard multi-forward/multi-backward scheduling).

To keep things robust, `root_backward_final_callback` acts as a safety net that, at the end of the global backward, drives each post_backward so sharding, collectives, and state cleanup still happen correctly even when per-module hooks were not registered or invoked.

examples/megatron_fsdp/fsdp_toy.py
```diff
@@ -0,0 +1,246 @@
...
+
+def build_fsdp_model(
+    dim: int,
+    n_layers: int,
+    use_megatron_fsdp: bool,
+) -> Tuple["FSDPModule", torch.distributed.device_mesh.DeviceMesh]:
+    if use_megatron_fsdp:
+        from megatron.core.distributed.fsdp.src.megatron_fsdp.uneven_dtensor import get_state_dict
+        from megatron.core.distributed.fsdp.src.megatron_fsdp.v2 import FSDPModule, fully_shard
+        sys.modules["get_state_dict"] = get_state_dict
```
wujingyue, line 79: I'm pretty nervous about this monkeypatch. I sort of know your intention of keeping the optimizer unchanged but this arguably hides too much from the user.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/allocator.py
```diff
@@ -0,0 +1,516 @@
+import dataclasses
+from typing import Dict, List, Optional, Tuple
+
+import torch
+
+from .utils import ParamGroupIdx
+
+
+@dataclasses.dataclass
+class Bucket:
```
wujingyue, line 32: This abstraction doesn't look very useful?

tests/unit_tests/distributed/megatron_fsdp/v2/test_mcore_nd_parallel.py
wujingyue, line 1: Can this PR test fully_shard directly?

There’s a ToyModel example, but it isn’t covered by CI and tests very little. I’m fine with the model being toy-sized, but I’d like key performance-related features, such as peak memory, to be tested continuously in CI.

wujingyue (changes requested):
Thanks! I'll review the high levels this week.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+
+from dataclasses import dataclass
+from typing import Optional
+
+import torch
+
+
+@dataclass(frozen=True)
+class FullyShardMixedPrecisionPolicy:
```
wujingyue, line 136:
```suggestion
class MixedPrecisionPolicy:
```
should be good enough. The import path should distinguish where it comes from.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/param_group.py
```diff
@@ -0,0 +1,244 @@
...
+    - dist_params: DTensor views into the buffer
+    - dist_grads: DTensor gradient views
+    """
+
+    def __init__(
+        self,
+        params: List[torch.nn.Parameter],
+        param_group_id: ParamGroupIdx,
+        *,
+        mp_policy: FullyShardMixedPrecisionPolicy,
```
wujingyue, line 59: Nit: Does ParameterGroup need to know **all** of MixedPrecisionPolicy? I'd probably flatten it and pass in only the attributes necessary.

tests/unit_tests/distributed/megatron_fsdp/v2/test_mcore_nd_parallel.py
shjwudp: This PR primarily focuses on functional regression for MCore, as we have a solid baseline for comparison.

Unit tests for the `fully_shard` API will also be included, but they are not the main focus at this stage. I will try leveraging AI to help generate validation checks, ensuring that critical functionalities are not broken during development.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/allocator.py
```diff
@@ -0,0 +1,564 @@
...
+
+    def free(
+        self, key: Optional[AllocatorKey] = None, *, param_group_id: Optional[AllocatorKey] = None
+    ) -> None:
+        key = _resolve_key(key, param_group_id)
+        if key in self.buckets:
+            _free_storage(self.buckets[key].data)
+
+
+class TracePoolAllocator(BucketAllocator):
```
wujingyue, line 131: I don't follow the name. Does this happen to be the old FixedPoolAllocator but with tracing?

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/allocator.py
```diff
@@ -0,0 +1,564 @@
...
+
+
+@dataclasses.dataclass
+class Bucket:
+    """Lightweight container for a temporary allocated tensor buffer."""
+
+    data: torch.Tensor
+
+
+class BucketAllocator:
```
wujingyue, line 38: I'm a bit overwhelmed by how many allocators have been implemented 😄 What allocation/deallocation algorithms do we really need? How much of it can be done with MemPool with custom allocators, which is torch native? E.g.
https://docs.pytorch.org/docs/2.12/symmetric_memory.html#torch.distributed._symmetric_memory.get_mem_pool

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/dp_buffer.py
```diff
@@ -0,0 +1,619 @@
...
+        self,
+        params: List[torch.nn.Parameter],
+        param_idx: Dict[torch.nn.Parameter, int],
+        dtype: torch.dtype,
+        device: torch.device,
+        dp_group: torch.distributed.ProcessGroup,
+        param_group_id: ParamGroupIdx,
+        *,
+        allocator: Optional[BucketAllocator] = None,
+        buffer_role: str = "model_weight",
```
wujingyue, line 302: Good to have for debugging! 

```suggestion
        name: str = "model_weight",
```
to be general. I may need it within an optimizer to convert between flat-sharded and tensor-atomic.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/dp_buffer.py
```diff
@@ -0,0 +1,619 @@
...
+        is_distributed: bool = False,
```
wujingyue, line 303: It's hard to define `is_distributed`. Some buffers might be replicated across dp_outer but sharded across dp_inner. Therefore, in the new design, I've been using Placements. 

That said, can this be computed from existing attributes? If so, it doesn't need to be another attribute or argument.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/dp_buffer.py
```diff
@@ -0,0 +1,619 @@
...
+        gradient_scaling_factor: Optional[float] = None,
+        chunk_size_factor: int = 1,
+        sharding_strategy: str = "no_shard",
```
wujingyue, line 306: This looks concerning — we’re leaking high-level abstractions like the global sharding strategy and buffer semantics into a low-level DataParallelBuffer.

I’d probably make DataParallelBuffer more self-contained, similar to DStorage in FlexShard. That feels like a cleaner abstraction and also makes the system much easier to reason about and debug.

wujingyue (commented):
About half way there--I'll review the rest by Monday!

megatron/core/distributed/fsdp/src/megatron_fsdp/fully_shard_rewrite/fully_shard.py
wujingyue: Thanks for the context! 

If a user chooses to skip the hooks, they should be responsible for invoking them manually and ensuring the calls remain symmetric (e.g., the same set of parameters are all-gathered and reduce-scattered). Correct?

We could still keep a final callback, but it should be limited to validating that the required constraints have been satisfied, rather than silently performing work on behalf of the user.

tests/unit_tests/distributed/megatron_fsdp/v2/test_mcore_nd_parallel.py
wujingyue: > Unit tests for the fully_shard API will also be included, but they are not the main focus at this stage.

I think this is really a matter of execution order.

First of all, we’ll need to split this PR into a few “substages” anyway 😄

When we do that, fully_shard would land before the MCore adapter changes. By extension, fully_shard would also be tested independently before the MCore adapter integration. Correct?

shjwudp: Sure, that’s a good suggestion—fingers crossed. It would be better if we can get the most critical parts merged first, and I’ll try to split this PR.

megatron/core/distributed/fsdp/src/megatron_fsdp/fully_shard_rewrite/fully_shard.py
shjwudp: In the final callback, the hook-related logic is primarily around gradient reduction. If users want to skip gradient reduction for certain modules or parameters, we could expose some interfaces — for example, a module/parameter tagging mechanism — to allow them to opt out selectively.

BTW, leveraging the final backward callback for state cleanup/reclamation is quite a neat and convenient pattern.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
shjwudp: I'm fine with both options. Let's leave a comment for @Autumn1998 to gather his input.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/allocator.py
shjwudp: This is an experimental allocator intended to replace FixedPoolAllocator. Its main advantage is easier debugging; however, it can fail in certain scenarios — e.g. when a validation pass (forward-only) is interleaved within the training loop, which breaks TracePoolAllocator.

You can find more details in [design.md#tracepoolallocator](https://github.com/shjwudp/Megatron-LM/blob/mfsdp_refactor_main/megatron/core/distributed/fsdp/src/megatron_fsdp/v2/design.md#tracepoolallocator).

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
```diff
@@ -0,0 +1,809 @@
...
+
+    def __init__(self):
+        self._is_root = True
+        self._post_backward_callback_queued = False
+
+
+@dataclass
+class _FSDPRootContext:
+    """
+    Runtime context shared across all FSDP modules within a single root.
```
wujingyue, line 52: To make sure I understand the definition of "root", a "root" FSDP module means an FSDP module that's not under any other FSDP modules. Correct? So a model can have multiple roots--just like roots in a forest.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
```diff
@@ -0,0 +1,809 @@
...
+    This object coordinates CUDA streams, execution ordering, and async
+    communication overlap (all-gather / reduce-scatter) during forward
+    and backward passes.
+    """
+
+    # ------------------------------------------------------------------
+    # CUDA streams (communication overlap)
+    # ------------------------------------------------------------------
+    ag_stream: torch.cuda.Stream  # all-gather / unshard stream
+    rs_stream: torch.cuda.Stream  # reduce-scatter stream
```
wujingyue, line 63: FYI, I think we'll need at least one extra for inter-node communication in HSDP/HFSDP.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
```diff
@@ -0,0 +1,809 @@
...
+    # ------------------------------------------------------------------
+    # Forward execution ordering
+    # ------------------------------------------------------------------
+    forward_order: List["FSDPModule"] = field(default_factory=list)
+    """
+    FSDP modules in actual forward execution order.
+
+    This ordering is used to:
+    - Schedule prefetching of parameters (unshard)
+    - Ensure correct overlap between compute and communication
```
wujingyue, line 86: I don't follow this. Even without an explicit order, correctness is guaranteed by letting forward wait_stream allgather and reduce_scatter wait_stream backward. What am I missing?

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
```diff
@@ -0,0 +1,809 @@
...
+
+    # ------------------------------------------------------------------
+    # Forward execution ordering
+    # ------------------------------------------------------------------
+    forward_order: List["FSDPModule"] = field(default_factory=list)
+    """
+    FSDP modules in actual forward execution order.
+
+    This ordering is used to:
+    - Schedule prefetching of parameters (unshard)
```
wujingyue, line 85: How important is prefetching with per-module control? The user can choose to fully_shard on a higher-level module if they want more parameters to be unsharded in one go, sacrificing memory. But I might be misunderstanding what you mean by prefetching.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/param_group.py
```diff
@@ -0,0 +1,329 @@
...
+        mesh: Optional[DeviceMesh] = None,
+        sharding_strategy: str = "optim_grads_params",
+        gradient_scaling_factor: Optional[float] = None,
```
wujingyue, line 62: Curious what this is for. It lacks documentation at this moment.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
```diff
@@ -0,0 +1,809 @@
...
+        # Build param name to param mapping for later lookup
+        param_to_name = {p: n for n, p in self.named_parameters()}
+        self._named_param_groups = []
+
+        for fsdp_param_group in fsdp_param_groups:
+            param_names = []
+            for param in fsdp_param_group.params:
+                param_name = param_to_name[param]
+                param_names.append(param_name)
+            self._named_param_groups.append((param_names, fsdp_param_group))
```
wujingyue, line 227: param_names should probably be put inside ParameterGroup. They are tightly coupled anyway.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
```diff
@@ -0,0 +1,809 @@
...
+        materialization_device = torch.cuda.current_device()
+        for name, m in self.named_modules():
+            if m in ignored_modules:
+                continue
+            # Skip modules that don't have meta parameters
+            if all(not p.is_meta for p in m.parameters(recurse=False)):
+                continue
+
+            m._apply(
+                lambda t: torch.empty_like(t, device=materialization_device) if t.is_meta else t,
```
wujingyue, line 283: This materializes full parameters instead of shards?

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
```diff
@@ -0,0 +1,809 @@
...
+                    "re-initializing FSDP state."
+                )
+            if ctx.reduce_grad_buckets.get(id(child_module)):
+                raise RuntimeError(
+                    "_init_fsdp_state cannot be called while a child FSDPModule "
+                    "has pending reduce-scatter operations. All children must have "
+                    "completed gradient reduction before re-initializing FSDP state."
+                )
+
+        root_context = _FSDPRootContext(
```
wujingyue, line 345: I'm sure I'm missing something--how did you make sure this is only constructed for the root module?

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/utils.py
```diff
@@ -0,0 +1,75 @@
...
+        ctx.post_backward = post_backward
+        return inputs
+
+    @staticmethod
+    def backward(ctx, *grads: torch.Tensor):
+        ctx.post_backward()
+        return (None,) + grads
+
+
+def _replace_module_parameter(module: nn.Module, name: str, new_param: nn.Parameter):
```
wujingyue, line 52:
```suggestion
def _replace_module_parameter(module: nn.Module, fqn: str, new_param: nn.Parameter):
```

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
```diff
@@ -0,0 +1,809 @@
...
+        # Create parameter groups
+        fsdp_param_groups = _get_module_fsdp_param_groups(
+            self,
+            mp_policy=mp_policy,
+            mesh=mesh,
+            ignored_params=ignored_params,
+            allocator=bucket_allocator,
+            gradient_scaling_factor=gradient_scaling_factor,
+        )
+        setattr(self, "_fsdp_param_groups", fsdp_param_groups)
```
wujingyue, line 216:
```suggestion
        self._param_groups = fsdp_param_groups
```

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
```diff
@@ -0,0 +1,809 @@
...
+        for module_index, candidate_module in enumerate(module_order):
+            if candidate_module is module:
+                if module_index + 1 >= len(module_order):
+                    return []
+                return [module_order[module_index + 1]]
+
+        raise AssertionError("Current module not found in forward module order")
+
+
+class FSDPModule(nn.Module):
```
wujingyue, line 166: Since it's a mixed in, I don't think it need/should be an nn.Module. Refer to https://github.com/pytorch/pytorch/blob/67464b8261bc79dd8516f95964e4095d0546c533/torch/distributed/fsdp/_fully_shard/_fully_shard.py#L318

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
```diff
@@ -0,0 +1,809 @@
...
+) -> List[ParameterGroup]:
+    """
+    Group module parameters by (device, dtype, requires_grad) and create ParameterGroups.
+
+    Parameters are grouped because they share the same buffer management
+    and sharding strategy. Each group gets its own DataParallelBuffer.
+    """
+    param_groups = {}
+
+    for param in module.parameters():
```
wujingyue, line 783: Something seems to be missing. Shouldn't we skip any submodule that is already an FSDPModule? Because it forms its own unit.

wujingyue (commented):
I made to fully_shard.py! I will focus on mixed precision tomorrow. 

In general, as I expected earlier, I’m more concerned about the low-level constructs like DataParallelBuffer and allocators and Optimizer. I’ll probably end up rewriting them yet again to properly support the different sharding strategies and formats. I started some of this work in https://github.com/NVIDIA/Megatron-LM/pull/4835 and added you as a reviewer. We should try to converge on the design.

The high-level constructs also need a fair amount of cleanup and/or clarification, but I can leave comments on the PRs after you split this one up.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/param_group.py
```diff
@@ -0,0 +1,329 @@
...
+            if self.main_weight_buffer is not None:
+                mbuf = self.main_weight_buffer
+                data = mbuf.get_item(self.param_idx[param], only_shard=is_param_shard)
+            elif self.model_weight_buffer is not None:
+                wbuf = self.model_weight_buffer
+                data = wbuf.get_item(self.param_idx[param], only_shard=is_param_shard)
+            else:
+                data = param.data.detach()
+
+            dist_param = torch.nn.Parameter(
```
wujingyue, line 300: Question: Conceptually, we need two sets of nn.Parameters, one for model computation (in low precision; views of model_weight_buffer) and the other for optimization (in high precision; views of main_weight_buffer). How's that handled in the code?

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
Autumn1998: I'm fine with both options also. `MixedPrecisionPolicy` also make sense to me

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/param_group.py
Autumn1998: The param group indeed doesn't need to know every member of MixedPrecisionPolicy, but it does need to use many parts of it. I think passing in the policy directly is cleaner and more beneficial for future extensibility.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/dp_buffer.py
shjwudp: I agree. Do you think implementing the placement idea based on this code would introduce any conflicts?

If not, I would strongly prefer that we move forward and incorporate this idea into v2. cc @Autumn1998

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
shjwudp: It is defined in this way, and the metaphor is very vivid.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
shjwudp: _Ensure correct overlap_ ere means allowing the communication and computation pipelines to run efficiently in parallel instead of random blocking each other. Because there are data dependencies between them, establishing the right execution order is crucial.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
shjwudp: I see, thanks for point it out.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/fsdp_module.py
shjwudp: Yes.

wujingyue: Is that problematic? I thought the contract is to only materialize the full parameter during forward/backward of that unit.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/param_group.py
wujingyue: I guess this is for the latter; the other "raw", full-shape parameter is probably constructed elsewhere I've yet to find.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/param_group.py
```diff
@@ -0,0 +1,329 @@
...
+    - dist_grads: DTensor gradient views
+    """
+
+    def __init__(
+        self,
+        params: List[torch.nn.Parameter],
+        param_group_id: ParamGroupIdx,
+        *,
+        mp_policy: FullyShardMixedPrecisionPolicy,
+        mesh: Optional[DeviceMesh] = None,
```
wujingyue, line 60:
```suggestion
        mesh: DeviceMesh,
```
Given it's an internal data structure, I don't see a really good reason to complicate the support surface.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,498 @@
...
+from contextlib import nullcontext
+from dataclasses import dataclass, field
+from importlib.metadata import version
+from typing import List, Optional, Tuple
+
+import torch
+from packaging.version import Version as PkgVersion
+
+try:
+    import transformer_engine as te
```
wujingyue, line 19: Do we still care about the non-TE path? It feels like the performance isn't going to be competitive at all without TE.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,498 @@
...
+    """Mixed precision policy owned by the v2 ``fully_shard`` path."""
+
+    main_params_dtype: Optional[torch.dtype] = None
+    main_grads_dtype: Optional[torch.dtype] = None
+    grad_comm_dtype: Optional[torch.dtype] = None
+    fp8_param_gather: bool = False
+    fp8_recipe: Optional[str] = None
+    keep_fp8_transpose_cache: bool = False
+    use_decoupled_grad: bool = False
+    fp8: Optional[FullyShardFP8Policy] = field(default=None, repr=False)
```
wujingyue, line 146: Request for comments. Also can't they all be merged into FullyShardFp8Policy?

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,498 @@
...
+            args["preserve_high_precision_init_val"] = True
+        return QUANTIZED_MODEL_INIT_CLASS(**args)
+
+    def group_key_dtype(self, tensor: torch.Tensor):
+        """Return the parameter grouping dtype key."""
+        if not self.is_fp8_param(tensor):
+            return tensor.dtype
+        return ("quantized", type(tensor).__name__, self.fp8.recipe)
+
+    def validate_param_group(self, params: List[torch.Tensor]) -> None:
```
wujingyue, line 185: I think many methods of this class (like this one) should really be moved to a different class or a helper method. E.g. ParameterGroup.validate(). This makes MixedPrecisionPolicy a leaf-level class with minimal dependency, following https://en.wikipedia.org/wiki/Single-responsibility_principle.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,498 @@
...
+        return [model_weight_buffer]
+
+    def needs_transpose_weight_buffer(self, tensor: torch.Tensor) -> bool:
+        """Return whether ``tensor`` needs an extra transpose/columnwise buffer."""
+        return HAVE_TE_MXFP8 and isinstance(tensor, MXFP8Tensor)
+
+    def main_params_dtype_for_param(self, tensor: torch.Tensor) -> Optional[torch.dtype]:
+        """Return the main-parameter dtype for a parameter group."""
+        if self.is_fp8_param(tensor) and self.main_params_dtype is None:
+            return torch.float32
```
wujingyue, line 282: Why? If a parameter is fp8, main_weight is fp32 regardless of `main_params_dtype`?

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,498 @@
...
+    enabled: bool = False
+    recipe: Optional[str] = None
+    keep_transpose_cache: bool = False
+
+
+@dataclass(frozen=True)
+class FullyShardMixedPrecisionPolicy:
+    """Mixed precision policy owned by the v2 ``fully_shard`` path."""
+
+    main_params_dtype: Optional[torch.dtype] = None
```
wujingyue, line 139: Naming is better to be consistent--we have main_params and main_weight_buffer.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def storage_tensors_to_free(
+        self,
+        tensor: torch.Tensor,
+        model_weight_buffer,
+        main_weight_buffer,
+    ) -> List[torch.Tensor]:
+        """Return original parameter storages that FSDP buffers have replaced."""
+        # The buffers are ownership signals, not data sources: non-FP8 params can
+        # be backed by either model or main weight storage, while FP8 params are
```
wujingyue, line 247: > non-FP8 params can be backed by either model or main weight storage

Really? I thought params are either sharded params (backed by main_weight and main_grad) or unsharded params (sort of model_weight but allgathered).

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,498 @@
...
+        data_parallel_group: torch.distributed.ProcessGroup,
+        model_weight_buffer,
+        main_weight_buffer,
+        transpose_weight_buffer=None,
+    ) -> None:
+        """Install optimized main weights into model compute weights."""
+        if main_weight_buffer is None:
+            return
+
+        assert model_weight_buffer is not None, "main weights require a model-weight buffer"
```
wujingyue, line 370: Side question: for high precision like bf16, main_weight_buffer == model_weight_buffer?

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,498 @@
...
+def quantize_main_weights_to_fp8(
+    model_params: List[torch.Tensor],
+    main_params: List[Optional[torch.Tensor]],
+    start_offsets: List[Optional[int]],
+    data_parallel_group: torch.distributed.ProcessGroup,
+    fsdp_shard_model_params: List[Tuple[Optional[torch.Tensor], Optional[torch.Tensor]]],
+) -> None:
+    """Quantize FP32 main-weight shards into FP8/MXFP8 model-weight shards."""
+    if len(model_params) == 0:
+        return
```
wujingyue, line 434:
```suggestion
```

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,498 @@
...
+
+    fsdp_shard_model_params = [x[0] if x[1] is None else x for x in fsdp_shard_model_params]
+    if HAVE_TE_CAST_MASTER_WEIGHTS_TO_FP8:
+        args = [
+            model_params,
+            main_params,
+            start_offsets,
+            data_parallel_group,
+            fsdp_shard_model_params,
+        ]
```
wujingyue, line 444:
```suggestion
```
I'd just inline them to the call site.

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
```diff
@@ -0,0 +1,498 @@
...
+    attr = "_rowwise_data" if hasattr(tensor, "_rowwise_data") else "_data"
+    return getattr(tensor, attr)
+
+
+def quantize_main_weights_to_fp8(
+    model_params: List[torch.Tensor],
+    main_params: List[Optional[torch.Tensor]],
+    start_offsets: List[Optional[int]],
+    data_parallel_group: torch.distributed.ProcessGroup,
+    fsdp_shard_model_params: List[Tuple[Optional[torch.Tensor], Optional[torch.Tensor]]],
```
wujingyue, line 430: What's this in addition to `model_params` and `main_params`?

megatron/core/distributed/fsdp/src/megatron_fsdp/v2/mixed_precision.py
shjwudp: I suspect this refers to the backed model weight buffer in its distributed form (distributed state) and main weight shard for high-precision version. I think the description here could be more precise. FYI @Autumn1998

PR #3553, opened 2026-02-24
Port DeepSeek Sparse Attention to `MambaModel`

janEbert:
# What does this PR do ?

Make experimental DeepSeek Sparse Attention (DSA) available to `MambaModel`. For now, we error out when a user tries to use both standard Attention and DSA. It's not supported right now to have different Attention flavors in the same hybrid model, this requires broader restructuring.

### Pre-checks

- [ ] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [X] I have added relevant unit tests
- [X] I have added relevant functional tests
- [X] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [X] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

The following process is enforced via the CODEOWNERS file for changes into `megatron/core`. For changes outside of `megatron/core`, it is up to the PR author whether or not to tag the Final Reviewer team.

<details>
<summary>For MRs into `main` branch</summary>

Feel free to message or comment the @mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

#### (Step 2): Collect the expert reviewers reviews

1. Attach the `Expert Review` label when your PR is ready for review.
2. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

:warning: Only proceed to the next step once all reviewers have approved, merge-conflict are resolved and the CI is passing.  
Final Review might get declined if these requirements are not fulfilled.

#### (Step 3): Final Review

1. Add `Final Review` label
2. GitHub auto-assigns final reviewers based on your changes. They will get notified and pick up your PR soon.

#### (Optional Step 4): Cherry-pick into release branch

If this PR also needs to be merged into `core_r*` release branches, after this PR has been merged, select `Cherry-pick` to open a new PR into the release branch.

</details>

<details>
<summary>For MRs into `dev` branch</summary>
The proposed review process for `dev` branch is under active discussion.

MRs are mergable after one approval by either `eharper@nvidia.com` or `zijiey@nvidia.com`.
</details>

### Merging your PR

Any member of [core-adlr](https://github.com/orgs/teams/NVIDIA/core-adlr) and [`core-nemo`](https://github.com/orgs/teams/NVIDIA/core-nemo) will be able to merge your PR.

PR #4452, opened 2026-04-23
Port Multi-Latent Attention to `HybridModel`

janEbert:
Add Multi-Latent Attention (MLA) support to `HybridModel` (similar to DSA support in https://github.com/NVIDIA/Megatron-LM/pull/3553).

Also includes a minor rename for Mamba→Hybrid that wasn't caught in the earlier large PRs.

Note that the QK norm spec resolution functionality also catches errors that the existing implementation doesn't catch when the spec is misconfigured, therefore we need to make some additional changes.

### DIFF
--- a/megatron/core/models/backends.py
+++ b/megatron/core/models/backends.py
@@ -204,15 +204,15 @@ class InferenceSpecProvider(BackendSpecProvider):
 
 def get_backend(
     transformer_impl: Literal["local", "transformer_engine", "inference_optimized"]
-) -> type[BackendSpecProvider]:
+) -> BackendSpecProvider:
     """Return the backend that's selected with the given `transformer_impl`."""
     if transformer_impl == "transformer_engine":
         from megatron.core.extensions.transformer_engine_spec_provider import TESpecProvider
 
-        return TESpecProvider
+        return TESpecProvider()
     elif transformer_impl == "inference_optimized":
-        return InferenceSpecProvider
+        return InferenceSpecProvider()
     elif transformer_impl == "local":
-        return LocalSpecProvider
+        return LocalSpecProvider()
     else:
         raise ValueError(f"unknown transformer_impl='{transformer_impl}'")
--- a/megatron/core/models/backends.py
+++ b/megatron/core/models/backends.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import warnings
 from abc import abstractmethod
 from functools import partial
-from typing import Optional, Protocol, cast
+from typing import Literal, Optional, Protocol, cast
 
 from megatron.core.extensions.transformer_engine import (
     TEColumnParallelGroupedLinear,
@@ -200,3 +200,19 @@ class InferenceSpecProvider(BackendSpecProvider):
                 activation_func=self.activation_func(),
             ),
         )
+
+
+def get_backend(
+    transformer_impl: Literal["local", "transformer_engine", "inference_optimized"]
+) -> type[BackendSpecProvider]:
+    """Return the backend that's selected with the given `transformer_impl`."""
+    if transformer_impl == "transformer_engine":
+        from megatron.core.extensions.transformer_engine_spec_provider import TESpecProvider
+
+        return TESpecProvider
+    elif transformer_impl == "inference_optimized":
+        return InferenceSpecProvider
+    elif transformer_impl == "local":
+        return LocalSpecProvider
+    else:
+        raise ValueError(f"unknown transformer_impl='{transformer_impl}'")
--- a/megatron/core/models/hybrid/hybrid_block.py
+++ b/megatron/core/models/hybrid/hybrid_block.py
@@ -5,6 +5,7 @@
 # This source code is licensed under the Apache license found in the
 # LICENSE file in the root directory of this source tree.
 
+import copy
 from contextlib import nullcontext
 from dataclasses import dataclass
 from typing import Optional, Tuple, Union
@@ -115,6 +116,8 @@ class HybridStack(MegatronModule):
         )
         self.layer_type_list = layer_type_list
 
+        submodules = self._maybe_fuse_mla_down_proj(submodules)
+
         # Build layers from the pre-selected segment
         self.layers = nn.ModuleList()
         for i, layer_type in enumerate(self.layer_type_list):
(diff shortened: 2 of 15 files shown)


Phlip79 (changes requested):
This MLA spec ignores `--qk-layernorm` and `--qk-l2-norm` and doesn't expose `--mla-down-proj-fusion` like GPTModel does.

janEbert:
> This MLA spec ignores `--qk-layernorm` and `--qk-l2-norm` and doesn't expose `--mla-down-proj-fusion` like GPTModel does.

Addressed; the logic for QK layernorm is quite complex, so this blew the PR up. I could put this into another PR to make this one easier to review.

janEbert:
Note that the `_validate_qk_norm_spec` method is just a safety measure that catches errors that the existing implementation doesn't catch when the spec is misconfigured. It's actually not directly related to the changes in this PR, although I do make use of the assumption that the validation provides.

tests/unit_tests/models/test_hybrid_model.py
```diff
@@ -325,13 +327,15 @@ def _get_attention_layer(self, model):
...
+        """Without qk_layernorm, attention has trivial q/k layernorm."""
+        from megatron.core.transformer.identity_op import IdentityOp
+
         model = self._build_model()
         attn = self._get_attention_layer(model)
         assert attn is not None
-        assert attn.q_layernorm is None
-        assert attn.k_layernorm is None
+        assert attn.q_layernorm is None or isinstance(attn.q_layernorm, IdentityOp)
+        assert attn.k_layernorm is None or isinstance(attn.q_layernorm, IdentityOp)
```
Phlip79:
```suggestion
        assert attn.k_layernorm is None or isinstance(attn.k_layernorm, IdentityOp)
```

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -186,6 +200,29 @@ def __init__(
...
             )
 
+    def _maybe_fuse_mla_down_proj(self, submodules: HybridStackSubmodules) -> HybridStackSubmodules:
+        if getattr(self.config, "mla_down_proj_fusion", False):
+            submodules = copy.deepcopy(submodules)
+            mla_spec = submodules.mla_layer
+            # We always fuse the input layernorm because Hybrid always uses TransformerEngine.
+
+            from megatron.core.extensions.transformer_engine import TELayerNormColumnParallelLinear
+            from megatron.core.transformer.multi_latent_attention import FusedMLASelfAttention
```
Phlip79: Why not import these along with `TENorm` above?

megatron/core/transformer/multi_latent_attention.py
```diff
@@ -598,18 +608,198 @@ def __init__(
...
+                        )
+                else:
+                    linear_q_proj_cls = submodules.linear_q_proj or TEColumnParallelLinear
+                    if linear_q_proj_cls is None:
+                        raise ValueError(
+                            "qk_layernorm requires TransformerEngine or "
+                            "q_layernorm/kv_layernorm to be set in the spec."
+                        )
+                    elif linear_q_proj_cls is not TEColumnParallelLinear:
+                        raise RuntimeError(
```
Phlip79: Inconsistent use of `ValueError` and `RuntimeError`

Phlip79 (approved):
Great tests!

tests/unit_tests/models/test_hybrid_model.py
janEbert: Great catch! Fixed.

megatron/core/transformer/multi_latent_attention.py
janEbert: Fixed; `RuntimeErrors` are only raised when `TransformerEngine` is missing.

megatron/core/models/hybrid/hybrid_block.py
janEbert: Since it's an optional feature, I felt it was good to also keep the imports optional. Moved them to the top, now.

janEbert:
Too much breakage from adding support for extra features, converted back to draft for now.

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -198,6 +213,27 @@ def __init__(
                 eps=self.config.layernorm_epsilon,
             )
 
+    def _maybe_fuse_mla_down_proj(self, submodules: HybridStackSubmodules) -> HybridStackSubmodules:
+        if getattr(self.config, "mla_down_proj_fusion", False):
+            # Do not deepcopy the original `submodules`, so its unrelated contents, such as
+            # `partial` functions, stay identical.
+            mla_spec = copy.deepcopy(submodules.mla_layer)
+            submodules = replace(submodules, mla_layer=mla_spec)
```
santhnm2: What is the purpose of deepcopying and then immediately assigning? Why not just directly assign to `submodules.mla_layer`?

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -198,6 +213,27 @@ def __init__(
                 eps=self.config.layernorm_epsilon,
             )
 
+    def _maybe_fuse_mla_down_proj(self, submodules: HybridStackSubmodules) -> HybridStackSubmodules:
+        if getattr(self.config, "mla_down_proj_fusion", False):
```
santhnm2: Nit: if this function is a no-op without `mla_down_proj_fusion` being set, should we just rename to `_fuse_mla_down_proj` and move the condition check to the constructor?

megatron/core/transformer/multi_latent_attention.py
```diff
@@ -613,18 +621,244 @@ def __init__(
...
+            and is_fused_norm_linear(submodules.linear_kv_up_proj)
+        ):
+            raise ValueError(
+                f"`kv_layernorm={submodules.kv_layernorm}` is non-trivial "
+                f"and `linear_kv_up_proj={submodules.linear_kv_up_proj}` is a "
+                f"fused norm+linear; either unset `kv_layernorm` or use a "
+                f"linear layer without norm fusion for `linear_kv_up_proj`"
+            )
+
+    def _resolve_qk_norm_config(
```
santhnm2, line 631: This function is very difficult to follow with all the nested if conditions. Can we take a step back here and enumerate the constraints at a high level first and then figure out the simplest way to enforce these constraints?

megatron/core/models/hybrid/hybrid_block.py
janEbert: Since we're modifying `mla_layer` specifically, it seemed cleaner to avoid mutating it. However, I agree that, since we're mutating `submodules` anyway, we could just forego the deepcopy.

Previously, this was:
```python
submodules = copy.deepcopy(submodules)
mla_spec = submodules.mla_layer
```

, which resulted in equivalency tests for the specs not passing, since `partial` functions don't compare equally when they are compared to copies of themselves.

What do you prefer? Remove copies and mutate or add copies and keep the original objects clean? Either way I agree the current solution is awkward.

megatron/core/models/hybrid/hybrid_block.py
janEbert: Sounds good.

megatron/core/models/hybrid/hybrid_block.py
janEbert: I went back to the old full deepcopy route.

megatron/core/transformer/multi_latent_attention.py
janEbert: Refactored it all to a new module with individual functions for the various checks.

megatron/core/models/backends.py
```diff
@@ -200,3 +200,19 @@ def grouped_mlp_modules(self, moe_use_grouped_gemm: bool) -> ExpertsBuilder:
                 activation_func=self.activation_func(),
             ),
         )
+
+
+def get_backend(
```
santhnm2, line 205: Where is this function used? Is this replacing existing functionality?

janEbert: [It's used in the QK norm config resolution](https://github.com/NVIDIA/Megatron-LM/pull/4452/files#diff-ea57a41780482f09515f303ae6e9b390ce0143177cf57ca4df807a23efc265c0R43). It doesn't replace existing functionality, I just needed a way to get the currently used backend to properly replace the existing layers. I was also surprised to find out that such functionality doesn't exist.

Phlip79 (approved):

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -198,6 +214,24 @@ def __init__(
                 eps=self.config.layernorm_epsilon,
             )
 
+    def _fuse_mla_down_proj(self, submodules: HybridStackSubmodules) -> HybridStackSubmodules:
+        submodules = copy.deepcopy(submodules)
```
santhnm2, line 224: Why is this deepcopy necessary?

santhnm2 (approved):

megatron/core/models/hybrid/hybrid_block.py
janEbert: Added small comment as discussed offline.
```
        # Avoid modifying the original object so users don't get surprised about their `submodules`
        # being modified underneath them.
```

jiemingz (approved):

PR #4463, opened 2026-04-24
Add Hybrid Transformer block fusion

janEbert:
Add fusion operator `[...]` to `--hybrid-layer-pattern`, enabling packing subsequent sequence mixer + channel mixer operations into one `TransformerLayer`. This way, we have better expectations of the operations and can leverage existing optimizations in `TransformerLayer`.

For checkpointing, we save the model as if it was unfused, so that we have a canonical format and backward-compatibility. The model can be loaded in fused form from the unfused checkpoint. The transformations to the state dict are only applied when saving or loading.

An important part of this PR is also extracting the "low-level" blocks from the hybrid spec, which were previously placed as parts of `TransformerLayer`'s submodules, to make them available as stand-alone specs for use in both combined/packed/"fused" and standalone (the previous/standard solution) `TransformerLayer`s.

Yields **3.79% throughput improvement** on a 20-layer 1.3B non-MoE model with no gradient accumulation and PP=1;
**1.92% througput improvement** on a 20-layer 8.4B MoE model with gradient accumulation of 4 and PP=2.
Peak memory use remains extremely similar across the fused/unfused setting, with a minimal improvement when using the fused version.

### DIFF
--- a/megatron/core/models/hybrid/hybrid_block.py
+++ b/megatron/core/models/hybrid/hybrid_block.py
@@ -442,7 +442,7 @@ class HybridStack(MegatronModule):
             )
 
             sharded_prefix = f'{layer_prefix}{global_layer_offset}.'
-            sharded_pp_offset = []
+            sharded_pp_offset: list[tuple[int, int, int]] = []
 
             layer_sharded_state_dict = layer.sharded_state_dict(
                 state_dict_prefix, sharded_pp_offset, metadata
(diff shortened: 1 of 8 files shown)


Phlip79 (commented):
- Should we reject MTP usage within brackets?
- Can you add a working forward/backward test? `TestFusedLayerValidation` just has failure path coverage.

janEbert:
Great points! Addressed both.

Phlip79 (approved):

Phlip79 (approved):

janEbert:
Closed in favor of https://github.com/NVIDIA/Megatron-LM/pull/4942.

PR #4467, opened 2026-04-24
[codex] Fix Mamba conv params under fine-grained FSDP gather

ilml:
## Summary

Fix fine-grained Megatron-FSDP parameter gathering for Mamba's fused conv path.

This branch is now stacked with the MambaLayer FSDP support from #4329 first, then the direct conv-param gather fix. #4329 makes `MambaLayer` an FSDP unit and fixes TP annotation for SSM parameters; this PR handles the remaining case where `MambaMixer.forward()` reads a child module's parameters without invoking that child module's forward hook.

Mamba's memory-efficient fused path reads `conv1d.weight` and `conv1d.bias` directly and passes them into `mamba_split_conv1d_scan_combined`, instead of calling `Conv1d.forward()`. With fine-grained Megatron-FSDP gather enabled, that means the child `conv1d` module's pre-forward gather hook never runs. After the first forward releases parameter storage, the second forward can pass a null-base sharded view into causal-conv, producing an illegal memory access.

This PR adds an opt-in `_fsdp_extra_forward_param_modules` hook for modules that directly read child-module params, and uses it from `MambaMixer` for `self.conv1d`. It also makes Mamba context-parallel parameter access resolve through the live mixer object so FSDP-updated parameters are not bypassed by stale cached references.

## Validation

- `python3 -m py_compile megatron/core/distributed/fsdp/mcore_fsdp_adapter.py megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py megatron/core/distributed/torch_fully_sharded_data_parallel.py megatron/core/ssm/mamba_context_parallel.py megatron/core/ssm/mamba_mixer.py tests/unit_tests/distributed/megatron_fsdp/test_mcore_tensor_parallelism_detect.py`
- `python3 -m pytest tests/unit_tests/distributed/megatron_fsdp/test_mcore_tensor_parallelism_detect.py -q` could not run on the login node because `pytest` is not installed there.
- Before stacking #4329, a clean interactive run of `/home/tolong/work/dsv3/interactive_nt.sh` completed through iteration 50 and saved checkpoint with no `CUDA error`, `illegal memory`, or `FAILED` in the log:
  `/lustre/fsw/coreai_dlalgo_llm/tolong/results/nemo_megatron/megatron/nemotron6/hybrid/debug/interactive_nt_debug/logs/interactive_nt_full_20260424_150942.log`
- Earlier two-iteration blocking-CUDA smoke test completed and saved checkpoint at iteration 2 after reproducing the pre-fix crash at that boundary.

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
@@ -759,6 +759,17 @@ class MegatronFSDP(torch.nn.Module):
             if self.enable_fine_grained_param_gather_hook:
                 param_list = list(module.parameters(recurse=False))
 
+            extra_forward_param_modules = getattr(module, "_fsdp_extra_forward_param_modules", ())
+            if isinstance(extra_forward_param_modules, nn.Module):
+                extra_forward_param_modules = (extra_forward_param_modules,)
+            if extra_forward_param_modules:
+                seen_param_ids = {id(param) for param in param_list}
+                for extra_module in extra_forward_param_modules:
+                    for extra_param in extra_module.parameters():
+                        if id(extra_param) not in seen_param_ids:
+                            param_list.append(extra_param)
+                            seen_param_ids.add(id(extra_param))
+
             # All-gather the parameters before the forward pass.
             self.all_gather_and_wait_parameters_ready(
                 params=param_list,
--- a/megatron/core/ssm/mamba_context_parallel.py
+++ b/megatron/core/ssm/mamba_context_parallel.py
@@ -70,6 +70,7 @@ class MambaContextParallel:
         A_log_cp1: torch.Tensor,
         D_cp1: torch.Tensor,
         D_has_hdim: bool,
+        mixer=None,
     ) -> None:
         if not HAVE_EINOPS:
             raise ImportError("einops is required by the Mamba model but cannot be imported")
@@ -84,6 +85,7 @@ class MambaContextParallel:
         self.A_log_cp1 = A_log_cp1
         self.D_cp1 = D_cp1
         self.D_has_hdim = D_has_hdim
+        self._mixer = mixer
 
         self.cp_size = self.cp_group.size()
 
@@ -231,24 +233,29 @@ class MambaContextParallel:
     def get_conv1d_weight(self) -> torch.Tensor:
         """Returns a slice of the conv1d weight relevant to the current context parallel rank"""
         # weight shape: [conv_dim, 1, d_conv]
-        return self._slice_conv_param(self.conv1d_cp1.weight)
+        conv1d = self._mixer.conv1d if self._mixer is not None else self.conv1d_cp1
+        return self._slice_conv_param(conv1d.weight)
 
     def get_conv1d_bias(self) -> torch.Tensor:
         """Returns a slice of the conv1d bias relevant to the current context parallel rank"""
         # bias shape: [conv_dim]
-        return self._slice_conv_param(self.conv1d_cp1.bias)
+        conv1d = self._mixer.conv1d if self._mixer is not None else self.conv1d_cp1
+        return self._slice_conv_param(conv1d.bias)
 
     def get_dt_bias(self) -> torch.Tensor:
         """Returns a slice of dt_bias relevant to the current context parallel rank"""
-        return self._slice_vector_param(self.dt_bias_cp1)
+        param = self._mixer.dt_bias if self._mixer is not None else self.dt_bias_cp1
+        return self._slice_vector_param(param)
 
     def get_A_log(self) -> torch.Tensor:
         """Returns a slice of A_log relevant to the current context parallel rank"""
-        return self._slice_vector_param(self.A_log_cp1)
+        param = self._mixer.A_log if self._mixer is not None else self.A_log_cp1
+        return self._slice_vector_param(param)
 
     def get_D(self) -> torch.Tensor:
         """Returns a slice of D relevant to the current context parallel rank"""
-        return self._slice_vector_param(self.D_cp1, has_hdim=self.D_has_hdim)
+        param = self._mixer.D if self._mixer is not None else self.D_cp1
+        return self._slice_vector_param(param, has_hdim=self.D_has_hdim)
 
     def _slice_conv_param(self, param: torch.Tensor) -> torch.Tensor:
         """
--- a/megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
+++ b/megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
@@ -40,6 +40,7 @@ from megatron.core.config_logger import has_config_logger_enabled, log_config_to
 from megatron.core.distributed.data_parallel_base import _BaseDataParallel
 from megatron.core.distributed.distributed_data_parallel_config import DistributedDataParallelConfig
 from megatron.core.process_groups_config import ProcessGroupCollection
+from megatron.core.ssm.mamba_layer import MambaLayer
 from megatron.core.transformer.transformer_config import TransformerConfig
 from megatron.core.transformer.transformer_layer import TransformerLayer
 from megatron.core.utils import is_te_min_version, log_single_rank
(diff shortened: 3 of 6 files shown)


cspades:
CC'd from DM's

Hmm, so the MambaMixer has a Conv1D submodule but the Conv1D submodule's weights are used in Autograd functions during the MambaMixer.forward() pass?

- I assume you are using fine_grained_all_gather since the Conv1D is actually a sub-module of MambaMixer so only in the case where the Conv1D weights would not be AG'd is if we are using fine-grained module AG hooks, in which case we never trigger the Conv1D hook.
- There is also a post-forward / post-backward hook that calls this function: [release_module_parameters](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L551). It needs to be called on the modules in extra_forward_param_modules so we can re-shard them. Try asking Codex what is the cleanest, most generalizable way to have the hooks support re-sharding these untouched modules as well.
- Are you setting fsdp_unit_modules=[TransformerLayer, MambaLayer]?  Either you hit the storage error on Step 1 (the very first AG) but if Mamba is not a FSDP unit, it will never be re-sharded, skipping the post-forward / post-backward hook.
  - @Phlip79 has a PR and is testing this as well: https://github.com/NVIDIA/Megatron-LM/pull/4329

cspades:
Also, if we do implement this, I think Torch FSDP2 has something similar for "hitch-hiking" parameters into the AG: `Tensor.fsdp_pre_all_gather` and `Tensor.fsdp_post_all_gather` (Example in TE: https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/tensor/mxfp8_tensor.py#L613)

That's per-parameter / per-Tensor though, so we're basically doing the same thing here with modules, I suppose.

megatron/core/ssm/mamba_context_parallel.py
```diff
@@ -231,24 +233,29 @@ def conv1d_channels(self):
     def get_conv1d_weight(self) -> torch.Tensor:
         """Returns a slice of the conv1d weight relevant to the current context parallel rank"""
         # weight shape: [conv_dim, 1, d_conv]
-        return self._slice_conv_param(self.conv1d_cp1.weight)
+        conv1d = self._mixer.conv1d if self._mixer is not None else self.conv1d_cp1
+        return self._slice_conv_param(conv1d.weight)
```
cspades, line 237: Have to ask, where/when do we have "stale" references such that we can no longer directly retrieve the weight?

cspades (commented):
Looks pretty good, per-module parameter hitch-hiking seems reasonable for me since our FSDP units are based on modules as well.

@shjwudp @Autumn1998 @wujingyue in the re-write we should expose this as an API, but this PR defines the feature, is this general enough?

wujingyue (commented):
Thanks for the fix! 

IIUC, https://github.com/NVIDIA/Megatron-LM/blob/42e396ef5a7da3ae2f3284184ff6f3e4c0146a17/megatron/core/ssm/mamba_context_parallel.py#L214 is where `def conv1d` calls the `F.conv1d` instead of a submodule. 

Instead, could MambaContextParallel's constructor create the submodule from `conv1d_cp1`? For your reference, https://gitlab-master.nvidia.com/clara-discovery/boltz/-/blob/dev/src/boltz/distributed/model/layers/triangular_attention.py#L1490 is an internal example that adopts this practice for context parallelism. 

cc @cspades

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -759,6 +759,17 @@ def _pre_forward_param_unshard(module: nn.Module, *unused):
...
+            if isinstance(extra_forward_param_modules, nn.Module):
+                extra_forward_param_modules = (extra_forward_param_modules,)
+            if extra_forward_param_modules:
+                seen_param_ids = {id(param) for param in param_list}
+                for extra_module in extra_forward_param_modules:
+                    for extra_param in extra_module.parameters():
+                        if id(extra_param) not in seen_param_ids:
+                            param_list.append(extra_param)
+                            seen_param_ids.add(id(extra_param))
+
```
cspades, line 772: > There is also a post-forward / post-backward hook that calls this function: [release_module_parameters](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L551). It needs to be called on the modules in extra_forward_param_modules so we can re-shard them.

Early note, what about the pre-backward param unshard?

We need to ensure that any rogue weights are re-sharded, and un-sharded during the backward pass. Did you check that the Conv-1D weights are re-sharded?

cspades (changes requested):
Still WIP, will approve when all features are implemented!

cspades:
Screenshot 2026-04-30 at 11 22 59 AM

Need to make sure the module re-shard is working, currently doesn't look like the case.

wujingyue (commented):
Request for clarification: did you run into this issue with enable_fine_grained_param_gather_hook? When MFSDP is applied to MambaLayer, it ought to find all parameters under that layer including the ones in conv1d. It wasn't clear to me how MFSDP missed the parameter in the first place. cc @ilml

cspades:
> Request for clarification: did you run into this issue with enable_fine_grained_param_gather_hook? When MFSDP is applied to MambaLayer, it ought to find all parameters under that layer including the ones in conv1d. It wasn't clear to me how MFSDP missed the parameter in the first place. cc @ilml

```
                enable_fine_grained_param_gather_hook=(
                    config.fp8_recipe == "mxfp8" and ddp_config.fp8_param_gather
                ),
```
It's automatically turned on for MXFP8. (I still forget exactly why.)

ilml:
> Request for clarification: did you run into this issue with enable_fine_grained_param_gather_hook? When MFSDP is applied to MambaLayer, it ought to find all parameters under that layer including the ones in conv1d. It wasn't clear to me how MFSDP missed the parameter in the first place. cc @ilml

yeah this is a very subtle bug:
but basically, its all because we are using third party [kernel](https://github.com/state-spaces/mamba/blob/316ed6036538405f767782132f76caf342256d33/mamba_ssm/ops/triton/ssd_combined.py#L815)  for mamba, and this kernel is NOT inheriting from nn.Module, and it exacts raw weights to passing the kernel for efficiency. and this escape the mfsdp sharding harness. which eventually cause the memory access bug.

wujingyue:
Similar to [what I said earlier](https://github.com/NVIDIA/Megatron-LM/pull/4467#pullrequestreview-4185859800), I think the right fix is to make `self.cp` an `nn.Module` that co-owns the parameters that it needs. 

The reasons are:
1. This is a good practice to follow in general--each parameter is owned by the `Module` that actually uses it. 
2. This avoids the `_fsdp_extra_forward_param_modules` hack completely. Model writers would have to know MFSDP implementation really well to know whether/where to add `_fsdp_extra_forward_param_modules`. It's also extra feature/code to maintain in MFSDP. 
3. This makes the model compatible with the future per-module MFSDP API. We expect `fully_shard` to be called on each `nn.Module` that forms an FSDP unit and to insert hooks to that `nn.Module`. With this PR as is, I couldn't find any `nn.Module` that we can call `fully_shard` on. `self.conv1` would be the wrong module because it's not even executed when CP is on. The `MambaMixer` would be the wrong module because that would put **all** parameters under into one ParameterGroup. 

Wdyt?

cspades:
> Screenshot 2026-04-30 at 11 22 59 AM
> Need to make sure the module re-shard is working, currently doesn't look like the case.

Made a mistake, these are ALL scaling factors. It's quite a lot larger than I expected. (9.3 GB large for this Nemotron model.)

Currently, Megatron-FSDP has this sharding strategy:
```
QuantizedTensorStorage:
    uint8 rowwise_data  # This is controlled by Megatron-FSDP's weight buffer.
    uint8 colwise_data  # This is controlled by Megatron-FSDP's transpose buffer.
    uint8 row_scale_inv  # This is NOT controlled by Megatron-FSDP.
    uint8 col_scale_inv  # This is NOT controlled by Megatron-FSDP.
```
We do deallocate (or reassign for `FixedPoolAllocator`) the storage for the row and column data, but we leave the scaling factors in-place as they already are 16x to 256x smaller than the original parameters, and we can use those Tensors as output buffers for quantization.

Will consider sharding the scaling factors in the future.

wujingyue:
I'm closing this PR for now. 

I'm testing a different, cleaner fix: https://github.com/NVIDIA/Megatron-LM/pull/4717.

PR #4473, opened 2026-04-26
[dev] feat(attention): Add rotary_base_per_layer for Step-3.5-Flash

shifangx:
# What does this PR do ?

feat(attention): Add rotary_base_per_layer for Step-3.5-Flash

- rotary_base_per_layer: Optional[List[float]] -- per-layer RoPE theta values.
  When set, each SelfAttention creates its own RotaryEmbedding
There are quite a few places that assume the GPT model includes RotaryEmbedding by default, so this PR needs to keep RotaryEmbedding within the GPT model. 

The new feature default to False/None and have no effect on existing models.

Main branch: https://github.com/NVIDIA/Megatron-LM/pull/4622

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/core/transformer/attention.py
```diff
@@ -379,6 +379,80 @@ def __init__(
...
+                self.config.hidden_size,
+                self.config.num_attention_heads,
+                config=self.config,
+                init_method=not_none(self.config.init_method),
+                gather_output=False,
+                bias=False,
+                skip_bias_add=False,
+                is_expert=False,
+                tp_comm_buffer_name='gate',
+                tp_group=self.pg_collection.tp,
```
Victarry: **[CRITICAL Architecture]** `g_proj` should be folded into `linear_qkv` and merged with the existing `attention_output_gate` path, rather than introducing a separate ColumnParallelLinear and a new forward branch.

**Reason**:
1. **Code duplication**: `attention_output_gate` (per-head per-channel; gate weights fused into `linear_qkv`, sliced out by `_split_qkv`, applied by `_apply_output_gate`) and `use_head_wise_attn_gate` (per-head scalar; independent `g_proj`; a new sigmoid-multiply block) are functionally near-parallel — they only differ in whether gate granularity is `hn` or `1`. Yet their implementation paths are completely independent. Future changes (GQA AG handling, CUDA graphs, FP8/FP4 compatibility) on either path will need to be replicated and will drift.
2. **TE/local backend divergence**: `g_proj = submodules.linear_qkv(...)` resolves to `TELayerNormColumnParallelLinear` (with a fused LN that has its own learnable parameters) under the TE backend, but plain `ColumnParallelLinear` under the local backend — **the same config produces different mathematics on the two backends**. Folding into `linear_qkv` makes QKV and gate share the same LN, which is consistent across backends and matches the pure-linear `g_proj` of the Step-3.5-Flash paper.
3. **One fewer GEMM kernel launch**: the per-head scalar gate occupies only `num_heads` columns; folding it into `linear_qkv` is essentially free.
4. **`_apply_output_gate` does not need a branch**: replace `gate.view(*x.shape)` with `view(*x.shape[:-1], -1)` (or remove it altogether), and PyTorch broadcasting naturally handles both gate shapes — `[sq,b,np,hn]` (existing) and `[sq,b,np,1]` (new) both multiply `core_attn_out [sq,b,np,hn]` correctly element-wise.

**Suggestion**:
- Branch in the `linear_qkv_out_dim` calculation:
  ```python
  if self.config.attention_output_gate:
      self.linear_qkv_out_dim += self.config.kv_channels * self.config.num_attention_heads  # per-head per-channel
  elif self.config.use_head_wise_attn_gate:
      self.linear_qkv_out_dim += self.config.num_attention_heads                            # per-head scalar
  ```
- In `_split_qkv`, slice the trailing `num_heads` scalars as a new tail segment. Since this is not a multiple of `hn`, slice it off before the group reshape, or implement an unequal-width split inside the group. The GQA AG path needs corresponding updates: after the all-gather, slice QKV by group and slice the scalar segment by q_head separately.
- Change `_apply_output_gate` to rely on broadcasting so both gates share the same function.
- **Delete** the `g_proj` module, `tp_comm_buffer_name='gate'`, and the new per-head gate forward block.
- Since only one gate type can occupy the `linear_qkv` tail, **the two flags become naturally mutually exclusive** (you may add an assert in `__post_init__` to make this explicit).

megatron/core/transformer/attention.py
```diff
@@ -975,6 +1049,11 @@ def forward(
         if no_rope:
             rotary_pos_emb = None
 
+        # Per-layer theta: override the model-level RoPE with this layer's own embedding.
+        if self.rotary_pos_emb is not None and rotary_pos_emb is not None:
+            seq_len = rotary_pos_emb.shape[0]
+            rotary_pos_emb = self.rotary_pos_emb(seq_len)
```
Victarry, line 1097: **[IMPORTANT Correctness]** The per-layer rotary override **depends on the model-level `rotary_pos_emb` already existing**, but the PR description states "shared model-level rotary_pos_emb in GPTModel is not created" — and this PR **does not modify `GPTModel`**.

**Reason**:
1. **Description does not match implementation**: the model-level rotary is still created by `GPTModel`, and each layer just overrides it. Both rotaries coexist; the model-level one is computed every forward and immediately discarded — wasted compute on every iteration.
2. **Silent skip**: if `GPTModel` ever stops passing `rotary_pos_emb` on some path (cross-attention, fully no-rope layers, user-disabled), the per-layer rotary will be silently skipped — there's no `assert`, and the behavior becomes unpredictable.
3. **`seq_len = rotary_pos_emb.shape[0]`** implicitly couples to the external rotary tensor shape. That variable is wrapped into `(rotary_pos_emb,) * 2` later in the forward; the override happens before the tuple wrap so it's currently correct, but the coupling is fragile.

**Suggestion**:
- Modify `GPTModel` in tandem: when `config.rotary_base_per_layer is not None`, skip creating the model-level `rotary_pos_emb` and just pass `seq_len` to each layer so each layer generates its own rotary.
- Or keep the override-style implementation but obtain `seq_len` from `inference_context` / `hidden_states.shape[0]` instead of relying on the external rotary shape.
- Add an assert: when per-layer rotary is enabled and `rotary_pos_emb is None`, raise explicitly instead of silently skipping.

megatron/core/transformer/attention.py
```diff
@@ -379,6 +379,80 @@ def __init__(
...
+                rotary_interleaved=self.config.rotary_interleaved,
+                seq_len_interpolation_factor=seq_len_interpolation_factor,
+                rotary_base=rotary_base,
+            )
+            self.mrope_section = self.config.mrope_section
+            assert (
+                self.mrope_section is not None
+            ), "mrope require mrope_section setting, but we got None from TransformerConfig"
+        else:
+            assert False, "Invalid position embedding type"
```
Victarry: **[IMPORTANT Maintenance]** `_build_per_layer_rotary_pos_emb` duplicates the `RotaryEmbedding` / `YarnRotaryEmbedding` / `MultimodalRotaryEmbedding` construction logic from `gpt_model.py`.

**Reason**: When `gpt_model.py` adds new rotary parameters (new `rope_scaling` parameters, new rotary types), this per-layer path will silently fall out of sync, causing per-layer rotary behavior to diverge from model-level rotary.

**Suggestion**: Extract a factory function (e.g. in `models/common/embeddings/rotary_pos_embedding.py`) with a signature like:
```python
def build_rotary_pos_emb(
    config,
    rotary_base: Optional[float] = None,
    cp_group=None,
) -> RotaryEmbedding | YarnRotaryEmbedding | MultimodalRotaryEmbedding: ...
```
called from both `GPTModel` and here, so the two sites stay in lockstep.

tests/unit_tests/transformer/test_head_wise_attn_gate.py
```diff
@@ -0,0 +1,212 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
Victarry: **[IMPORTANT Tests]** Tests cover only `use_head_wise_attn_gate`; there are no tests for `rotary_base_per_layer`.

**Reason**: `rotary_base_per_layer` involves two independent paths (per-layer rotary construction, forward-time override of the model-level rotary), each with three branches (`rope` / `yarn` / `mrope`). Zero test coverage means any future refactor (including adopting the factory-function suggestion) has no safety net.

**Suggestion**: Add at minimum:
- A test that triggers the `__post_init__` assert when `len(rotary_base_per_layer) != num_layers`.
- A multi-layer model with distinct `rotary_base` per layer; verify each layer's `self.rotary_pos_emb` is non-`None` and that `inv_freq` matches the corresponding `rotary_base`.
- A forward-pass test where each layer's rotary output numerically matches a standalone `RotaryEmbedding` constructed with the same `rotary_base`.
- A test that with `rotary_base_per_layer=None`, `self.rotary_pos_emb is None` and forward runs without error.

megatron/core/transformer/attention.py
Victarry: **[SUGGESTION Code]** `assert False, "Invalid position embedding type"` does not include the actual `position_embedding_type` value, and `assert False` is semantically off — this is not an invariant violation, it's an unsupported configuration.

**Suggestion**:
```python
else:
    raise NotImplementedError(
        f"rotary_base_per_layer does not support "
        f"position_embedding_type={self.config.position_embedding_type!r} "
        f"(only 'rope' / 'yarn' / 'mrope' are supported)."
    )
```

megatron/core/transformer/attention.py
```diff
@@ -379,6 +379,80 @@ def __init__(
             # the quantized tensor.
             set_save_original_input(self.linear_proj)
 
+        # Per-layer RotaryEmbedding (used when rotary_base_per_layer is set in config).
+        self.rotary_pos_emb = None
+        if getattr(self.config, 'rotary_base_per_layer', None):
```
Victarry, line 394: **[SUGGESTION Code]** `getattr(self.config, 'rotary_base_per_layer', None)` uses `getattr`, which suggests the field might not exist — but the field is added by this PR to `TransformerConfig`, so it **always exists**.

**Suggestion**:
```python
if self.config.rotary_base_per_layer is not None:
```
This avoids unnecessary cognitive overhead and prevents readers from wondering whether some other config type lacks this field.

Victarry (changes requested):
PR adds two opt-in features (`use_head_wise_attn_gate` + `rotary_base_per_layer`) for Step-3.5-Flash; defaults are off, existing models unaffected.

**Main suggestions (see inline comments for details)**:

1. **[CRITICAL]** Fold `use_head_wise_attn_gate` into `linear_qkv` and merge it with the existing `attention_output_gate` path. Share `_split_qkv` / `_apply_output_gate` (broadcasting handles both gate shapes — no new branch needed). Benefits:
   - Eliminates the TE/local backend divergence in `g_proj` math (`submodules.linear_qkv` resolves to `TELayerNormColumnParallelLinear` under TE, which adds an independent learnable LayerNorm).
   - Avoids two near-parallel "output gating" implementations coexisting long-term and drifting.
   - Saves one GEMM kernel launch.
   - Makes the two gate flags naturally mutually exclusive.

2. **[IMPORTANT]** The `rotary_base_per_layer` forward override depends on the model-level `rotary_pos_emb` already existing, contradicting the PR description's claim that "model-level rotary_pos_emb is not created" — this PR does not modify `GPTModel`, so both rotaries actually coexist.

3. **[IMPORTANT]** `_build_per_layer_rotary_pos_emb` duplicates the rotary-construction logic from `gpt_model.py`; recommend extracting a shared factory function.

4. **[IMPORTANT]** No test coverage at all for `rotary_base_per_layer`.

5/6. **[SUGGESTION]** `assert False, "Invalid position embedding type"` is uninformative; `getattr(self.config, 'rotary_base_per_layer', None)` is unnecessary (the field is added by this PR).

megatron/core/transformer/attention.py
shifangx: Have folded use_head_wise_attn_gate into linear_qkv and merged it with the existing attention_output_gate path.

megatron/core/transformer/attention.py
shifangx: There are quite a few places that assume the GPT model includes RoPE by default, so this PR needs to keep RoPE within the GPT model. The description in the PR README has also been updated.

megatron/core/transformer/attention.py
shifangx: Prefer to keep the code in the GPT model unchanged.
The per-layer RoPE initialization parameters in the attention module are not exactly the same as those in the GPT model.
To minimize the impact on the existing GPT model code, the implementations in these two places are kept separate for now.

megatron/core/transformer/attention.py
shifangx: done

tests/unit_tests/transformer/test_head_wise_attn_gate.py
shifangx: added test for rotary_base_per_layer

megatron/core/transformer/attention.py
shifangx: done

shifangx:
/ok to test [7485da4](https://github.com/NVIDIA/Megatron-LM/pull/4473/commits/7485da49c716213c6acf11af777fbba5348b5777)

shifangx:
/ok to test [c75616a](https://github.com/NVIDIA/Megatron-LM/pull/4473/commits/c75616af11281935a78aab296d0943b9429eaaeb)

shifangx:
/ok to test [5671195](https://github.com/NVIDIA/Megatron-LM/pull/4473/commits/5671195236a6df980d8220af86d61d421646b5f9)

shifangx:
/ok to test [bea28f6](https://github.com/NVIDIA/Megatron-LM/pull/4473/commits/bea28f6d0d08e046cf277a1e73221fa5109796d6)

Victarry:
Please also submit a PR to the main branch, Thanks!

Victarry (approved):

shifangx:
/ok to test [027d8d2](https://github.com/NVIDIA/Megatron-LM/pull/4473/commits/027d8d2cffaebabefb5d2be4b5aa96415ff2b872)

shifangx:
hi, @NVIDIA/core-nemo, could you help to review this pr?

PR #4475, opened 2026-04-26
Muon CPU offload

pengdurice:
# What does this PR do ?
Add a simple CPU offload for Muon optimizer
This is a simple CPU offloading implementation such that the  fp32 master weights and optimizer states are moved to CPU after the optimizer's step is called and they are reloaded before the next step() call. This doesn't reduce the GPU memory requirement  during the optimizer step but reduces the peak memory usages during the model's forward / backward calls. 

## Tests
This is tested with GLM5 training and unblocked the OOM issue without this change.
Multiple unit tests are added.

## Issue tracking
https://github.com/NVIDIA/Megatron-LM/issues/4691

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


pengdurice:
@janEbert , Hi, nice to meet you! would you mind review this PR to enable Muon CPU offload? thank you!

megatron/core/optimizer/layer_wise_optimizer.py
```diff
@@ -744,7 +771,13 @@ def step_with_ready_grads(self) -> bool:
         runs when this optimizer is a child of an outer ``ChainedOptimizer``,
         which calls ``step_with_ready_grads`` directly on each child and bypasses
         ``step``.
+
+        When CPU offloading is enabled, orchestrates the full cycle:
+        reload states to GPU -> optimizer step -> param sync -> offload states to CPU.
         """
+        if self._cpu_offload:
+            self.reload_optimizer_states()
```
janEbert, line 793: See 2. in the main review comment.

megatron/core/optimizer/__init__.py
```diff
@@ -894,6 +891,12 @@ def _get_megatron_emerging_optimizer(
                         "the legacy LayerWise ping-pong path for MoE models."
                     )
                 fallback_config.use_distributed_optimizer = True
+                # Disable per-optimizer CPU offload (HybridDeviceOptimizer) for the
+                # Adam fallback when LayerWiseDistributedOptimizer is active.
+                # CPU offloading is handled uniformly by LayerWiseDistributedOptimizer
+                # for all sub-optimizers (Muon + Adam), preventing double-offloading.
+                if use_layer_wise:
+                    fallback_config.optimizer_cpu_offload = False
```
janEbert: See 3. in the main review comment.

megatron/core/optimizer/__init__.py
```diff
@@ -10,6 +10,41 @@
 from torch.optim import SGD as CPUSGD
```
janEbert, line 10: Don't have a good place for this, but here is a placeholder comment so you can resolve 1. from the main review comment. :)

janEbert (changes requested):
Hey, thanks for the implementation and additional feature! I couldn't get the code to run immediately; have you actually tested this?

1. `--optimizer-cpu-offload` requires `--use-precision-aware-optimizer` (`megatron/training/arguments.py:1665`).
   However, `--use-precision-aware-optimizer` requires Adam use (`megatron/core/optimizer/optimizer_config.py:419`).

   So the code currently can't run with Muon. Maybe allow `--optimizer-cpu-offload` without `--use-precision-aware-optimizer` when also using `--use-layer-wise-distributed-optimizer`.
2. It's too late to call `LayerWiseDistributedOptimizer.reload_optimizer_states` in `LayerWiseDistributedOptimizer.step_with_ready_grads`:
   `prepare_grads` would already have been called by `ChainedOptimizer.step` before `step_with_ready_grads` is executed. And in `MixedPrecisionOptimizer.prepare_grads`, `_copy_model_grads_to_main_grads` tries to copy CUDA `main_param.main_grad` tensors to the CPU `main_param.grad` tensors, resulting in a device mismatch `RuntimeError`.

   You'd need to call `reload_optimizer_states` in a new `LayerWiseDistributedOptimizer.prepare_grads` override method before executing the `super` method. Note that you may still want to leave the original location where `reload_optimizer_states` is called in, in case `step_with_ready_grads` is called directly (without `prepare_grads`).
3. Setting `fallback_config.optimizer_cpu_offload = False` in `megatron/core/optimizer/__init__.py:899` prevents Adam parameters from being offloaded in this case, because a separate `HybridDeviceOptimizer` is created, which is _not_ managed by the `LayerWiseDistributedOptimizer` (refer to `megatron/core/optimizer/__init__.py:965`).

   So this should be removed. Checking correct offloading of Adam-managed parameters in both cases (`use_separate_distributed_optimizer=True` and `use_separate_distributed_optimizer=False`) to the tests would also make a lot of sense!

pengdurice:
> Hey, thanks for the implementation and additional feature! I couldn't get the code to run immediately; have you actually tested this?
> 
> 1. `--optimizer-cpu-offload` requires `--use-precision-aware-optimizer` (`megatron/training/arguments.py:1665`).
>    However, `--use-precision-aware-optimizer` requires Adam use (`megatron/core/optimizer/optimizer_config.py:419`).
>    So the code currently can't run with Muon. Maybe allow `--optimizer-cpu-offload` without `--use-precision-aware-optimizer` when also using `--use-layer-wise-distributed-optimizer`.
> 2. It's too late to call `LayerWiseDistributedOptimizer.reload_optimizer_states` in `LayerWiseDistributedOptimizer.step_with_ready_grads`:
>    `prepare_grads` would already have been called by `ChainedOptimizer.step` before `step_with_ready_grads` is executed. And in `MixedPrecisionOptimizer.prepare_grads`, `_copy_model_grads_to_main_grads` tries to copy CUDA `main_param.main_grad` tensors to the CPU `main_param.grad` tensors, resulting in a device mismatch `RuntimeError`.
>    You'd need to call `reload_optimizer_states` in a new `LayerWiseDistributedOptimizer.prepare_grads` override method before executing the `super` method. Note that you may still want to leave the original location where `reload_optimizer_states` is called in, in case `step_with_ready_grads` is called directly (without `prepare_grads`).
> 3. Setting `fallback_config.optimizer_cpu_offload = False` in `megatron/core/optimizer/__init__.py:899` prevents Adam parameters from being offloaded in this case, because a separate `HybridDeviceOptimizer` is created, which is _not_ managed by the `LayerWiseDistributedOptimizer` (refer to `megatron/core/optimizer/__init__.py:965`).
>    So this should be removed. Checking correct offloading of Adam-managed parameters in both cases (`use_separate_distributed_optimizer=True` and `use_separate_distributed_optimizer=False`) to the tests would also make a lot of sense!


> Hey, thanks for the implementation and additional feature! I couldn't get the code to run immediately; have you actually tested this?
> 
> 1. `--optimizer-cpu-offload` requires `--use-precision-aware-optimizer` (`megatron/training/arguments.py:1665`).
>    However, `--use-precision-aware-optimizer` requires Adam use (`megatron/core/optimizer/optimizer_config.py:419`).
>    So the code currently can't run with Muon. Maybe allow `--optimizer-cpu-offload` without `--use-precision-aware-optimizer` when also using `--use-layer-wise-distributed-optimizer`.
> 2. It's too late to call `LayerWiseDistributedOptimizer.reload_optimizer_states` in `LayerWiseDistributedOptimizer.step_with_ready_grads`:
>    `prepare_grads` would already have been called by `ChainedOptimizer.step` before `step_with_ready_grads` is executed. And in `MixedPrecisionOptimizer.prepare_grads`, `_copy_model_grads_to_main_grads` tries to copy CUDA `main_param.main_grad` tensors to the CPU `main_param.grad` tensors, resulting in a device mismatch `RuntimeError`.
>    You'd need to call `reload_optimizer_states` in a new `LayerWiseDistributedOptimizer.prepare_grads` override method before executing the `super` method. Note that you may still want to leave the original location where `reload_optimizer_states` is called in, in case `step_with_ready_grads` is called directly (without `prepare_grads`).
> 3. Setting `fallback_config.optimizer_cpu_offload = False` in `megatron/core/optimizer/__init__.py:899` prevents Adam parameters from being offloaded in this case, because a separate `HybridDeviceOptimizer` is created, which is _not_ managed by the `LayerWiseDistributedOptimizer` (refer to `megatron/core/optimizer/__init__.py:965`).
>    So this should be removed. Checking correct offloading of Adam-managed parameters in both cases (`use_separate_distributed_optimizer=True` and `use_separate_distributed_optimizer=False`) to the tests would also make a lot of sense!

@janEbert, thank you so much for your review. In summary: yes, I have done unit tests and a GLM5 training with this code, but for various reasons, the 3 issues were hidden.

For my unit test cases, since the test was created bypassing arguments initialization, so issue 1 is skipped. For issue 2: since the old tests used use_cpu_initialization=True, model params were fp32 (torch.cuda.FloatTensor), which routes through fp32_from_fp32_groups — these are never offloaded. The device mismatch only triggers on the float16_groups → fp32_from_float16_groups path (where master weights get offloaded to CPU). So the buggy path was never exercised. For issue 3: the tests use use_distributed_optimizer=False, which only takes the legacy path where Adam is a child inside LayerWise (correctly has offload disabled). The use_separate_distributed_optimizer=True path (where Adam's DistOpt is a sibling and should manage its own offloading) requires full DDP setup that the unit tests didn't cover.

For my GLM5 training, since I did the training several weeks ago when def step() was still there, so I did:

    @torch.no_grad()
    def step(self):  # type: ignore[no-untyped-def]
        """step function for layer-wise optimizer."""
        if self._cpu_offload:
            self.reload_optimizer_states()
and now there is no longer def step() — instead we have prepare_grads + step_with_ready_grads. Naively putting self.reload_optimizer_states into step_with_ready_grads causes the device mismatch because prepare_grads() (which calls _copy_model_grads_to_main_grads) runs before step_with_ready_grads.

Fixes applied:

Issue 1: Relaxed the assert to also accept --use-layer-wise-distributed-optimizer.
Issue 2: Added a prepare_grads() override that reloads states before super().prepare_grads().
Issue 3: Removed fallback_config.optimizer_cpu_offload = False from the use_separate_distributed_optimizer=True path (Adam's DistOpt is a sibling, not a child — it manages its own HybridDeviceOptimizer offloading). Kept it in the legacy path where Adam feeds into LayerWise, also some more tests on the legacy path is added.
LMK if you have additional questions. thank you!

megatron/core/optimizer/layer_wise_optimizer.py
pengdurice: mentioned in https://github.com/NVIDIA/Megatron-LM/pull/4475#issuecomment-4580203252

megatron/core/optimizer/__init__.py
pengdurice: mentioned in https://github.com/NVIDIA/Megatron-LM/pull/4475#issuecomment-4580203252,

megatron/core/optimizer/__init__.py
pengdurice: thank you for your review, response is in https://github.com/NVIDIA/Megatron-LM/pull/4475#issuecomment-4580203252 ;-)

janEbert:
Thanks a lot for the changes! The code is still lacking a test for `use_separate_distributed_optimizer=True`, which would be nice to have to assert that the offloading behavior for Adam params doesn't regress at some point. :)

> In the separate DistributedOptimizer path (use_distributed_optimizer=True),
> Adam's DistOpt is a sibling of LayerWise — it manages its own offloading.
> That path requires full DDP setup and is validated via integration tests.

(From [6583dd4223280d87376b6bf1e0f561213748c588](https://github.com/NVIDIA/Megatron-LM/pull/4475/commits/6583dd4223280d87376b6bf1e0f561213748c588#diff-b618f5de2c241cdc25e026433b07ba1f667c6e9709063008003b616a46c1f070R125-R127).)

Maybe this can be mocked.

pengdurice:
> Thanks a lot for the changes! The code is still lacking a test for `use_separate_distributed_optimizer=True`, which would be nice to have to assert that the offloading behavior for Adam params doesn't regress at some point. :)
> 
> > In the separate DistributedOptimizer path (use_distributed_optimizer=True),
> > Adam's DistOpt is a sibling of LayerWise — it manages its own offloading.
> > That path requires full DDP setup and is validated via integration tests.
> 
> (From [6583dd4223280d87376b6bf1e0f561213748c588](https://github.com/NVIDIA/Megatron-LM/pull/4475/commits/6583dd4223280d87376b6bf1e0f561213748c588#diff-b618f5de2c241cdc25e026433b07ba1f667c6e9709063008003b616a46c1f070R125-R127).)
> 
> Maybe this can be mocked.

Addressed by adding setting the two flags to true so that it is using separate and it passed. thank you!

janEbert:
Could you also address the AI review comments? They seem very reasonable to me. Thanks a lot!

pengdurice:
> Could you also address the AI review comments? They seem very reasonable to me. Thanks a lot!

@janEbert ,  just addressed, thank you!

janEbert:
Awesome, thank you so much!

janEbert (approved):

janEbert:
Hey, I ran `tools/autoformat.sh` because there were some linting errors. Just so you know for the future. :)

janEbert (changes requested):
Dear @pengdurice, sorry to be the bearer of bad news. We're moving toward a unification of layer- and elementwise optimizers and these changes lead to further divergence between these types of optimizers.

In order not to merge the PR by accident, I'll block it for now while we figure out how to continue with this. For now, I appreciate greatly that you took the time to address all the requests and am sorry that we cannot move forward yet.

pengdurice:
> Dear @pengdurice, sorry to be the bearer of bad news. We're moving toward a unification of layer- and elementwise optimizers and these changes lead to further divergence between these types of optimizers.
> 
> In order not to merge the PR by accident, I'll block it for now while we figure out how to continue with this. For now, I appreciate greatly that you took the time to address all the requests and am sorry that we cannot move forward yet.

Hi @janEbert, thanks for the heads up — totally understandable and no worries at all! I appreciate you taking the time to review and provide thorough feedback throughout the process. It was a great learning experience.

May I stay in the loop on the unification effort? Once the new design for layer/ elementwise optimizer unification is taking shape, I'd be happy to rebase or rewrite the CPU offload feature to align with the new architecture. 
If there's a tracking issue or design doc where I could follow along (or provide feedback from the perspective of a user running Muon + CPU offload in production), I'd welcome the chance to contribute. Either way, just let me know when the time is right and I'll adapt accordingly.

Thanks again for all the guidance!

janEbert:
Thank you for being so understanding! I'm currently not involved in the unification effort, so it's hard to say for me whether it's possible to refactor this PR.

I'll also convert it back to draft for now so it doesn't clutter the list.

pengdurice:
> Thank you for being so understanding! I'm currently not involved in the unification effort, so it's hard to say for me whether it's possible to refactor this PR.
> 
> I'll also convert it back to draft for now so it doesn't clutter the list.

Got you, just let me know if you need anything from me in the future;)

PR #4481, opened 2026-04-27
[dev] [DeepSeek-v4] Part 2: Hash MoE and SwiGLU clamp

hxbai:
# What does this PR do ?
We will create several PRs to functionally support DeepSeek-v4 training. This is the second one.

Add DeepSeek-v4 Hash MoE and SwiGLU clamp.

- Add new argument `--moe-n-hash-layers`.
- Add SwiGLU support to `--activation-func-clamp-value`.

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/core/transformer/transformer_block.py
```diff
@@ -381,6 +382,12 @@ def build_layer(layer_spec, layer_number):
                 hidden_size=self.config.hidden_size,
                 eps=self.config.layernorm_epsilon,
             )
+            if self.config.enable_hyper_connections:
+                hc_mult = self.config.num_residual_streams
+                hc_dim = self.config.hidden_size * hc_mult
+                self.hc_head_fn = nn.Linear(hc_dim, hc_mult, bias=False)
+                self.hc_head_base = nn.Parameter(torch.empty(hc_mult))
+                self.hc_head_scale = nn.Parameter(torch.empty(1))
```
Victarry: **[CRITICAL]** `hc_head_base` and `hc_head_scale` are created with `torch.empty(...)`, which returns uninitialized memory. For from-scratch training (no checkpoint load), the model will start with arbitrary garbage values, and the values may differ across ranks because they depend on the memory allocator state. This breaks distributed-training determinism and likely causes loss instability/NaN.

The DSv4-Pro inference code uses `torch.empty` too, but it is inference-only and always loads a real checkpoint, so this is safe there. For Megatron training we must initialize explicitly.

**Suggestion:**
```python
self.hc_head_base = nn.Parameter(torch.zeros(hc_mult))
self.hc_head_scale = nn.Parameter(torch.ones(1))
```
or call an explicit init function. `hc_head_fn` is `nn.Linear`, so `reset_parameters` already covers it; only the two raw `nn.Parameter`s need explicit init.

megatron/core/transformer/moe/router.py
```diff
@@ -163,13 +162,37 @@ def __init__(
...
+            # Use a round-robin approach to initialize the tid2eid buffer.
+            vocab_size = self.config.actual_vocab_size
+            num_experts = self.config.num_moe_experts
+            ids = torch.arange(vocab_size, device=torch.cuda.current_device())
+            tid2eid = torch.stack(
+                [(ids + k) % num_experts for k in range(self.topk)], dim=1,
+            ).to(torch.int32)
+            self.register_buffer('tid2eid', tid2eid)
+        else:
+            self.tid2eid = None
```
Victarry, line 198: **[IMPORTANT]** Round-robin `tid2eid` initialization is a placeholder that is unsafe for real training, and there is no warning to flag this to users.

The DSv4-Pro inference checkpoint ships a learned `tid2eid` table (loaded as `nn.Parameter(requires_grad=False)`); the inference code does not reveal the training-time init recipe. A round-robin `(token_id + k) % num_experts` mapping has period `num_experts`, so when `vocab_size >> num_experts` (DSv4: 129280 / 384 ≈ 337 full periods), every expert combination is reused ~337 times. Combined with skewed token frequencies, this leads to severe expert load imbalance and a likely collapse of hash-layer routing in early training.

**Suggestion:** emit a one-time warning when a hash router is constructed and document the limitation, e.g.:
```python
import logging
logger.warning_once(
    "Hash MoE layer initialized with placeholder round-robin tid2eid. "
    "For real training, you MUST either (a) load tid2eid from a "
    "pre-trained DSv4 checkpoint, or (b) provide a frequency-aware "
    "initialization (e.g., Sinkhorn-balanced over token frequency). "
    "Round-robin will cause severe expert imbalance."
)
```
Optionally expose an `init_tid2eid(method=...)` hook so users can plug in a custom strategy without monkey-patching.

megatron/core/fusions/fused_bias_swiglu.py
```diff
@@ -48,6 +48,21 @@ def weighted_swiglu(y, weights):
     return res.to(dtype)
 
 
+@jit_fuser
+def clamped_swiglu(y, clamp_value):
+    y_1, y_2 = torch.chunk(y, 2, -1)
+    y_1 = y_1.clamp(min=None, max=clamp_value)
+    y_2 = y_2.clamp(min=-clamp_value, max=clamp_value)
+    return F.silu(y_1) * y_2
+
```
Victarry, line 59: **[IMPORTANT Precision]** Clamped SwiGLU is computed entirely in the input dtype (typically bf16), but the DSv4-Pro reference implementation explicitly upcasts to fp32 before clamping/SiLU.

DSv4 `Expert.forward` (`inference/model.py`):
```python
gate = self.w1(x).float()
up = self.w3(x).float()
if self.swiglu_limit > 0:
    up = torch.clamp(up, min=-self.swiglu_limit, max=self.swiglu_limit)
    gate = torch.clamp(gate, max=self.swiglu_limit)
x = F.silu(gate) * up
```

With `clamp_value` around 10.0 and bf16 inputs, the SiLU region near the clamp boundary loses precision and the elementwise product can drift from fp32 reference. This is the same reason RMSNorm is conventionally promoted to fp32.

**Suggestion:** keep the clamp + SiLU in fp32 inside the fused kernel, then cast back to the input dtype on output. Apply the same change to `clamped_weighted_swiglu` and the corresponding backward `clamped_swiglu_back` / `clamped_weighted_swiglu_back` so forward/backward stay numerically consistent.

megatron/core/transformer/transformer_block.py
```diff
@@ -922,9 +934,17 @@ def forward(
...
+            # DSv4 modifies the output contract implementation
+            rsqrt = torch.rsqrt(
+                hidden_states.square().mean(-1, keepdim=True) + self.config.layernorm_epsilon
+            )
+            mixes = self.hc_head_fn(hidden_states) * rsqrt
+            pre = torch.sigmoid(mixes * self.hc_head_scale + self.hc_head_base) + 1e-6
+            n = self.config.num_residual_streams
+            hidden_states = torch.sum(
+                pre.unsqueeze(-1) * hidden_states.view(*hidden_states.shape[:-1], n, -1), dim=-2
+            )
```
Victarry: **[IMPORTANT Precision]** The new mHC final contract runs `square().mean()`, `rsqrt`, `sigmoid` and the weighted sum entirely in the input dtype (typically bf16). The DSv4-Pro reference (`ParallelHead.hc_head` in inference/model.py) explicitly promotes the input to fp32 for the whole block:

```python
def hc_head(self, x, hc_fn, hc_scale, hc_base):
    shape, dtype = x.size(), x.dtype
    x = x.flatten(2).float()                       # <-- fp32
    rsqrt = torch.rsqrt(x.square().mean(-1, keepdim=True) + self.norm_eps)
    mixes = F.linear(x, hc_fn) * rsqrt
    pre = torch.sigmoid(mixes * hc_scale + hc_base) + self.hc_eps
    y = torch.sum(pre.unsqueeze(-1) * x.view(shape), dim=2)
    return y.to(dtype)
```

`square().mean()` in bf16 is the same precision pitfall as RMSNorm — small variance values get rounded out. For deep stacks the residual-stream mixing can compound this drift.

**Suggestion:** mirror DSv4 — cast `hidden_states` to fp32 before the rsqrt/linear/sigmoid block, and cast the final summed output back to the input dtype.

megatron/core/transformer/moe/router.py
```diff
@@ -44,10 +45,11 @@ def __init__(
             is_mtp_layer (bool): Flag indicating if this router is part of an MTP layer.
         """
         super().__init__(config)
+        assert layer_number is not None, "layer_number is required for the router."
```
Victarry: **[IMPORTANT Compatibility]** Making `layer_number` a required constructor argument and removing `set_layer_number` is a breaking change to a `megatron/core/` public surface (`Router`, `MoELayer`).

Any external code that builds these via `submodules.router(config=..., pg_collection=..., is_mtp_layer=...)` without `layer_number`, or that relied on the two-step `__init__` + `set_layer_number(...)` flow, will hit this assertion immediately.

**Suggestion:** keep the new constructor argument, but
- default `layer_number=None`,
- emit a `DeprecationWarning` when `layer_number is None` instead of asserting,
- temporarily preserve `set_layer_number` as a deprecated shim that sets the field and warns.
Alternatively, document the breakage prominently in release notes / CHANGELOG and provide a migration recipe.

megatron/core/transformer/transformer_layer.py
```diff
@@ -784,6 +789,10 @@ def _forward_mlp(
             self.config.inference_fuse_tp_communication
         )
 
+        moe_kwargs = {}
+        if self.is_moe_layer and input_ids is not None:
+            moe_kwargs["input_ids"] = input_ids
```
Victarry, line 794: **[IMPORTANT]** `moe_kwargs` is set up here, but the chunked-prefill branch at line ~822 (and the analogous one in `HyperConnectionTransformerLayer._forward_mlp` near line ~1775) bypasses both `padding_mask` and `**moe_kwargs`:
```python
outputs = [self.mlp(chunk) for chunk in chunks]
```
For hash-routing layers, the router will hit `assert input_ids is not None` during prefill chunking and crash.

**Suggestion:** either forward `padding_mask` and `**moe_kwargs` into each chunked call (and decide whether `input_ids` should also be chunked along the sequence dimension), or add an `__post_init__` check that `moe_n_hash_layers > 0` is incompatible with `mlp_chunks_for_prefill`, mirroring the existing assert that excludes `overlap_moe_expert_parallel_comm`.

megatron/core/transformer/moe/router.py
```diff
@@ -583,14 +606,64 @@ def _apply_expert_bias(
...
+        num_tokens, num_experts = logits.shape
+
+        if self.score_function == "softmax":
+            scores = torch.softmax(logits, dim=-1, dtype=torch.float32).type_as(logits)
+        elif self.score_function == "sigmoid":
+            scores = torch.sigmoid(logits.float()).type_as(logits)
+        elif self.score_function == "sqrtsoftplus":
+            scores = torch.nn.functional.softplus(logits.float()).sqrt().type_as(logits)
+        else:
+            raise ValueError(f"Invalid score_function: {self.score_function}")
```
Victarry, line 638: **[SUGGESTION Precision]** `_hash_routing` casts the scores back to the input dtype (`type_as(logits)`) immediately after computing them in fp32, then performs `gather`, scatter and the optional renormalization in the input dtype. The DSv4 reference keeps the whole gating pipeline in fp32 until the final return:

```python
scores = linear(x.float(), self.weight.float())   # fp32
...
weights = original_scores.gather(1, indices)
if score_func != "softmax":
    weights /= weights.sum(dim=-1, keepdim=True)
weights *= self.route_scale
```

For `score_function='sqrtsoftplus'` in particular, the `softplus -> sqrt -> /sum` chain in bf16 can lose meaningful precision in the routing weights.

**Suggestion:** keep `scores` in fp32 through `gather`, the optional sum-renormalization and the `topk_scaling_factor` multiplication; cast to the original dtype only at the final scatter (or at the `routing_probs` return).

megatron/core/transformer/transformer_config.py
```diff
@@ -200,7 +200,7 @@ class TransformerConfig(ModelParallelConfig):
 
     activation_func_clamp_value: Optional[float] = None
     """Clamp the output of the linear_fc1 in the activation function. Only used when activation_func
```
Victarry, line 202: **[SUGGESTION]** The docstring now claims `activation_func_clamp_value` works for `quick_gelu` or `swiglu`, but in this PR clamp is only wired through the *weighted* SwiGLU path (`weighted_bias_swiglu_impl` -> `WeightedSwiGLUFunction`). The dense path `bias_swiglu_impl` (used by non-MoE / non-token-weighted SwiGLU MLP) does not accept `clamp_value` and silently ignores it.

So a user setting `activation_func_clamp_value` on a model with dense SwiGLU MLP layers will see zero effect, with no warning.

**Suggestion:** either
- extend `bias_swiglu_impl` / `BiasSwiGLUFunction` / `SwiGLUFunction` to also accept and respect `clamp_value`, or
- narrow the docstring to "weighted SwiGLU (MoE) only" and add a runtime check that warns or asserts when `activation_func_clamp_value > 0` is set on a dense-SwiGLU configuration.

megatron/core/transformer/moe/router.py
```diff
@@ -163,13 +162,37 @@ def __init__(
...
         self.input_jitter = None
 
-        self.enable_expert_bias = self.config.moe_router_enable_expert_bias
+        self.is_hash_layer = (
+            self.config.moe_n_hash_layers > 0
+            and layer_number <= self.config.moe_n_hash_layers
+        )
+        if self.is_hash_layer:
+            # DSv4 paper does not provide the initialization method for the hash buffer.
+            # Use a round-robin approach to initialize the tid2eid buffer.
```
Victarry: **[SUGGESTION]** The comment `DSv4 paper does not provide the initialization method for the hash buffer.` is a bit misleading: as of now there is no public DSv4 paper at all, so a future reader may waste time trying to find it. The real situation is that the DSv4-Pro inference checkpoint ships a *learned* `tid2eid` table loaded from safetensors, and the training-time init recipe is not published anywhere.

**Suggestion:** rewrite as something like:
```python
# DSv4-Pro ships a pre-trained tid2eid table in its inference checkpoint;
# no public initialization recipe is documented. Round-robin is used here
# only as a placeholder so the layer is runnable from scratch — see the
# warning above and the docs for moe_n_hash_layers.
```

Victarry (commented):
Generally looks good to me. Co-reviewd with AI, please take a look~

megatron/core/transformer/transformer_block.py
hxbai: fixed

megatron/core/transformer/moe/router.py
hxbai: added the logging in transformer_config.py

megatron/core/fusions/fused_bias_swiglu.py
hxbai: fixed

megatron/core/transformer/transformer_block.py
hxbai: fixed and moved to a standalone function call with torch.compile

megatron/core/transformer/moe/router.py
hxbai: I restored the `set_layer_number` API for outside usage to avoid breaking backward compatibility.
But in MCore, I still modified to use the new layer_number arg to support the hash router.

megatron/core/transformer/transformer_layer.py
hxbai: We do not care about the inference path for now. I'll leave it as this.

megatron/core/transformer/moe/router.py
hxbai: This is outside the scope of the PR. The dtype logic was updated in this PR https://github.com/NVIDIA/Megatron-LM/pull/4193

megatron/core/transformer/transformer_config.py
hxbai: fixed the docstring

megatron/core/transformer/moe/router.py
hxbai: fixed

megatron/core/transformer/transformer_block.py
hxbai: I removed the mHC modification from this PR and moved to https://github.com/NVIDIA/Megatron-LM/pull/4518/changes for consistency

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4481/commits/df71a39502239a95a27f66850b39f34986d21e1d

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4481/commits/c51d4616213d0cbd2707f97dbace07b10e84e523

Victarry:
LGTM

Victarry (approved):

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4481/commits/608a1b25b9d4bdf39955a3fdbc95223cdf1e93b3

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4481/commits/dfadf2e3b1b6fe4182cb2d401cdf7e13ab0c1be4

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4481/commits/d6a9445dab7c7f8cde8075ba4436199e2920f16a

PR #4486, opened 2026-04-27
Add preliminary Muon+M-FSDP support

janEbert:
Introduce Muon support to M-FSDP. Currently 1.5×–2.7× as slow compared to an Adam baseline with a 1B–8B DeepSeek-V3 proxy model. Peak memory slightly lower than with Adam (4–7 % less).

### DIFF
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -741,8 +741,9 @@ def _build_megatron_fsdp_emerging_optimizer(
 
       1. Builds the linear-only Muon optimizer (plain `TensorParallelMuon` for
          `no_shard`; `FSDPZeROTensorParallelMuon` for sharded strategies, which
-         allgathers DP row-shards before NS) over a frozen-grad view of the
-         model so `_get_param_groups` only emits Muon-managed params.
+         gathers only split rank-boundary DTensor parameters before NS) over a
+         frozen-grad view of the model so `_get_param_groups` only emits
+         Muon-managed params.
       2. Falls back to `get_megatron_optimizer` for the non-linear params,
          which routes through the standard Megatron-FSDP path for Adam.
       3. Wraps the chained result with `FSDPMuonChainedOptimizer`, which drives
@@ -756,9 +757,9 @@ def _build_megatron_fsdp_emerging_optimizer(
     # Choose Muon variant based on the inner-DP sharding strategy.
     # - "no_shard" (ZeRO-0): params/grads are full Replicate DTensors, so the
     #   plain TensorParallelMuon works without any DP communication.
-    # - Sharded strategies (ZeRO-1/2/3): finish_grad_sync() reduce-scatters
-    #   gradients into Shard(0) row-shards. FSDPZeROTensorParallelMuon
-    #   allgathers across the DP group before NS and re-shards the result.
+    # - Sharded strategies (ZeRO-1/2/3): FSDPZeROTensorParallelMuon gathers
+    #   only split boundary parameters across the DP group before NS and
+    #   re-shards the result.
     fsdp_strategy = getattr(
         model_chunks[0].ddp_config, "data_parallel_sharding_strategy", "no_shard"
     )
@@ -813,8 +814,8 @@ def _build_megatron_fsdp_emerging_optimizer(
     optimizers: List[Any] = []
     init_fns: List[Callable] = []
     if linear_param_groups:
-        # ZeRO-1/2/3: gradients are reduce-scattered over `dp_cp` for dense
-        # params, so the FSDPZero variant must allgather over the same group.
+        # ZeRO-1/2/3: split dense boundary params are sharded over `dp_cp`, so
+        # the FSDPZero variant must gather over the same group.
         dense_kwargs = dict(eopt_kwargs)
         if use_fsdp_zero_muon:
             dense_kwargs["dp_group"] = pg_collection.dp_cp
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -64,7 +64,9 @@ from .distrib_optimizer import DistributedOptimizer
 from .emerging_optimizers import (
     _EMERGING_OPTIMIZERS,
     HAVE_EMERGING_OPTIMIZERS,
+    FSDPMuonChainedOptimizer,
     _create_emerging_optimizer,
+    _get_mfsdp_models,
 )
 from .grad_scaler import ConstantGradScaler, DynamicGradScaler
 from .layer_wise_optimizer import LayerWiseDistributedOptimizer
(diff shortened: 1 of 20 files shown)


megatron/core/optimizer/__init__.py
```diff
@@ -721,6 +724,193 @@ def check_config_overrides_consistency(
...
+                nonlinear_params.append(param)
+
+    # Build the Muon optimizer for linear params.
+
+    for p in nonlinear_params:
+        p.requires_grad = False
+    linear_param_groups = _get_param_groups(model_chunks, config, config_overrides)
+
+    expert_param_groups: List[Dict[str, Any]] = []
+    if not use_layer_wise:
```
cspades: Is this because layer-wise <> EP isn't compatible, so we just use Adam? (Gathering expert params is too heavyweight? Also, another complexity is EP-DP sharding, so we need to gather both.)

megatron/core/optimizer/__init__.py
```diff
@@ -721,6 +724,193 @@ def check_config_overrides_consistency(
...
+    adam_optimizers = list(getattr(chained_adam, "chained_optimizers", [chained_adam]))
+    optimizers += adam_optimizers
+
+    def _adam_init_state_fn(opt: Any, config: Any = None) -> None:
+        for group in opt.param_groups:
+            for p in group["params"]:
+                if len(opt.state[p]) == 0:
+                    if config is None or not config.use_precision_aware_optimizer:
+                        opt.state[p]["exp_avg"] = torch.zeros_like(p.data)
+                        opt.state[p]["exp_avg_sq"] = torch.zeros_like(p.data)
```
cspades: Also what does this case capture? With MFSDP, I think we need DTensors so we usually do a "dummy step" to get DTensor state based on the DTensor params.

Two examples:

- MCore: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/optimizer/distrib_optimizer.py#L1298 (`_init_optimizer_states_with_dummy_values`)
- Torch DCP: https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/state_dict.py#L609 (`_init_optim_state`)

megatron/core/optimizer/__init__.py
```diff
@@ -721,6 +724,193 @@ def check_config_overrides_consistency(
...
+                        opt.initialize_state(p)
+
+    init_fns += [_adam_init_state_fn] * len(adam_optimizers)
+
+    # Combine and wrap with the FSDP-protocol adapter.
+
+    if use_layer_wise:
+        # Float16OptimizerWithFloat16Params is incompatible with FSDP DTensor
+        # params (no .main_grad), so temporarily clear bf16 to prevent the
+        # LayerWiseDistributedOptimizer from re-wrapping each sub-optimizer.
```
cspades: Either we should extend the logic in `LayerWiseDistributedOptimizer` or the comment should be simpler, like this:
> `LayerWiseDistributedOptimizer(config.bf16)` wraps the optimizer with a Float16 Megatron optimizer if BF16 is used. Megatron-FSDP only supports DistributedOptimizer and LayerWiseDistributedOptimizer.

Note: Layer-wise appears to be a permutation of optimizer params into different layer-specific groups distributed across DP-CP (or EDP). I am not sure if this is compatible with Megatron-FSDP, since we have un-even shards. Is it possible that, say, non-empty Params are all allocated to a different DP rank, but that rank's equivalent Params are empty, so no optimizer update?

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -146,6 +161,67 @@ def _get_qkv_split_shapes(model_cfg) -> List[int]:
...
+
+def _get_mfsdp_models(model_chunks):
+    """Extract list of MegatronFSDP instances from FSDP-wrapped model chunks."""
+    mfsdp_models = []
+    for chunk in model_chunks:
+        # FullyShardedDataParallel delegates finish_grad_sync / start_param_sync
+        # from its .module (MegatronFSDP). install_optimized_model_weights lives
+        # directly on MegatronFSDP, so we need the inner module reference.
+        if hasattr(chunk, "finish_grad_sync") and hasattr(chunk, "module"):
+            mfsdp_models.append(chunk.module)
```
cspades: Can we just check directly for `FullyShardedDataParallel`? Many Megatron wrappers satisfy the current if condition!

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -276,6 +352,208 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+            return None
+        if isinstance(value, _DTensor):
+            return value
+        data = getattr(value, "data", None)
+        if isinstance(data, _DTensor):
+            return data
+        return None
+
+    def _is_nonempty_dtensor_param(self, param: torch.Tensor) -> bool:
+        dtensor = self._as_dtensor(param)
```
cspades: If I didn't mention this before, param should be from the optimizer.param_groups which should already be `DTensor` right? We really only need to go to Tensor land for any GEMMs or Muon ops.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -276,6 +352,208 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+        if self.dp_group is None or get_pg_size(self.dp_group) == 1:
+            if p.grad is None:
+                return
+            return self._local_muon_update(p, p.grad, group)
+
+        value = self._as_dtensor(p)
+        if value is None:
+            if p.grad is None:
+                return
+            return self._local_muon_update(p, p.grad, group)
```
cspades: Hmm, why would `value` be `None` here? (Why would `_as_dtensor` not have data, and also can we just guard `_HAVE_DTENSOR` for FSDP use?)

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -276,6 +352,208 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+            return
+
+        state = self.state[p]
+        mom_buffer = state["momentum_buffer"]
+        mom_dtensor = self._as_dtensor(mom_buffer)
+        mom_local = mom_dtensor.to_local() if mom_dtensor is not None else mom_buffer
+
+        grad = p.grad
+        if grad is None:
+            local_grad = torch.zeros_like(mom_local)
```
cspades: Should be empty like the param to get a zero DTensor grad. Ideally, this should be guarded by requires grad and we error out for missing gradients right?

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -276,6 +352,208 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+            idx for idx, param in enumerate(params) if self._needs_boundary_gather(param)
+        ]
+
+        if self.dp_group is None or get_pg_size(self.dp_group) == 1:
+            return set(local_boundary_indices)
+
+        gathered_indices: list[list[int] | None] = [None] * get_pg_size(self.dp_group)
+        torch.distributed.all_gather_object(
+            gathered_indices, local_boundary_indices, group=self.dp_group
+        )
```
cspades: I think we don't need this AG. The current DP rank doesn't need to know what other DP ranks param indices need gathering. We know everything from the current rank's parameter's global shape and local shape.

megatron/training/arguments.py
```diff
@@ -1468,12 +1471,34 @@ def validate_args(args, defaults={}):
...
 
         if args.use_distributed_optimizer:
-            args.use_layer_wise_distributed_optimizer = True
-            args.use_distributed_optimizer = False
+            # Megatron-FSDP already shards optimizer state itself (ZeRO-1/2/3
+            # via `--data-parallel-sharding-strategy`). Layering
+            # `LayerWiseDistributedOptimizer` on top would double-shard and,
+            # in practice, trips `TypeErrors` in its constructor call from
+            # `_build_megatron_fsdp_emerging_optimizer`. The M-FSDP factory
+            # already handles the "distributed" part via `FSDPMuonChainedOptimizer`.
```
cspades: The "double shard" here means MFSDP shards tensors unit-wise, while this layer-wise distopt distributes parameters as well. I think I mentioned this above, this needs to be brutally checked and confirmed, and this comment needs to be updated to not be completely vague / meaningless.

megatron/training/arguments.py
```diff
@@ -1468,12 +1471,34 @@ def validate_args(args, defaults={}):
...
-        assert not args.use_megatron_fsdp, "Emerging optimizer does not support Megatron-FSDP for now."
-        assert args.ckpt_format in ["torch", "torch_dist"], "Emerging optimizer supports torch and torch_dist checkpoint format."
+        if args.use_megatron_fsdp:
+            assert args.optimizer == "muon", (
+                "Emerging optimizer with Megatron-FSDP is currently only supported for Muon."
+            )
+            assert args.outer_dp_sharding_strategy == "no_shard", (
+                "Emerging optimizer with Megatron-FSDP does not support HSDP "
+                "(--outer-dp-sharding-strategy != no_shard) yet."
+            )
```
cspades: This is a red flag. Because we are using DP-CP group, it should work OOTB, the optimizer doesn't care about Megatron-FSDP internal sharding strategies (like DP-inner or DP-outer). It should only see the fully-sharded optimizer state and main parameters on the cumulative DP-CP group. We def need support for HFSDP.

megatron/training/arguments.py
```diff
@@ -1468,12 +1471,34 @@ def validate_args(args, defaults={}):
...
+            # Megatron-FSDP itself requires `fsdp_dtensor` (asserted above), so
+            # the emerging-optimizer path must accept it here to avoid a
+            # contradictory assertion pair.
+            assert args.ckpt_format == "fsdp_dtensor", (
+                "Emerging optimizer with Megatron-FSDP requires "
+                "--ckpt-format fsdp_dtensor."
+            )
```
cspades: What does the optimizer state look like to DCP with layer-wise, if you could share, thanks!

If there is parameter distribution, we need to be careful about writing offsets.

cspades (changes requested):
Definitely many nits, ambiguities, and red flags. Will also test out on my side and comment on solutions. Good progress though! The `step()` implementation is as intended.

megatron/core/optimizer/emerging_optimizers.py
cspades: Since we removed the `_as_dtesnor` function that return `None` check if we still need this.

megatron/core/optimizer/emerging_optimizers.py
janEbert: Addressed in 127e03b60514e6efd3986d84ec48621b53060a37.

megatron/core/optimizer/emerging_optimizers.py
janEbert: Adressed in c78801406c746ddf7a53735783b903ba297ae102.

janEbert: Reverted for now because this didn't account for empty boundary params (i.e., params for which `p_local.numel() == 0 and needs_gather`, where we always have `p.grad is None`).

If you'd prefer another solution than empty dummy parameters, let's discuss.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -276,6 +297,556 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+
+    def _get_boundary_gather_param_indices(self, group: Dict[str, Any]) -> set[int]:
+        """Return globally-agreed parameters whose flat items cross FSDP shard boundaries."""
+        params = group["params"]
+        cache_key = tuple(id(param) for param in params)
+        cached_indices = self._boundary_gather_indices_cache.get(cache_key)
+        if cached_indices is not None:
+            return cached_indices
+
+        has_mfsdp_params = any(getattr(param, "orig_param", None) is not None for param in params)
```
cspades, line 924: BTW this isn't a good way to tell if we are using Megatron-FSDP, plus we really only care that the `DTensor.shape != DTensor._local_tensor.shape` for when we need to all-gather, which does not care about MFSDP.

megatron/core/optimizer/optimizer_config.py
```diff
@@ -289,6 +289,11 @@ class OptimizerConfig:
     """Optimizer for nonlinear parameters (embeddings, biases, norms) when using muon.
     One of 'adam' or 'lion'. Defaults to 'adam'."""
 
+    muon_fsdp_batched_all_gather: bool = False
+    """If True, batch Muon+M-FSDP boundary parameter all-gathers by dtype/device/group.
+    This can reduce collective count but increases temporary peak memory.
+    """
```
cspades, line 322: FYI, dtype and device are always the same for all parameters. Group here as in DP-Outer / DP-Inner or parameter groups / FSDP units?

janEbert: "Parameter group"; improved clarity.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -276,6 +297,556 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+            raise ValueError(
+                f"Expected exactly one chunk metadata per rank, got {len(chunk_metadata_list)}."
+            )
+
+        local_tensor = dtensor_ref.to_local()
+        local_chunk_metadata = chunk_metadata_list[0]
+        local_chunks_info = [
+            {"shape": torch.Size(local_tensor.shape), "offset": tuple(local_chunk_metadata.offsets)}
+        ]
+        shard_group = dtensor_ref.device_mesh.get_group(shard_mesh_dims[0])
```
cspades: This doesn't look right, why is the shard group just the first mesh dim? HFSDP has 2 sharding groups, for instance? (Or you need to flatten them.)

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -276,6 +297,556 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+            full_tensor = redistribute_uneven_dtensor_to_replicated(local_dtensor).to_local()
+            self._copy_dtensor_chunk_metadata(dtensor_ref, local_dtensor)
+            return None if local_tensor.numel() == 0 else full_tensor
+
+        local_buffer = local_tensor.contiguous().view(-1)
+        group_tensors = [
+            torch.empty(chunk_info["numel"], dtype=local_tensor.dtype, device=local_tensor.device)
+            for chunk_info in plan["chunk_infos"]
+        ]
+        torch.distributed.all_gather(group_tensors, local_buffer, group=plan["shard_group"])
```
cspades: See above, I don't know how this could work with HFSDP. You need to all-gather across two groups, not one. `chunk_info["numel"]` looks to be a full-shard, and I assume this all-gathers into a partial shard, likely in the wrong order (inner first [1], then outer [0]).

PR #4509, opened 2026-04-29
Integrate LayerWiseDistributedOptimizer with DDP buffer infrastructure

deepakn94:
## Summary

Adds a **shard-aligned parameter layout** for `LayerWiseDistributedOptimizer` that guarantees no parameter is split across shard boundaries — the invariant optimizers like Muon need so Newton-Schulz iteration can run on full weight matrices. The optimizer publishes this layout to DDP, which means DDP can manage the buffers exactly as it does for `DistributedOptimizer`:

- **Gradient reduction**: reduce-scatter via `use_distributed_optimizer=True`, replacing the previous all-reduce-then-pick-your-shard scheme.
- **Parameter sync**: DDP's standard buffer all-gather via `model_chunk.start_param_sync()`, replacing the legacy flatten / `all_gather_v` / unflatten path that copied data on both ends.

## Why

Before this PR, `LayerWiseDistributedOptimizer` ran outside DDP's buffer infrastructure:

- DDP was configured with `use_distributed_optimizer=False`, so gradients were all-reduced rather than reduce-scattered. Each rank held a full reduced gradient buffer even though it would only update its own shard.
- Parameter sync was a hand-rolled `all_gather_v` that flattened each rank's params into a contiguous tensor, all-gathered, then unflattened and `copy_`-ed the results back into model params. Two extra full-size copies per sync.

Both fell out of "the optimizer doesn't know how the buffer is laid out, so DDP can't be told to slice it." Once the optimizer can publish a layout that guarantees "every param fits inside one DP shard," DDP's existing reduce-scatter and buffer all-gather work directly.

### Shard-aligned size-matching layout

`LayerWiseDistributedOptimizer._compute_per_buffer_param_layout` produces a shard-aligned layout per buffer using a size-matching algorithm: each round claims one param for shard 0 and assigns same-sized params (or padding) to shards 1..dp-1, so all shards grow by exactly the same amount in lockstep. Shard sizes stay equal by construction; for repeated-layer models (N identical transformer blocks) padding overhead is zero. Every param ends up fully contained within a single DP-rank's shard — the invariant Muon's whole-tensor update requires.

Shared (tied) embeddings get isolated buckets — they are placed alone in shard 0 of their own bucket, with shards 1..dp-1 padded to the same size. This preserves the no-shard-crossing invariant but pays a `(dp_size - 1) * pad(numel)` cost per shared embedding. Eliminating that cost is the goal of a follow-up PR that routes Adam-managed params through a separate `DistributedOptimizer`.

### Optimizer state partitioning derived from the layout

`_shard_params_from_layout` derives each parameter's owning rank directly from the published layout (`(start - bucket_start) // shard_size`) instead of the previous independent ping-pong-by-numel assignment. Optimizer and DDP now agree on shard assignments by construction, eliminating the class of bugs where the two could disagree and the optimizer would step on a parameter whose gradient was reduce-scattered to a different rank. A defensive assertion in `_shard_params_from_layout` catches any param straddling a shard boundary, so layout regressions fail loudly rather than silently corrupting the emerging optimizer's update.

### Step flow

DDP is now configured with `use_distributed_optimizer=True` for layerwise mode, so model params are views into `bucket.param_data` and grads are reduce-scattered into each rank's shard of the gradient buffer during backward. On step:

1. `LayerWise.step` runs the Muon update on its local-rank shard of every layerwise buffer; the standard "main → model_param" copy updates the param buffer in place (because model_param *is* a view into the buffer).
2. `model_chunk.start_param_sync(force_sync=True)` syncs the buffer across DP ranks via DDP's standard all-gather — no flatten/unflatten copies.

### DDP wiring centralised in a helper

`wrap_model_chunks_with_ddp` (new, in `megatron/training/training.py`) centralises the DDP-construction logic shared between `get_model` and unit tests:

- Forces `ddp_config.use_distributed_optimizer = True` when `use_layer_wise_distributed_optimizer=True` (needed for reduce-scatter).
- Computes the per-chunk `full_param_layout` via `LayerWiseDistributedOptimizer.compute_full_param_layout` (layerwise) or `DistributedOptimizer.compute_full_param_layout` (standard distopt).
- Wraps each chunk with the layout passed through.

### Legacy paths kept as fallback (for the no-layout case)

The pre-PR sync paths are retained for callers that don't supply a layout yet and are marked for removal once all call sites pass one:

- `LayerWiseDistributedOptimizer.allgather_params` (flatten / `all_gather_v` / unflatten path).
- `LayerWiseDistributedOptimizer.set_bucket_layerwise_params_list`.
- `LayerWiseDistributedOptimizer._shard_params_ping_pong` (the ping-pong-by-numel fallback used when no layout is supplied).
- The variable-size all-gather branch in `_ParamAndGradBucketGroup.start_param_sync`.

## Tests

- New `tests/unit_tests/distributed/test_layerwise_param_layout.py`: covers the size-matching layout (uniform / mixed / shared-embedding / dp-divisibility / backprop ordering / `dp_size ∈ {1, 2, 4, 8}`).
- `test_emerging_optimizers` and `dist_checkpointing/test_layer_wise_optimizer`: updated to construct DDP via the new `wrap_model_chunks_with_ddp` helper so the layerwise wiring matches `training.get_model`.

## Test plan

- [x] Verify `test_layerwise_param_layout.py` passes (shard divisor, size-matching layout, shared-embedding isolation, bucket alignment, backprop ordering, full layout grouping)
- [x] Verify existing `test_param_layout.py` tests still pass
- [x] Verify `test_emerging_optimizers` tests pass against `LayerWiseDistributedOptimizer`
- [x] Verify `dist_checkpointing/test_layer_wise_optimizer` tests pass
- [x] Run Muon training with `--use-layer-wise-distributed-optimizer` and verify convergence matches the existing layerwise baseline
- [x] Verify reduce-scatter is used (not allreduce) when the layerwise optimizer is active — i.e. `ddp_config.use_distributed_optimizer=True` is in effect and grads are reduce-scattered into per-rank shards
- [x] Cluster: small Muon training run at `dp_size > 1` to verify gradient flow + convergence end-to-end

image

## Follow-up

Routing Adam-managed parameters through a separate `DistributedOptimizer` (sub-tensor sharding) so tied embeddings avoid the `(dp_size - 1) * pad(numel)` per-rank padding cost. That work is staged on a separate branch and will land as a follow-up PR.

### DIFF
--- a/megatron/core/distributed/distributed_data_parallel.py
+++ b/megatron/core/distributed/distributed_data_parallel.py
@@ -165,6 +165,8 @@ class DistributedDataParallel(_BaseDataParallel):
                     param_indices == layout.param_indices
                 ), f"param_indices for {buffer_key} do not match between grouping and layout"
 
+        self.full_param_layout = full_param_layout
+
         # Compute gradient scaling factors.
         if config.calculate_per_token_loss:
             assert (
--- a/megatron/core/distributed/param_and_grad_buffer.py
+++ b/megatron/core/distributed/param_and_grad_buffer.py
@@ -322,8 +322,9 @@ class _ParamAndGradBucketGroup:
         async_op = self.ddp_config.overlap_param_gather and not force_sync
 
         if not self.ddp_config.use_distributed_optimizer:
-            # Layer-wise optimizer path: use all_gather for variable-size
-            # param gather.
+            # Legacy layer-wise optimizer path: use all_gather for variable-size
+            # param gather.  Once all layerwise call sites set
+            # ddp_config.use_distributed_optimizer=True, this branch can be removed.
             #
             # Each rank may own a different number of params per bucket, so
             # layerwise_param_flat_sizes can vary across ranks.  PyTorch's NCCL
--- a/megatron/core/distributed/param_and_grad_buffer.py
+++ b/megatron/core/distributed/param_and_grad_buffer.py
@@ -208,6 +208,16 @@ class _ParamAndGradBucketGroup:
                 not self.ddp_config.reduce_scatter_with_fp32_accumulation
             ), "RS w/ FP32 accumulation not supported with num_distributed_optimizer_instances > 1"
 
+        reduction_collective = (
+            "reduce-scatter" if self.ddp_config.use_distributed_optimizer else "all-reduce"
+        )
+        log_single_rank(
+            logger,
+            logging.INFO,
+            f"Using {reduction_collective} for gradient reductions because "
+            f"{self.ddp_config.use_distributed_optimizer=}",
+        )
+
         global dist_reduce_scatter_func
         if self.ddp_config.reduce_scatter_with_fp32_accumulation:
             dist_reduce_scatter_func = reduce_scatter_with_fp32_accumulation
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -850,7 +850,7 @@ def _get_megatron_emerging_optimizer(
             config,
             pg_collection,
             init_state_fn_list=list(init_fns),
-            model_chunks=model_chunks if config.overlap_param_gather else None,
+            model_chunks=model_chunks,
         )
 
     return ChainedOptimizer(results)
--- a/megatron/core/optimizer/layer_wise_optimizer.py
+++ b/megatron/core/optimizer/layer_wise_optimizer.py
@@ -1,15 +1,18 @@
 # Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
 import logging
-from typing import Callable, List, Optional
+import math
+from collections import defaultdict
+from typing import Callable, Dict, List, Optional, Tuple
 
 import torch
 from torch._utils import _flatten_dense_tensors, _unflatten_dense_tensors
 
 from megatron.core.dist_checkpointing.dict_utils import nested_values
 from megatron.core.dist_checkpointing.mapping import LocalNonpersistentObject, ShardedStateDict
+from megatron.core.distributed.param_and_grad_buffer import group_params_for_buffers
 from megatron.core.process_groups_config import ProcessGroupCollection
-from megatron.core.utils import get_pg_rank, get_pg_size
+from megatron.core.utils import get_pg_rank, get_pg_size, log_single_rank
 
 from .clip_grads import count_zeros_fp32, get_grad_norm_fp32
 from .optimizer import (
@@ -19,6 +22,13 @@ from .optimizer import (
     MegatronOptimizer,
 )
 from .optimizer_config import OptimizerConfig
+from .param_layout import (
+    FullParamLayout,
+    PerBufferParamLayout,
+    bucket_end_divisor,
+    pad_param_start,
+    pad_to_divisor,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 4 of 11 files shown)


megatron/core/optimizer/layer_wise_optimizer.py
```diff
@@ -39,6 +49,304 @@ class LayerWiseDistributedOptimizer(ChainedOptimizer):
...
+            shard_assignments = [[] for _ in range(dp_size)]
+            shard_pos = 0
+            bucket_unpadded = 0
+
+        while True:
+            param = _next_unassigned()
+            if param is None:
+                break
+
+            S = param.data.nelement()
```
deepakn94: Let's use the same variable naming convention used in `distrib_optimizer.py`.

deepakn94:
Addressed review comment: renamed `S` to `param_numel` throughout `_compute_per_buffer_param_layout` and updated the docstring to match `distrib_optimizer.py` conventions.

Also in the latest push: all buffers now use the shard-aligned layout (not just Muon buffers). This ensures no param is split across shard boundaries regardless of optimizer type. The separate `BufferKey` grouping still keeps Muon and Adam params in different buffers to minimize size-matching padding.

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -124,9 +124,25 @@ def __init__(
...
+        # Group parameters by (param_dtype, grad_dtype, is_expert_parallel,
+        # use_layerwise_distributed_optimizer).
         buffer_groups = group_params_for_buffers(all_params, self.ddp_config.grad_reduce_in_fp32)
 
+        # Validate layerwise buffer requirements.
+        has_layerwise_buffers = any(
+            key.use_layerwise_distributed_optimizer for key in buffer_groups
+        )
+        if has_layerwise_buffers:
+            assert self.ddp_config.use_distributed_optimizer, (
```
deepakn94: Why does `has_layerwise_buffers=True` imply that `ddp_config.use_distributed_optimizer` needs to be True? Can't I just do gradient all-reductions?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -923,6 +934,10 @@ def __init__(
         self.ddp_config = ddp_config
         self.params = [param for (param, _) in params_with_names]
         self.param_indices = param_indices
+        # The BufferKey this buffer was created from, when known. Lets
+        # downstream code (e.g. layerwise/dist optimizer dispatch) ask which
```
deepakn94: "Lets downstream code determine the buffer_key without re-inspecting parameters."

megatron/core/optimizer/__init__.py
```diff
@@ -828,32 +830,220 @@ def _get_megatron_emerging_optimizer(
...
+
+    The two paths are combined with :class:`ChainedOptimizer`.
+    """
+    assert not config.overlap_param_gather_with_optimizer_step, (
+        "overlap_param_gather_with_optimizer_step is not supported with "
+        "use_layer_wise_distributed_optimizer=True."
+    )
+
+    # Partition param groups.
+    layerwise_groups: Dict[str, List[Dict]] = defaultdict(list)  # opt_name -> groups
```
deepakn94: We don't need separate groups for dense and expert layers?

megatron/core/optimizer/__init__.py
```diff
@@ -828,32 +830,220 @@ def _get_megatron_emerging_optimizer(
...
+
+
+def _build_distopt_optimizers_for_layerwise_adam_groups(
+    config: OptimizerConfig,
+    model_chunks: List[MegatronModule],
+    distopt_dense_groups: List[Dict],
+    distopt_expert_groups: List[Dict],
+    pg_collection: ProcessGroupCollection,
+) -> List[MegatronOptimizer]:
+    """Build :class:`DistributedOptimizer` instances for Adam-routed buckets.
```
deepakn94: How does this differ from existing code to create distopt_optimizers? Can we re-use that code?

megatron/core/distributed/distributed_data_parallel.py
deepakn94: Updated the assert message to spell this out. Short version: in principle yes, you can do gradient all-reductions, but only via the legacy ping-pong fallback path that doesn't use the layout — that path is still available when `full_param_layout` isn't supplied. The new layout-driven path consumed by `_shard_params_from_layout` derives ownership from buffer offsets and assumes reduce-scatter has populated each rank's shard, so it's specifically tied to `use_distributed_optimizer=True`. The assert documents the requirement; the all-reduce path is still reachable via the no-layout fallback.

megatron/core/distributed/param_and_grad_buffer.py
deepakn94: Done — applied the rewording verbatim.

megatron/core/optimizer/__init__.py
deepakn94: Asymmetric: LayerWise mixes dense + expert in one wrapper because it manages expert sharding internally; DistOpt does not, so its dense and expert groups go into separate `DistributedOptimizer`s with the right DP groups. Added a clarifying comment on the partitioning block.

megatron/core/optimizer/__init__.py
deepakn94: Extracted the duplicated process-group setup into `_setup_distributed_optimizer_process_groups()` and called it from both this layerwise path and `get_megatron_optimizer`'s standard path. Dedupes the ~10 lines of `setup_process_groups_for_optimizer` + `mp_rank` + `distributed_optimizer_instance_id` computation.

The remaining differences between the two paths are inherent:
- Param groups are pre-filtered here (we already have `all_param_groups` from `_get_param_groups`), versus the existing path's `_get_param_groups_and_buffers` which re-runs `_get_param_groups` and filters by `is_expert_parallel`.
- Buffers are filtered by `BufferKey.use_layerwise_distributed_optimizer` here, versus by attribute name (`buffers` vs `expert_parallel_buffers`) in the existing path.
- The layerwise path doesn't iterate the `overlap_param_gather_with_optimizer_step` chunk split (asserted out).

A larger dedup of the DistOpt-construction loop would require parameterizing the existing path on a buffer filter and on the chunk-split flag, which felt like a bigger change than this PR should swallow.

ko3n1g:
/ok to test [88655b2](https://github.com/NVIDIA/Megatron-LM/pull/4509/commits/88655b2d336aca3e92165d81f81ab4b149354370)

megatron/core/optimizer/layer_wise_optimizer.py
```diff
@@ -39,6 +43,279 @@ class LayerWiseDistributedOptimizer(ChainedOptimizer):
     6. allgather updated params to every rank
     """
 
+    @staticmethod
+    def _shard_divisor(data_parallel_world_size: int, ddp_config) -> int:
```
deepakn94, line 53: Can we not re-use the methods used in `DistributedOptimizer`?

megatron/core/optimizer/layer_wise_optimizer.py
```diff
@@ -39,6 +43,279 @@ class LayerWiseDistributedOptimizer(ChainedOptimizer):
...
+           insert padding of that numel.  Every shard grows by the same amount,
+           so all shards stay the same size.
+        5. When the bucket total reaches *bucket_size*, finalise the bucket
+           (pad shard size to :meth:`_shard_divisor`) and start a new one.
+        6. Repeat from 3 until all parameters are assigned.
+
+        Because repeated layers produce many parameters of the same shape,
+        size-matching naturally keeps whole parameters together without any
+        name-parsing heuristic.  For a model with *N* identical transformer
+        layers, padding overhead is zero.
```
deepakn94: "padding overhead is low (depending on number of layers and number of shards)"

FDecaYed (changes requested):
Could we make the use of layout default False before we can split large weights like embeddings out from layerwise? (if I'm understanding this correctly, currently it is True by default is muon is used)
Otherwise looks fine

FDecaYed (approved):

jaredcasper (approved):

yaoyu-33 (approved):

PR #4517, opened 2026-04-29
[build] fix: move mamba-ssm and causal-conv1d to optional [ssm] extra

ko3n1g:

## Problem

`mamba-ssm` and `causal-conv1d` are listed in the `[dev]` and `[lts]` optional extras. Both packages have no pre-built wheels for common CUDA configurations and must be compiled from source, taking ~10 minutes each. This makes any downstream package that depends on `megatron-core[dev]` — notably the Megatron-Bridge wheel build CI — routinely exceed k8s runner step limits, causing intermittent failures.

## Fix

Remove `mamba-ssm` and `causal-conv1d` from `[dev]` and `[lts]`, and expose them as a dedicated `[ssm]` extra. This is safe because both packages are imported inside `try/except ImportError` blocks in `mamba_mixer.py` — non-SSM workflows are unaffected.

**Usage for Mamba support:**
```bash
pip install megatron-core[dev,ssm]
```

</details>

### DIFF
--- a/docker/Dockerfile.ci.dev
+++ b/docker/Dockerfile.ci.dev
@@ -41,7 +41,7 @@ RUN --mount=type=cache,target=/root/.cache/uv \
     export NVTE_CUDA_ARCHS="80;90;100"
     uv venv ${UV_PROJECT_ENVIRONMENT} --system-site-packages
     uv sync --only-group build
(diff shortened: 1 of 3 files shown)


ko3n1g:
/ok to test [9525793](https://github.com/NVIDIA/Megatron-LM/pull/4517/commits/9525793923896f89225b1700179f6d3d35266a90)

chtruong814 (approved):

PR #3828, opened 2026-03-12
[dev] mHC kernel fusion 

jingqiny-99:
# What does this PR do ?

kernel fusion (using cuTile) for mHC.

Verified using 27B Model (introduced in original mHC paper) with GB200

Throughput: 
image

Validation LM Loss:
image

Iteration Time:
image

### Config: `s=8192, b=4, n=4, C=7168`

| Kernel | Framework | Fwd (ms) | Bwd (ms) | E2E (ms) | Speedup |
|--------|-----------|---------:|---------:|---------:|--------:|
| sinkhorn | compiled | 0.346 | 1.018 | 1.364 | 1.00x |
| sinkhorn | cutile | 0.065 | 0.133 | 0.198 | 6.89x |
| h_aggregate | compiled | 0.338 | 0.808 | 1.146 | 1.00x |
| h_aggregate | cutile | 0.349 | 0.661 | 1.010 | 1.13x |
| h_post_bda | compiled | 3.829 | 6.663 | 10.492 | 1.00x |
| h_post_bda | cutile | 0.684 | 2.549 | 3.234 | 3.24x |
| proj_rms | compiled | 0.554 | 2.204 | 2.758 | 1.00x |
| proj_rms | cutile | 0.426 | 1.541 | 1.967 | 1.40x |

## Prerequisites

Linux x86_64, Linux aarch64 or Windows x86_64

A GPU with compute capability 8.x 10.x, 11.x or 12.x

NVIDIA Driver r580 or later

Python version 3.10, 3.11, 3.12 or 3.13

## Usage

Install CuTile: `pip install cuda-tile[tileiras]`

Enable fusion kernel with `--use-fused-mhc`

### Pre-checks

- [x] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [x] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
[...]

PR #2943, opened 2026-01-14
[dev] feat(mHC): Add basic pytorch implementation of manifold hyper connection(mHC).

jingqiny-99:
# What does this PR do ?

Phase 1 of design proposal in #2919 

Main PR #3430 

Tested with Qwen3-30B-A3B, TP=1, PP=4 in 8x8xH100 Nodes. 

The loss curve below compares runs with and without mHC. The loss with mHC is slightly lower than without mHC, and the magnitude of the reduction is close to the results reported in the paper.

image

Figure 1: Two runs. 

loss_diff_mHC_vs_nonmHC

Figure 2: Our results. 

image

Figure 3: main results in mHC paper
:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

TODO:
- [ ] Move hyper connection expand to embedding layer?
- [x] Make discard and output trigger in TransformerBlock
- [x] Make mhc recompute logics in TransformerBlock in one place
- [x] Refactor random.py for new save/load logics

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
        C1[Expert Review] --> C2[Final Review]
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [ ] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

The following process is enforced via the CODEOWNERS file for changes into `megatron/core`. For changes outside of `megatron/core`, it is up to the PR author whether or not to tag the Final Reviewer team.

<details>
<summary>For MRs into `main` branch</summary>

Feel free to message or comment the @mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

#### (Step 2): Collect the expert reviewers reviews

1. Attach the `Expert Review` label when your PR is ready for review.
2. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

[...]

PR #4531, opened 2026-04-29
[Split 1/N of #3430] feat(mHC): basic pytorch implementation of manifold hyper connection

Connor-XY:
## What does this PR do?

This is the **first** of a sequence of split PRs derived from #3430
(mirror of upstream #2943 + kernel-fusion #3828). The goal of splitting is
to keep each PR small and limit the number of CODEOWNERS reviewer groups
required, making review faster and easier. Subsequent splits will be
opened **only after this one merges** — strictly one open at a time.

This PR contains **only** the core Manifold Hyper Connection (mHC) module
and its directly-supporting changes:

- `megatron/core/transformer/hyper_connection.py` (new mHC module)
- `megatron/core/transformer/transformer_block.py` (mHC integration into the block)
- `megatron/core/transformer/transformer_layer.py` (`HyperConnectionTransformerLayer` + recompute hooks)
- `megatron/core/transformer/transformer_config.py` (new mHC config fields)
- `megatron/core/transformer/__init__.py` (re-exports of mHC symbols)
- `megatron/core/transformer/cuda_graphs.py` (copyright-year bump only)
- `megatron/core/tensor_parallel/random.py` (`CheckpointWithoutOutput` / `CheckpointManager` refactor used by mHC recompute)
- `megatron/core/fusions/fused_bias_dropout.py` (extends `get_bias_dropout_add` with an optional `mhc_recompute_manager` path)

Reviewer groups touched: **core-adlr, core-nemo, transformer, cuda-graphs**.

### Deferred to follow-up split PRs (NOT in this PR)

- GPT model wiring — `gpt_layer_specs.py`, experimental specs, `gpt_builders.py`, `training/initialize.py`, model/fp8 tests
- Pipeline-parallel support — `pipeline_parallel/schedules.py`, `test_pp_mhc_compatibility.py`
- Fused mHC cuTile kernels — the `#3828` kernel side (`fused_mhc_kernels.py`, kernel-related fields in `transformer_config.py`, kernel-using paths in `hyper_connection.py`, `test_fused_mhc_kernels.py`)
- Functional-test recipe — `tests/functional_tests/test_cases/gpt/gpt3_mcore_te_tp2_pp2_mhc/*`, `tests/test_utils/recipes/h100/gpt.yaml`

## Unit-test coverage for components in this PR

Every component added or modified by this PR has a corresponding unit test
included in this same PR:

| Component | Unit test (in this PR) |
|---|---|
| `hyper_connection.py` (new module) | `tests/unit_tests/transformer/test_hyper_connection_recompute.py`, `tests/unit_tests/transformer/test_mhc_block_manager.py`, `tests/unit_tests/transformer/test_transformer_layer.py` (HyperConnectionTransformerLayer cases) |
| `transformer_layer.py` (mHC variant + recompute hooks) | `tests/unit_tests/transformer/test_transformer_layer.py` |
| `transformer_block.py` (mHC integration) | `tests/unit_tests/transformer/test_mhc_block_manager.py`, `tests/unit_tests/transformer/test_transformer_layer.py` |
| `transformer_config.py` (new mHC fields) | exercised by configs in `tests/unit_tests/transformer/test_transformer_layer.py` |
| `tensor_parallel/random.py` (`CheckpointWithoutOutput` + `ckpt_manager`) | `tests/unit_tests/transformer/test_hyper_connection_recompute.py` |
| `fusions/fused_bias_dropout.py` (`mhc_recompute_manager` path) | exercised indirectly via the mHC MLP/recompute path in `tests/unit_tests/transformer/test_hyper_connection_recompute.py` and `tests/unit_tests/transformer/test_transformer_layer.py` |
| `transformer/__init__.py` | export-only (no dedicated test — symbols are imported and exercised by the tests above) |
| `transformer/cuda_graphs.py` | copyright-year bump only (no test required) |

The functional test (golden-values + model_config) for the end-to-end mHC
configuration lives in the deferred functional-test split PR, since that
file is `@NVIDIA/ci`-owned and orthogonal to the transformer reviewers.

## Credit

All changes here are a **re-slice** of @jingqiny-99's work in #3430
(upstream basic pytorch implementation #2943). No new functionality has
been added, removed, or rewritten in this split — only repackaged for
narrower review scope.

## Pre-checks

- [x] Re-slice of an existing PR; no new feature
- [x] Original unit tests included for all components touched
- [ ] Functional test deferred to a later split PR (per split plan)
- [x] PR opened as **draft** per repo contributing guidelines


Connor-XY:
/ok to test [4a6e07c](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/4a6e07cc7172ef9be9da46f27b9d40e57799910d)

Connor-XY:
/ok to test [6ee487e](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/6ee487e452664681c407a33481b5b9517f265591)

Connor-XY:
/ok to test [0481591](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/048159121c3d4d80ed64224509a2464f55842e04)

Connor-XY:
/ok to test [54d47ce](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/54d47ce00f493f96a943d2642f26f2f5db857a2f)

Connor-XY:
/ok to test [1577a19](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/1577a1974b4e325c94077a932487a247013dac01)

megatron/core/tensor_parallel/random.py
```diff
@@ -685,10 +747,56 @@ def backward(ctx, *args):
...
+        grads = tuple(inp.grad if isinstance(inp, torch.Tensor) else None for inp in inputs)
         return (None, None) + grads
 
 
+class CheckpointManager:
+    """
+    Manages multiple CheckpointWithoutOutput objects within a TransformerBlock
+    cross layer recomputations, enabling unified recomputation during backward pass.
+    This is particularly useful for scenarios where multiple checkpoint operations have
+    sequential dependencies (i.e., the output of one checkpoint is the input of the next).
```
mathemakitten, line 802:
```suggestion
class CheckpointManager:
    """
    Coordinates activation recomputation across multiple CheckpointWithoutOutput instances 
    within a TransformerBlock, enabling unified recomputation during backward pass.
    This is particularly useful for scenarios where multiple checkpoint operations have
    sequential dependencies (i.e., the output of one checkpoint is the input of the next).
```
Would it be reasonable to name this ActivationRecomputeManager or CheckpointWithoutOutputManager or similar? while "checkpoint manager" makes sense in the context of this file, across the codebase, "checkpoint" is overloaded.

megatron/core/transformer/hyper_connection.py
```diff
@@ -0,0 +1,696 @@
...
+            # Compute dL/dM_input (i.e., dL/dM_init) via autograd
+            (grad_M_init,) = torch.autograd.grad(
+                outputs=M_current,
+                inputs=M_input,
+                grad_outputs=grad_output,
+                create_graph=False,
+                retain_graph=False,
+            )
+        # Apply chain rule: dL/dH = dL/dM_init * dM_init/dH = dL/dM_init * M_init
+        # Since M_init = exp(H_res_logits), we have d(exp(x))/dx = exp(x) = M_init
```
mathemakitten, line 143: So directly in `forward` above you have `M_init = torch.exp(H_res_logits - H_res_logits.max(dim=-1, keepdim=True).values)`. Is the intention here to simplify based on the fact that the first iteration of Sinkhorn zeroes out any row scaling of the input? If so, can you document this clearly? As a reader it was not obvious to me why we were able to drop the max (i.e. it's basically a constant / Sinkhorn output is invariant to row-scaling of the input and does not contribute to grad).

megatron/core/transformer/transformer_config.py
```diff
@@ -1458,6 +1492,50 @@ def __post_init__(self):
...
+        ):
+            warnings.warn(
+                "HyperConnections are enabled but 'mhc' is not in "
+                "recompute_modules with selective recompute. Consider adding 'mhc' to "
+                "recompute_modules with selective recompute to reduce activation memory."
+            )
+
+        # Validation for hyper_connections with MTP
+        if self.enable_hyper_connections and self.mtp_num_layers is not None:
+            raise ValueError(
```
mathemakitten, line 1994: This should be a NotImplementedError, right?

megatron/core/transformer/transformer_layer.py
```diff
@@ -700,6 +704,11 @@ def forward(self, *args, **kwargs):
         This method calls the core computation of a transformer layer, including
         self-attention, cross-attention (if applicable), and feed-forward operations.
         """
+        # Injected by __call__ for cuda graph keying; not a real forward arg.
+        kwargs.pop("dynamic_inference_decode_only", None)
+        assert (
+            not self.config.enable_hyper_connections
+        ), "Please use HyperConnectionTransformerLayer instead"
```
mathemakitten: Recommend to move this out of forward() and into module instantiation to fail early and avoid calling on every forward.

megatron/core/transformer/transformer_layer.py
```diff
@@ -1280,6 +1289,33 @@ def _should_call_local_cudagraph(self, *args, **kwargs):
...
+        cg_index = microbatch_idx % len(self.cuda_graphs)
+        if not hasattr(self.cuda_graphs[cg_index], 'backward_dw'):
+            return
+        self.cuda_graphs[cg_index].backward_dw()
+
+    def __call__(self, *args, **kwargs):
+        # Extract mhc_recompute_manager before CUDA graph manager processes kwargs,
+        # since CheckpointManager is not a CUDA-graph-supported type.
+        self._mhc_recompute_manager = kwargs.pop("mhc_recompute_manager", None)
+        kwargs.pop("is_last_layer_in_recompute_block", None)
```
mathemakitten: Right now this is on `TransformerLayer`, which is used everywhere. `self._mhc_recompute_manager` is now an attribute on every base TransformerLayer instance even when mHC is off, which is most of the time. Can we scope this down to HyperConnectionTransformerLayer only?

mathemakitten: Also, is `is_last_layer_in_recompute_block` actually passed as a kwarg still? It looks like it has been changed to be a class attr passed around?

megatron/core/transformer/transformer_layer.py
```diff
@@ -1280,6 +1289,33 @@ def _should_call_local_cudagraph(self, *args, **kwargs):
...
+
+        if self._should_call_local_cudagraph(*args, **kwargs):
```
mathemakitten: Why is this newly needed? `TransformerLayer` was running without his before, and it was only necessary at the block level, not the layer level.

megatron/core/transformer/transformer_layer.py
```diff
@@ -1289,6 +1325,373 @@ def get_layer_norm_weights(self):
...
+        packed_seq_params: Optional[PackedSeqParams] = None,
+        sequence_len_offset: Optional[Tensor] = None,
+        padding_mask: Optional[Tensor] = None,
+        mhc_recompute_manager: Optional['CheckpointManager'] = None,
+        *,
+        inference_params: Optional[Any] = None,
+    ):
+        """Forward attention with hyper connection pre/post processing on self-attention."""
+        from megatron.core.pipeline_parallel.fine_grained_activation_offload import (
+            FineGrainedActivationOffloadingInterface as off_interface,
```
mathemakitten:
```suggestion
            FineGrainedActivationOffloadingInterface as offload_interface,
```

megatron/core/transformer/transformer_layer.py
```diff
@@ -1289,6 +1325,373 @@ def get_layer_norm_weights(self):
         return
 
 
+class HyperConnectionTransformerLayer(TransformerLayer):
```
mathemakitten, line 1631: This currently works for dense FFN + mHC but does not compose with MoE. What is the extensibility plan for MoE, to also later inherit from MoETransformerLayer? (I assume this is for DSv4).

megatron/core/transformer/transformer_layer.py
```diff
@@ -1289,6 +1325,373 @@ def get_layer_norm_weights(self):
...
+            context = attention_output_with_bias["context"]
+
+        with self.bias_dropout_add_exec_handler():
+            hidden_states = self.cross_attn_bda(self.training, self.config.bias_dropout_fusion)(
+                attention_output_with_bias, residual, self.hidden_dropout
+            )
+
+        return hidden_states, context
+
+    def _forward_mlp(
```
mathemakitten: This is mostly a duplication of TransformerLayer._forward_mlp. Is it possible to make mHC. awrapper around an inner layer's _forward_mlp instead? I am afraid that we will quickly incur drift between the two / it becomes a maintenance burden.

megatron/core/transformer/transformer_layer.py
```diff
@@ -1289,6 +1325,373 @@ def get_layer_norm_weights(self):
...
+        """
+        submodules = super()._get_submodules_under_cudagraphs()
+
+        if not self.config.cuda_graph_scope:
+            return submodules
+
+        if CudaGraphScope.attn in self.config.cuda_graph_scope:
+            submodules.append(self.self_attention_hyper_connection)
+        if (not self.is_moe_layer and CudaGraphScope.mlp in self.config.cuda_graph_scope) or (
+            self.is_moe_layer and CudaGraphScope.moe in self.config.cuda_graph_scope
```
mathemakitten: Is this a preemptive check or actually needed? HyperConnectionTransformerLayer extends  TransformerLayer, not MoETransformerLayer.

megatron/core/transformer/transformer_layer.py
```diff
@@ -1280,6 +1289,33 @@ def _should_call_local_cudagraph(self, *args, **kwargs):
                 return True
         return False
 
+    def backward_dw_cudagraph(self, microbatch_idx):
```
mathemakitten: same question as below: why is this needed on base TransformerLayer? I don't see it referenced or called anywhere else?

megatron/core/tensor_parallel/random.py
Connor-XY: Renamed to `CheckpointWithoutOutputManager` and updated the docstring along the lines you suggested. The class strictly manages `CheckpointWithoutOutput` instances, so the new name mirrors what it coordinates and avoids the broader "checkpoint" overloading.

megatron/core/transformer/hyper_connection.py
Connor-XY: Good catch. I replaced the comment with a sharper justification: the per-row max shift is invariant under Sinkhorn-Knopp normalization (the first row-normalization `T_r` divides each row by its sum, which cancels any per-row scalar), so `exp(H_ij - c_i)` and `exp(H_ij)` reach the same fixed point and gradient. The shift only changes the numerical stability of the `exp`.

megatron/core/transformer/transformer_config.py
Connor-XY: switched the `fine_grained_activation_offloading` incompatibility to `NotImplementedError`. The error message describes a known unimplemented interaction (the mHC recompute hook fires before the offloading backward chunk is initialized), so the new class matches intent.

megatron/core/transformer/transformer_layer.py
Connor-XY: Done. `_mhc_recompute_manager` is now declared in `HyperConnectionTransformerLayer.__init__` (defaults to `None`), so `forward` reads it directly off `self` without a `getattr` fallback. The `__call__` itself is also scoped to the HC subclass now.

megatron/core/transformer/transformer_layer.py
Connor-XY: Agreed. I moved everything mHC-specific out of base `TransformerLayer`:

- The new `__call__` override is dropped from base `TransformerLayer` entirely. `HyperConnectionTransformerLayer` now defines its own `__call__` that pops `mhc_recompute_manager` / `is_last_layer_in_recompute_block` and chains to `super().__call__`.
- `_mhc_recompute_manager` is initialized in `HyperConnectionTransformerLayer.__init__` only. Base TransformerLayer never sees it.
- The `dynamic_inference_decode_only` injection is also dropped: `cuda_graphs.py` reads `inference_context.is_decode_only()` directly, so the kwarg injection was redundant.

Connor-XY: Removed the pop:
`is_last_layer_in_recompute_block` is not a kwarg in the current code path. It's set as a dynamic attribute on the manager object in `transformer_block.py` before the layer is invoked.

megatron/core/transformer/transformer_layer.py
Connor-XY: Removed from this PR. `backward_dw_cudagraph` was added on base `TransformerLayer` but had no caller in this split. Will reintroduce in that split where it's actually needed.

megatron/core/transformer/transformer_layer.py
Connor-XY: Renamed `off_interface` → `offload_interface`

megatron/core/transformer/transformer_layer.py
Connor-XY: The split scope and design intent here are deliberately to *not* compose mHC with MoE inside a single `TransformerLayer`. Two reasons:

1. The mHC algorithm wraps the MLP submodule in a fused `apply_h_res + apply_h_post + bias-dropout-add` kernel. Plumbing that through MoE's router + dispatcher + shared-expert paths inside one layer would couple two orthogonal feature ladders and is not part of the original #3430 PR being re-sliced.
2. We have a clean composition path already: wrap the MoE layer inside a `HybridStack` layer via `HyperConnectionHybridLayer` (in the deferred follow-up split). The mHC pre/post happens at the hybrid-layer boundary, with MoE staying a vanilla `MoETransformerLayer` underneath. DSv4 is expected to ride that path.

A native `HyperConnectionMoETransformerLayer` (inheriting from `MoETransformerLayer` and replacing the post-MLP path) is plausible if a future workload needs it, but it's not on the roadmap for this re-slice.

To make the intent explicit at code level, this PR now raises `NotImplementedError` from `HyperConnectionTransformerLayer.__init__` when `submodules.mlp` resolves to an MoE layer, with a message pointing at the `HyperConnectionHybridLayer` path. (a09ffbd52)

megatron/core/transformer/transformer_layer.py
Connor-XY: Refactored. Extracted a `_run_mlp(self, pre_mlp_layernorm_output, residual, padding_mask, inference_context)` helper on `TransformerLayer` that holds the recompute / chunked-prefill / fp8-fp4 / plain-MLP branching in one place. Both `TransformerLayer._forward_mlp` and `HyperConnectionTransformerLayer._forward_mlp` now call it — eliminates ~80 lines of duplication.

I also looked at the lighter "HC wraps `super()._forward_mlp`" shape but couldn't make it work: base `_forward_mlp` always finishes with `_forward_post_mlp`, which consumes the `mlp_output_with_bias` tuple and the residual to produce the bda'd output. HC needs both those values intact to drive its `fused_h_res_h_post_bda` kernel, so we need the cut point *before* `_forward_post_mlp`. The helper exposes that cut.

megatron/core/transformer/transformer_layer.py
Connor-XY: You're right that the runtime branch (`is_moe_layer` clause inside `_get_submodules_under_cudagraphs`) was dead — HC doesn't accept MoE submodules. Two changes:

1. Added an explicit `NotImplementedError` raise in `HyperConnectionTransformerLayer.__init__` when `self.is_moe_layer` is `True` after `super().__init__(...)`. The message points at the `HyperConnectionHybridLayer` path so a misconfigured spec fails fast and clearly at setup.
2. Simplified `_get_submodules_under_cudagraphs` to drop the dead MoE branch.

megatron/core/transformer/transformer_layer.py
Connor-XY: `backward_dw_cudagraph` is removed from base `TransformerLayer` in this PR. It had no caller here; its consumer lives in the deferred PP-schedules split, where it'll be reintroduced.

megatron/core/transformer/transformer_layer.py
Connor-XY: Good catch on the follow-up — confirmed there's no caller anywhere that passes `is_last_layer_in_recompute_block` as a kwarg, so the pop is genuinely dead. Removed in 6d3ac1749.

megatron/core/transformer/transformer_layer.py
mathemakitten: Thank you, looks great. Can we do something similar for `_forward_attention`?

Connor-XY: I split the _forward_attention of base TransformerLayer into three functions. HC overrides two of the three funcs (input-layernorm + bda-step), and inherits cross-attention. HC-specific (h_res, h_post) flow between the two overrides as an opaque attn_state tuple.

Connor-XY:
/ok to test [75588dc](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/75588dcccbc5c3d3c2f0bbbebb08622357c8bedb)

Victarry:
@Connor-XY Thanks for the PR. Do we have any training loss curve with and without mHC for comparison?

megatron/core/transformer/transformer_layer.py
mathemakitten: This is great! Actually, can the `_forward_mlp` function be modularized the same way as `_forward_attention`? e.g.

```
pre_mlp_state = self._run_pre_mlp_layernorm(hidden_states)         
mlp_output_with_bias = self._run_mlp(...)
return self._apply_mlp_bda_step(mlp_output_with_bias, residual, pre_mlp_state)
```

Then `_forward_post_mlp_with_fused_hyper_connection` becomes just an `_apply_mlp_bda_step` override.

Connor-XY:
/ok to test [389e9e2](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/389e9e20f0d11ce76d1dd0eebfbf81e5e0d7e588)

megatron/core/transformer/transformer_layer.py
Connor-XY: I've mirrored the _forward_attention refactor for the MLP path.

Connor-XY:
/ok to test [cd2abae](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/cd2abaed6db1b173ea5b48ae97fae52c5cd2d24a)

mathemakitten (approved):

Connor-XY:
> @Connor-XY Thanks for the PR. Do we have any training loss curve with and without mHC for comparison?

  Hi @Victarry — thanks for the question. Quick summary:

  Upstream evidence (canonical): the original Qwen3-30B-A3B with-vs-without-mHC loss curves are on the upstream
  PyTorch reference implementation in #2943 — those are the convergence-validation figures that should be cited as
  the algorithmic story.

  Why this PR doesn't have an end-to-end loss curve of its own: #4531 is split 1/N of #3430 (the integration of
  #2943 into core). The goal of splitting was to keep each PR small and limit the number of CODEOWNERS reviewer
  groups required, so this PR contains only the core mHC module and the transformer-layer / transformer-block
  integration. Several pieces needed to actually train end-to-end with mHC are intentionally deferred to follow-up
  splits (per the PR description):

  - gpt_layer_specs.py — the enable_hyper_connection flag plumbing that selects HyperConnectionTransformerLayer
  instead of plain TransformerLayer. Without this, pretrain_gpt.py can't construct an mHC model.
  - Pipeline-parallel mHC compatibility (pipeline_parallel/schedules.py).
  - Fused mHC cuTile kernels (#3828).
  - Functional-test recipe (gpt3_mcore_te_tp2_pp2_mhc).

  What I ran instead (sanity check): a small side-by-side at the parent integration branch where the full feature
  already lives — yxu1/mhc-hybridmodel-dsv4, which is the branch this split was sliced from. Tiny GPT-3 (12 layers,
   hidden=512, TP=2 PP=2) on 4 H100s for 500 iterations, deterministic mode, the_pile data, identical args with
  only --enable-hyper-connections toggled.

  Result (lm loss on the training set):
  
image

  Caveat: the run above used the integration branch (which has the full implementation including post-review
  numerical fixes like the fp32 Sinkhorn cast). The split itself was sliced before some of those numerical fixes
  landed; pulling them forward into a follow-up commit on this PR is on my list.

  Let me know if you'd like a longer or larger-config run as additional evidence.

Connor-XY:
/ok to test [16c60e0](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/16c60e02bcb6f1155e4e186f760b478efd87bc47)

Connor-XY:
/ok to test [8e1bde4](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/8e1bde49ba60eaabe03274948173db53a6fa7cb7)

Connor-XY:
Hi @yaoyu-33, Heads-up that this PR has a downstream-affecting rename worth your attention.

  **What changed:** `TransformerLayer._forward_post_mlp` was renamed to `_apply_mlp_bda_step` and gained a third parameter `mlp_state: tuple = ()` since we need to support mHC.

  **Why the rename:** review feedback on this PR asked us to dedup `HyperConnectionTransformerLayer._forward_mlp`, which was a near-copy of the base. We extracted the MLP path into hooks `_run_pre_mlp_layernorm` → `_run_mlp` → `_apply_mlp_bda_step`, mirroring the attention-side hooks (`_run_input_layernorm` → `self_attention` → `_apply_self_attn_bda_step`) introduced in the same review pass. The names now match symmetrically: a `_run_*_layernorm` hook returns an opaque `state` slot (`attn_state` / `mlp_state`) that the corresponding `_apply_*_bda_step` hook consumes — so subclasses can thread extra intermediates between the two halves with the same pattern on both sides. `_forward_post_mlp` was retired in favor of the symmetric name.

  **Why this matters for Bridge:** `Gemma4TransformerLayer` in
  `Megatron-Bridge/src/megatron/bridge/models/gemma/gemma4_provider.py:217` (commit [`2faedbf6fe`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/commit/2faedbf6fe3c422835a44b2b360cadcb2a116a54)) overrides `_forward_post_mlp` to apply Gemma-4-specific math. After Bridge bumps its mcore submodule past this PR, the override is orphaned: nothing calls `_forward_post_mlp` anymore, the inherited base `_apply_mlp_bda_step` runs, and the Gemma-specific work is silently skipped. 

  **Mitigation already in this PR:** I just pushed commit `d50caaf0` (`refactor(mHC): add back-compat shim for renamed _forward_post_mlp hook`). The shim detects a legacy `_forward_post_mlp` override on any subclass and routes through it with a `DeprecationWarning`, so existing Bridge code keeps producing correct numerics during the migration window. The shim will be removed in a future release.

  **Bridge-side fix when convenient:** 1. Rename the override `_forward_post_mlp` → `_apply_mlp_bda_step`. 2. Add the third parameter, accepted but unused for Gemma-4: `def _apply_mlp_bda_step(self, mlp_output_with_bias, residual, mlp_state=()):`

Would you have a few minutes to review this PR? Particularly happy to hear if you spot any other Bridge subclass that's affected. Thanks!

Connor-XY:
/ok to test [c0ced14](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/c0ced146b6a927c36487907eeff05d93f1456a77)

Phlip79 (approved):

Connor-XY:
/ok to test [20dbf8e](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/20dbf8e9095f9e50260e643366dd825e1b0c71f4)

Connor-XY:
/ok to test [461af34](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/461af3466251adc5ad5d4967451645de05f4e590)

ericharper (approved):

Connor-XY:
/ok to test [27fef06](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/27fef06db041bb19039ec7f0eec9cec3399cadc6)

Connor-XY:
/ok to test [d59082b](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/d59082b728cc6080c56a8b45d2a26eebb7ec2023)

Connor-XY:
/ok to test [f1df30c](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/f1df30c1db8c2ece4df1e0465b89f67f12337e7b)

megatron/core/transformer/transformer_block.py
```diff
@@ -619,6 +670,18 @@ def forward(
             use_inner_quantization_context = False
             outer_quantization_context = nullcontext()
 
+        # Determine if MHC recompute should be used
+        # Only enable when: training mode AND hyper connections AND 'mhc' in recompute_modules
+        use_mhc_recompute = (
```
jiemingz: could this moved to init?

megatron/core/transformer/transformer_config.py
```diff
@@ -1064,6 +1068,35 @@ class TransformerConfig(ModelParallelConfig):
     CudaGraphScope instances deserialized from pre-refactor checkpoints are converted to their
     string names before normalization so existing CUDA_GRAPH_MODULES_DEPRECATIONS handles them."""
 
+    ####################
+    # Hyper-Connection Configuration
+    ####################
```
jiemingz, line 1117: can we try to use a consistent naming and vocabulary here? like `enable_mhc_connections` , `mhc_num_residual_streams`

megatron/core/transformer/transformer_block.py
```diff
@@ -483,6 +487,46 @@ def __call__(self, *args, **kwargs):
             return super().__call__(*args, **kwargs)[0]
         return super().__call__(*args, **kwargs)
 
+    def _build_mhc_recompute_layer_plan(
```
jiemingz: can we move as much of this into hyper_connection.py? it looks like this function doesnt touch much transformer_block internals

megatron/core/transformer/transformer_layer.py
```diff
@@ -799,6 +841,35 @@ def _maybe_reflatten_from_moe(self, output, packed_seq_params, mbs):
             return output
         return output.transpose(0, 1).reshape(mbs * packed_seq_params.tokens_per_sample, 1, -1)
 
+    def _run_pre_mlp_layernorm(self, hidden_states):
```
jiemingz: can we use `_run_pre_mlp_layernorm` in MoETransformerLayer as well? to reduce duplicating code

jiemingz: `_run_pre_mlp_layernorm` and `_forward_pre_mlp_layernorm` sound too confusing and similar, maybe `_pre_mlp_layernorm_and_residual` ?

Connor-XY:
/ok to test [2cfc2c9](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/2cfc2c905ca5d47f2fe058d70128166d645cc3af)

Connor-XY:
/ok to test [07ddaf3](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/07ddaf312c80f0a15cce901cecee5de2a78c254c)

FDecaYed (approved):

jiemingz (approved):

Connor-XY:
/ok to test [cc80c69](https://github.com/NVIDIA/Megatron-LM/pull/4531/commits/cc80c69ab53298817de59ade9e7c1dff9a1348ba)

PR #4574, opened 2026-05-01
refactor(skills): add when_to_use frontmatter, split ci-test-system, enforce skill workflow

ko3n1g:

Ports the learnings from [NVIDIA-NeMo/Megatron-Bridge#3621](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3621) to Megatron-LM.

## Changes

**Add `when_to_use` field to all skill frontmatters** — replaces non-standard `TRIGGER when:` / `DO NOT TRIGGER when:` fields with the canonical Anthropic spec field. Strips embedded trigger text from `description:` fields and moves it to `when_to_use:` trigger phrases.

**Add mandatory 3-step skill workflow to AGENTS.md** — makes skill invocation structural: read artifact → invoke skill → answer.

**Add `.claude/settings.json` with UserPromptSubmit hook** — enforces the mandatory read→skill→answer workflow on every prompt.

**Add `Read` tool to `claude_review.yml` and inject skill-loading instruction** — both light and strict review prompts now instruct Claude to read relevant SKILL.md files before reviewing.

**Split `ci-test-system` into `testing` and `cicd`** — `testing` covers test layout, recipe YAML, adding tests, and golden values; `cicd` covers pipeline structure, scope labels, internal CI triggering, and failure investigation.

**Add `linting-and-formatting` skill** — extracted from `build-and-dependency`: `tools/autoformat.sh`, ruff/black/isort/pylint/mypy, import ordering, code style rules.

**Update `build-and-dependency`** — removes linting section (now in `linting-and-formatting`), tightens description and `when_to_use`.

</details>


chtruong814 (approved):

AGENTS.md
```diff
@@ -1,5 +1,23 @@
...
+
+**Workflow — mandatory order for every task:**
+1. **Pull information first.** Read the commit, PR, error log, file, or
+   whatever artifact the task is about. Do not reason about it yet.
+2. **Select and invoke the skill.** Based on what you just read, identify
+   the relevant skill and invoke it before forming any answer or plan.
+3. **Answer or implement.** Only after the skill is loaded, use its context
+   to reason, diagnose, or write code.
+
+Never skip or reorder these steps. Do not wait for the user to name the right
```
Phlip79, line 18: Not sure I agree with this. Claude/Codex productivity and performance goes down as context gets larger. My concern with this is the context will get "polluted" without the user's approval.

ko3n1g: Well but this is relevant context, so this is the part we actually want. For instance, when we submit a short prompt “upgrade dependencies” it often bypasses skills and then end up discovering for itself that it needs to run inside a container. These settings here enforce to load the relevant skill into context so that the resolution path becomes more linear

Phlip79: I would prefer forcing the user to be explicit with skill usage: to prevent Claude from adding a skill to its context when it shouldn't. For example, let's say a user has a PR and wants to create a new, basic unit test: Claude might load the cicd skill, even though it's not necessary.

ko3n1g: I think if this happens, we need to revisit the trigger_when keywords to ensure this doesnt happen. in practice, it also doesnt happen so far.  

Screenshot 2026-05-01 at 23 19 05

ericharper: How much of this is necessary vs what's baked into claude code and codex? https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview#how-skills-work

ericharper: For me the main idea is progressive disclosure, we don't want agents loading too much into context.

ko3n1g: this PR is implementing anthropics schema:) I've updated the front-matter so that the description and when_to_invoke fields are matching the content.

However, i've experienced a few times that for very short prompts, claude feels overconfident and ignores looking for skills. It reliably works when we elaborate the task in the prompt more extensively.

But for short prompt like: "update my lockfile" it skips the skills since it believes it knows that to do. our skills however share Nemo/Megatron-specific information, like "uv lock needs to run inside a container". 

So the `.claude/settings.json` enforces to always go via skills, even for claude believes it might not be necessary. I think its the right thing to do, since our skills always have more specific knowledge around our own stack than claude could have.

ko3n1g: This is also relevant for headless agents: Its not efficient to always instruct the agent which skill to open. Users might even not know which skills exist, shouldn't need to know, respectively. We want the agent to be able to pick up skills reliably to find a way to solution in the most linear way

ko3n1g: And to clarify: It doesn't load _all_ skills into context. Only the one it believes is relevant for solving the task. That's the beautiful thing about skills, its token-efficient

ericharper: Would be great to give your feedback to claude code devs re: 

But for short prompt like: "update my lockfile" it skips the skills since it believes it knows that to do. our skills however share Nemo/Megatron-specific information, like "uv lock needs to run inside a container".

That seems like a bug?

ko3n1g: Yeah to me it feels like a bug. When asking why it skipped reading the skills it responds something along the lines of “oh yeah, you’re right, I totally overlooked the skills”. Kind of as skill discovery requires a minimum number of tokens.

PR #4604, opened 2026-05-04
Enable async scheduling for decode-only inference

lmcafee-nvidia:
## Summary
This PR adds opt-in decode-only async scheduling for dynamic inference. During eligible decode-only steps, the controller can accept sampled tokens from the previous step, prepare next-step metadata, and launch or reuse a speculative forward so CPU bookkeeping can run while GPU work is already queued. Async scheduling remains disabled by default and is gated on supported decode-only shapes and features.

The implementation shares decode metadata preparation between the synchronous and async paths, adds explicit eligibility and diagnostic tracking, and uses full-iteration CUDA graphs for the optimized async path. It starts with the GPT decode path, then extends the same scheduling machinery to hybrid, MoE, expert-parallel, and MTP decode-only workloads, including row-mapped pending-forward reuse and graph-backed hidden-state handling for hybrid MTP.

The scheduler now reconciles lifecycle boundaries instead of invalidating an entire speculative step. Finished rows can be trimmed, new admissions can be deferred until pending async work is resolved, KV block reservations and releases are managed one step ahead, and pause, evict, and finish transitions keep request state and block ownership consistent.

The branch also broadens feature compatibility for production decode requests. Async decode now supports generated logprobs and top-n logprobs, prompt logprob interactions, stop-word termination, top-k and top-p sampling, stable per-request RNG ownership across accept/reject, chunked/prefix-cache decode-only eligibility, logging-step barriers, and server compatibility when forced logprobs are disabled.

For expert-parallel server/coordinator runs, this PR adds a tagged EP async step protocol that keeps real and dummy ranks in the same step phases. The protocol coordinates work consensus, pending-forward reuse/discard, graph-shape sync, async handoff launch/skip, and step completion, then sends coordinator replies after EP completion to avoid collective-ordering stalls.

## Tests
- Focused 8-GPU unit tests for async scheduling, dynamic engine/context behavior, controller speculative sampling, logprobs, and EP coordinator paths.
- `bash tools/autoformat.sh` and `git diff --check`.
- Nano-v3 MTP 300M async-scheduling smoke benchmark.


megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -2001,147 +2023,112 @@ def add_dummy_requests_for_expert_parallel_step(
...
-        best_graph = CUDAGraphBatchDimensionBuilder.match_graph_config(
+    def _match_cuda_graph_batch_dimensions(
+        self, batch_dimensions: InferenceBatchDimensions, *, strict: bool
+    ) -> Optional[InferenceBatchDimensions]:
+        """Return the cuda graph batch dimensions for ``batch_dimensions`` if one matches."""
+        return CUDAGraphBatchDimensionBuilder.match_graph_config(
             batch_dimensions,
             self.cuda_graph_batch_dimensions_list,
-            strict=self.is_hybrid_model,
+            smallest_non_decode_cuda_graph_size=self.smallest_non_decode_cuda_graph_size,
```
sidsingh-nvidia, line 2368: Why are these arguments coming back? The match_graph_config function was simplified when the NVLS dispatcher was merged in.

sidsingh-nvidia (changes requested):

PR #4623, opened 2026-05-05
Fix Megatron-FSDP optimizer CPU offload and checkpointing

wplf:
## Summary
- Handle Megatron-FSDP DTensor parameters and gradients by operating on local shards before optimizer CPU offload copies.
- Save `HybridDeviceOptimizer` state for `fsdp_dtensor` checkpoints as deterministic FSDP DTensors, skipping duplicate `master_param` entries and attaching local DCP chunk metadata without metadata collectives.
- Allow SWiGLU/GDN optimizer checkpoint preprocessing to split flat optimizer DTensors using the corresponding model parameter metadata, and restore loaded DTensor optimizer state back to local tensors.

Fixes https://github.com/NVIDIA/Megatron-LM/issues/4910.

## Tests
- `uv run isort megatron/core/optimizer/distrib_optimizer.py`
- `python -m py_compile megatron/core/optimizer/distrib_optimizer.py megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py megatron/core/transformer/fsdp_dtensor_checkpoint.py`
- `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. uv run python -m pytest tests/unit_tests/test_optimizer_cpu_offloading.py -q` (`72 passed`)
- Local Qwen3.5-VL proxy repro with `optimizer_cpu_offload=True`, `optimizer_offload_fraction=1`, `overlap_cpu_optimizer_d2h_h2d=True`, `use_precision_aware_optimizer=True`, and `fsdp_dtensor` optimizer checkpointing: checkpoint saves at iterations 10 and 12 completed successfully.

### DIFF
--- a/megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py
+++ b/megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py
@@ -4,6 +4,20 @@ from typing import Dict
 
 import torch
 
+try:
+    from torch.distributed._tensor import DTensor
+
+    HAVE_DTENSOR = True
+except ImportError:
+    HAVE_DTENSOR = False
+    DTensor = None
+
+
+def _to_local_if_dtensor(tensor):
+    if HAVE_DTENSOR and isinstance(tensor, DTensor):
+        return tensor.to_local()
+    return tensor
+
 
 def _param_generator(cpu_optimizer):
     for group in cpu_optimizer.param_groups:
@@ -90,6 +104,7 @@ class HybridDeviceOptimizer(torch.optim.Optimizer):
                 fp32_param = self.param_to_fp32_param[param]
                 grad = getattr(param, "decoupled_grad", param.grad)
                 if grad is not None:
+                    grad = _to_local_if_dtensor(grad)
                     fp32_param.grad = grad.to(fp32_param.dtype)
                     fp32_param.requires_grad = True
                 else:
@@ -105,6 +120,7 @@ class HybridDeviceOptimizer(torch.optim.Optimizer):
                     continue
 
                 param.requires_grad = False
+                grad = _to_local_if_dtensor(grad)
                 if param not in self.cpu_copy_map_grad:
                     self.cpu_copy_map_grad[param] = torch.empty(
                         param.shape, dtype=param.dtype, pin_memory=self.pin_cpu_grads, device="cpu"
@@ -121,6 +137,7 @@ class HybridDeviceOptimizer(torch.optim.Optimizer):
                 with torch.cuda.stream(self._h2d_stream):
                     for param in _param_generator(optimizer):
                         gpu_param = self.cpu_copys_map_gpu_param[param]
+                        gpu_param = _to_local_if_dtensor(gpu_param)
                         gpu_param.data.copy_(param.data, non_blocking=True)
                 self._h2d_stream.record_event().wait(torch.cuda.current_stream())
 
@@ -137,6 +154,7 @@ class HybridDeviceOptimizer(torch.optim.Optimizer):
 
                         if param in self.param_to_fp32_param:
                             fp32_param = self.param_to_fp32_param[param]
+                            param = _to_local_if_dtensor(param)
                             param.data.copy_(fp32_param.data)
 
             return fp32_param_copy_back_gpu_hook
@@ -252,8 +270,14 @@ class HybridDeviceOptimizer(torch.optim.Optimizer):
         params = []
         for group in self.param_groups:
             params.extend(group["params"])
-        params_total_numel = sum([param.numel() for param in params])
-        gpu_params_total_numel = sum([param.numel() for param in params if param.is_cuda])
+        params_total_numel = sum([_to_local_if_dtensor(param).numel() for param in params])
+        gpu_params_total_numel = sum(
+            [
+                _to_local_if_dtensor(param).numel()
+                for param in params
+                if _to_local_if_dtensor(param).is_cuda
+            ]
+        )
         cpu_params_total_numel = params_total_numel - gpu_params_total_numel
         offload_threshold = gpu_params_total_numel * offload_fraction
         offload_params_numel = 0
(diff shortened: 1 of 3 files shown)


wplf:
/ok to test [5da7d67](https://github.com/NVIDIA/Megatron-LM/pull/4623/commits/5da7d67fdfd63544368a20fdd9074f5fabc5526d)

megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py
```diff
@@ -4,6 +4,20 @@
...
+
+    HAVE_DTENSOR = True
+except ImportError:
+    HAVE_DTENSOR = False
+    DTensor = None
+
+
+def _to_local_if_dtensor(tensor):
+    if HAVE_DTENSOR and isinstance(tensor, DTensor):
+        return tensor.to_local()
```
cspades: BTW, if we don't need differentiability (such as in this optimizer step), sometimes `dtensor._local_tensor` is more robust since we can be sure we do not receive a copy of the original local Tensor. (Don't ask me why I know this.)

cspades (commented):
@wplf Could you share some information on what the bug was? Curious what was the root cause and how it is related to DTensors!

megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py
wplf: Hi, @cspades .

Currently, when using `megatron-fsdp` and `--optimizer-cpu-offload`, it will crash. Root cause is hybrid-optimizer does not take mfsdp into consideration. Use Dtensor._local_tensor will resolve this issue.

But I haven't check this fix thoroughly for training convergence. 

I'll check training stability and convert `Dtensor.to_local()` to `Dtensor._local_tensor` as soon as possible.

wplf: Addressed in `eff7bfb06`: `_to_local_if_dtensor()` now returns `dtensor._local_tensor` instead of `dtensor.to_local()` for the hybrid optimizer path. Re-ran `python -m py_compile` on the touched checkpoint/optimizer files and `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. uv run python -m pytest tests/unit_tests/test_optimizer_cpu_offloading.py -q` (`72 passed`).

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -1594,15 +1616,167 @@ def sharded_param_state_fsdp_dtensor(self, is_loading: bool = False):
...
+        # each local shard as replicated data. Its fp32 master params duplicate
+        # Megatron-FSDP's model-state main weights, so they are intentionally not
+        # stored in the optimizer state.
+        if isinstance(self.optimizer, HybridDeviceOptimizer):
+            packed_state = self._pack_hybrid_optimizer_fsdp_state_dict()
+        else:
+            packed_state = {
+                (self._param_name(k) if isinstance(k, torch.Tensor) else k): v
+                for k, v in self.state.items()
+            }
```
cspades, line 1631: This optimizer isn't just used for FSDP right? We should check that the parameters are FSDP parameters before doing this.

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -1594,15 +1616,167 @@ def sharded_param_state_fsdp_dtensor(self, is_loading: bool = False):
...
+                )
+                continue
+            packed_state[key] = state
+
+        return packed_state
+
+    def _pack_hybrid_optimizer_fsdp_state(
+        self, param: torch.nn.Parameter, param_name: str, state: dict[str, Any]
+    ) -> dict[str, Any]:
+        """Convert HybridDeviceOptimizer FSDP-local tensor state to DTensors."""
```
cspades, line 1673: Why not just use [`preprocess_state_dict_for_uneven_dtensor` or `update_uneven_dtensor_chunk_metadata`](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py#L238) instead of re-implementing it here? (Also, we should have all of these types of utilities inside the Megatron-FSDP source directory unless it absolutely needs to be in the training loop / optimizer code!)

megatron/core/transformer/fsdp_dtensor_checkpoint.py
```diff
@@ -304,11 +304,14 @@ def split_swiglu_linear_fc1(data, dist_param, swiglu_shard_axis, is_expert_param
...
-            )
-            local_tensor = data.to_local()
+            if data.shape == global_shape:
+                local_tensor = data.to_local()
+            else:
+                assert data.numel() == dist_param.numel(), (
+                    f"DTensor shape mismatch: data.shape={data.shape} vs "
+                    f"dist_param.shape={global_shape}"
+                )
+                local_tensor = data.to_local().view(-1)
```
cspades, line 314: These can also be `DTensor._local_tensor` BTW! Plus more instances of this in this same file.

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -1594,15 +1616,167 @@ def sharded_param_state_fsdp_dtensor(self, is_loading: bool = False):
...
+                        value.data.view(-1),
+                        flat_param,
+                        dist_index=param.megatron_fsdp_dist_index,
+                        is_expert_param=is_expert_param,
+                        run_check=False,
+                        update_uneven_dtensor_chunk_meta=False,
+                    )
+                    self._set_flat_fsdp_dtensor_chunk_metadata(
+                        value, param.megatron_fsdp_slice
+                    )
```
cspades, line 1735: So TL;DR I assume the `packed_state` is our offloaded state, and we want them to be DTensors as well?

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -1594,15 +1616,167 @@ def sharded_param_state_fsdp_dtensor(self, is_loading: bool = False):
...
+                        f"Failed to pack HybridDeviceOptimizer state '{key}' for "
+                        f"FSDP parameter '{param_name}' on rank {rank}."
+                    ) from exc
+            packed_state[key] = value
+
+        return packed_state
+
+    @staticmethod
+    def _set_flat_fsdp_dtensor_chunk_metadata(tensor: DTensor, fsdp_slice: slice) -> None:
+        """Attach DCP chunk metadata for a flat FSDP-local shard without collectives."""
```
cspades, line 1748: We can just re-factor the existing function in `uneven_dtensor.py` to skip the metadata AG update! Less code outside MFSDP the better!

cspades (commented):
Some important nits, basically if we can move most of this code into MFSDP source it would be much cleaner and better.

INFO: Could you also add more commentary explaining the lifecycle of the offloaded DTensors here? Just want to make sure if this is the right way to implement this.

PR #4624, opened 2026-05-05
[dev]: faster implementation of mHC fused kernels

jingqiny-99:
# What does this PR do ?

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/core/fusions/fused_mhc_kernels.py
```diff
@@ -1,40 +1,840 @@
...
+        _CUTILE_DEVICE_SUPPORT_CACHE = False
+        return False
+
+    output = f"{result.stdout}\n{result.stderr}"
+    if "Cannot find option named" in output and arch in output:
+        _CUTILE_DEVICE_SUPPORT_ERROR = output.strip().splitlines()[0]
+        _CUTILE_DEVICE_SUPPORT_CACHE = False
+        return False
+
+    _CUTILE_DEVICE_SUPPORT_CACHE = True
```
Victarry, line 205: [P2] This caches cuTile support as true even when `tileiras` exits non-zero for reasons other than the specific unsupported-arch stderr pattern above. A broken compiler or unsupported target could then make the fused path dispatch to cuTile and fail later. Please treat any non-zero `result.returncode` as unsupported, preserving stderr/stdout in `_CUTILE_DEVICE_SUPPORT_ERROR`, before caching true.

megatron/core/fusions/fused_mhc_kernels.py
```diff
@@ -729,101 +1725,1421 @@ def _next_power_of_2(n: int) -> int:
...
+    if _BACKEND_INFO_LOGGED:
+        return
+    _BACKEND_INFO_LOGGED = True
+    backend_selection, all_native = _mhc_backend_status()
+    log_single_rank(
+        logger,
+        logging.WARNING if all_native else logging.INFO,
+        f"[mHC] fused backend selection: {backend_selection}",
+    )
+    if all_native and safe_get_rank() == 0:
```
Victarry, line 3032: [P2] If `use_fused_mhc=True` but both Triton and cuTile are unavailable, this currently only warns and then continues with all-native torch implementations. I think this should be a hard error. The flag expresses an intent to use accelerated/fused mHC kernels, and silently running all-native can make large jobs look correctly configured while getting none of the expected performance benefit. Users who want the reference implementation can leave `use_fused_mhc=False`.

megatron/core/fusions/fused_mhc_kernels.py
```diff
@@ -729,101 +1725,1421 @@ def _next_power_of_2(n: int) -> int:
...
+    _raise_mhc_backend_validation_error()
+    global _BACKEND_INFO_LOGGED
+    if _BACKEND_INFO_LOGGED:
+        return
+    _BACKEND_INFO_LOGGED = True
+    backend_selection, all_native = _mhc_backend_status()
+    log_single_rank(
+        logger,
+        logging.WARNING if all_native else logging.INFO,
+        f"[mHC] fused backend selection: {backend_selection}",
```
Victarry, line 3030: [P3] This backend summary is emitted at INFO unless all ops are native. In `auto` mode with partial fallback, e.g. Triton is available but cuTile is not, some ops can run on Triton while `proj_rms`/`proj_rms_compute_h` fall back to native. That is functionally valid, but it has a meaningful performance implication and is easy to miss. Please consider warning when any fused op falls back to native while another op uses an accelerated backend, or otherwise make partial fallback explicitly visible.

megatron/core/transformer/hyper_connection.py
```diff
@@ -92,6 +95,39 @@ def native_proj_rms(x: Tensor, weight: Tensor, eps: float = 1e-6) -> Tuple[Tenso
...
+    def backward(ctx, grad1, grad2, grad3):
+        """Combine gradients from the three broadcast aliases."""
+        grads = [g for g in (grad1, grad2, grad3) if g is not None]
+        if len(grads) == 0:
+            return None, None
+        if len(grads) == 1:
+            return grads[0], None
+        if len(grads) == 2:
+            return grads[0] + grads[1], None
+        return ctx.fused_add_3_fn(grad1, grad2, grad3), None
```
Victarry, line 128: [P3] `BroadcastTensorFused` takes `fused_add_3_fn` through `apply()` and stores it on `ctx`, but every caller passes `self._fused_add_3_op`, which is always `native_fused_add_3`. The nearby comment also says this add intentionally always uses the torch.compile native path regardless of selected backend. Consider hardcoding `native_fused_add_3` in this backward and removing `_fused_add_3_op` plus the extra `apply()` argument; if a real backend-specific add is needed later, the indirection can be reintroduced then.

Victarry (commented):
Reviewed the fused mHC backend/control-flow changes. Main concerns are around backend availability semantics and fallback visibility.

Victarry (approved):

ydx980521:
Hi author, I have a question about the _triton_hpb_bwd_g_hp_hr_kernel kernel. Why is the matrix multiplication implemented using FP32 multiplication + summation (* + sum) instead of using tl.dot? Based on my estimation, tl.dot should offer better performance. Could you please clarify the reason?

jingqiny-99:
> Hi author, I have a question about the _triton_hpb_bwd_g_hp_hr_kernel kernel. Why is the matrix multiplication implemented using FP32 multiplication + summation (* + sum) instead of using tl.dot? Based on my estimation, tl.dot should offer better performance. Could you please clarify the reason?

Thanks for your good question. In short, we  have tried `tl.dot`, it came out slower, so `* + sum` here is on
purpose.

This kernel is limited by memory bandwidth, not by compute. `g_hr` is a batched
(N x C) @ (C x N) matmul with N = num_residual_streams = 4, and the current
version already runs at ~80-89% of peak HBM bandwidth. `tl.dot` does make the
math cheaper, but MMA needs the data in a specific layout, so it has to go
through shared memory first. That extra shared memory traffic pulls DRAM
bandwidth down from ~81% to ~61%, and the kernel ends up slower overall: 4-17%
on H100 and 9-50% on GB200, depending on the shape (both autotuned).

To make `tl.dot` pay off we'd probably have to tweak a few other places around
it as well, which gets complicated. And since we're already close to peak
memory bandwidth, there isn't much left to gain anyway.

PR #4633, opened 2026-05-05
Treat MambaLayer as an FSDP unit module

wujingyue:
Based on #4717.


cspades (commented):
Tested this (without CG) on my CG branch with a simple HybridModel MEMEMEME and it works, were you able to reproduce the original bug and why does making non-units persistent fix it? Is it basically just that some pointer gets dereferenced after the first reset and we AG the wrong variable? More details would be super nice!

wujingyue:
> were you able to reproduce the original bug and why does making non-units persistent fix it? 

Good question. I struggled a lot to reproduce the so-called "original bug". My best attempt so far is https://github.com/wujingyue/Megatron-LM/blob/mamba_repro/repro_illegal_memory_access.sh, but it only triggers an IMA when MambaLayer is non-unit.

cc @xrennvidia for a better repro or confirm the "original bug" that you noticed no longer shows up

PR #4324, opened 2026-04-15
Fix checkpoint loading with `load_main_params_from_ckpt=True` for grouped weight

ksivaman:
# What does this PR do ?

Support loading a multi-weight `GroupedLinear` checkpoint into a single weight module and vice-versa. This affects the path where the optimizer params directly load the checkpoint. The normal (model) loading path is already supported.

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

Once all required reviewers have approved, the `Approved` label is applied **automatically**.

### Merge

Any member of [mcore-engineers](https://github.com/orgs/NVIDIA/teams/mcore-engineers) will be able to merge your PR.

<details>
<summary>For MRs into `dev` branch</summary>
The proposed review process for `dev` branch is under active discussion.

MRs are mergable after one approval by either `eharper@nvidia.com` or `zijiey@nvidia.com`.
</details>

PR #4071, opened 2026-03-31
[Dev] Skip routed expert padding for graph-safe MoE

zhongbozhu:
# What does this PR do ?

This PR is a follow up fix to https://github.com/NVIDIA/Megatron-LM/pull/3890, which introduced TE Fused group mlp layer for graph safe moe. However, padding itself as a stand-alone operation can also be graph-unsafe if the padding was implemented based on knowing problem shapes on host, which is the case for router padding and explicit Fp8Padding module. 

HybridEp provides the ability of doing the padding fused with local permute, which is done on-device, so it's cuda graph safe. This PR checks whether padding has already been done so that we can safely skip the graph-unsafe padding, this will be used together with the paged stashing PR: https://github.com/NVIDIA/Megatron-LM/pull/2690. 

PR to main: TBD (waiting on https://github.com/NVIDIA/Megatron-LM/pull/3971 to be merged to main)

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

[...]

PR #3971, opened 2026-03-20
Support GEMM + Swiglu fused MLP

ksivaman:
# What does this PR do ?

This PR supports GEMM + Swiglu fused MLP via Transformer Engine sequential ops, including:

- TE `GroupedTensor` support in distributed optimizer.
- TE `GroupedTensor` checkpointing support for cases to load a multi-param (1 param per expert) checkpoint into a single param config.

**Note**: Support is already added in dev branch and this is cherry-picking #3890.

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

Once all required reviewers have approved, the `Approved` label is applied **automatically**.

### Merge

Any member of [mcore-engineers](https://github.com/orgs/NVIDIA/teams/mcore-engineers) will be able to merge your PR.

<details>
<summary>For MRs into `dev` branch</summary>
The proposed review process for `dev` branch is under active discussion.

MRs are mergable after one approval by either `eharper@nvidia.com` or `zijiey@nvidia.com`.
</details>

PR #4636, opened 2026-05-05
Combine GEMM + SwiGLU fused MLP PRs (3890, 4071, 4095, 4219, 4311, 4324) → main

Connor-XY:
## What does this PR do?

This PR consolidates six `dev`-branch PRs related to TE fused grouped MLP into a single PR targeting `main`. It supersedes #3971 (which mirrored only #3890 to `main`) and brings the full feature set in one merge.

### Source PRs (all merged to `dev`)

| dev PR | Title | What it adds |
|---|---|---|
| [#3890](https://github.com/NVIDIA/Megatron-LM/pull/3890) | Support GEMM + Swiglu fused MLP | TE op-fuser path for `TEGroupedMLP` (`_is_fused_impl_supported`, `_make_fused_ops`, `_fused_forward`, `_make_fused_impl_pre_forward_hook`); GLU-interleaved layout (`moe_mlp_glu_interleave_size`); grouped-quantized-tensor support in `DistributedOptimizer` (`_is_grouped_quantized_tensor`, `_expand_quantized_param_shard_for_cast`); `_normalize_grouped_parameter_keys` checkpoint-key compat hook on `TEGroupedLinear`; CUTLASS 256-byte alignment in `get_align_size_for_quantization` when op-fuser is on; new CLI flag `--use-transformer-engine-op-fuser`. |
| [#4071](https://github.com/NVIDIA/Megatron-LM/pull/4071) | Skip routed expert padding for graph-safe MoE | Refactors the `moe_router_padding_for_quantization` check into a `skip_routed_expert_padding(config)` helper that *also* skips quantization padding when the token dispatcher is `flex` with `hybridep` backend (the dispatcher applies padding itself). Eliminates a double-pad on graph-safe HybridEP paths. |
| [#4095](https://github.com/NVIDIA/Megatron-LM/pull/4095) | TE fused grouped MLP with grouped bias and delayed wgrad | Adds `single_grouped_bias` to the op-fuser path (renames `single_grouped_parameter` → `single_grouped_weight`); threads `delay_wgrad_compute` through `te.pytorch.ops.GroupedLinear`; drops the `need_backward_dw()` gate; gates `_is_fused_impl_supported` on TE ≥ 2.14.0; adds a fused-aware `backward_dw()` override. |
| [#4219](https://github.com/NVIDIA/Megatron-LM/pull/4219) | Enabled fused grouped MLP for `quick_gelu` and add config for grouped params | Extends `_is_fused_impl_supported` and `_make_fused_ops` to support `quick_gelu` via `te.pytorch.ops.ScaledClampedQGeGLU`; adds `moe_single_grouped_weight` / `moe_single_grouped_bias` `TransformerConfig` fields with validation; threads them through `TEGroupedLinear.__init__` and `_set_arg` for checkpoint loading; updates the `test_hybrid_moe_model` GOLDEN_CONFIG. |
| [#4311](https://github.com/NVIDIA/Megatron-LM/pull/4311) | Fix fused grouped MLP wgrad hooks for DDP reduce-scatter | After the fused children's `backward_dw()` call, explicitly invoke `linear_fc{1,2}._trigger_wgrad_accumulation_and_reduce_hooks()`. The wgrad hooks (DDP reduce-scatter, etc.) live on the original `linear_fc1/fc2` modules but `backward_dw()` is called on the new GroupedLinear instances created by `_make_fused_ops()`. Without the explicit trigger, `param.grad` is never zeroed and `AccumulateGrad` performs a spurious `add_` into `main_grad`. |
| [#4324](https://github.com/NVIDIA/Megatron-LM/pull/4324) | Fix checkpoint loading with `load_main_params_from_ckpt=True` for grouped weight | Adds `_normalize_state_dict_for_grouped_params` to `DistributedOptimizer._build_model_param_to_state_dict_param_map`. Mirrors the `normalize_grouped_parameter_keys` hook (which only fires during `load_state_dict`) so the optimizer reload path also handles single-grouped ↔ indexed weight layout drift. |

### Rebase-fix work on top

Beyond the six source PRs, this branch carries fixes needed to land the combined feature on current `main`:

- Fix the `self.bias_act_func` → local `bias_act_func` typo on the non-recompute forward path (the closure is defined locally; calling it via `self.` AttributeError'd at runtime).
- Promote `normalize_grouped_parameter_keys` from an inline `__init__` closure to a method `_normalize_grouped_parameter_keys` so it's directly testable.
- Use `device="meta"` for op-shell construction in `_make_fused_ops` (existing weights are reattached after, so the GPU allocation was a wasted transient).
- **Read the wrappers' combined `delay_wgrad_compute`** (`config.delay_wgrad_compute or config.overlap_dispatch_backward_with_experts_wgrad`) in `_make_fused_ops` and the `backward_dw()` gate, not the raw config flag — otherwise runs that enable wgrad delay through `overlap_dispatch_backward_with_experts_wgrad` would silently lose the overlap optimization.
- Fix `_make_fused_ops` activation comment typo "GEGL" → "GeGLU".
- Add unit-test coverage that the original PRs didn't carry: `_apply_bias`, `_remove_glu_interleaving`, `_make_fused_impl_pre_forward_hook`, `_make_fused_ops` shapes (with mocked TE) including the `quick_gelu`/ScaledClampedQGeGLU branch and the `single_grouped_bias=True` path, `_fused_forward` arg threading, `_split_grouped_checkpoint_tensor` (all branches), `_normalize_grouped_parameter_keys` (8 cases including round-trip), `_normalize_state_dict_for_grouped_params` (8 cases — indexed↔grouped, bias path, `module.`-prefix stripping, ambiguous-match guard, incomplete-set guard, target-already-present guard, non-TE-grouped skip), `backward_dw` fused dispatch including PR 4311 hook trigger, `backward_dw` fallback when delay is off, `_expand_quantized_param_shard_for_cast`, `_is_grouped_quantized_tensor`. Plus a GPU-gated `test_gpu_make_fused_ops_constructs_with_real_te` for the meta-device + weight-reattach path against real TE.
- Comment improvements in `param_and_grad_buffer.py` (no behavior change), addressing reviewer questions on the original PR.
- Various lint and config alignments to keep CI green.

### Files changed (vs `main`)

```
megatron/core/distributed/param_and_grad_buffer.py            (comments only)
megatron/core/extensions/transformer_engine.py                (_normalize_grouped_parameter_keys + _split_grouped_checkpoint_tensor + extra_kwargs for grouped params)
megatron/core/optimizer/distrib_optimizer.py                  (grouped-quantized-tensor support + _normalize_state_dict_for_grouped_params)
megatron/core/transformer/moe/experts.py                      (TEGroupedMLP fused path + DDP wgrad hook trigger + skip_routed_expert_padding)
megatron/core/transformer/moe/moe_utils.py                    (CUTLASS alignment + skip_routed_expert_padding helper)
megatron/core/transformer/transformer_config.py               (new flags: use_transformer_engine_op_fuser, moe_mlp_glu_interleave_size, moe_single_grouped_{weight,bias})
megatron/training/checkpointing.py                            (_set_arg for new flags)
tests/unit_tests/models/test_hybrid_moe_model.py              (GOLDEN_CONFIG)
tests/unit_tests/optimizer/test_distrib_optimizer_grouped_quantized.py   (NEW)
tests/unit_tests/transformer/moe/test_grouped_mlp.py
tests/unit_tests/transformer/test_transformer_engine_grouped_linear.py   (NEW)
```

### Validation

- Linters clean (`black --skip-magic-trailing-comma --skip-string-normalization --check`, `isort --check`, `pylint` 10/10, `ruff check`) on the full lint scope.
- CPU mock + GPU integration tests pass on draco's `batch_singlenode` partition with the canonical post-rename TE container (TE `2.14.0+f031cf87`): **46 passed in 6.54s**, including `test_gpu_forward_backward`, `test_gpu_make_fused_ops_constructs_with_real_te`, all the new coverage above, and the existing `_split_grouped_checkpoint_tensor` / `_normalize_grouped_parameter_keys` round-trip tests.

## Contribution process

- [x] Cherry-picks of #3890, #4071, #4095, #4219, #4311, #4324 applied in their `dev` merge order (rebase-fix work for #3890 follows #3971's pattern).
- [x] Lint clean against the project autoformat config.
- [x] Existing unit tests pass; new tests added for previously-uncovered surface, including the `_normalize_state_dict_for_grouped_params` cross-module ambiguity guard.

### Outstanding follow-up

- @zhongbozhu noted on `distrib_optimizer.py:1083` that the single-grouped-weight path needs more numerical fixes from `dev`. Awaiting pointers to specific commits/PRs to either fold into this PR or track in a follow-up.


Connor-XY:
/ok to test [fe61201](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/fe6120189fa2b9f908c123a89e8bba1bedc66e91)

Connor-XY:
/ok to test [4b0b739](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/4b0b739431ffc097697eb1d98605cb757d1410f3)

Connor-XY:
/ok to test [1858cdc](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/1858cdc9aae8f9ee3916cac7120678952477063b)

Connor-XY:
/ok to test [681ffa7](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/681ffa79465a0ff3929203ab228eb0a72960e4e2)

yaox12 (approved):

gdengk:
Hi @Connor-XY  could we also ship this dev PR together with this one https://github.com/NVIDIA/Megatron-LM/pull/4311
It's also part of fused gemm feature to support delayed wgrad.

Connor-XY:
/ok to test [e0f2b22](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/e0f2b226a33aca3d3a8500d1e02aec059611b8cc)

kvareddy:
@Mellonta can you please review this change?

kvareddy:
@jiemingz can you please review this PR?

zhongbozhu:
We might also need to cherry-pick this small fix into this PR, which is a fix to the TE grouped mlp module: https://github.com/NVIDIA/Megatron-LM/pull/4071

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -1062,6 +1066,70 @@ def _get_main_param_and_optimizer_states(self, model_param):
...
+            and callable(tensor.split_into_quantized_tensors)
+            and getattr(tensor, "quantizer", None) is not None
+        )
+
+    @classmethod
+    def _is_distopt_quantized_param(cls, tensor: torch.Tensor) -> bool:
+        """Check if tensor should follow quantized parameter path in dist optimizer."""
+        return is_float8tensor(tensor) or cls._is_grouped_quantized_tensor(tensor)
+
+    def _expand_quantized_param_shard_for_cast(
```
zhongbozhu, line 1083: this part of the code is related with the single grouped weight feature, which still needs more fixes (from dev too) to make it numerically correct

Connor-XY:
@zhongbozhu thanks — picked PR #4071 in `64dcdc85d` (cherry-pick) and `d2f3040c4` (test fixup for the new `moe_token_dispatcher_type` / `moe_flex_dispatcher_backend` reads in `skip_routed_expert_padding`). Updated the PR title/description to list it.

Re: the inline comment on `distrib_optimizer.py:1083` — understood that the single-grouped-weight path needs more numerical fixes from `dev`. Could you point me at the specific commits/PRs you have in mind so I can pull them in here too? Otherwise we'll merge what's currently staged and follow up in a separate PR.

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -244,6 +244,10 @@ def __init__(
         # or bucket.grad_data.
         self.cached_param_buffer_shard_list = [None] * len(self.buckets)
         self.cached_grad_buffer_shard_list = [None] * len(self.buckets)
+        # Cached views remember the grad-mode context in which they were created. Rebuild them
+        # after a mode change so param all-gathers never mutate no_grad-created views while
+        # autograd is enabled.
+        self._cached_param_buffer_shards_grad_enabled = None
```
zhongbozhu: this line of code looks already a dangling pointer in dev branch, but somehow it's still referenced in this PR, this might be related with merge conflict resolving, but we need more testing for this part

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -406,23 +410,31 @@ def start_param_sync(self, force_sync: bool = False):
...
             # does not need a copy-back step, so coalescing works correctly.
-            with _coalescing_manager(
-                self.intra_distributed_optimizer_instance_group, async_ops=async_op
-            ) as cm:
-                for idx, bucket in enumerate(self.buckets):
-                    if self.cached_param_buffer_shard_list[idx] is None:
-                        self.cached_param_buffer_shard_list[idx] = shard_buffer(
-                            bucket.param_data, self.intra_distributed_optimizer_instance_size
+            current_grad_enabled = torch.is_grad_enabled()
+            if self._cached_param_buffer_shards_grad_enabled != current_grad_enabled:
```
zhongbozhu: this line has been removed in dev branch, @ksivaman can you double check the logic for this part?

zhongbozhu:
> @zhongbozhu thanks — picked PR #4071 in `64dcdc85d` (cherry-pick) and `d2f3040c4` (test fixup for the new `moe_token_dispatcher_type` / `moe_flex_dispatcher_backend` reads in `skip_routed_expert_padding`). Updated the PR title/description to list it.
> 
> Re: the inline comment on `distrib_optimizer.py:1083` — understood that the single-grouped-weight path needs more numerical fixes from `dev`. Could you point me at the specific commits/PRs you have in mind so I can pull them in here too? Otherwise we'll merge what's currently staged and follow up in a separate PR.

Yeah, we are meeting with @WanZzzzzz to see if we can fix single weight, currently not sure if that fix should block this PR, but ideally we want to merge a PR without known bug, but we can discuss about timeline and maybe do it in another PR.

megatron/core/distributed/param_and_grad_buffer.py
Connor-XY: @zhongbozhu thanks — I traced this across `main` / `dev` / this PR. The state is more nuanced than fully-dangling:

| | `_cached_param_buffer_shards_grad_enabled = None` (init) | `current_grad_enabled` check + cache reset (read site, ~L413) | `with torch.no_grad():` wrap (~L420) |
|---|---|---|---|
| `nvidia/main` | absent | absent | absent |
| `nvidia/dev` | present (dangling — never read) | **removed** | **removed** |
| this PR | present | **present** | **present** |

So this PR is _not_ dangling — it carries both the init AND the read site (the read at L413 + the `no_grad` wrap that PR #3890 originally introduced). What was removed on dev is the consumption side, leaving only the orphan init.

Recent dev work (#4560, #4563, #4649, #4609) replaced the grad-mode tracking with explicit `post_all_gather_processing` hooks for MXFP8 / NVFP4 / layerwise overlap. Those hooks are NOT carried by this PR (different scope), so on `main` the grad-mode protection here is the only mechanism guarding against the autograd-mid-AG issue PR #3890 was written to fix.

@ksivaman can you confirm whether keeping the L413 read site + `no_grad` wrap is still required when #3890 lands on `main` ahead of the dev-side post-processing refactors, or whether the migration plan is to bring those over too? If the protection is no longer needed, happy to delete the dead init + read site here in one commit.

megatron/core/distributed/param_and_grad_buffer.py
Connor-XY: Addressed inline at https://github.com/NVIDIA/Megatron-LM/pull/4636#discussion_r3211339106 — TL;DR: this PR is not dangling, the read site at L413 still uses this field. It is dangling on `dev` (consumption side removed by later refactors), but `main` doesn't have those refactors yet, so removing the protection here would re-introduce the autograd-mid-AG issue PR #3890 was originally guarding against. Awaiting @ksivaman's call on whether to delete or keep.

Connor-XY:
/ok to test [d2f3040](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/d2f3040c45c4f75522479b9e80ecb8b3d881803e)

Connor-XY:
/ok to test [02d7e3b](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/02d7e3b3b0f5a737ab0f9895ce6d55ef26b24df5)

maanug-nv (approved):
wrt `megatron/training/checkpointing.py` changes, deferring to @NVIDIA/training-adlr , I don't think the additions to `load_args_from_checkpoint()` affect Megatron-Bridge.

megatron/core/distributed/param_and_grad_buffer.py
zhongbozhu: @Connor-XY https://github.com/NVIDIA/Megatron-LM/pull/4560 and https://github.com/NVIDIA/Megatron-LM/pull/4649 are PRs to a temporary staging branch, so https://github.com/NVIDIA/Megatron-LM/pull/4563 is the actual dev2main PR that we should consider if it should be shipped together this PR.

Connor-XY: Thanks for the pointer to #4563 (and for clarifying that #4560 / #4649 are staging-branch PRs, not dev2main candidates).

**On bundling #4563:** I'd recommend keeping them separate.
- Different concern. #4563 fixes FP8/MXFP8/NVFP4 param-gather post-processing during eval-interleaved training (Qiyu's `_post_param_sync` extraction). This PR is TE fused grouped MLP. Same file, distinct logic — no functional overlap.
- It does **not** subsume the earlier question (whether to keep the `_cached_param_buffer_shards_grad_enabled` + `no_grad` wrap from #3890). That's a separate decision still waiting on @ksivaman.
- Cherry-picks cleanly onto main on its own — I diffed `param_and_grad_buffer.py` between this PR head and #4563 head, and the overlap is contextual only.
- Reviewer surface is already wide here (6 PRs aggregated). Adding FP8/MXFP8/NVFP4 eval semantics widens it further.

Happy to flip the call if you'd rather bundle to amortize a single review cycle.

Connor-XY:
/ok to test [e20cc6a](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/e20cc6a091cdc94d188c16b3dc6af9be3c40102e)

Mellonta:
Just wanted to flag a tricky case I saw. When the op-fuser flag is on with HybridEP, Megatron's _is_fused_impl_supported() passes but TE's fusion pass can still silently fall back to the basic-op path with no log message. The fused and basic-op paths have different size contracts under HybridEP, so the silent fallback turns into a runtime crash. It makes debugging quite difficult since the log shows nothing. Not sure what the best practice is here.

Connor-XY:
> Just wanted to flag a tricky case I saw. When the op-fuser flag is on with HybridEP, Megatron's _is_fused_impl_supported() passes but TE's fusion pass can still silently fall back to the basic-op path with no log message. The fused and basic-op paths have different size contracts under HybridEP, so the silent fallback turns into a runtime crash. It makes debugging quite difficult since the log shows nothing. Not sure what the best practice is here.

Disable the fused path explicitly in `_is_fused_impl_supported` whenever the active dispatcher is flex+hybridep, so the construction-time assert fires immediately instead of the silent-fallback runtime crash.

Connor-XY:
/ok to test [1de5c99](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/1de5c99c3f5b7d3c0378e11140bb367663eb569a)

Connor-XY:
/ok to test [b30ee43](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/b30ee43d10fadff3ede673c5b5a642b18cee709a)

Connor-XY:
/ok to test [df7435a](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/df7435a5d724585f4ca68dd0aba4ca237094f551)

Mellonta:
> Disable the fused path explicitly in `_is_fused_impl_supported` whenever the active dispatcher is flex+hybridep, so the construction-time assert fires immediately instead of the silent-fallback runtime crash.

Sorry, I wasn't clear. flex+hybridep is the intended use case, so we want to keep it on. My concern is what happens when something else goes subtly wrong: `NVTE_CUTEDSL_FUSED_GROUPED_MLP` not set, FC dims not 256-aligned, `moe_mlp_glu_interleave_size` not set, etc. Any of these silently knocks TE off the fused path, and then HybridEP's padding doesn't match what the basic-op path expects, so the user gets a `split_with_sizes` crash mid-training instead of a clear error at startup.

I think the easy fix is to mirror those static TE gates in `_is_fused_impl_supported()`, but it may cause duplicate code in two code bases.

chtruong814 (approved):

jiemingz (approved):

Connor-XY:
Thanks for clarifying — I misread the first round and reverted the wrong guard. Now removed in cf56b7b0c (commit 5fd0c0b on the PR head): both the `flex+hybridep` rejection in `_is_fused_impl_supported` and the matching test are gone, since flex+hybridep is the intended supported path.

On mirroring TE's static gates: agreed in principle, that's the cleanest user-facing behavior. Two questions before I write it:

1. **What's the canonical list?** From your message I have:
   - `NVTE_CUTEDSL_FUSED_GROUPED_MLP` env var must be set (truthy?)
   - FC dims (FFN-hidden? hidden?) 256-aligned
   - `moe_mlp_glu_interleave_size` must be set (non-`None`?)
   
   Is there a doc / TE source file you can point me at for the authoritative set, including any I'm missing? I'd rather copy from a single source than reconstruct from this comment thread.

2. **Where should the check live?** Inlining in `_is_fused_impl_supported` works but, as you noted, duplicates the gate. A cleaner option is a small TE-provided predicate (e.g. `te.pytorch.ops.is_grouped_mlp_fusion_supported(...)`) that we just call here. Open to whichever the TE team prefers — if there's appetite for a TE-side predicate I can hold off on the inline mirror.

For now this PR just removes the wrong guard. Happy to do the proper gate-mirror in a follow-up commit on this PR or a separate PR, whichever fits the review path better.

Connor-XY:
/ok to test [cf56b7b](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/cf56b7b0c6468b448051eba4ae672f79f2a851f6)

tests/unit_tests/transformer/moe/test_grouped_mlp.py
```diff
@@ -186,3 +731,174 @@ def test_gpu_forward_backward_with_no_tokens_allocated(self):
...
+                )
+        for idx in range(experts.linear_fc2.num_gemms):
+            if not getattr(experts.linear_fc2, "single_grouped_weight", False):
+                assert getattr(ops[2], f"weight{idx}") is getattr(
+                    experts.linear_fc2, f"weight{idx}"
+                )
+
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+    @pytest.mark.internal
+    def test_gpu_fused_path_loss_decreases(self):
```
nanz-nv, line 811: Is the env var NVTE_CUTEDSL_FUSED_GROUPED_MLP being set correctly to enable the fused grouped GEMM in TE, according to the requirement [here](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py#L75).
Same for other tests. 
In my previous attempt to include these fused grouped GEMM in the unit test, setting this env var at unit test level doesn't help because the support is checked at the time TE is loaded, which doesn't happen at unit test level.

Connor-XY: Excellent catch! You're right on both points and the same gap was in my new functional recipe.

**Functional recipe (fixed in ec1d1e055):** Added `NVTE_CUTEDSL_FUSED_GROUPED_MLP: 1` to the `ENV_VARS` block of `tests/functional_tests/test_cases/moe/gpt3_mcore_te_tp1_pp1_te_4experts_groupedGEMM_op_fuser/model_config.yaml`. The recipe runner exports env vars before launching `pretrain_gpt.py`, so TE imports with the var in place.

**GPU unit tests (`test_gpu_make_fused_ops_constructs_with_real_te`, `test_gpu_fused_path_loss_decreases`):** Updated the docstring of the convergence test to make the limitation explicit — these only exercise the Megatron-side wrapper plumbing (op-sequence construction, fused-impl pre-forward hook, weight aliasing into the TE op wrappers, (delayed-)wgrad backward), not TE's underlying cuDSL kernel selection. I left them in because the wrapper plumbing is what this PR adds and they still catch wrapper-side regressions (e.g., wgrad-hook breakage in PR #4311).

Connor-XY: Quick follow-up after probing this on actual hardware. Sharing in case it shifts the picture for you too.

Looking at TE's [`is_supported()`](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py#L73): the env var is one of several gates. The kernel additionally requires **SM100 (Blackwell)** plus the import-time `register_forward_fusion` step. A100 is SM80, H100 is SM90, neither qualifies. So on the current CI matrix (dgx_h100), this PR's fused-op-fuser path always runs TE's basic-op fallback regardless of how we set the env var.

Net for this PR:
- I kept `NVTE_CUTEDSL_FUSED_GROUPED_MLP=1` in the functional recipe ENV_VARS — it's a no-op on H100 today but means no recipe change when Blackwell joins the matrix.
- Updated the recipe comment and unit-test docstring to make the SM100 requirement explicit, so this PR's coverage claim is honest: "exercises the Megatron op-fuser wrapper end-to-end", not "exercises the cuDSL kernel."

Connor-XY:
/ok to test [46be5d1](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/46be5d172cf218562286f0d5910eb8a6a120b5ae)

Connor-XY:
Picking up this thread now that #4621 is visible.

Status check:
- #4621 ("Fix single grouped weight when enabling MXFP8 primary weight") is still **open** against `dev`.
- The buggy code path in #4621 is the dist-optimizer quantized-param shard logic (`_is_distopt_quantized_param` / `_expand_quantized_param_shard_for_cast`) plus the param-AG quantized-storage handling in `param_and_grad_buffer.py`. All of that is **quantized-tensor-only** — biases don't enter those paths in mxfp8 training.
- This PR's tests don't exercise the buggy combo (no fp8/mxfp8, just bf16+groupedGEMM).

To address your "ideally no known bug" concern without coupling this PR's schedule to #4621, I added a fail-fast `__post_init__` guard in `transformer_config.py` (commit 8d3920d8a). It rejects `moe_single_grouped_weight=True` unless the user is also on fp8 mode with the mxfp8 recipe:

```python
if self.moe_single_grouped_weight:
    if self.fp4 or not self.fp8 or self.fp8_recipe != Fp8Recipe.mxfp8:
        raise ValueError(
            "moe_single_grouped_weight is currently supported only with fp8 mode "
            "and fp8_recipe='mxfp8'."
        )
```

megatron/core/transformer/moe/experts.py
```diff
@@ -453,6 +744,31 @@ def backward_dw(self):
...
+                # (those are in the nn.Module tree), but backward_dw() is called on the
+                # NEW GroupedLinear instances created by _make_fused_ops().  We must
+                # explicitly fire the hooks on the originals so DDP can zero param.grad
+                # and trigger reduce-scatter – otherwise param.grad is never cleared and
+                # AccumulateGrad performs a spurious add_ into main_grad.
+                # TODO: find a better place to invoke _trigger_wgrad_accumulation_and_reduce_hooks.
+                # The wgrad hook registration lives in TE while the trigger is issued here
+                # in MCore, so the hook lifecycle is split across both codebases. Consolidate
+                # ownership on one side (either register+trigger entirely in TE, or expose
+                # the fused backward_dw through MCore) to remove this fragmentation.
```
asolergi-nv, line 768: Is there a plan to address this?

Connor-XY:
/ok to test [8d3920d](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/8d3920d8aef2a6dec2f927a9fce16c61c83f1c36)

megatron/core/transformer/moe/experts.py
Connor-XY: cc @gdengk

asolergi-nv (approved):

fanshiqing (approved):

megatron/core/optimizer/distrib_optimizer.py
Connor-XY: I added a fail-fast `__post_init__` guard in `transformer_config.py` (commit 8d3920d8a). It rejects `moe_single_grouped_weight=True` unless the user is also on fp8 mode with the mxfp8 recipe:

```python
if self.moe_single_grouped_weight:
    if self.fp4 or not self.fp8 or self.fp8_recipe != Fp8Recipe.mxfp8:
        raise ValueError(
            "moe_single_grouped_weight is currently supported only with fp8 mode "
            "and fp8_recipe='mxfp8'."
        )
```

yashaswikarnati (approved):

Connor-XY:
/ok to test [d49ad27](https://github.com/NVIDIA/Megatron-LM/pull/4636/commits/d49ad27420447d545c11df45bb8f405fdc47725c)

zhongbozhu (approved):
For future reference, let's aggregate all the PRs necessary and run E2E convergence test with optimization toggles turned on.

ericharper (approved):

kvareddy (approved):

ksivaman:
Confirmed numerics with TE tests are equivalent as that with `dev` branch. LGTM.

tests/test_utils/recipes/h100/unit-tests.yaml
```diff
@@ -43,6 +43,14 @@ spec:
...
     fi
 
+    # Enable TE's cuDSL fused grouped MLP path for MoE buckets. The kernel
+    # additionally requires SM100 (Blackwell), so on H100/A100 CI this is a
+    # no-op; setting it here means the kernel is picked up automatically when
+    # Blackwell hardware joins the unit-test matrix.
+    if [[ "$BUCKET" == *"transformer/moe"* || "$BUCKET" == *"test_moe_experts"* ]]; then
+        export NVTE_CUTEDSL_FUSED_GROUPED_MLP=1
+    fi
+
```
ko3n1g, line 53: We shouldn't set env vars here.. it makes it difficult to spot how/where behavior comes from. We should only set env vars within the unit tests (i.e. with a fixture).

PR #4638, opened 2026-05-06
Fix DistributedOptimizer OOM when resuming from checkpoint

### LINKED ISSUE #3576: [bug] Preraining requires more memory after restoring from checkpoint (26.04.00)
### Problem

The same pretraining script requires more memory when training is conducted after restoring from a checkpoint as opposed to pretraining from scratch. 

Attached a simple training recipe to test on H100 and logs. 
1) Run training on a single H100, wait for a checkpoint to be saved, interrupt the run.
2) Run the same script again. The training should fail with torch.OutOfMemoryError: CUDA out of memory.

[01_from_scratch.log](https://github.com/user-attachments/files/27208807/01_from_scratch.log)
[02_restore_from_checkpoint.log](https://github.com/user-attachments/files/27208806/02_restore_from_checkpoint.log)
[repro.py](https://github.com/user-attachments/files/27208805/repro.py)

### Minimal repro

```shell
1. python repro.py |& tee 01_from_scratch.log, cancel after first saving the checkpoint at 10 steps
2. python repro.py |& tee 02_restore_from_checkpoint.log. Checkpoint after 10 steps will be loaded and the job should with OOM on the first training step
```

### Expected behavior

Training when restoring from checkpoint requires the same amount of GPU memory.

### Affected area

area:ckpt

### Regression?

Not sure

### Environment

nvcr.io/nvidia/nemo:26.04.00 container, single H100

### Logs

```shell

```

cuichenx:
## What

Fix `DistributedOptimizer` OOM when resuming from a checkpoint.

When resuming from a checkpoint, `DistributedOptimizer.load_state_dict()` calls
`self.optimizer.load_state_dict()` (PyTorch's native Adam), which casts all Adam
state tensors (`exp_avg`, `exp_avg_sq`) to match the parameter device (GPU).
These tensors then sit on GPU throughout the first forward pass, causing a peak
GPU memory spike that doesn't occur during fresh training — and can cause OOM.

In fresh training, Adam states are allocated lazily inside the first
`optimizer.step()` call — after the forward and backward passes have freed
activations. The resumed case should match this behaviour exactly.

## Root cause

```
sharded_state_dict(is_loading=True)
  └── self.load_state_dict(self.state_dict())
        └── self.optimizer.load_state_dict(...)   # ← casts all state to GPU
```

The GPU allocation happens before training starts, so the first forward pass
runs with: model params + FP32 main params + Adam states (2× FP32 params) all
on GPU simultaneously — much higher than fresh training where Adam states aren't
present yet.

## Fix

Two changes to `distrib_optimizer.py`:

1. **After `self.optimizer.load_state_dict()`**, immediately offload all Adam
   state tensors to CPU pinned memory. This reduces GPU footprint back to the
   fresh-training baseline before the first forward pass.

2. **`_ensure_adam_states_on_gpu()`** — new method called at the start of
   `step_with_ready_grads()`. Restores CPU tensors to GPU just before the first
   Adam step (i.e., after activations have been freed). On subsequent steps all
   tensors are already on GPU, so this is a cheap no-op.

## Verification

Tested with Llama 3.2 1B (4 layers) and Qwen3 1.7B (4 layers), measuring
step-1 peak GPU memory above the pre-step baseline:

| | Fresh | Resume (before fix) | Resume (after fix) |
|---|---|---|---|
| Llama 3.2 1B | 4.59 GB | ~10 GB | 3.82 GB ✓ |
| Qwen3 1.7B   | same pattern | OOM in original report | matched ✓ |

Peak == current in both cases after fix, confirming Adam states are not on GPU
during the forward pass.

Fixes: NVIDIA-NeMo/Megatron-Bridge#3576

### DIFF
--- a/megatron/core/optimizer/distrib_optimizer.py
+++ b/megatron/core/optimizer/distrib_optimizer.py
@@ -651,6 +651,9 @@ class DistributedOptimizer(MixedPrecisionOptimizer):
             log_config_to_disk(config, locals(), prefix=type(self).__name__)
 
         super().__init__(optimizer, config, grad_scaler, init_state_fn)
+        # True when load_state_dict() has pinned Adam states to CPU; cleared after the first
+        # optimizer step restores them to GPU via _ensure_adam_states_on_gpu().
+        self._has_cpu_adam_states = False
         self.model_chunks = model_chunks
         self.ddp_config = self.model_chunks[0].ddp_config
         for model_chunk in self.model_chunks:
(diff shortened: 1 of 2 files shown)


deepakn94:
What's the point of this? You need to allocate space for the optimizer state anyway eventually? I would actually argue that we should not lazily allocate Adam state for the "start from initialization" case if we want feature parity...

cuichenx:
> What's the point of this? You need to allocate space for the optimizer state anyway eventually? I would actually argue that we should not lazily allocate Adam state for the "start from initialization" case if we want feature parity...

Actually you're right, this PR works by reordering the cache allocator rather than actually reducing memory. The customer issue can also be resolved by setting `expandable_segments:True`. Closing

Issue #4641, opened 2026-05-06
[BUG] Muon + PP + MTP: tied word_embeddings on MTP-only last PP stage is routed to Muon instead of Adam


yezhengmao1:
[@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) When enabling Muon optimizer together with pipeline parallel (PP) and Multi-Token Prediction (MTP), the tied word_embeddings.weight replicated on the last PP stage for the MTP module is incorrectly routed to the Muon optimizer, while the original embedding on PP rank 0 stays on Adam (the default routing for embedding/output parameters).

This produces two symptoms:

* Divergent optimizer updates across the tied replicas. PP rank 0 updates the embedding via Adam; the last PP stage updates its tied copy via Muon. The two replicas drift apart between embedding-grad all-reduces, which corrupts training.
* Orphan optimizer state at checkpoint time. The Muon-owned replica on the last PP stage has no main replica at checkpoint save/load (the embedding group only expects an Adam-owned counterpart), causing save/load mismatches.

Root cause: 
in https://github.com/NVIDIA/Megatron-LM/blob/c325855e47e2700e6f64a4b2b09911c01e6fd7f8/megatron/core/models/common/language_module/language_module.py#L205  the `is_embedding_or_output_parameter` attributes are only set when `self.pre_process is true`. 

On an MTP-only last PP stage (where `pre_process=False` but an MTP module holds a tied `word_embeddings`), the param is left untagged, so the Muon param-group routing rule (which excludes embedding/output params) does not exclude it, and Muon claims it.

PR: https://github.com/NVIDIA/Megatron-LM/pull/4642

janEbert:
Thanks a lot for the bug report and PR!

janEbert:
Fixed by #5034.

PR #4642, opened 2026-05-06
Fix incorrect optimizer routing for tied word_embeddings.weight on the MTP-only last PP stage when Muon + PP + MTP are enabled together.

yezhengmao1:
## Root cause

In `LanguageModule`, the `is_embedding_or_output_parameter`  tags are only set when `self.pre_process is true`. On an MTP-only last PP stage (`pre_process=False` but holds a tied word_embeddings replica via MTP), the param is left untagged. 

Muon's param-group routing then claims it, while PP rank 0 keeps its replica on Adam — causing divergent updates and orphan optimizer state at checkpoint time.

Also see: https://github.com/NVIDIA/Megatron-LM/issues/4641

## Fix

Also tag the embedding param on MTP-process stages so it is grouped with the PP=0 replica and stays on Adam


yashaswikarnati (approved):

yaoyu-33 (approved):

jaredcasper (approved):

jaredcasper:
Note that it is probably not a good idea to use Muon on embeddings.

dimapihtar (approved):
LGTM. Thank you!

janEbert:
Hey @yezhengmao1, thanks a lot for the useful contribution. I'm sorry we didn't merge this earlier.
By now, the much later https://github.com/NVIDIA/Megatron-LM/pull/5034 has been merged with the same change, so I'm closing this. Sorry that the duplicate was merged instead of your earlier solution.

PR #4697, opened 2026-05-08
Simple and stable Inference APIs

YangFei1990:
# What does this PR do ?

## Motivation And Goals

The current Megatron inference APIs expose many internal building blocks:

- `InferenceConfig`
- `DynamicInferenceContext`
- `GPTInferenceWrapper`
- `TextGenerationController`
- `DynamicInferenceEngine`
- `InferenceClient`
- `DataParallelInferenceCoordinator`
- dynamic text generation server helpers

This is powerful, but it makes simple usage verbose. A user who wants to run offline generation or serve requests must understand engine construction, context selection, tokenizer setup, coordinator lifecycle, and per-rank behavior.

## APIs

Inspired by vLLM, we propose two dimensions for the API design.

### Sync and Async APIs

We propose two major APIs for Megatron inference:

- `MegatronLLM`: synchronous offline inference API. Calls block for final
  outputs.
- `MegatronAsyncLLM`: asyncio-native generation, online serving (OpenAI
  compatible).

Both classes support offline inference, lifecycle control (`pause`/`unpause`/`suspend`/`resume`), and access to the underlying engine for expert use. The differentiators are: `MegatronAsyncLLM` exposes async methods and the online HTTP server (`serve(...)`); `MegatronLLM` exposes sync methods.

The underlying primitive APIs can also be accessed through corresponding
property attributes (`engine`, `context`, `controller`).

### Coordinator

Both sync and async APIs support direct mode and coordinator mode, specified by the `use_coordinator` argument in the API constructor. We also provide an `is_primary_rank` property to help users understand which rank should feed data and collect outputs.

Without coordinator, all ranks are treated as user-managed ranks, and users need to handle load balancing between different DP/EP ranks. Every rank's `is_primary_rank` returns true: the API does not decide which rank should receive which prompts or which rank should emit output. Users must split data across different DP/EP ranks, ensure consistent inputs across TP/PP/CP ranks, and gather/aggregate results from different DP/EP ranks. If users do not shard inputs correctly, they may duplicate work or violate TP/PP/EP/DP group expectations.

With coordinator, the coordinator manages load balancing. Users feed data on the coordinator (primary) rank and collect output on that rank. `is_primary_rank` returns true only on the coordinator rank, which is global rank 0. Online serving mode requires `use_coordinator=True` when DP/EP size is greater than 1.

Lifecycle methods (`pause`/`unpause`/`suspend`/`resume`) are only meaningful in coordinator mode. They raise `RuntimeError` in direct mode.

## Examples
Here we list some common examples, for details check `examples/inference`

### Offline Sync Generation With Coordinator

```python
from megatron.inference import MegatronLLM, SamplingParams

llm = MegatronLLM(
    ...,
    use_coordinator=True,
)

if llm.is_primary_rank:
    outputs = llm.generate(prompts, SamplingParams(num_tokens_to_generate=128))
    for output in outputs:
        print(output.generated_text)

llm.shutdown()
```

### Concurrent Async Generation With Multiple Prompts

```python
from megatron.inference import MegatronAsyncLLM, SamplingParams

llm = MegatronAsyncLLM(
    ...,
    use_coordinator=True,
    coordinator_host="10.0.0.1",
    coordinator_port=6000,
)

if llm.is_primary_rank:
    sampling_params = SamplingParams(num_tokens_to_generate=64)
    results = await llm.generate(prompts, sampling_params)
    for result in results:
        print(result.generated_text)

await llm.shutdown()
```
### Programmatic OpenAI-Compatible Server

```python
from megatron.inference import MegatronAsyncLLM, ServeConfig

llm = MegatronAsyncLLM(
    ...,
    use_coordinator=True,
    coordinator_host="10.0.0.1",  # Internal/routable host for coordinator ZMQ.
    coordinator_port=6000,
)

# All ranks enter llm.serve, but only the primary rank hosts the HTTP server.
# `blocking=True` (default) keeps serve() awaiting until the server stops.
await llm.serve(
    ServeConfig(
        host="0.0.0.0",  # HTTP bind host.
        port=5000,
    ),
)

# Users can send OpenAI-compatible requests to the primary rank's HTTP endpoint.
# For example, from another process:
#
# from openai import OpenAI
#
# client = OpenAI(api_key="EMPTY", base_url="http://<primary-host>:5000/v1")
# response = client.chat.completions.create(
#     model="megatron-gpt",
#     messages=[{"role": "user", "content": "Explain Megatron inference."}],
#     max_tokens=128,
#     temperature=0.8,
#     top_p=0.95,
#     extra_body={"top_k": 40},  # Sent to the server as top-level {"top_k": 40}.
# )
# print(response.choices[0].message.content)
#
# Equivalent raw HTTP request:
#
# curl http://<primary-host>:5000/v1/chat/completions \
#   -H "Content-Type: application/json" \
#   -d '{
#     "model": "megatron-gpt",
#     "messages": [{"role": "user", "content": "Explain Megatron inference."}],
#     "max_tokens": 128,
#     "temperature": 0.8,
#     "top_p": 0.95,
#     "top_k": 40
#   }'
```

## PR review
The major files to review are newly added examples in `examples/inference` and the high level implementations in `megatron/inference`, the rest are most test coverage and doc changes.

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


YangFei1990:
/ok to test [5f651b7](https://github.com/NVIDIA/Megatron-LM/pull/4697/commits/5f651b710d50c9f162320de0d95159091d97472e)

YangFei1990:
/ok to test [963f663](https://github.com/NVIDIA/Megatron-LM/pull/4697/commits/963f663635febbffcf7eaf81c80f50249571e0f3)

megatron/core/inference/apis/_llm_base.py
```diff
@@ -0,0 +1,419 @@
...
+                coord_runtime = _CoordinatorRuntime(
+                    engine,
+                    is_primary=is_primary_rank,
+                    coordinator_host=coordinator_host,
+                    coordinator_port=coordinator_port,
+                )
+                loop_manager.run_sync(coord_runtime.setup(loop=loop_manager.loop))
+            except BaseException:
+                loop_manager.stop()
+                raise
```
tdene, line 319: This does not teardown, it just raises. The coordinator process stays alive.

megatron/core/inference/apis/llm.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def shutdown(self) -> None:
+        """Tear down the engine and runtime. Idempotent. Direct mode is a no-op."""
+        if self._shutdown_called:
+            return
+        self._shutdown_called = True
+        if not self._use_coordinator:
+            return  # direct mode: nothing to tear down
+        assert self._loop_manager is not None
+        self._loop_manager.run_sync(self._shutdown_impl())
```
tdene, line 138: This will deadlock if we're already inside the `_loop_manager` when we call it.

megatron/core/inference/apis/_llm_base.py
```diff
@@ -0,0 +1,419 @@
...
+        if self._use_coordinator:
+            # ``add_request`` calls ``asyncio.get_running_loop().create_future()``
+            # so it must be invoked from a coroutine on the runtime loop. This
+            # coroutine runs on that same loop, so ``asyncio.gather`` over the
+            # returned futures is safe.
+            assert self._coord_runtime is not None and self._coord_runtime.client is not None
+            futures = [self._coord_runtime.client.add_request(p, sp) for p in prompts]
+            return list(await asyncio.gather(*futures))
+        # TODO: replace with an upstream ``engine.async_generate`` so direct-mode
+        # async generate doesn't block the caller's event loop.
```
tdene, line 418: I think this is a very big TODO. As-is, I don't see any benefit to using `MegatronAsyncLLM(use_coordinator=False)` over `MegatronLLM`, because it just ends up calling `generate` either way.

I agree this is a bigger chance that should be part of a follow-up PR, but this TODO should really be highlighted.

megatron/core/inference/apis/async_llm.py
```diff
@@ -0,0 +1,246 @@
...
+
+    async def wait_for_shutdown(self) -> None:
+        """Block until the engine's background loop task terminates.
+
+        No-op in direct mode.
+        """
+        if not self._use_coordinator:
+            return
+        assert self._loop_manager is not None
+        await self._loop_manager.run_async(self._wait_for_shutdown_impl())
```
tdene, line 225: I think this is unsafe to call after `shutdown`, because `shutdown` has already called `_loop_manager.stop()`. Is this a possible race condition?

megatron/core/inference/apis/llm.py
```diff
@@ -0,0 +1,153 @@
...
+    def generate(
+        self,
+        prompts: Union[str, List[int], List[str], List[List[int]]],
+        sampling_params: Optional[SamplingParams] = None,
+    ) -> List["DynamicInferenceRequest"]:
+        """Run inference for one prompt or a batch.
+
+        Returns ``list[DynamicInferenceRequest]`` in input order. Single-prompt
+        input returns a one-element list -- the always-list shape is the
+        deliberate sync-vs-async asymmetry.
```
tdene, line 63: Why do we have this deliberate asymmetry?

megatron/core/inference/apis/_llm_base.py
```diff
@@ -0,0 +1,419 @@
...
+        if (coordinator_host is not None or coordinator_port is not None) and not use_coordinator:
+            raise ValueError("coordinator_host/port require use_coordinator=True")
+
+        if inference_config is None:
+            inference_config = InferenceConfig()
+
+        # Build the engine pipeline. Mirrors examples/inference/gpt/gpt_dynamic_inference.py.
+        context = DynamicInferenceContext(model.config, inference_config)
+        # TODO: extend for non-GPT models in a future iteration.
+        wrapper = GPTInferenceWrapper(model, context)
```
tdene, line 282: The primary models NVIDIA uses are non-GPT models.

I think this entire bloc can perhaps be simplified by using `megatron.inference.get_model_for_inference()`?

megatron/inference/_llm_base.py
```diff
@@ -0,0 +1,419 @@
...
+class _EventLoopManager:
+    """Per-instance background daemon thread + persistent asyncio event loop.
+
+    Bridges sync and async user-thread callers to coroutines that run on the
+    background loop via ``asyncio.run_coroutine_threadsafe``.
+    """
+
+    def __init__(self) -> None:
+        self._loop: asyncio.AbstractEventLoop | None = None
+        self._thread: threading.Thread | None = None
```
tdene: Use either `Optional` or `|`, not both in the same file.

megatron/inference/_llm_base.py
```diff
@@ -0,0 +1,419 @@
...
+            loop.run_forever()
+
+        self._thread = threading.Thread(target=_run_loop, daemon=True)
+        self._thread.start()
+
+        # Wait for the loop to be created and running before returning so
+        # callers can use ``submit`` immediately. Mirrors NeMo RL's polling
+        # approach.
+        while self._loop is None or not self._loop.is_running():
+            time.sleep(0.001)
```
tdene: Can you use an Event and await it instead?

examples/inference/offline_inference.py
```diff
@@ -0,0 +1,278 @@
...
+                results = await llm.generate(prompts_list, sampling_params)
+                torch.cuda.synchronize()
+                total_time = get_curr_time() - t
+
+                total_output_tokens = sum(len(r.generated_tokens) for r in results)
+                throughputs.append(total_output_tokens / total_time)
+            captured = _capture_engine_stats(llm)
+
+    peak_mem_stats = get_global_peak_memory_stats_bytes()
+    _report_results(args, setup_prefix, results, throughputs, total_time, peak_mem_stats, captured)
```
tdene, line 235: Can there be code re-use here?

examples/inference/launch_inference_server.py
```diff
@@ -0,0 +1,106 @@
...
+    args = get_args()
+
+    # Match the legacy tool's NVTX gating.
+    if args.profile and args.nvtx_ranges:
+        configure_nvtx_profiling(True)
+
+    # Required for lm-eval loglikelihood compatibility: keeps prompt logits
+    # materialized so echo=True / logprob requests work end-to-end. Matches
+    # tools/run_dynamic_text_generation_server.py.
+    args.return_log_probs = True
```
tdene: It feels wrong to bury this here. The right place to put it would be inside the default setting for this argument, just changing it from `False` to `True`. There can definitely be use-cases that want to launch inference with `return_log_probs` set to `False` in order to save memory.

tdene (commented):
A couple of significant issues, and some more minor ones.

megatron/core/inference/apis/_llm_base.py
YangFei1990: I think that is feasible to wrap the model in `megatron.inference.get_model_for_inference()`, but what do you mean primary models NVIDIA uses are non-GPT models?

megatron/core/inference/apis/_llm_base.py
YangFei1990: Yes, currently it is for completeness that both `MegatronAsyncLLM` and `MegatronLLM` support coordinator and direct mode, so user can use with/without asyncio freely. In general I think most common use case would be that `MegatronLLM` runs in direct mode and `MegatronAsyncLLM` runs in coordinator mode. 
We do not have a contract to claim that all 5 modes (including server) to achieve the same perf, so I did not prioritize this. But if you do see strong use cases for async + direct, we can highlight this.

megatron/core/inference/apis/async_llm.py
YangFei1990: This `wait_for_shutdown` together with the `_wait_for_shutdown_impl` is mainly defined for the non-primary ranks, where with the coordinator mode the primary rank will do the generate / serve call, so the non-primary ranks need to wait, they can call this. 

Currently most of the wait logic is handled inside `shutdown` ready, so this function is only used in the `serve` when `blocking` is true. I do not see a case when we could call this after `shutdown`, let me know if I missed anything

megatron/core/inference/apis/llm.py
YangFei1990: That is to align with vLLM (and also Tinker's sample APIs). vLLM's async generate does not take batch, it always take single input and output a single element. Our async generate takes both batch and single input, so I want to make the single input behaves like this.

megatron/core/inference/apis/llm.py
YangFei1990: Could you elaborate? This is the sync API why would we be in an event loop?

tdene: You're right: we should not be in an event loop. But I can see a scenario where the customer accidentally misuses the API.

It's a nit, and feel free to disagree, but I think it's worth checking here if we are inside the event loop to prevent the customer from experiencing a hang.

megatron/core/inference/apis/_llm_base.py
tdene: No, I fully agree with you that this is not a priority. I see no use case.

I just feel like this is such a significant TODO that it should be documented in a more visible place as well, if possible.

megatron/core/inference/apis/async_llm.py
tdene: That's fair, I just meant to ask the question. I personally also could not find a race condition, but I do not understand the code as well as you.

megatron/core/inference/apis/llm.py
tdene: Understood!

megatron/core/inference/apis/_llm_base.py
tdene: Sorry, I brought up two separate topics.

The first topic is just a question of whether most of this surrounding code can go away if we use `megatron.inference.utils.get_model_for_inference()` or `get_dynamic_inference_engine()`. And now that I look at it again, I believe the answer is a clear *no*, there is no code reuse to be had here.

The second topic is that this hardcodes to `GPTInferenceWrapper`, when NVIDIA mainly cares about hybrid models... and on this point I am completely wrong, apologies. I got confused between inference code and training code. `GPTInferenceWrapper` applies to both hybrid and non-hybrid model.

tdene: Resolving; this entire comment was a hallucination on my side.

megatron/inference/_llm_base.py
```diff
@@ -0,0 +1,419 @@
...
+    ) -> None:
+        if (coordinator_host is not None or coordinator_port is not None) and not use_coordinator:
+            raise ValueError("coordinator_host/port require use_coordinator=True")
+
+        if inference_config is None:
+            inference_config = InferenceConfig()
+
+        # Build the engine pipeline. Mirrors examples/inference/gpt/gpt_dynamic_inference.py.
+        context = DynamicInferenceContext(model.config, inference_config)
+        # TODO: extend for non-GPT models in a future iteration.
```
shanmugamr1992: Given that GPTInferenceWrapper addresses all models (hybrid models as well), can you get rid of this comment so that it snot confusing.

shanmugamr1992:
@YangFei1990  lets rename legacy to advanced , and add a little about why it exists in the doc (Saying it can be sued to customize the individual compoentnts ) e.g if you want to change tokenization/sampling you could overwrite text gen controller and use that etc.

examples/inference/README.md
```diff
@@ -1,288 +1,112 @@
...
-Refer to [gpt_inference_wrapper.py](../../megatron/core/inference/model_inference_wrappers/gpt/gpt_inference_wrapper.py) for an example of implementing this for GPTModel.
-
-<br>
+Send requests with any OpenAI-compatible client. The dynamic server
+currently returns `"model": "EMPTY"` and does not validate the request
+`model` field — pass anything you like.
 
-##### 3.3. Modify Inference Parameters
-We use  [common inference params](../../megatron/core/inference/sampling_params.py) for text generation. Customize this to change `top_p`, `top_k`, number of tokens to generate etc. Other attributes can be added for the inference loop as shown below.
+### Legacy examples
```
shanmugamr1992: Lets change ethis to advanced

examples/inference/README.md
```diff
@@ -1,288 +1,112 @@
...
-##### 1.2 Running The Code
-An example Slurm script is shown below. Set the tokenizer paths, inference params, and other settings appropriately. 
-
-For a recap on sampling parameters, refer to [this blog](https://ivibudh.medium.com/a-guide-to-controlling-llm-model-output-exploring-top-k-top-p-and-temperature-parameters-ed6a31313910).
 
+- [What's in here](#whats-in-here)
+- [Offline inference](#offline-inference)
+- [OpenAI-compatible inference server](#openai-compatible-inference-server)
+- [Legacy examples](#legacy-examples)
+- [See also](#see-also)
```
shanmugamr1992, line 10: Lets also add a TODO (Or upcomign section) to add the other thigns you mentioned like streaming, CLI etc.

shanmugamr1992:
@YangFei1990 one more very important thing is to move the async llm , llm etc into megatron/core/inference (Coz thats the pip package). so anyone installing mcore pip package wont have access to the apis

shanmugamr1992:
@YangFei1990  add assert to make sure EP works only with coordinator mode

docs/llama_mistral.md
```diff
@@ -314,14 +304,6 @@ python Megatron-Bridge/examples/conversion/convert_checkpoints.py import \
 
 After this conversion, we are ready to load the checkpoints into a Megatron GPT model.
 
-## (Optional) Validate checkpoints
-
-A Megatron-LM text generation server for Mistral-7B can be launched using the script `examples/inference/llama_mistral/run_text_generation_mistral.sh <PATH_TO_CONVERTED_MCORE_CHECKPOINT> <PATH_TO_DOWNLOADED_HUGGINGFACE_CHECKPOINT>`.
-
-Once running, query the server with `curl 'http://<TEXT_GENERATION_SERVER_IP>:5000/api' -X 'PUT' -H 'Content-Type: application/json; charset=UTF-8'  -d '{"prompts":["<SOME_PROMPT>"], "tokens_to_generate":100, "top_k":1}'`.
-
-A reference generation for comparison can be obtained from the Huggingface transformers library by running `python examples/inference/llama_mistral/huggingface_reference.py --model_path <PATH_TO_DOWNLOADED_HUGGINGFACE_CHECKPOINT> --prompt <SOME_PROMPT>`.
```
shanmugamr1992, line 323: This is what @santhnm2  was talking about. Why these are removed ?

YangFei1990: This is because we no longer have inference example for the llama_mistral.

examples/inference/offline_inference.py
YangFei1990: Do you mean the `_run_sync` and `_run_async`? If so I actually keep them separate intentionally, as this is an tutorial to help user to understand how to use these APIs, so I thought it would be good to make it explicit for the core loop logic. Let me know your thoughts.

tdene: That's fair!

YangFei1990:
/ok to test [a3aade9](https://github.com/NVIDIA/Megatron-LM/pull/4697/commits/a3aade9d2fb6ced6e49414e24d892ea680035cb7)

shanmugamr1992 (approved):

megatron/core/inference/apis/__init__.py
santhnm2, line 1: Nit: Can we make this folder `megatron/core/inference/api` instead of `megatron/core/inference/apis`?

megatron/core/inference/apis/_llm_base.py
```diff
@@ -0,0 +1,463 @@
...
+    ``model`` must be in eval mode before construction; this class does not
+    modify the model state.
+    """
+
+    def __init__(
+        self,
+        *,
+        model,
+        tokenizer,
+        inference_config: Optional[InferenceConfig] = None,
```
santhnm2, line 259: Why is this allowed to be None?

megatron/core/inference/apis/_llm_base.py
```diff
@@ -0,0 +1,463 @@
...
+
+        # Build the engine pipeline. Mirrors examples/inference/gpt/gpt_dynamic_inference.py.
+        context = DynamicInferenceContext(model.config, inference_config)
+        wrapper = GPTInferenceWrapper(model, context)
+        controller = TextGenerationController(inference_wrapped_model=wrapper, tokenizer=tokenizer)
+        engine = DynamicInferenceEngine(controller=controller, context=context)
+
+        if use_coordinator:
+            # Lazy import so the module imports cleanly without torch installed.
+            import torch.distributed as dist
```
santhnm2: I think torch should be a required prerequisite, can we remove this?

tests/unit_tests/inference/high_level_api/test_async_llm_serve_guard.py
tdene, line 1: We don't need an entire file just for this single specific test.

tests/unit_tests/inference/high_level_api/test_event_loop_manager.py
```diff
@@ -0,0 +1,119 @@
...
+import pytest
+
+from megatron.core.inference.apis._llm_base import _EventLoopManager
+
+
+class TestEventLoopManager:
+    def test_loop_property_raises_before_start(self):
+        mgr = _EventLoopManager()
+        with pytest.raises(RuntimeError, match="start"):
+            _ = mgr.loop
```
tdene: I don't see how this test is useful.

tests/unit_tests/inference/high_level_api/test_event_loop_manager.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def test_start_is_idempotent_and_loop_accessible(self):
+        mgr = _EventLoopManager()
+        mgr.start()
+        try:
+            loop1 = mgr.loop
+            mgr.start()  # double-start should be a no-op
+            loop2 = mgr.loop
+            assert loop1 is loop2
+            assert loop1.is_running()
```
tdene: I don't see how this test is useful.

tests/unit_tests/inference/high_level_api/test_event_loop_manager.py
```diff
@@ -0,0 +1,119 @@
...
+        finally:
+            mgr.stop()
+
+    def test_submit_run_sync_run_async(self):
```
tdene: This seems to be testing underlying functionality of Python itself, instead of any of the code actually in this class.

tests/unit_tests/inference/high_level_api/test_event_loop_manager.py
```diff
@@ -0,0 +1,119 @@
...
+            assert mgr.run_sync(coro()) == 42
+
+            # run_async awaits the future from another loop
+            import asyncio
+
+            assert asyncio.run(mgr.run_async(coro())) == 42
+        finally:
+            mgr.stop()
+
+    def test_stop_is_idempotent_and_joins_thread(self):
```
tdene: I don't believe this tests much.

tests/unit_tests/inference/high_level_api/test_event_loop_manager.py
```diff
@@ -0,0 +1,119 @@
...
+
+        monkeypatch.setattr(torch.cuda, "is_available", lambda: False)
+        monkeypatch.setattr(torch.cuda, "set_device", lambda d: called.append(d))
+
+        mgr = _EventLoopManager()
+        mgr.start()
+        try:
+            assert called == []
+        finally:
+            mgr.stop()
```
tdene: I don't understand the use of these tests. They are just checking if `torch.cuda.set_device(parent_device)` works?

tests/unit_tests/inference/high_level_api/test_lifecycle_guards.py
tdene, line 1: We do not need 4 separate tests in this class; they could all be one single, parameterized, test. That kind of test would be more readable, easier to update, and more thorough.

tests/unit_tests/inference/high_level_api/test_normalize_prompts.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def _normalize(prompts):
+    """Bypass __init__ (which builds the engine pipeline) and call the method
+    on a bare instance."""
+    obj = _MegatronLLMBase.__new__(_MegatronLLMBase)
+    return obj._normalize_prompts(prompts)
+
+
+class TestNormalizePrompts:
```
tdene: All of these tests should be combined into 1 single test function.

tests/unit_tests/inference/high_level_api/test_serve_config.py
tdene, line 1: These tests do not test anything.

tests/unit_tests/inference/high_level_api/test_validation.py
tdene, line 1: These tests should be merged.

tdene (commented):
For the tests, in general, they should all be merged into a single file. I believe that all of the current tests can be covered by only 1 or 2 individual (parameterized) tests. Keeping them spread out across multiple files and classes like this makes it difficult to maintain, understand, and read.

YangFei1990:
/ok to test [1bbdada](https://github.com/NVIDIA/Megatron-LM/pull/4697/commits/1bbdada6aa02582b70dc956597587bb2201c6b7d)

tdene (approved):

YangFei1990:
/ok to test [ec780e5](https://github.com/NVIDIA/Megatron-LM/pull/4697/commits/ec780e58956c4d2ffea9ae9678ecb5122b0c0696)

wdykas (approved):

megatron/core/inference/apis/async_llm.py
```diff
@@ -0,0 +1,232 @@
...
+    """Async high-level inference API for Megatron.
+
+    Asyncio-native wrapper over the shared engine + runtime managed by
+    :class:`_MegatronLLMBase` -- see that class for execution modes
+    (direct vs coordinator), caller responsibilities, and the
+    ``model.eval()`` contract.
+
+    On top of the base, this class provides:
+
+    - ``async generate`` accepting single or batched prompts. In direct mode
```
lmcafee-nvidia: direct mode is not support, given the `ValueError` below. we should update this docstring

megatron/core/inference/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+- **`llm.engine.reset()` is unsafe in coordinator mode.** Two failure modes, both upstream in `dynamic_engine.py`:
+  - *Deadlock*: `reset()` *rebinds* (does not mutate in-place) `_cond` / `_state_events`. Any coroutine on the engine-loop task that is `await`ing one of those primitives holds a reference to the OLD object in its suspended frame. Subsequent `notify_all()` / `set()` calls hit the NEW objects, leaving the suspended waiter stranded; the next `generate()` hangs.
+  - *Silent corruption*: `reset()` also sets `self.use_coordinator = False`, which silently re-routes failed-request handling, scheduling notification, and `suspend()`'s state machine to direct-mode branches. Outcome: not-a-hang but wrong behavior, harder to diagnose.
+  - The example `offline_inference.py` blocks `--inference-repeat-n > 1` with `--use-coordinator` for these reasons. Direct-mode reset is safe.
+
+- **Async-direct `generate()` is single-caller.** Concurrent `await llm.generate(...)` (e.g. via `asyncio.gather`) in direct mode raises `RuntimeError`. Pass batched prompts instead, or switch to coordinator mode.
```
lmcafee-nvidia: direct mode isn't supported, so we should update this line

megatron/core/inference/apis/_llm_base.py
YangFei1990: We have default values defined (just in the same `__init__` function), so if user does not pass anything we use default values. But in our practical examples we have feed it.

megatron/core/inference/apis/__init__.py
YangFei1990: Could you share the motivation? Technically there are more than one api.

santhnm2: True, but in my opinion the distinction between `MegatronLLM` and `MegatronAsyncLLM` is not substantial enough to call them separate APIs, to me these are more like two instantiations of the same high-level interface. I think calling it `apis` implies there will be a series of APIs that get added over time rather than 1 single-source-of-truth API which I think is closer to the intent. But this is a very minor point so I'm ok with `apis` too.

santhnm2 (approved):

lmcafee-nvidia (approved):

kvareddy (approved):

shanmugamr1992:
/ok to test [7b513a0](https://github.com/NVIDIA/Megatron-LM/pull/4697/commits/7b513a0f44871644ad09cfa47ae17053420a9785)

YangFei1990:
/ok to test [74eb926](https://github.com/NVIDIA/Megatron-LM/pull/4697/commits/74eb9263c4a1b8d63238cb863af55155df2ac3c3)

YangFei1990:
/ok to test [0b44fb8](https://github.com/NVIDIA/Megatron-LM/pull/4697/commits/0b44fb822a5adfcbee0a400c22902182c9d35f59)

ko3n1g (approved):

ko3n1g (changes requested):
This will require changes in the NeMo side. Can we work with @oyilmaz-nvidia before the merge?

ko3n1g (approved):
Thanks for sharing more background in our offline discussion, all good now

YangFei1990:
/ok to test [4e3269d](https://github.com/NVIDIA/Megatron-LM/pull/4697/commits/4e3269d4ecf832f5d03950f9d7d20474247ee012)

YangFei1990:
/ok to test [87ba21c](https://github.com/NVIDIA/Megatron-LM/pull/4697/commits/87ba21c3bcc4f6a5095c58e07efe60575f74bfd3)

PR #4717, opened 2026-05-11
Preserve non-FSDP-unit buckets across AllGatherPipeline reset

wujingyue:
## Summary

- Preserve non-FSDP-unit parameter buckets across `AllGatherPipeline.reset()` by default. Non-unit parameters may be read across module boundaries by fused kernels or wrapper modules, so there is no safe generic point where their gather scratch can be freed and later guaranteed to be re-gathered before use.
- Add `BucketStatus.PRESERVED` to distinguish retained bucket storage that is not ready for AG use from truly empty/deallocated buckets. A later all-gather refreshes preserved non-unit bucket storage in place.
- Keep a full-release path as `reset(preserve_non_fsdp_units=False)` for debugging only, when the model will not be reused.

## Root Cause

The Mamba repro fails on iteration 2 when `MambaMixer.forward` reads `self.conv1d.weight` directly through the fused `causal_conv1d_fn(weight=self.conv1d.weight, ...)` path instead of calling `self.conv1d(...)`.

Two otherwise reasonable behaviors combine into an unsafe dependency on opportunistic prefetch:

1. `MambaMixer.conv1d`'s pre-forward hook never fires. Since `nn.Module.__call__` is not invoked on the `conv1d` submodule, FSDP's `_pre_forward_param_unshard` hook for `conv1d` does not run. The parent `MambaMixer` hook does run, but for non-FSDP-unit modules it gathers only `module.parameters(recurse=False)`, so it covers direct `MambaMixer` params such as `A_log`, `D`, `dt_bias`, and `norm.weight`, not `conv1d.weight` in the child module's bucket.

   ```
   MambaMixer (M) -> Conv1D (M) -> weight (P)
     |                                 ^
     +-> MambaContextParallel(non-M) --+
   ```

2. Prefetch masks that missing gather on the first iteration but can disappear on the second. The first pre-forward hook can prefetch enough adjacent buckets to incidentally gather `conv1d.weight`'s bucket. On the failing repro, however, the input has `requires_grad=False`, so the FSDP unit can miss post-backward cleanup and remain in `PRE_BACKWARD`. FSDP intentionally disables forward prefetch in `PRE_BACKWARD` to avoid speculative gathers during activation recomputation. On iteration 2, that means no prefetch chain gathers `conv1d.weight`.

Before this PR, `finalize_model_grads()` eventually called `synchronize_param_gather()`, which called `AllGatherPipeline.reset()` and freed every bucket, including non-FSDP-unit buckets. With the `conv1d.weight` bucket freed and no guaranteed hook/prefetch path to re-gather it on iteration 2, the fused convolution kernel could dereference freed storage and hit a CUDA illegal memory access.

Setting `requires_grad=True` on the input masks the failure by restoring post-backward cleanup and re-enabling prefetch. That makes the repro pass, but it does not provide a reliable gather contract for raw cross-module parameter reads.

## Fix

- Preserve non-FSDP-unit bucket storage during the normal reset path.
- Mark preserved buckets as `PRESERVED` so retained storage is not treated as freshly gathered or as deallocated.
- Continue releasing FSDP-unit-owned buckets through the normal `can_be_released` path.

This fixes the immediate illegal memory access at the bucket-reset step. The underlying Mamba pattern and FSDP hook/prefetch interaction remain worth tightening separately.

## Golden Values

- Bumped `mem-allocated-bytes` in `tests/functional_tests/test_cases/moe/deepseek_proxy_fsdp_ep2_fsdp2_1node/golden_values_dev_dgx_gb200.json` (~594 MB -> ~800 MB) and `tests/functional_tests/test_cases/moe/deepseek_proxy_fsdp_ep2_fsdp2_ep_overlap/golden_values_dev_dgx_h100.json` (~407 MB -> ~513 MB).
- The increase is expected: this PR keeps non-FSDP-unit bucket gather scratch persistently allocated. This affects idle-time/current allocation (`mem-allocated-bytes`), not peak allocation (`mem-max-allocated-bytes`), so it is less concerning than a peak-memory regression.

## Test Plan

- [x] Reproduce the previously failing Mamba illegal-memory-access scenario and confirm it no longer triggers.
- [x] Local targeted regression: `/opt/venv/bin/python -m torch.distributed.run --standalone --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/megatron_fsdp/test_mcore_fully_sharded_data_parallel.py::TestFsdpNonUnitBucketPreservation::test_non_unit_bucket_preserved_across_param_sync`
- [ ] CI `Run tests` for the latest head, triggered with `/ok to test` on May 19, 2026.


cspades (commented):
Tested this (without CG) and it works even past ToT, could you elaborate on the exact RCA, basically our reset (which I assume happens at the end of a global training step) destroys non-units that we would like to keep around, and Mamba had a weight that Megatron-FSDP cannot re-gather?

Would be great if you can provide a bit more details to accompany this clean fix as well!

@shjwudp remind me, why do we release all buckets before the optimizer step?

wujingyue:
> basically our reset (which I assume happens at the end of a global training step) destroys non-units that we would like to keep around, and Mamba had a weight that Megatron-FSDP cannot re-gather

That's correct, but let me provide more details after lunch. I'm going to have a meeting in five minutes.

wujingyue:
First of all, the way MambaMixer is structured invites all sorts of corner cases. self.conv1d.weight is technically owned by the self.conv1d submodule, but MambaMixer accesses it directly, bypassing the submodule abstraction entirely. While the description in #4467 is not quite accurate, we are seeing similarly unusual usage patterns in practice. It’s probably worth cleaning this up in a separate PR, although I’m waiting for a bit more motivation before diving into that.

For this particular PR and the added repro, the illegal memory access in `MambaMixer.forward` on iteration 2 (PR #4467) comes from the interaction of two behaviors in the code, intentional or not. Neither triggers IMA on its own but their composition leaves `conv1d.weight`'s bucket ungathered on iteration 2 after it has been freed.

  ## Behavior 1 — `MambaMixer.conv1d`'s pre-forward hook never fires, so its bucket is not guaranteed to be all-gathered before forward

  `MambaMixer.forward` reads `self.conv1d.weight` as a raw tensor via `causal_conv1d_fn(weight=self.conv1d.weight, ...)` instead of calling `self.conv1d(input)` (a deliberate choice — the fused kernel is much faster than `nn.Conv1d`'s general path). Because `nn.Module.__call__` is never invoked on the `conv1d` submodule, none of its registered forward hooks (including FSDP's `_pre_forward_param_unshard`) ever fire. So `conv1d.weight`'s bucket is never gathered through any
  per-iteration hook on `conv1d` itself.

  The parent `MambaMixer`'s pre-forward hook *does* fire, but for non-FSDP-unit modules it only gathers `module.parameters(recurse=False)` (`megatron_fsdp.py` line 753-757) — i.e. `MambaMixer`'s direct params `{A_log, D, dt_bias, norm.weight}`. The `recurse=False` is also deliberate: FSDP wants to minimize per-hook memory by gathering only what this module's forward will directly use. `conv1d.weight` lives in a different bucket and is excluded.

  **Net effect:** there is no per-iteration gather path that guarantees `conv1d.weight`'s bucket is allocated before `MambaMixer.forward` runs. The two design choices (raw-tensor access, `recurse=False`) are individually reasonable but together leave a gap.

  **Evidence:** instrumented `conv1d._training_state` stays at `IDLE` across the first forward — it would have flipped to `FORWARD` if its pre-forward hook had fired.

  ## Behavior 2 — The 500 MB prefetch chain that masks Behavior 1 stops working on iteration 2

  The first forward gets away with Behavior 1 because `_pre_forward_param_unshard` calls `all_gather_params(prefetch=True)`, which walks `bucket_to_bucket_group` and keeps extending the gather list up to `suggested_AG_prefetch_size = 500_000_000` (`param_and_grad_buffer.py` line 4028). The toy model in this repro fits entirely under 500 MB, so the very first pre-forward hook to fire (`attn_layer`'s) ends up prefetching every bucket — including `conv1d.weight`'s. Prefetch is, by
  design, opportunistic rather than guaranteed.

  Prefetch is suppressed when the calling module's `_training_state == PRE_BACKWARD` (`megatron_fsdp.py` line 745-749):

  ```python
  if input_training_state == TrainingState.PRE_BACKWARD:
      fsdp_forward_prefetch = False
  ```

  This is intentional: during activation recomputation a forward is re-running inside backward, and speculative gathers in that context are wasted work.

  On iteration 2, `attn_layer._training_state` is stuck at `PRE_BACKWARD` because its post-backward cleanup (`_post_backward_release_module`, line 637-657) was never installed. The reason is in `_register_post_backward_hook` (line 792-800):

  ```python
  for i, obj in enumerate(args_kwargs_list):
      if torch.is_tensor(obj) and obj.requires_grad:    # gate
          inp_tensor_indices.append(i)
          inp_tensors.append(obj)

  if len(inp_tensors) == 0:
      return args, kwargs                               # bail — no hook attached
  ```

  The post-backward hook is attached via an autograd `Function` inserted on inputs with `requires_grad=True`. The gate is necessary — there's no autograd graph to splice into otherwise — but it also means an FSDP unit receiving an input without `requires_grad` gets no post-backward cleanup. The original repro's `x = torch.randn(..., requires_grad=False)` falls in that case.

  **Result on second forward:** `attn_layer` enters its pre-forward hook with state `PRE_BACKWARD` → `fsdp_forward_prefetch = False` → no prefetch chain → `conv1d.weight`'s bucket (freed by `AllGatherPipeline.reset()` during `finalize_model_grads`) stays unallocated → `causal_conv1d_fn` dereferences freed storage → CUDA illegal memory access.

  Setting `requires_grad=True` on the input restores `attn_layer`'s post-backward cleanup, which restores prefetch on iteration 2, which incidentally rescues `conv1d.weight`'s bucket — masking Behavior 1 again.

  ### TrainingState timeline (with fix reverted)

  `requires_grad=False` on input (crashes on second forward):

  ```
  [pre-forward]                  attn_layer(TL,unit)=IDLE          mamba_layer=IDLE          mixer=IDLE          conv1d=IDLE
  [post-forward]                 attn_layer(TL,unit)=IDLE          mamba_layer=FORWARD       mixer=FORWARD       conv1d=IDLE
  [pre-backward]                 attn_layer(TL,unit)=IDLE          mamba_layer=FORWARD       mixer=FORWARD       conv1d=IDLE
  [post-backward]                attn_layer(TL,unit)=PRE_BACKWARD  mamba_layer=PRE_BACKWARD  mixer=PRE_BACKWARD  conv1d=PRE_BACKWARD
  [post-finalize_model_grads]    attn_layer(TL,unit)=PRE_BACKWARD  mamba_layer=PRE_BACKWARD  mixer=PRE_BACKWARD  conv1d=PRE_BACKWARD
  [post-second-forward]          ← CUDA illegal memory access
  ```

  `requires_grad=True` on input (passes — masks the failure):

  ```
  [pre-forward]                  attn_layer(TL,unit)=IDLE  mamba_layer=IDLE          mixer=IDLE          conv1d=IDLE
  [post-forward]                 attn_layer(TL,unit)=IDLE  mamba_layer=FORWARD       mixer=FORWARD       conv1d=IDLE
  [pre-backward]                 attn_layer(TL,unit)=IDLE  mamba_layer=FORWARD       mixer=FORWARD       conv1d=IDLE
  [post-backward]                attn_layer(TL,unit)=IDLE  mamba_layer=PRE_BACKWARD  mixer=PRE_BACKWARD  conv1d=PRE_BACKWARD
  [post-finalize_model_grads]    attn_layer(TL,unit)=IDLE  mamba_layer=PRE_BACKWARD  mixer=PRE_BACKWARD  conv1d=PRE_BACKWARD
  [post-second-forward]          attn_layer(TL,unit)=IDLE  mamba_layer=PRE_BACKWARD  mixer=PRE_BACKWARD  conv1d=PRE_BACKWARD
  ```

  The only meaningful difference is `attn_layer` at `[post-backward]`: `PRE_BACKWARD` vs. `IDLE`. That single bit flip restores prefetch on iteration 2.

  ## How they combine

  - **Behavior 1 alone** is silent as long as something else gathers the bucket.
  - **Behavior 2 alone** is silent if every module that fires a pre-forward hook on iteration 2 covers its own params (the normal case where each module's forward goes through `__call__`).
  - **Together:** Behavior 1 makes `conv1d.weight` invisible to per-iteration hooks, leaving it dependent on prefetch; Behavior 2 disables prefetch precisely on the iteration that needs it; the bucket reset between iterations turns "not gathered" into "freed storage dereferenced" → IMA.

  The PR fixes the IMA at the bucket-reset step. Behaviors 1 and 2 remain load-bearing-on-prefetch in the codebase. It's probably worth considering tightening the contract in separate PRs.

cspades:
On my radar, still need a tiny bit more time to check a few things!

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -3921,9 +3921,22 @@ def reset(self):
...
+        # already-allocated bucket_id, so no new memory is allocated.
         for bucket_id in range(self.num_buckets):
+            is_unit_bucket = self.buffer.parameter_groups[bucket_id].fsdp_unit_id is not None
             for bwd in [False, True]:
-                self.bucket_can_be_released[self.get_bucket_key(bucket_id, bwd)] = True
+                bucket_key = self.get_bucket_key(bucket_id, bwd)
+                if is_unit_bucket:
+                    self.bucket_can_be_released[bucket_key] = True
+                else:
+                    self.bucket_status[bucket_key] = BucketStatus.EMPTY
```
cspades: Nit: Setting the bucket status to EMPTY could be misleading for when the bucket is persistently allocated. If using this status, I'd add a new status `Enum`.

Important: Also, not a fan of making this change in `reset()` since it means `reset` isn't actually resetting the state of the AG, it's resetting everything except the persistently-allocated non-FSDP units. I'd replace the `reset` call in MCore with another method for deallocating FSDP units only, instead of modifying this function, or use an optional argument like `persist_non_fsdp_units: bool` or something!

wujingyue: > Also, not a fan of making this change in reset() since it means reset isn't actually resetting the state of the AG, it's resetting everything except the persistently-allocated non-FSDP units.

Agreed that the name `reset` is confusing. 

> I'd replace the reset call in MCore with another method for deallocating FSDP units only, instead of modifying this function, or use an optional argument like persist_non_fsdp_units: bool or something!

I'm not sure about this. If the contract is indeed https://github.com/NVIDIA/Megatron-LM/blob/815c83d9bcc5854030a456b66f096ae3862292f6/megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py#L4174-L4178, why should `reset` ever release non-FSDP units? cc @cspades

cspades: ~~Also BTW I didn't notice this docstring, I didn't know this was the intended use case! But yeah, let's drop this sentence in the documentation and just have an optional argument to control it!~~ (JK, don't delete this docstring it's actually informative. You could move it somewhere else or think of this as an extremely low-level function that will unconditionally destroy your buckets, thus this is a warning against that.)

cspades: Also this is the documentation for `release_bucket`, not for `reset` right? When I wrote this sentence I really just meant that during FSDP runtime this function shouldn't be called on non-units, so this comment was more like "during FSDP runtime" but `reset` function that calls this function is just clean reset of the FSDP pipeline.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -3921,12 +3929,29 @@ def reset(self):
...
+        # async_bucket_gather refreshes wbuf.data from the post-optimizer-step shard
+        # in place. TemporaryBucketAllocator.allocate() is idempotent for an already
+        # allocated bucket_id, so no new memory is allocated.
         for bucket_id in range(self.num_buckets):
+            is_unit_bucket = self.buffer.parameter_groups[bucket_id].fsdp_unit_id is not None
             for bwd in [False, True]:
-                self.bucket_can_be_released[self.get_bucket_key(bucket_id, bwd)] = True
+                bucket_key = self.get_bucket_key(bucket_id, bwd)
+                if preserve_non_fsdp_units and not is_unit_bucket:
+                    self.bucket_status[bucket_key] = BucketStatus.NEEDS_REFRESH
```
cspades: How about something slightly more general like: `BucketStatus.RESERVED`

"Needs refresh" kinda implies that we HAVE to refresh it or else it doesn't work anymore, that's not necessarily true and really is up to the optimizer / user. Sorry for being so picky, this term rubbed me the wrong way.

wujingyue: Changed to PRESERVED to be consistent with the argument name preserve_non_fsdp_units

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -3921,12 +3929,30 @@ def reset(self):
...
                 self.wait_bucket_ready(bucket_id, bwd)
+        # Unit buckets follow the normal release path: mark can_be_released and let
+        # recycle_unused_buckets free the gather scratch. Preserved non-unit
+        # buckets keep their storage pinned per the release_bucket contract
+        # (cross-module readers such as fused kernels that bypass nn.Module.forward
+        # may dereference them), but the AG pipeline should not treat them as
+        # freshly gathered. If a later all-gather targets the bucket,
+        # async_bucket_gather refreshes wbuf.data in place from the current shard.
+        # TemporaryBucketAllocator.allocate() is idempotent for an already allocated
+        # bucket_id, so no new memory is allocated.
```
cspades: Nit: This comment is a bit dense to read, for simplicity can you just say something along these lines, which I think is a little bit more succinct to how this logic affects the behavior of this method:

> If `preserve_non_fsdp_units` is set, then do not release buckets associated with FSDP non-units, instead of fully resetting the state of the pipeline.

I think the current documentation + in-line commentary elsewhere already covers why we can keep around non-units, as FSDP doesn't shard them.

cspades (approved):

cspades:
Just to check:
> Bumped mem-allocated-bytes in tests/functional_tests/test_cases/moe/deepseek_proxy_fsdp_ep2_fsdp2_1node/golden_values_dev_dgx_gb200.json (~594 MB -> ~800 MB)

This really only affects the steady-state part of the memory profile, but the FWD-BWD section should remain the same right? I didn't check where this sensor (`"mem-allocated-bytes"`) is activated, but it kind of suggests a memory ceiling increase...

cc @gautham-kollu @shjwudp @rapatel

ericharper (approved):

jaredcasper (approved):

PR #4718, opened 2026-05-11
Revert "Add Python-side guardrail for HybridEP InfiniBand limit and rename seq_len (#4094)"

ko3n1g:
## What

Revert PR #4094 ("Add Python-side guardrail for HybridEP InfiniBand limit and rename seq_len").

## Why

The guardrail at `megatron/core/transformer/moe/fused_a2a.py:398` is **overly conservative** — it raises `ValueError` for `tx_depth = 3*num_tokens + 1 >= 65536`, citing InfiniBand's `tx_depth < 65536` limit. Empirically, DeepEP / the underlying RDMA stack already handles dispatch at `tx_depth = 98305` without crashing on the affected GB300/B300 clusters; the Python pre-check rejects workloads that *would have run fine*.

## Evidence

End-to-end re-test of every test that the guardrail was preventing from running, with the revert applied:

- **Revert SHA:** [`9924c3f205`](https://github.com/NVIDIA/Megatron-LM/commit/9924c3f2050724e877f61b042f411d0506a4e707) (this PR's HEAD)
- **Test image:** `dgxctestingtemp-nemofw-nightly.50777079` (the same nightly that contains the guardrail baked-in, with the revert overlaid at runtime)
- **Pinning:** MBridge `main` (original `MBS=4/8` values, *not* halved)
- **Verification pipeline (nemo-ci internal):** [50889489](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/50889489)
- **Test cases:** all 9 tests the guardrail was rejecting:

| Test | Cluster | MBS | Result with revert |
|---|---|---|---|
| `nemotron_3_nano_8gpu_b300_bf16_perf` | bia (b300) | 4 | (still running at write time) |
| `nemotron_3_nano_8gpu_b300_fp8_mx_perf` | bia (b300) | 4 | (still running at write time) |
| `nemotron_3_nano_8gpu_gb300_bf16_perf` | lyris (gb300) | 4 | ✅ success |
| `nemotron_3_nano_8gpu_gb300_fp8_mx_perf` | lyris (gb300) | 4 | ✅ success |
| `qwen3_30b_a3b_64gpu_b300_bf16_perf` | bia (b300) | 8 | ✅ success |
| `qwen3_30b_a3b_64gpu_b300_fp8_mx_perf` | bia (b300) | 8 | ✅ success |
| `qwen3_30b_a3b_64gpu_gb300_bf16_perf` | lyris (gb300) | 8 | ✅ success |
[...]

PR #4719, opened 2026-05-11
Add Python-side guardrail for DeepEP IB limits

### LINKED ISSUE #3999: [Bug] HybridEP dispatcher passes incorrect max_num_of_tokens_per_rank to DeepEP, causing RDMA QP assertion failure

janEbert:
Compared to the previous attempt in #4094, which didn't properly check (was overly conservative; [this was only noticed in Megatron-Bridge tests](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3768)) and had to be reverted (#4718), we add a lot more code to basically re-compute the DeepEP internal handling. This is necessary to handle all cases that DeepEP considers properly.

Another approach we could try to take is filtering the raised error from DeepEP, and giving the more helpful error message when we encounter the known one.

Ref #4094, #4718, fix #3999.

### DIFF
--- a/megatron/core/transformer/moe/fused_a2a.py
+++ b/megatron/core/transformer/moe/fused_a2a.py
@@ -3,6 +3,7 @@
 # Copyright (c) 2025 DeepSeek
 # Licensed under the MIT License - https://github.com/deepseek-ai/DeepEP/blob/main/LICENSE
 
+import os
 from typing import Optional
 
 from megatron.core.utils import internal_api
@@ -268,6 +269,7 @@ else:
 
 
 try:
+    import hybrid_ep_cpp
     from deep_ep import HybridEPBuffer
 
     HAVE_HYBRIDEP = True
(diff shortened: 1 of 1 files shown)


ko3n1g:
PR passes internal MBridge tests

PR #4724, opened 2026-05-11
Use Protocols to type-check SelfAttention module of TransformerLayer

nschank:
# What does this PR do ?
Defines Protocols for the `SelfAttention` module in `TransformerLayer`, uses it in `TransformerLayerSubmodules`, and update upstream/downstream code appropriately to fix type errors discovered. Also adds some simple introspection/modification tooling for `attn_mask_type`.

I made `cp_comm_type` and `pp_layer_offset` non-optional parameters which are `None` when unused, which seems simpler and provides better type checking.

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


dimapihtar:
Hi @nschank 

Thanks a lot for the contribution! Could you resolve merge conflict, please?

nschank:
Fixed merge conflicts!

yashaswikarnati:
/ok to test [b65557f](https://github.com/NVIDIA/Megatron-LM/pull/4724/commits/b65557f45ac3b58d8e6d9dee998862d5d680bb0e)

yashaswikarnati (approved):

Phlip79 (approved):

janEbert:
What's the recommended way to check the type of the specs at runtime? Since they're now becoming anonymous functions.

nschank:
@janEbert 

That wasn't entirely possible even before - consider `WrappedTorchNorm` which is a "type" but the only way to figure out the return type is to construct one!

However, much like `get_submodules` and `get_param` I can add a `get_result_type` if desired - I haven't really run into code needing it yet, but if it would be useful it's not particularly difficult to write! But in practice I would say that it's not a great idea to need any of these; it's usually going to be better to extend the Protocols themselves in some way (such as a classmethod/ClassVar that identifies properties you might want to know), or in some cases to overwrite the callable instead with something like:

```python
def new_callable(*args, **kwargs): 
  result = original_callable(*args, **kwargs)
  match result:
    case ExampleModule() | OtherExampleModule():
      verify_something()
    case BadModule():
      raise ValueError(...)
    case _:
      logger.error('Unexpected type, you may need to support this!')
  return result
```

If you have any example usages you can point me to, I'd be happy to adapt these approaches/provide the tools to make them simpler to do!

nschank:
Fixed merge conflicts and fixed failing unit tests!

yashaswikarnati:
/ok to test [ed60482](https://github.com/NVIDIA/Megatron-LM/pull/4724/commits/ed604827f19c6334dca49d7e7b0e598c3e3ecfc5)

yaoyu-33 (approved):

nschank:
@janEbert Ah look at that, I did end up needing to introspect the constructed type for the last remaining failing test! Added `try_get_constructed_type`, which I tried to make a little more obviously scary than the others with the name since it's the least trustworthy of them.

janEbert:
Thank you for the example! Being able to check what a spec actually resolves to is becoming more relevant as we're trying to move away from config logic inside the spec files and would instead want to pre-define possible specs or create them at runtime, if necessary (e.g., it would be combinatorically infeasible to pre-define all combinations of sequence mixer + channel mixer).
The spec defines the backends used for different layers, while the config defines the architecture built from those backend specs. The two (hardcoded backend in spec, dynamic logic making use of those specs depending on the config at runtime) should clearly be separated. This is not the case for the `GPTModel`, but we're moving the `HybridModel` towards this.

nschank:
> Thank you for the example! Being able to check what a spec actually resolves to is becoming more relevant as we're trying to move away from config logic inside the spec files and would instead want to pre-define possible specs or create them at runtime, if necessary (e.g., it would be combinatorically infeasible to pre-define all combinations of sequence mixer + channel mixer).

@janEbert  For something like this, I'd personally try to structure things so the spec is less associated with a particular type, and is instead a fully-featured factory! You can pass any callable that will return something of the right 'shape' to these submodules, so there's no particular reason you need to build a "spec" that describes the types ahead of time. You can simply create a factory method called e.g. `build_sequence_mixer` which you pass directly to `HybridModelSubmodules.sequence_mixer`, and it can construct the correct type of sequence mixer dynamically.

This _does_ make it more difficult to do validation of the 'spec' _before_ you build the model, but I'd argue that's probably a good thing - it would make more sense to have a centralized validation step _after_ the model is built (e.g. using a visitor pattern and with modules supporting some mutability of their state at this stage for things like `attn_mask`), where there's no longer any guessing about how the spec is going to be interpreted. Introspecting a config object is inherently lossy, because the actual code can dynamically opt to do something with the config that wasn't expected by whoever originally wrote the configuration; for instance, if I add a subclass of attention which ignores the `attn_mask` parameter and has a hardcoded kind of attention mask, the validation will no longer be checking the correct thing. Something like a visitor pattern allows for even custom modules defined outside the library to register themselves to be validated correctly after construction, since setters/getters can _also_ be encoded in the type `Protocols` being added here to provide an actual API for how the validation is implemented.

janEbert:
Wow @nschank thank you for your very appreciated thoughts! Your idea of passing a factory is flexible and clean, but I don't think it's quite what we want, since it again couples the spec (what's supposed to become only a backend implementation specification) to the runtime configuration.

I do agree with the validation step making sense; ideally, we'd have validation both before and after model construction.

> Introspecting a config object is inherently lossy, because the actual code can dynamically opt to do something with the config that wasn't expected by whoever originally wrote the configuration

I agree, and there are already minor pain points because of this. Using a Visitor may make the most sense, I'll keep that in mind. Thanks again for your suggestions!

PR #4750, opened 2026-05-12
[dev] [4/5] Qwen3.5 support: Interleaved MRoPE layout

wplf:
## Qwen3.5 support series

This is part of a 5-PR series adding Qwen3.5-VL support, split for review clarity.

**Dev PRs (this series):**
- [1/5] MTP packed-seq CP+THD fix — #4494
- [2/5] FSDP DTensor Bridge checkpoint compatibility — #4748
- [3/5] SharedExpertMLP meta init — #4749
- [4/5] Interleaved MRoPE layout — #4750 ← **this PR**
- [5/5] Qwen3.5-VL training example — #4751

**Main PRs (corresponding mirrors):**
- [1/5] #4495
- [2/5] #4753
- [3/5] #4754
- [4/5] #4755
- [5/5] #4756

---
## Summary

Add the interleaved MRoPE layout used by Qwen3.5-VL, gated by a new config flag.

- New `TransformerConfig.mrope_interleaved: bool = False`.
- New `MultimodalRotaryEmbedding(interleaved_mrope=...)` argument.
- New helper `_apply_interleaved_mrope(freqs, mrope_section)` that converts the per-channel outer-product layout `(3, bs, seq_len, dim)` into the interleaved single-tensor layout `(bs, seq_len, dim)`.
- `GPTModel` passes `config.mrope_interleaved` through to the embedding.

## Why

`MultimodalRotaryEmbedding` currently produces the section-based T/H/W cycling used by Qwen2-VL. Qwen3.5-VL (and the matching HuggingFace `Qwen3VLTextRotaryEmbedding.apply_interleaved_mrope` unified on 2026-02-24) use a different layout where:
- T freqs occupy stride-3 positions `{0, 3, 6, ...}`
- H freqs occupy `{1, 4, 7, ...}`
- W freqs occupy `{2, 5, 8, ...}`

Both layouts are now supported; the flag picks between them.

## Risk

`mrope_interleaved` defaults to `False`, so existing Qwen2-VL / GPT users see no behavior change.

## Test plan

- [ ] Existing MRoPE tests still pass with `mrope_interleaved=False` (default).
- [ ] New layout matches HF `apply_interleaved_mrope` reference on a fixed `(position_ids, mrope_section)` input.

## Notes

This is the core prerequisite for an upcoming `examples/multimodal_dev/` PR that adds Qwen3.5-VL.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4749, opened 2026-05-12
[dev] [3/5] Qwen3.5 support: SharedExpertMLP meta init

wplf:
## Qwen3.5 support series

This is part of a 5-PR series adding Qwen3.5-VL support, split for review clarity.

**Dev PRs (this series):**
- [1/5] MTP packed-seq CP+THD fix — #4494
- [2/5] FSDP DTensor Bridge checkpoint compatibility — #4748
- [3/5] SharedExpertMLP meta init — #4749 ← **this PR**
- [4/5] Interleaved MRoPE layout — #4750
- [5/5] Qwen3.5-VL training example — #4751

**Main PRs (corresponding mirrors):**
- [1/5] #4495
- [2/5] #4753
- [3/5] #4754
- [4/5] #4755
- [5/5] #4756

---
## Summary

Add `_reset_parameters` to `SharedExpertMLP` so the directly-owned `gate_weight` is materialized off the meta device when `use_shared_expert_gate=True`.

## Why

Without this, meta-init leaves `gate_weight` on the meta device and the first forward fails with a meta-tensor error. Submodules already have their own `_reset_parameters`, so only the directly-owned `gate_weight` needs handling.

The implementation mirrors the standard per-parameter init pattern:
- run `init_method` when `config.perform_initialization`
- cast to `config.params_dtype`
- set `sequence_parallel` attribute

## Risk

Only fires when `use_shared_expert_gate=True` and `gate_weight is not None` — no effect on existing paths.

## Test plan

- [ ] Meta-init a config with `moe_shared_expert_overlap` and `use_shared_expert_gate=True`; verify forward succeeds.
- [ ] Existing shared-expert tests still pass.

## Notes

Extracted from a larger Qwen3.5-VL development branch to keep this fix reviewable on its own.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4751, opened 2026-05-12
[dev] [5/5] Qwen3.5 support: Qwen3.5-VL training example

wplf:
## Qwen3.5 support series

This is part of a 5-PR series adding Qwen3.5-VL support, split for review clarity.

**Dev PRs (this series):**
- [1/5] MTP packed-seq CP+THD fix — #4494
- [2/5] FSDP DTensor Bridge checkpoint compatibility — #4748
- [3/5] SharedExpertMLP meta init — #4749
- [4/5] Interleaved MRoPE layout — #4750
- [5/5] Qwen3.5-VL training example — #4751 ← **this PR**

**Main PRs (corresponding mirrors):**
- [1/5] #4495
- [2/5] #4753
- [3/5] #4754
- [4/5] #4755
- [5/5] #4756

---
## Summary

Adds a standalone VLM training playground under `examples/multimodal_dev/` with Qwen3.5-VL end-to-end.

**Model-agnostic harness**
- `pretrain_multimodal.py` entry point and `MODEL_REGISTRY` so a new architecture is just a registry entry + backing module.
- `models/base.py`, `forward_step.py`, `arguments.py`, `data/` (mock + CORD-V2 dataset with THD pack/pad in collate).

**Qwen3.5-VL**
- Full model: `models/qwen35_vl/` — vision encoder, MRoPE (pre-computed for THD), decoder, factory, specs, configurations for proxy / 9B / 397B-A17B variants.
- Run script + README, plus tests: `tests/test_mrope_parity.py`, `test_cp_correctness.py`, `test_cp_support.py`, `test_cp_thd_correctness.py`, `test_thd_correctness.py`, `test_thd_e2e.py`.

**One-line training infra change**
- `megatron/training/datasets/data_samplers.py`: enable the vanilla-collate torch DataLoader path when the new arg `use_vanilla_collate_fn` is set (needed for CORD-V2 under BSHD).

## Dependency

This example sets `mrope_interleaved=True` in its `TransformerConfig` and relies on the core MRoPE interleaved layout introduced in **#4750**. The diff here is self-contained (only `examples/` + the 1-line `data_samplers.py` change), but the example won't run end-to-end until #4750 merges.

## Functionality support
Resume training loss curve
bc2482cd-7d3a-4bda-8106-dee44c03b1f0

## CP / THD correctness verification

`tests/test_cp_thd_correctness.py` runs CP=1 and CP=4 in a single `torchrun --nproc-per-node 4` invocation (in-process `destroy_model_parallel` + `initialize_model_parallel` between phases, weights pinned by a `state_dict` snapshot, identical inputs via a seeded `torch.Generator`). Loss aggregated via `AllReduce(SUM)` on `(num, den)`; grad_norm aggregated via `AllReduce(SUM)` of gradients on the CP group then divided by `cp_size`, so each rank holds the CP-mean gradient that matches CP=1's backward on the full-batch mean loss.

Default config (B=2, S=64, H=256, L=2, bf16):

| Test | CP=1 | CP=4 | abs diff | rel diff |
|---|---|---|---|---|
| BSHD loss | 7.03250265 | 7.03217983 | 3.23e-04 | 4.59e-05 |
| BSHD grad_norm | 4.84910854 | 4.84744710 | 1.66e-03 | 3.43e-04 |
| THD loss | 7.03250265 | 7.03241825 | 8.44e-05 | 1.20e-05 |
| THD grad_norm | 4.84910839 | 4.84912564 | 1.73e-05 | 3.56e-06 |

Cross-check: BSHD CP=1 loss ≡ THD CP=1 loss = `7.03250265`, and BSHD CP=1 grad_norm ≈ THD CP=1 grad_norm to 7 decimals — equal-length sequences make the two attention paths mathematically identical, so the CP=1 grad_norm match confirms BSHD/THD parity at the gradient level as well.

## Checkpoint conversion (HF → Megatron-FSDP DTensor)

The example consumes a Megatron-FSDP DTensor checkpoint, converted from the HuggingFace release via [Megatron-Bridge](https://github.com/NVIDIA-NeMo/Megatron-Bridge).

**Setup** — clone Bridge and pin its `3rdparty/Megatron-LM` submodule to this branch:

```bash
git clone --recurse-submodules https://github.com/NVIDIA-NeMo/Megatron-Bridge.git
cd Megatron-Bridge/3rdparty/Megatron-LM
git remote add wplf https://github.com/wplf/Megatron-LM.git
git fetch wplf feat/qwen35-vl-example
git checkout feat/qwen35-vl-example
cd ../..
```

**Convert** (single 8×H100 node, EP=8 / TP=CP=1; `--hf-model` can be any Qwen3.5 variant, e.g. `Qwen/Qwen3.5-35B-A3B`):

```bash
PYTHONPATH=./src:./3rdparty/Megatron-LM/ \
  torchrun --nproc_per_node=8 \
  examples/conversion/mfsdp/convert_checkpoints_fsdp.py import \
  --hf-model Qwen/Qwen3.5-35B-A3B \
  --megatron-path ${WORKSPACE}/models/Qwen/Qwen3.5-35B-A3B-fsdp \
  --ckpt-format fsdp_dtensor \
  --ep 8
```

HF weights are auto-fetched on first run via `huggingface_hub`. Adjust `--tp` / `--cp` / `--ep` to match the training topology (must satisfy `WORLD_SIZE % (TP*CP*EP) == 0`).

**Output**

```
${WORKSPACE}/models/Qwen/Qwen3.5-35B-A3B-fsdp/
├── iter_0000000/
│   ├── __0_0.distcp .. __7_0.distcp   # FSDP DTensor shards, one per rank (~18 GB each for 35B-A3B)
│   ├── .metadata
│   ├── run_config.yaml
│   └── train_state.pt
├── latest_checkpointed_iteration.txt
└── latest_train_state.pt
```

**Bridge dependency** — requires [NVIDIA-NeMo/Megatron-Bridge#3987](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3987) (skip tokenizer save in `convert_checkpoints_fsdp.py`). Without that fix the checkpoint is still written correctly but the script exits non-zero after save with `AttributeError: 'TokenizerConfig' object has no attribute 'make_vocab_size_divisible_by'` against this branch's `megatron.core.tokenizers.utils.build_tokenizer`.

## Risk

- All new files under `examples/multimodal_dev/`.
- `data_samplers.py` change is fully backwards-compatible: behavior is unchanged unless `use_vanilla_collate_fn` is explicitly set.

## Test plan

- [ ] `pytest examples/multimodal_dev/tests/` passes.
- [ ] `scripts/run_qwen35_vl.sh` proxy variant trains a few steps on mock data.
- [ ] CORD-V2 dataset loads with `--use-vanilla-collate-fn` and trains a few steps.
- [ ] `torchrun --nproc-per-node 4 examples/multimodal_dev/tests/test_cp_thd_correctness.py` — CP=1 vs CP=4 BSHD/THD loss + grad_norm within tolerance.

### DIFF
--- a/examples/multimodal_dev/data/cord_v2.py
+++ b/examples/multimodal_dev/data/cord_v2.py
@@ -1,11 +1,11 @@
 # Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
 
-"""Simple VLM dataset for multimodal_dev training.
+"""CORD-V2 VLM dataset for multimodal_dev training.
 
 Single-turn image-text dataset using a HuggingFace ``AutoProcessor`` for
-tokenization and image preprocessing.  Currently supports CORD-V2 (receipt
-OCR).  No multi-turn support — each sample is one image + question →
-answer pair.
+tokenization and image preprocessing.  This module is the reference
+implementation for the CORD-V2 receipt-OCR dataset.  No multi-turn support —
+each sample is one image + question → answer pair.
 
 Each image is preprocessed via ``qwen_vl_utils.process_vision_info`` and
 fed to the processor with Qwen-VL's recommended ``min_pixels`` /
@@ -20,6 +20,32 @@ Usage::
         --model-arch qwen35_vl --dataset-provider cord_v2 \\
         --hf-processor-path Qwen/Qwen3.5-397B-A17B \\
         --total-seq-length 4096 --use-vanilla-collate-fn
+
+Adding another VLM dataset
+--------------------------
+
+The dataset layer mirrors the model layer's registry pattern: each dataset
+ships its own module and a ``train_valid_test_datasets_provider`` factory,
+and the model's registry entry maps a ``--dataset-provider`` name to that
+factory's dotted path.  To add a new dataset (e.g. NLVR2):
+
+1. Create ``examples/multimodal_dev/data/<name>.py`` with::
+
+       def train_valid_test_datasets_provider(train_val_test_num_samples):
+           ...  # build datasets using args from get_args()
+           return train_ds, val_ds, test_ds
+
+2. Register it under the relevant model in
+   ``examples/multimodal_dev/models/__init__.py``::
+
+       MODEL_REGISTRY["qwen35_vl"]["dataset_providers"]["<name>"] = (
+           "examples.multimodal_dev.data.<name>"
+           ".train_valid_test_datasets_provider"
+       )
+
+3. Launch with ``--dataset-provider <name>``.
+
+No edits to ``pretrain_multimodal.py`` or ``forward_step.py`` are required.
 """
 
 import json
--- a/examples/multimodal_dev/models/qwen35_vl/mrope.py
+++ b/examples/multimodal_dev/models/qwen35_vl/mrope.py
@@ -1,0 +1,378 @@
+# Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+
+"""MRoPE (Multimodal Rotary Position Embedding) position ID computation.
+
+Computes 3D position IDs for Qwen3.5-VL: for text tokens all three
+dimensions share sequential positions; for image/video tokens the three
+dimensions encode (temporal, height, width) in the merged spatial grid.
+
+Supports two input layouts:
+
+* **BSHD** — ``input_ids`` is ``[B, S]``; each row is an independent
+  sample (possibly padded) and ``attention_mask`` marks valid tokens.
+* **THD** — ``input_ids`` is ``[1, T]``, a concatenation of ``N``
+  sub-sequences. ``packed_seq_params.cu_seqlens_q_padded`` gives the
+  physical segment boundaries in the packed tensor and
+  ``cu_seqlens_q`` gives the valid (unpadded) token count inside each
+  segment. Position IDs restart at 0 at every segment boundary; image
+  / video grid rows are consumed in packed order across segments.
+
+Ported from Megatron-Bridge ``get_rope_index`` (which itself is adapted
+from HF ``Qwen3VLForConditionalGeneration.get_rope_index``).  The inner
+loop iterates over vision occurrences, not individual tokens.
+"""
+
+from typing import Optional
+
+import torch
+from torch import Tensor
+
+from megatron.core.packed_seq_params import PackedSeqParams
+
+
+def _build_sample_mrope_positions(
+    sample_input_ids: Tensor,
+    image_grid_thw: Optional[Tensor],
+    video_grid_thw: Optional[Tensor],
+    image_index: int,
+    video_index: int,
+    spatial_merge_size: int,
+    image_token_id: int,
+    video_token_id: int,
+    vision_start_token_id: int,
+) -> tuple[Tensor, int, int]:
+    """Compute MRoPE position IDs for a single sub-sequence.
+
+    Walks vision occurrences in ``sample_input_ids`` and produces a
+    ``[3, L]`` position tensor whose values start at 0. Advances
+    ``image_index`` / ``video_index`` through ``image_grid_thw`` /
+    ``video_grid_thw`` so callers can keep a running cursor across
+    multiple sub-sequences.
+    """
+    vision_start_indices = torch.argwhere(
+        sample_input_ids == vision_start_token_id,
+    ).squeeze(1)
+    vision_tokens = sample_input_ids[vision_start_indices + 1]
+    image_nums = int((vision_tokens == image_token_id).sum())
+    video_nums = int((vision_tokens == video_token_id).sum())
+    input_tokens = sample_input_ids.tolist()
+    llm_pos_ids_list: list = []
+    st = 0
+    remain_images, remain_videos = image_nums, video_nums
+
+    for _ in range(image_nums + video_nums):
+        if image_token_id in input_tokens and remain_images > 0:
+            ed_image = input_tokens.index(image_token_id, st)
+        else:
+            ed_image = len(input_tokens) + 1
+        if video_token_id in input_tokens and remain_videos > 0:
+            ed_video = input_tokens.index(video_token_id, st)
+        else:
+            ed_video = len(input_tokens) + 1
+
+        if ed_image < ed_video:
+            t, h, w = (
+                image_grid_thw[image_index][0],
+                image_grid_thw[image_index][1],
+                image_grid_thw[image_index][2],
+            )
+            image_index += 1
+            remain_images -= 1
+            ed = ed_image
+        else:
+            t, h, w = (
+                video_grid_thw[video_index][0],
+                video_grid_thw[video_index][1],
+                video_grid_thw[video_index][2],
+            )
+            video_index += 1
+            remain_videos -= 1
+            ed = ed_video
+
+        llm_grid_t, llm_grid_h, llm_grid_w = (
+            t.item(),
+            h.item() // spatial_merge_size,
+            w.item() // spatial_merge_size,
+        )
+        text_len = ed - st
+
+        st_idx = (
+            llm_pos_ids_list[-1].max() + 1
+            if llm_pos_ids_list
+            else 0
+        )
+        llm_pos_ids_list.append(
+            torch.arange(text_len).view(1, -1).expand(3, -1)
+            + st_idx
+        )
+
+        t_index = (
+            torch.arange(llm_grid_t)
+            .view(-1, 1)
+            .expand(-1, llm_grid_h * llm_grid_w)
+            .flatten()
+        )
+        h_index = (
+            torch.arange(llm_grid_h)
+            .view(1, -1, 1)
+            .expand(llm_grid_t, -1, llm_grid_w)
+            .flatten()
+        )
+        w_index = (
+            torch.arange(llm_grid_w)
+            .view(1, 1, -1)
+            .expand(llm_grid_t, llm_grid_h, -1)
+            .flatten()
+        )
+        llm_pos_ids_list.append(
+            torch.stack([t_index, h_index, w_index])
+            + text_len
+            + st_idx
+        )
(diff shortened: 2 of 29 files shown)


Victarry:
While running the PR's `examples/multimodal_dev/tests` tests, I hit a collection failure on the original branch: `test_thd_e2e.py` and `test_thd_correctness.py` import `_pack_batch` from `examples.multimodal_dev.forward_step`, but `forward_step.py` does not define/export `_pack_batch`.

Could you add that helper or update the tests to use the intended packing API?

BestJuly:
Please add the latest checkpoint conversion guide so users to resume from HF checkpoint and run. Previously we record the steps in this [issue](https://github.com/BestJuly/Megatron-LM/issues/7) and there should have some updates now.

wplf:
OK, I'm working on the UT of CP now. Checkpoint conversion guide may be finished this afternoon.

wplf:
/ok to test [6399e29](https://github.com/NVIDIA/Megatron-LM/pull/4751/commits/6399e292d9eab85042e253658cce7b3ee16b14f1)

examples/multimodal_dev/data/cord_v2.py
```diff
@@ -0,0 +1,371 @@
...
+            "pixel_values": pixel_values,
+            "image_grid_thw": image_grid_thw,
+        }
+
+
+# ---------------------------------------------------------------------------
+# Megatron dataset provider interface
+# ---------------------------------------------------------------------------
+
+def train_valid_test_datasets_provider(train_val_test_num_samples):
```
Victarry, line 334: **[SUGGESTION Design]** The real dataset path is currently hard-wired around CORD-V2, while the model side already has a registry/factory pattern. Could we make the dataset layer similarly extensible?

Concretely, please consider passing a dataset name into `train_valid_test_datasets_provider` (or an equivalent dataset factory), dispatching by dataset name instead of binding the implementation directly to CORD-V2, and documenting how to add a new dataset provider. That would make this example easier to extend beyond CORD-V2 without editing model-specific registry code each time.

wplf:
/ok to test [89ccadf](https://github.com/NVIDIA/Megatron-LM/pull/4751/commits/89ccadf2f145d2f3f6c8e6887f83d16909e5ac20)

wplf:
/ok to test [334d4e1](https://github.com/NVIDIA/Megatron-LM/pull/4751/commits/334d4e18e81f5cb8a107a0e90c5343d77af5a62d)

cryoco:
May I learn about the corresponding TE version in qwen3.5 support?

wplf:
> May I learn about the corresponding TE version in qwen3.5 support?
FYI, TE dependency is fairly loose.

I'm using the latest TE with a cherry-pick from [TE pr 2932](https://github.com/NVIDIA/TransformerEngine/pull/2932)  on gb.
If you're using Hopper, please ensure your cuDNN version is greater than 9.19.0.
Using Hopper with fused attention + thd + cuDNN below 9.19.0 will cause NaN issues during THD training.

cryoco:
> > May I learn about the corresponding TE version in qwen3.5 support?
> > FYI, TE dependency is fairly loose.
> 
> I'm using the latest TE with a cherry-pick from [TE pr 2932](https://github.com/NVIDIA/TransformerEngine/pull/2932) on gb. If you're using Hopper, please ensure your cuDNN version is greater than 9.19.0. Using Hopper with fused attention + thd + cuDNN below 9.19.0 will cause NaN issues during THD training.

Thanks. BTW, is there any current performance benchmark on hopper or blackwell?

wplf:
> > > May I learn about the corresponding TE version in qwen3.5 support?
> > > FYI, TE dependency is fairly loose.
> > 
> > 
> > I'm using the latest TE with a cherry-pick from [TE pr 2932](https://github.com/NVIDIA/TransformerEngine/pull/2932) on gb. If you're using Hopper, please ensure your cuDNN version is greater than 9.19.0. Using Hopper with fused attention + thd + cuDNN below 9.19.0 will cause NaN issues during THD training.
> 
> Thanks. BTW, is there any current performance benchmark on hopper or blackwell?

We are still working on it.

examples/multimodal_dev/scripts/run_qwen35_vl.sh
```diff
@@ -0,0 +1,503 @@
...
+    DISTRIBUTED_ARGS+=(
+        --master_addr "${MASTER_ADDR:-localhost}"
+        --master_port "${MASTER_PORT:-6000}"
+    )
+fi
+
+# --- Parallelism ---
+MODEL_PARALLEL_ARGS=(
+    --tensor-model-parallel-size "$TP"
+    --pipeline-model-parallel-size "$PP"
```
Victarry, line 232: **[IMPORTANT Correctness]** This script exposes `PP` and forwards it into `--pipeline-model-parallel-size`, but the multimodal model provider currently builds a full model on every rank and ignores the `pre_process` / `post_process` stage flags.

With `PP>1`, this can violate Megatron's pipeline-parallel contract by constructing full pre/post-processing on each pipeline rank instead of partitioning the model. Please either reject `pipeline_model_parallel_size > 1` for this standalone path, or wire the model construction through the standard pre/post-process pipeline-stage semantics.

examples/multimodal_dev/scripts/run_qwen35_vl.sh
```diff
@@ -0,0 +1,503 @@
...
+    --expert-model-parallel-size "$EP"
```
Victarry, line 233: **[IMPORTANT Correctness]** The script defaults `EP=2` and always forwards `--expert-model-parallel-size`, but the dense variants set `NUM_EXPERTS=0` and do not add MoE args.

As a result, a simple dense run such as `MODEL_VARIANT=9b ./run_qwen35_vl.sh` will fail Megatron argument validation because expert parallelism is enabled without experts. Please default dense variants to `EP=1`, or fail fast with a clear script-level error when `NUM_EXPERTS=0 && EP>1`.

examples/multimodal_dev/scripts/run_qwen35_vl.sh
```diff
@@ -0,0 +1,503 @@
...
+    --hidden-dropout 0.0
+    --experimental-attention-variant gated_delta_net
+    --linear-attention-freq 4
+    --linear-conv-kernel-dim 4
+    --linear-key-head-dim 128
+    --linear-value-head-dim 128
+    --linear-num-key-heads 16
+    --linear-num-value-heads "$LINEAR_NUM_VALUE_HEADS"
+    --make-vocab-size-divisible-by 485
+    --moe-router-force-load-balancing
```
Victarry: **[IMPORTANT Training Quality]** `--moe-router-force-load-balancing` is useful for performance benchmarking, but it should not be enabled unconditionally in the default finetuning recipe.

For real CORD-V2 finetuning/convergence checks, forcing router balance prevents the router from learning data-dependent routing decisions and makes the loss curve easy to misinterpret. Please gate this behind an explicit perf/mock-data mode, or default it off for real finetuning runs.

examples/multimodal_dev/models/qwen35_vl/vision_encoder.py
```diff
@@ -0,0 +1,593 @@
...
+                order from the ViT transformer blocks.
+
+        Returns:
+            ``[total_merged_patches, out_hidden_size]``.
+        """
+        hidden_states = self.patch_norm(hidden_states)
+        merged = hidden_states.view(-1, self.merge_dim)
+        merged, _ = self.linear_fc1(merged)
+        # NOTE: Official HuggingFace uses default approximate='none' in Qwen3VLVisionPatchMerger.
+        merged = torch.nn.functional.gelu(merged, approximate="tanh")
```
Victarry: **[IMPORTANT Checkpoint Parity]** The comment says the official HuggingFace PatchMerger uses the default GELU approximation (`approximate='none'`), but the implementation calls `gelu(..., approximate='tanh')`.

This changes the vision tower numerics after HF-to-Megatron checkpoint conversion. Please match the HF implementation here, or add a parity test/justification showing that `approximate='tanh'` is intentional and checkpoint-compatible.

examples/multimodal_dev/models/qwen35_vl/mrope.py
```diff
@@ -0,0 +1,378 @@
...
+    ``video_grid_thw`` so callers can keep a running cursor across
+    multiple sub-sequences.
+    """
+    vision_start_indices = torch.argwhere(
+        sample_input_ids == vision_start_token_id,
+    ).squeeze(1)
+    vision_tokens = sample_input_ids[vision_start_indices + 1]
+    image_nums = int((vision_tokens == image_token_id).sum())
+    video_nums = int((vision_tokens == video_token_id).sum())
+    input_tokens = sample_input_ids.tolist()
```
Victarry, line 59: **[SUGGESTION Performance]** This forward path repeatedly moves runtime tensors to Python (`tolist()` / `.item()` here and in the vision grid helpers). At scale this introduces GPU-to-CPU synchronization in every step and also makes CUDA graph capture difficult.

Could we precompute the MRoPE/grid metadata in the data pipeline, or cache it by sequence/grid shape so the hot model forward path does not need host-side tensor inspection each iteration?

Victarry (commented):
Thanks for this PR!
Reproduced the E2E pipeline from weight conversion to model training and worked well. Left a few comments.

examples/multimodal_dev/models/qwen35_vl/mrope.py
wplf: We will resolve this by @BestJuly 's mrope fusion PR.

examples/multimodal_dev/scripts/run_qwen35_vl.sh
wplf: Resolved

Victarry (approved):

PR #4424, opened 2026-04-22
[DEV] fix(megatron-fsdp): compute SWiGLU/GDN split in item coordinates for non-DTensor optimizer states

xuwchen:
main PR: #4423 

# What does this PR do ?
`split_swiglu_linear_fc1` and `split_gdn_fused` originally used `data.numel() // tp_mesh.mesh.numel()` to compute component slices (`w_slice`/`v_slice` for SWiGLU, `comp_slice` for GDN), then intersected them with `fsdp_slice` to determine each rank's portion. When `data` is a DTensor, `data.numel()` returns the global element count (=`param.numel()`), so component slices and `fsdp_slice` are both in item coordinates and the intersection is valid. However, when `data` is a plain Tensor (optimizer states created by FusedAdam), `data.numel()` returns the local shard size on this rank, making the component slices's range far smaller than the item coordinate space and resulting in an intersection across incompatible coordinate systems.

The fix replaces `data.numel()` with `dist_param.numel()` (`dist_param` is always a DTensor and therefore reflects the global element count), ensuring `fsdp_slice` and all component slices are in item coordinates.

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

[...]

PR #4393, opened 2026-04-20
Revert "fix mfsdp unwrap stuck at MegatronFSDP [dev] (#4273)"

wplf:
This reverts Pull request #4273 , which breaks ckpt load&save and others.

PR #4753, opened 2026-05-12
[main] [2/5] Qwen3.5 support: FSDP DTensor Bridge checkpoint compatibility

wplf:
## Qwen3.5 support series

This is part of a 5-PR series adding Qwen3.5-VL support, split for review clarity.

**Main PRs (this series):**
- [1/5] MTP packed-seq CP+THD fix — #4495
- [2/5] FSDP DTensor Bridge checkpoint compatibility — #4753 ← **this PR**
- [3/5] SharedExpertMLP meta init — #4754
- [4/5] Interleaved MRoPE layout — #4755
- [5/5] Qwen3.5-VL training example — #4756

**Dev PRs (corresponding mirrors):**
- [1/5] #4494
- [2/5] #4748
- [3/5] #4749
- [4/5] #4750
- [5/5] #4751

---
## Summary

Four related fixes that together let `fsdp_dtensor` resume from a checkpoint converted with **Megatron-Bridge**:

1. **`split_swiglu_linear_fc1` / `split_gdn_fused`**: accept plain `Tensor` (not only `DTensor`) for `data.to_local()`. Bridge writes some already-local tensors that previously crashed the splitter.

2. **TP-local GDN split**: when the GDN tensor is already TP-local (`data.shape[split_dim] == sum(split_sizes)`), use `split_dtensor(..., update_uneven_dtensor_chunk_meta=True)` instead of the TP-mesh-based split that assumes the full unsharded shape.
3. **Stash missing top-level keys in `_load_base_checkpoint`**: keys not present as DCP storage metadata (e.g. `args`, `iteration`) are stashed before `dcp.load` and restored after. Megatron-saved checkpoints contain these as `BytesStorageMetadata`; Bridge-converted ones don't, and DCP errors without this stash.
4. **`DistributedOptimizer._copy_model_params_to_main_params`**: under `use_megatron_fsdp`, return instead of raising. DCP already loads directly into the fp32 main buffer, and a post-load hook copies main → model weights — both are correct at this call site.

5. **`split_swiglu_linear_fc1` / `split_gdn_fused` use `dist_param` coordinates** *(cherry-pick of #4424 by @xuwchen)*: when `data` is a plain `Tensor` (FusedAdam optimizer state), `data.numel()` / `data.shape` describe the local item layout, not the global TP-sharded layout, so the splitter computed wrong `data_size`, `view_shape`, and `per_tp_rank_shape`. Switch to `global_shape = dist_param.shape` and `dist_param.numel()` everywhere, with a DTensor-shape consistency assert.

Also wires `handle_gdn_in_state_dict` into `preprocess_fsdp_dtensor_state_dict`.

## Why bundled

All four fixes are needed together for the Bridge → `fsdp_dtensor` resume path to work end-to-end.

## Note on key-unwrapping

An earlier revision of this branch also rewrote `model.get_parameter(f"module.{key}")` → `model.get_parameter(key)`. That root cause was resolved in #4393 ("Revert 'fix mfsdp unwrap stuck at MegatronFSDP [dev] (#4273)'"), so the `module.` prefix is correct again and those four lookup edits have been dropped.

## Notes

Mirror of #4748 (same patch, targeting `main` instead of `dev`).

## Update (2026-05-14): model→main copy reverted

A previous revision of this PR (commit `4c2c3cbfb` on dev / `f4ba23b11` on main) implemented an FSDP `copy_model_weights_to_main_weights` and called it from `_copy_model_params_to_main_params`, on the theory that the silent `return` left fp32 main params uninitialized under `--no-load-optim` / `--finetune`.

Empirical buffer-diff probe showed that's wrong: under `fsdp_dtensor`, `main_weight_buffer` is exposed as the module's `nn.Parameter`s via `MegatronFSDP`'s `state_dict` pre-hook (`_replace_param_with_distributed_if_needed`), so the preceding `model.load_state_dict(state_dict["model"])` writes the checkpoint's fp32 tensors **directly into `main_weight_buffer`** — full fp32 precision, even with `--no-load-optim`, because main params live in the model state dict (not the optimizer state dict) under `fsdp_dtensor`. The load_state_dict post-hook (`copy_main_weights_to_model_weights`) then casts main → bf16 to refresh `model_weight_buffer`.

A 50-step finetune+resume on a Qwen3.5-VL 0.8B model with and without the copy gave indistinguishable loss curves (wandb runs `q8d7l7nm` / `o4tyuf5z` / `wwuz7y4z` under https://wandb.ai/wplf/fsdp-fix-verify/). The pre-copy buffer diff was exactly bf16 rounding noise; the post-copy diff dropped to 0 only because main had been downgraded to bf16 precision.

The model→main copy commit has been reverted. The plain `return` is correct, and the comment in `_copy_model_params_to_main_params` now documents the mechanism so this isn't re-litigated.
## Update (2026-05-18): TP-local GDN fast-path restored on main

The TP-local GDN fast-path is back on this main-target PR. NVIDIA/Megatron-LM#4799 (`conver334`) provides the same fast-path against `dev`, and the dev-target PR #4748 dropped its copy to avoid a textual conflict; but #4799 has **not** yet landed on `main`, so this PR still needs to provide it. Once #4799's main mirror lands, this bullet can be removed here too (same flow as PR #4748).


wplf:
/ok to test [bb5dbe7](https://github.com/NVIDIA/Megatron-LM/pull/4753/commits/bb5dbe759f4c4a0c63cd7969d809355b7fbce2f7)

wplf:
/ok to test [6fd7b35](https://github.com/NVIDIA/Megatron-LM/pull/4753/commits/6fd7b35b4a0af2e7e71b4252ce54624dc825f853)

shjwudp (approved):

PR #4771, opened 2026-05-13
Route non-Muon params through DistributedOptimizer

deepakn94:
## Summary

Splits the DDP buffer for layer-wise optimizers so Muon-managed matrices stay in a shard-aligned `LayerWiseDistributedOptimizer` buffer, while everything else (embeddings, biases, LayerNorm weights) lands in a separate byte-level `DistributedOptimizer` buffer. The top-level optimizer becomes `ChainedOptimizer([LayerWise(Muon), DistOpt(Adam)])`. This drops the massive `(dp_size - 1) × embedding_numel` padding overhead that the all-LayerWise layout pays when there are unique-large params (typical for embeddings).

## What changed

- `BufferKey` gains an `is_layer_wise_distributed_optimizer` field; `group_params_for_buffers` reads it from a param attribute set by `tag_params_for_buffer_routing` before DDP wrap.
- `LayerWiseDistributedOptimizer.compute_full_param_layout` dispatches per-buffer: LayerWise's shard-aligned layout for its buffers, `DistributedOptimizer`'s byte-level layout for the rest.
- `LayerWiseDistributedOptimizer._shard_params_from_layout` and `set_bucket_layerwise_params_list` filter to LayerWise-managed buckets only (`_bucket_is_layer_wise_managed` helper).
- `_get_megatron_emerging_optimizer` constructs a real `DistributedOptimizer` for non-emerging-optimizer (Adam) groups when the layer-wise + layout path is active, chained at the top level alongside LayerWise.
- `partition_buckets` uniqueness check relaxed to only enforce on uint8 buffers (so two bf16 buffers can coexist in one DDP wrapper).
- New `start_param_sync_for_bucket_group_subset` helper: when LayerWise and DistOpt are chained, each calls `start_param_sync` on only its own bucket groups (filtered via a predicate) so the same buckets aren't all-gathered twice per step.
- LayerWise param layout: replaced cross-layer size-matching with greedy-LPT bin-packing in contiguous backprop chunks, restoring the `bucket_id == cur + 1` invariant in `_ParamAndGradBuffer.__init__` and preserving `overlap_grad_reduce` dispatch timing.
- Bucket finalisation now uses `threshold = max(bucket_size, dp_size * chunk_max_param * 0.9)` so high-dp configs don't emit narrow buckets where the chunk's largest param dominates shard size — keeps per-bucket padding overhead under ~11%.
- `_build_sharded_state_dict_metadata` treats `use_layer_wise_distributed_optimizer` as equivalent to `use_distributed_optimizer` for sharding-type purposes, fixing the deprecated `fully_sharded_model_space` fallback that crashed checkpoint save post-#4509.

## Convergence

image

## Performance

8B Llama-style (TP=4 PP=1, mb=1, seq=8192), `--ddp-num-buckets 6`, sweep over DP ∈ {8, 16, 32} × GA ∈ {1, 2, 4, 8}. Per-cell numbers are medians across the 4 GA points at that DP, hybrid minus legacy:

8b_throughput_vs_ga

**Take-away:** hybrid is throughput-neutral to slightly-better than legacy at every DP.

## Guard assertions for not-yet-supported configurations

The split path fails fast (clear error message) on:

- **`num_distributed_optimizer_instances > 1`** — `distributed_optimizer_instance_id` is hardcoded to 0 in this path; needs proper computation to lift.
- **Expert-parallel non-Muon param groups** — would need a second DistOpt instance with the expert-DP process group + expert buffers.

Both fall back cleanly when `use_layer_wise_param_layout=False` (the legacy LayerWise ping-pong path).

## Test plan

- [x] Unit tests: 62 LayerWise + 6 metadata-builder tests pass on cw-dfw.
- [x] End-to-end save_checkpoint/load_checkpoint roundtrip on the LayerWise + DistOpt chain (`test_optimizer_common_state_dict_hybrid` in `tests/unit_tests/dist_checkpointing/test_layer_wise_optimizer.py`).
- [x] 8B Transformer sweep across {Adam, Muon legacy, Muon hybrid} × {4, 8, 16} nodes × GA ∈ {1, 2, 4, 8}.


deepakn94:
/ok to test [34cd925](https://github.com/NVIDIA/Megatron-LM/commit/34cd92505427a85b5ae7524c939f4b8e4b5b8068)

deepakn94:
/ok to test [d52da35](https://github.com/NVIDIA/Megatron-LM/commit/d52da3560773a9cd1a01652af80c7fab7cbd53b7)

megatron/core/optimizer/__init__.py
```diff
@@ -726,6 +726,7 @@ def _get_megatron_emerging_optimizer(
     model_chunks: List[MegatronModule],
     config_overrides: Optional[Dict[ParamKey, Any]] = None,
     pg_collection: Optional[ProcessGroupCollection] = None,
+    use_gloo_process_groups: bool = True,
```
kvareddy: Why do we need gloo process groups?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1498,11 +1497,16 @@ def partition_buckets(
...
+    # (e.g. LayerWise-managed bf16 weights + Adam-managed bf16 biases live in
+    # separate buffers but share the bf16 ``param_dtype``), so the uniqueness
+    # check is restricted to uint8.
     dtype_to_buffer_map = {}
     for buffer in buffers:
         dtype = buffer.param_dtype
-        # Make sure that the param_dtype of any two buffers is different.
-        assert dtype not in dtype_to_buffer_map
+        if dtype == torch.uint8:
+            assert dtype not in dtype_to_buffer_map
```
deepakn94: We can probably get rid of dtype_to_buffer_map here?

megatron/core/optimizer/__init__.py
deepakn94: We don't, can probably remove this.

FDecaYed (approved):
Other part looks ok but I have some concerns on the tagging layerwise managed param and DDP wrapping part. I think we'll need another PR to refactor it. current approach of hard coding muon split seems not correct in longer term

My understanding is optimizer setup needs to happen after DDP wrapping due to it changing param/grad buffer. But on the other hand, we want a "optimizer managed layout" that can make comm optimal. If done naturally, this means creating optimizer first. 
So overall I think we should a single source of code dealing with adam/muon/soap/etc split, as well as producing layout. all run in the beginning and same information is feed into both DDP and optimizers init.

kvareddy (approved):

jaredcasper (approved):

lmcafee-nvidia (approved):

yaoyu-33 (approved):

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -3004,8 +3032,10 @@ def step_with_ready_grads(self) -> bool:
             # the first all-gather is launched asynchronously in the next optimizer.zero_grad()
             # call and subsequent all-gathers are launched in the forward pre-hook.
             if not self.ddp_config.overlap_param_gather:
-                for model_chunk in self.model_chunks:
-                    model_chunk.start_param_sync()
+                # Only sync DistOpt-managed bucket groups so a sibling
+                # LayerWiseDistributedOptimizer's own ``start_param_sync`` call
+                # is not duplicated for the same buckets.
+                self.start_param_sync_for_bucket_group_subset()
```
yashaswikarnati, line 3038: I'm trying to get my head around whats the behaviour with overlaps like overlap_param_gather_with_optimizer_step. I wonder if in the overlap path as well we need a filtered call like this in https://github.com/deepakn94/Megatron-LM/blob/d642193cae4995feadf5a157a1061f3476bd4eba/megatron/core/optimizer/optimizer.py#L1292

yashaswikarnati (approved):

PR #4773, opened 2026-05-13
Add dev-feature preservation gate and change schedule

Phlip79:
# What does this PR do ?

Adds a deterministic dev-feature preservation audit to the nightly main-to-dev sync instructions, and changes the scheduled sync cadence from daily to twice weekly.

## Summary

- Adds a hard-abort "Dev-feature preservation audit" to the `skills/nightly-sync/SKILL.md` pre-push invariant checks.
- The audit scans files touched by the sync and reports lines that are present on `origin/dev`, absent from `origin/main`, and missing from the merged tree. This catches the common `-X theirs` regression where a dev-only feature is silently dropped during main-to-dev conflict resolution.
- Keeps the existing skip/override semantics explicit:
  - skips `pyproject.toml`, `uv.lock`, `docker/Dockerfile.ci.dev`, and `.github/CODEOWNERS`, which are already covered by dedicated rules;
  - skips the skill's intentional "Files to Override from Main" list, where main may legitimately win.
- Documents how to classify audit hits:
  - if main intentionally removed the line, document the specific main commit SHA in the PR body;
  - otherwise treat it as a merge accident and restore the dev line before pushing.
- Records recent regressions from #4659 and #4716 that this audit would have flagged, including dropped dynamic-context-parallel references, missing MoE/router call sites, and the missing packing scheduler documentation.
- Changes `.github/workflows/nightly-sync-main-to-dev.yml` from daily at `21:00 UTC` to Monday and Thursday at `15:00 UTC` (`8 AM PDT` / `7 AM PST`).

### DIFF
--- a/.github/workflows/nightly-sync-main-to-dev.yml
+++ b/.github/workflows/nightly-sync-main-to-dev.yml
@@ -17,9 +17,10 @@ name: Nightly Sync Main to Dev
 on:
   workflow_dispatch:
   schedule:
-    # 21:00 UTC = 2 PM PDT (1 PM PST during winter — GitHub Actions cron
-    # is UTC-only and does not follow DST).
-    - cron: '0 21 * * *'
+    # Twice-weekly cadence: Monday and Thursday at 15:00 UTC.
+    # 15:00 UTC = 8 AM PDT (7 AM PST during winter — GitHub Actions cron
+    # is UTC-only and does not follow DST). Days-of-week: 1=Mon, 4=Thu.
+    - cron: '0 15 * * 1,4'
 
 concurrency:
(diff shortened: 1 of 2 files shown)


FDecaYed (approved):
Let's try one with this. my intuition is this would catch lots of cases where feature is already merged into main and we're syncing after T1(so `theirs` being correct strat cases)
and eventually we still need to let the agent figure out which type it is after catching dropped dev code
- if we're post T1 and main has latest code, take main
- if we're before T1 and main don't have this at all, take dev

I'm thinking maybe we can let agent compare git commits, or read PR where we already have manually linked dev/main PR to decide

balasaajay (approved):

PR #4785, opened 2026-05-13
Add --qad-train-target {base|mtp|both} for QAD / MTP QAT (frozen-base, frozen-MTP, or co-train)

yeyu-nvidia:
## Summary

Enable MTP (multi-token prediction) head training **and export** on ModelOpt-quantized models, for both GPT and hybrid (Mamba+MoE) architectures. Three commits:

1. **`--freeze-base-for-mtp`** — freeze all base-model parameters and train only MTP heads. Adds `_freeze_base_for_mtp()` (sets `requires_grad=False` on all non-`mtp.layers.*` params) and builds `mtp_block_spec` in `modelopt_gpt_hybrid_builder()` so MTP heads can be added to a ModelOpt-quantized GPT model.
2. **Unit tests** for the freeze behavior.
3. **MTP block spec for the ModelOpt HybridModel local spec** (`model_specs.py`) — adds `mtp_block_spec` (enorm / hnorm / eh_proj / final layernorm via non-TE `Norm` + `ColumnParallelLinear`) to `HybridStackSubmodules`. Without this, `MCoreHybridModel` cannot build or export MTP for hybrid architectures, so MTP weights were silently dropped on export.

**Use case:** After NVFP4 QAD, load the quantized checkpoint with `--mtp-num-layers 1 --mtp-hybrid-override-pattern "*E" --freeze-base-for-mtp`. MTP heads are added and trained with lm_loss while the quantized base stays frozen; the MTP weights then export to HF format for vLLM speculative decoding.

## Test plan

- [x] Unit test: build model with `freeze_base_for_mtp=True`, verify only `mtp.layers.*` params have `requires_grad=True` (`tests/unit_tests/post_training/test_freeze_base_for_mtp.py`)
- [x] End-to-end (hybrid): full NVFP4 QAD → QAT-MTP → export → vLLM pipeline on Nemotron-3.5-Nano (OCI-HSG `cicd_1779387587`). Export emits 1059 MTP tensors (NVFP4) / 270 (BF16); MTP QAT loss decreases with base frozen.
- [x] Acceptance length on MT-Bench (num_speculative_tokens=3): BF16 AL=2.077 (3.077 w/ bonus), NVFP4 AL=1.832 (2.832 w/ bonus), vs AL=1.00 / 0 accepted before the fix.

## Notes

- Rebased onto current `main` (de-conflicted; the only conflict was a now-redundant local import of `get_hybrid_stack_modelopt_spec`, which `main` already imports at module scope).
- Downstream integration: omniml/integration/nmm-sandbox!159 pins this commit.

### DIFF
--- a/megatron/post_training/model_builder.py
+++ b/megatron/post_training/model_builder.py
@@ -157,6 +157,67 @@ def _build_teacher_model(config, config_raw: Namespace, model_kwargs: Dict[str, 
     return teacher
 
 
+def _freeze_for_qad(model, target):
+    """Select which side of an MTP model trains during QAD / MTP QAT.
+
+    Splits parameters into the MTP heads (``mtp.layers.*``) and the base model,
+    and freezes one side so controlled QAD+MTP experiments can be run:
+
+    * ``"mtp"``  — train the MTP heads only, freeze the base. Used after QAD:
+      load a quantized checkpoint, add MTP heads, and train them while the
+      quantized base stays fixed (the production two-phase recipe).
+    * ``"base"`` — train the base only, freeze the MTP heads. QAD on the base
+      with the MTP head held at its init (e.g. measuring how well a frozen MTP
+      head rides on a quantizing base).
+    * ``"both"`` — train the base and the MTP heads together (QAD co-training).
+    """
+    if target not in ("mtp", "base", "both"):
+        raise ValueError(f"qad train target must be one of mtp/base/both, got {target!r}")
+
+    if target == "both":
+        for param in model.parameters():
+            param.requires_grad = True
+        # Nothing is frozen, so no router expert_bias should be pinned.
+        for module in model.modules():
+            if hasattr(module, 'expert_bias'):
+                module.frozen_expert_bias = False
+        print_rank_0("QAD train target 'both': all parameters trainable")
+        return
+
+    train_mtp = target == "mtp"
+    trainable, frozen = 0, 0
+    for name, param in model.named_parameters():
+        is_mtp = 'mtp.layers.' in name
+        param.requires_grad = is_mtp == train_mtp
+        if param.requires_grad:
+            trainable += 1
+        else:
+            frozen += 1
+
+    # The MoE router's expert bias is updated from load-balancing token counts in
+    # finalize_model_grads._update_router_expert_bias, independently of requires_grad.
+    # Setting requires_grad=False does NOT stop it, so the frozen side would keep
+    # drifting. Flag the frozen side's routers so the update is skipped; the trainable
+    # side's routers must keep updating (so we clear the flag there).
+    frozen_bias = 0
+    for name, module in model.named_modules():
+        if hasattr(module, 'expert_bias'):
+            is_mtp = 'mtp.layers.' in name
+            freeze_this = is_mtp != train_mtp
+            module.frozen_expert_bias = freeze_this
+            if freeze_this:
+                frozen_bias += 1
+    print_rank_0(
+        f"QAD train target '{target}': training {'MTP' if train_mtp else 'base'} "
+        f"({trainable} trainable, {frozen} frozen, {frozen_bias} router expert_bias frozen)"
+    )
+
+
+def _freeze_base_for_mtp(model):
+    """Deprecated alias for ``_freeze_for_qad(model, "mtp")``."""
+    _freeze_for_qad(model, "mtp")
+
+
 def modelopt_gpt_hybrid_builder(
     args,
     pre_process,
@@ -260,6 +321,22 @@ def modelopt_gpt_hybrid_builder(
                 use_arbitrary_attention_mask=False,
             )
 
+        # Build MTP block spec if MTP is enabled.
+        mtp_block_spec = None
+        if args.mtp_num_layers is not None:
+            from megatron.core.models.gpt.gpt_layer_specs import (
+                get_gpt_decoder_layer_specs,
+                get_gpt_mtp_block_spec,
+            )
+
+            use_te = args.transformer_impl == "transformer_engine"
+            decoder_layer_specs = get_gpt_decoder_layer_specs(
+                config, use_transformer_engine=use_te,
+            )
+            mtp_block_spec = get_gpt_mtp_block_spec(
+                config, decoder_layer_specs[-1], use_transformer_engine=use_te,
+            )
+
         model_kwargs = {
             "transformer_layer_spec": transformer_layer_spec,
             "vocab_size": args.padded_vocab_size,
@@ -273,6 +350,7 @@ def modelopt_gpt_hybrid_builder(
             "rotary_percent": args.rotary_percent,
             "rotary_base": args.rotary_base,
             "rope_scaling": args.use_rope_scaling,
+            "mtp_block_spec": mtp_block_spec,
             "pg_collection": pg_collection,
         }
         model = MCoreGPTModel(config=config, **model_kwargs)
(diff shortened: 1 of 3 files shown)


AAnoosheh:
I have something almost the same in [my PR](https://github.com/NVIDIA/Megatron-LM/pull/5019) (`--freeze-all-layers`) except mine also stops the router bias from updating (which is sneakily independent of requires_grad !)

Let's see if we can consolidate into one argument if possible.

yeyu-nvidia:
Thanks @AAnoosheh — good catch, confirmed. `_update_router_expert_bias` in `finalize_model_grads.py` gates only on `module.training` and updates `expert_bias` via `copy_()` from token counts, so `requires_grad=False` doesn't touch it. On Nemotron-3.5-Nano (`--moe-router-enable-expert-bias`) my "frozen" base was silently drifting during MTP-head training.

I've pushed a fix using the **same `frozen_expert_bias` flag** as your PR, so we're already aligned on the mechanism:
- `_freeze_base_for_mtp` sets `module.frozen_expert_bias = True` on non-MTP routers (the MTP block's own routers stay updatable)
- same gate in `_update_router_expert_bias`: `... and not getattr(module, 'frozen_expert_bias', False)`

The only real difference is scope: yours freezes **all** layers, mine freezes **all-except-MTP**. Happy to consolidate — proposal:
1. Factor a shared helper like `freeze_parameters(model, except_patterns=[...])` that sets both `requires_grad=False` and `frozen_expert_bias=True`; `--freeze-all-layers` → `except=[]`, `--freeze-base-for-mtp` → `except=['mtp.layers.']`.
2. Heads-up on merge order: we both add the identical gate line in `finalize_model_grads.py`, so whichever lands second will conflict there — let's make sure only one of us owns that hunk (or I rebase onto yours).

Which would you prefer — I extract the shared helper into this PR, or fold the MTP carve-out into #5019?

megatron/core/post_training/modelopt/hybrid/model_specs.py
```diff
@@ -166,12 +172,31 @@ def _get_hybrid_stack_local_spec(
         ),
     )
 
+    mtp_block_spec = ModuleSpec(
```
jenchen13, line 240: does MTP training also work in TE spec (not local spec)? we may deprecate local spec in the future

jenchen13 (commented):
is there an example in our internal sandbox CI?

megatron/core/post_training/modelopt/hybrid/model_specs.py
yeyu-nvidia: Yes — the TE spec already supports MTP. `get_hybrid_stack_modelopt_spec(use_default_te_spec=True)` returns the standard `hybrid_stack_spec`, which carries `mtp_block_spec=_hybrid_mtp_block_spec` (TENorm-based) at `hybrid_layer_specs.py:189`, and the no-grouped-gemm TE variant forwards it explicitly at `model_specs.py:49`. This change only adds the equivalent to the **local** (non-TE) spec, which was the path missing it — bringing local to parity rather than being MTP's only route.

My end-to-end validation (Nemotron-3.5-Nano NVFP4 QAD → MTP QAT → export) ran through the default local spec, so that's the exercised path; I haven't run the full pipeline through `use_default_te_spec=True` yet — happy to. Since TE already covers MTP, this addition is low-risk and isolated to the non-TE branch, so it can be dropped cleanly if/when local spec is removed.

megatron/post_training/model_builder.py
```diff
@@ -157,6 +157,37 @@ def _build_teacher_model(config, config_raw: Namespace, model_kwargs: Dict[str,
...
+            frozen += 1
+
+    # The MoE router's expert bias is updated from load-balancing token counts
+    # in finalize_model_grads._update_router_expert_bias, independently of
+    # requires_grad. Setting requires_grad=False above does NOT stop it, so the
+    # base would keep drifting during MTP training. Flag non-MTP routers so the
+    # update is skipped (the MTP block's own routers must keep updating).
+    frozen_bias = 0
+    for name, module in model.named_modules():
+        if hasattr(module, 'expert_bias') and 'mtp.layers.' not in name:
```
AAnoosheh: Might be safer to do `hasattr(module, 'frozen_expert_bias')` – not sure if ever possible that we have something with `.expert_bias` but not `.frozen_expert_bias`

AAnoosheh (approved):
Other than that, LGTM

megatron/post_training/model_builder.py
yeyu-nvidia: The gate in `finalize_model_grads._update_router_expert_bias` reads this as `getattr(module, 'frozen_expert_bias', False)`, so a module with `.expert_bias` but no `.frozen_expert_bias` is handled safely — it's simply treated as not-frozen. That's actually the normal case for the MTP block's own routers (which intentionally keep updating), so the asymmetry is deliberate rather than a gap.

I set the flag only on base routers here instead of initializing `frozen_expert_bias=False` in the router `__init__` (as your PR does) to keep this isolated to the post-training path. When we consolidate with `--freeze-all-layers`, happy to make the router `__init__` the single source of truth and drop the dynamic set. Thanks for the review + approve!

dimapihtar (approved):

yeyu-nvidia:
Rebased onto current `main` and slimmed the diff.

While rebasing I found `main` has **independently landed** both pieces this PR originally carried in `megatron/core/`:
- the `frozen_expert_bias` guard in `finalize_model_grads._update_router_expert_bias`, and
- `mtp_block_spec` in the hybrid stack spec.

So the only remaining content in those two core files was explanatory comments. I dropped those two comment-only hunks, so the PR now touches **only `megatron/post_training/`**:
- `arguments.py` — the `--freeze-base-for-mtp` flag
- `model_builder.py` — `_freeze_base_for_mtp()`, which freezes everything except `mtp.layers.*` and sets `frozen_expert_bias=True` on the frozen base routers (read by the already-merged guard in `finalize_model_grads`)
- `tests/unit_tests/post_training/test_freeze_base_for_mtp.py`

**Pipeclean (B200):** validated end-to-end — NVFP4 QAD → MTP-QAT on frozen quantized base → export produced a HF checkpoint with **270 MTP tensors**; vLLM loads it with MTP detected and target-model weight sharing. AL benchmark re-running now.

Now `MERGEABLE` / `APPROVED`. Ready to merge.

yeyu-nvidia:
**Scope update** (per request): generalized the binary `--freeze-base-for-mtp` into **`--qad-train-target {base|mtp|both}`** so the full controlled QAD+MTP experiment can be driven from one knob, instead of a flag now + a rename later.

| `--qad-train-target` | trains | freezes |
|---|---|---|
| `mtp` | MTP heads | base (the previous `--freeze-base-for-mtp` behavior; production two-phase recipe) |
| `base` | base | MTP heads |
| `both` | base + MTP | nothing |

`_freeze_for_qad(model, target)` freezes the non-target side and, for MoE, pins `frozen_expert_bias` on the **frozen** side's routers (and clears it on the trainable side), since that update bypasses `requires_grad`. **`--freeze-base-for-mtp` is kept as a deprecated alias** for `--qad-train-target mtp`, so nothing downstream breaks.

Why: this supports a controlled study of how the MTP head should be treated during NVFP4 QAD — base-trained with MTP frozen (`base`), co-trained (`both`), vs the existing MTP-on-frozen-base recipe (`mtp`). Tests extended to cover all three directions + the alias. Lint/DCO green locally with the pinned tools.

PR #4790, opened 2026-05-14
Fix MIMO optimizer setup for frozen modules

liding-nv:
## Summary

  The MIMO optimizer's per-module `param_state_sharding_type` was lost across a distributed-checkpoint save/load round-trip. In non-colocated MIMO each rank owns a different module subset (e.g. rank 0 owns language, rank 1 owns vision), so this metadata is not common across ranks and dist_checkpointing's common-state path drops it. After load, the inner optimizer then mis-sharded its state. Wrap it in a `ShardedObject` at save time and unwrap it on load so each rank's value survives.

  - New `_extract_param_state_sharding_type` / `_restore_param_state_sharding_type` helpers in `mimo/optimizer.py` round-trip the value through a per-module `ShardedObject` keyed `optimizer.mimo.{module}{suffix}.param_state_sharding_type`.
  - `_extract_param_groups`: drop the now-empty `'optimizer'` sub-dict after popping `param_groups` so dist_checkpointing's common-state path doesn't carry an empty dict that gets discarded inconsistently.
  - `MimoOptimizer` save/load callsites invoke the new helpers alongside the existing `_extract_param_groups` / `_restore_param_groups` and  `_extract_grad_scaler` / `_restore_grad_scaler` calls.

  ## Tests

  CPU-only unit tests added inside the existing file:
  - `test_mimo_checkpoint.py::TestOptimizerCheckpointHelpers` — `_extract_param_state_sharding_type` happy path / suffix /  no-op-when-missing; `_restore_param_state_sharding_type` rename / no-op-when-missing; _extract_param_groups` empty-optimizer-cleanup /  negative case; `_restore_param_groups` recreates missing wrapper.


liding-nv:
@yashaswikarnati  can u take a look? thx

megatron/core/optimizer/__init__.py
```diff
@@ -966,6 +972,7 @@ def get_megatron_optimizer(
     use_gloo_process_groups: bool = True,
     pg_collection: Optional[ProcessGroupCollection] = None,
     dump_param_to_param_group_map: Optional[str] = None,
+    param_group_process_group: Optional[torch.distributed.ProcessGroup] = None,
```
yashaswikarnati: I wonder if pg_collection.intra_dist_opt can be directly used here instead of passing a new arg  param_group_process_group. pg_collection passed here ideally should have all the process group that the optimizer would need

liding-nv: Yeah that's actually what I had first by auto-deriving it from `pg_collection.intra_dist_opt` inside `get_megatron_optimizer`. Turns out the layer-wise distopt also passes a `pg_collection` (test`test_overlap_param_gather_hook_enable_disable_cycle[False]`) and its `intra_dist_opt` isn't a globally-consistent group for that `all_gather`, so the collective deadlocked → CI hung 60min and got SIGTERM'd.

lmk if there's a sane way to keep layer-wise working.

PR #4810, opened 2026-05-14
Fix CUDA IMA in fsdp_double_buffer when an FSDP unit's bucket doesn't fit the pool

wujingyue:
## Summary

When `DistributedDataParallelConfig.fsdp_double_buffer=True` is enabled together with a hybrid model whose FSDP units have differently-shaped buckets (e.g. attention + Mamba), training fails with a CUDA illegal memory access during backward.

This PR replaces `FixedPoolAllocator`'s backup allocator with `StorageResizeBasedBucketAllocator` (already the default for the non-double-buffer path) and adds a regression test.

## Reproducer

A new test, `TestFsdpHybridModelDoubleBuffer::test_train_steps_with_double_buffer`, builds a 1-attention-layer + 1-mamba-layer stack, marks both as FSDP unit modules, and runs 5 training steps with `fsdp_double_buffer=True`. Without the fix the test aborts with `CUDA error: an illegal memory access was encountered` during the first backward; with the fix it passes.

## Root cause

When `fsdp_double_buffer=True`, `FixedPoolAllocator` is used as the weight allocator. Its pool only serves FSDP units whose bucket structure matches the rep-unit (same per-bucket dtype + total size). Any unit that doesn't match falls through to the pool's `backup_allocator`.

The hybrid model triggers exactly this: bucket 0 (TransformerLayer, 263 KB) and bucket 1 (MambaLayer, 933 KB) don't satisfy `_is_two_bucket_group_equal`, so only TransformerLayer is pooled; MambaLayer goes through the backup.

The backup was `TemporaryBucketAllocator`, whose `free` calls `_free_storage` *and* deletes the `Bucket` entry — discarding the `Storage` object. The next `allocate` builds a fresh `Bucket` whose `data = torch.empty(...)` has a brand-new `Storage`.

This interacts badly with how `causal_conv1d_fn` saves its weight argument:

```python
xBC = causal_conv1d_fn(
    x=xBC,
    weight=rearrange(self.cp.get_conv1d_weight(), "d 1 w -> d w"),  # non-leaf view
    bias=self.cp.get_conv1d_bias(),
    ...
)
```

`CausalConv1dFn.forward` then calls `ctx.save_for_backward(x, weight, bias, ...)`. PyTorch's `SavedVariable` for a **non-leaf** tensor freezes the tensor's `Storage` reference at save time — unlike a leaf, where it re-resolves `.data` on `unpack()`. So the saved `weight` permanently references the same `Storage` Python/C++ object.

Lifecycle in the failing case (verified empirically):

```
fetch_bucket(bid=1, set_param_data=True) → Storage S0 allocated (1.78 MB), conv1d.weight rebound, causal_conv1d_fn saves view of S0
post-fwd: free_bucket_storage(bid=1)     → S0 is _resize_(0)'d AND the Bucket+Storage entry is deleted (TBA path)
pre-bwd:  fetch_bucket(bid=1, ...)       → Storage S1 allocated (different Storage object, possibly at same caching-allocator address)
                                            conv1d.weight.data rebound to view of S1 — but the saved view in autograd still references S0
backward: causal_conv1d_bwd derefs saved view → reads from S0 (size 0 forever) → IMA
```

`StorageResizeBasedBucketAllocator` avoids this by retaining `self.buckets[bucket_id]` (and therefore its `Storage` object) across `free`/`allocate` cycles: `free` calls `_resize_(0)`, the next `allocate` calls `_resize_(size)`. The caching allocator may return a different address on `_resize_(size)` than the forward pass had — and that's fine. A non-leaf `SavedVariable` stores a shallow copy of the TensorImpl that *shares* the original `Storage`; the saved view's sizes/strides/storage_offset are frozen, but its `data_ptr()` is computed at access time as `storage.data_ptr() + storage_offset * itemsize`, reading the storage pointer fresh on every dereference. So when the kernel dispatches at backward, it picks up whatever address the pre-bwd `_resize_(size)` just installed (and that's the address the all-gather wrote to).

## Why this hasn't shown up before

Most FSDP-managed params are passed directly to ops (e.g. `F.linear(x, self.weight)`), so they're saved as **leaves**, and autograd re-resolves `.data` at unpack — `p.data = new_view` propagates. MambaMixer is the first case in the tree that constructs a non-leaf view of a bucket-managed param (`rearrange(...)`) *before* a `save_for_backward`, defeating the leaf re-resolution path. The crash therefore requires the conjunction of (a) `fsdp_double_buffer=True`, (b) a non-pooled FSDP unit, and (c) a kernel that saves a non-leaf view of one of that unit's params.

## Test plan

- [x] `tests/unit_tests/distributed/megatron_fsdp/test_mcore_fully_sharded_data_parallel.py::TestFsdpHybridModelDoubleBuffer` — fails with IMA before the fix, passes after.
- [x] Run pretrain_hybrid.py to compare base and PR for a non-Mamba, hybrid model: https://github.com/NVIDIA/Megatron-LM/pull/4810#issuecomment-4513899983
- [x] Run pretrain_hybrid.py with a TransformerLayer+MambaLayer hybrid model: https://github.com/NVIDIA/Megatron-LM/pull/4810#issuecomment-4513912606
- [x] CI: full FSDP unit-test suite.


wujingyue:
# PR 4810 FSDP Double-Buffer Backup Allocator Experiment

## Goal

Test the effect of replacing the FSDP fixed-pool fallback allocator:

- Base: `TemporaryBucketAllocator`
- PR 4810: `StorageResizeBasedBucketAllocator`

The focus was double buffering when one FSDP unit fits the fixed pool and another unit falls back to the backup allocator.

## Hardware

Ran locally on **2 x NVIDIA GB200 GPUs** out of a 4-GPU GB200 node.

Each GPU:

- `NVIDIA GB200`
- `189471 MiB`
- driver `580.159.04`

The training command used `--nproc_per_node=2`.

## Test Setup

Used `pretrain_hybrid.py` directly.

Key args:

```text
--use-megatron-fsdp
--fsdp-double-buffer
--overlap-grad-reduce
--overlap-param-gather
--data-parallel-sharding-strategy optim_grads_params
--ddp-bucket-size 4096
--ckpt-format fsdp_dtensor
--hybrid-layer-pattern "*-"
--spec megatron.core.models.hybrid.hybrid_layer_specs hybrid_stack_spec
```

The `*-` hybrid pattern avoids `MambaLayer` and creates attention + MLP units with different bucket sizes. That still exercises the fixed-pool fallback path while isolating the allocator change from Mamba-specific behavior.

## Result

Both runs completed 5 iterations and saved checkpoints successfully.

| Run | Allocator | Status | Max allocated | Max reserved |
| --- | --- | --- | --- | --- |
| Base | `TemporaryBucketAllocator` | passed 5 iters | `252.78 MB` | `344 MB` |
| PR 4810 | `StorageResizeBasedBucketAllocator` | passed 5 iters | `139.53 MB` | `188 MB` |

Both runs had the same model size:

```text
number of parameters: 2,903,808
```

Current allocated memory after iterations was identical:

```text
allocated: 127.46 MB
```

So the PR did not reduce persistent model state. It reduced transient peak memory.

## Interpretation

`StorageResizeBasedBucketAllocator` was meaningfully better in this repro:

- peak allocated dropped by about `113 MB`
- peak reserved dropped by about `156 MB`
- roughly `45%` lower peak allocated memory in this small run

The likely reason is allocator behavior. The base allocator frees storage and deletes the bucket entry, so later use creates a fresh CUDA tensor/storage. The PR keeps the bucket object stable and resizes its storage to/from zero. Under overlapped all-gather and reduce-scatter, that avoids extra CUDA allocator churn and lowers transient memory pressure.

## Conclusion

PR 4810 does not regress the double-buffer memory profile in the no-Mamba hybrid repro. It improves peak memory behavior for the fallback allocator path while preserving successful training behavior.

wujingyue:
# PR 4810 Mamba FSDP-Unit Validation

## Goal

Validate PR 4810 with a hybrid model where both `TransformerLayer` and `MambaLayer` are explicit Megatron FSDP units.

This specifically exercises the case where:

- the `TransformerLayer` FSDP unit fits in the FSDP double-buffer fixed pool
- the `MambaLayer` FSDP unit does not fit in that fixed pool
- the `MambaLayer` path therefore uses the fixed-pool fallback allocator

No base comparison was run, because base is expected to crash due to the bug fixed by PR 4810.

## Local Code Change

Applied this local-only change:

```python
from megatron.core.ssm.mamba_layer import MambaLayer
```

and changed the default Megatron FSDP unit list for `optim_grads_params` from:

```python
self.fsdp_unit_modules = [TransformerLayer]
```

to:

```python
self.fsdp_unit_modules = [TransformerLayer, MambaLayer]
```

This change was intentionally kept local and was not added to the PR.

## Hardware

Ran locally on **2 x NVIDIA GB200 GPUs** out of a 4-GPU GB200 node.

Each GPU:

- `NVIDIA GB200`
- `189471 MiB`
- driver `580.159.04`

The training command used `--nproc_per_node=2`.

## Test Setup

Used `pretrain_hybrid.py` directly.

Key args:

```text
--use-megatron-fsdp
--fsdp-double-buffer
--overlap-grad-reduce
--overlap-param-gather
--data-parallel-sharding-strategy optim_grads_params
--ddp-bucket-size 4096
--ckpt-format fsdp_dtensor
--hybrid-layer-pattern "*M"
--spec megatron.core.models.hybrid.hybrid_layer_specs hybrid_stack_spec
--train-iters 50
```

The run used the PR 4810 allocator behavior:

```python
self.backup_allocator = StorageResizeBasedBucketAllocator()
```

## Result

The run completed successfully.

```text
exit status: 0
completed iterations: 50
checkpoint saved: yes
```

## Model Shape

The model used the `*M` hybrid pattern:

- layer 0: `TransformerLayer`
- layer 1: `MambaLayer`

Parameter counts from the log:

```text
TransformerLayer params: 263680
MambaLayer params:      933400
total params:           3311128
```

This confirms the intended heterogeneous FSDP-unit case: the Transformer unit is much smaller than the Mamba unit.

## Memory

Peak rank-0 memory reported by Megatron:

```text
max allocated: 389.96 MB
max reserved:  604.00 MB
```

Final memory after checkpoint save:

```text
allocated:     132.12 MB
max allocated: 389.96 MB
reserved:      340.00 MB
max reserved:  604.00 MB
```

## Loss Trend

The loss decreased over the run.

```text
iter 1 loss:   9.063955
iter 50 loss:  7.059249
first 5 avg:   9.010403
last 5 avg:    7.117695
min loss:      6.971731 at iter 36
```

Selected iteration losses:

```text
1  9.063955
2  9.069407
3  9.007366
4  9.028999
5  8.882289
...
46 6.980014
47 7.249215
48 7.219358
49 7.080637
50 7.059249
```

## Conclusion

PR 4810 successfully handled the hybrid `TransformerLayer + MambaLayer` Megatron FSDP double-buffer case when `MambaLayer` was included as an explicit FSDP unit.

The run did not crash, completed 50 iterations, saved a checkpoint, and showed a clear downward loss trend. This validates the target scenario for PR 4810: a hybrid model where one FSDP unit fits in the double-buffer fixed pool and the larger Mamba FSDP unit falls back to the new `StorageResizeBasedBucketAllocator`.

wujingyue:
@Autumn1998 @shjwudp @cspades I added more test results to the PR description. PTAL

cspades:
BTW, I think this can be fixed with my new MaxPoolAllocator that supports all FSDP units (among many other things, it's just straight up better than FixedPoolAlloc at some cost to memory), so we might not need this anymore!

However, I am not sure if we still need this PR, i.e. avoid CUDA cache dealloc and always resize in the presence of NCCL UBR. Do we have a NCCL UBR storage-resize on registered buffer test yet?

wujingyue:
> BTW, I think this can be fixed with my new MaxPoolAllocator that supports all FSDP units (among many other things, it's just straight up better than FixedPoolAlloc at some cost to memory), so we might not need this anymore!

If we remove FixedPoolAllocator **completely**, I agree this fix is not necessary. 

> Do we have a NCCL UBR storage-resize on registered buffer test yet?

I'm not sure how that's related to this PR. This PR fixes the allocation/deallocation of **non-UBR unit**.

cspades (approved):
Can't see much wrong with this, so if nobody is objecting I'll approve.

Phlip79 (approved):

jaredcasper (approved):

PR #4811, opened 2026-05-14
ci: declare workflow-level `contents: read` on 7 read-only workflows

arpitjain099:
# What does this PR do?

Pins the default `GITHUB_TOKEN` to `contents: read` on the 7 workflows in `.github/workflows/` that don't actually use the default token for any write operation:

- `auto-reminder-bot.yml`: scheduled cron reminder, no GitHub API.
- `auto-update-copy-pr-bot.yml`: regenerates `.github/copy-pr-bot.yaml` from mcore team membership. The `gh api` calls and the final `git push` both go through `secrets.PAT`, not `GITHUB_TOKEN`.
- `cicd-approve-test-queue.yml`: runs `gh release download v0.1.0`. Release-asset read is covered by `contents: read`.
- `community-bot.yml`: community-PR labeling helper, no API mutation in the GITHUB_TOKEN path.
- `review-trigger.yml`: triggers downstream workflows, no `GITHUB_TOKEN` use.
- `sync-skills.yml`: filesystem sync only.
- `trigger-mbridge-tests.yml`: triggers a downstream test pipeline.

Left implicit on purpose:

- `build-docs.yml`, `copyright-check.yml`, `install-test.yml`, `multi-approval-bot.yml` all use `gh run view` with `github.token`. That needs `actions: read` on top of `contents: read`, which I'd rather leave to a maintainer who knows whether to grant it at workflow or job scope.
- `sync-team-usergroups.yml` falls back to `GITHUB_TOKEN` when no `NVIDIA_MCORE_ONCALL_TOKEN`/`PAT` is configured, and the script's actual scope needs aren't clear from outside.
- The release/cherry-pick/stale/close-inactive workflows obviously need write scopes and are not touched here.

## Issue tracking

Linked issue: N/A. Standalone CI hygiene change.

## Pre-checks

- [x] YAML validated locally with `yaml.safe_load` on each touched file.
- [x] No functional change. Only the scope of the implicit `GITHUB_TOKEN` is narrowed.
- [x] Workflows whose writes use `secrets.PAT` are unaffected by this scope change.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. The leaked token retained whatever scope was issued at the workflow level. Pinning per workflow caps that runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

### DIFF
--- a/.github/workflows/auto-reminder-bot.yml
+++ b/.github/workflows/auto-reminder-bot.yml
@@ -7,6 +7,9 @@ on:
   schedule:
     - cron: "0 12 * * *"
 
+permissions:
+  contents: read
+
 jobs:
   run-script:
     name: Run Auto Reminder Bot
--- a/.github/workflows/auto-update-copy-pr-bot.yml
+++ b/.github/workflows/auto-update-copy-pr-bot.yml
@@ -5,6 +5,9 @@ on:
   schedule:
     - cron: "0 0 * * *"
 
+permissions:
+  contents: read
+
 jobs:
   auto-update-copy-pr-bot:
     runs-on: ubuntu-latest
--- a/.github/workflows/cicd-approve-test-queue.yml
+++ b/.github/workflows/cicd-approve-test-queue.yml
@@ -19,6 +19,9 @@ on:
     - cron: "*/5 * * * *" # Runs every 5 minutes
   workflow_dispatch: # Allows manual triggering
 
+permissions:
+  contents: read
+
 jobs:
   approve-queue:
     runs-on: ubuntu-latest
--- a/.github/workflows/community-bot.yml
+++ b/.github/workflows/community-bot.yml
@@ -19,6 +19,9 @@ on:
   issue_comment:
     types: [created, edited, deleted]
 
+permissions:
+  contents: read
+
 jobs:
   community-bot:
     uses: NVIDIA-NeMo/FW-CI-templates/.github/workflows/_community_bot.yml@v0.65.10
--- a/.github/workflows/community-bot.yml
+++ b/.github/workflows/community-bot.yml
@@ -21,6 +21,8 @@ on:
 
 permissions:
   contents: read
+  issues: write
+  pull-requests: write
 
 jobs:
   community-bot:
--- a/.github/workflows/review-trigger.yml
+++ b/.github/workflows/review-trigger.yml
@@ -9,6 +9,9 @@ on:
   pull_request_review:
     types: [submitted]
 
+permissions:
+  contents: read
+
 jobs:
   signal:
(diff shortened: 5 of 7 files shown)


Connor-XY:
Hello @arpitjain099, thanks for the workflow-permissions hardening.

I checked the changed files and this is limited to read-only GitHub Actions workflows where declaring workflow-level `contents: read` should reduce the token scope without changing repository behavior. The main review question is whether any of these workflows rely on implicit write permissions through a called action or future step.

I’m routing this to @ko3n1g for the CI/workflow ownership path.

arpitjain099:
Thanks @Connor-XY, and good question - that is the exact thing worth pinning down here. I went back through all seven files and traced which token each write actually uses.

Six of the seven do not touch the default GITHUB_TOKEN for any mutation:

- auto-reminder-bot.yml: only checkout + a python script that talks to Slack and GitHub via secrets.PAT (GH_TOKEN).
- auto-update-copy-pr-bot.yml: the gh api calls and the git push both run with secrets.PAT (checkout also pulls with the PAT), so the default token is read-only.
- cicd-approve-test-queue.yml: the deployment-approval POST runs with GITHUB_TOKEN set to secrets.PAT, and the release download uses NVIDIA_MANAGEMENT_ORG_PAT. The default token only does checkout.
- review-trigger.yml: just writes a PR number to disk and uploads it with actions/upload-artifact. Artifact upload does not need a repo write scope.
- sync-skills.yml: calls _sync_skills.yml, which does git push, but its checkout runs with token: secrets.PAT, so the push is authenticated by the persisted PAT credential, not the default token.
- trigger-mbridge-tests.yml: convictional/trigger-workflow-and-wait is passed github_token: secrets.PAT to dispatch the downstream repo.

The one that does rely on an implicit write is community-bot.yml. It calls NVIDIA-NeMo/FW-CI-templates _community_bot.yml, and inside that reusable workflow the job env sets GITHUB_TOKEN to secrets.GITHUB_TOKEN (the default token). The "Add community-request label" and "Remove community-request label" steps have no per-step token override, so their gh issue edit calls use that default token, which is why the reusable workflow declares issues: write and pull-requests: write at the job level. Since a caller permissions block is the ceiling for a called workflow, capping the caller at contents: read alone would drop those scopes and the label steps would 403.

I pushed a fix for that: community-bot.yml now grants contents: read plus issues: write and pull-requests: write, matching what the reusable workflow needs. The other six stay at contents: read. Happy to adjust if @ko3n1g would rather scope community-bot at the job level or leave it untouched.

PR #4817, opened 2026-05-15
[Dev][opt] Optimize e_proj and h_proj TP communication for MTP with mHC

Baibaifan:
### Optimize MTP mHC projection gather

In the mHC MTP path, `e_proj` and `h_proj` are both column-parallel projections on the same tensor-parallel group with `gather_output=False`. Their outputs therefore contain matching hidden partitions.

Before

image

Cost:
- 2x TP all-gather
- 2x full hidden activations before add

After

image

Cost:
- 1x TP all-gather
- no extra full hidden activation before add

Because e_proj and h_proj produce the same TP hidden partition, we can add locally first and gather once.


Victarry:
LGTM. cc @hxbai

Victarry (approved):

hxbai (approved):
LGTM. Thanks for your optimization.

PR #4822, opened 2026-05-15
Allow YAML MoE configs to use model specs

chawkins-nvidia:
## Summary

Remove a YAML-only validation assertion that rejects MoE configs when `spec` is set.

Hybrid model launches use `spec` to construct the model stack, and the argparse validation path already allows MoE configs with a model spec. This makes the `--yaml-cfg` path consistent with that behavior while preserving the existing MoE tensor-parallel and expert-parallel checks.

## Validation

- `python3 -m py_compile megatron/training/yaml_arguments.py`


maanug-nv (approved):

deepakn94 (approved):

deepakn94:
/ok to test [8ae1bef](https://github.com/NVIDIA/Megatron-LM/pull/4822/commits/8ae1bef640b8c4364196af41cf737d2420d1df08)

PR #4825, opened 2026-05-15
Add opt-in MXFP8 LM-head output projection

gdengk:
## What does this PR do ?

Adds an opt-in TE-based LM-head ColumnParallelLinear that runs under the MXFP8 autocast context. Controlled by the new `fp8_output_proj` config flag; active only when `fp8=True` and `fp8_recipe='mxfp8'`. Defaults to off, so existing flows are unaffected.

This is the main-branch equivalent of #4484 and #4489 (merged to `26.04-alpha`), ported as a self-contained module instead of layering on the alpha-only `LinearCrossEntropyModule` wrapper.

### Changes
- New module `megatron/core/transformer/mxfp8_output_proj.py` with `TELinearCrossEntropyModule` and `is_te_mxfp8_output_proj_active`.
- `GPTModel.__init__` conditionally swaps the output layer when the gate is on.
- New `TransformerConfig.fp8_output_proj` field (default `False`).

### Tests
`tests/unit_tests/transformer/test_mxfp8_output_proj.py`:
- Gating helper across config combinations (pure-Python).
- Constructor validations for unsupported options.
- GPTModel default uses `ColumnParallelLinear`.
- GPTModel uses `TELinearCrossEntropyModule` under mxfp8 (Blackwell-gated).

### Pre-checks


megatron/core/transformer/mxfp8_output_proj.py
```diff
@@ -0,0 +1,142 @@
...
+                f"{prefix}_extra_state"
+            )
+        )
+
+    def get_extra_state(self):
+        # MXFP8 has no persistent recipe state; keep LM head _extra_state empty
+        # so GPTModel.sharded_state_dict's no-extra-state invariant still holds.
+        return None
+
+    def set_extra_state(self, state):
```
Phlip79: Add a warning here in case someone attempts to call `set_extra_state`?

Phlip79 (approved):

santhnm2 (approved):
LGTM, but is this only applicable for `GPTModel` or is this also relevant for `HybridModel`?

Phlip79:
> LGTM, but is this only applicable for `GPTModel` or is this also relevant for `HybridModel`?

I'll add support in a follow-up PR for this. There are a few other features that also need to be added to `HybridModel`.

megatron/core/transformer/mxfp8_output_proj.py
```diff
@@ -0,0 +1,144 @@
...
+        return False
+    if not getattr(config, "fp8", False):
+        return False
+
+    fp8_recipe = getattr(config, "fp8_recipe", None)
+    recipe_value = getattr(fp8_recipe, "value", fp8_recipe)
+    return str(recipe_value).lower() == "mxfp8" or str(fp8_recipe).lower().endswith(".mxfp8")
+
+
+class TELinearCrossEntropyModule(TEColumnParallelLinear):
```
jaredcasper: Why is this called TELinearCrossEntropyModule if it doesn't have cross entropy in it? A more descriptive name that says how it's different than normal TEColumnParallelLinear layer would be better? And we typically don't have "Module" in the name like that.

Is the only different that it runs under an MXFP8 context? But the normal TELinear has a quant_context() that it runs under? Why is this whole separate module necessary? It's not clear from the description.

megatron/core/transformer/mxfp8_output_proj.py
```diff
@@ -0,0 +1,144 @@
...
+    """TE ColumnParallelLinear variant for the LM head under MXFP8.
+
+    Active only when ``fp8_output_proj=True`` with ``fp8_recipe='mxfp8'``. Acts
+    as a drop-in replacement for the plain ColumnParallelLinear used as the GPT
```
jaredcasper: This does not seem to be specific to GPT models?

megatron/core/transformer/mxfp8_output_proj.py
```diff
@@ -0,0 +1,144 @@
...
+    if not getattr(config, "fp8", False):
+        return False
+
+    fp8_recipe = getattr(config, "fp8_recipe", None)
+    recipe_value = getattr(fp8_recipe, "value", fp8_recipe)
+    return str(recipe_value).lower() == "mxfp8" or str(fp8_recipe).lower().endswith(".mxfp8")
+
+
+class TELinearCrossEntropyModule(TEColumnParallelLinear):
+    """TE ColumnParallelLinear variant for the LM head under MXFP8.
```
jaredcasper: What makes this particular for the LM head?

megatron/core/transformer/mxfp8_output_proj.py
gdengk: Updated the naming and docstring.
The subclass mostly want to include the LM-head behavior tensor_parallel.ColumnParallelLinear provides for the bf16 LM head today, which TEColumnParallelLinear does not have yet:
- Forces delay_wgrad_compute=False so backward_dw is a no-op to match with bf16 tensor_parallel.ColumnParallelLinear behavior
- Empty get/set_extra_state so the LM-head state dict stays compatible across bf16↔MXFP8 swaps.
- Preserves the ColumnParallelLinear kwarg surface (keep_master_weight_for_test, skip_weight_param_allocation, defer_embedding_wgrad_compute buffers, disable_grad_reduce) so we don't need a special branch if called from TEColumnParallelLinear.

megatron/core/transformer/mxfp8_output_proj.py
gdengk: Replied in the previous thread.

gdengk:
/ok to test [2b624cc](https://github.com/NVIDIA/Megatron-LM/pull/4825/commits/2b624cccffb0a2256c87f74decdb1d4c3b26e42a)

jaredcasper (approved):

ericharper (approved):

PR #4381, opened 2026-04-19
General scaling policy for HP transfer + Depth MuP recipe

plugyawn:
# What does this PR do ?
Addresses part of #4088, introducing a scaling policy for transfer recipes.
This also subsumes and refactors #3058 and #3715 into a `mup` recipe. 
For now, I've kept the --use-mup flag connected (since it was part of the last release), but I'll add in a warning for deprecation? 

Current recipes:
 - `none` (default) being the standard Megatron parameterization
 - `mup` being the refactored existing MuP behavior, preserved exactly.
 - `depth_mup` (draft) with support for Adam/AdamW + dense residual transformers to demonstrate support for other scaling recipes.

### `depth_mup` recipe

We do the following:
- dense residual branch multiplier: `depth_mult^-1`
- hidden matrix-like LR depth multiplier: `depth_mult^0`
- hidden Adam epsilon depth multiplier: `depth_mult^-1`

And also,
- a dense block output-projection init compensation: `depth_mult^+0.5`
  - Megatron already uses layer-count-based residual output-projection initialization. This allows us to avoid accidentally double-counting depth at initialization while making the residual branch scaling explicit in the forward graph.

Chiefly, this PR adds a new `megatron.core.parameterization` package, which handles:
- canonical scaling recipe configs
- final resolved model layer and trainer/optimizer scaling policies
- parameter-role helpers for embedding/output/shared/vector/matrix-like parameters, for support extending to Muon.

For unsupported optimizers like SGD, we raise an error for now. Some math discovery might be necessary here. SGD depth transfer appears to need explicit hidden-weight, hidden-bias laws that would be complex to implement in one go. 

Plots for MuP: Adam, SGD, Muon respectively:
image
image
image
240 iterations, wikitext8, on an A100 80 GB.

Needs some discussion on the YAML changing and the resume loop. 

Some more thought: I think a scaling policy and training recipes could be a great feature moving forward as we can do more with the CLI, considering codex seems quite adept at using CLIs in general! I understand it's a more convenience-based idea, but would be exciting to see first-class support for autonomous _and_ principled pretraining runs.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
[...]

PR #4829, opened 2026-05-16
[1/4] Hyperparameter Transfer: add scaling policy infrastructure

plugyawn:
# What does this PR do ?

PR 1/4 of the planned scaling-recipes stack. 
Design doc: [PR_MuP.pdf](https://github.com/user-attachments/files/27854711/PR_MuP.pdf)
Adds
- internal scaling-policy infrastructure used by the follow-up `mup`, `depth_mup`, and Muon interaction PRs, without introducing the public scaling-recipe surface yet, as planned in #4381. 

This PR:
  - Adds `megatron.core.parameterization` with resolved scaling context, model policy, training policy, and parameter-role helpers.
  - Routes existing model-side MuP-sensitive behavior through policy hooks for embeddings, logits, attention scaling, initialization, and transformer-layer construction.
  - Routes optimizer override construction through the resolved training policy while preserving existing behavior.
  - Preserves scaling metadata through model and FSDP parameter paths.
  - Makes optimizer param-group resume matching stable for older checkpoints by:
    - excluding mutable scheduler values such as `max_lr` / `min_lr` from group identity;
    - default-normalizing missing legacy identifier fields to runtime defaults.
  - Adds internal documentation and regression coverage for the policy seam and optimizer group compatibility.

  This is intentionally the infrastructure slice only. Follow-up PRs layer on:

  1. `feature/scaling-recipe-mup`: canonical width-MuP recipe API.
  2. `feature/depth-mup`: experimental dense-GPT depth-MuP support.
  3. `feature/muon-spectral-mup`: Muon scalar optimizer interaction.

  ## Issue tracking

  Linked issue: Related to #4381.

  ## Testing

  - Added unit coverage for scaling-policy infrastructure behavior.
  - Added optimizer regression coverage for stable group matching and legacy missing identifier defaults.
  - Local smoke: optimizer group identifiers from legacy checkpoint-style groups match runtime defaults.
  - `BASE_REF=main SKIP_DOCS=true ./tools/autoformat.sh`
  - `git diff --check`
  - `git diff --name-only upstream/main...HEAD -- '*.py' | xargs python3 -m py_compile`

  ## Contribution process

  ### Pre-checks


plugyawn:
Hi @janEbert @mkhona-nvidia! Planning to split this into 4 PRs as described. Design doc is attached.

plugyawn:
Tagging @NVIDIA/mcore-oncall; the design discussion for this PR happened in #4381.

plugyawn:
@janEbert should I start adding the next PRs too?

janEbert:
If they're already done, why not? It's nice to see the overarching "story" and wrap-up, too. Thank you for the additional work in making this more manageable!

plugyawn:
Added!

plugyawn:
Made a few changes: some naming changes (ResolvedScalingContext --> ScalingContext, ResolvedScalingPolicy --> ModelScalingPolicy, and a few others), optimizer-group resume tests. 
One question: the old #3058 MuP implementation can be completely deprecated after #4998; should we have a separate PR to drop it, or remove it in #4998?

plugyawn:
Hi @mkhona-nvidia @janEbert! Have added the rest of the PRs, waiting on the reviews!

plugyawn:
@janEbert just a bump!

janEbert:
Hey, what I like about the initial design here is that it has a very explicit goal, hardcoding things where it makes sense to enable parametrization support but nothing else, which keeps things simple.

However, I think that's also a limitation that we'd like to avoid integrating before other design decisions. I'd prefer we start with general user-facing optimizer and initialization configuration APIs, which would allow other modern related configurations as well. I've had the feeling that modern LLM trainings really like to make use of per-layer configuration and enabling more general configuration options would benefit both goals in the long-term, enabling another dimension of user-freedom. The proposed scaling recipes could then be worked on top of those general APIs. (Although we'd prefer these to stay as examples. We would probably not maintain a list of parametrization policies in the core codebase.)

I think it makes sense to first consider these questions, which were also raised in https://github.com/NVIDIA/Megatron-LM/issues/5732, and work from there.

PR #4840, opened 2026-05-18
[Dev][feat] Fix MUON optimizer mxfp8 input padding

Wohox:
# What does this PR do ?

Pad each parameter's start index up to a 64-element boundary in the default `_ParamAndGradBuffer` layout when the **layer-wise distributed optimizer** (Muon and other non-Adam/SGD optimizers) is active, so that `main_grad` slices reach the cuBLASLt-required D-pointer alignment for MXFP8 wgrad. Without this, mid-bucket params land at 16-byte D-pointer alignment and `cublasLtMatmulAlgoGetHeuristic` returns `CUBLAS_STATUS_NOT_SUPPORTED`, killing the training job.

The `main`-branch counterpart is #4889.

## Issue tracking

Linked issue: N/A (bug fix)

## Details

**Symptom.** Training a model with `--use-precision-aware-optimizer` MXFP8 wgrad + layer-wise distributed optimizer (`use_layer_wise_distributed_optimizer=True`) crashes inside cuBLASLt on the first wgrad call for a non-first param in any bucket:

```
RuntimeError: cuBLAS Error: CUBLAS_STATUS_NOT_SUPPORTED when calling cublasLtMatmulAlgoGetHeuristic
```

**Root cause.** Under the layer-wise distributed optimizer, `arguments.py` sets `use_distributed_optimizer=False` and routes through the legacy `allgather_params` sync path, so the `get_model` site in `training.py` does **not** supply a pre-padded `full_param_layout` to DDP. DDP then falls back to `_compute_default_per_buffer_param_layout`, which packs params back-to-back with no padding. The resulting `main_grad` slice for any param whose offset isn't a multiple of 64 elements (≥128 bytes for BF16/MXFP8) has only 16-byte D-pointer alignment, which cuBLASLt 12.8.x rejects for MXFP8 wgrad.

The standard distributed optimizer is unaffected because it precomputes a shard-aligned `full_param_layout` that already aligns per-param starts.

**Fix.**

1. `param_and_grad_buffer.py` — add `pad_param_starts: bool = False` to `_compute_default_per_buffer_param_layout` and `_ParamAndGradBuffer.__init__`. When set, each param's `param_start_index` is rounded up to a 64-element boundary via `pad_param_start` (the same helper the distributed optimizer already uses). Cost: ≤ 126 bytes per param.
2. `distributed_data_parallel.py` — forward the new `pad_param_starts` kwarg from `DistributedDataParallel.__init__` down to `_ParamAndGradBuffer`.
3. `training.py` (`get_model`) — when `DP is DDP and use_layer_wise_distributed_optimizer=True`, pass `pad_param_starts=True` to every wrapped chunk. The flag is ignored when a `full_param_layout` is supplied, so callers that already provide a layout are unaffected.

### Pre-checks

- [x] I have added relevant documentation (inline docstring on new args)

### Test plan

- Reproduce: launch any Muon + MXFP8 wgrad training (e.g. `configs/benchmarking/recipes/deepseek_v4_flash_proxy/gb200/mxfp8_*muon*.yaml`) on `dev` without this patch → `CUBLAS_STATUS_NOT_SUPPORTED` on iter 1.
- Verify: same launch with this patch → training proceeds; per-iter throughput unchanged vs. Adam/standard distributed optimizer baseline.
- Regression: `--use-distributed-optimizer` (standard) path is untouched (still supplies its own padded `full_param_layout`); BF16 layer-wise path also runs cleanly (padding is benign).

### Code review

<details>

</details>


FDecaYed (approved):
LGTM

Wohox:
/ok to test [bff38ab](https://github.com/NVIDIA/Megatron-LM/pull/4840/commits/bff38ab958cc92ab9483a3b4c2887298cfd6259b)

Wohox:
Closed since the fix is already included in #4771

PR #4841, opened 2026-05-18
[dev] feat(attention): add use_head_wise_attn_gate for Step-3.5-Flash

shifangx:
# What does this PR do ?

  - use_head_wise_attn_gate: bool -- enable a per-head scalar output gate
    (a.k.a. g_proj). The gate is fused into linear_qkv as the trailing
    num_attention_heads rows; in the forward pass each rank slices off its
    num_attention_heads_per_partition gate scalars, applies sigmoid, and
    scales each attention head's output before linear_proj.
  - Distinct from attention_output_gate, which fuses a full head_dim gate
    into linear_qkv. The per-head gate adds only num_attention_heads extra
    output rows on linear_qkv (vs. num_attention_heads * head_dim), so the
    parameter and FLOP overhead is negligible.
  - Under tensor parallel, this requires that both num_query_groups and
    num_attention_heads be divisible by the TP world size so the QKV and
    gate sub-ranges stay aligned per rank.

  The new feature defaults to False and has no effect on existing models.

  for main branch: https://github.com/NVIDIA/Megatron-LM/pull/4848

  Related: per-layer RoPE for Step-3.5-Flash is tracked separately in
  [#4473](https://github.com/NVIDIA/Megatron-LM/pull/4473).

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>

### DIFF
--- a/megatron/core/transformer/attention.py
+++ b/megatron/core/transformer/attention.py
@@ -1158,9 +1158,7 @@ class Attention(MegatronModule, ABC):
         )
         output_gate = self.config.attention_output_gate
         # head_wise_attn_gate is self-attention only (gate rows live in linear_qkv).
-        head_wise_gate_enabled = (
-            self.config.head_wise_attn_gate and self.attention_type != "cross"
-        )
+        head_wise_gate_enabled = self.config.head_wise_attn_gate and self.attention_type != "cross"
         # Check if fused_single_qkv_rope is requested but either unavailable or not
         # supported for the current use case.
         if self.attention_type != "cross":
@@ -1170,9 +1168,7 @@ class Attention(MegatronModule, ABC):
         if output_gate:
             assert split_qkv, "output_gate is not supported for unsplit mixed_qkv tensor."
         if head_wise_gate_enabled:
-            assert split_qkv, (
-                "head_wise_attn_gate is not supported for unsplit mixed_qkv tensor."
-            )
+            assert split_qkv, "head_wise_attn_gate is not supported for unsplit mixed_qkv tensor."
 
         qkv_linear_manager = off_interface(self.offload_qkv_linear, hidden_states, "qkv_linear")
         with qkv_linear_manager as hidden_states:
@@ -1401,8 +1397,8 @@ class Attention(MegatronModule, ABC):
             nvtx_range_push(suffix="head_wise_attn_gate")
             gate_states = head_wise_gate.view(*head_wise_gate.shape[:2], -1, 1)
             core_attn_out = core_attn_out.view(*gate_states.shape[:3], -1)
-            core_attn_out = (
-                core_attn_out * torch.sigmoid(gate_states.float()).to(core_attn_out.dtype)
+            core_attn_out = core_attn_out * torch.sigmoid(gate_states.float()).to(
+                core_attn_out.dtype
             )
             core_attn_out = core_attn_out.view(*gate_states.shape[:2], -1)
             nvtx_range_pop(suffix="head_wise_attn_gate")
@@ -1908,10 +1904,7 @@ class SelfAttention(Attention):
         return weight_updated
 
     def sharded_state_dict(
-        self,
-        prefix: str = "",
-        sharded_offsets: tuple = (),
-        metadata: Optional[dict] = None,
+        self, prefix: str = "", sharded_offsets: tuple = (), metadata: Optional[dict] = None
     ) -> ShardedStateDict:
         """Wrap linear_qkv.{weight,bias} with a factory that reshards QKV and
         gate sub-blocks independently, analogous to apply_swiglu_sharded_factory."""
@@ -1930,10 +1923,7 @@ class SelfAttention(Attention):
 
 # pylint: disable=missing-function-docstring
 def apply_head_wise_attn_gate_sharded_factory(
-    original_sh_ten,
-    sharded_offsets,
-    qkv_size: int,
-    gate_size: int,
+    original_sh_ten, sharded_offsets, qkv_size: int, gate_size: int
 ):
     """Split linear_qkv.{weight,bias} into QKV and gate sub-tensors with
     distinct keys so each is resharded independently along axis 0 across TP.
@@ -1953,17 +1943,13 @@ def apply_head_wise_attn_gate_sharded_factory(
         f"({gate_size}) divisible by TP world_size ({axis_frag})."
     )
     assert local_axis_size == qkv_local_size + gate_local_size
-    rank_offset = (
-        original_sh_ten.global_offset[shard_axis + prepend_axis_num] // local_axis_size
-    )
+    rank_offset = original_sh_ten.global_offset[shard_axis + prepend_axis_num] // local_axis_size
 
     @torch.no_grad()
     def sh_ten_build_fn(
         key: str, t: torch.Tensor, replica_id: ReplicaId, flattened_range: Optional[slice]
     ):
-        tensor_qkv, tensor_gate = torch.split(
-            t, [qkv_local_size, gate_local_size], dim=shard_axis
-        )
+        tensor_qkv, tensor_gate = torch.split(t, [qkv_local_size, gate_local_size], dim=shard_axis)
         offset_qkv = (shard_axis + prepend_axis_num, rank_offset, axis_frag)
         offset_gate = (shard_axis + prepend_axis_num, rank_offset, axis_frag)
         return [
--- a/megatron/core/transformer/attention.py
+++ b/megatron/core/transformer/attention.py
@@ -1516,7 +1516,40 @@ class SelfAttention(Attention):
         if self.config.use_head_wise_attn_gate:
             # Fuse per-head scalar gate (Step-3.5-Flash g_proj) into linear_qkv.
             # Gate rows are appended after the QKV block, so the trailing
-            # num_attention_heads rows of linear_qkv.weight are the gate weights.
+            # num_attention_heads rows of linear_qkv.weight are the gate
+            # weights. Two layout invariants must hold for the forward-time
+            # peel-off (in get_query_key_value_tensors) to pick up actual gate
+            # rows on every rank; enforce both at init so the failure mode is
+            # a clear constructor error rather than a silent "wrong rows =
+            # gate" misalignment during training.
+            tp_world_size = get_pg_size(self.pg_collection.tp)
+            # (a) num_query_groups >= world_size: the head-wise gate slice
+            #     runs BEFORE the num_query_groups < world_size AllGather +
+            #     reslice fallback in get_query_key_value_tensors. Under that
+            #     fallback, num_attention_heads_per_partition is inflated to
+            #     num_attention_heads / num_query_groups (the pre-AG count),
+            #     which is larger than the actual per-rank gate scalar count
+            #     (num_attention_heads / world_size). The slice would then
+            #     take V/K rows as gate. Disallow the combination outright.
+            assert self.config.num_query_groups >= tp_world_size, (
+                f"use_head_wise_attn_gate requires num_query_groups "
+                f"({self.config.num_query_groups}) >= tensor-model-parallel "
+                f"world_size ({tp_world_size}). The gate peel-off happens "
+                f"before the num_query_groups < world_size AllGather + "
+                f"reslice path; under that fallback the gate slice would "
+                f"over-take and pick up V/K rows instead of gate rows."
+            )
+            # (b) num_attention_heads % world_size == 0: each rank must
+            #     carry an integer number of gate scalars. (Redundant with
+            #     parent's divide(num_attention_heads, world_size) under
+            #     invariant (a), but kept for an explicit error message.)
+            assert self.config.num_attention_heads % tp_world_size == 0, (
+                f"use_head_wise_attn_gate requires num_attention_heads "
+                f"({self.config.num_attention_heads}) to be divisible by the "
+                f"tensor-model-parallel world_size ({tp_world_size}); the "
+                f"trailing-rows-as-gate layout cannot place an integer number "
+                f"of gate rows on every rank otherwise."
+            )
             self.linear_qkv_out_dim += self.config.num_attention_heads
         self.linear_qkv = submodules.linear_qkv(
             self.config.hidden_size,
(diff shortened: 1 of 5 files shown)


megatron/core/transformer/attention.py
```diff
@@ -1445,6 +1471,11 @@ def __init__(
         self.linear_qkv_out_dim = self.query_projection_size + 2 * self.kv_projection_size
         if self.config.attention_output_gate:
             self.linear_qkv_out_dim += self.config.kv_channels * self.config.num_attention_heads
+        if self.config.use_head_wise_attn_gate:
+            # Fuse per-head scalar gate (Step-3.5-Flash g_proj) into linear_qkv.
+            # Gate rows are appended after the QKV block, so the trailing
+            # num_attention_heads rows of linear_qkv.weight are the gate weights.
+            self.linear_qkv_out_dim += self.config.num_attention_heads
```
Victarry: **[CRITICAL Checkpoint]** Fused gate rows break dist-checkpoint TP resharding silently.

With `use_head_wise_attn_gate=True`, `linear_qkv.weight` is laid out globally as `[QKV_block, Gate_block]` (gate rows appended at the tail). `SelfAttention` does not override `sharded_state_dict`, so it falls through to `ColumnParallelLinear` / `TELayerNormColumnParallelLinear`, which declares `{"weight": 0}` — no information about the QKV/Gate sub-blocks.

Under TP resharding (save TP=N → load TP=M, M≠N), the dist-ckpt framework re-splits the global `(QKV_dim + H, hidden_size)` tensor uniformly along axis 0. Whenever `(QKV_dim + H) / load_ws < QKV_dim` (the common case — e.g. H=8, load_ws=4, QKV=1536, gate=8, per-rank=386 ≪ 1536), the trailing `num_attention_heads_per_partition` rows on new ranks `0..M-2` lie inside the QKV block, not the gate block. The forward then peels those V/K rows, runs `sigmoid`, and multiplies them per-head into `core_attn_out`. No shape error — the model silently computes a different function.

This is the same class of bug that SwiGLU avoids via `apply_swiglu_sharded_factory` (`megatron/core/transformer/mlp.py:354-361`). For contrast, `attention_output_gate` is safe under resharding because its gate rows are interleaved per-group, composing naturally with the existing QKV TP AllGather+reslice path in `get_query_key_value_tensors`.

**Suggestion:** Either
1. Add a `SelfAttention.sharded_state_dict` override that uses a `ShardedTensorFactory` to chunk `linear_qkv.weight` into `[QKV_part, gate_part]` on save and cat on load (each independently axis-0 sharded), analogous to `apply_swiglu_sharded_factory`.
2. Or change the layout to per-group inline (like `attention_output_gate`), avoiding the need for a factory.

Until fixed, document "not supported with TP resharding" in the config docstring, and add a save TP=1 → load TP=2 logits-equality test under `tests/unit_tests/dist_checkpointing/`.

megatron/core/transformer/attention.py
```diff
@@ -1594,6 +1625,21 @@ def get_query_key_value_tensors(
...
+        # uniformly, so each rank gets num_attention_heads_per_partition gate
+        # scalars at the tail of mixed_qkv's last dim. Requires that both
+        # num_query_groups and num_attention_heads be divisible by tp world_size
+        # so the QKV and gate sub-ranges stay aligned per rank.
+        if self.config.use_head_wise_attn_gate:
+            gate_size = self.num_attention_heads_per_partition
+            mixed_qkv, gate_states = torch.split(
+                mixed_qkv, [mixed_qkv.size(-1) - gate_size, gate_size], dim=-1
+            )
+            self._head_wise_gate_states = gate_states
```
Victarry: **[IMPORTANT TP Correctness]** Trailing-rows-as-gate assumption is undocumented and untested under TP>1.

The peel-off here assumes each rank's local `mixed_qkv` has its `num_attention_heads_per_partition` gate scalars at the tail. This relies on three implicit preconditions:

1. `ColumnParallelLinear` splits the weight along axis 0 contiguously and uniformly (no row interleave / stride reordering).
2. `num_attention_heads % world_size == 0`.
3. The gate block is globally appended after the entire QKV block (`[QKV..., Gate...]`).

Precondition (1) is a silent dependency on the linear backend. If a future change ever adds row interleave to `linear_qkv`, the slice picks V rows instead of gate rows with no shape error.

The unit tests run `Utils.initialize_model_parallel(1, 1)` exclusively, so the TP>1 slicing alignment is not actually verified by tests.

Additionally, the gate split here happens *before* the `num_query_groups < world_size` AllGather+reslice branch below. The gate does not participate in that fallback path, which is what saves `attention_output_gate` under resharding and atypical group/ws ratios.

**Suggestions:**
1. Add a TP=2 unit test where gate weight rows are set to a recognizable pattern (e.g. `-inf` for some heads, `+inf` for others) and verify the corresponding output heads are zeroed / passed through on each rank — this back-validates the layout assumption rather than relying on prose.
2. Add an init-time assertion in `SelfAttention.__init__` when `use_head_wise_attn_gate=True`: `assert self.config.num_attention_heads % world_size == 0`.
3. Document the three preconditions in the comment block above this `if` statement, making the dependency on `ColumnParallelLinear`'s contiguous row split explicit.

megatron/core/transformer/attention.py
```diff
@@ -395,6 +395,13 @@ def __init__(
             rotary_base = self.config.rotary_base_per_layer[self.layer_number - 1]
             self._build_per_layer_rotary_pos_emb(rotary_base)
 
+        # Per-head scalar output gate (Step-3.5-Flash g_proj) is fused into
+        # linear_qkv when use_head_wise_attn_gate is True; gate weights live in
+        # the trailing num_attention_heads rows of linear_qkv.weight. The gate
+        # tensor is sliced out in SelfAttention.get_query_key_value_tensors and
+        # consumed in Attention.forward.
+        self._head_wise_gate_states = None
```
Victarry: **[IMPORTANT Robustness]** Passing gate states via `self._head_wise_gate_states` adds a stateful side-channel between `get_query_key_value_tensors` and `forward`. Concrete hazards:

1. **Cross-attention path**: `CrossAttention.get_query_key_value_tensors` never sets this attribute, but `Attention.forward` still reads it. If `use_head_wise_attn_gate=True` is set on a model that has cross-attention layers, the `assert gate_states is not None` at line 1388 trips with a message that does not explain the real cause.
2. **Concurrent forward passes**: under pipeline-parallel overlap schedules, two microbatches' forwards can in principle be in flight on the same module instance; an instance attribute will be overwritten before the first call consumes it. The existing `attention_output_gate` passes `gate` via the return tuple precisely to avoid this.
3. **Recompute / instance-attr lifetime**: with `set_for_recompute_input_layernorm` enabled (FP8 selective recompute), if the recompute path enters `get_query_key_value_tensors` again, gate_states get re-stashed but may never be consumed, leaving a `[sq, b, np_per_rank]` bf16 tensor referenced by the module until the next forward.

The pre-existing `attention_output_gate` returns `(query, key, value, gate)` from `get_query_key_value_tensors` and threads `gate` through `forward` as a local. This PR should follow the same pattern.

**Suggestion:** drop the instance attribute and thread `head_wise_gate` through the return tuple of `get_query_key_value_tensors` (e.g. as an optional 5th element when `use_head_wise_attn_gate=True`), then unpack at the call site in `Attention.forward` (~line 1161-1176).

If you keep the instance-attribute design as-is, at minimum:
- Add an init-time guard in `Attention.__init__`: `assert self.attention_type == "self" or not self.config.use_head_wise_attn_gate, "use_head_wise_attn_gate only supports self-attention"`.
- Rewrite the assert message at line 1388 to point at the cross-attention / config-mismatch cause.

megatron/core/transformer/transformer_config.py
```diff
@@ -263,6 +263,13 @@ class TransformerConfig(ModelParallelConfig):
     SelfAttention layer creates its own RotaryEmbedding with the corresponding base;
     the shared model-level rotary_pos_emb is not created."""
 
+    use_head_wise_attn_gate: bool = False
+    """Apply a per-head scalar output gate (e.g., Step-3.5-Flash g_proj).
+    The gate weights are fused into linear_qkv as the trailing
+    num_attention_heads rows; the sigmoid of those scalars gates each
+    attention head independently. Distinct from attention_output_gate,
+    which fuses a full head_dim gate into linear_qkv."""
+
```
Victarry: **[IMPORTANT Validation]** `use_head_wise_attn_gate` lacks `__post_init__` validation.

Three missing checks:

1. **Mutual exclusion / composition with `attention_output_gate`**: both flags grow `linear_qkv_out_dim` and rely on different slicing logic; the PR does not document or test whether they compose. If they do not, it should be an explicit error.
2. **TP divisibility**: the PR description claims `num_attention_heads % world_size == 0` is required, but this is not enforced anywhere.
3. **Self-attention only**: applying the gate to cross-attention trips an obscure assert in `forward`; the config should refuse the combination up front.

**Suggestion:** add to `TransformerConfig.__post_init__`:

```python
if self.use_head_wise_attn_gate:
    assert not self.attention_output_gate, (
        "use_head_wise_attn_gate and attention_output_gate cannot both be enabled "
        "(both append rows to linear_qkv with incompatible layouts)"
    )
    assert self.num_attention_heads % self.tensor_model_parallel_size == 0, (
        "use_head_wise_attn_gate requires num_attention_heads divisible by "
        "tensor_model_parallel_size"
    )
```

If the two gates are meant to compose, please document the combined semantics in the docstring and add a test for the joint path.

megatron/core/transformer/attention.py
Victarry, line 1491: **[SUGGESTION Init]** Gate rows share `init_method` with QKV, so `sigmoid(gate·x)` starts near 0.5 at init — every head's output is halved from step 0.

Most "output gate" designs initialize the gate to make `sigmoid(gate) ≈ 1` at init (approximate identity), avoiding extra optimization burden during warmup. If the Step-3.5-Flash reference specifies a particular init, please follow it.

Additionally, under FP8 the `linear_qkv` weight uses a single tensor-level scale for both the QKV and gate blocks. If gate weights are initialized to a much smaller range than QKV (e.g. zero-init), they may underflow in FP8 quantization. So any custom init should keep gate-weight magnitudes comparable to QKV, or expose a config knob.

**Suggestion:** initialize the gate rows separately after `self.linear_qkv` is constructed, e.g.:

```python
if self.config.use_head_wise_attn_gate:
    with torch.no_grad():
        # Bias gate-rows so sigmoid is near 1 at init (approximate identity)
        self.linear_qkv.weight[-self.config.num_attention_heads:].mul_(0.1)
        if self.linear_qkv.bias is not None:
            self.linear_qkv.bias[-self.config.num_attention_heads:].fill_(2.0)  # sigmoid(2) ~ 0.88
```

At minimum, document the current init behavior (gate ≈ 0.5 at start) in the config docstring so users aren't surprised.

megatron/core/transformer/transformer_config.py
```diff
@@ -263,6 +263,13 @@ class TransformerConfig(ModelParallelConfig):
     SelfAttention layer creates its own RotaryEmbedding with the corresponding base;
     the shared model-level rotary_pos_emb is not created."""
 
+    use_head_wise_attn_gate: bool = False
```
Victarry: **[SUGGESTION Naming]** Naming inconsistency with the sibling flag.

The existing flag is `attention_output_gate: bool` (no `use_` prefix). The new flag is `use_head_wise_attn_gate: bool` (with `use_` prefix). Users writing YAML configs will not remember which one has the prefix.

**Suggestion:** rename to `head_wise_attn_gate` to match the existing convention, or — if you want to standardize on `use_` going forward — file a follow-up to rename both consistently with deprecation aliases. Either way, add a cross-reference in the docstring pointing at `attention_output_gate` and explaining the per-head-scalar vs full-head-dim difference (the PR description already has this prose; surface it in the config docstring).

tests/unit_tests/transformer/test_head_wise_attn_gate.py
```diff
@@ -0,0 +1,219 @@
...
+    return SelfAttention(config, submodules, layer_number=1)
+
+
+@pytest.mark.parametrize("transformer_impl", ["transformer_engine", "native"])
+class TestHeadWiseAttnGateInit:
+    """Verify linear_qkv is sized to include the fused gate rows iff enabled."""
+
+    @pytest.fixture(autouse=True)
+    def setup_teardown(self, transformer_impl):
+        Utils.initialize_model_parallel(1, 1)
```
Victarry, line 134: **[SUGGESTION Tests]** Test coverage gaps for a new `megatron/core` feature.

Current tests run `Utils.initialize_model_parallel(1, 1)` exclusively — TP=1, bf16, no bias, no packed_seq, no CP. `test_zero_gate_halves_output` indirectly proves the path is wired, but several risk-bearing paths are unverified:

- **TP>1 gate row alignment** — the core unverified concern (see CRITICAL Checkpoint comment).
- **CP** — `gate_states` is sliced before any sequence-parallel / CP split; verify shape compatibility with `core_attn_out` at `[sq/cp, b, np*hn]` per CP rank.
- **Backward** — gradient flow to gate rows (should work but no assertion).
- **`add_qkv_bias=True`** — gate rows get bias, affects sigmoid input.
- **`packed_seq_params.qkv_format == 'thd'`** — `core_attn_out` is reshaped to `[t, 1, h]`; the `.view(*gate_states.shape[:3], -1)` chain interacts with the THD reshape.
- **fp8** — interaction with `set_save_original_input(self.linear_qkv)` and the `linear_qkv` weight scale shared between QKV and gate blocks.
- **dist-checkpoint save/load TP resharding** (see CRITICAL Checkpoint comment).

**Suggestion:** at minimum add:
1. A TP=2 test parametrized over `num_query_groups in (2, 8)` comparing forward output against a TP=1 reference with identical weights replicated to TP=2 layout.
2. One `add_qkv_bias=True` case.
3. A dist-ckpt resharding test (save TP=1 → load TP=2, compare logits) under `tests/unit_tests/dist_checkpointing/`. There are existing patterns there you can adapt.

CP / fp8 / backward can be filed as follow-ups but should at least be acknowledged in the PR description.

megatron/core/transformer/attention.py
Victarry, line 1491: **[IMPORTANT FP8 Alignment]** `linear_qkv_out_dim` increment may break FP8 GEMM alignment.

The increment here is `num_attention_heads` (e.g. 8, 12, 28, ...). FP8 GEMM in Transformer Engine requires the per-partition output dim to be 16-aligned (FP4: 32-aligned). The existing `attention_output_gate` adds `num_attention_heads * kv_channels` rows — since `kv_channels` is typically 64 or 128, alignment is automatic. This PR's much smaller increment can break it.

Concrete failure cases:
- H=8, G=8 (MHA), D=128, TP=1 → `linear_qkv_out_dim = 3*8*128 + 8 = 3080`; `3080 % 16 = 8` (FAIL)
- H=8, G=8, D=128, TP=2 → per-rank `1540`; `1540 % 16 = 4` (FAIL)
- H=16, G=2 (GQA), D=128, TP=1 → `2576`; `2576 % 16 = 0` (OK)
- H=32, G=8, D=128, TP=2 → per-rank `3104`; `3104 % 16 = 0` (OK)

TE will either raise an alignment error in the FP8 GEMM, or (worse) silently pad the weight and produce shifted outputs. This is a new FP8-only precondition that does not exist for `attention_output_gate`.

**Suggestion:** either
1. Enforce alignment in `TransformerConfig.__post_init__` when FP8/FP4 is on:

   ```python
   if self.use_head_wise_attn_gate and (self.fp8 or self.fp4):
       align = 32 if self.fp4 else 16
       per_partition = self.linear_qkv_out_dim // self.tensor_model_parallel_size
       assert per_partition % align == 0, (
           f"use_head_wise_attn_gate under fp8/fp4 requires linear_qkv_out_dim_per_partition "
           f"({per_partition}) to be aligned to {align}; num_attention_heads={self.num_attention_heads}"
       )
   ```

2. Or pad the gate-rows block to the alignment boundary (waste a few weight rows but avoid user-visible constraint), and ignore the padded rows in the forward slice.

shifangx: Thanks for advice.

shifangx: Done

megatron/core/transformer/attention.py
shifangx: done

megatron/core/transformer/attention.py
shifangx: done

megatron/core/transformer/attention.py
shifangx: fixed

megatron/core/transformer/transformer_config.py
shifangx: fixed

megatron/core/transformer/transformer_config.py
shifangx: fixed

tests/unit_tests/transformer/test_head_wise_attn_gate.py
shifangx: fixed

megatron/core/transformer/attention.py
shifangx: fixed

shifangx:
/ok to test [462a280](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/462a280a90c363c134da98c1155abb76b6f88f2b)

shifangx:
/ok to test [a93633a](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/a93633aa4c67338b9c3d7d6711766f1cc5f7ca02)

shifangx:
/ok to test [e141466](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/e141466df2137f3427b577ab9dd00bd608d58826)

shifangx:
/ok to test [9262719](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/92627197034c134bb6061020ee81a27ee82126e4)

shifangx:
/ok to test [bb1287c](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/bb1287cf15f2b19454eca43704879b920ed4132b)

shifangx:
/ok to test [87056f4](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/87056f4a731e08eb19e7089016bac088c1d40d76)

shifangx:
/ok to test [549478c](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/549478cdbe3d24a926e13df08fb54ef275d65484)

shifangx:
/ok to test [dd94d20](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/dd94d20148ce4f1e06f818c9fc0b96ff19e52e8a)

shifangx:
/ok to test [22255f2](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/22255f28989af4a884631f650af464d66e2671d5)

shifangx:
/ok to test [b89aeda](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/b89aeda0da877c509d0f2a97623aca7db25bf220)

shifangx:
/ok to test [2bfcbf4](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/2bfcbf41d8bf85d0600357ab1aa12a70f83959ac)

Victarry (approved):

shifangx:
/ok to test [45eba11](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/45eba116780e1b9b95d9de2482ebf0b22316c474)

shifangx:
/ok to test [05f336c](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/05f336ce6cd2460ade10db1a232b6baa4428ff48)

shifangx:
/ok to test [f01a5d4](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/f01a5d4acc22cb63d137680232459e48c7eb33fd)

shifangx:
/ok test [522ae61](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/522ae618094bca16ee3384e94092416c5fa60795)

shifangx:
/ok to test [684d7c5](https://github.com/NVIDIA/Megatron-LM/pull/4841/commits/684d7c5027760d2cc9d6d2533029316862d210d5)

PR #4846, opened 2026-05-18
Remove DeepEP hardware limit check

janEbert:
[As discussed](https://github.com/NVIDIA/Megatron-LM/pull/4846#discussion_r3278427726), the HybridEP IB hardware limit check is removed but the variable rename kept for better legibility.

CC @Autumn1998
Ref #3999

### DIFF
--- a/megatron/core/transformer/moe/fused_a2a.py
+++ b/megatron/core/transformer/moe/fused_a2a.py
@@ -280,7 +280,7 @@ _hybrid_ep_buffer = None
 def init_hybrid_ep_buffer(
     group: torch.distributed.ProcessGroup,
     hidden_dim: int,
-    seq_len: int,
+    num_tokens: int,
     num_local_experts: int,
     num_sms_dispatch_api: Optional[int] = None,
     num_sms_combine_api: Optional[int] = None,
@@ -302,8 +302,8 @@ def init_hybrid_ep_buffer(
             Process group for HybridEP all-to-all communication.
         hidden_dim (int):
             Hidden dimension of the input tensor.
-        seq_len (int):
-            Maximum sequence length of the input tensor.
+        num_tokens (int):
+            Maximum token count of the input tensor.
         num_local_experts (int):
             Number of local experts.
         num_sms_dispatch_api (Optional[int]):
@@ -335,7 +335,7 @@ def init_hybrid_ep_buffer(
     _hybrid_ep_buffer = HybridEPBuffer(
         group=group,
         hidden_dim=hidden_dim,
-        max_num_of_tokens_per_rank=seq_len,
+        max_num_of_tokens_per_rank=num_tokens,
         num_local_experts=num_local_experts,
         use_fp8=fp8_dispatch,
         **kwargs,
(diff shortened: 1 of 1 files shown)


megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -451,13 +451,12 @@ def forward(
...
 
         num_tokens, hidden_dim = x.shape[-2:]
-        _validate_hybrid_ep_ib_tx_depth(num_tokens, group)
         if _hybrid_ep_buffer is None:
             fp8_dispatch = False  # Currently, we do not support fp8 dispatch
             init_hybrid_ep_buffer(
                 group,
                 hidden_dim,
-                num_tokens,
+                _hybrid_ep_initial_buffer_tokens(num_tokens),
```
Autumn1998:
```
def _hybrid_ep_initial_buffer_tokens(num_tokens: int) -> int:
    if _hybrid_ep_ib_tx_depth(num_tokens) <= _HYBRID_EP_IB_QP_MAX_DEPTH:
        return num_tokens
    return _hybrid_ep_max_supported_tokens()
```

We should not enter this test **_unconditionally,_** because `_HYBRID_EP_IB_QP_MAX_DEPTH` should not be used for checking in a pure NVL domain. We need to use `_hybrid_ep_uses_nic` to determine this first, before applying the `_HYBRID_EP_IB_QP_MAX_DEPTH` check. Furthermore, theoretically these checks already exist in hep — why are we adding duplicate monitoring in mcore? This makes the code look unnecessarily complex.

Autumn1998: If we feel the error message isn't clear enough, I can add a Python-level assert in hep, which would spare us from having to modify fuse a2a.

Autumn1998: https://github.com/deepseek-ai/DeepEP/pull/638/commits
PR in hybrid-ep

janEbert: The buffer will grow anyway if it gets initialized with too few tokens. I don't see how it causes an issue when in a pure NVL domain, except that the very first usage would have one additional buffer growth.

What I like about having the error message in MLM is that it gives the user concrete advice on how to try and fix/work around the error. While your error message in DeepEP is more clear than the previous one, it doesn't offer a user of MLM additional advice.

But if we conclude that the user-facing message in MLM doesn't offer enough value for the additional complexity it requires in the code (where I completely agree with you – it hurts the file's cleanliness and has already shown to be complex to handle correctly), then we can close this and revert the previous commits. I would like to have a discussion on this, though, with the points I made above. :)

Autumn1998: In general, the buffer expansion in Hybrid EP is quite costly, it frees the original buffer and reallocates everything from scratch. Although it only happens on the first step, in my mind it feels like this would introduce two expensive initialization operations, which doesn't make sense to me. The code complexity is also a concern; it would make the code harder for users to read.

But as you said, at the HEP layer we can only surface a message like "too many tokens", whereas at the mcore layer we can advise the user to enable a larger TP — that is a valid point. If you feel this is important, we can keep it. However, I lean towards keeping the code clean, since I assume users will have a certain level of background knowledge in LLMs.

janEbert: Okay, I agree with that; in order to hopefully prevent [further issues](https://github.com/NVIDIA/Megatron-LM/issues/3999) being raised by using a better variable name, I'll convert this PR to:
- reverting the previous checks and instead
- [keeping the variable rename from `seq_len` to `num_tokens`](https://github.com/NVIDIA/Megatron-LM/pull/4719/files#diff-44f7813bc805db685b73a89b2be32f651363b013d488ea71941ba61d512ae280R338-R391) for better readability.

Thank you for your thoughts!

yaox12 (approved):

yashaswikarnati (approved):

ericharper (approved):

ko3n1g (approved):

deepakn94 (approved):

PR #4855, opened 2026-05-18
Fix incorrect inference metadata tensor dtypes

santhnm2:
# What does this PR do ?
Ensures `token_to_block_idx` uses `int64` consistently. Otherwise, we were seeing numerical overflow when the number of blocks exceeded 2^16. Also makes `batch_indices_decode` int64 consistently.

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

<details>

</details>


megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -1142,7 +1143,7 @@ def initialize_all_tensors(self) -> None:
         if self.is_hybrid_model:
             self._cpu_mamba_batch_indices_decode = self._cpu_bookkeeping_buf[
```
lmcafee-nvidia, line 1170: reviewing this with my AI, it thinks this is still buggy ...

> `batch_indices_decode` is now an int64 view, but this slot still starts after several int32/float32 sections, so `_off` is not guaranteed to be 8-byte aligned. For example, with max_requests=1, max_tokens=256, max_kv_blocks=64, the Mamba section starts at byte 9524, so `_off % 8 == 4`. Reinterpreting a uint8 slice as torch.int64 requires 8-byte alignment, so some hybrid configurations can fail during context construction. Could we either move this field into the int64 section or insert explicit padding before this view, and mirror the same padding in ContextGPUView?

megatron/core/inference/contexts/gpu_view.py
```diff
@@ -119,8 +119,8 @@ def __init__(
         off += tok_int64_bytes
         self.token_to_pos_ids = self._buf[off : off + tok_int64_bytes].view(torch.long)
         off += tok_int64_bytes
-        self.token_to_block_idx = self._buf[off : off + tok_int32_bytes].view(torch.int32)
-        off += tok_int32_bytes
+        self.token_to_block_idx = self._buf[off : off + tok_int64_bytes].view(torch.int64)
```
lmcafee-nvidia, line 131: This changes `token_to_block_idx` to int64, but the existing `tests/unit_tests/inference/contexts/test_gpu_view.py` expectations were not updated: `TOKEN_VIEWS_INT32` still includes `token_to_block_idx`, and the Mamba view loop still asserts every Mamba tensor is int32. This should fail `TestContextGPUView.test_layout_with_and_without_mamba` on CUDA CI. Please update that test to expect `token_to_block_idx` and `mamba_batch_indices_decode` as int64.

tests/unit_tests/inference/contexts/test_fused_kv_append_kernel.py
```diff
@@ -0,0 +1,115 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+import pytest
+import torch
+
+try:
```
lmcafee-nvidia: This try/import does not actually detect whether Triton is available. `fused_kv_append_kernel.py` catches ImportError internally and still exports `triton_append_key_value_cache` backed by mocks, so this test sets HAVE_TRITON=True even when Triton is missing. On a CUDA worker without Triton, the skip marker would not skip. Could we import/check the module-level `HAVE_TRITON` from `fused_kv_append_kernel` instead?

tdene:
@santhnm2 @lmcafee-nvidia why not just keep all the metadata in `int64` to prevent any such issues from popping up? It's not going to be a large memory cost.

tests/unit_tests/inference/contexts/test_fused_kv_append_kernel.py
santhnm2: Done

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Addressed

megatron/core/inference/contexts/gpu_view.py
santhnm2: Updated the test

santhnm2:
> @santhnm2 @lmcafee-nvidia why not just keep all the metadata in `int64` to prevent any such issues from popping up? It's not going to be a large memory cost.

Yea I think we need to clean this up, will defer to a later PR though because it may end up breaking other things unintentionally

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -358,15 +358,15 @@ def create_cuda_graphs(self, reset_context: bool = True):
         # decoder graphs within the same loop rather than in a separate pass.
         unwrapped = unwrap_model(controller.inference_wrapped_model.model)
         mtp_warmup_enabled = (
-            controller.num_mtp_heads > 0
+            controller.num_mtp_depths > 0
```
lmcafee-nvidia: `num_mtp_depths` is only assigned in `_init_mtp_sampling_tensors()` after the `if not self.num_speculative_tokens: ... return` branch. For non-speculative runs, the controller can therefore be constructed without this attribute. This CUDA graph warmup path evaluates `controller.num_mtp_depths > 0` before checking `controller.num_speculative_tokens`, so local CUDA graph capture with `num_speculative_tokens == 0` can raise `AttributeError`. Could we initialize `self.num_mtp_depths = 0` in `TextGenerationController.__init__` or before the early return in `_init_mtp_sampling_tensors()`, or reorder this condition to check `num_speculative_tokens` first?

lmcafee-nvidia (approved):

kvareddy (approved):

shanmugamr1992 (approved):

PR #4885, opened 2026-05-20
[dev] Add experimental Megatron Lite as agentic exploration

ISEEKYAN:
# What does this PR do ?

Adds an experimental `megatron.lite` package under `experimental/lite` with a lite runtime, common primitives, and lite-only Qwen3/Qwen3.5 model implementations.

This is an experimental package layout that is source-tree importable with:

```bash
export PYTHONPATH=$PWD/experimental/lite:$PYTHONPATH
```

and user-facing imports use:

```python
from megatron.lite.runtime import LiteConfig, RuntimeConfig, create_runtime
```

Included:
- Lite runtime API and backend under `megatron.lite.runtime`
- Common lite primitives under `megatron.lite.primitive`
- Lite-only Qwen3 MoE and Qwen3.5 MoE model implementations
- HF safetensors load/export helpers
- Megatron-Core optimizer wrapping
- README and initial design docs under `experimental/lite/docs`

Intentionally excluded from this PR:
- Hybrid model implementations
- Bridge model/runtime implementations
- FSDP2 optimizer primitives
- Benchmark scripts
- Repository-level packaging changes

## Issue tracking

Linked issue: TODO

### Pre-checks

Validation run locally:

```bash
python -B -m compileall -q experimental/lite/megatron/lite
PYTHONPATH=$PWD/experimental/lite:$PWD python -B -c "import megatron.lite as ml; from megatron.lite.runtime import RuntimeConfig, LiteConfig; from megatron.lite.model.registry import resolve_model_type_from_hf; print(ml.__name__, RuntimeConfig().backend, LiteConfig.__name__, resolve_model_type_from_hf({'model_type': 'qwen3_moe'}))"
```

Expected output includes:

```text
megatron.lite lite LiteConfig qwen3
```


sbhavani:
@ISEEKYAN Thanks for these efforts to improve Megatron user experience! I think this branch is useful as an exploration, but the PR has grown beyond what we should try to land as the first step.

I’d recommend converting this PR to a draft and opening a new minimal PR that only lands a minimal skeleton:

- docs explaining the motivation and non-goals
- API contracts / interfaces only
- skills or instructions for agentic experimentation
- scripts for validation/parity workflow
- no full Qwen model implementation, use a toy/dummy example
- no FSDP2, LoRA, checkpoint conversion, or VERL integration yet

The larger implementation pieces can later be split into follow-up PRs, one experiment at a time. That would make review much easier and keep the first landed change aligned with the goal: faster Megatron experimentation.

Also for naming, please avoid the 'lite' naming. I think something like `examples/agent_compose/` or `examples/megatron_compose/` would better match the intended goal.

ISEEKYAN:
Thanks @sbhavani, this makes sense from a reviewability perspective, and I agree this has grown too large for a first main-branch landing PR.

One point I want to clarify is the role of “skills” in this exploration. Based on my experiments, the skills/instructions alone are not the most important part of making this agentic. The critical part is the combination of skills with modular, isolated, runnable reference code. The code gives the agent concrete composition boundaries and implementation examples to follow; the skills describe how to use and extend those boundaries. In practice these two parts are tightly coupled, and if we land only abstract contracts or skills without the real modular implementation, we lose much of the agentic value.

I’m also a bit concerned that a toy/dummy example would not exercise the actual hard integration surfaces we are trying to explore: real model layouts, MoE, checkpoint conversion, optimizer state continuity, HF export, VERL integration, offload, etc. Those are exactly the places where users and downstream systems currently struggle, and where a runnable reference implementation can provide useful feedback.

That said, I agree the path to main should be split. For main-branch PRs, I can follow your suggestion and break this down into smaller reviewable pieces: motivation/non-goals, API contracts, validation workflow, skills/instructions, then separate follow-ups for model implementations, checkpoint/export, VERL, FSDP2/offload, etc.

For this PR specifically, since it targets `dev` and is under an experimental area, my preference would be to merge the complete exploration into `dev` first so community users and customers can try the current OOTB VERL training path and give feedback on the actual end-to-end experience. We can then use that feedback to decide what should be promoted to main and in what shape.

On naming, I’m open to changing it. I used `lite` to emphasize a smaller runtime surface, but if `agent_compose` / `megatron_compose` better matches the long-term intent, I’m happy to discuss or rename as part of the split main-branch PRs.

sbhavani:
@ISEEKYAN thanks for being flexible and I'm glad we can start with the skeleton for the initial PR direct to main.

For this PR, I'd still recommend to keep it as a draft or create a new branch. The dev branch is regularly synced with main and we would eventually need an equivalent main PR. The code changes are also big enough that it could be a separate branch.

FDecaYed (approved):

sbhavani:
Could we rename the dir from lite to "megatron-compose"? 

And why not keep it in examples/? I don't think we need a new experimental dir.

PR #4891, opened 2026-05-20
Fix hybrid M-FSDP numerics

janEbert:
Fixes M-FSDP numerics for hybrid topologies by correctly preserving `DTensor` gradient metadata through grad-stat collection and completing delayed outer-DP gradient reductions for HSDP/HFSDP before optimizer gradients are attached.

Testing Adam on `main` (currently 859b7197f19119f53113ab6d30d1901b87ca1197) vs. this PR branch over 100 training steps on a 1B model.

| Mode    | main mean / last loss abs diff | fix mean / last loss abs diff |
|---------|-------------------------------:|------------------------------:|
| M-FSDP  |          2.268e-01 / 3.986e-02 |         2.265e-01 / 4.621e-02 |
| M-HSDP  |          1.527e+00 / 3.017e-01 |         2.546e-01 / 4.859e-02 |
| M-HFSDP |          6.713e-01 / 1.252e-01 |         2.723e-01 / 1.628e-02 |

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
@@ -4742,6 +4742,8 @@ def make_fsdp_dtensor(
         shape=param.shape,
         stride=param.stride(),
     )
+    setattr(fsdp_tensor, "megatron_fsdp_dist_index", dist_index)
+    setattr(fsdp_tensor, "megatron_fsdp_is_expert_param", is_expert_param)
 
     if run_check:
         validate_uneven_dtensor(fsdp_tensor)
--- a/megatron/core/optimizer/optimizer.py
+++ b/megatron/core/optimizer/optimizer.py
@@ -198,16 +198,10 @@ class MegatronOptimizer(ABC):
                 and getattr(param, "__fsdp_param__", False)
             ):
                 grad = param.decoupled_grad if hasattr(param, "decoupled_grad") else None
-                if (
-                    getattr(param, "__fsdp_param__", False)
-                    and grad is not None
-                    and hasattr(grad, "_local_tensor")
-                ):
-                    # Megatron-FSDP gradients are DTensors.
-                    grad = grad._local_tensor
             elif getattr(param, "__fsdp_param__", False):
-                # Megatron-FSDP gradients are DTensors.
-                grad = param.grad._local_tensor if param.grad is not None else None
+                # Megatron-FSDP gradients are DTensors. Keep the DTensor
+                # wrapper so grad-norm code can reduce over FSDP shard groups.
+                grad = param.grad
             else:
                 grad = param.grad
             grad_not_none = grad is not None
--- a/megatron/core/utils.py
+++ b/megatron/core/utils.py
@@ -40,6 +40,8 @@ try:
     HAVE_DTENSOR = True
 except ImportError:
     HAVE_DTENSOR = False
+    DTensor = None
+    Shard = None
 
 from megatron.core import parallel_state
 from megatron.core.dist_checkpointing.mapping import ShardedTensor
@@ -1111,15 +1113,39 @@ def to_local_if_dtensor(tensor: Union[torch.Tensor, "DTensor"]) -> torch.Tensor:
         return tensor.to_local() if HAVE_DTENSOR and isinstance(tensor, DTensor) else tensor
 
 
+def is_same_process_group(
+    group: torch.distributed.ProcessGroup | None, other: torch.distributed.ProcessGroup | None
+) -> bool:
+    """Returns whether two process groups contain the same ranks."""
+    if group is None or other is None:
+        return False
+    if group is other:
+        return True
+    try:
+        return torch.distributed.get_process_group_ranks(
+            group
+        ) == torch.distributed.get_process_group_ranks(other)
+    except (RuntimeError, ValueError):
+        return False
+
+
 def get_data_parallel_group_if_dtensor(
     tensor: Union[torch.Tensor, "DTensor"], data_parallel_group: "ProcessGroup" = None
 ) -> Optional["ProcessGroup"]:
     """Gets the data parallel group of the given tensor if it is a DTensor."""
     if HAVE_DTENSOR and isinstance(tensor, DTensor):
-        current_group = tensor.device_mesh.get_group()
-        assert data_parallel_group is None or current_group == data_parallel_group
+        dist_index = getattr(tensor, "megatron_fsdp_dist_index", None)
+        if dist_index is not None:
+            current_group = dist_index.get_dp_group(
+                is_expert_parallel=getattr(tensor, "megatron_fsdp_is_expert_param", False)
+            )
+        else:
+            current_group = tensor.device_mesh.get_group()
+        assert data_parallel_group is None or is_same_process_group(
+            current_group, data_parallel_group
+        )
         return current_group
-    return None
+    return data_parallel_group
 
 
 def prepare_input_tensors_for_wgrad_compute(grad_output, all_gathered_input):
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
@@ -320,8 +320,11 @@ class MegatronFSDP(torch.nn.Module):
             # Default to overlapped parameter gather when fully-sharding.
             self.ddp_config.overlap_param_gather = True
         if self.ddp_config.data_parallel_sharding_strategy in ["optim_grads_params", "optim_grads"]:
-            # Default to overlapped gradient reduce-scatter when sharding gradients.
-            self.ddp_config.overlap_grad_reduce = True
+            if not self.ddp_config.overlap_grad_reduce:
+                raise ValueError(
+                    "Megatron-FSDP requires `overlap_grad_reduce=True` when "
+                    "`data_parallel_sharding_strategy` is 'optim_grads' or 'optim_grads_params'."
+                )
         if not self.is_delay_grad_reduce:
             # Gradient reduce-scatter must be overlapped when using sharding optimizer
             # and gradients.
--- a/megatron/core/optimizer/clip_grads.py
+++ b/megatron/core/optimizer/clip_grads.py
@@ -49,7 +49,7 @@ except ImportError:
 
 from ..tensor_parallel import param_is_not_tensor_parallel_duplicate
 from ..transformer.module import param_is_not_shared
-from ..utils import get_data_parallel_group_if_dtensor, to_local_if_dtensor
+from ..utils import get_data_parallel_group_if_dtensor, is_same_process_group, to_local_if_dtensor
 
 
 def get_grad_norm_fp32(
@@ -99,9 +99,10 @@ def get_grad_norm_fp32(
             torch.distributed.all_reduce(
                 total_norm_cuda, op=torch.distributed.ReduceOp.MAX, group=data_parallel_group
             )
-        torch.distributed.all_reduce(
-            total_norm_cuda, op=torch.distributed.ReduceOp.MAX, group=grad_stats_parallel_group
-        )
+        if not is_same_process_group(data_parallel_group, grad_stats_parallel_group):
+            torch.distributed.all_reduce(
+                total_norm_cuda, op=torch.distributed.ReduceOp.MAX, group=grad_stats_parallel_group
+            )
         total_norm = total_norm_cuda[0].item()
 
     else:
@@ -133,9 +134,10 @@ def get_grad_norm_fp32(
             torch.distributed.all_reduce(
                 total_norm, op=torch.distributed.ReduceOp.SUM, group=data_parallel_group
             )
-        torch.distributed.all_reduce(
-            total_norm, op=torch.distributed.ReduceOp.SUM, group=grad_stats_parallel_group
-        )
+        if not is_same_process_group(data_parallel_group, grad_stats_parallel_group):
+            torch.distributed.all_reduce(
+                total_norm, op=torch.distributed.ReduceOp.SUM, group=grad_stats_parallel_group
+            )
         if multi_tensor_scale_tensor_impl is not None:
             total_norm = total_norm.pow(1.0 / norm_type)
         else:
@@ -232,18 +234,9 @@ def count_zeros_fp32(
     #   - should not be a replica due to tensor model parallelism
     total_num_zeros = torch.zeros(1, dtype=torch.int64, device='cuda')
     data_parallel_group = None
-    use_megatron_fsdp = False
     for param in parameters:
         grad_attr = "decoupled_grad" if use_decoupled_grad else "grad"
         grad_not_none = hasattr(param, grad_attr) and getattr(param, grad_attr) is not None
-        if getattr(param, "__fsdp_param__", False) and grad_not_none:
-            # If the parameter is managed by Megatron FSDP, the gradient is
-            # an FSDP-sharded DTensor, and we should use the local shard.
-            use_megatron_fsdp = True
-            grad = getattr(param, grad_attr)._local_tensor
-            num_zeros = grad.numel() - torch.count_nonzero(grad)
-            total_num_zeros += num_zeros
-            continue
         is_not_shared = param_is_not_shared(param)
         is_not_tp_duplicate = param_is_not_tensor_parallel_duplicate(param, tp_group=tp_group)
         if grad_not_none and is_not_shared and is_not_tp_duplicate:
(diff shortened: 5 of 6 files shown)


megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -3795,6 +3887,7 @@ def _bucket_group_gradient_reduce(
                     main_grad_buffer.copy_(reduced_grad)
 
             reduce_scatter_view_out_event = self.outer_fsdp_group_grad_reduce_stream.record_event()
+            self._pending_outer_fsdp_grad_reduce_buckets.difference_update(bucket_group)
```
cspades: Why do we need this? The outer reduce-scatter should only happen on the optimizer step, and it should be waited on during `finalize_model_grads`: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py#L3848

megatron/core/utils.py
```diff
@@ -1052,6 +1060,67 @@ def to_local_if_dtensor(tensor: Union[torch.Tensor, "DTensor"]) -> torch.Tensor:
...
+    for mesh_dim, placement in enumerate(tensor.placements):
+        if not isinstance(placement, shard_placement_types):
+            continue
+        mesh_dim_name = None
+        if mesh_dim_names is not None and mesh_dim < len(mesh_dim_names):
+            mesh_dim_name = mesh_dim_names[mesh_dim]
+        if mesh_dim_name == "tp":
+            continue
+        append_unique_process_group(groups, mesh.get_group(mesh_dim))
+    return groups
```
cspades: Instead of digging out the groups, can we just get the full DP group (including both inner and outer DP) from `FSDPDistributedIndex` for a single all-reduce call on all DP ranks instead of 2 calls? All group retrieval should come from there.

janEbert: Done.

megatron/core/optimizer/optimizer.py
```diff
@@ -194,16 +194,10 @@ def _filter_grads_for_norm(
...
-                    and hasattr(grad, "_local_tensor")
-                ):
-                    # Megatron-FSDP gradients are DTensors.
-                    grad = grad._local_tensor
             elif getattr(param, "__fsdp_param__", False):
-                # Megatron-FSDP gradients are DTensors.
-                grad = param.grad._local_tensor if param.grad is not None else None
+                # Megatron-FSDP gradients are DTensors. Keep the DTensor
+                # wrapper so grad-norm code can reduce over FSDP shard groups.
+                grad = param.grad
```
cspades, line 204: Why are we no longer using DTensor gradients (to mirror the DTensor optimizer parameters, or else you get the `DTensor <> Tensor operations not supported` error for instance). Is this relevant to this PR?

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: Still not sure what this is doing: `finish_pending_outer_fsdp_grad_reduce`

When we call `finalize_model_grads` all the way back in the FWD-BWD function in `schedules.py`, DP-Outer should have finished because we wait on `reduce_scatter_view_out_event = self.outer_fsdp_group_grad_reduce_stream.record_event()`. If we are using a full CUDA graph, I'm not even sure you will be able to call `current_stream.wait_stream(self.outer_fsdp_group_grad_reduce_stream)` without an error saying you are not allowed to wait on a stream that is under the CG's capture scope.

janEbert: In short, the `--overlap-grad-reduce` argument and the `MegatronFSDP.ddp_config.overlap_grad_reduce` settings are in conflict.

There are a few steps/interactions to it:

1. I did _not_ pass `--overlap-grad-reduce`, but I passed `--data-parallel-sharding-strategy optim_grads_params`, so [this code section was entered](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L322-L324). I.e., `MegatronFSDP.ddp_config.overlap_grad_reduce = True`, even though `--overlap-grad-reduce` wasn't passed.
2. [This code section is only entered](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/training/training.py#L3384-L3395) when `--overlap-grad-reduce` is passed. It sets up [`config.no_sync_func` (which, by default, is `None`](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/core/model_parallel_config.py#L135-L139); this fact is also asserted in the code block). Specifically, it _would_ set `config.no_sync_func = [model_chunk.no_sync for model_chunk in model]`, which _would_ resolve to the [MegatronFSDP.no_sync](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L1110-L1122) method in our case. However, remember that this does **not** happen and `config.no_sync_func` remains `None`, so [`no_sync_func` becomes a `nullcontext`](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/core/pipeline_parallel/schedules.py#L733-L735).
3. In turn, `MegatronFSDP.is_last_microbatch` is never set and [defaults to `False` here in this code block](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L708-L722). (You can ignore `MegatronFSDP.model_auto_sync` here, since it'll never be `True` in the current Megatron-LM integration, as [we don't pass `sync_model_each_microbatch`](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/core/distributed/fsdp/mcore_fsdp_adapter.py#L188-L213) and [the `MegatronFSDP.sync` method](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L1124-L1134) isn't called anywhere.)
4. While [we do enter the code block linked in the previous step](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L708-L722), [the `outer_fsdp_group_grad_reduce` condition will now **always** be `False`](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L717-L721). Therefore, [the DP-Outer grad reduction in `GradReducePipeline._bucket_group_gradient_reduce`](https://github.com/NVIDIA/Megatron-LM/blob/0552f29307204801ae18a805f2f49775f6652064/megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py#L3727-L3818) is not executed. That code block was copy-pasted to `MegatronFSDP.finish_pending_outer_fsdp_grad_reduce` to ensure it's executed.

I'm testing a fix now so that the problem in step 1 doesn't occur anymore, because `MegatronFSDP.ddp_config.overlap_grad_reduce` is now also respected. I'll report back and update here when the tests pass.

cspades: Just a warning that some of the points claimed above (last microbatch for sure) do not seem to be true (or if true, would suggest much bigger problems that would be instantly noticeable, for instance we would have ZeRO-1 and HFSDP call inner/outer reduce-scatter every microbatch instead of every global step, which we even have seen for Muon testing on your code is not the case). Proceed carefully!

janEbert: Which of the statements do you not agree with? I linked all corresponding code sections. Maybe you never see the issues because you always pass `--overlap-grad-reduce`?

megatron/core/optimizer/optimizer.py
janEbert: The DTensor metadata enables us to reliably get a `data_parallel_group` in `clip_grads.get_grad_norm_fp32`.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: Right, we always pass RS overlap when using everything except the no-shard / DDP strategy (as it is the only one that uses a gradient all-reduce in-place the gradient buffer). So if there is a problem with that logic, it should be fixed there instead of changing everything else.

I didn't take a close look at all of your linked code since I am pretty sure it is detached from the actual issue, but if this is where we are truly at then I can take a closer look.

The goal is to not have this extra DP-Outer logic since it is being strung along for the ride.

cspades: Can you just add some arguments.py code that sets RS overlap to true if using anything except no-shard, and AG overlap to true if we are using optim-grads-params?

is_last_microbatch=True must be set at every optimizer step!

megatron/core/optimizer/optimizer.py
cspades: What do you mean by "reliably"? You can always attach the FSDP Dist Index as an attribute of the DTensor, if that is what you are suggesting?

janEbert: Yes, exactly, that's the attribute we need (to access the `get_dp_group` method). Are you suggesting that we should keep the unwrapping code and instead attach a new attribute to the local tensors? To me, that would seem like a hack.

What I haven't understood yet is, what's the reason to keep the unwrapping? I.e., why do we not keep the (in my eyes) more simple/API-compliant solution that keeps the tensor wrapped for longer? You mentioned an error, which I haven't seen in my tests, but maybe I need to expand the scope to include the settings that trigger it.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
janEbert: I can go with your suggestion; however, what do you think of 1e26b9fb5a instead?

cspades: I think setting it in args may catch any other gaps there may be in the codebase using this argument congruently to DDPConfig's version of it.

megatron/core/optimizer/optimizer.py
cspades: So it won't matter for clip grads (since it is a unary op but there's more to consider than that, see below) but in general we should use Tensors until MLM, TE, etc. actually becomes DTensor compliant.

This can manifest itself through nullptrs (DTensor->data = nullptr while DTensor->_local_tensor->data is valid) which can both silently or non-silently produce functional errors, or if this DTensor is used in some future code in this function that expects this to be a Tensor for binary ops, you will hit an error there too.

I'd just extract the DP group above without changing this unwrapping logic to avoid future issues, until there is a better design pattern. Requires discussions beyond the scope of this PR.

megatron/core/utils.py
```diff
@@ -1111,15 +1113,39 @@ def to_local_if_dtensor(tensor: Union[torch.Tensor, "DTensor"]) -> torch.Tensor:
...
+    except (RuntimeError, ValueError):
+        return False
+
+
 def get_data_parallel_group_if_dtensor(
     tensor: Union[torch.Tensor, "DTensor"], data_parallel_group: "ProcessGroup" = None
 ) -> Optional["ProcessGroup"]:
     """Gets the data parallel group of the given tensor if it is a DTensor."""
     if HAVE_DTENSOR and isinstance(tensor, DTensor):
-        current_group = tensor.device_mesh.get_group()
```
cspades, line 1119: Note to self, report this to PyTorch after checking what this usually returns. Something is fishy here.

megatron/core/optimizer/optimizer.py
janEbert: Got it, thanks a lot for the explanation.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
janEbert: Okay, I'll move it. Note that your suggestion hardcodes some assumptions and may not catch other potential future cases, so there's a tradeoff either way.

cspades: I get the hard-coded stuff part (plus need to mirror to MBridge and maybe move all of this stuff to validation arg containers), but isn't this the best we can do to make sure all these args are consistent everywhere?

Only setting the variable in training.py (but not in args) means only those functions will have the correct logic. If there's a better idea to make sure it's consistent with MFSDP DDPConfig's setting then let's do that.

janEbert: What do you think of the following?

In [`megatron_fsdp.py::MegatronFSDP::__init__`](https://github.com/NVIDIA/Megatron-LM/blob/b9f7fb68a6eda06af359cbc8745b44d93e1383c5/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L322-L324):
```patch
         if self.ddp_config.data_parallel_sharding_strategy in ["optim_grads_params", "optim_grads"]:
-            # Default to overlapped gradient reduce-scatter when sharding gradients.
-            self.ddp_config.overlap_grad_reduce = True
+            if not self.ddp_config.overlap_grad_reduce:
+                    "Megatron-FSDP requires `overlap_grad_reduce=True` when "
+                    "`data_parallel_sharding_strategy` is 'optim_grads' or "
+                    "'optim_grads_params'."
+                )
```

That way, we require `--overlap-grad-reduce` to be set (or `overlap_grad_reduce=True` in the case of standalone use) instead of silently changing the config. It makes OOTB support slightly worse due to less magic being applied, as it moves responsibility for correct configuration to the user.

megatron/core/optimizer/optimizer.py
cspades: Just noticed that the DTensors are localized later right before the TE `multi_tensor_applier` kernel. So I guess we can remove the localization here.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: I think this at least covers ZeRO-2 and FSDP, but I need to take a closer look at ZeRO-1 and DDP later.

megatron/core/optimizer/optimizer.py
janEbert: Just to be really clear, this means you're fine with keeping this change? Even considering potential errors in other places you mentioned?

cspades (approved):
@shjwudp @wujingyue Please take a look as well.

megatron/core/optimizer/optimizer.py
cspades: Yes, can you double-check every place where the output of this function is used? Report back if you find any situations that could hurt the developer experience, i.e. if there's a long section of code where these DTensors could pass through a region where users would expect Tensors.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
janEbert: Regarding `--overlap-param-gather`, these could potentially lead to breakage: https://github.com/NVIDIA/Megatron-LM/blob/872442adc71fbff72db46ca35674dff9e4dabe83/megatron/training/arguments.py#L1616-L1626 and https://github.com/NVIDIA/Megatron-LM/blob/872442adc71fbff72db46ca35674dff9e4dabe83/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L319-L321

I could apply the same change above for `overlap_param_gather`, just for safety.

cspades: The complexity in all of this is that this variable is completely overloaded. It may very well be the case that False in the training loop but True in MFSDP is totally fine. For overlapping the grad reduction, I think it is okay to force them to be consistent. But for param gather, frozen parameters may just work fine with FSDP but not work for DDP all-gather post-optim step.

I think you should let me handle this in a future PR. As long as your gradient reductions are working, this should hopefully be a upgrade.

janEbert: Kk, I deem this thread resolved, then. Won't apply the symmetric solution to `overlap_param_gather`.

cspades:
@janEbert Test failures are trivial, just add `overlap_grad_reduce=True` to the DDPConfig of all the tests that failed.

megatron/core/optimizer/optimizer.py
cspades: It should just be `get_grad_norm_fp32`, which localizes the DTensor, and that's the end of this path.

janEbert: Yeah, all uses are put straight into `get_grad_norm_fp32`, I agree. That path uses `.to_local()` though, which you said may be suboptimal in cases, right? E.g., it may cause copies. Is it still good enough?

cspades: Main issue with copies is it doesn't bring over attributes and maybe perf if you do a lot of copies, both of which are not super important. I also think the copies depends on some other mechanics like differentiability, which is also not relevant here and may not be triggered at all. It's fine TL;DR

Phlip79 (approved):

jaredcasper (approved):

PR #4898, opened 2026-05-20
test: skip cuTile fused mHC tests on Hopper

Phlip79:
Not only should we skip cuTile if it's not available, we should also skip it if we're running on Hopper GPUs.


balasaajay (approved):

yaox12 (approved):

Phlip79:
Closed in favor of https://github.com/NVIDIA/Megatron-LM/pull/4886

PR #4899, opened 2026-05-20
Make Mamba conv params direct mixer params

wujingyue:
## Summary
- Make Mamba causal-conv weight and bias direct `MambaMixer` parameters instead of parameters owned by a nested `nn.Conv1d` module.
- Keep the fused conv paths reading the same tensors through `MambaContextParallel`, while allowing Megatron FSDP's non-unit `MambaMixer` pre-forward hook to see and gather those params directly.
- Add a unit-test repro that disables AG prefetch and verifies the fused Mamba conv path does not depend on incidental prefetch of the old nested conv bucket.

## Issue
As discussed in https://github.com/NVIDIA/Megatron-LM/pull/4717#issuecomment-4445336140, `MambaMixer.forward` uses the fused causal-conv kernels and reads the conv weight/bias as raw tensors instead of calling the `conv1d` submodule. Since `nn.Module.__call__` is never invoked on that submodule, the submodule's FSDP pre-forward hook does not fire.

The parent `MambaMixer` hook does fire, but for non-FSDP-unit modules it gathers only `module.parameters(recurse=False)`. With the old structure, `conv1d.weight` and `conv1d.bias` were nested submodule params, so they were invisible to that gather path and only worked when some earlier opportunistic AG prefetch happened to gather their bucket. When prefetch is suppressed, the fused conv path can dereference an ungathered/freed bucket and hit CUDA illegal memory access.

This PR removes that mismatch by making the conv params shallow `MambaMixer` params, so the existing non-unit gather path covers the tensors that `MambaMixer.forward` directly reads. This is arguably a targeted workaround for the current FSDP failure mode, but this ownership mismatch has triggered enough corner cases that it is also the cleaner module contract: if `MambaMixer.forward` directly uses these tensors, `MambaMixer` should own them directly.

There is a similar ownership problem in `GatedDeltaNet`; this PR leaves it out of scope and we will address it separately.

## Tests
- `git diff --check`
- `python -m py_compile megatron/core/ssm/mamba_mixer.py megatron/core/ssm/mamba_context_parallel.py tests/unit_tests/ssm/test_mamba_context_parallel.py tests/unit_tests/distributed/megatron_fsdp/test_mcore_fully_sharded_data_parallel.py`
- Full unit test not run locally; PR has `Run tests` label for CI.


jaredcasper (approved):

Phlip79:
This looks like it would break loading of "old" checkpoints. @ericharper what is the precedent here? Add a hook to `MambaMixer` that rewrites `conv1d.weight` to `conv1d_weight` etc...?

wujingyue:
> This looks like it would break loading of "old" checkpoints. @ericharper what is the precedent here? Add a hook to `MambaMixer` that rewrites `conv1d.weight` to `conv1d_weight` etc...?

That’s indeed reflected in CI. I’ll update MambaMixer.sharded_state_dict to map the new names back to the old ones, though let me know if there’s a cleaner approach.

I know some projects handle checkpoint compatibility through model versioning — for example, allowing incompatibility across major versions but preserving compatibility across minor versions. I’m curious whether we have a similar release/versioning process in place.

megatron/core/ssm/mamba_mixer.py
```diff
@@ -243,6 +243,7 @@ def __init__(
         assert self.nheads % self.ngroups == 0, "nheads must be evenly divisible by ngroups"
 
         assert not bias
+        assert conv_bias
```
janEbert, line 251: It wouldn't be too much hassle to support `conv_bias=False`, right? [`torch.nn.functional.conv1d`](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.conv1d.html) supports `bias=None`, but it would require some extra handling. Not sure if that feature would ever be used, though.

janEbert (approved):
Thanks for the lovely issue description. ~One main concern I have is backward-compatibility with regard to checkpoints. We could keep it by transforming the `conv1d_{weight,bias}` keys in the state dict so they look like before (i.e., they'd become `conv1d.{weight,bias}`).~

Oops, I just saw the handling logic and previous discussion. I'm not sure if it's correct to have the transformation in `sharded_state_dict`; instead, it could be placed in `megatron.training.checkpointing.generate_state_dict`. Maybe @cspades could also chime in here if the `sharded_state_dict` approach is okay after all.

megatron/core/ssm/mamba_mixer.py
wujingyue: The old code makes this assumption already, just not explicitly asserted. I could make code work with conv_bias==None but chose not to for https://google.github.io/eng-practices/review/developer/small-cls.html

wujingyue:
> instead, it could be placed in `megatron.training.checkpointing.generate_state_dict`

Layer-specific adapter code should live alongside the corresponding layer for readability and maintenance.

wujingyue:
Thanks for the review, @janEbert !

ericharper:
> > This looks like it would break loading of "old" checkpoints. @ericharper what is the precedent here? Add a hook to `MambaMixer` that rewrites `conv1d.weight` to `conv1d_weight` etc...?
> 
> That’s indeed reflected in CI. I’ll update MambaMixer.sharded_state_dict to map the new names back to the old ones, though let me know if there’s a cleaner approach.
> 
> I know some projects handle checkpoint compatibility through model versioning — for example, allowing incompatibility across major versions but preserving compatibility across minor versions. I’m curious whether we have a similar release/versioning process in place.

@NVIDIA/dist-checkpointing, @Phlip79, let's align on our checkpoint compatibility strategy and clearly document our guidelines.

santhnm2 (approved):

wujingyue:
@ericharper your turn :)

ericharper:
@balasaajay, I'm seeing these failures in the unit tests:   4-task-1-0/0 [default0]:FAILED tests/unit_tests/elastification/test_flextron_mamba_elasticity_manager.py::TestFlextronMambaElasticityManager::test_attach_produces_hook_handles - AttributeError: 'MambaMixer' object has no attribute 'conv1d'
  4-task-1-0/0 [default0]:FAILED tests/unit_tests/elastification/test_flextron_mamba_elasticity_manager.py::TestFlextronMambaElasticityManager::test_current_router_none_preserves_output - AttributeError: 'MambaMixer' object has no attribute 'conv1d'
  4-task-1-0/0 [default0]:FAILED tests/unit_tests/elastification/test_flextron_mamba_elasticity_manager.py::TestFlextronMambaElasticityManager::test_full_budget_one_hot_approx_matches_baseline - AttributeError: 'MambaMixer' object has no attribute 'conv1d'
  4-task-1-0/0 [default0]:FAILED tests/unit_tests/elastification/test_flextron_mamba_elasticity_manager.py::TestFlextronMambaElasticityManager::test_small_budget_one_hot_changes_output - AttributeError: 'MambaMixer' object has no attribute 'conv1d'
  4-task-1-0/0 [default0]:FAILED tests/unit_tests/elastification/test_flextron_mamba_elasticity_manager.py::TestFlextronMambaElasticityManager::test_detach_restores_baseline - AttributeError: 'MambaMixer' object has no attribute 'conv1d'
  4-task-1-0/0 [default0]:FAILED tests/unit_tests/elastification/test_flextron_mamba_elasticity_manager.py::TestAddFlextronMambaElasticity::test_factory_returns_manager - AttributeError: 'MambaMixer' object has no attribute 'conv1d'
  4-task-1-0/0 [default0]:= 6 failed, 1469 passed, 60 skipped, 85 deselected, 5775 warnings in 912.41s (0:15:12) =

But the unit tests are showing as passing.

@wujingyue , can you take a look?

ericharper:
I think this breaks bridge as well? https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/94dc04baf65157463181eef0c19549d5a6e4ccec/src/megatron/bridge/models/nemotronh/nemotron_h_bridge.py#L501

@yaoyu-33 , @cuichenx  can you take a loook?

wujingyue:
Nice catch, @ericharper! I didn't expect CI to show green when it's actually not. Will take a look at the actual results...

ericharper:
> Nice catch, @ericharper! I didn't expect CI to show green when it's actually not. Will take a look at the actual results...

Yeah it's not supposed to :) But CI team is aware and doing impact analysis.

megatron/core/ssm/mamba_mixer.py
janEbert: Makes sense.

janEbert:
>> instead, it could be placed in `megatron.training.checkpointing.generate_state_dict`
>
> Layer-specific adapter code should live alongside the corresponding layer for readability > and maintenance.

I looked back into my discussion with @cspades on this back when I faced the same choice. The main reason was to keep `sharded_state_dict` mapping to the original FQNs and that due to the method's ubiquity, it shouldn't be modified in model-specific ways. I'm also finding the divergence with regard to `state_dict` potentially confusing.

That said, I do agree with your statement as well and that was my original preferred solution as well. If using `sharded_state_dict` doesn't mess things up, it is certainly a much cleaner solution than placing additional transformations that would break the checkpoints by default when the required `sharded_state_dict` transformations are missing.

megatron/elastification/flextron_elasticity_hooks.py
```diff
@@ -44,9 +59,10 @@ def _init_embedding_masks(self):
         """Initialize embedding dimension masks."""
         mask_list = []
         for emb_int in self.config.emb_int_list:
-            assert (
-                0 <= emb_int <= self.config.hidden_size
-            ), f'emb_int_list entries must be in [0, hidden_size={self.config.hidden_size}], got {emb_int}.'
+            assert 0 <= emb_int <= self.config.hidden_size, (
```
wujingyue, line 62: lint only

megatron/elastification/flextron_elasticity_hooks.py
```diff
@@ -118,12 +134,14 @@ def _init_mamba_masks(self):
 
         tp_size = parallel_state.get_tensor_model_parallel_world_size()
         for mamba_int in self.config.mamba_int_list:
-            assert (
-                0 <= mamba_int <= self.mamba_mixer.nheads
-            ), f"mamba_int_list entries must be in [0, nheads={self.mamba_mixer.nheads}], got {mamba_int}."
-            assert (
-                mamba_int % tp_size == 0
-            ), f"mamba_int_list entries must be evenly divisible by tp_size={tp_size}, got {mamba_int}."
+            assert 0 <= mamba_int <= self.mamba_mixer.nheads, (
```
wujingyue, line 137: lint only

wujingyue:
ok to test

sheliang-nv (approved):

ericharper (approved):

PR #4919, opened 2026-05-21
inference(MLA): swap dispatch to FlashInfer BatchMLAPagedAttentionWrapper (2/3)

shanmugamr1992:
## Summary

**Part 2 of 3** in the MLA dynamic-inference rewrite series. This PR is
where the actual kernel swap happens. Builds on
[PR #4918](https://github.com/NVIDIA/Megatron-LM/pull/4918) (which dropped
the FlashMLA-specific block-size guard).

## What changes

The MLA dynamic-batching path no longer goes through ``flash_decode_and_prefill``
(which routed to FlashMLA for decode + Flash-Attention paged for prefill,
with mutually incompatible page-size requirements). It now goes through a
single ``DynamicInferenceContext.mla_paged_attention(...)`` helper backed by
FlashInfer's ``BatchMLAPagedAttentionWrapper``, which:

- handles decode + (incremental) prefill in one kernel,
- accepts arbitrary page sizes,
- runs in compressed-KV / ``kv_lora_rank`` space (matches what vLLM and
  SGLang do for DeepSeek-style models).

### File-level breakdown

| File | What |
|---|---|
| ``megatron/core/inference/contexts/dynamic_context.py`` | Adds ``_build_mla_paged_attention_wrappers()`` (one wrapper per CUDA-graph shape + eager, sharing a 128MB float workspace) and ``mla_paged_attention()`` (picks the wrapper, builds FlashInfer indptrs from the existing per-step CPU mirrors, plans + runs against zero-copy ckv/kpe slices of the layer cache). Stores ``mla_kv_lora_rank`` / ``mla_qk_pos_emb_head_dim`` on the context. |
| ``megatron/core/transformer/multi_latent_attention.py`` | Q absorption (``einsum sbhd,hdk->sbhk``) is now unconditional whenever ``cache_mla_latents=True`` — matches what FlashInfer expects. Dispatches to ``inference_context.mla_paged_attention(...)``. Lifts the ``up_v_weight`` einsum out of the ``is_decode_only()`` guard so prefill steps also get the latent → ``v_head_dim`` map-back. Drops the now-unused ``rearrange`` import and the prefill-only output rearrange branch. |
| ``megatron/core/transformer/attention.py`` | Removes the MLA branch inside ``flash_decode_and_prefill`` (function is no longer MLA-aware). Removes the optional ``flash_mla`` import and the ``MLATransformerConfig`` import that only gated the deleted branch. |
| ``pyproject.toml``, ``uv.lock`` | Drops the ``flash_mla`` git source, the ``no_pypi_wheels`` entry, and the ``no-build-isolation-package`` entry. |
| ``megatron/core/transformer/transformer_config.py``, ``megatron/training/arguments.py`` | Updates the ``cache_mla_latents`` docstring and CLI help to reference FlashInfer. |

### Mathematical equivalence

Both the existing decode-only absorbed path and the prefill non-absorbed
path produce the same logits up to numerical precision. The change here
makes both paths use the absorbed kernel, which is mathematically
equivalent (``W_UK @ W_UK^T`` absorbed into the Q projection) but
numerically different. Since the prior prefill path was broken anyway
(FA's 256-divisibility check vs. FlashMLA's 64-only page size — the two
kernels cannot coexist on the same paged cache), there are no existing
golden values to preserve. PR 3 of this series adds the functional-test
coverage that should have existed all along.

## Series roadmap

- ✅ **PR 1** ([#4918](https://github.com/NVIDIA/Megatron-LM/pull/4918)) — Drop the FlashMLA-specific page-size guard.
- ✅ **This PR** — Swap the dispatch to FlashInfer; drop the ``flash_mla`` dependency.
- ⏳ **PR 3 (upcoming)** — Functional logits-match test + recipe + golden values.

## Test plan

- [x] Unit: ``tests/unit_tests/inference/contexts/test_dynamic_context.py::TestDynamicContext::test_mla_cache_block_size_relaxed`` — the test now also exercises ``_build_mla_paged_attention_wrappers()`` since wrapper allocation is invoked from ``__init__`` whenever ``cache_mla_latents=True``. **Verified on cw-dfw H100** in PR 1.
- [ ] Re-verify on cw-dfw H100 after Stage 2 changes (verification job in flight at time of opening; will update with the result).
- [ ] Functional logits-match — comes in PR 3.
- [ ] CI: ``Run tests`` + ``container::lts`` (because this changes ``pyproject.toml``).

## Notes for reviewers

The non-obvious correctness points:

1. **CPU mirrors → FlashInfer indptrs**. MCore already maintains
   ``_cpu_mha_cu_query_seq_lengths``, ``_cpu_mha_cu_kv_seq_lengths``,
   ``_cpu_mha_kv_seq_lengths`` and ``_cpu_mha_block_table`` per step (with
   ``-1`` sentinels in the block-table padding rows). ``mla_paged_attention``
   reuses those CPU mirrors and converts the ``[batch, max_kv_blocks]``
   block-table to FlashInfer's flat ``(kv_indices, kv_indptr)`` form,
   dropping the sentinels naturally. Padded rows have ``kv_len=0``, so the
   conversion loop contributes 0 blocks for them.

2. **CUDA-graph wrapper buffers**. Each wrapper in
   ``self._mla_graph_wrappers`` is sized to its ``cuda_graph_batch_dimensions_list``
   entry's ``req_count``. ``plan()`` then ``.copy_()``s host-side indptrs into
   the pre-allocated GPU buffers, so the wrapper's ``run()`` is captureable.

3. **Always-absorbed Q layout**. The query packed in
   ``qkv_up_proj_and_rope_apply_for_cached_latent_kv`` is now
   ``[s, b, h, kv_lora_rank + qk_pos_emb_head_dim]`` for both decode and
   prefill (previously ``[..., qk_head_dim + qk_pos_emb_head_dim]`` for
   prefill). The forward then splits it back to ``q_nope`` /``q_pe`` to
   call FlashInfer.

### DIFF
--- a/megatron/core/inference/contexts/dynamic_context.py
+++ b/megatron/core/inference/contexts/dynamic_context.py
@@ -58,9 +58,11 @@ except ImportError:
 
 try:
     import flashinfer  # type: ignore # pylint: disable=unused-import
+    from flashinfer.mla import BatchMLAPagedAttentionWrapper
 
     HAVE_FLASHINFER = True
 except ImportError:
+    BatchMLAPagedAttentionWrapper = None  # type: ignore[assignment,misc]
     HAVE_FLASHINFER = False
 
 try:
@@ -390,7 +392,9 @@ class DynamicInferenceContext(BaseInferenceContext):
         self.block_size_tokens = inference_config.block_size_tokens
         if self.cache_mla_latent:
             #   one vector  c_t  (rank)  +  optional RoPE phase slice
-            self.kv_reduced_dim = model_config.kv_lora_rank + model_config.qk_pos_emb_head_dim
+            self.mla_kv_lora_rank = model_config.kv_lora_rank
+            self.mla_qk_pos_emb_head_dim = model_config.qk_pos_emb_head_dim
+            self.kv_reduced_dim = self.mla_kv_lora_rank + self.mla_qk_pos_emb_head_dim
             self.block_size_bytes = (
                 kv_dtype_size_bytes
                 * self.num_attention_layers
@@ -691,6 +695,12 @@ class DynamicInferenceContext(BaseInferenceContext):
         # Allocate GPU state.
         self.is_tensor_state_allocated = False
         self.initialize_all_tensors()
+
+        # FlashInfer MLA paged-attention wrappers (one per CUDA-graph shape + eager).
+        # Allocated last because they depend on the bookkeeping/metadata config that
+        # initialize_all_tensors() finalizes.
+        if self.cache_mla_latent:
+            self._build_mla_paged_attention_wrappers()
 
         # Print info.
         active_blocks = self.kv_block_allocator.active_count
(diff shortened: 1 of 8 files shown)


shanmugamr1992:
**Stage 2 verification update**: Just confirmed on cw-dfw H100 (job 11976112) — the cog job rebuilt uv env from scratch (because `pyproject.toml` changed), then ran the unit test:

```
4 passed, 39 warnings in 2.52s
PASSED test_mla_cache_block_size_relaxed[16]
PASSED test_mla_cache_block_size_relaxed[32]
PASSED test_mla_cache_block_size_relaxed[64]
PASSED test_mla_cache_block_size_relaxed[128]
```

What this verifies:
- `uv sync --locked` succeeds with the trimmed `uv.lock` (no dangling `flash-mla` references).
- `_build_mla_paged_attention_wrappers()` allocates the FlashInfer wrapper objects + workspace buffer correctly at all four block sizes.
- The construction path doesn't import-error on the new `flashinfer.mla.BatchMLAPagedAttentionWrapper` import.

What's still pending end-to-end verification: `mla_paged_attention()` being actually invoked from `MLASelfAttention.forward()`. That happens in the PR 3 functional test (logits-match against a real H100 inference run). I'll link the result here once PR 3's golden-value run lands.

shanmugamr1992:
**PR2 runtime validated on H100 via PR3's functional test.**

PR3's first green H100 run (cw-dfw, session `mla-debug`) ran the MLA + FlashInfer dispatch end-to-end through 30 generated tokens, validating the runtime correctness of this PR.

One PR2 bug was caught in the process: `DynamicInferenceContext.mla_paged_attention` was passing **non-contiguous** ckv/kpe slices of `layer_buf[..., :kv_lora_rank]` / `[..., kv_lora_rank:]` to FlashInfer. Slicing the innermost dim of a contiguous tensor produces a non-contig view, which FlashInfer reads with the wrong stride and triggers an illegal memory access (which surfaces asynchronously at the next CUBLAS call). Fixed by adding `.contiguous()` at the call site (committed on PR3 as part of the bundle since that's where it was exposed — happy to rebase it onto PR2 if reviewers prefer).

The proper long-term fix is to allocate ckv and kpe as separate buffers in `_allocate_memory_buffer`, avoiding the per-step copy that `.contiguous()` introduces. Marked as a TODO in the code.

Run trace:
```
44m | dynamic | requests: cli, n 1, g 30, r/step 4 | bf: 512 MB
throughput: 2.170 tok/s … total time: 13.822s … steps: 30
```

mem peak ~2.3 GiB, deterministic token output across runs.

PR #4920, opened 2026-05-21
tests/inference: add MLA dynamic inference logits-match test (3/3)

shanmugamr1992:
## Summary

**Part 3 of 3** in the MLA dynamic-inference rewrite series. Adds the
first functional test for the ``cache_mla_latents=True`` path, which
until now had **zero** functional or unit coverage. Builds on:

- [PR #4918](https://github.com/NVIDIA/Megatron-LM/pull/4918) — drop the
  FlashMLA-specific page-size guard.
- [PR #4919](https://github.com/NVIDIA/Megatron-LM/pull/4919) — swap the
  MLA dispatch to FlashInfer's ``BatchMLAPagedAttentionWrapper``; drop
  the ``flash_mla`` dependency.

## What this PR adds

1. ``tests/functional_tests/test_cases/gpt/gpt_mla_dynamic_inference_tp1_pp1_logitsmatch_zmq/model_config.yaml`` —
   a small (4 layer, hidden 512, ``kv_lora_rank=64``) random-init MLA
   inference test case. Mirrors the ``gpt3_mcore_te_tp2_pp2_mla`` training
   shape, downsized to TP=PP=1. Uses ``NullTokenizer`` so the test is
   self-contained (no checkpoint or real tokenizer needed).
2. A single ``test_case`` entry in
   ``tests/test_utils/recipes/h100/gpt-dynamic-inference.yaml`` that wires
   it into the standard dynamic-inference functional-test runner.

## Golden values: deliberately not yet committed

This PR is opened against the assumption that PR 2 lands first. Once it
does, the standard frozen-start workflow generates the
``golden_values_dev_dgx_h100.json`` file on the first green H100 run —
that file then becomes the baseline for future runs. Since the
``cache_mla_latents`` path has never produced any logits in CI before,
there is no prior baseline to preserve.

The intended sequence:

1. PR 2 merges.
2. CI on this PR (with the ``Run functional tests`` label) runs the new
   test case. The first run captures logprobs + generated tokens.
3. ``python tests/test_utils/python_scripts/download_golden_values.py
   --source github --pipeline-id <run-id>`` pulls them.
4. The downloaded ``golden_values_dev_dgx_h100.json`` lands in a
   follow-up commit on this branch.

## Why a single combined test case (decode + prefill + mixed) instead of separate cases

The issue write-up enumerated separate unit-test scenarios
(``test_mla_decode_only``, ``test_mla_prefill_only``,
``test_mla_mixed_batch``, ``test_mla_chunked_prefill``,
``test_mla_with_cuda_graphs``). All of those are exercised by this single
functional test because the test runs a real prompt through prefill +
30-step decode, which by construction hits prefill and decode (and
mixed-batch transitions) within the same engine invocation. Splitting
them into separate unit tests is a follow-up that becomes worthwhile once
the functional test is stable and a regression is suspected at a specific
phase.

## Test plan

- [x] Test case YAML + recipe entry land.
- [ ] CI: ``Run functional tests`` label after PR 2 merges. First run
      records the baseline; second run verifies determinism.
- [ ] Commit the generated ``golden_values_dev_dgx_h100.json`` in a
      follow-up.

## Notes for reviewers

This PR can be reviewed independently of PR 2 — the test case + recipe
addition is pure scaffolding. It will start passing in CI once PR 2 is
merged (or rebased onto a branch with PR 2 applied).


shanmugamr1992:
/ok to test [d1b45c5](https://github.com/NVIDIA/Megatron-LM/pull/4920/commits/d1b45c5358ed88c79f2a760e63455641fcb831a3)

shanmugamr1992:
/ok to test [ea89886](https://github.com/NVIDIA/Megatron-LM/pull/4920/commits/ea89886ff0db8525deb80dd28ad16d021aa36b72)

shanmugamr1992:
/ok to test [9bbdd42](https://github.com/NVIDIA/Megatron-LM/pull/4920/commits/9bbdd42f9259ad502fa7f8423d89d9da1012df0b)

shanmugamr1992:
/ok to test [caa73bb](https://github.com/NVIDIA/Megatron-LM/pull/4920/commits/caa73bbf8abc7a9ff8ab90b2f8836bd9efd7cd56)

shanmugamr1992:
**Update — PR3 is end-to-end green on cw-dfw H100.**

Spent today iterating on the failure cascade. Final fix bundle (commit `776cf0dd5`):

1. **Cache-contiguity bug in PR2** — `DynamicInferenceContext.mla_paged_attention` was slicing `layer_buf[..., :kv_lora_rank]` / `layer_buf[..., kv_lora_rank:]` and passing those views to FlashInfer. Slicing the innermost dim of a contiguous tensor produces a non-contiguous view; FlashInfer reads it with the wrong stride and triggers an illegal memory access (surfaces async at the downstream `up_v_weight` einsum as a CUBLAS error). Fixed with `.contiguous()` plus a TODO for the proper long-term fix (separate ckv/kpe buffers in `_allocate_memory_buffer`). **This is really a PR2 fix that landed on PR3 because the test scaffolding here was what exposed it.**
2. **`NullTokenizer`** had no `bos`/`bos_id` (the controller probes `tokenizer.bos` to strip a leading BOS).
3. **Sub-tensor-core MLA head dims**. The original test config used `qk_pos_emb_head_dim=8`, which JIT-compiles but the H100 SM90 kernel can't actually execute. Switched to FlashInfer's officially-supported `smaller_mla_dimensions` shape (qk_nope=64, qk_rope=64, v=128, kv_lora=256).
4. **Oversized inference buffer for a tiny model** — `buffer_size_gb=4` / tiny KV per request → `max_requests=116504` > `max_tokens=16384`. Pinned `buffer_size_gb=0.5` and `max_requests=32`.

**Run result (cw-dfw, session `mla-debug`):**
```
44m | dynamic | requests: cli, n 1, g 30, r/step 4 | bf: 512 MB
throughput: 2.170 tok/s … total time: 13.822s … steps: 30
```

Golden values captured. Metric set is intentionally `generated_tokens` only — random-init bf16 produces per-token logits big enough to overflow softmax in places, giving NaN logprobs that `math.isclose` can't match. Token-level determinism is the meaningful signal for a path-correctness test. Logprob comparison becomes available against a real checkpoint, which is out of scope here.

PR #4921, opened 2026-05-21
feat: skip final full recompute unit

### LINKED ISSUE #4546: Activation recompute can be skipped on the final instance of the recomputed modules to improve performance.

- While benchmarking Megatron-FSDP and FSDP2 in Automodel, we noticed that the final instance of the modules specified for recompute in MCore doesn't just store its activations in preparation for the backward pass. Can we add an argument to MCore that supports this? It will improve performance across the board. Are we doing something wrong?

@NVIDIA/mcore-oncall @NVIDIA/mcore-mlperf

- Ideally, we just keep the final activations in memory as they will be used immediately in the backward pass.

N/A

**Additional context**

@Autumn1998 @ZhiyuLi-Nvidia for more context.

Glitchfix:

# What does this PR do ?
Adds an opt-in MCore flag to skip checkpointing the final full-recompute unit in each pipeline stage, retaining its forward activations because backward consumes them first.

## Issue tracking

Linked issue: Fixes #4546

## Scope

This PR intentionally scopes the optimization to `recompute_granularity='full'`. `skip_final_recompute` now raises a config error when used without full activation recomputation, so the flag cannot silently no-op for selective recompute.

### Pre-checks

## Validation

- `LD_LIBRARY_PATH=/home/shivanjanc/.pyenv/versions/3.12.9/lib/python3.12/site-packages/nvidia/cudnn/lib:$LD_LIBRARY_PATH SKIP_DOCS=true bash tools/autoformat.sh`
  - Completed successfully; the script still reports the known non-fatal mypy internal error and exits 0.
- `LD_LIBRARY_PATH=/home/shivanjanc/.pyenv/versions/3.12.9/lib/python3.12/site-packages/nvidia/cudnn/lib:$LD_LIBRARY_PATH python -m pytest tests/unit_tests/transformer/test_recompute.py -q`
  - `5 passed`
- `git diff --check`
  - clean
- CLI/config path checked with `--recompute-granularity full --recompute-method uniform --recompute-num-layers 1 --skip-final-recompute`; both parsed args and `TransformerConfig` set `skip_final_recompute=True`.

## Notes

I validated the functional recompute behavior with focused autograd tests. I did not claim or run a large-scale multi-GPU performance benchmark locally.

### DIFF
--- a/megatron/core/recompute.py
+++ b/megatron/core/recompute.py
@@ -136,6 +136,8 @@ def checkpointed_forward(
             if (start + layer_offset) in extract_layer_indices:
                 intermediate_hidden_states.append(hidden_states)
 
+    skip_final_recompute = getattr(self.config, "skip_final_recompute", False)
+
     if self.config.recompute_method == 'uniform':
         # Uniformly divide the total number of layers and checkpoint
         # the input activation of each divided chunk.
@@ -144,7 +146,10 @@ def checkpointed_forward(
             chunk_end = min(
                 layer_idx + self.config.recompute_num_layers, self.num_layers_per_pipeline_rank
             )
-            chunk_runner(layer_idx, chunk_end, True)
+            use_checkpoint = not (
+                skip_final_recompute and chunk_end == self.num_layers_per_pipeline_rank
+            )
+            chunk_runner(layer_idx, chunk_end, use_checkpoint)
             layer_idx += self.config.recompute_num_layers
     elif self.config.recompute_method == 'block':
         # Checkpoint the input activation of only a set number of individual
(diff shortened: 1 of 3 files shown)


chengcuiping:
Thanks for working on this. I can help validate the change on 4×A100, but I think the intended meaning of “final instance” in #4546 should be clarified first.

The current patch interprets it spatially on every forward invocation:

* `uniform`: bypass checkpointing for the final recompute group;
* `block`: bypass checkpointing for the final recomputed layer.

For example, with 6 layers and `recompute_num_layers=2`, the uniform path retains the full final two-layer group. Is that the intended unit, or should only the final module/layer be retained? Also, does “final instance” mean the final layer/unit, or the final outstanding forward/microbatch instance before backward?

The current 5-layer / group-size-2 unit test leaves a one-layer final group, so it does not distinguish these interpretations.

If the intended contract is confirmed, I can help provide:

* output, input-gradient, and parameter-gradient parity;
* uniform tests with a multi-layer final group;
* multiple-microbatch PP/VPP coverage;
* FP8/TE coverage;
* iteration-time and `max_memory_allocated` measurements across several microbatch counts.

This would help quantify the recompute saving versus the additional retained-activation memory.

PR #4922, opened 2026-05-21
Inference: Do not route pad/dummy tokens to any expert

sidsingh-nvidia:

# What does this PR do ?
Prevents routing of padding / dummy tokens to any of the routed experts. For small decodes, these can end up unnecessarily activating experts. This PR makes it such that only real tokens activate experts. This can help us move away from recording way too many fine grained cuda-graphs. 

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -696,6 +701,16 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
         self.is_tensor_state_allocated = False
         self.initialize_all_tensors()
 
+        # Bind the GPU real-token-count tensor onto the NVLS dispatcher class
+        # so it can mask out CUDA-graph padding tokens during routing. The
+        # tensor lives inside gpu_view._buf (fixed address) and is refreshed
+        # each step by transfer_bookkeeping_to_gpu(). NVLS-only — the NCCL
+        # dispatcher requires equal token counts across ranks already.
+        if self._nvls_dispatcher:
+            NVLSAllGatherVDispatcher.set_real_token_count_tensor(
```
lmcafee-nvidia: This only wires the real-token mask for NVLS, but the NCCL CUDA-graph path still allgathers the unmasked routing map and sets valid_tokens from the padded local token count. Equal token counts across ranks do not mean the rows are real tokens, so decode graph padding/dummy requests can still route to experts when inference_moe_token_dispatcher_type='nccl'. Can we either add equivalent masking/valid-token handling for NCCL or explicitly scope this PR as NVLS-only?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -2347,6 +2387,15 @@ def transfer_bookkeeping_to_gpu(self) -> None:
             self._staging_request_query_lengths[n_active:padded_active] = 0
             self._staging_request_kv_length_offsets[n_active:padded_active] = 0
 
+        # Real (unpadded) token count for this step. CUDA-graph replay pads
+        # the token dim to a captured size; MoE routing reads this on GPU and
+        # rewrites padding rows' routing entries to -1 so they don't go to
+        # any expert. Set to 0 on CUDA-graph capture / dummy EP steps so
+        # every row gets masked out.
+        self._staging_real_token_count[0] = (
```
lmcafee-nvidia, line 2497: I don't think this stays zero during graph construction or EP dummy forwards in the controller path. _dynamic_step_context_init() calls initialize_attention_state(...), then immediately calls context.transfer_bookkeeping_to_gpu() again with the default no_real_work=False, so the second copy overwrites gpu_view.real_token_count with batch_dimensions.token_count. That means warmup/dummy graph capture still routes the dummy tokens. Can we propagate no_real_work to the controller call, or store this as context state so the second transfer remains idempotent?

megatron/core/inference/contexts/dynamic_context.py
sidsingh-nvidia: That's a good point. The scope of this PR is NVLS-only. In theory, we can do the same optimization for the nccl dispatcher, but since it is simply a fall-back I haven't done it yet.

megatron/core/inference/contexts/dynamic_context.py
sidsingh-nvidia: I have made it a state variable instead of an arg.

megatron/core/transformer/moe/token_dispatcher_inference.py
```diff
@@ -524,6 +546,15 @@ def token_dispatch(self, hidden_states, probs):
         if self._runs_metadata_sync:
             self.update_metadata(hidden_states.shape[0])
 
+        # Mask out CUDA-graph padding rows of the local routing map so the AGV
+        # propagates -1 into agv_r for those slots; padding tokens then route
+        # to no expert. _real_token_count_tensor is wired by the context and
+        # holds the *global* unpadded token count, so we pass self.sp_rank to
+        # shift local rows into the global frame for the comparison.
+        mask_routing_padding(
```
yaox12: **[IMPORTANT Correctness]** This mask uses the context's decoder `real_token_count` for every NVLS dispatch, but serial MTP MoE forwards are request-count shaped rather than decoder-token-count shaped. With speculative decoding, `batch_dimensions.token_count` is `active_request_count * (num_speculative_tokens + 1)`, while `_compute_serial_mtp_and_sample()` pads MTP to `padded_count == req_count`; if the scalar is larger than the MTP routing map row count, `global_rows >= real_count` is never true and MTP padding rows still route to experts.

Please make the mask consume the real row count for the current MoE forward. For MTP, that should be `active_request_count`, not the decoder token count. It would also be good to add coverage with `num_speculative_tokens > 0`, a padded MTP graph count greater than active requests, and NVLS MoE in the MTP layer.

megatron/core/transformer/moe/token_dispatcher_inference.py
```diff
@@ -313,6 +314,12 @@ class NVLSAllGatherVDispatcher(InferenceAllGatherDispatcherBase):
     _step_metadata: Optional[torch.Tensor] = None  # [3] int32
     _per_rank_worst_case_token_count: int = 2048  # round_up_tokens(max_tokens) // tp_size
 
+    # [1] int32 view onto context.gpu_view.real_token_count. Fixed GPU address;
+    # written each step by the context's transfer_bookkeeping_to_gpu(). Holds the
+    # real (unpadded) local token count so the dispatcher can mask routing for
+    # CUDA-graph padding tokens. Wired once by the context after gpu_view init.
+    _real_token_count_tensor: Optional[torch.Tensor] = None
```
yaox12, line 321: **[IMPORTANT Compatibility]** `_real_token_count_tensor` is optional, but `token_dispatch()` now unconditionally passes it into `mask_routing_padding()`. Existing dispatcher-level paths, including `TestNVLSAllGatherVDispatcher`, allocate and use `NVLSAllGatherVDispatcher` directly without a `DynamicInferenceContext`, so this class field remains `None` and the dispatch path fails before the collectives.

Please either initialize a safe default in `allocate_buffers()` / `update_metadata()`, or guard the mask so standalone dispatcher use treats all rows as real. A small direct-dispatcher unit test without a context-bound tensor would cover this path.

lmcafee-nvidia (approved):

megatron/core/transformer/moe/token_dispatcher_inference.py
sidsingh-nvidia: Done.

megatron/core/transformer/moe/token_dispatcher_inference.py
sidsingh-nvidia: Fixed.

yaox12 (approved):

kvareddy:
@santhnm2 can you please take a look at this MR?

kvareddy (approved):

shanmugamr1992 (approved):

Phlip79:
@sidsingh-nvidia please fix DCO error

sidsingh-nvidia:
@Phlip79 the suggested rebase commands always end up messing the commit-tree. I am having to do this manually.

PR #4930, opened 2026-05-22
test: move NVTE_CUTEDSL_FUSED_GROUPED_MLP env into pytest fixture

ko3n1g:

## What does this PR do?

Move `NVTE_CUTEDSL_FUSED_GROUPED_MLP=1` out of the H100 unit-test CI recipe and into pytest fixtures alongside the tests that need it. Addresses review feedback on #4636: env vars belong with the test code so the source of behavior is visible from the test itself, not from a bucket-conditional export in `tests/test_utils/recipes/h100/unit-tests.yaml`.

The kernel additionally requires SM100 (Blackwell), so on H100/A100 CI this is a no-op; the fixture still wires it up so the path is picked up automatically once Blackwell hardware joins the unit-test matrix.

## Changes

- `tests/test_utils/recipes/h100/unit-tests.yaml` — drop the bucket-conditional `export NVTE_CUTEDSL_FUSED_GROUPED_MLP=1` block.
- `tests/unit_tests/transformer/moe/conftest.py` — set the env var in the existing autouse `set_env` fixture so it covers the whole `tests/unit_tests/transformer/moe/**` bucket.
- `tests/unit_tests/dist_checkpointing/models/test_moe_experts.py` — add a module-scoped autouse fixture that sets the env var and restores the previous value on teardown.

## Example

Before (recipe yaml):

```bash
if [[ "$BUCKET" == *"transformer/moe"* || "$BUCKET" == *"test_moe_experts"* ]]; then
    export NVTE_CUTEDSL_FUSED_GROUPED_MLP=1
fi
```

After (test fixture):

```python
@pytest.fixture(scope="function", autouse=True)
def set_env():
    ...
    os.environ['NVTE_CUTEDSL_FUSED_GROUPED_MLP'] = '1'
```

## Notes

- This PR targets `main` and is independent of #4636 — the env-var lines being removed exist only on the #4636 branch, so once both land, the YAML change is harmless and the fixture change becomes the canonical source.

</details>


ko3n1g:
Closing — re-opening with a clean branch off main that carries only the fixture changes.

PR #4953, opened 2026-05-23
Fix Torch FSDP2 crash: add force_all_reduce kwarg to base finish_grad_sync

### LINKED ISSUE #4669: Torch FSDP2 crashes in `finalize_model_grads` because `_BaseDataParallel.finish_grad_sync()` does not accept `force_all_reduce`
## Summary

  When running a pretraining job with Torch FSDP2 enabled, training fails during gradient finalization with:

  
  `TypeError: _BaseDataParallel.finish_grad_sync() got an unexpected keyword argument 'force_all_reduce'`

  This appears to be an interface mismatch introduced after force_all_reduce was added to the gradient finalization path.

## Analysis

  finalize_model_grads() now calls:

  model_chunk.finish_grad_sync(force_all_reduce=force_all_reduce)

  This works for DistributedDataParallel, whose method signature is:

  def finish_grad_sync(self, force_all_reduce: Optional[bool] = False):

  It also works for Megatron FSDP, which accepts the same compatibility argument.

  However, TorchFullyShardedDataParallel does not define its own finish_grad_sync() method, so it inherits the base implementation from _BaseDataParallel:

```
  def finish_grad_sync(self):
      pass
```

  As a result, any Torch FSDP2 run that reaches finalize_model_grads() receives the unexpected keyword argument and fails before the optimizer step.

## Expected behavior

  Torch FSDP2 should be able to ignore force_all_reduce if it does not need it, instead of failing due to the base class method signature.

## Suggested fix

  Update the base data-parallel interface in:

  megatron/core/distributed/data_parallel_base.py

  to accept the optional compatibility argument:

```
  diff --git a/megatron/core/distributed/data_parallel_base.py b/megatron/core/distributed/data_parallel_base.py
  index ... .. ...
  --- a/megatron/core/distributed/data_parallel_base.py
  +++ b/megatron/core/distributed/data_parallel_base.py
  @@ -1,6 +1,7 @@
   # Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.

   from contextlib import contextmanager
  +from typing import Optional

   import torch

  @@ -46,7 +47,7 @@ class _BaseDataParallel(MegatronModule):
       def scale_gradients(self, scaling_factor: float) -> None:
           """Scale all gradients inside the buffers by `scaling_factor`."""
           pass

  -    def finish_grad_sync(self):
  +    def finish_grad_sync(self, force_all_reduce: Optional[bool] = False):
           """
           Finishes grad sync (all-reduce or reduce-scatter) communication operations
           for all model gradients.
  @@ -55,6 +56,10 @@ class _BaseDataParallel(MegatronModule):
           calls to complete. When overlap_grad_reduce is set to False, calls synchronous
           communication ops.
  +
  +        Args:
  +            force_all_reduce: Optional compatibility argument used by implementations
  +                that may need to force all-reduce instead of reduce-scatter.
           """
           pass
```

  This keeps the base class interface consistent with the current call site and with existing DDP/Megatron-FSDP implementations. It also allows Torch FSDP2 to continue using the base no-op
  behavior without crashing.

factnn:
## Summary

`finalize_model_grads()` calls `model_chunk.finish_grad_sync(force_all_reduce=...)`, which works for DDP and Megatron FSDP but crashes for Torch FSDP2 with:

```
TypeError: _BaseDataParallel.finish_grad_sync() got an unexpected keyword argument 'force_all_reduce'
```

The fix adds the optional `force_all_reduce` parameter to the base class method signature, matching the interface already used by all subclass implementations (`DistributedDataParallel`, `MegatronFSDP`).

## Changes

- `megatron/core/distributed/data_parallel_base.py`: Add `force_all_reduce: Optional[bool] = False` to `finish_grad_sync()` signature and import `Optional`.

Fixes #4669


yashaswikarnati (approved):

cspades:
Darn, I don't like that Megatron-FSDP also has a no-op optional argument for this either.

@factnn Can you instead make the change here: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/distributed/torch_fully_sharded_data_parallel.py

```
    def finish_grad_sync(self, force_all_reduce=False):
        """
        Finishes grad sync (all-reduce or reduce-scatter) communication operations
        for all model gradients.

        When overlap_grad_reduce is set to True, waits for asynchronous communication
        calls to complete. When overlap_grad_reduce is set to False, calls synchronous
        communication ops.
        """
        super().finish_grad_sync()  # or pass, this function does nothing anyway
```
Will keep the base API cleaner, basically equivalent to what Megatron-FSDP does, a no-op.

Will approve immediately after.

megatron/core/distributed/data_parallel_base.py
```diff
@@ -46,7 +47,7 @@ def scale_gradients(self, scaling_factor: float) -> None:
         """Scale all gradients inside the buffers by `scaling_factor`."""
         pass
 
-    def finish_grad_sync(self):
+    def finish_grad_sync(self, force_all_reduce: Optional[bool] = False):
```
wujingyue: Why `Optional[bool]` instead of `bool`?

factnn:
Updated per your suggestion — moved the `force_all_reduce` kwarg out of `_BaseDataParallel` and into a `TorchFullyShardedDataParallel.finish_grad_sync()` override instead:

```python
def finish_grad_sync(self, force_all_reduce=False):
    super().finish_grad_sync()
```

This keeps the base API clean while still accepting the kwarg from `finalize_model_grads()` → no more `TypeError`. Equivalent to what Megatron-FSDP does (a no-op).

factnn:
@cspades Hi there, I've updated the PR per your suggestion — moved the `force_all_reduce` kwarg out of `_BaseDataParallel` and into a `TorchFullyShardedDataParallel.finish_grad_sync()` override instead. Could you take another look?

cspades (approved):

cspades:
/ok to test [efa2c61](https://github.com/NVIDIA/Megatron-LM/pull/4953/commits/efa2c61bd2df914f32c5ab685d77ecf3dcddf1fd)

factnn:
@cspades Hi there, would you mind also taking a look at #4954 when you get a chance? It's a similar small fix — just adding a missing `name` kwarg to `build_mlp()` in a test and reverting a flaky marker. Thanks\!

cspades:
@ko3n1g can help with that, I don't have approval power for that module!

ericharper (approved):

factnn:
@cspades @yashaswikarnati @ericharper Hi! This one has 3 approvals now — could it be merged? The issue is still present on main. Just rebased onto latest main. Thanks!

cspades:
Needs @NVIDIA/core-adlr approval, we'll try to raise attention after the company break. Always the last reviewer that bottlenecks things, appreciate the patience!

factnn:
> Needs @NVIDIA/core-adlr approval, we'll try to raise attention after the company break. Always the last reviewer that bottlenecks things, appreciate the patience!

 thanks for the feedback! 
 Really glad to know you're still following this PR. Enjoy the holiday! I'll be here when you get back.

factnn:
Hi maintainers! This PR has 3 approvals (from @yashaswikarnati, @cspades, @ericharper). Could someone help merge it? Thanks!

deepakn94 (approved):

cspades:
/ok to test [f351ef2](https://github.com/NVIDIA/Megatron-LM/pull/4953/commits/f351ef29e91baa9487f33a3fa1b6a21af55a8577)

cspades:
@factnn Turned on auto-merge, will check back soon, thanks for your patience and bug-fix here.

cspades:
Pretty sure some MBridge CI timed out, sending it back in.

PR #4959, opened 2026-05-23
fix(fsdp): import os in safe_get_rank fallback

### LINKED ISSUE #4958: [BUG] Megatron-FSDP safe_get_rank() fallback raises NameError because os is not imported
**Describe the bug**

@NVIDIA/mcore-oncall

```python

safe_get_rank()
```

**Expected behavior**

`safe_get_rank()` should:

- otherwise default to `0`

**Additional context**

fallintoplace:
Fixes https://github.com/NVIDIA/Megatron-LM/issues/4958.

## Summary
- add the missing `os` import in Megatron-FSDP utils
- add a focused regression test for `safe_get_rank()` env fallback

## Root cause
`safe_get_rank()` falls back to `os.environ["RANK"]` when `torch.distributed` is not initialized, but `megatron_fsdp/utils.py` did not import `os`, so that fallback path raised `NameError`.

## Validation
- direct function-level validation: loading `megatron_fsdp/utils.py` and verifying `safe_get_rank()` returns `7` when `RANK=7` and `torch.distributed.is_initialized()` is mocked to `False`
- attempted targeted pytest for `tests/unit_tests/distributed/megatron_fsdp/test_mcore_tensor_parallelism_detect.py -k safe_get_rank`
- local test harness is currently blocked by missing optional dependencies in this environment (`pytest` on the system interpreter, then `triton` during repo `conftest` import)

## Notes
- scope is intentionally minimal: one import fix plus one regression test

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/utils.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/utils.py
@@ -16,6 +16,7 @@ import contextlib
 import inspect
 import logging
 import operator
+import os
 from contextlib import nullcontext
 from functools import reduce
 from importlib.metadata import version
--- a/tests/unit_tests/distributed/mfsdp_v1/test_mcore_tensor_parallelism_detect.py
+++ b/tests/unit_tests/distributed/mfsdp_v1/test_mcore_tensor_parallelism_detect.py
@@ -7,6 +7,7 @@ from megatron.core.distributed.fsdp.mcore_fsdp_adapter import FullyShardedDataPa
 from megatron.core.distributed.fsdp.src.megatron_fsdp.utils import (
     get_mcore_tensor_parallel_partition_dim,
     is_mcore_tensor_parallel_duplicated,
+    safe_get_rank,
     using_tensor_parallel,
 )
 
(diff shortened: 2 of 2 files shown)


megatron/core/distributed/fsdp/src/megatron_fsdp/utils.py
wujingyue, line 479:
```suggestion
# If torch.distributed is not initialized, try to read environment variables.
return int(os.environ.get("RANK", 0))
```
If os.environ["RANK"] is not an integer, I'd rather raise this ValueError to the user instead of quietly assuming 0.

wujingyue (approved):
Thanks for the fix!

megatron/core/distributed/fsdp/src/megatron_fsdp/utils.py
fallintoplace: Thanks! I kept the existing fallback semantics intentionally and only fixed the missing import in this PR.

The current implementation already treats malformed `RANK` values as a safe fallback case, and the canonical helper in `megatron/core/_rank_utils.py` also catches `ValueError` / `TypeError` and returns `0`. Changing this here to raise would broaden the PR from an import fix into a behavioral change, and would make the FSDP helper inconsistent with the existing helper.

I’d prefer to keep this PR scoped to the missing import. If you want, I can open a follow-up to discuss whether both helpers should raise on invalid `RANK` instead.

wujingyue: > I kept the existing fallback semantics intentionally and only fixed the missing import in this PR.

I understood that. 

> I can open a follow-up to discuss whether both helpers should raise on invalid RANK instead.

Yes, please.

fallintoplace:
@wujingyue Do you still need anything from my side for this ticket?

wujingyue:
LGTM. I launched CI but you may have to fix DCO by signing your commits and lint.

fallintoplace:
@wujingyue I will do this in a few hours and let you know.

Phlip79 (approved):

jiemingz (approved):

fallintoplace:
Now there is a merge conflict, should I fix it and push?

ericharper:
/ok to test [f7c5a88](https://github.com/NVIDIA/Megatron-LM/pull/4959/commits/f7c5a88daf7d8cd9ea0eb234f585ee1c57316b96)

PR #4967, opened 2026-05-25
[feat] Generalized Tensor Parallelism (GTP)

fanshiqing:

# What does this PR do ?

**Generalized Tensor Parallelism (GTP)** is a light-weight, high-performance and memory-efficient distributed training strategy implemented in Megatron-LM and TransformerEngine. It shards weight tensors across an GTP process group and reconstructs them on-demand via async all-gather, enabling training of larger models without sacrificing throughput by overlapping communication with computation.

- [GTP_README.md](https://github.com/NVIDIA/Megatron-LM/blob/main/docs/api-guide/core/generalized_tensor_parallel.md)

    Outline:
    
    [Features](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#1-features)
        1.1 [Fine-grained, per-weight materialization & gradient reduction](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#11-fine-grained-per-weight-materialization--gradient-reduction)
        1.2 [CUDA graph compatibility](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#12-cuda-graph-compatibility)
        1.3 [Low-precision gather (native FP8 / NVFP4 param)](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#13-low-precision-gather-native-fp8--nvfp4-param)
        1.4 [Composability with TP / SP / EP / DDP](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#14-composability-with-tp--sp--ep--ddp)
        1.5 [Opt-in, minimally invasive integration](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#15-opt-in-minimally-invasive-integration)
        1.6 [Optimizer-agnostic (Adam + Muon)](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#16-optimizer-agnostic-adam--muon)
        1.7 [Scaling](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#17-scaling)
        1.8 [Native distributed checkpointing (DCP)](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#18-native-distributed-checkpointing-dcp)
        [Usage](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#2-usage)
        2.1 [Required flags](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#21-required-flags)
        2.2 [High-priority streams (Blackwell and later)](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#22-high-priority-streams-blackwell-and-later)
        2.3 [Minimal end-to-end example](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#23-minimal-end-to-end-example)
        2.4 [Tuning knobs](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#24-tuning-knobs)
        [Implementation details](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#3-implementation-details)
        3.1 [GTP_remat architecture (Mcore ↔ TE integration)](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#31-gtp_remat-architecture-mcore--te-integration)
        [Class hierarchy: which linears shard](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#class-hierarchy-which-linears-shard)
        3.2 [DDP buckets with (E)GTP_remat](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#32-ddp-buckets-with-egtp_remat)
        3.3 [Distributed checkpointing (DCP)](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#33-distributed-checkpointing-dcp)
        3.4 [Prefetch-chain construction and its design assumptions](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#34-prefetch-chain-construction-and-its-design-assumptions)
        [Testing](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#4-testing)

- TE related MR: [TE-3005](https://github.com/NVIDIA/TransformerEngine/pull/3005) (Merged)

## GTP's Architecture (Mcore + TE)

• TE ships self-contained: Protocol + dispatcher no-op.
• GTP lives 100% in Mcore as one protocol implementer.

image

## Changes Summary

image

## Convergence Test (07/25)

image

**Config**: 1/4 of NT3 Ultra,27Layers512Experts, 32xGB200, GTP16_EP8EGTP2_m1g32_mxfp8_CG, Adam Optimizer, forceBalance, vs. 3D parallel baseline with parallelism=TP8_EP16DP2_m1g32, ~10k steps, dataset=blend_files/1T-phase1var-moresft.json

GTP matches well with the 3d baseline in `lm loss` curve:

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


megatron/core/extensions/transformer_engine.py
```diff
@@ -893,6 +894,9 @@ def __init__(
             self.te_quant_params, torch.is_grad_enabled()
         )
 
+        if is_te_min_version("2.14.0"):
```
fanshiqing: @fanshiqing When https://github.com/NVIDIA/TransformerEngine/pull/3005 is merged, switch '2.14' to the corresponding TE version

megatron/core/extensions/transformer_engine.py
```diff
@@ -893,6 +894,9 @@ def __init__(
...
+            self.gtp_size = get_pg_size(gtp_group) if gtp_group is not None else 1
+            extra_kwargs["gtp_group"] = gtp_group if torch.distributed.is_initialized() else None
```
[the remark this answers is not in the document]
fanshiqing: That's a known issue and I will upgrade the te version at line:897 when the TE's related gtp MR is get merged.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -0,0 +1,1753 @@
...
+            else:
+                # MXFP8 scale_inv shapes (see MXFP8Quantizer.get_scale_shape):
+                #   rowwise_scale_inv:    [round_up(M, 128),     round_up(K//32, 4)]
+                #   columnwise_scale_inv: [round_up(M//32, 4),   round_up(K, 128)]
+                # GTP shards M (dim 0 of the weight), so strip to the unpadded sizes.
+                if metadata.get("rowwise_scale_inv") is not None:
+                    m_rows = round_up_to_nearest_multiple(M, 128)
+                    metadata["rowwise_scale_inv"] = metadata["rowwise_scale_inv"][:m_rows]
+                if metadata.get("columnwise_scale_inv") is not None:
+                    m_tiles = round_up_to_nearest_multiple(M // MXFP8_BLOCK_SCALING_SIZE, 4)
```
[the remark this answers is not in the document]
fanshiqing: The floor/ceil asymmetry is intentional and correct, do NOT change it.
 Each GTP line exactly mirrors its own TE quantizer: mxfp8_tensor.py and nvfp4_tensor.py

megatron/core/transformer/cuda_graphs.py
```diff
@@ -771,6 +928,51 @@ def __init__(
                 self.fp4_recipe = get_fp4_recipe(self.base_module.config)
                 _set_skip_fp8_weight_update_tensor(False)
 
+    def _register_side_stream(self, side_streams, stream):
```
jiemingz: this function is only a single line; can we get rid of it or have it absorb more functionality?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -416,6 +441,12 @@ def create_cudagraphs(cls):
                     "https://github.com/NVIDIA/TransformerEngine/blob/v2.10/transformer_engine/pytorch/utils.py#L759"  # pylint: disable=line-too-long
                 )
 
+        if any(r[0].gtp_remat for r in cls.cudagraph_record):
+            reallocate_gtp_cache_to_mempool(
```
jiemingz: This function was originally create because the mempool was in megatron but the GTP implementation was in TE, so an API was needed.
Now both are in megatron so do we still need this? Could we just ensure params are created in the cudagraph mempool in the first place?

megatron/core/transformer/cuda_graphs.py
fanshiqing: Done. refined

megatron/core/transformer/cuda_graphs.py
fanshiqing: good point. I have refactored the implementation to allocate GRAPHED buffers / quantization tensors into CG mempool at creation.

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -233,9 +303,22 @@ def __init__(
                 else:
                     assert scaling_factor == target_gradient_scaling_factor
 
-            param_layout = (
-                full_param_layout.layouts.get(buffer_key) if full_param_layout is not None else None
-            )
+            # With GTP: full_param_layout contains stray GTP entries not in this buffer,
```
deepakn94: This needs to be refactored out, since we don't want to hardcode `DistributedOptimizer` vs. `LayerwiseDistributedOptimizer` information here.

deepakn94: Update `full_param_layout` here: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/optimizer/distrib_optimizer.py#L567. I think `data_parallel_world_size` in that function should be a `Union[int, Dict[BufferKey, int]` and maybe we can also fold `expert_data_parallel_world_size` into this. Similar changes needed at https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/optimizer/layer_wise_optimizer.py#L292.

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -256,6 +339,75 @@ def __init__(
             else:
                 self.buffers.append(buffer)
 
+        # GTP-sharded params have already been RS'd over the GTP axis by GTP itself,
+        # so DDP must use the GTP-peer-excluded group here. full_param_layout is not
```
deepakn94: `full_param_layout` should actually be applied to GTP buffers.

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -491,6 +491,35 @@ def finalize_model_grads(
         pos_emb_group = parallel_state.get_position_embedding_group(check_initialized=False)
         dp_cp_group = parallel_state.get_data_parallel_group(with_context_parallel=True)
 
+    # Drain any in-flight GTP reduce-scatters on rs_stream before the DP gradient sync.
```
deepakn94: Let's re-factor this into a separate method.

megatron/core/optimizer/optimizer.py
```diff
@@ -208,25 +243,59 @@ def step_with_ready_grads(self) -> bool:
         """Step the optimizer with ready gradients, return successful."""
         return True
 
+    def _compute_grad_norm_with_gtp(self, non_gtp_grads, gtp_grads):
```
deepakn94: Is there no way to leverage the existing code that computes grad_norms across TP groups for this?

megatron/core/optimizer/__init__.py
```diff
@@ -1152,21 +1143,58 @@ def get_megatron_optimizer(
         )
         model_chunk_offset += 1
 
-    moe_param_groups, moe_buffers = _get_param_groups_and_buffers(
+    # GTP params: separate optimizer with with_gtp DP group.
```
deepakn94: This is becoming extremely clunky (there's a combinatorial explosion in the number of param_groups as we shard things differently). Can we clean this up?

deepakn94: Seems like this should become a zipped for loop at minimum.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -760,6 +885,44 @@ def __init__(
             self.fp4_enabled = self.base_module.config.fp4 is not None
             self.fp8_runtime_enabled = None
             self.fp4_runtime_enabled = None
+            self.gtp_remat = self.base_module.config.generalized_tensor_parallel_remat_size > 1
+
+            # Ensure internal warmup (inside create_fwd_graph) has >= 2 steps
+            # for GTP: 1st builds chain + tickets, 2nd exercises prefetch path.
+            if self.gtp_remat:
```
jiemingz, line 843: why break this condition out from the block below?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1092,11 +1335,16 @@ def create_bwd_graph(self):
                 out_grad.requires_grad = True
             self.static_grad_outputs.append(out_grad)
 
+        torch.cuda.synchronize()
```
jiemingz: why is this needed? Does this mirror the sync against bwd streams in the fwd pass?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1106,10 +1354,63 @@ def create_bwd_graph(self):
...
+                    self.bwd_ag_fence_event.record(egtp_graphed_ag)
+                    torch.cuda.current_stream().wait_event(self.bwd_ag_fence_event)
+
+                # Record completion AFTER AG drain + fence but BEFORE RS drain,
+                # so main_stream can trigger the next runner while RS is still
+                # in flight on rs_stream.
+                self.bwd_completion_event.record()
+
+                # Phase 2: in-graph RS drain + finalize.
+                wait_async_comms(GTPChain.GRAPHED.value, finalize_after_drain=True)
```
jiemingz, line 1322: Could we combine the `wait_async_comms` + stream sync into a new "`sync_gtp_comms_against_cur_stream(ag_sync=True, rs_sync=True, fwd=True)`" function?
Whenever we call wait_async_comms, we always want the comms to finish with respect to something (or else calling wait_async_comms doesnt actually do anything). So its a natural way to organize the code.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1106,10 +1354,63 @@ def create_bwd_graph(self):
...
+                from megatron.core.parallel_state import (
+                    get_expert_generalized_tensor_parallel_remat_group,
+                    get_expert_generalized_tensor_parallel_remat_world_size,
+                    get_generalized_tensor_parallel_remat_group,
+                )
+
+                gtp_group = get_generalized_tensor_parallel_remat_group()
+                graphed_ag = get_ag_stream(GTPChain.GRAPHED.value, gtp_group)
+                self.bwd_ag_fence_event.record(graphed_ag)
+                torch.cuda.current_stream().wait_event(self.bwd_ag_fence_event)
```
jiemingz: Is there a specific reason why we cant do `torch.cuda.current_stream.wait_stream(ag_stream) ` here?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1060,6 +1290,19 @@ def create_bwd_graph(self):
         # to 0 when activation checkpointing is used. See [interaction with recompute].
         global bwd_buffer_reuse_ref_count
 
+        # Tag cross-graph chain-tail GTP params: those whose prev_w lives in
+        # another runner's params_to_backprop. Read by TE's wgrad_reduce_scatter
+        # cascade and wait_async_comms to split the captured RS wait/add across
+        # producer and consumer graphs (avoids cross-capture cudaStreamWaitEvent
+        # on c10d Work.postEvent).
+        if self.gtp_remat:
```
jiemingz: Where is this used?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -684,10 +765,46 @@ def backward(ctx, *grads):
...
+            dense_group = parallel_state.get_generalized_tensor_parallel_remat_group()
+            expert_group = parallel_state.get_expert_generalized_tensor_parallel_remat_group()
+            params_by_group = defaultdict(list)
+            for param in runner.finalized_during_bwd_capture:
+                is_expert = not getattr(param, 'allreduce', True)
+                params_by_group[expert_group if is_expert else dense_group].append(param)
+            for group, params in params_by_group.items():
+                gtp_rs_stream = get_rs_stream(GTPChain.GRAPHED.value, group)
+                gtp_rs_stream.wait_event(runner.bwd_phase2_completion_event)
+                with torch.cuda.stream(gtp_rs_stream):
```
jiemingz, line 753: Can we move as much of this out of the replay call? For instance, can we create `params_by_group` before runtime?
Why do we need to wait against bwd_phase2_completion_event here?

megatron/core/transformer/cuda_graphs.py
fanshiqing: Yes, `wait_stream(ag_stream)` is just `wait_event(ag_stream.record_event())` with an **anonymous** event that gets destroyed right after the call.

 A captured partial graph node must point at an event that **stays alive for every future replay**, so I  keep the event as a member (`self.bwd_ag_fence_event`) and does record + wait_event by hand, then this event cannot be garbaged collected out under the graph.

```
def wait_stream(self, stream):
    self.wait_event(stream.record_event())   # record_event()->Event() : a brand-new, NAMELESS event
```

megatron/core/transformer/cuda_graphs.py
fanshiqing: It's dead code for now, let me remove it

megatron/core/transformer/cuda_graphs.py
fanshiqing: > why wait here

We need to make sure previous graph's async GTP-ReduceScatter is  truly done, then it can trigger the DDP async-reduceScatter. Otherwise, DDP might read stale main_grad data, result in Inf/NaN issue

fanshiqing: >Can we move as much of this out of the replay call? For instance, can we create params_by_group before runtime?

done

megatron/core/transformer/cuda_graphs.py
fanshiqing: honestly, I did not remember when and why we add this sync after the git commit stashing, let me experiment first to check what it's gurading for.

fanshiqing: Let's first remove it.
local UTs and middle-scale convergence test are working well.

Let's add back if later we do find we need this sync explicitly.

megatron/core/transformer/cuda_graphs.py
fanshiqing: It seems not necessary at least for the current 2-stage bwd event implementation, we don't save too much for this unified function?
I recommend first let's keep these code as they are.

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: That's a real concern.

I am working in progress to make GTP truly independent of DPxCP groups so that we can avoid touching the current DDP logics and keep the original 2 buffers(dense+moe) as it is intead of 4 which is tight coupling and also requires modifications on layerwise_optimizer.

megatron/core/transformer/cuda_graphs.py
jiemingz: IIUC, any operation placed on runner.stream, after the `self._wait_side_streams(self.bwd_side_streams)`,  is guaranteed to be ordered after the RS. Furthermore `self.bwd_phase2_completion_event.record()` is the very last operation on the `runner.stream`

So to order the hook execution after the RS couldnt we just do:
```
with torch.cuda.stream(runner.stream):
        hook()

```
which orders the hook call after the RS is synchronized back to the graph stream.
Or if we would like to use gtp_rs_stream (actually why do we need to call the hook on gtp_rs_stream anyways?):
```

gtp_rs_stream.wait_stream(runner.stream)
with torch.cuda.stream(gtp_rs_stream):
        hook()
```

would that work?

megatron/core/transformer/cuda_graphs.py
jiemingz: Wait but why would using an anonymous event not work? Yes an event would be created on the fly each replay, but that seems fine to me?
So does that mean our `_wait_side_streams` which uses `wait_stream` in the same fashion is not working? It doesnt actually order the side_streams against the current stream?

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -233,9 +268,22 @@ def __init__(
...
 
-            param_layout = (
-                full_param_layout.layouts.get(buffer_key) if full_param_layout is not None else None
-            )
+            # With GTP active, a caller-provided full_param_layout is sized for the full DP group,
+            # not the replicate group the buffer shards over, so recompute it below.
+            if full_param_layout is not None and not gtp_active:
+                param_layout = full_param_layout.layouts.get(buffer_key)
+            elif self.ddp_config.use_distributed_optimizer:
+                from ..optimizer.distrib_optimizer import DistributedOptimizer
```
fanshiqing: @fanshiqing avoid this hardcoding

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: This issue has been resolved in my latest gtp-standalone-axis refactoring, will close this comment and please review the latest code again

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -81,9 +81,32 @@ def __init__(
         # Assign all required process groups
         self.dp_group = process_group_dict['dp_group']
         self.dp_cp_group = process_group_dict['dp_cp_group']
-        self.intra_dp_cp_group = process_group_dict['intra_dp_cp_group']
         self.expt_dp_group = process_group_dict['expt_dp_group']
-        self.intra_expt_dp_group = process_group_dict['intra_expt_dp_group']
+        # DDP reduces every bucket over the GTP/EGTP-EXCLUDED replicate group (the gtp axis is
```
deepakn94: So the `self.intra_dp_cp_group` is used to compute the parameter layout as a fallback? Is `intra_dp_cp_no_gtp_group` basically the DP+CP group assuming the TP group is basically the full ETP group?

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -81,9 +81,32 @@ def __init__(
...
+        # over ARE the *_no_gtp variants. They alias the regular intra groups when GTP is off.
+        self.intra_dp_cp_group = process_group_dict.get(
+            'intra_dp_cp_no_gtp_group', process_group_dict['intra_dp_cp_group']
+        )
+        self.intra_expt_dp_group = process_group_dict.get(
+            'intra_expt_dp_no_egtp_group', process_group_dict['intra_expt_dp_group']
+        )
+        # Full cross-instance, GTP-peer-EXCLUDED groups for broadcast_params (init-time weight
+        # sync must reach all true replicas). Fall back to the full DP groups when GTP is off.
+        self.dp_cp_no_gtp_group = process_group_dict.get('dp_cp_no_gtp_group', self.dp_cp_group)
```
deepakn94: Where is this used?

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -81,9 +81,32 @@ def __init__(
...
+        self.expt_dp_no_egtp_group = process_group_dict.get(
+            'expt_dp_no_egtp_group', self.expt_dp_group
+        )
+        # GTP is "active" when the replicate groups are strictly smaller than the full DP groups.
```
deepakn94: What do you mean by "replicate groups" here?

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -81,9 +81,32 @@ def __init__(
...
+        gtp_active = (
+            self.dp_cp_no_gtp_group.size() != self.dp_cp_group.size()
+            or self.expt_dp_no_egtp_group.size() != self.expt_dp_group.size()
+        )
+        if gtp_active and self.ddp_config.average_in_collective:
+            raise NotImplementedError(
+                "Orthogonal GTP currently supports average_in_collective=False (the default); "
```
deepakn94: "Orthogonal"?

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -491,6 +536,15 @@ def finalize_model_grads(
...
         dp_cp_group = parallel_state.get_data_parallel_group(with_context_parallel=True)
 
+    # Fence the current stream against all GTP backward grad work before the DP gradient sync.
+    if (
+        config.generalized_tensor_parallel_remat_size > 1
+        or config.expert_generalized_tensor_parallel_remat_size > 1
+    ):
+        from megatron.experimental.gtp import wait_gtp_grads_on_current_stream
+
+        wait_gtp_grads_on_current_stream()
```
deepakn94: Re-name this to `wait_for_gtp_grad_reduction_on_current_stream`?

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: >So the self.intra_dp_cp_group is used to compute the parameter layout as a fallback

Yes

>Is intra_dp_cp_no_gtp_group basically the DP+CP group assuming the TP group is basically the full GTP group?

No. `intra_dp_cp_no_gtp_group` = DPxCP / GTP, nothing to do with TP.

fanshiqing: `DPxCP`: split input tokens (batch_size x seq dims) across replica group x gtp_group
 -  because GTP also need to shard the activation across GTP group, so here for simplification we just reuse the vanilla batch-distribution logic across all ranks. 

`DPxCP/GTP`: get the pure weight replica group, which is needed for DDP for following up buffer process, gradient reduce and scaling. 
here we factored out GTP because GTP handles the grdients interally (GTP-ReduceScatter per op level)

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: in `broadcast_params`

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: vanilla DPxCP: the full size of where the input tokens (=batch_size x seqlen) is split over.
This size is then split into two parts:
- gtp_size: where the input is shard by gtp_size AND weight is also shard across gtp_size
- the left DPxCP/gtp_size: the ture recplia groups, where only the input is sharded while keep the weights duplicated across `DPxCP/gtp_size` group

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: Here I mean compared to the version before this refactoring, GTP is not a standalone parallelism axis which coupled with DDP/DistOpt, for now it's independent and orthoganl to DP/CP/TP, the only exception is that it reuse the data-distribution logic of DPxCP during the data-loading pass becase I don't want to change the `gradient_accumulation_size` computation logics too.

fanshiqing: Well here should just remove `Orthogonal` which is confusing.

megatron/core/distributed/finalize_model_grads.py
fanshiqing: sounds good

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -81,9 +81,32 @@ def __init__(
...
+        # DDP reduces every bucket over the GTP/EGTP-EXCLUDED replicate group (the gtp axis is
+        # completed separately by the RS + finalize all-reduce), so the intra groups DDP shards
+        # over ARE the *_no_gtp variants. They alias the regular intra groups when GTP is off.
+        self.intra_dp_cp_group = process_group_dict.get(
+            'intra_dp_cp_no_gtp_group', process_group_dict['intra_dp_cp_group']
+        )
+        self.intra_expt_dp_group = process_group_dict.get(
+            'intra_expt_dp_no_egtp_group', process_group_dict['intra_expt_dp_group']
+        )
+        # Full cross-instance, GTP-peer-EXCLUDED groups for broadcast_params (init-time weight
```
deepakn94: I don't find "GTP-peer-EXCLUDED" that intuitive :( Maybe provide an example above with tp=2, gtp=64 and world_size=1024 (what would each of these groups' sizes be?)?

fanshiqing: okay, let me add one explicit example here

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -442,6 +442,51 @@ def _allreduce_non_tensor_model_parallel_grads(
 _allreduce_layernorm_grads = _allreduce_non_tensor_model_parallel_grads
 
 
+def _allreduce_replicated_grads_over_gtp_group(model: List[torch.nn.Module]):
+    """SUM NON-GTP (replicated) param grads over the gtp / egtp group.
+
+    DDP already reduced these params over the gtp-EXCLUDED replicate group with 1/full scaling,
```
deepakn94: I always find the wgrad scaling confusing.

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: what about the refined comment here with one concrete example:
```
        # Process-group sizes, e.g. TP2 x GTP64, world_size=1024 (PP=CP=EP=1, one distopt instance):
        #   model_size = TP x PP x CP x GTP = 2 x 64 = 128  ->  weight-replicate DP = 1024/128 = 8
        #     dp_cp_group (batch-shard, GTP-INCLUDED) = GTP x weight-replicate = 64 x 8 = 512
        #     dp_cp_no_gtp_group (weight-replicate, GTP-EXCLUDED) = 8
        #     gtp group (weight-shard axis) = 64
        #     tp group = 2
        #   Use: dp_cp_group sizes the 1/(GTP x weight-replicate) grad scaling + reduces/broadcasts
        #        the REPLICATED (non-GTP) params; dp_cp_no_gtp_group does the GTP SHARDS.
        #
        # DDP reduces each bucket over the weight-replicate (GTP-EXCLUDED) group; the GTP axis is
        # completed separately by the weight reduce-scatter, and one 1/(GTP x weight-replicate)
        # factor scales both. See gtp/README.md section 3.2 (incl. why average_in_collective=False).
        # When GTP is off, the *_no_gtp groups alias the regular DP groups.
```

megatron/core/distributed/finalize_model_grads.py
fanshiqing: Yes, at first glance it does.

The current choices in summary under the goal of **minimal changes to DDP/DistOpt + easy to maintain** is: 
- **One scaling site (in `distributed_data_parallel.py`** (a single pre-scale in the buffer), gtp-agnostic, no per-buffer cases.
- **Collectives are pure SUMs**: DP, the GTP RS, and the finalize AR all just sum; none carries a scale factor.
- **Exact, and numerically simpler** (one division vs chained per-collective divisions).

If GTP handle's its gradient scaling itself, then it will break (1) and (3).

Do you have any other proposal?

fanshiqing: > I always find the wgrad scaling confusing.

Essentially, each GTP rank takes different batch just like common DP rank, and gradient-SUM is performed by GTP itself. So just like what DDP does, gradient scaling for GTP gradients is also required

megatron/core/distributed/distributed_data_parallel.py
deepakn94: That looks pretty good. Minor suggested fixes:

```python
# Example process-group sizes (e.g., TP=2, GTP=64, world_size=1024 with PP=CP=EP=1 and single
# DistOpt instance):
#   model_size = TP x PP x CP x GTP = 2 x 64 = 128  ->  DP = 1024 / 128 = 8.
#   The model weights are replicated DP (= 8) times.
#     dp_cp_group (degree of batch sharding, includes GTP) = GTP x DP = 64 * 8 = 512.
#     dp_cp_no_gtp_group (degree of weight replication, excludes GTP) = 8.
#     gtp_group = 64.
#     tp_group = 2.
#
# Data-parallel gradient reductions for each bucket are performed over dp_cp_no_gtp_group
# (GTP-excluded group). Data-parallel gradient reductions over the GTP group are completed
# separately in the model backward pass.
#
# See Section 3.2 in `gtp/README.md` for more details (including why average_in_collective=False).
#
# When GTP is disabled, the *_no_gtp groups alias the regular DP groups.
```

deepakn94: I didn't understand this part:

```python
        #   Use: dp_cp_group sizes the 1/(GTP x weight-replicate) grad scaling + reduces/broadcasts
        #        the REPLICATED (non-GTP) params; dp_cp_no_gtp_group does the GTP SHARDS.
```

megatron/core/model_parallel_config.py
```diff
@@ -20,6 +20,15 @@ class ModelParallelConfig:
     tensor_model_parallel_size: int = 1
     """Intra-layer model parallelism. Splits tensors across GPU ranks."""
 
+    generalized_tensor_parallel_remat_size: int = 1
```
deepakn94: Align argument names according to this discussion: https://nvidia.slack.com/archives/C0B2CHTGZ54/p1780897811626509.

megatron/core/model_parallel_config.py
```diff
@@ -20,6 +20,15 @@ class ModelParallelConfig:
...
+    """Generalized tensor parallelism with weight rematerialization. Shards model weights
+       across GPU ranks along ``out_features``; each weight is rematerialized independently
+       (per-weight, not per-layer) via async all-gather on every forward AND backward pass.
+       Carved out of the data-parallel axis, so increasing this size shrinks per-rank weight
```
deepakn94: Do we need this?

megatron/core/tensor_parallel/layers.py
```diff
@@ -281,6 +307,17 @@ def __init__(
                     tensor=self.weight, is_parallel=True, dim=0, stride=1
                 )
 
+        self.gtp_size = 1
+        if gtp_group is not None and gtp_group.size() > 1:
+            from megatron.experimental.gtp import wrap_module_params_gtp
```
deepakn94: cc @jaredcasper.

This is not going to fly...`megatron/core` can't import stuff not in `megatron/core` since it needs to be self-contained.

megatron/core/optimizer/__init__.py
```diff
@@ -1180,7 +1207,7 @@ def get_megatron_optimizer(
                 param_groups=moe_param_groups,
                 per_model_buffers=moe_buffers,
                 model_parallel_group=expt_tp_pp_group,
```
deepakn94: I think this needs to be updated to `expt_tp_pp_with_egtp_group`.

megatron/core/optimizer/__init__.py
```diff
@@ -1036,13 +1036,32 @@ def get_megatron_optimizer(
...
+    # the *_no_gtp (gtp/egtp-EXCLUDED) replicate group — which aliases the full DP group when GTP
+    # is inactive. GTP is "active" when that group is strictly smaller (no Gloo state path then).
+    # A None group means the axis is unused (e.g. no expert parallelism) → not active.
+    def _gtp_active_for(sub, full):
+        return sub is not None and full is not None and sub.size() != full.size()
+
+    gtp_active = _gtp_active_for(intra_dp_cp_no_gtp_group, intra_dp_cp_group) or _gtp_active_for(
+        intra_expt_dp_no_egtp_group, intra_expt_dp_group
+    )
+    main_dp_group = intra_dp_cp_no_gtp_group
```
deepakn94: Why the `main_` prefix here?

megatron/core/optimizer/clip_grads.py
```diff
@@ -218,6 +218,8 @@ def count_zeros_fp32(
         use_decoupled_grad (bool, optional): If True, reads from the
             '.decoupled_grad' attribute instead of the standard '.grad'.
             Defaults to False.
+        tp_group (ProcessGroup, optional): TP group for the TP-duplicate filter.
```
deepakn94: Delete? Not used?

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: Here I want to explain what's the use case of `dp_cp_group` and `dp_cp_no_gtp_group`

fanshiqing: `dp_cp_group` is used when perfrom the grad scaling over DPxCPxGTP groups

megatron/core/model_parallel_config.py
fanshiqing: Right, will updating all of these arguments in the last commit

megatron/core/tensor_parallel/layers.py
fanshiqing: Oh god... then the current placement of gtp which under `megatron/experimental/gtp` does break this requirement :(

megatron/core/optimizer/clip_grads.py
fanshiqing: No, it's used in https://github.com/NVIDIA/Megatron-LM/pull/4967/changes#diff-072e3c3a7be90072dc0c7b29725fb4d6d79dd52c932f79afefb9677e07b68861R251

megatron/core/optimizer/__init__.py
fanshiqing: the `main_` prefix marks the `effective` DP shard group passed to the optimizer, distinguishing if from the other DP-ish groups that are not used for state sharding.

fanshiqing: you are right to find it confusing...
what about these candidates:
`main_dp_group` --> `dense_optim_dp_shard_group`
`main_dp_group_gloo` --> `dense_optim_dp_group_gloo`
`main_expt_dp_group` --> `expert_optim_dp_shard_group`

megatron/core/model_parallel_config.py
fanshiqing: outdated line, will refine

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -442,6 +442,49 @@ def _allreduce_non_tensor_model_parallel_grads(
...
+        if not grads or group is None or group.size() <= 1:
+            continue
+        coalesced = _flatten_dense_tensors(grads)
+        torch.distributed.all_reduce(coalesced, op=torch.distributed.ReduceOp.SUM, group=group)
+        for param, buf, synced in zip(params, grads, _unflatten_dense_tensors(coalesced, grads)):
+            buf.copy_(synced)
+            grad_attr = _get_main_grad_attr(param)
+            orig_grad = getattr(param, grad_attr)
+            setattr(param, grad_attr, _reshard_if_dtensor(buf, orig_grad))
+
```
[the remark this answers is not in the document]
fanshiqing: no bug, no change needed. The concern assumes `_allreduce_non_tensor_model_parallel_grads` does a DP reduction, it doesn't, it reduces over `tp_group`

megatron/core/tensor_parallel/layers.py
```diff
@@ -101,6 +103,29 @@ def param_is_not_tensor_parallel_duplicate(param, tp_group=None):
...
+    rank 0 of the gtp/egtp axis (else counted gtp/egtp times). When GTP is off the rank is 0,
+    so every param is kept.
+    """
+    if getattr(param, "is_gtp", False):
+        return True
+    is_expert = not getattr(param, "allreduce", True)
+    if is_expert:
+        return get_expert_gtp_weight_remat_rank() == 0
+    return get_gtp_weight_remat_rank() == 0
+
```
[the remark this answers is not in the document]
fanshiqing: No, the current name is fine: 
-  "GTP" is the established umbrella term in this codebase,  it subsumes EGTP.
-  The docstring already documents both axes (lines 115-119: "counted once across the GTP/EGTP axis", "GTP/EGTP shards"
-  Symmetry with `param_is_not_tensor_parallel_duplicate`

So I prefer leave the name as-is

megatron/core/parallel_state.py
```diff
@@ -1455,6 +1698,42 @@ def get_tensor_model_parallel_group(check_initialized=True):
...
+        return 0
+
+
+def get_gtp_weight_remat_rank():
+    """Return caller's rank in the parameter-sharding group."""
+    if torch.distributed.is_available() and torch.distributed.is_initialized():
+        group = get_gtp_weight_remat_group(check_initialized=False)
+        return group.rank() if group is not None else 0
+    else:
+        return 0
```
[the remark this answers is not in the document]
fanshiqing: That's fine, it matches all of these `get_xx_rank()` in this file, just keep it as-is.

megatron/training/utils/common_utils.py
```diff
@@ -71,129 +100,110 @@ def calc_params_l2_norm(model, force_create_fp32_copy=False):
...
+    sharded_params_data = []        # Dense, sharded → reduce over dp_cp
+    gtp_params_data = []            # GTP, non-sharded
+    gtp_sharded_params_data = []    # GTP, sharded → reduce over dp_cp_no_gtp
+    moe_params_data = []            # MoE, non-sharded
+    moe_sharded_params_data = []    # MoE, sharded → reduce over expert_dp
+    moe_gtp_params_data = []        # MoE-GTP, non-sharded
+    moe_gtp_sharded_params_data = []  # MoE-GTP, sharded → reduce over expert_dp_no_gtp
+
+    gtp_rank = mpu.get_gtp_weight_remat_rank()
+    egtp_rank = mpu.get_expert_gtp_weight_remat_rank()
```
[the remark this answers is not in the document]
fanshiqing: No code change is needed, and the suggested gating would actually introduce a deadlock.
- Point 1 (rank calls at 114-115): safe by design, no real masking risk
- Point 2 (line 191): the group is always initialized, so it can't crash and gating it would be a bug

megatron/training/utils/common_utils.py
```diff
@@ -71,129 +100,110 @@ def calc_params_l2_norm(model, force_create_fp32_copy=False):
...
+    _sum_reduce(moe_gtp_sharded_norm_2, mpu.get_expert_data_parallel_group(no_gtp=True))
 
-    # Account for MoE norm even if current rank doesn't have any expert params to prevent
-    # hang in models with un-even numbers of MoE layers.
-    # See details in https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/issues/409
-    else:
-        moe_norm_2 = torch.zeros_like(norm_2)
+    # --- Combine dense + GTP norms ---
+    # model_parallel group = TP×GTP×PP, so GTP reduction is implicit.
+    norm_2 = params_norm_2 + sharded_norm_2 + gtp_norm_2 + gtp_sharded_norm_2
```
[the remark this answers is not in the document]
fanshiqing: The invariant always holds a silent GTP under-count is structurally impossible. Three layers guarantee it:

1. GTP is injected into the order programmatically, not user-controlled (parallel_state.py:859):
decoder_order = _inject_gtp(order, after="tp")   # inserts "gtp" right after "tp"
1. The user passes the standard order (e.g. 'tp-cp-ep-dp-pp', no gtp); _inject_gtp always adds it. So "someone initializes without GTP in the order string" can't happen they don't own that string for the GTP axis.
2. The model-parallel group query token is hardcoded (parallel_state.py:1151):
for ranks in decoder_rank_generator.get_ranks('tp-gtp-pp'):   # GTP explicit
2. Not derived from the user order it literally names gtp.
3. Backstop: RankGenerator fails loudly on mismatch (__init__, lines 43-47): if gtp > 1 but 'gtp' somehow isn't in the order, it raises ("you haven't specified the order"). So the only failure mode is a construction-time crash, never a silent under-count. (When gtp == 1, gtp is a size-1 identity dim auto-appended harmlessly and there are no GTP params to under-count anyway.)

So get_model_parallel_group() provably spans TPxGTPxPP, and adding gtp_norm_2/gtp_sharded_norm_2 into norm_2 without a separate GTP reduction (common_utils.py:185) is correct.

No changes needed, just keep as-is.

megatron/core/process_groups_config.py
```diff
@@ -133,6 +148,10 @@ class ProcessGroupCollection:
     # _INTRA_EXPERT_DATA_PARALLEL_GROUP
     intra_expt_dp: torch.distributed.ProcessGroup = field(init=False)
 
+    # _EXPERT_DATA_PARALLEL_GROUP_NO_GTP — expert DP excluding EGTP peers (true expert
+    # weight replicas). Identical to ``expt_dp`` when EGTP=1.
+    intra_expt_dp_no_egtp: torch.distributed.ProcessGroup = field(init=False)
+
     # _INTER_PARTIAL_EXPERT_DATA_PARALLEL_GROUP
```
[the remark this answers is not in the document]
fanshiqing: The current naming is fine, just keep it as-is.

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -446,6 +442,49 @@ def _allreduce_non_tensor_model_parallel_grads(
 _allreduce_layernorm_grads = _allreduce_non_tensor_model_parallel_grads
 
 
+def _allreduce_replicated_grads_over_gtp_group(model: List[torch.nn.Module]):
+    """SUM NON-GTP (replicated) param grads over the gtp / egtp group.
```
deepakn94: "Sum wgrads for replicated parameters over the gtp / egtp group."

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -446,6 +442,49 @@ def _allreduce_non_tensor_model_parallel_grads(
...
+
+    DDP already reduced these params over the gtp-EXCLUDED replicate group with 1/full scaling,
```
deepakn94: "The data-parallel collective already reduced wgrads over the GTP-excluded process groups with 1/full scaling".

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -446,6 +442,49 @@ def _allreduce_non_tensor_model_parallel_grads(
...
+    leaving them 1/gtp short. SUM (not AVG) over the gtp/egtp group recovers the full mean.
```
deepakn94: "leaving them 1/gtp short" is confusing. I think.

megatron/core/transformer/cuda_graphs.py
fanshiqing: - Your first option works but serializes the RS onto the graph stream -> loses cross-graph overlap.
- Your second option is equivalent to today's code I think. wait_event(phase2_event) is just the more precise

megatron/core/tensor_parallel/layers.py
ericharper: I don't think there's anything experimental about these changes. This is sort of all or nothing IMO.

deepakn94 (approved):

megatron/core/transformer/cuda_graphs.py
jiemingz: In that case, would it be cleaner to remove `phase2_event` entirely? And replace it with the above code?
and `wait_for_gtp_grad_reduction_on_current_stream` could just be:
```
for runner in _CudagraphGlobalRecord.cudagraph_record:
        torch.cuda.current_stream.wait_stream(runner.stream)
```

fanshiqing: No, that's not right.

```
wait_for_gtp_grad_reduction_on_current_stream could just be:

for runner in _CudagraphGlobalRecord.cudagraph_record:
        torch.cuda.current_stream.wait_stream(runner.stream)
```
The above code only covers the GRAPHED weights. 
For ungraphed weights, waiting for RS_streams is still necessary.

fanshiqing: And  `cudagraph_record` is **empty** during normal training, so iterating it would silently do nothing. So even though if you remove `phase2_event`, you still need to keep a list of all graph's runner_stream for `wait_for_gtp_grad_reduction_on_current_stream`'s usage

which did not make the code looks cleaner right?

fanshiqing: ok finally I removed the `bwd_phase2_completion_event` and use `runner_stream` instead to make the code look a bit cleaner~~

fanshiqing:
@ericharper @jaredcasper Hey Eric and Jared, can you help talk a look for this MR? Let me know if you have any concern for this MR so that I can change accordingly~ 

Any comments are welcome!

jiemingz (approved):

megatron/core/tensor_parallel/layers.py
```diff
@@ -1191,6 +1282,7 @@ def __init__(
         tp_comm_buffer_name: str | None = None,  # Not used
         tp_group: Optional[torch.distributed.ProcessGroup] = None,
```
kvareddy, line 1289: make tp_group a list of [tp_group, gtp_remat_group].

kvareddy (approved):

megatron/core/tensor_parallel/layers.py
fanshiqing: Finally I take a more elegant solution **without changing/adding** existing APIs for upper-level classes like {MLP/Attention/Mamba/SharedExpert/Experts}, now only the Megatron's `transformer_engine.py` decides per linear class whether to GTP_remat-shard, 

- detailed explained [here](https://github.com/fanshiqing/Megatron-LM/blob/gtp_release/docs/api-guide/core/generalized_tensor_parallel.md#class-hierarchy-which-linears-shard), 
- related [commit](https://github.com/NVIDIA/Megatron-LM/pull/4967/changes/78efe54847d7eb94988a9a8417bd1a86071c3e27).

megatron/core/utils.py
```diff
@@ -539,6 +539,17 @@ def get_tensor_model_parallel_group_if_none(tp_group, is_expert=False, check_ini
...
 
 
+def get_gtp_weight_remat_group(is_expert=False):
+    """Return the active GTP weight-remat group from parallel_state, or None if inactive."""
+    if not (torch.distributed.is_available() and torch.distributed.is_initialized()):
+        return None
+    if not parallel_state.is_initialized():
+        return None
+    if is_expert:
+        return parallel_state.get_expert_gtp_weight_remat_group(check_initialized=False)
```
deepakn94: I think we try to refrain from using `parallel_state` directly :/ See the above method with the deprecation notice.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -0,0 +1,1941 @@
...
+                slot.reduce_scatter,
+                slot.fwd,
+            )
+
+        return slot.buf
+
+    def release(self, ticket: int):
+        """Return the buffer to the pool (ticket stays valid).
+
+        slot.buf is intentionally NOT cleared: get() must stay idempotent so CUDA-graph-captured
```
Victarry, line 1788: **[CRITICAL Correctness]** release() publishes slot.buf back to the pool while deliberately retaining the same object in the persistent ticket. A later same-key ticket can therefore pop that buffer while the old ticket still returns it, so multiple weights permanently alias one full-weight storage.

This becomes a real read/write race for adjacent same-key weights: all_gather_and_prefetch() issues the next weight's async AG before TransformerEngine launches the current weight's GEMM. The next AG can overwrite the shared output buffer while the current GEMM is reading it. The existing two-same-shape second-pass test reaches this path but only checks isfinite, which cannot detect finite outputs computed with the wrong weight.

Please add exclusive buffer liveness (or a capture-stable double/ring buffer with a consumer-done event) and a numerical parity test that gives the adjacent weights deliberately different values.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -0,0 +1,1941 @@
...
+
+    shard_size = tensor.shape[0] // gtp_remat_size
+    shard = tensor[gtp_rank * shard_size : (gtp_rank + 1) * shard_size]
+    gtp_shard = GTPShardedParam(shard.clone())
+    gtp_shard.pad_length = pad_length
+    # Preserve the source weight's TP attributes (dropped when wrapping into GTPShardedParam),
+    # so param_is_not_tensor_parallel_duplicate still classifies it without GTP-specific code.
+    from megatron.core.tensor_parallel import copy_tensor_model_parallel_attributes
+
+    copy_tensor_model_parallel_attributes(gtp_shard, param)
```
Victarry, line 442: **[CRITICAL Distributed Correctness]** Native ColumnParallelLinear / RowParallelLinear set weight.allreduce=False for expert-parallel weights before calling wrap_module_params_gtp(), but this replacement only copies _MODEL_PARALLEL_ATTRIBUTE_DEFAULTS; that list does not include allreduce. The new GTPShardedParam consequently defaults to allreduce=True.

For native SequentialMLP with EGTP, DDP and _get_param_groups() then classify the expert shard as dense and select the dense DDP/optimizer/broadcast groups instead of the expert groups. Please preserve allreduce (and any other expert ownership metadata) when replacing the parameter, and add a native SequentialMLP + EGTP test; the current MoE coverage uses TE grouped experts and does not exercise this path.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -760,6 +872,33 @@ def __init__(
             self.fp4_enabled = self.base_module.config.fp4 is not None
             self.fp8_runtime_enabled = None
             self.fp4_runtime_enabled = None
+            self.gtp_remat = self.base_module.config.gtp_weight_remat_size > 1
```
Victarry, line 841: **[CRITICAL CUDA Graph]** This gate only checks dense gtp_weight_remat_size. In the supported dense GTP=1, expert EGTP>1 configuration, a captured MoE module still issues EGTP AG/RS, but runner.gtp_remat remains false, so the runner skips EGTP side-stream registration/drains and never builds the post-replay GTP finalize-hook plan.

The missing hook path is deterministic with overlap-grad-reduce: GTP parameters use a manually driven DDP hook, that hook returns during graph capture, and the only post-replay invocation is also guarded by runner.gtp_remat. EGTP parameters are therefore never registered grad-ready. Please include expert_gtp_weight_remat_size > 1 in this decision (or detect GTP params owned by the runner) and add an EGTP-only MoE CUDA-graph test.

megatron/core/optimizer/__init__.py
```diff
@@ -1047,13 +1049,33 @@ def get_megatron_optimizer(
 
     dp_cp_group = process_groups_dict['dp_cp_group']
     intra_dp_cp_group = process_groups_dict['intra_dp_cp_group']
-    intra_expt_dp_group = process_groups_dict['intra_expt_dp_group']
+    intra_dp_cp_no_gtp_remat_group = process_groups_dict['intra_dp_cp_no_gtp_remat_group']
+    intra_expt_dp_no_egtp_remat_group = process_groups_dict['intra_expt_dp_no_egtp_remat_group']
     mp_group = process_groups_dict['mp_group']
     expt_tp_pp_group = process_groups_dict['expt_tp_pp_group']
+    expt_tp_pp_with_egtp_remat_group = process_groups_dict['expt_tp_pp_with_egtp_remat_group']
```
Victarry, line 1067: **[CRITICAL Numerical Correctness]** The new EGTP-merged model-parallel group is used by the standard Adam/SGD expert path below, but _get_megatron_emerging_optimizer() returns before reaching this code and still assigns pg_collection.tp_ep_pp to non-layer-wise expert Muon/AdaptiveMuon groups. That group excludes the EGTP axis.

As a result, each EGTP rank reduces gradient statistics for only its local expert-weight shard. When clipping (or a grad-norm skip threshold) is enabled, peers can compute different partial norms and apply different clipping coefficients instead of one norm over all EGTP shards. Please use tp_ep_pp_with_egtp_remat for expert emerging-optimizer grad statistics as well, and add a non-layer-wise Muon + EGTP clipping test.

Victarry (approved):
LGTM for the changes in MoE.

megatron/core/utils.py
fanshiqing: Resolved in the latest commit: https://github.com/NVIDIA/Megatron-LM/pull/4967/changes/a75a6af335dc9b547194b48e9e52a482e111394b

yuzhongw-nvidia (approved):

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -0,0 +1,2109 @@
...
+    wait_async_comms()
+    cur = torch.cuda.current_stream()
+    for s in _AG_STREAMS.values():
+        cur.wait_stream(s)
+    for s in _RS_STREAMS.values():
+        cur.wait_stream(s)
+    # Local import: cuda_graphs imports this module, so a module-level import would be circular.
+    from megatron.core.transformer.cuda_graphs import get_gtp_runner_streams
+
+    for s in get_gtp_runner_streams():
```
jiemingz, line 298: [we already have an event that synchronizes the backward graph with the async wgrad reduce scatter here
](https://github.com/fanshiqing/Megatron-LM/blob/1937dd015f5da39fa7e72b51da1d6f4db1fe6130/megatron/core/distributed/param_and_grad_buffer.py#L613)

Could we use that preexisting infrastructure instead? So instead of waiting on cudagraph runner streams we just have to make sure we properly record `_cudagraph_wgrad_ready_event`

megatron/core/transformer/cuda_graphs.py
```diff
@@ -666,6 +745,20 @@ def backward(ctx, *grads):
...
+        # rs_stream, after wait_stream(runner.stream) fences Phase 2) to let DDP RS overlap bwd.
+        if runner.gtp_remat:
+            for gtp_rs_stream, params in runner._gtp_finalize_hook_plan:
+                gtp_rs_stream.wait_stream(runner.stream)
+                with torch.cuda.stream(gtp_rs_stream):
+                    for param in params:
+                        param.grad = None
+                        if hasattr(param, 'grad_added_to_main_grad'):
+                            param.grad_added_to_main_grad = True
+                        hook = getattr(param, '_grad_accum_hook', None)
```
jiemingz, line 758: could we set the param's `_cudagraph_wgrad_ready_event` on the param here? in param_and_grad_buffer.py it will wait on this event before copying the main_grad.

megatron/training/utils/common_utils.py
```diff
@@ -48,6 +48,35 @@
 from megatron.training import get_adlr_autoresume, get_args, get_timers
 
 
+
+def _compute_norm_2(params_list):
+    """Compute squared L2 norm of a list of tensors. Returns a CUDA scalar."""
+    if len(params_list) > 0:
+        dummy_overflow_buf = torch.tensor([0], dtype=torch.int, device='cuda')
```
asolergi-nv, line 55:
```suggestion
        dummy_overflow_buf = torch.tensor([0], dtype=torch.int32, device='cuda')
```

shanmugamr1992 (approved):

megatron/training/training.py
```diff
@@ -2458,7 +2484,12 @@ def _save_state_dict(attr_name, label):
 
     # Update learning rate.
     if update_successful:
-        increment = get_num_microbatches() * args.micro_batch_size * args.data_parallel_size
+        increment = (
+            get_num_microbatches()
+            * args.micro_batch_size
+            * args.data_parallel_size
```
asolergi-nv, line 2509: https://github.com/fanshiqing/Megatron-LM/blob/1937dd015f5da39fa7e72b51da1d6f4db1fe6130/megatron/training/global_vars.py#L132

Doesn't `args.data_parallel_size` already `* args.gtp_weight_remat_size`? Should we leave a comment?

megatron/training/training.py
```diff
@@ -2590,7 +2621,12 @@ def training_log(
         timers_to_log.extend(RL_LOGGABLE_TIMER_NAMES)
 
     # Calculate batch size.
-    batch_size = args.micro_batch_size * args.data_parallel_size * get_num_microbatches()
+    batch_size = (
+        args.micro_batch_size
+        * args.data_parallel_size
```
asolergi-nv, line 2647: Same

megatron/training/training.py
```diff
@@ -4082,7 +4104,9 @@ def evaluate(
     # make validation batch size independent from training batch size
     eval_batch_size = args.eval_global_batch_size
     eval_micro_batch_size = args.eval_micro_batch_size
-    eval_num_microbatches = eval_batch_size // (eval_micro_batch_size * args.data_parallel_size)
+    eval_num_microbatches = eval_batch_size // (
+        eval_micro_batch_size * args.data_parallel_size * args.gtp_weight_remat_size
```
asolergi-nv, line 4180: Same

megatron/training/training.py
fanshiqing: No, `data_parallel_size` excludes the GTP-remat axis (it's folded into total_model_size at `arguments.py:446`; each gtp-remat peer consumes a distinct microbatch, so multiply it back in for the sample count, will add comments here.

megatron/training/utils/common_utils.py
fanshiqing: This is not necessary as `torch.int is torch.int32` is True

asolergi-nv (approved):

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
fanshiqing: Yeah I think so, but I would prefer to add these changes into the existing parallel MR (https://github.com/fanshiqing/Megatron-LM/pull/9) and keep the current changes as-is

megatron/core/transformer/cuda_graphs.py
fanshiqing: Same as above, we should cover this too in the following up https://github.com/fanshiqing/Megatron-LM/pull/9 to make the gtp related TE changes looks cleaner

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -230,6 +230,41 @@ def scaled_orthogonalize_fn(
...
 
+    def scaled_orthogonalize_fn_with_gtp_remat(self, p, grad, tp_group, partition_dim):
+        """All-gather grad along GTP_remat/EGTP_remat dim 0, orthogonalize, then slice back.
+
+        GTP_remat shards weights along dim 0 independently of TP's partition_dim. Newton-Schulz
+        needs the full weight matrix, so we reconstruct the GTP_remat dimension before running
+        the TP-aware orthogonalization, then extract the local GTP_remat shard from the result.
+        When GTP_remat is inactive this is a plain passthrough to scaled_orthogonalize_fn.
+        """
+        is_expert = getattr(p, 'expert_tp', False)
```
mkhona-nvidia, line 242: @deepakn94 @philipcmonk 

Flagging this occurance if and when the is_expert functionality is removed or changed

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -230,6 +230,41 @@ def scaled_orthogonalize_fn(
...
+        gtp_remat_group = (
```
mkhona-nvidia, line 243: @deepakn94 @kvareddy 

This path is going through a similar if/else structure like the regular MoE path in the rest of megatron-lm, which might lead to bugs in the future. Flagging here, the logic is OK but long term we should unify the API and make it more robust rather than have this if/else structure

Could we add a TODO here?

mkhona-nvidia (changes requested):
Please leave TODOs in the 2 places where a comment has been added

megatron/core/optimizer/emerging_optimizers.py
philipcmonk: `expert_tp` is a parallel way of determining whether something is an expert that's different from `allreduce=False`.  It's specific to emerging optimizers and it [inspects the name](https://github.com/NVIDIA/Megatron-LM/blob/ff603601630797924ae174a7057cf1e57405ed2a/megatron/core/optimizer/__init__.py#L783-L784).

If it's only used to choose reduction groups (I think it is?), then it should probably switch to using `allreduce=False`, but if it's used to choose which parameters are subject to Muon, then keeping it separate is reasonable.

This function appears to be a direct translation of [this function](https://github.com/NVIDIA/Megatron-LM/blob/ff603601630797924ae174a7057cf1e57405ed2a/megatron/core/optimizer/emerging_optimizers.py#L246-L251) which uses `expert_tp` in the same way.  So for this PR it should probably stay as it is, and when we clean this up later we can change all uses of `expert_tp` at the same time.  #5916 does not touch expert_tp, so there won't be a conflict.

megatron/core/optimizer/emerging_optimizers.py
deepakn94: @fanshiqing , let's add a TODO before these two lines for now like "TODO: Clean up code that determines if parameter is a MoE layer and which TP group to use."

yashaswikarnati (approved):

ericharper (approved):

mkhona-nvidia (approved):

PR #4976, opened 2026-05-26
Add experimental Megatron-FSDP fully_shard implementation

wujingyue:
Stacked on top of #4835.

# What does this PR do ?
Adds an experimental per-module Megatron-FSDP `fully_shard` path that uses DBuffer primitives to shard parameters, materialize full weights for compute, and reduce gradients back into sharded optimizer state.

Meta-parameter materialization is intentionally split out to the follow-up draft PR at https://github.com/NVIDIA/Megatron-LM/pull/5369.

## Issue tracking

Linked issue: N/A

## Summary
- add the experimental `fully_shard(...)` entry point plus `FsdpModule`, `ParameterGroup`, and `Placements` runtime state
- group parameters by dtype and `requires_grad`, manage sharded main/model weight buffers, and sync optimizer-updated weights into compute weights
- allocate model/gradient buffers from `DBuffer.device` instead of reaching through `local_buffer.device`
- install forward/backward hooks for unshard, reshard, and parameter-completion-based gradient reduction, including accumulation across backward calls
- extend distributed coverage for DBuffer layout/redistribution and the experimental FSDP path, including nested ownership, frozen parameters, optimizer-step visibility, CPU-initialized parameter sharding, autograd storage reuse, loss parity, and peak-memory reduction

## Testing
- `BASE_REF=pull-request/4835 CHECK_ONLY=true bash tools/autoformat.sh`
- `git diff --check origin/pull-request/4835..HEAD`

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
@@ -116,6 +116,20 @@ class DBuffer:
         """Device of the local buffer."""
         return self.local_buffer.device
 
+    def reallocate_storage(self) -> None:
+        """Restore the local buffer's backing storage to its logical size."""
+        self._resize_storage(self.local_buffer.numel())
+
+    def release_storage(self) -> None:
+        """Release local buffer storage without replacing the Storage object."""
+        # Autograd may save views that share this Storage object. Resizing the
+        # existing Storage releases the allocation while preserving those aliases
+        # for a later reallocate_storage().
+        self._resize_storage(0)
+
+    def _resize_storage(self, numel: int) -> None:
+        self.local_buffer.untyped_storage().resize_(numel * self.local_buffer.element_size())
+
     def _get_owned_range(self, tensor_index: int) -> _OwnedRange | None:
         """Return this buffer's owned range for logical tensor ``tensor_index``."""
         tensor_start = self.layout.tensor_to_offset[tensor_index]
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
@@ -268,6 +268,21 @@ class DBuffer:
             raise ValueError(f"Expected out device {self.device}, got {out.device}.")
         return out
 
+    def cast(self, dtype: torch.dtype) -> "DBuffer":
+        """Return this buffer with the same layout and placements in ``dtype``."""
+        if self.dtype == dtype:
+            return self
+
+        destination = DBuffer(
+            mesh=self.mesh,
+            placements=self.placements,
+            tensor_shapes=self.layout.tensor_shapes,
+            dtype=dtype,
+            device=self.device,
+        )
+        destination.local_buffer.copy_(self.local_buffer)
+        return destination
+
     def redistribute(
         self, new_placements: Iterable[Placement], *, out: "DBuffer | None" = None
     ) -> "DBuffer":
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
@@ -42,6 +42,7 @@ def _validate_mesh_axis(mesh: DeviceMesh, axis: int) -> None:
 
 
 def _validate_placements(placements: Iterable[Placement]) -> None:
+    """Validate DBuffer placements form a supported contiguous local layout."""
     seen_flat = False
     for placement in placements:
         if not isinstance(placement, (Replicate, Partial, Flat)):
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fsdp_module.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fsdp_module.py
@@ -31,13 +31,13 @@ class FsdpModule:
 
     _parameter_groups: tuple[ParameterGroup, ...]
     _ready_grad_parameters: set[nn.Parameter]
-    num_training_parameters: int
+    _num_training_parameters: int
 
     def __init__(
         self, mesh: DeviceMesh, placements: Placements, mixed_precision_policy: MixedPrecisionPolicy
     ) -> None:
         """Initialize FSDP runtime state on an already-constructed module."""
-        owned_parameters = _materialize_and_collect_owned_parameters(self, _mesh_device(mesh))
+        owned_parameters = _collect_owned_parameters(self)
         axis_indices = tuple(_axis_index(mesh, axis) for axis in placements.dp_axes)
         assert axis_indices == tuple(
             range(mesh.ndim)
@@ -54,7 +54,7 @@ class FsdpModule:
         ]
         self._parameter_groups = tuple(parameter_groups)
         self._ready_grad_parameters = set()
-        self.num_training_parameters = sum(
+        self._num_training_parameters = sum(
             len(group.sharded_parameters) for group in self._parameter_groups if group.requires_grad
         )
         self._register_hooks()
@@ -77,7 +77,7 @@ class FsdpModule:
     def _make_grad_hook(self, parameter: nn.Parameter) -> Callable[[nn.Parameter], None]:
         def grad_hook(_parameter: nn.Parameter) -> None:
             self._ready_grad_parameters.add(parameter)
-            if len(self._ready_grad_parameters) == self.num_training_parameters:
+            if len(self._ready_grad_parameters) == self._num_training_parameters:
                 self.post_backward()
 
         return grad_hook
@@ -86,6 +86,7 @@ class FsdpModule:
         """Prepare full parameters for forward compute."""
         self._ready_grad_parameters.clear()
         for group in self._parameter_groups:
+            group.sync_model_weight_from_main_weight()
             group.unshard_parameters()
 
     def post_forward(self) -> None:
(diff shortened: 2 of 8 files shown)


tests/unit_tests/distributed/megatron_fsdp/test_experimental_fully_shard.py
```diff
@@ -0,0 +1,388 @@
...
+
+    fully_shard(
+        model,
+        mesh=mesh,
+        placements=_flat_placements(),
+        mixed_precision_policy=MixedPrecisionPolicy(main_params_dtype=torch.float32),
+    )
+    # SGD's foreach/fused CUDA paths require matching parameter and gradient dtypes.
+    # Use the scalar path to exercise FP32 main weights with default BF16 main grads.
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.25, foreach=False)
```
wujingyue, line 224: I'm obviously taking a shortcut here. 

Right now we have fairly complicated logic around precision mismatches between main_weight and main_grad (for example, use_decoupled_grad). I’m considering simplifying this for the new design. 

Algorithmically, I think this ultimately has to be solved at the optimizer layer — e.g., by making optimizers precision-aware, or by wrapping existing optimizers when we don’t control their implementations. I don’t really see an alternative given the following constraints:
1. We need to support main_weight in fp32 because optimizers generally prefer higher precision.
2. We need to support main_grad in bf16 because computing fp32 gradients during backward is too expensive.
3. We need to support vanilla torch.optim.Optimizers, almost all of which assume param.data.dtype == param.grad.dtype.

Am I missing anything? cc @skyw

skyw: There are more combinations. a made up example (not sure megatron does this)
* Do gradient accumulation step in main_grad in fp32 while grad itself per MBS is computed in bf16.

Probably should have a list of mixed precision "recipes" that need to be supported. Not necessarily everything non FSDP path.

wujingyue: > Do gradient accumulation step in main_grad in fp32 while grad itself per MBS is computed in bf16.

Good to know — I’ve been setting .grad_dtype in this PR, which would be problematic for this use case. I’ll fix that before sending the PR out for review.

That said, I think this is a different problem that can be solved in post_backward. The optimizer sees sharded/DTensor parameters backed by main_weight and main_grad; in this particular case, both are fp32.

Back to my original question: I’m trying to see whether I can simplify the algorithm. It sounds like the answer from you is no, and that we may even need to support more combinations!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
```diff
@@ -0,0 +1,494 @@
...
+        ):
+            self.model_weight.redistribute(
+                self._unsharded_model_weight.placements, out=self._unsharded_model_weight
+            )
+        self._switch_to_unsharded_parameters()
+
+    def reshard_parameters(self) -> None:
+        """Install sharded DTensor parameters on the owning modules."""
+        self._switch_to_sharded_parameters()
+        self._unsharded_model_weight.release_storage()
```
wujingyue: This is an overkill for post backward. We can simply drop the tensor reference for backward; we don't have to hold on to the same tensor object.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
```diff
@@ -0,0 +1,494 @@
...
+
+        if not has_sharded_grads:
+            for index, parameter in enumerate(self.sharded_parameters):
+                parameter.grad = self.main_grad.get_dtensor(index)
+
+        for parameter in self.unsharded_parameters:
+            parameter.grad = None
+
+
+class FsdpModule:
```
wujingyue:
```suggestion
class FsdpUnit:
```
? In fact, this class is not even a nn.Module but a mixin.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
```diff
@@ -0,0 +1,494 @@
...
+        for name, placements in (
+            ("parameter", self.parameter),
+            ("gradient", self.gradient),
+            ("optimizer", self.optimizer),
+        ):
+            if len(placements) != axis_count:
+                raise ValueError(f"Expected {axis_count} {name} placements, got {len(placements)}.")
+
+
+class ParameterGroup:
```
wujingyue:
```suggestion
class FsdpParameterGroup:
```
so it's hard to confuse with optimizers' parameter groups.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
```diff
@@ -254,6 +269,21 @@ def _create_or_validate_out(
...
+            return self
+
+        destination = DBuffer(
+            mesh=self.mesh,
+            placements=self.placements,
+            tensor_shapes=self.layout.tensor_shapes,
+            dtype=dtype,
+            device=self.device,
+        )
+        destination.local_buffer.copy_(self.local_buffer)
```
Autumn1998, line 284: Do we need to consider the quantization that mixed precision might introduce?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -0,0 +1,259 @@
...
+        if self.main_weight is self.model_weight:
+            return
+
+        self.main_weight.cast(self.model_weight.local_buffer.dtype).redistribute(
+            self.model_weight.placements, out=self.model_weight
+        )
+
+    def unshard_parameters(self) -> None:
+        """Install full parameters for local compute."""
+        self.sync_model_weight_from_main_weight()
```
Autumn1998: Is this saying that we need to cast from the main weights to the model weights before every unshard? I think we only trigger this once per global batch.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
wujingyue: It's out of the scope of this PR, but it certainly needs to be addressed later.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue: Good catch! This used to be in a post-optimizer-step hook and got moved here after I decided to split optimizer changes into a separate PR. 

For now, I moved this to pre-forward so at least it's not called unnecessarily in pre-backward. 

https://github.com/NVIDIA/Megatron-LM/pull/5378 is a better fix, which sets microbatch states. Feel free to review that PR. I believe these states are necessary for HSDP/HFSDP anyway. 

Another option is to sync weights after each optimizer step.

wujingyue: From @shjwudp: The `microbatch` scope could be problematic for 1F1B, because it's hard for the user to identify the first/last microbatch.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -0,0 +1,258 @@
...
+            tensor_shapes=tensor_shapes,
+            dtype=self.dtype,
+            device=self.main_weight.device,
+        )
+        if main_weight_dtype == self.dtype and main_weight_placements == model_weight_placements:
+            self.model_weight = self.main_weight
+        else:
+            self.model_weight = DBuffer(
+                mesh=self.mesh,
+                placements=model_weight_placements,
```
wujingyue, line 115: From @Autumn1998: seems similar to how we create unsharded_model_weight. Consider dedup.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -0,0 +1,258 @@
...
+            self.model_weight.placements, out=self.model_weight
+        )
+
+    def unshard_parameters(self) -> None:
+        """Install full parameters for local compute."""
+        self._unsharded_model_weight.reallocate_storage()
+        # This buffer backs unsharded Parameters whose views may be saved by autograd.
+        # Materializing FSDP-managed storage should not look like a user mutation.
+        with torch.autograd._unsafe_preserve_version_counter(
+            self._unsharded_model_weight.local_buffer
```
wujingyue, line 191: From @shjwudp: understand how it is different from no_grad.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -0,0 +1,258 @@
...
+                placements=model_weight_placements,
+                tensor_shapes=tensor_shapes,
+                dtype=self.dtype,
+                device=self.main_weight.device,
+            )
+
+        self.main_grad = None
+        if self.requires_grad:
+            grad_dtype = mixed_precision_policy.main_grads_dtype or self.dtype
+            self.main_grad = DBuffer(
```
wujingyue, line 124: From @shjwudp: When micro batch size is one, some users prefer to not store main_grad persistently. The benefit is to not allocate main_grad during forward.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fsdp_module.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def _make_grad_hook(self, parameter: nn.Parameter) -> Callable[[nn.Parameter], None]:
+        def grad_hook(_parameter: nn.Parameter) -> None:
+            self._ready_grad_parameters.add(parameter)
+            if len(self._ready_grad_parameters) == self._num_training_parameters:
+                self.post_backward()
+
+        return grad_hook
+
+    def pre_forward(self) -> None:
```
wujingyue, line 85: From @Autumn1998 : pre_forward/forward/post_forward may be called multiple times when activation recomputation is on.

wujingyue: @Autumn1998 to provide the code pointer

shjwudp (approved):

wujingyue:
Recovered as replacement draft PR #5387 after this PR was closed due to the deleted base ref. The branch has been rebased directly on main: https://github.com/NVIDIA/Megatron-LM/pull/5387

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue: Clarified in 23c47a847fcdde714903f2bffb3d630388bbd6e1

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
wujingyue: Decided to not introduce another term.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
wujingyue: Done

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue: ea39f57762995d52d9f2de10d7d289a4a7ad86e0 adds a code comment about this.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fsdp_module.py
wujingyue: The user should probably call module.forward instead of `module(...)` when the intention is to avoid triggering hooks.

PR #4987, opened 2026-05-26
feat(optim): support --fp8-param-gather for muon + mxfp8 in LayerWise

Wohox:
## Summary

`--fp8-param-gather` + `--reuse-grad-buf-for-mxfp8-param-ag` (mxfp8) was already wired for the standard `DistributedOptimizer`, but the muon `LayerWiseDistributedOptimizer` was missing the pieces needed to take that same path. This PR fills those gaps so `--fp8-param-gather` works for muon-managed 2D weights when `--fp8-recipe=mxfp8`, on top of #4889's muon-aware bucket alignment, with **bitwise parity against the `--fp8-param-gather=off` baseline**.

Design principle: **reuse the DistOpt machinery instead of growing a parallel LayerWise path.** The bf16-staging write is exposed as a duck-typed `_copy_main_params_to_param_buffer()` so `train_step` drives both optimizers through one code path, and the per-bucket param-sync filtering mirrors DistOpt's exactly.

Background: with this feature the model params are kept as **persistent MXFP8** tensors; the bf16 `param_buffer` is aliased onto the idle grad buffer; the all-gather ships bf16 and then quantizes back to MXFP8 in place on every rank. Every gap below is a consequence of the muon/LayerWise path not yet handling some part of that lifecycle that DistOpt already did.

The gaps fall into four areas.

---

## Area 1 — Master-weight numerics for persistent FP8 params

These two gaps made the muon loss drift from the `fp8_param_gather=off` baseline. Both exist *only* because the params are now persistent MXFP8 tensors.

**1a — fp32 master was initialized from dequantized MXFP8.**
The inner `Float16OptimizerWithFloat16Params.__init__` builds the master via `param.detach().clone().float()`. For an MXFP8 param, `.float()` dequantizes, so the master is born from FP8-rounded values and carries ~FP8 noise from iter 0. DistOpt avoids this by reading `model_param.get_high_precision_init_val()` (the bf16 init TE preserves on CPU). Without the fix, the iter-10 lm-loss gap is ≈0.27 nats.
→ filled by `_restore_high_precision_init_val` (mirrors `distrib_optimizer.py:405-416`).

**1b — the inner step re-quantized MXFP8 params redundantly.**
`_copy_main_params_to_model_params` would run `model.data.copy_(main_fp32)` on each MXFP8 model param, triggering an extra `QuantizedTensor` `quantize_(fp32)` that the post-AG `quantize_(bf16)` then overwrites — wasted work that also perturbs muon convergence. DistOpt's `reuse_grad_buf` step path never touches MXFP8 storage at the inner step.
→ filled by `_skip_mxfp8_in_copy_main_to_model` (+ a shallow-copied `inner_config` that hides `reuse_grad_buf_for_mxfp8_param_ag` from the inner optimizer; `self.config` is restored after `super().__init__` so the outer ChainedOptimizer's shared-config assertion still holds).

## Area 2 — Staging the bf16 param buffer before the all-gather

With `reuse_grad_buf_for_mxfp8_param_ag` the bf16 `param_buffer` is separate from the MXFP8 storage, and the all-gather ships `param_buffer`. Two gaps left it holding the wrong bytes.

**2a — the updated master was never written into the staging buffer.**
The optimizer step updates the fp32 master, but nothing copied it (cast to bf16) into `param_buffer` before the AG, so the AG shipped stale bytes.
→ filled by `_copy_main_params_to_param_buffer` — a duck-typed twin of `DistributedOptimizer._copy_main_params_to_param_buffer`, so `train_step` can call it polymorphically over `(DistributedOptimizer, LayerWiseDistributedOptimizer)`. Self-guards on `use_buffer_param_sync` + `reuse_grad_buf_for_mxfp8_param_ag`, so it is a no-op for configs that don't need it.

**2b — the staging buffer is zeroed before the deferred all-gather (overlap path).**
Because the staging buffer is *aliased onto the grad buffer*, the next iteration's `zero_grad_buffer()` zeroes it **after** the in-step write but **before** the deferred forward-pre-hook all-gather. Under `--overlap-param-gather` the AG therefore shipped a zeroed buffer and the muon-managed weights never updated (loss frozen). The fix is **not** to disable overlap, but to re-stage post-zero on every site that force-syncs through the param buffer:
- `train_step` (training loop) — the same hook DistOpt already uses;
- the **eval** block (`disable_forward_pre_hook(param_sync=True)` force-syncs params for eval);
- `paged_stash` `_try_copy_main_params` (reachable when `moe_expert_rank_capacity_factor` is set with `reuse_grad_buf + overlap`).
All three now re-stage `LayerWiseDistributedOptimizer` too, not just `DistributedOptimizer`.

## Area 3 — Disjoint param-sync when LayerWise and DistOpt coexist

muon manages 2D weights via LayerWise; everything else goes through a sibling `DistributedOptimizer` (#4771). They share the same DDP buffers, so the gap is making each optimizer all-gather **only its own buckets**.

**3a — per-bucket ownership / mixed bucket groups.**
A bucket group can be mixed-ownership: in `partition_buckets` Case 3 (FP8 present, `reduce_scatter_with_fp32_accumulation=False`) non-FP8 DistOpt-managed bf16 buckets (biases, layernorms) get appended into the last FP8 bucket group. Checking only `buckets[0]` and dispatching AG on the whole group double-syncs the foreign buckets. Both optimizers now filter **per-bucket** and, for mixed groups, synthesize a thin transient bucket group containing only the buckets they own. Applied symmetrically in three places: `DistributedOptimizer.start_param_sync_for_bucket_group_subset`, `LayerWiseDistributedOptimizer.start_param_sync_for_bucket_group_subset`, and the deferred-sync path in `optimizer.py` (which also recurses the chained-optimizer tree so nested DistOpt instances are reached).

## Area 4 — Fail-fast configuration guards

The remaining gap was silent misconfiguration: combinations that would quietly all-gather stale storage.

**4a** — mxfp8 + LayerWise + `--fp8-param-gather` now requires `--reuse-grad-buf-for-mxfp8-param-ag` (the two must be co-enabled).
**4b** — non-mxfp8 + LayerWise + `--fp8-param-gather` is explicitly rejected (only MXFP8's `_rowwise`/`_columnwise` round-trip is wired up today).

---

## Deliberately NOT in this PR

- **Quantizer dual-usage re-assert.** An earlier revision re-asserted `quantizer.set_usage(rowwise=True, columnwise=True)` before each AG, guessing TE's bwd toggled `columnwise_usage`. Running the bitwise test in `deterministic_mode()` **without** that re-assert still reproduces `atol=rtol=0` ON-vs-OFF parity, proving it unnecessary; the helper was removed to keep the LayerWise surface minimal.
- **Blockwise FP8 + LayerWise + `--fp8-param-gather`** — rejected at arg-parse (4b) and skipped in the test; needs a separate `_rowwise`/`_columnwise` layout + post-AG path.

## Test plan

* `tests/unit_tests/test_muon_mxfp8_fp8_param_gather.py::test_on_vs_off_bitwise_identical` — builds the same small GPT twice (`fp8_param_gather=False`/`True`), restores the ON run's init state from the OFF snapshot so both start bit-identical, runs 5 muon steps in `deterministic_mode()`, and asserts `atol=rtol=0` on loss, forward output, per-parameter `main_grad`, fp32 master, and (non-MXFP8) model params. Parametrized over **`overlap_param_gather ∈ {False, True}`** — the `True` case is the regression guard for Area 2b. Skips automatically when MXFP8 is unsupported (pre-Blackwell, sm_120+, TE without MXFP8). Verified on GB200: `[mxfp8-False]` and `[mxfp8-True]` both pass; blockwise variants skip by design.
* End-to-end convergence smoke (DSV4 proxy, GB200, muon, e4m3 mxfp8): ON tracks the OFF baseline (iter-10 within ≈0.03 nats; ≈0.27 nats without Area 1a). Overlap ON: before Area 2b the loss was frozen; after, the overlap run tracks the non-overlap run.

## Changes by file

| File | Area |
|---|---|
| `megatron/core/optimizer/layer_wise_optimizer.py` | 1a, 1b, 2a, 2b, 3a |
| `megatron/core/optimizer/distrib_optimizer.py` | 3a (per-bucket filter / mixed-group sub-group) |
| `megatron/core/optimizer/optimizer.py` | 3a (deferred-sync per-bucket filter + recursion) |
| `megatron/training/training.py` | 2b (train-loop + eval re-stage) |
| `megatron/core/transformer/moe/paged_stash.py` | 2b (paged-stash re-stage) |
| `megatron/training/arguments.py` | 4a, 4b |
| `tests/unit_tests/test_muon_mxfp8_fp8_param_gather.py` | bitwise ON-vs-OFF, overlap on/off |


megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
wujingyue, line 1: I'm not sure about the Megatron convention. Does the license year go with when the file was originally written or when the file was latest updated? @ericharper

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -3003,11 +3003,28 @@ def start_param_sync_for_bucket_group_subset(self) -> None:
...
+
+                if len(distopt_buckets) == len(bucket_group.buckets):
+                    model_chunk._start_bucket_group_param_sync(bucket_group, force_sync=False)
+                else:
+                    distopt_bucket_group = type(bucket_group)(
+                        distopt_buckets,
+                        bucket_group.ddp_config,
+                        bucket_group.intra_distributed_optimizer_instance_group,
+                        bucket_group.intra_distributed_optimizer_instance_size,
+                    )
```
wujingyue, line 3046: I think this is added logic and I've yet to understand what it does. Some comments would help.

wujingyue (commented):
Thanks for the PR! I'll wait until expert reviews (esp dist-optimizer) are done.

megatron/core/optimizer/distrib_optimizer.py
ericharper: It's fine to update them.

Wohox: Thanks @ericharper — keeping the 2026 bump. Resolving.

megatron/core/optimizer/distrib_optimizer.py
Wohox: Good call — added an extended docstring + inline comments on each branch in `70c5487dc`. The three cases now read:

1. **Entire group LayerWise-owned** → `continue` (LayerWise side syncs it).
2. **Pure DistOpt group** → dispatch the original bucket group as-is so DDP-side per-group state (handles, gather lists, etc.) is reused.
3. **Mixed group** (only possible in `partition_buckets` Case 3 — FP8 present + `reduce_scatter_with_fp32_accumulation=False` — which appends non-FP8 DistOpt bf16 buckets into the last FP8 bucket group): build a thin transient `type(bucket_group)(distopt_buckets, …)` that reuses the parent ddp_config / DP group / DP world size so the AG collective lands on the same comm. The LayerWise mirror of this routine independently syncs the LayerWise-tagged buckets from the same parent group.

The docstring was also extended to call out the Case 3 scenario explicitly so anyone reading the function next time has the context up front.

megatron/training/arguments.py
```diff
@@ -1075,6 +1075,49 @@ def validate_args(args, defaults={}):
...
+        # grad buffer (the same ``reuse_grad_buf_for_mxfp8_param_ag`` trick
+        # used by the standard DistributedOptimizer). For MXFP8 + LayerWise the
+        # model params keep their own ``_rowwise_data`` / ``_columnwise_data``
+        # storage and the bf16 ``param_buffer`` is separate; if we don't also
+        # write the optimizer-step output into ``param_buffer`` before the AG,
+        # the AG ships stale bytes. The LayerWise code path that does that
+        # write is gated on ``reuse_grad_buf_for_mxfp8_param_ag``, so require
+        # the flag explicitly to fail fast on misconfiguration.
+        will_use_layer_wise_distributed_optimizer = (
+            args.optimizer not in ('sgd', 'adam') and args.use_distributed_optimizer
```
FDecaYed, line 1093: @kunlunl to also comment. 
My rough understanding is fp8-gather and reuse flags are recommended to be used together even before any muon changes for mxfp8? and at least for mxfp8 there is no reason to use only fp8-gather without reuse
Maybe let's take the opportunity to finally clean up these flag that doesn't quite show what they do as name suggests. I'm thinking something like:

- mxfp8 + fp8-param-gather: equivalent to old fp8-param-gather+reuse, deprecate the reuse flag, support layerwise
- non-mxfp8 + fp8-param-gather: equivalent to before, layerwise not supported

tests/unit_tests/test_muon_mxfp8_fp8_param_gather.py
```diff
@@ -0,0 +1,466 @@
...
+        args.muon_coefficient_type = 'quintic'
+        args.muon_tp_mode = 'duplicated'
+        args.use_precision_aware_optimizer = False
+        args.exp_avg_dtype = 'fp32'
+        args.exp_avg_sq_dtype = 'fp32'
+        args.use_distributed_optimizer = True
+        # Disable AG / RS overlap so the timing is deterministic and the test
+        # exercises the synchronous post-AG quantize path (see
+        # ``_post_param_sync`` in ``param_and_grad_buffer.py``).
+        args.overlap_param_gather = False
```
FDecaYed: I think we should also test the overlapping path since that's more important

megatron/core/optimizer/layer_wise_optimizer.py
```diff
@@ -717,24 +849,149 @@ def count_zeros(self):
...
+                    lw_bucket_group = type(bucket_group)(
+                        lw_buckets,
+                        bucket_group.ddp_config,
+                        bucket_group.intra_distributed_optimizer_instance_group,
+                        bucket_group.intra_distributed_optimizer_instance_size,
+                    )
+                    model_chunk._start_bucket_group_param_sync(lw_bucket_group, force_sync=False)
+
+    @torch.no_grad()
+    def _write_owned_mxfp8_masters_to_param_buffer(self) -> None:
```
FDecaYed: I have a hunch that this is all over-complicated because we originally designed layerwise optimizer to 'replace' dist-opt, so it inherited chainedoptimizer, while distopt inherited mixprecision optimizer
since now we already made up our mind to get layerwise/distopt work alongside each other, maybe we change that? so something like a chained optimizer chaining both layerwise and distopt instance. This way, both can implement _copy_main_params_to_param_buffer that being invoked with same code path?

That said, the code here make sense, and if we want to get this in first and do refactor later, I'm also fine with that

FDecaYed (changes requested):

megatron/core/optimizer/layer_wise_optimizer.py
Wohox: To summarise here, we have 2 pending refactorings:
* Unify `reuse_grad_buf_for_mxfp8_param_ag` and `fp8_param_gather`.
* Let LayerwiseDistOpt inherit from MixedPrecisionOpt instead of ChainedOpt

I would suggest to keep this PR simple and support the above refactorings in another PR.

megatron/training/arguments.py
Wohox: I agree with the proposal but have a concern regarding the concurrent use of FP8 parameter gather, reuse gradient reduce buffer, and extra overlap gradient reduce. Our testing showed that enabling these features simultaneously can trigger NaN issues.

Currently, customers can workaround this by enabling reuse gradient reduce while keeping overlap parameter gather disabled. The proposed design would remove this bypass.

@kunlunl Could you confirm if this issue has been resolved or if there is a PR tracking it?

kunlunl: As far as I know, all NaN bugs related to mxfp8 / fp8 param gather / reuse grad buf / overlap param gather are resolved in dev branch, except an open PR (https://github.com/NVIDIA/Megatron-LM/pull/4994) that also will be merged soon.

And I think the proposal by @FDecaYed doesn't force use overlap param gather. It just deprecate the `reuse_xxx`, because it's actually not an option, if you want to use fp8 param + mxfp8, you must enable it. if you want to use fp8 param + other fp8 recipe, you must disable `reuse_xxx`, so no meaning to keep it.

FDecaYed:
Closing this PR because Muon + MXFP8 LayerWise parameter gather is no longer required by the current DeepSeek-V4 HybridModel delivery plan (#5795). The frozen DSv4 recipe uses the standard DistributedOptimizer rather than Muon, so this is not a prerequisite; if the generic optimizer feature is still desired independently, it can be reopened or resubmitted outside the DSv4 plan after its design threads are resolved.

PR #4990, opened 2026-05-26
Fix TEGroupedMLP pre-backward unshard in fine-grained FSDP hooks for …

rapatel:
…1F1B EP overlap.

Dev PR - [Megatron-LM/pull/4989](https://github.com/NVIDIA/Megatron-LM/pull/4989)

# What does this PR do ?

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
@@ -179,8 +179,8 @@ class MegatronFSDP(torch.nn.Module):
             which can improve performance when using MXFP8 parameters with activation recomputation.
         enable_fine_grained_param_gather_backward_hook (bool): Register pre-backward unshard hooks
             on each submodule (used by 1F1B EP overlap and similar schedules).
-        fine_grained_pre_backward_recurse_module_types (Optional[Tuple[Type[nn.Module], ...]]):
-            Module classes for which fine-grained pre-backward unshard uses
+        fine_grained_recurse_module_types (Optional[Tuple[Type[nn.Module], ...]]):
+            Module classes for which fine-grained pre-forward / pre-backward unshard uses
             ``parameters(recurse=True)`` (container modules whose sharded weights live on
             children). Checked with :func:`isinstance`. Defaults to empty (none).
         report_nan_in_param_grad (bool): Whether to enable precise NaN-checking for parameter wgrad.
@@ -223,9 +223,7 @@ class MegatronFSDP(torch.nn.Module):
         disable_symmetric_registration: bool = False,
         enable_fine_grained_param_gather_hook: bool = False,
         enable_fine_grained_param_gather_backward_hook: bool = False,
-        fine_grained_pre_backward_recurse_module_types: Optional[
-            Tuple[Type[nn.Module], ...]
-        ] = None,
+        fine_grained_recurse_module_types: Optional[Tuple[Type[nn.Module], ...]] = None,
         report_nan_in_param_grad: bool = False,
     ):
         super().__init__()
@@ -281,10 +279,8 @@ class MegatronFSDP(torch.nn.Module):
         self.enable_fine_grained_param_gather_backward_hook = (
             enable_fine_grained_param_gather_backward_hook
         )
-        recurse_types = fine_grained_pre_backward_recurse_module_types or ()
-        self.fine_grained_pre_backward_recurse_module_types: Tuple[
-            Type[nn.Module], ...
-        ] = recurse_types
+        recurse_types = fine_grained_recurse_module_types or ()
+        self.fine_grained_recurse_module_types: Tuple[Type[nn.Module], ...] = recurse_types
         self.report_nan_in_param_grad = report_nan_in_param_grad
 
         # FSDPDistributedIndex stores the process groups and meshes used by Megatron-FSDP.
@@ -559,6 +555,25 @@ class MegatronFSDP(torch.nn.Module):
         """
         fsdp_unit_modules = self.fsdp_unit_modules
 
+        def _param_list_for_submodule_unshard(
+            module: nn.Module, *, for_backward: bool
+        ) -> List[nn.Parameter]:
+            """Build the parameter list for fine-grained or FSDP-unit unshard hooks."""
+            if isinstance(module, tuple(fsdp_unit_modules)):
+                return list(module.parameters())
+            fine_grained_enabled = (
+                self.enable_fine_grained_param_gather_backward_hook
+                if for_backward
+                else self.enable_fine_grained_param_gather_hook
+            )
+            if (
+                fine_grained_enabled
+                and self.fine_grained_recurse_module_types
+                and isinstance(module, self.fine_grained_recurse_module_types)
+            ):
+                return list(module.parameters(recurse=True))
+            return list(module.parameters(recurse=False))
+
         def release_module_parameters(module, bwd, lazy=False, *unused):
             """
             Release the parameters of a given module after completing the forward
@@ -749,16 +764,7 @@ class MegatronFSDP(torch.nn.Module):
             else:
                 module._training_state = TrainingState.FORWARD
 
-            if isinstance(module, tuple(fsdp_unit_modules)):
-                param_list = list(module.parameters())
-            else:
-                # All-gather the shallow parameters in every forward pass for modules
-                # that are not FSDP units. Do not recurse unless absolutely necessary,
-                # to allocate as little memory as possible for this forward pass.
-                param_list = list(module.parameters(recurse=False))
-
-            if self.enable_fine_grained_param_gather_hook:
-                param_list = list(module.parameters(recurse=False))
+            param_list = _param_list_for_submodule_unshard(module, for_backward=False)
 
             # All-gather the parameters before the forward pass.
             self.all_gather_and_wait_parameters_ready(
@@ -874,17 +880,7 @@ class MegatronFSDP(torch.nn.Module):
             for sub_module in module.modules():
                 sub_module._training_state = TrainingState.PRE_BACKWARD
 
-            if isinstance(module, tuple(fsdp_unit_modules)):
-                param_list = list(module.parameters())
-            elif (
-                self.enable_fine_grained_param_gather_backward_hook
-                and self.fine_grained_pre_backward_recurse_module_types
-                and isinstance(module, self.fine_grained_pre_backward_recurse_module_types)
-            ):
-                # Container modules (e.g. TEGroupedMLP): sharded weights are on children.
-                param_list = list(module.parameters(recurse=True))
-            else:
-                param_list = list(module.parameters(recurse=False))
+            param_list = _param_list_for_submodule_unshard(module, for_backward=True)
 
             # All-gather / unshard the module parameters before the backward pass.
             self.all_gather_and_wait_parameters_ready(
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
@@ -558,21 +558,45 @@ class MegatronFSDP(torch.nn.Module):
         def _param_list_for_submodule_unshard(
             module: nn.Module, pass_direction: Literal["forward", "backward"]
         ) -> List[nn.Parameter]:
-            """Build the parameter list for fine-grained or FSDP-unit unshard hooks."""
-            if isinstance(module, tuple(fsdp_unit_modules)):
-                return list(module.parameters())
+            """Build the parameter list for fine-grained or FSDP-unit unshard hooks.
+
+            Parameter buckets designated by this function are all-gathered and may
+            pre-fetch subsequent buckets in FSDP bucket order during runtime.
+            """
+            # Fine-grained hooks are attached to all sub-modules; this function
+            # controls which parameters each hook should unshard.
             fine_grained_enabled = (
                 self.enable_fine_grained_param_gather_backward_hook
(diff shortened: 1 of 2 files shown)


rapatel:
@cspades PTAL. Thanks!

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -874,6 +887,13 @@ def _pre_backward_param_unshard(module: nn.Module, *unused):
 
             if isinstance(module, tuple(fsdp_unit_modules)):
                 param_list = list(module.parameters())
+            elif (
+                self.enable_fine_grained_param_gather_backward_hook
+                and self.fine_grained_pre_backward_recurse_module_types
+                and isinstance(module, self.fine_grained_pre_backward_recurse_module_types)
+            ):
+                # Container modules (e.g. TEGroupedMLP): sharded weights are on children.
+                param_list = list(module.parameters(recurse=True))
```
wujingyue: This is at least asymmetric with https://github.com/NVIDIA/Megatron-LM/blob/859b7197f19119f53113ab6d30d1901b87ca1197/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L758-L759

Can you clarify the intention? I've yet to understand the combination of fine-grained **and** recurse. When enable_fine_grained_param_gather_backward_hook is on, I assume you want to register hooks on the children modules in addition to the root. Therefore, children modules are supposed to take care of their own parameters and `recurse=False`.

wujingyue (commented):
Thanks for the PR!

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
rapatel: Good question - the confusion is mostly about what “fine-grained” registers vs what actually runs in the 1F1B overlap schedule.

Fine-grained mode registers a pre-bwd hook on every submodule so overlap can call attn / mlp / experts directly instead of TransformerLayer.forward(). For normal modules, recurse=False is right - only unshard that module’s own parameters.

TEGroupedMLP is a container with no direct parameters (weights are on linear_fc1 / linear_fc2), so recurse=False gives an empty param list and the hook does nothing. Child linears don’t save us in the fused/overlap path because their forward often doesn’t run; backward is tied to TEGroupedMLP’s output, so that hook must use recurse=True. This is an allowlist for that container case, not “recurse everywhere.”

This PR is to allow the 1F1B implementation from @Wohox - https://github.com/NVIDIA/Megatron-LM/pull/3797 - to run with fused version of TEGroupedMLP. Without this change, we miss the un-sharding (param AG) of linear fc1/2 weights during bprop.

wujingyue: > TEGroupedMLP is a container with no direct parameters (weights are on linear_fc1 / linear_fc2), so recurse=False gives an empty param list and the hook does nothing. Child linears don’t save us in the fused/overlap path because their forward often doesn’t run

Interesting -- I saw a similar pattern with MambaLayer. #4899 is for fixing that because we own the source code. 

How does this PR fix this problem? If hooks are attached to TEGroupedMLP, communication is triggered before/after TEGroupedMLP and therefore not fine-grained enough for overlapping. If to submodules, the parameters may be used outside the unshard-to-reshard window, which is unsafe. What am I missing?

rapatel: Thanks I believe this is the similar idea as MambaLayer (#4899) (container + fine-grained hooks), but for MoE experts under overlap.

I am not adding new hooks - with enable_fine_grained_param_gather_backward_hook, hooks are already on every submodule, including TEGroupedMLP. The bug is the param list - recurse=False on the container -> no params -> no unshard -> MMU fault. We only fix that list so that we grab the appropriate params.

Overlap (1F1B) already runs at submodule granularity (experts = TEGroupedMLP). On the fused path, linear_fc1/linear_fc2 don’t run forward, so backward only sees the TEGroupedMLP hook. We’re not making overlap coarser - we’re making that hook actually gather child weights.

Also, we’re not moving hooks to children only. Child hooks may exist but don’t run if their forward didn’t run. The failure was no unshard before expert backward, not params used outside the window.

@cspades Please chime in as I may be missing something and you folks are a lot more familiar with this codebase. Want to make sure my understanding is correct. Thanks!

wujingyue: Got it -- there are actually three levels: the parent of TEGroupedMLP (which is an FSDP unit), TEGroupedMLP, and the direct children of TEGroupedMLP. TEGroupedMLP is actually at the correct FSDP granularity. 

I'll review this again soon and thanks for the context!

rapatel: Sure - thanks for reviewing it for me.

BTW - if you take a look at the [1f1b schedule logic](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/models/common/model_chunk_schedule_plan.py#L230) - you can see which module is getting invoked at what time. You'll see the following:

```python
        if b_layer is not None:
            b_grad = b_layer.mlp.backward(b_grad)
```

The fsdp unit is `transformer_layer` (parent of TEGroupedMLP), but the unit itself is never invoked.

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -177,6 +177,12 @@ class MegatronFSDP(torch.nn.Module):
             userbuffer registration when nccl_ub is set.
         enable_fine_grained_param_gather (bool): Whether to enable "fine-grained" param all-gather,
             which can improve performance when using MXFP8 parameters with activation recomputation.
+        enable_fine_grained_param_gather_backward_hook (bool): Register pre-backward unshard hooks
```
wujingyue, line 180: Note to myself: this PR doesn't add this flag; it merely adds a missing comment.

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
wujingyue: The PR makes more sense to me now. 

These are probably pre-existing issues, but what still confuse me:
1. There's an asymmetry between pre_forward_param_unshard and pre_backward_param_unshard: https://github.com/NVIDIA/Megatron-LM/blob/146d1719f8cf6631ddd325c40a9ed72710a1a986/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L753-L760. Shouldn't this change apply to pre-forward as well otherwise linear_fc1/2's parameters might not be ready before forward? Is the problem being hidden by prefetching again? 
2. Why not recurse=True unconditionally? The comment [here](https://github.com/NVIDIA/Megatron-LM/blob/146d1719f8cf6631ddd325c40a9ed72710a1a986/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py#L754-L756) is plausible. However, don't we have a correctness issue whenever `module` is a submodule of an FSDP unit and doesn't directly own every parameter under it? 
3. Looks like unshard is fine-grained but reduce_grads is not (it's still on TransformerLayer). Does that create a problem for 1f1b overlapping? 

Regarding the PR, if you could rename fine_grained_pre_backward_recurse_module_types to fine_grained_recurse_module_types and apply it both pre-forward and pre-backward, the PR will LGTM. The blast radius would be small because fine_grained_pre_backward_recurse_module_types is a new opt-in knob. 

cc @shjwudp

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -553,6 +562,25 @@ def _register_fsdp_hooks(self, root_module):
         """
         fsdp_unit_modules = self.fsdp_unit_modules
 
+        def _param_list_for_submodule_unshard(
+            module: nn.Module, *, for_backward: bool
```
wujingyue: Consider https://typing.python.org/en/latest/spec/literal.html instead of a boolean. e.g. `Literal["forward", "backward"]`. We could do this for optimizer and it's not immediately clear what `for_backward=False` means. 

Ref: https://www.teamten.com/lawrence/programming/prefer-enums-over-booleans.html

wujingyue (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
rapatel: Great suggestion - applied. Thanks!

Phlip79 (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
```diff
@@ -546,6 +555,25 @@ def _register_fsdp_hooks(self, root_module):
...
+                if pass_direction == "backward"
+                else self.enable_fine_grained_param_gather_hook
+            )
+            if (
+                fine_grained_enabled
+                and self.fine_grained_recurse_module_types
+                and isinstance(module, self.fine_grained_recurse_module_types)
+            ):
+                return list(module.parameters(recurse=True))
+            return list(module.parameters(recurse=False))
```
cspades: Wait, looking at this again, shouldn't the logic be:

- If not an FSDP unit, then don't recurse.
  - Always fine-grained AG, since we don't want to over-gather parameters that we may not need yet, as this module doesn't have a batch specification (or user-guaranteed FSDP-compliant schedule) for how many parameters need to be communicated. **Attach hooks to _everything_!**
- If an FSDP unit...
  - If not fine-grained, then recurse.
    - This is the vanilla case, an FSDP unit module that is not fine-grained should fetch all of its parameters.
  - If fine-grained and is not a recursing module...
    - If we are gathering forward-pass / row-wise parameters (`_pre_forward_param_unshard`), then don't recurse.
      - **_Even if the module is an FSDP unit module, we want to fine-grain AG the forward pass / row-wise parameters!_** ❗ 
      - Used to deal with activation recompute (with forward pass parameters). This is because activation recompute is at a fine-grained module level (not at the FSDP unit level), so without a fine-grained hook, module forward parameters will not be gathered and thus activation recompute will be functionally incorrect.
        - Without MXFP8, backward AG == forward AG. So recompute is no-problemo, because backward AG always recurses.
        - With MXFP8, we absolutely need to fine-grain AG the row-wise parameters.
    - If we are gathering backward-pass / column-wise parameters (`_pre_backward_param_unshard`), then recurse.
      - Vanilla case. Backward pass parameters don't need fine-grained AG, we can just gather everything since we're going to use everything anyway.
  - If fine-grained and is a recursing module, then recurse.
    - Like `TEFusedMLP` where its parameters lose their membership (and hooks) to the top-level `TEFusedMLP` module. This is your new override logic. 🍀 

The logic currently looks like:

- If FSDP unit, then recurse. (`recurse=True` is the default!)
- If not FSDP unit and fine-grained and not a recursing module, then don't recurse.
- If not FSDP unit and fine-grained but it is a recursing module, then recurse.

TL;DR FSDP unit modules can still be fine-grained AG for sub-module recompute.

cspades (changes requested):

megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
cspades: cc @Autumn1998 to check if this logic makes sense.

cspades: cc @wujingyue @shjwudp as well. We need better APIs here for all of these cases in V2.

rapatel: @cspades Yeah I think you're right... but I don't think anything outside of TransformerLayer right now is an FSDP unit. And in the 1F1B case, we saw the TEGroupedMLP module gets invoked, and we never recurse it for the param list, so it returns [], so we need to recurse it to grab the param list from fc1 and fc2.

Let me think about it and come up with a proposed change. Good catch though - thanks!

Autumn1998: I understand that during recompute, a backward pass may trigger the forward of any individual part. Because of that, we need to attach fine-grained hooks to each part of the forward, so that backward can trigger them in a fine-grained way and avoid unnecessary communication. From my perspective, the logic looks correct. For the 1F1B scenario there may be more complex issues, since we have to manually call the forward and backward of some submodules, and @shjwudp  has been working on that recently.

rapatel: Addressed and reviewed changes with @cspades. See 338dfbb91c2af6d9c319e5b28762e2ba662b7f3a.

cspades (approved):

deepakn94 (approved):

cspades:
/ok to test [c0422cb](https://github.com/NVIDIA/Megatron-LM/pull/4990/commits/c0422cbdef5e5b1933af10fcd9d19d171d794e86)

cspades:
/ok to test [00ddbc6](https://github.com/NVIDIA/Megatron-LM/pull/4990/commits/00ddbc67074c40ba037bf711c6e8ec03a976ad21)

cspades:
/ok to test [cc78b37](https://github.com/NVIDIA/Megatron-LM/pull/4990/commits/cc78b379a420e37fcb68c35d08723c54ea7295a9)

Issue #4991, opened 2026-05-26
[DISCUSSION] Should safe_get_rank() warn or raise on invalid RANK values instead of falling back to 0?


fallintoplace:
**Describe the bug / discussion topic**

@NVIDIA/mcore-oncall

During review on PR #4959, a question came up about whether `safe_get_rank()` should continue to silently fall back to `0` when `RANK` is present but not parseable as an integer.

Today, both helpers keep longstanding fallback semantics:

- `megatron/core/_rank_utils.py`
- `megatron/core/distributed/fsdp/src/megatron_fsdp/utils.py`

Both currently catch `ValueError` / `TypeError` and return `0`.

This issue is to discuss whether that should remain the desired behavior, or whether invalid `RANK` values should instead surface a warning or raise an error.

**Steps/Code to reproduce bug**

Current behavior with a malformed environment variable is effectively:

```python
import os

os.environ["RANK"] = "not-an-int"

# safe_get_rank() currently returns 0
```

This happens because both implementations catch `ValueError` / `TypeError` from `int(...)` conversion and default to `0`.

**Expected behavior**

It would be useful to clarify the intended behavior for malformed `RANK` values. Reasonable options seem to be:

- keep current behavior: fall back to `0`
- warn, then fall back to `0`
- raise `ValueError` to surface misconfiguration immediately

**Additional context**

- This behavior is longstanding in the canonical helper: `safe_get_rank()` was introduced in `megatron/core/_rank_utils.py` in commit `6cf285b23` with the current fallback-to-`0` semantics.
- The Megatron-FSDP helper was added later in commit `b969f7642` and copied the same behavior.
- PR #4959 intentionally fixes only the missing `os` import in the FSDP helper and does not change these semantics.

If maintainers prefer a stricter or more explicit behavior here, I can send a follow-up PR to update both helpers consistently.

Connor-XY:
Hi @fallintoplace, thanks for splitting this behavior question out from #4959.

I checked the linked PRs. #4959 is only the missing-import fix and should not decide invalid `RANK` semantics. #4993 is the right follow-up and now updates both helpers consistently after maintainer feedback: invalid `RANK` should surface as a hard error rather than silently falling back to 0. That makes this issue the policy/tracking item for the behavior change rather than a separate runtime bug.

I assigned this to @wujingyue, who requested the follow-up and approved the direction on #4993. Once #4993 lands, this issue can close against it.

PR #4993, opened 2026-05-26
fix(core): warn on invalid RANK env values

fallintoplace:
## Summary
- warn when `RANK` is present but not parseable in both `safe_get_rank()` helpers
- preserve the existing fallback-to-`0` behavior instead of raising
- add focused regression coverage for the canonical helper and the Megatron-FSDP helper

## Motivation
This is the conservative follow-up discussed in #4991 and during review on #4959.

`safe_get_rank()` has historically treated malformed `RANK` values as a non-throwing fallback case. This change surfaces the misconfiguration with a warning while keeping existing callers on the same return-path contract.

## What changed
- `megatron/core/_rank_utils.py`
  - warn on invalid `RANK` values before falling back to `0`
  - keep the existing SLURM fallback and no-rank warning behavior
- `megatron/core/distributed/fsdp/src/megatron_fsdp/utils.py`
  - add the missing `os` import required by the fallback path
  - warn on invalid `RANK` values before falling back to `0`
- tests
  - add canonical helper coverage in `tests/unit_tests/test_utils.py`
  - add Megatron-FSDP helper coverage in `tests/unit_tests/distributed/megatron_fsdp/test_mcore_tensor_parallelism_detect.py`

## Validation
- `python3 -m py_compile` on all changed modules/tests
- direct function-level validation for both helpers:
  - valid `RANK` returns the parsed integer
  - invalid `RANK` emits a warning and returns `0`
- attempted targeted pytest collection for the Megatron-FSDP test file
  - blocked in this environment by an unrelated optional dependency import (`triton`) during module collection

## Notes
- This PR intentionally does **not** raise on invalid `RANK`; it keeps the longstanding non-throwing semantics and makes the fallback more visible instead.
- Related discussion: #4991


megatron/core/distributed/fsdp/src/megatron_fsdp/utils.py
```diff
@@ -471,11 +473,18 @@ def safe_get_rank() -> int:
...
+            return int(rank_env)
+        except (ValueError, TypeError):
+            warnings.warn(
+                f"Invalid RANK environment variable value {rank_env!r}. Defaulting to rank 0.",
+                stacklevel=2,
+            )
+            return 0
+
+    # Return rank 0 regardless of the actual rank.
+    return 0
```
wujingyue:
```suggestion
    return int(os.environ.get("RANK", 0))
```
I understood your intention to avoid semantic changes. But in this case I think it's a good thing to expose an invalid RANK as a hard error. And it simplifies too. 

(The same goes to the other helper.)

fallintoplace: Here is the commit that addressed this. [260fa78](https://github.com/NVIDIA/Megatron-LM/pull/4993/commits/260fa78db582dc075028ba5564df1264c606533e)

wujingyue (approved):

ericharper (approved):

ericharper:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4993/commits/260fa78db582dc075028ba5564df1264c606533e

ericharper:
Fixes #4991

ericharper:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/4993/commits/260fa78db582dc075028ba5564df1264c606533e

fallintoplace:
Let me know if I need to do something.

wujingyue:
> Let me know if I need to do something.

Can you fix the lint errors shown in the CI?

fallintoplace:
@wujingyue I will push the fix now.

Phlip79:
> @wujingyue I will push the fix now.

@fallintoplace I am using your PR as a test for a new automation we have. This new automation should be fixing the lint issue soon!

fallintoplace:
@Phlip79 I just pushed the lint fix, I apologize that I saw your message just now.

Phlip79:
> @Phlip79 I just pushed the lint fix, I apologize that I saw your message just now.

No worries! I just kicked off CI again.

guihong-nv:
@fallintoplace Please resolve the conflicts for merging.

fallintoplace:
Let me have a look and push the commit.

guihong-nv:
/ok to test [20f7c0e](https://github.com/NVIDIA/Megatron-LM/pull/4993/commits/20f7c0e9036304cece905ef068b571d320299200)

wujingyue:
@guihong-nv you could also comment `/claude fix` FYI

fallintoplace:
Should I rebase this? Thank you for your attention.

megatron/core/_rank_utils.py
```diff
@@ -28,21 +28,18 @@ def safe_get_rank() -> int:
...
-    except (ValueError, TypeError):
-        return 0
+    if "RANK" in os.environ:
+        return int(os.environ["RANK"])
+
+    slurm_rank = resolve_slurm_rank()
+    if slurm_rank is not None:
+        return slurm_rank
+
+    warnings.warn(
```
deepakn94, line 38: Can we decrease the log clutter here? Having this on every rank overwhelms our logfiles at large scale.

wujingyue: Good point in general, but this warning predates this change and its trigger is unchanged. Since rank discovery failed, we also cannot reliably restrict it to rank 0. I suggest handling log-volume reduction as a separate follow-up.

fallintoplace: Thanks. What should I do now?

wujingyue: cc @deepakn94

deepakn94 (approved):

PR #4818, opened 2026-05-15
[MXFP8] Mirror fixes in Mbridge for mxfp8 param gather 

zhongbozhu:
# What does this PR do ?

This PR is supposed to be dev-only, which mirrors this [commit ](https://github.com/NVIDIA/Megatron-LM/pull/4562/changes/b7a124349ba8b1babce17fab97fafa4953ffa236) from https://github.com/NVIDIA/Megatron-LM/pull/4562

Mbridge PR: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3727

## Issue tracking

For PRs from open-source community contributors:

- **New features**: a linked issue is **required**. Please open a [feature request](https://github.com/NVIDIA/Megatron-LM/issues/new?template=feature_request.md) and reference it here before submitting the PR.
- **Small updates (bug fixes, minor improvements)**: a linked issue is **recommended** and will accelerate the PR review process.

Linked issue: 

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

Once all required reviewers have approved, the `Approved` label is applied **automatically**.

### Merge

[...]

PR #4994, opened 2026-05-26
[Dev] Generalized fix for mxfp8 param gather 

zhongbozhu:

Assistant: GPT-5.5 xhigh 

# What does this PR do ?
Follow up of https://github.com/NVIDIA/Megatron-LM/pull/4818 https://github.com/NVIDIA/Megatron-LM/pull/4563 

This PR fixes a general correctness issue in the interaction between:

  --fp8-param-gather
  --reuse-grad-buf-for-mxfp8-param-ag
  --overlap-param-gather

With MXFP8 grad-buffer reuse, DDP param_data is not a persistent parameter buffer. It is a temporary high-precision all-gather staging buffer shared with the gradient buffer. After param all-gather completes, the gathered values are copied/quantized into TE MXFP8 parameter storage, and the shared buffer is zeroed so the next backward pass can safely accumulate gradients into it.

That means any explicit/forced param sync must first ask the optimizer to restage the updated FP32 master-weight shards into the DDP param buffer. Otherwise DDP may all-gather stale data or zeroes and overwrite model parameters.

Previously checkpoint path did roughly:
```
  if should_disable_forward_pre_hook(args):
      force_param_sync(model)

  save_checkpoint(...)
```

Then force_param_sync(model) called:
```
  model_chunk.start_param_sync(force_sync=True)
```

DDP does not all-gather from optimizer master weights. It all-gathers from the local shard view of `bucket.param_data`

So if nobody first did `_copy_main_params_to_param_buffer()`, then the checkpoint forced AG reads whatever is currently in the shared DDP staging buffer.

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


zhongbozhu:
cc @WanZzzzzz  @kunlunl

megatron/core/optimizer/optimizer.py
```diff
@@ -1271,6 +1275,43 @@ def load_state_dict(self, state_dict):
             optimizer.load_state_dict(state)
         self._synchronize_steps()
 
+    @override
```
zhongbozhu, line 1278: This is to avoid the below issue:
```
  dense optimizer prepare:
    zero whole model_chunk buffer
    copy dense ranges

  expert optimizer prepare:
    zero whole model_chunk buffer
    copy expert ranges
```
then the second zero wipes out dense ranges

WanZzzzzz: I don't think this will be an issue. For the same model chunk, the dense buffer and expert buffer are separate. 
My guess: your claude code takes the context of debugging the issue when dp overlap is off, where the dense optimizer syncs the dense+expert buffer and expert optimizer also syncs the dense+expert buffer. This is indeed a bug, but here we don't have this problem.

In short, I think the below code can be simpified.

megatron/core/optimizer/optimizer.py
```diff
@@ -1271,6 +1275,43 @@ def load_state_dict(self, state_dict):
...
+    @torch.no_grad()
+    def prepare_model_params_for_param_sync(self) -> None:
+        """Stage params once per DDP model chunk before explicit param sync."""
+        use_reused_grad_buffer = (
+            self.config.reuse_grad_buf_for_mxfp8_param_ag and self.config.overlap_param_gather
+        )
+        if not use_reused_grad_buffer:
```
WanZzzzzz, line 1285: Not sure why we need this block. When we don't use reuse and overlap, prepare_model_params_for_param_sync() will be a no-op. Why do we need to still call it?

zhongbozhu: yes, it will go to another prepare_model_params_for_param_sync in the base class, which will then be a no-op

megatron/core/optimizer/optimizer.py
zhongbozhu: I think numerically, you might be right that this isn't gonna be a issue.

But notice this code from Mcore main: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/training.py#L3687C1-L3691C75 

The current behavior is zero mode chunk once, and then copy data. If we simplify this code, we will zero the model chunk twice and adds unnecessary overhead?

WanZzzzzz:
LGTM. Tested convergence for DSv3:
image

kunlunl (approved):
LGTM

yaox12 (approved):

PR #5004, opened 2026-05-27
docs: document mmap_bin_files=False as parallel-FS mmap-quota mitigation

yaoyu-33:
## What

Expand the docstring of `BlendedMegatronDatasetConfig.mmap_bin_files` to call out that setting it to `False` is the right knob for **parallel-filesystem mmap-quota exhaustion**, not only for S3 / object storage.

Current docstring (one line):

> Whether to mmap the .bin files or use file pointers.

New docstring lists the two known cases where users should disable mmap:

1. Object storage backends (S3, etc.).
2. Parallel FS (Lustre, WekaFS, ...) when a blend contains many prefixes — the FS client can enforce a per-process live-mmap quota lower than `vm.max_map_count` and raise `PermissionError: [Errno 1] Operation not permitted` (EPERM) from `numpy.memmap` partway through `IndexedDataset` construction.

## Why

Docs-only. Came out of [NVIDIA-NeMo/Megatron-Bridge#3968](https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/3968), where a 444-prefix × {train, valid} = 888-dataset blend on a WekaFS mount reproducibly EPERM'd at `indexed_dataset.py:280` after ~590 successful builds — single-rank, no concurrency.

The reporter spent real time ruling out three other "DataLoader init speedup" flags (`path_to_cache`, `defer_npy_index_mmap`, `fast_cache_load`) — none of those reduce the per-prefix live-mmap count, which is what the FS client is actually sensitive to. Only `mmap_bin_files=False` works, because it swaps `_MMapBinReader` for `_FileBinReader` and halves the per-dataset mmap count (the `.idx` mmap in `_IndexReader.__init__` is still unconditional).

The current docstring gives users no hint that this is a possibility on parallel FSes, so they go down the wrong rabbit holes. This is the cheapest possible mitigation: document the workaround so the next person finds it from the config.

## Out of scope

- Making the `.idx` mmap optional (would need a new flag and is a bigger change — could be a follow-up).
- Auto-detecting parallel FS and flipping the default.
- Working with Weka on a client-side fix.

## Test

Docs-only, no code path changes. CI should be a no-op.


asolergi-nv:
Hey! Do you think it makes sense to also add this to our [dataloading docs](https://github.com/NVIDIA/Megatron-LM/blob/main/docs/user-guide/data-loading.md#flag-reference) too?

asolergi-nv (approved):

Phlip79 (approved):

jaredcasper (approved):

PR #5006, opened 2026-05-27
DDP wrap pg size fixes

maanug-nv:

# What does this PR do?

Fixes `_ddp_wrap` in `megatron/training/models/dist_utils.py` to use the explicitly-passed `ProcessGroupCollection` instead of falling back to `megatron.core.parallel_state` (`mpu`) globals.

### Changes

1. **Read process-group sizes/ranks from `pg_collection`, not `mpu`.** Three call sites in `_ddp_wrap` were querying `mpu.get_data_parallel_world_size(with_context_parallel=True)`, `mpu.get_pipeline_model_parallel_rank()`, and `mpu.get_expert_data_parallel_world_size()`. These now read from `pg_collection.dp_cp.size()`, `pg_collection.pp.rank()`, and `pg_collection.expt_dp.size()` respectively. The `mpu` import is dropped.

   The previous behavior silently ignored the caller's `pg_collection` for these three values, which meant non-default process groups (e.g. multi-module setups, or callers that build their own `ProcessGroupCollection`) got DDP bucket sizing and full-param layouts computed against the global mpu state rather than the groups they actually intended to use. And if the caller had not initialized the global parallel state at all, these `mpu` calls would have failed outright — even though a fully-populated `pg_collection` was already in hand.

2. **Forward `pg_collection` to `DistributedDataParallel`, not just megatron-FSDP.** The kwarg was previously gated on `if use_megatron_fsdp:`, so the plain `DistributedDataParallel` path never received it and had to fall back to its own mpu lookup. The gate is now `if not use_torch_fsdp2:` — DDP and megatron-FSDP both get it; TorchFSDP2 (which doesn't accept the kwarg) still skips.

## Issue tracking

Linked issue: N/A — small bug fix found while auditing `megatron.training.models` for stale `mpu` dependencies.

### Pre-checks

### DIFF
--- a/megatron/training/models/dist_utils.py
+++ b/megatron/training/models/dist_utils.py
@@ -8,7 +8,7 @@ logger = logging.getLogger(__name__)
 from typing import Any, Callable
 
 import torch
-from megatron.core import tensor_parallel, mpu
+from megatron.core import tensor_parallel
 from megatron.core.distributed import (
     DistributedDataParallel,
     DistributedDataParallelConfig,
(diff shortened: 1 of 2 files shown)


yaoyu-33 (approved):

wujingyue (commented):
Hey @maanug-nv , can you update the PR description? It's currently a template providing me no context :)

deepakn94 (approved):

PR #3386, opened 2026-02-12
Add E2E support for THD format

xiaoyao0115:
## Description

This PR adds **Sequence Packing (THD format)** E2E support to MCore. Dev branch PR:https://github.com/NVIDIA/Megatron-LM/pull/2924

The core missing functionalities of THD in MCore are:
* data iterator cannot handle THD meta data, like cu_seqlens, max_seqlens.
* num_microbatches is fixed.
* PackParams are not passing between PP ranks.

#### 1. Add a data_iterator wrapper (megatron/core/datasets/data_schedule.py::wrap_dataloader)

A wrapper function that intercepts the data iterator to perform scheduling and packing:

- **Schedule & Pack**: Extracts data from the data iterator, schedules sequences across DP×CP ranks, and packs them into microbatches with `cu_seqlens` metadata.
- **Returns packing results**: Returns the packed `num_microbatches` along with two parameters for FLOPs calculation: `num_total_tokens_this_global_batch` and `sequence_square_sum_this_global_batch`.
- **TP broadcast**: Broadcasts `num_microbatches` and FLOPs parameters across TP ranks since only TP rank 0 has access to the data iterator.
- **PP broadcast**: When using PP, middle PP stages (not first or last) require metadata (`cu_seqlens`, `cu_seqlens_padded`, `max_seqlen`, etc.) to be broadcast from PP rank 0 for correct computation.

#### 2. Mock SFT Dataset Support

Supports mock datasets for testing and benchmarking with configurable sequence length distributions.
There are two modes of mock sft dataset:
- **File mode**: Load sequence lengths from an external CSV, example json:
  ```json
  {"mode": "file", "path": "/path/to/seqlens.csv"}
  ```
- **Distribution mode**: Generate sequence lengths from a distribution (currently supports lognormal), example json:
  ```json
  {"mode": "distribution", "type": "lognormal", "min_seq_len": 1024, "max_seq_len": 8192, "mean_seq_len": 4096, "lognormal_sigma": 1.1}
  ```

#### Before vs After

```mermaid
graph LR
    subgraph Before
        A1[DataIterator] --> B1[get_batch]
        B1 --> C1[forward_backward]
        C1 --> D1[Fixed seq_len FLOPs]
    end
    subgraph After
        A2[DataIterator] --> W[wrap_dataloader]
        W -->|schedule + pack| B2[PackedDataIterator]
        W -->|broadcast| M[num_microbatches + flops_params]
        B2 --> C2[get_batch_for_sequence_packing]
[...]

PR #5008, opened 2026-05-27
[main] add thd sequence packing dispatcher support for main

HaochenYuan:

# What does this PR do ?
Corresponding main branch PR for dev branch [PR 4816](https://github.com/NVIDIA/Megatron-LM/pull/2924) & [PR 5048](https://github.com/NVIDIA/Megatron-LM/pull/5048).  

Previous dev [PR 2924](https://github.com/NVIDIA/Megatron-LM/pull/2924) added support for thd format in training, but for MoE dispatcher, only all2all type is supported. This PR adds the deepep & hybridep backend.  

While currently the thd e2e support [PR 3386](https://github.com/NVIDIA/Megatron-LM/pull/3386) is not merged, this PR will remain draft state until 3386 proceeds.

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


yaox12 (approved):

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1032,19 +1033,53 @@ def __init__(
...
-        self.routing_map = routing_map.reshape(num_tokens, self.num_experts)
-        self.token_probs = probs.reshape(num_tokens, self.num_experts)
+        self._original_num_tokens = num_tokens
+
+        padded_num_tokens = num_tokens
+        equalize_thd_token_counts = (
+            self.config.sequence_packing_scheduler is not None
+            or self.config.moe_hybridep_pad_variable_tokens
+        )
+        if equalize_thd_token_counts:
```
zhongbozhu: Maybe we should completely let `moe_hybridep_pad_variable_tokens` control the padding? In future work, once we add cuda graph support, this padding becomes a blocker for moe cuda graph because it has dynamic shapes. With THD cuda graph padding entire input sequence with zeros, padding for HybridEP is not going to be needed in the first place.

zhongbozhu (approved):
LGTM

yaoyu-33 (approved):

tests/unit_tests/models/test_hybrid_moe_model.py
```diff
@@ -319,6 +313,7 @@
     "use_transformer_engine_op_fuser": False,
     "moe_single_grouped_weight": False,
     "moe_single_grouped_bias": False,
+    "moe_hybridep_pad_variable_tokens": False,
```
yashaswikarnati: could we add a test with moe_hybridep_pad_variable_tokens=True as this PR introduces this code path ?

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -52,9 +53,12 @@ def get_buffer(group: torch.distributed.ProcessGroup, hidden_bytes: int):
         num_nvl_bytes = max(
             config.get_nvl_buffer_size_hint(hidden_bytes, group.size()), num_nvl_bytes
         )
-        num_rdma_bytes = max(
-            config.get_rdma_buffer_size_hint(hidden_bytes, group.size()), num_rdma_bytes
-        )
+        # Local-only EP groups do not need an RDMA buffer, and DeepEP builds
+        # without internode support may not expose RDMA size hints.
+        if group.size() > torch.cuda.device_count():
```
yashaswikarnati: Could you help me understand why we need this condition, seems a little brittle

Victarry (approved):
LGTM

megatron/core/transformer/moe/fused_a2a.py
HaochenYuan: Removed.

tests/unit_tests/models/test_hybrid_moe_model.py
HaochenYuan: UT added.

megatron/core/transformer/transformer_config.py
```diff
@@ -864,6 +864,12 @@ class TransformerConfig(ModelParallelConfig):
     moe_permute_fusion_into_hybridep: bool = False
     """Fuse token rearrangement ops during token dispatching for HybridEP."""
 
+    moe_hybridep_pad_variable_tokens: bool = False
```
jiemingz: Is this feature only intended to be used for sequence packing? If so it feels awkward that if a user were to enable a sequence packed dataset that produced uneven sequence lengths, they would also need to know about this knob (which doesnt mention anything about sequence packing)

I think we generally want to avoid creating new configs when we could otherwise reasonably infer them from elsewhere. Could we set this path by doing something like detecting the presence of `packed_seq_params` by plumbing it down the moe_layer?

zhongbozhu: Yes we expect users to turn this toggle on specifically for HybridEP + THD and we will guide our customers when should we turn this on. On the other hand, having packed sequence doesn't always mean we needs to pad if we apply offline packing and the dataset was created without the need to pad. 

However, the accuracy impact of doing offline packing & memory overhead concerns of always padding to max seq-len are often key factors for us to introduce this specific toggle to do minimum amount of padding while also being able to use HybridEP.

jiemingz: can we rename this and update the description to reflect this reasoning? maybe `moe_hybridep_pad_uneven_sequence_packed_inputs` ?

zhongbozhu: I agree that documentation might need to be updated, but `moe_hybridep_pad_variable_tokens` might be a better name because this is a general enough feature to handle dynamic sized inputs across EP ranks, while technically sequence packed inputs is orthogonal, because we can have the following two cases:
1) if I choose not to do seq-packing, but processing single document (accuracy concerns about offline seqpacking might lead to this), I will need this toggle to do minimum amount of padding. 
2) if I choose to do seq-packing, there are cases where my data distribution is good enough that it's actually acceptable to pad to MBS * seqlen because the utilization rate is very high, so that we don't really need to pad right before HybridEP, for example Nemotron RL seems to this case. 
Overall, I avoided to bring in seq-packing & thd into the naming in the first place.

jiemingz: then maybe: `moe_hybridep_pad_uneven_dispatch_inputs` ? to disambiguate we're padding the input tokens which vary across ranks

zhongbozhu: I am okay with this one

HaochenYuan: renamed.

jiemingz (approved):

yashaswikarnati (approved):

PR #4728, opened 2026-05-11
Fix Muon QKV split for gated attention

Moozy23232:
# What does this PR do ?

Fix Muon QKV splitting for gated-attention models where `attention_output_gate=True`.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

Muon's QKV split path assumed that `linear_qkv.weight` is laid out as three
blocks per query group:

```text
[q, k, v]
```

However, Megatron-Core's attention implementation increases `linear_qkv_out_dim`
when `attention_output_gate` is enabled:

```python
self.linear_qkv_out_dim = self.query_projection_size + 2 * self.kv_projection_size
if self.config.attention_output_gate:
    self.linear_qkv_out_dim += self.config.kv_channels * self.config.num_attention_heads
```

That makes the gated-attention layout effectively:

```text
[q_gate, q, k, v]
```

In Qwen3.5-style gated attention, this can produce a `linear_qkv.weight` first
dimension such as `4608 = 2048 + 2048 + 256 + 256`. The old Muon split shape
was still computed as `[2048, 256, 256]`, which can make the reshape in Muon
orthogonalization invalid.

This PR:

- computes gated-QKV split shapes as `[q_gate, q, k, v]` when
  `attention_output_gate=True`;
- records per-parameter `qkv_split_shapes` and `muon_param_name` metadata on
  valid `linear_qkv.weight` parameters;
- uses per-parameter split metadata in the Muon orthogonalization path;
- preserves Muon QKV metadata when tensor-parallel attributes are copied to
  optimizer/master parameters;
- adds unit coverage for gated-QKV split shapes and metadata preservation.

## Issue tracking

Linked issue: Fixes #4731 

## Root cause

The failing case is structural rather than specific to one launcher. I first hit
it while training a Qwen3.5-style MoE model with:

```text
optimizer: dist_muon
muon_split_qkv: true
attention_output_gate: true
```

For this class of model, `linear_qkv.weight` includes a query-sized gate block in
addition to the regular Q, K, and V blocks. The Muon split code did not account
for this gated-QKV layout and kept using a 3-way QKV split.

There is also a metadata propagation issue: `copy_tensor_model_parallel_attributes()`
previously copied `is_qkv`, but did not copy `qkv_split_shapes` or a parameter
name/debug identifier. This can lose the information needed by the optimizer
when operating on copied optimizer/master parameters.

## Minimal reproduction

[...]

PR #5015, opened 2026-05-27
Add MLA support to Muon

### LINKED ISSUE #4091: Add MLA support for Muon optimizer
Currently, the Muon optimizer implementation in `megatron/core/optimizer/muon.py` lacks support for MLA (Multi-Latent Attention) parameters, as indicated by the TODO comment (`TODO(deyuf): support MLA`). 

MLA-related parameters are not tagged with `is_qkv` and thus undergo direct orthogonalization. Maybe it lead to some unpredictable result.

Add MLA support to the Muon optimizer.

janEbert:

# What does this PR do ?

Add MLA support to Muon by extending the existing metadata tagging scheme for splitting the MLA up-projections (and the sole Q-projection when `q_lora_rank=None`).

The down-projs are deliberately left out of the scope of this PR and will be included in a follow-up PR with the down-proj splitting being off-by-default behind an additional feature flag.

## Issue tracking

Fix #4091.

**Rebased on top of https://github.com/NVIDIA/Megatron-LM/pull/4728.**

### DIFF
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -725,7 +725,72 @@ def check_config_overrides_consistency(
 def _tag_muon_split_qkv_parameters(
     config: OptimizerConfig, model_chunks: list[MegatronModule]
 ) -> None:
-    """Tag standard QKV and MLA up-projection weights with Muon split metadata."""
+    """Tag standard QKV and MLA up-projection weights with Muon split metadata.
+
+    Note that MLA down-projection weights aren't split with the current implementation.
+
+    The applied tags are:
+
+    - `is_qkv` (bool): whether the layer is a fused/packed collection of standard Multi-Head
+      Attention Q, K, and V projections. The layer contains a single weight, which includes all of
+      the Q, K, and V projections.
+    - `qkv_split_shapes` (tuple[int, ...] | None): per-group row counts for the projection slices
+      packed in this weight, if applicable. The gradients for the tagged parameter will be split
+      along the row dimension. The splits will be orthogonalized independently.
+
+      For example, without an Attention output gate, we have the following for the fused/packed
+      standard Multi-Head Attention QKV projection when not using tensor model parallelism:
+
+      ```
+      # Remember that PyTorch uses transposed weights.
+      linear_qkv.weight.shape == (
+          (
+              num_attention_heads * kv_channels  # <- Qs
+              + kv_channels * num_query_groups  # <- Ks
+              + kv_channels * num_query_groups  # <- Vs
+          ),
+          hidden_size,
+      )
+
+      linear_qkv.qkv_split_shapes = (
+          num_attention_heads // num_query_groups * kv_channels,  # <- Q
+          kv_channels,  # <- K
+          kv_channels,  # <- V
+      )
+      ```
+
+      In the optimizer step, we transform the gradient like the following:
+      ```
+      # The initial shape is like `linear_qkv.weight.shape` before.
+      linear_qkv.grad.shape == linear_qkv.weight.shape
+
+      qs_grad, ks_grad, vs_grad = muon_split_grads(linear_qkv.grad, linear_qkv.qkv_split_shapes)
+
+      qs_grad.shape == (
+          num_query_groups,
+          num_attention_heads // num_query_groups * kv_channels,
+          hidden_size,
+      )
+      ks_grad.shape == vs_grad.shape == (
+          num_query_groups,
+          kv_channels,
+          hidden_size,
+      )
+
+      # Contract first two dimensions.
+      qs_grad = qs_grad.reshape(num_attention_heads * kv_channels, hidden_size)
+      qs_grad = orthogonalize(qs_grad)
+
+      ks_grad = ks_grad.reshape(num_query_groups * kv_channels, hidden_size)
+      ks_grad = orthogonalize(ks_grad)
+
+      vs_grad = vs_grad.reshape(num_query_groups * kv_channels, hidden_size)
+      vs_grad = orthogonalize(vs_grad)
+
+      result_grad = muon_concat_grads(qs_grad, ks_grad, vs_grad)
+      result_grad.shape == linear_qkv.grad.shape
+      ```
+    """
     split_qkv = getattr(config, 'muon_split_qkv', True)
 
     def _is_muon_split_metadata_managed_param(
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -4,7 +4,7 @@ import logging
 import warnings
 from collections import defaultdict
 from dataclasses import astuple
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Union
 
 import torch
 from torch.optim import SGD as CPUSGD
@@ -728,14 +728,16 @@ def _tag_muon_split_qkv_parameters(
     """Tag standard QKV and MLA up-projection weights with Muon split metadata."""
     split_qkv = getattr(config, 'muon_split_qkv', True)
 
-    def _is_muon_split_metadata_managed_param(name: str, is_mla: bool) -> bool:
+    def _is_muon_split_metadata_managed_param(
+        name: str, attn_variant: Literal["mha", "mla"]
+    ) -> bool:
         """Return whether the parameter with the given name should be managed by Muon split tags.
 
         E.g., standard QKV or MLA projection weights.
         """
         if 'linear_qkv.weight' in name:
             return True
-        if not is_mla:
+        if attn_variant != "mla":
             return False
         return any(
             f'{projection}.weight' in name
@@ -751,11 +753,11 @@ def _tag_muon_split_qkv_parameters(
 
     for model_chunk in model_chunks:
         model_cfg = get_model_config(model_chunk)
-        is_mla = getattr(model_cfg, 'multi_latent_attention', False)
+        attn_variant = "mla" if getattr(model_cfg, 'multi_latent_attention', False) else "mha"
         standard_qkv_split_shapes = None
         mla_q_split_shapes = None
         mla_kv_split_shapes = None
-        if is_mla:
+        if attn_variant == "mla":
             mla_q_split_shapes = (model_cfg.qk_head_dim, model_cfg.qk_pos_emb_head_dim)
             mla_kv_split_shapes = (model_cfg.qk_head_dim, model_cfg.v_head_dim)
 
@@ -763,7 +765,7 @@ def _tag_muon_split_qkv_parameters(
             if not param.requires_grad:
                 continue
 
-            is_managed_param = _is_muon_split_metadata_managed_param(name, is_mla)
+            is_managed_param = _is_muon_split_metadata_managed_param(name, attn_variant)
             if is_managed_param:
                 # Remove existing Muon split tags from a parameter.
                 for attr in ("is_qkv", "qkv_split_shapes"):
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -722,6 +722,69 @@ def check_config_overrides_consistency(
     return True
 
 
+def _tag_muon_split_qkv_parameters(
+    config: OptimizerConfig, model_chunks: list[MegatronModule]
+) -> None:
+    """Tag standard QKV and MLA up-projection weights with Muon split metadata."""
+    split_qkv = getattr(config, 'muon_split_qkv', True)
+
+    def _is_muon_split_metadata_managed_param(name: str, is_mla: bool) -> bool:
+        """Return whether the parameter with the given name should be managed by Muon split tags.
+
+        E.g., standard QKV or MLA projection weights.
+        """
+        if 'linear_qkv.weight' in name:
+            return True
+        if not is_mla:
+            return False
+        return any(
+            f'{projection}.weight' in name
+            for projection in (
+                'linear_q_proj',
+                'linear_q_up_proj',
+                'linear_kv_up_proj',
+                'linear_q_down_proj',
+                'linear_kv_down_proj',
+                'linear_qkv_down_proj',
+            )
+        )
+
+    for model_chunk in model_chunks:
+        model_cfg = get_model_config(model_chunk)
+        is_mla = getattr(model_cfg, 'multi_latent_attention', False)
+        standard_qkv_split_shapes = None
+        mla_q_split_shapes = None
+        mla_kv_split_shapes = None
+        if is_mla:
+            mla_q_split_shapes = (model_cfg.qk_head_dim, model_cfg.qk_pos_emb_head_dim)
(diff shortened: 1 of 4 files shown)


ericharper:
@Phlip79 , this code is not being covered by an expert group. Can we cover it?

megatron/core/optimizer/__init__.py
```diff
@@ -721,6 +722,69 @@ def check_config_overrides_consistency(
     return True
 
 
+def _tag_muon_split_qkv_parameters(
+    config: OptimizerConfig, model_chunks: list[MegatronModule]
+) -> None:
+    """Tag standard QKV and MLA up-projection weights with Muon split metadata."""
+    split_qkv = getattr(config, 'muon_split_qkv', True)
+
+    def _is_muon_split_metadata_managed_param(name: str, is_mla: bool) -> bool:
```
wujingyue: https://www.teamten.com/lawrence/programming/prefer-enums-over-booleans.html

Consider `Literal["mla", "mha"]` or an enum. Otherwise, it's unclear what's the opposite for MLA.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -249,28 +301,9 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
-
-            qkv_grads = [
-                self.scaled_orthogonalize_fn(g, tp_group, partition_dim).view(
-                    num_query_groups, -1, grad_shape[-1]
-                )
-                for g in qkv_grads
-            ]
-            grad = torch.cat(qkv_grads, dim=1).view(grad_shape)
+        split_shapes = self._get_muon_split_shapes(p) if self.split_qkv else None
+        if split_shapes is not None:
```
wujingyue:
```
if self.split_qkv:
  split_shapes = ...
  assert split_shapes
  grad = ...
```

megatron/core/optimizer/__init__.py
```diff
@@ -721,6 +722,69 @@ def check_config_overrides_consistency(
     return True
 
 
+def _tag_muon_split_qkv_parameters(
+    config: OptimizerConfig, model_chunks: list[MegatronModule]
+) -> None:
+    """Tag standard QKV and MLA up-projection weights with Muon split metadata."""
```
wujingyue: More code comments could help. I've yet to understand the expected behavior of the function, e.g., what tags are applied to what modules and why.

wujingyue (commented):
Getting some expert reviews can help too.

megatron/core/optimizer/emerging_optimizers.py
janEbert: I like the simplification idea, but this breaks for parameters which don't have the `muon_split_shapes` attribute and for which `is_qkv_fn` is either `None` or evaluates to `False` (i.e., any non-QKV params).
In other words, `TensorParallelMuon._get_muon_split_shapes` can return `None` without it being an error, which the suggestion wouldn't allow.

megatron/core/optimizer/__init__.py
janEbert: In favor of keeping this PR small, I've gone with the inexplicit enum version with `Literal` for now.

I think it makes sense to introduce an enum for all sequence and channel mixer layers that Megatron-LM contains in the future (in a more accessible location such as `megatron/core/enums.py`). Possibly with a registry that is automatically updated when a new `SequenceMixer` subclass is defined. That enum could also be used here.

megatron/core/optimizer/__init__.py
janEbert: I've added additional documentation including some pseudocode to better explain the behavior. Let me know if it's good or maybe too much.

I also realized, maybe it makes sense to refactor out this whole tagging logic into a new module. Then we could also improve the type-safety for the attributes without blowing up `optimizer/__init__.py` even more. Could even consider adding a whole new struct just for the Muon-related attributes to modularize them. I'd also consider that out of scope for this PR, though.

megatron/core/optimizer/emerging_optimizers.py
janEbert: However, coming back to this, I've realized that – assuming we don't need the custom behavior it enables – the `TensorParallelMuon.qkv_split_shapes` attribute is now redundant and could be removed, since `muon_split_shapes` should always be set for any matching parameter. Through this, the checks surrounding `is_qkv_fn` in `_get_muon_split_shapes` could also be removed, simplifying that function. That is, unless we need the behavior all this enables of supporting other custom `is_qkv_fn` functions and `qkv_split_shapes`.

@FDecaYed and/or @skyw could you comment on this since you seem to have implemented the design? Otherwise I'd also keep the above in mind for a future simplification.

skyw: The comment was about
```python
a = foo() if condition else None
if a is not None:
    # do something
```
A less error prone and more readable way is
```python
if condition:
    a = foo()
    assert a is not None
    # do something
```

It is less about split qkv logic but pure code style.

wujingyue: > It is less about split qkv logic but pure code style.

That's right. Here I'm simply proposing to combine the two `if`s into one. 

(My rule of thumb for measuring code complexity is counting the number of if-conditions! This is because humans are bad at reasoning about branches.)

> In other words, TensorParallelMuon._get_muon_split_shapes can return None

TensorParallelMuon._get_muon_split_shapes can return None **even when** self.split_qkv is true? Otherwise, my pseudo-code ought to work.

megatron/core/optimizer/__init__.py
```diff
@@ -721,6 +722,136 @@ def check_config_overrides_consistency(
...
+    def _is_muon_split_metadata_managed_param(
+        name: str, attn_variant: Literal["mha", "mla"]
+    ) -> bool:
+        """Return whether the parameter with the given name should be managed by Muon split tags.
+
+        E.g., standard QKV or MLA projection weights.
+        """
+        if 'linear_qkv.weight' in name:
+            return True
+        if attn_variant != "mha":
```
gaojingwei: if attn_variant=“mla”，this place will return False，the below code will run contine，so the param can not set the muon_split_shapes.

 if not is_managed_param or not split_qkv or len(param.shape) != 2:
                continue

megatron/core/optimizer/emerging_optimizers.py
janEbert: I'm not sure if we have a misunderstanding. I understand your change was a simplification suggestion. However, it doesn't work with the current code logic, since `a` can legally be `None` even if `condition=True`. That's what I tried to state in the first comment.

My second comment was simply a way to (1) make the suggestion work and (2) simplify the surrounding code even further. So I wish there was feedback on that instead. :)

>> In other words, TensorParallelMuon._get_muon_split_shapes can return None
>
> TensorParallelMuon._get_muon_split_shapes can return None **even when** self.split_qkv is true? Otherwise, my pseudo-code ought to work.

Yes, that's what I tried to explain in the first comment. In fact, this is the normal behavior for all the parameters that should not be split.

megatron/core/optimizer/__init__.py
janEbert: Thanks a lot for noticing this, fixed the typo.

megatron/core/optimizer/emerging_optimizers.py
wujingyue: > TensorParallelMuon._get_muon_split_shapes can return None even when self.split_qkv is true

Thanks for clarifying, @janEbert! 

@skyw Does this sound like a bug? How could Muon know how to split without split shapes? 😕

janEbert: Sorry, it seems like it's still unclear. Any parameter can be passed to `TensorParallelMuon._get_muon_split_shapes`. However, only packed parameters, in which a single tensor is used to represent different functionality, are actually split. E.g., the Q, K, and V projections are commonly packed into the weight of a single `Linear` layer instead of three separate ones. Parameters for which it doesn't make sense to split them (e.g., any of the layers in the Transformer block MLP) don't have a split shape and thus we return `None`.

Basically, `TensorParallelMuon._get_muon_split_shapes` combines (a) checking for whether the passed parameter is splittable (which we _need_ to check because of the customization `TensorParallelMuon.is_qkv_fn` and `TensorParallelMuon.qkv_split_shapes` allow) and (b) returning the parameter's split shape, or `None` if the parameter isn't splittable. The returned value encodes whether the parameter is splittable or not and what its split shape is.

FDecaYed (commented):
the code change itself looks ok, but at high level I feel splitting fused should be a general thing we support with a refined API. we cannot sustain keep adding new cases as the workload coming. Ideally we should generalize this.

Also kinda related, we had the change in Emerging-optimizers to support batched muon processing. instead of supporting arbitrary batch of different shaped weights, it only support batch of weight all with same shape, and it could be useful for cases where same shape weights are fused in the first place

ericharper (approved):

skyw:
> the code change itself looks ok, but at high level I feel splitting fused should be a general thing we support with a refined API. we cannot sustain keep adding new cases as the workload coming. Ideally we should generalize this.
> 
> Also kinda related, we had the change in Emerging-optimizers to support batched muon processing. instead of supporting arbitrary batch of different shaped weights, it only support batch of weight all with same shape, and it could be useful for cases where same shape weights are fused in the first place

Specific for that, the plan is for a version of grouped linear has 3d batch shaped weights in TE or other places.

megatron/core/optimizer/emerging_optimizers.py
skyw: > Sorry, it seems like it's still unclear. Any parameter can be passed to `TensorParallelMuon._get_muon_split_shapes`. However, only packed parameters, in which a single tensor is used to represent different functionality, are actually split. E.g., the Q, K, and V projections are commonly packed into the weight of a single `Linear` layer instead of three separate ones. Parameters for which it doesn't make sense to split them (e.g., any of the layers in the Transformer block MLP) don't have a split shape and thus we return `None`.
> 
> Basically, `TensorParallelMuon._get_muon_split_shapes` combines (a) checking for whether the passed parameter is splittable (which we _need_ to check because of the customization `TensorParallelMuon.is_qkv_fn` and `TensorParallelMuon.qkv_split_shapes` allow) and (b) returning the parameter's split shape, or `None` if the parameter isn't splittable. The returned value encodes whether the parameter is splittable or not and what its split shape is.

in code like
```python
if condition:
    a = foo()
    assert a is not None
    # do something
```
If a can be None which has specific meaning, it is recommended to nest
```python
if condition:
    a = foo()
    if a is not None:
      # do something
    else:
      # bar?
```
should also note that although the code can work, it suggests sub-optimal logic, i.e. overloaded meanings of None

janEbert: But then what we'd end up with is the following:

Currently:

```python
if condition:
    a = an_optional()
else:
    a = None
if a is not None:
    do_something()
else:
    do_something_else()
```

With your suggestion, we would get:

```python
if condition:
    a = an_optional()
    if a is not None:
        do_something()
    else:
        do_something_else()
else:
    do_something_else()
```

Alternatively, we could avoid the repetition with a boolean:

```python
should_do_something_else = not condition
if condition:
    a = an_optional()
    if a is not None:
        do_something()
    else:
        should_do_something_else = True
if should_do_something_else:
    do_something_else()
```

I still fail to see the improvement here. But if you'd prefer we go in this direction, which of the two solutions would you prefer?

Or should we get rid of the `None` entirely (i.e., make the return type non-`Optional`)? We can refactor out the checking logic, doing it in a separate function, and refactor the existing method (i.e., `TensorParallelMuon._get_muon_split_shapes`, or `do_something()` in the examples), so that it must not return an `Optional`. Again, I'd also like to go back to the questions for enabling a simplification in https://github.com/NVIDIA/Megatron-LM/pull/5015/files#r3324076370.
What I don't like about that way of doing things is that the current optional typing of the `TensorParallelMuon.qkv_split_shapes` attribute lends itself well to the current logic, since we need to access and query the attribute anyway. I also don't see `None` being overloaded here. It signals the absence of a split shape for the given parameter, which, to my understanding, is what optionality is about.

skyw: Although it is ok to have (many code does that)
```python
if condition:
    a = an_optional()
else:
    a = None
if a is not None:
    do_something()
else:
    do_something_else()
```
But note that "if a is not None:" implies "condition" implicitly, which is less readable.

Agreed on qkv split part is not best designed, Deyu also expressed similar idea. We should revisit it outside of this PR.

janEbert: Agree with the implicitness. I think I prefer more "flat" code, since I consider nesting more detrimental to readability than the implicitness. I'll change it to the version with the boolean I suggested above to avoid the implicitness. Thank you for the input!

FDecaYed (approved):

megatron/core/optimizer/emerging_optimizers.py
janEbert: Applied the suggestion.

```patch
-        split_shapes = self._get_muon_split_shapes(p) if self.split_qkv else None
-        if split_shapes is not None:
-            grad = self._orthogonalize_split_grad(grad, split_shapes, tp_group, partition_dim)
-        else:
+        grad_was_split = False
+        if self.split_qkv:
+            split_shapes = self._get_muon_split_shapes(p)
+            if split_shapes is not None:
+                grad = self._orthogonalize_split_grad(grad, split_shapes, tp_group, partition_dim)
+                grad_was_split = True
+        if not grad_was_split:
             grad = self.scaled_orthogonalize_fn(grad, tp_group, partition_dim)
```

yaoyu-33 (approved):

buptzyb:
@janEbert Hi, any updates on this PR? Are we going to merge it? And what's the plan for the KV down proj split? Thanks!

janEbert:
Hey @buptzyb, there are sadly still missing expert reviews on this PR. As for the down-proj split, I didn't want to work on it before this PR is actually merged.

FDecaYed (approved):

mkhona-nvidia:
hi @janEbert 

can we merge this PR and https://github.com/NVIDIA/Megatron-LM/pull/6326 ?

The `split_qkv` feature of Muon has been a point of contention for a while and newer muon runs do per-head splitting. We can take the opportunity to improve this `split_qkv` feature

janEbert:
Sounds good, but what do you mean by "merge"? It feels like both our PRs don't really address the design point, if I see correctly.

mkhona-nvidia:
> Sounds good, but what do you mean by "merge"? It feels like both our PRs don't really address the design point, if I see correctly.

Sorry by merge, I meant unify the PRs into a common interface that is more general than either of the 2 PRs

janEbert:
I see, yeah, sounds good! Do you (or some of the others already in the discussion) already have something in mind? If not, I can give it some thought.

mkhona-nvidia:
> I see, yeah, sounds good! Do you (or some of the others already in the discussion) already have something in mind? If not, I can give it some thought.

No one else is working on it, though @philipcmonk might have thoughts. There's also an interaction with gated attention, where we split into `QKV` and also the `output_gate`

PR #5016, opened 2026-05-27
Fix Megatron-FSDP safe_get_rank fallback

### LINKED ISSUE #4958: [BUG] Megatron-FSDP safe_get_rank() fallback raises NameError because os is not imported

Devil1716:
## Summary
- import os in Megatron-FSDP utils so safe_get_rank() can read RANK when torch.distributed is not initialized
- add regression coverage for valid, missing, and invalid RANK fallback behavior

Fixes #4958

## Testing
- python -m py_compile megatron/core/distributed/fsdp/src/megatron_fsdp/utils.py tests/unit_tests/distributed/megatron_fsdp/test_mcore_tensor_parallelism_detect.py
- git diff --check

Could not run the focused pytest locally because this environment is missing pytest and torch.

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/utils.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/utils.py
@@ -16,6 +16,7 @@ import contextlib
 import inspect
 import logging
 import operator
+import os
 from contextlib import nullcontext
 from functools import reduce
 from importlib.metadata import version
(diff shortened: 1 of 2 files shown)


wujingyue:
Similar PRs are already approved. https://github.com/NVIDIA/Megatron-LM/pull/4959

PR #5023, opened 2026-05-27
fix(test): widen iter-time steady-state window for short tests

ko3n1g:

## What

Make the iter-time steady-state window adaptive so it does not break tests that run fewer than 30 steps.

## Why

[`#5010`](https://github.com/NVIDIA/Megatron-LM/pull/5010) (commit `71223d5`) hard-coded the iter-time aggregation window to `range(30, 46)` to avoid the warmup step dominating the median. For tests that train for only ~25 steps (e.g. `gpt3_7b_tp1_pp4_memory_speed`, which sets `--train-iters: 25` and `--exit-interval: 25`), no step lands in that window, so both `actual_value_list` and `golden_value_list` end up empty. `np.median([])` returns `NaN`, `np.isclose(NaN, NaN)` is `False`, and the test fails with no useful diagnostic.

## How

In `tests/functional_tests/python_test_utils/common.py:pipeline`, derive the window from the highest step present in the golden values:

```python
max_golden_step = max(golden_value.values.keys()) if golden_value.values else 0
steady_window = range(5, 21) if max_golden_step <= 25 else range(30, 46)
```

- Short tests (≤25 logged steps) → `range(5, 21)` — skips the first warmup step, averages over ~15 measured points.
- Everything else → `range(30, 46)` — original behaviour, untouched.

## Test plan

- [ ] Functional CI green on `gpt3_7b_tp1_pp4_memory_speed`.
- [ ] No regression on long iter-time tests (median identical to pre-change, since `range(30, 46)` path is taken).

</details>

### DIFF
--- a/tests/functional_tests/python_test_utils/common.py
+++ b/tests/functional_tests/python_test_utils/common.py
@@ -208,9 +208,8 @@ def pipeline(
                 ]
 
                 if metric_name == "iteration-time":
-                    # Restrict iter-time aggregation to the steady-state window
-                    # (steps 30-45) so the warmup step does not dominate the median.
-                    steady_window = range(30, 46)
+                    max_golden_step = max(golden_value.values.keys()) if golden_value.values else 0
+                    steady_window = range(5, 21) if max_golden_step <= 25 else range(30, 46)
                     actual_value_list = [
                         value
                         for value_step, value in actual_values[metric_name].values.items()


tests/functional_tests/python_test_utils/common.py
```diff
@@ -208,9 +208,8 @@ def pipeline(
                 ]
 
                 if metric_name == "iteration-time":
-                    # Restrict iter-time aggregation to the steady-state window
-                    # (steps 30-45) so the warmup step does not dominate the median.
-                    steady_window = range(30, 46)
+                    max_golden_step = max(golden_value.values.keys()) if golden_value.values else 0
```
wujingyue, line 211:
```suggestion
                    max_golden_step = max(golden_value.values.keys(), default=0)
```

tests/functional_tests/python_test_utils/common.py
```diff
@@ -208,9 +208,8 @@ def pipeline(
...
+                    steady_window = range(5, 21) if max_golden_step <= 25 else range(30, 46)
```
wujingyue, line 212: Sounds a bit arbitrary. Conceptually, if warmup is still ongoing after iteration 5, steady_windows isn't steady. If not, we should just set the range to `range(5, 21)` instead of an if-then-else. Wdyt?

wujingyue (approved):

balasaajay (approved):

tests/functional_tests/python_test_utils/common.py
ko3n1g: This PR is a stop gap to another issue, so I’ll need to merge it. But open to discuss another follow-up. 

Usually our tests run for 50 steps, so we take 30:46 for averaging. 
however some tests only for 25 steps, so the best I could come up with was to choose 5:21. 

Not sure how to de-select noise steps better for those tests..

wujingyue: > This PR is a stop gap to another issue, so I’ll need to merge it

Sure. 

> Not sure how to de-select noise steps better for those tests..

In that case, the warmup threshold should probably be set by individual tests rather than being speculated by the test infra here.

PR #5047, opened 2026-05-28
Fix MoE aux_loss / z_loss gradient scaling with TP > 1

deepakn94:
## Summary

When `--calculate-per-token-loss` is set, MoE `aux_loss` and `z_loss` were scaled by `num_local_tokens` before being handed to `MoEAuxLossAutoScaler`. With `tp_cp_group.size() > 1`, this leaves the resulting gradient contribution dependent on TP/CP size, producing a seq-loss discrepancy between otherwise-identical runs with different TP. The first commit multiplies the scale by `tp_cp_group.size()` so the effective per-parameter scaling is invariant to TP and CP.

The first commit is a cherry-pick of internal commit `ed6fe919` ([gitlab link](https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/commit/ed6fe9190828088eeb3914276418b73afc4b4e42)).

## Commits

1. **Cherry-pick** — fix the scaling, keeping the in-comment rationale identical to the internal commit.
2. **Comment clarification** (no behavioural change) — restructures the aux_loss and z_loss comments around the two concrete quantities a reader needs to follow the fix: (a) the target final scaling on these gradients, `1 / (num_micro_batches * dp_cp_size)`, matching the `!calculate_per_token_loss` path; (b) what `--calculate-per-token-loss` already divides by — every parameter gradient is divided by `total_global_tokens` in `finalize_model_grads`, which under uniform tokens is `num_micro_batches * dp_cp_size * num_local_tokens`. From there the `num_tokens * tp_cp_group.size()` pre-multiply falls out as the cancellation that takes the actual divisor to the target. Also splits the "TP+CP gradient SUM all-reduce" shorthand into the two distinct reductions that actually run (TP-replicated sequence-parallel grads on the TP group in `_allreduce_non_tensor_model_parallel_grads`; CP grads as part of the DP+CP grad sync), and notes that the pre-existing `/tp_cp_group.size()` on `moe_z_loss_coeff` is a separate forward-side averaging correction.

## Test plan
- [ ] Compare loss curves for a small MoE config with TP=1 vs TP>1 and `--calculate-per-token-loss` set, before and after this change.
- [ ] CI: `Run functional tests` label (touches MoE routing — could affect numerics).

### DIFF
--- a/megatron/core/transformer/moe/router.py
+++ b/megatron/core/transformer/moe/router.py
@@ -479,15 +479,15 @@ class TopKRouter(Router):
             needs_dp_avg=needs_dp_avg,
         )
         if self.calculate_per_token_loss:
-            # Scale the aux_loss by the number of tokens.
+            # Scale the aux_loss by the global number of tokens (across TP and CP ranks).
             # The expected final scaling for aux_loss gradients is 1/(num_micro_batches * dp_size).
-            # After commit 02648000, Megatron started using the number of total tokens to scale
-            # gradients under the argument of calculate_per_token_loss,
-            # which scales both the main_loss gradient and aux_loss gradient by
-            # 1/(num_local_tokens * dp_size * num_micro_batches) in finalize_model_grads function.
-            # To correct this scaling, we need to scale the aux_loss by num_local_tokens here.
+            # finalize_model_grads divides all gradients by total_global_tokens (all-reduced across
+            # DP+CP but not TP). To correct this, we scale the aux_loss by the global token count
+            # (local tokens * tp_cp_group size) so that after TP+CP gradient SUM all-reduce and
+            # the finalize_model_grads division, the effective scaling is correct.
             # Use valid_token_count (excluding padding) if provided, otherwise use total tokens.
             num_tokens = valid_token_count if valid_token_count is not None else activation.shape[0]
+            num_tokens = num_tokens * self.tp_cp_group.size()
             activation = MoEAuxLossAutoScaler.apply(activation, aux_loss * num_tokens)
         else:
             activation = MoEAuxLossAutoScaler.apply(activation, aux_loss)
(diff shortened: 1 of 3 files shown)


megatron/core/transformer/moe/router.py
```diff
@@ -479,16 +479,36 @@ def attach_and_log_load_balancing_loss(
...
+            # So pre-multiplying aux_loss by num_local_tokens * tp_cp_group.size() cancels
+            # that same factor in total_global_tokens above, leaving 1 / (num_micro_batches *
+            # dp_size) as the effective scaling on the aux_loss gradient — the target.
             # Use valid_token_count (excluding padding) if provided, otherwise use total tokens.
-            num_tokens = valid_token_count if valid_token_count is not None else activation.shape[0]
-            activation = MoEAuxLossAutoScaler.apply(activation, aux_loss * num_tokens)
+            num_local_tokens = (
+                valid_token_count if valid_token_count is not None else activation.shape[0]
+            )
+            activation = MoEAuxLossAutoScaler.apply(
```
yashaswikarnati, line 509: would this assume valid token count is same on different tp cp ranks? In SFT/multimodal cases this might be different I guess

Victarry (approved):
LGTM. 
I did similar fix in https://github.com/NVIDIA/Megatron-LM/pull/3768/changes and verified that with this fix the load balancing loss results similar loss curve under difference parallelism settings.

deepakn94:
> LGTM. I did similar fix in https://github.com/NVIDIA/Megatron-LM/pull/3768/changes and verified that with this fix the load balancing loss results similar loss curve under difference parallelism settings.

Maybe we should port this commit to dev instead of !3768 to make sure you don't run into issues when `main` is merged into `dev` next.

ericharper (approved):

kvareddy (approved):

PR #5060, opened 2026-05-29
use param_group_name as the canonical identifier for param groups

yezhengmao1:
# What does this PR do ?

  Make optimizer param-group identity stable under µP — where multiple
  groups intentionally share `wd_mult` / `lr_mult` but have different
  LR schedules.

  ## Why

  Group identity used to be the tuple `(wd_mult, lr_mult, is_expert_parallel, is_decoupled_lr)`.

  Under µP, per-region LR differences are expressed via `max_lr` / `min_lr` /
  `init_lr` (the schedule), not via `lr_mult` — and `wd` is typically held
  the same across regions. So semantically distinct groups (e.g. embedding /
  hidden / output / layernorm) all collapse to the same identifier:

  | group | wd_mult | lr_mult | is_ep | is_decoupled_lr | max_lr |
  |---|---|---|---|---|---|
  | hidden | 1.0 | 1.0 | F | F | 3e-4 |
  | embedding | 1.0 | 1.0 | F | F | 6e-4 |
  | output | 1.0 | 1.0 | F | F | 1e-4 |

  → identifier tuple collides → checkpoint reload matches the wrong group
  (or fails).

  ## What

  Switch identity to a stable **name** computed from the matched
  `ParamKey`s, not from override values:

  - New `ParamKey.canonical_name()` — derived from `name` / `attr` /
    `predicate` / `with_name_predicate`, independent of any numeric
    override.
  - `_get_param_groups` records the matched `ParamKey.canonical_name()`s
    per group, all-gathers across ranks (PP/EP shards each see only a
    subset of params), and writes `param_group_name` on the group.
  - `param_group_identifier_keys = ('param_group_name', 'is_expert_parallel')`.

  Groups remain distinguishable regardless of how identical their numeric
  overrides happen to be.

 ### Pre-checks

  - [x] I have added proper typing
  - [ ] I have run autoformatter.sh


Connor-XY:
Hello @yezhengmao1, thanks for the clear writeup on the µP param-group collision.

I checked the diff: this changes optimizer param-group identity from the numeric tuple `(wd_mult, lr_mult, is_expert_parallel, is_decoupled_lr)` to a stable `param_group_name` derived from matched `ParamKey.canonical_name()` values, then all-gathers those matched names so PP/EP ranks agree on group identity. That directly addresses the case where embedding/hidden/output groups have the same multipliers but different LR schedules and would otherwise collide during checkpoint reload.

The key review points are distributed-optimizer checkpoint compatibility and whether older checkpoints/NeMo-style prefixed group keys need a migration or fallback path. I'm assigned as current MCore oncall to route the dist-optimizer/training review and make the remaining blocker explicit.

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -888,21 +888,7 @@ def load_state_dict(self, state_dict):
...
-                elif f"pre_{key}" in param_group:
-                    key = f"pre_{key}"
-                else:
-                    raise ValueError(
-                        f"Key {key} (or pre_{key}) not found in param_group {param_group}."
-                    )
-                needed_groups.append(param_group[key])
-            needed_groups = tuple(needed_groups)
-            return needed_groups
+            return tuple(param_group[key] for key in param_group_identifier_keys)
```
Connor-XY, line 891: This new key path will raise `KeyError` on optimizer checkpoints saved before this PR because their `param_groups` only carry the old identifier fields (`wd_mult`, `lr_mult`, `is_expert_parallel`, `is_decoupled_lr`) or the NeMo `pre_*` variants. Could we keep a migration/fallback path here, for example prefer `param_group_name` when present but fall back to the old tuple when loading legacy checkpoints? Otherwise resume from existing distributed-optimizer checkpoints regresses.

tests/unit_tests/test_optimizer.py
```diff
@@ -168,7 +170,14 @@ def test_get_param_groups_multiple_matches(mock_get_world_size):
     check_config_overrides_consistency(opt_config, config_overrides)
     param_groups2 = _get_param_groups([net], opt_config, config_overrides)
     assert len(param_groups) == 2
-    assert param_groups == param_groups2
+    # `param_group_name` differs by how the matching ParamKeys are spelled
```
Connor-XY, line 173: This test highlights a compatibility gap: two semantically equivalent `ParamKey` configurations can produce different `param_group_name` values, but `param_group_name` is now the checkpoint identity. Please add a load/reorder test that either makes the canonical name stable for equivalent configs or documents/handles that config spelling changes are incompatible.

Connor-XY (commented):
Thanks for the patch; I agree the old numeric tuple can collide for µP param groups. I started the oncall pass and will trigger CI on the current head. Before merge, I think we need to close the checkpoint-identity gaps below so existing distributed-optimizer checkpoints and equivalent config spellings do not regress.

maanug-nv (approved):

PR #5062, opened 2026-05-29
[mimo] Support bridge fan-out for variable modality tokens

liding-nv:
## Summary

Non-colocated bridge fan-out previously sent equal batch-dim slices to every peer. For VLM/AVLM data the per-sample modality-token counts vary, so the LLM received the wrong embeddings. Tag encoder outputs with per-sample split sizes and have the bridge honor them when fanning activations/gradients out.

  - `MimoModel._attach_modality_split_sizes` annotates flat encoder outputs with `_mimo_bridge_split_sizes` derived from  `special_token_ids`.
  - `MimoModel._empty_encoder_output` returns a `(0, hidden_size)` placeholder when a text-only sample reaches a non-colocated encoder rank; `_has_encoder_tokens` guards against silent loss of modality data.
  - `BridgeCommunicator._split_tensor_at_batch_dim` consumes `_mimo_bridge_split_sizes` (per-peer or per-sample) with sum/length checks.
  - `BridgeCommunicator._communicate_shapes` accepts a per-peer tensor list via the new `_as_per_peer_tensors` helper.
  - `schedules.backward_step_multimodule` unwraps single-element tensor lists so the per-peer paths feed back cleanly.

  ## Tests

  CPU-only unit tests added inside existing files:
  - `test_bridge_communicator.py::TestBridgeCommunicatorSplitMetadata` — `_split_tensor_at_batch_dim` sum-mismatch + `num_splits=1`short-circuit; `_as_per_peer_tensors` broadcast / pass-through / count-mismatch.
  - `test_mimo_model.py::TestMimoModelFanoutHelpers` — `_attach_modality_split_sizes` happy path + skip branches; `_has_encoder_tokens` presence/absence.

### DIFF
--- a/megatron/core/models/mimo/model/base.py
+++ b/megatron/core/models/mimo/model/base.py
@@ -423,7 +423,7 @@ class MimoModel(MegatronModule):
                     raise RuntimeError(
                         f"{encoder_name} inputs are missing, but matching special tokens exist"
                     )
-                output = self._empty_encoder_output(submodule, input_ids)
+                output = self._empty_encoder_output(encoder_name)
 
             if output is not None:
                 self._attach_modality_split_sizes(output, input_ids, encoder_name)
@@ -476,9 +476,7 @@ class MimoModel(MegatronModule):
             return False
         return bool((input_ids == self.special_token_ids[encoder_name]).any().item())
 
-    def _empty_encoder_output(
-        self, submodule: torch.nn.Module, input_ids: Optional[torch.Tensor]
-    ) -> torch.Tensor:
+    def _empty_encoder_output(self, encoder_name: str) -> torch.Tensor:
         """Return the bridge payload for text-only non-colocated batches."""
         language_config = self.mimo_config.language_model_spec.params['config']
         hidden_size = getattr(language_config, 'hidden_size', None)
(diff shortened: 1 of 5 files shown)


megatron/core/pipeline_parallel/schedules.py
```diff
@@ -550,13 +557,14 @@ def backward_step_multimodule(
 
     # Apply grad scaling if needed (for last stage only).
     for module_name in output_tensor.keys():
-        if output_tensor_grad[module_name] is None and config.grad_scale_func is not None:
+        output_tensor_grad_module = _unwrap_single_tensor_list(output_tensor_grad[module_name])
```
yashaswikarnati, line 560: May be I'm missing something.. would we need this change in schedules

liding-nv: Yes. We hit this in the Qwen3.5-27B MIMO validation run: language TP4/PP4/DP2, image encoder TP1/PP1/DP1, with `variable_seq_lengths=True`.

  In that path, `get_tensor_shapes()` returns `[()]`. The regular P2P communicator then returns a single-element grad list, so the multimodule dict can contain:

  {"language": [grad_tensor]}

  `backward_step_multimodule()` already unwraps this shape for `input_tensor`, but not for `output_tensor_grad`. Without this, we pass `[grad_tensor]` into backward instead of `grad_tensor`.

  So this is just making multimodule backward handle the same single-tensor-list convention as the existing single-module schedule.

megatron/core/models/mimo/model/base.py
```diff
@@ -414,16 +415,93 @@ def _forward_encoders(
...
+            device = ref_tensor.device
+        elif input_ids is not None:
+            device = input_ids.device
+        else:
+            device = torch.device("cuda", torch.cuda.current_device())
+        dtype = ref_tensor.dtype if ref_tensor is not None else self.config.params_dtype
+        if dtype is None:
+            dtype = torch.float32
+
+        return torch.empty((0, hidden_size), device=device, dtype=dtype, requires_grad=True)
```
yashaswikarnati: just a nit; can dtype also derived from vision_config and can the device be just the current device, instead of above if/else blocks

yashaswikarnati:
lgtm. lets revist and modify this slightly after the current code freeze/release

yashaswikarnati (approved):

yaoyu-33 (approved):

jaredcasper (approved):

yaoyu-33 (approved):

PR #5071, opened 2026-05-30
add mcore fix to the optimizer working on random weights issue

pengdurice:

# What does this PR do ?

Fixes a silent correctness bug where, with CPU optimizer offloading enabled,
loading a model checkpoint *without* the optimizer state (i.e. fine-tuning from
pretrained weights) leaves the `HybridDeviceOptimizer`'s internal parameter
copies at their stale, random-initialized values. The first `step()` then
computes the Adam update from — and copies back — those stale weights, silently
discarding the freshly loaded checkpoint. Training effectively restarts from
random init.

## Issue tracking

Linked issue: Related to NVIDIA-NeMo/RL#2371

This is the upstream (Megatron-Core) fix for the bug worked around downstream in
NVIDIA-NeMo/RL#2372 and mirrored in Megatron-Bridge#4082. Once this lands, both
NeMo-RL and Megatron-Bridge can drop their private-attribute workarounds.

## What is the bug

The bug manifests with a specific (but common for fine-tuning) combination:

> `optimizer_cpu_offload=True` + distributed optimizer + BF16 + initializing the
> model from a checkpoint (no optimizer state loaded).

When CPU offloading is on, the distributed optimizer's inner optimizer is a
`HybridDeviceOptimizer` (HDO). For each master parameter, HDO can keep up to two
additional copies that the sub-optimizers actually step against:

1. **FP32 shard master** (`shard_fp32_from_float16_groups`) — the per-DP-rank
   FP32 GPU shard used as the Adam master parameter. (Owned by the distributed
   optimizer.)
2. **CPU clones** (`HybridDeviceOptimizer.gpu_params_map_cpu_copy`) — pinned CPU
   copies the CPU sub-optimizer steps against; this is what makes offload
   actually offload.
3. **FP32 working copies** (`HybridDeviceOptimizer.param_to_fp32_param`) — extra
   FP32 copies kept for mixed-precision updates.

On a fine-tune-style load, the BF16 model parameters are refreshed and
`reload_model_params()` is called to re-seed the optimizer's master params. For
HDO, `DistributedOptimizer._copy_model_params_to_main_params()` took an early
return:

```python
if isinstance(self.optimizer, HybridDeviceOptimizer):
    self.optimizer.update_fp32_param_by_new_param()
    return
```

This early return is incorrect:

- It **skips `copy_group_params(...)`**, so the FP32 shard masters (level 1)
  are never refreshed from the loaded weights.
- `update_fp32_param_by_new_param()` only touches the FP32 working copies
  (level 3), which copy *from* the (still-stale) master params.
- The **CPU clones (level 2) are never refreshed at all.**

Net effect: all three copies remain at random init, and the first optimizer step
overwrites the loaded weights with garbage. Observed downstream as a loss of
~10 (log-prob of a uniform prediction over a ~100K vocab) on the very first
step of a Qwen2.5 SFT run with CPU offload enabled.

## The fix

`_copy_model_params_to_main_params()` now follows the normal path for HDO
instead of returning early:

1. Run `copy_group_params(...)` so the FP32 shard masters (level 1) are
   refreshed from the model params / `state_dict`, exactly as for a non-offload
   distributed optimizer.
2. Then call a new `HybridDeviceOptimizer.reload_model_params()` which re-seeds
   the CPU clones (level 2) and FP32 working copies (level 3) from the freshly
   updated master params.

The HDO master params (`gpu_params_map_cpu_copy` / `param_to_fp32_param` keys)
are the same tensor objects as `shard_fp32_groups` / `shard_fp32_from_float16_groups`
that `copy_group_params` updates in place, so the re-seed reads the loaded
values.

### Files changed

- `megatron/core/optimizer/cpu_offloading/hybrid_optimizer.py`: add
  `HybridDeviceOptimizer.reload_model_params()` that refreshes the CPU clones
  and FP32 working copies from the current parameter values.
- `megatron/core/optimizer/distrib_optimizer.py`: remove the HDO early return in
  `_copy_model_params_to_main_params()`; run the normal master-param copy, then
  call `self.optimizer.reload_model_params()` for HDO.
- `tests/unit_tests/test_optimizer_cpu_offloading.py`: add
  `test_reload_model_params_reseeds_internal_copies` (parametrized over
  `offload_fraction` and `param_update_in_fp32`) which populates the internal
  copies, simulates a checkpoint load by overwriting the model params in place,
  asserts the copies are stale before reload, and verifies `reload_model_params()`
  re-seeds them.

## Usage

No API change. The fix is transparent: any code path that already calls
`optimizer.reload_model_params()` after a checkpoint load (e.g. fine-tuning
without optimizer state) now correctly re-seeds all HDO parameter copies.

### Pre-checks


megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -3017,6 +3013,15 @@ def _copy_model_params_to_main_params(self, state_dict=None):
...
         if self.config.use_precision_aware_optimizer_no_fp8_or_ds_fp8:
+            # In this mode the params given to self.optimizer are views onto the model
+            # params, which the caller has already refreshed, so no shard copy is needed
+            # here. A HybridDeviceOptimizer, however, additionally keeps detached CPU
+            # clones and FP32 working copies of those params, which nothing else refreshes;
+            # re-seed them before returning. Note that `optimizer_cpu_offload` forces this
+            # branch to be taken whenever `use_precision_aware_optimizer` is set (see
+            # OptimizerConfig), so this is the path CPU offloading takes in practice.
+            if isinstance(self.optimizer, HybridDeviceOptimizer):
+                self.optimizer.reload_model_params()
```
janEbert: We now error out when using M-FSDP before allowing the optimizer to be a `HybridDeviceOptimizer`. Doesn't this break previously allowed behavior when M-FSDP is used?

janEbert: Similarly, we now only call `update_fp32_param_by_new_param` when `use_precision_aware_optimizer_no_fp8_or_ds_fp8`. Does this not cause breakage?

pengdurice: Thank you for the comments, fixed by honoring the original code structure;-)

pengdurice:
@janEbert , would you mind take another look? thank you!

pengdurice:
> @janEbert , would you mind take another look? thank you!

Hi @janEbert , thank you for kicking off the claude review, I added a comment without deleting the code since downstream repos may still use this method `update_fp32_param_by_new_param` to be on the safe side. Thank you!

PR #5074, opened 2026-05-30
ci(test): isolate ckpt-resume tensorboard per phase

### LINKED ISSUE #4903: 🐛 CI failure(nightlies and MR runs): ckpt-resume tests on dgx_gb200 fail with IndexError when reading 2nd-run tensorboard events
**Describe the bug**

Multiple `ckpt-resume` functional tests on `dgx_gb200` are failing in nightly CI with the same fingerprint. The first run (train-from-scratch) passes its golden-value comparison, but the post-resume `--is-second-run` analyzer crashes because no second tensorboard event file is present in the expected directory.

Tag @NVIDIA/mcore-oncall to get oncall's attention to this issue.

Failing test cases observed across two recent nightly runs (all `environment: dev`, `platforms: dgx_gb200`):

- `tests/functional_tests/test_cases/gpt/gpt3_mcore_te_tp2_pp1_resume_torch_dist_multi_dist_optimizer_instances`
- `tests/functional_tests/test_cases/moe/gpt3_mcore_te_tp2_pp1_resume_torch_dist_te_8experts2parallel_multi_dist_optimizer_instances`
- `tests/functional_tests/test_cases/gpt/gpt3_mcore_te_tp4_pp1_resume_torch_dist_dist_optimizer_overlap_grad_reduce_param_gather`
- `tests/functional_tests/test_cases/moe/gpt3_moe_mcore_te_tp4_ep2_etp2_pp2_resume_torch_dist_dist_optimizer`

Common pattern: every failing case matches `_resume_torch_dist_.*_dist_optimizer.*` and uses the distributed-optimizer + `--ckpt-format torch_dist` + (typically) `--dist-ckpt-optim-fully-reshardable` / `--num-distributed-optimizer-instances 2` resume path. Tests that do not exercise the dist-optimizer reshardable-resume code path on the same hardware are unaffected.

**Error**

`test_pretraining_regular_pipeline.py::test_regular_pipeline` (1st-run goldens) **PASSES** in all four jobs:

```
INFO  common.py:263 DETERMINISTIC test for metric lm loss: PASSED
INFO  common.py:263 APPROXIMATE  test for metric lm loss: PASSED
INFO  common.py:263 DETERMINISTIC test for metric num-zeros: PASSED
INFO  common.py:263 APPROXIMATE  test for metric num-zeros: PASSED
PASSED tests/functional_tests/python_test_utils/test_pretraining_regular_pipeline.py::test_regular_pipeline
```

The job then fails when `run_ci_test.sh` re-invokes the analyzer with `--is-second-run` against the same `$TENSORBOARD_PATH`:

```
+ uv run --no-sync python /opt/megatron-lm/tests/functional_tests/python_test_utils/get_test_results_from_tensorboard_logs.py \
    --logs-dir /jet/assets/basic/<test-case>/1/tensorboard \
    --train-iters 100 \
    --output-path /jet/assets/basic/<test-case>/golden_values_dev_dgx_gb200_2nd.json \
    --is-second-run --is-normal-test --step-size 1

  File "/opt/megatron-lm/tests/functional_tests/python_test_utils/get_test_results_from_tensorboard_logs.py", line 82, in <module>
    collect_train_test_metrics()
  File "/opt/megatron-lm/tests/functional_tests/python_test_utils/get_test_results_from_tensorboard_logs.py", line 43, in collect_train_test_metrics
    summaries = common.read_tb_logs_as_list(
  File "/opt/megatron-lm/tests/functional_tests/python_test_utils/common.py", line 124, in read_tb_logs_as_list
    event_file = files[index]
                 ~~~~~^^^^^^^
IndexError: list index out of range
INFO:__main__:Pipeline terminated with status FAILED
```

`get_test_results_from_tensorboard_logs.py` passes `index=1` when `--is-second-run` is set (line 45), expecting two `events.out.tfevents.*` files in the supplied `--logs-dir`. The IndexError means only one event file is present.

**Artifact evidence**

Inspecting the failing job artifacts confirms the resume run produced an event file in a *different* sub-directory than the first run, instead of appending to the same one:

```
results/iteration=0/restart=0/assets/basic/<test-case>/1/tensorboard/events.out.tfevents.<ts1>...   (~181 KB — first run)
results/iteration=0/restart=0/assets/basic/<test-case>/2/tensorboard/events.out.tfevents.<ts2>...   (~181 KB — resume run)
```

`run_ci_test.sh` exports `TENSORBOARD_PATH=$DIR/$i/$FILE` once per outer-loop iteration `$i` and re-uses it for both `RUN_NUMBER=1` and `RUN_NUMBER=2`, so both runs are expected to write into the same `<dir>/<i>/tensorboard/`. The artifact layout shows the resume run instead wrote into `<dir>/<i+1>/tensorboard/`, leaving exactly one event file in the directory the analyzer reads — hence `files[index=1]` fails. Both failing pipelines exhibit the same artifact layout.

The training itself completes (W&B logs from rank 7's `stderr.log` show iteration counts past 100 in the resume run), so the regression is in how `--tensorboard-dir` is honored on the resume invocation, not in dist-optimizer correctness.

Inside the dev container, run any of the failing test cases as a `ckpt-resume` job on a 2-node × 4-GPU GB200 configuration. Minimum repro args (from the moe job's training cmdline):

```bash
ft_launcher --nproc_per_node 4 --nnodes 2 --master_addr <addr> --master_port <port> --node_rank 0 \
  --max-restarts=3 pretrain_gpt.py \
  --num-layers 12 --hidden-size 512 --num-attention-heads 8 \
  --tensorboard-dir <SHARED_DIR>/1/tensorboard \
  --micro-batch-size 4 --global-batch-size 32 --seq-length 1024 --max-position-embeddings 1024 \
  --train-iters 100 --exit-interval 100 \
  --save <SHARED_DIR>/checkpoints --load /tmp/checkpoints/ \
  --transformer-impl transformer_engine \
  --tensor-model-parallel-size 2 --pipeline-model-parallel-size 1 \
  --use-distributed-optimizer --num-distributed-optimizer-instances 2 \
  --overlap-grad-reduce --overlap-param-gather \
  --deterministic-mode --no-gradient-accumulation-fusion --attention-softmax-in-fp32 \
  --use-checkpoint-opt_param-scheduler --use-mcore-models \
  --ckpt-format torch_dist --dist-ckpt-optim-fully-reshardable --dist-ckpt-strictness log_all \
  --bf16 --async-save --use-persistent-ckpt-worker
```

Then re-run the same command with `--load <SHARED_DIR>/checkpoints` (i.e. `RUN_NUMBER=2` from `run_ci_test.sh`) and check whether a second `events.out.tfevents.*` file lands in `<SHARED_DIR>/1/tensorboard/` (expected) or in a sibling `<SHARED_DIR>/2/tensorboard/` (observed).

**Additional context**

- All four failing jobs run GB200, 2 nodes × 4 GPUs.
- The first run passes its `test_regular_pipeline` check on the recorded goldens, so the dist-optimizer + torch_dist save side is intact; the regression is on the resume + tensorboard-dir reuse path.
- Same failure signature reproduces across two independent recent nightly runs, indicating a deterministic regression rather than infrastructure flakiness.
- Suspect commits to investigate: recent changes to `tests/functional_tests/shell_test_utils/run_ci_test.sh` and to the megatron training-init path that constructs the tensorboard SummaryWriter on resume.

Triaged automatically via `/create-issue`.

ko3n1g:

## What

`ckpt-resume` and `frozen-resume` functional tests previously wrote tensorboard events from **both** training phases into the same `$TENSORBOARD_PATH` directory and relied on `files[1]` (mtime-sorted glob `index=1`) to identify the resume run's data. That assumption is racy:

- if only one events file lands in the dir → `IndexError` (the failure mode in #4903),
- if the indexed file happens to have no scalar events flushed yet, or is some unrelated extra file → `read_tb_logs_as_list` returns `{}`, the `_2nd.json` file is written as `{}`, and the resume pytest fails in fixture setup with a pydantic `ValidationError: input_value=PydanticUndefined` out of `read_golden_values_from_json` at `tests/functional_tests/python_test_utils/common.py:168` — the latest h100 incarnation observed on the `gpt3_mcore_te_tp2_pp1_resume_torch_dist_te_8experts2parallel_multi_dist_optimizer_instances` test case (same test enumerated in #4903).

This PR removes the glob-ordering dependency by giving each phase its own directory.

## How

`run_ci_test.sh`

- Snapshot the per-repeat tensorboard base into `_REPEAT_TENSORBOARD_PATH`.
- For `ckpt-resume` / `frozen-resume`, before each training phase set:
  ```sh
  export TENSORBOARD_PATH="$_REPEAT_TENSORBOARD_PATH/run_${RUN_NUMBER}"
  mkdir -p "$TENSORBOARD_PATH"
  ```
  Training picks up the new path via the existing `--tensorboard-dir: ${TENSORBOARD_PATH}` in the recipe YAML.
- First-run analyzer now points at `${_REPEAT_TENSORBOARD_PATH}/run_1` for resume tests (unchanged for everything else).
- Second-run analyzer points at `${_REPEAT_TENSORBOARD_PATH}/run_2` and no longer passes `--is-second-run`.

`get_test_results_from_tensorboard_logs.py`

- The `--is-second-run` flag and the `index=1` branch it controlled are removed. The single caller now passes an explicit per-phase `--logs-dir`, so the script always reads `files[0]`.

## Layout, before / after

```text
# before (both phases collide; resume analyzer guesses via mtime sort)
$TENSORBOARD_PATH/
  events.out.tfevents.<ts1>.<host>.<pid1>    ← run 1
  events.out.tfevents.<ts2>.<host>.<pid2>    ← run 2 (analyzer: files[1])

# after (per-phase isolation; analyzer reads explicit subdir)
$TENSORBOARD_PATH/
  run_1/events.out.tfevents.<...>
  run_2/events.out.tfevents.<...>
```

## Scope

- Behavior unchanged for `frozen-start`, `release`, `checkpoint-consistency`, inference, and RL modes — they don't write to per-run subdirs.
- No change to golden values; same scalars, same metric filter list.

## Related

- Same failure family as the closed #4903 (gb200 `IndexError` variant). The multi-node sync fix in #4924 didn't address single-node h100 because the underlying cause is the analyzer's index-into-sorted-glob, not multi-node phase ordering.
</details>

### DIFF
--- a/tests/functional_tests/python_test_utils/get_test_results_from_tensorboard_logs.py
+++ b/tests/functional_tests/python_test_utils/get_test_results_from_tensorboard_logs.py
@@ -1,3 +1,4 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 import os
 
 os.environ["OPENBLAS_NUM_THREADS"] = "1"
--- a/tests/functional_tests/python_test_utils/get_test_results_from_tensorboard_logs.py
+++ b/tests/functional_tests/python_test_utils/get_test_results_from_tensorboard_logs.py
@@ -22,27 +22,13 @@ logger = logging.getLogger(__name__)
     help="Use first or all tensorboard logs",
     default=False,
 )
-@click.option(
-    "--is-second-run/--is-not-second-run",
-    type=bool,
-    help="Use second run of tensorboard logs",
-    default=False,
-)
 @click.option("--step-size", required=False, default=5, type=int, help="Step size of sampling")
 def collect_train_test_metrics(
-    logs_dir: str,
-    train_iters: str,
-    output_path: str,
-    is_convergence_test: bool,
-    is_second_run: bool,
-    step_size: int,
+    logs_dir: str, train_iters: str, output_path: str, is_convergence_test: bool, step_size: int
 ):
-    if is_convergence_test and is_second_run:
-        raise ValueError("Convergence test cannot be run on second run of tensorboard logs")
-
     summaries = common.read_tb_logs_as_list(
         logs_dir,
-        index=(-1 if is_convergence_test else (1 if is_second_run else 0)),
+        index=(-1 if is_convergence_test else 0),
         train_iters=train_iters,
         start_idx=1,
         step_size=step_size,
--- a/tests/functional_tests/shell_test_utils/run_ci_test.sh
+++ b/tests/functional_tests/shell_test_utils/run_ci_test.sh
@@ -226,6 +226,11 @@ for i in $(seq 1 $N_REPEAT); do
     FILE=$(basename "$_TENSORBOARD_PATH")
     export TENSORBOARD_PATH=$DIR/$i/$FILE
     mkdir -p $(dirname $TENSORBOARD_PATH)
+    # Per-repeat base; for ckpt-resume / frozen-resume tests each training
+    # phase writes to its own ${_REPEAT_TENSORBOARD_PATH}/run_N subdir so the
+    # analyzer can read by explicit path rather than guessing phase from
+    # mtime-sorted glob order.
+    _REPEAT_TENSORBOARD_PATH=$TENSORBOARD_PATH
     export REPEAT=$i
 
     wait_for_all_nodes "repeat_${REPEAT}_start"
@@ -294,6 +299,10 @@ for i in $(seq 1 $N_REPEAT); do
         done
     else
         # The standard single-run test that otherwise runs
+        if [[ "$TEST_TYPE" == "ckpt-resume" || "$TEST_TYPE" == "frozen-resume" ]]; then
+            export TENSORBOARD_PATH="$_REPEAT_TENSORBOARD_PATH/run_${RUN_NUMBER}"
+            mkdir -p "$TENSORBOARD_PATH"
+        fi
         run_training_phase "repeat_${REPEAT}_run_${RUN_NUMBER}"
     fi
 
@@ -316,6 +325,8 @@ for i in $(seq 1 $N_REPEAT); do
         echo $((TRAIN_ITERS / 2)) >$CHECKPOINT_LOAD_PATH/latest_checkpointed_iteration.txt
 
         export RUN_NUMBER=2
+        export TENSORBOARD_PATH="$_REPEAT_TENSORBOARD_PATH/run_${RUN_NUMBER}"
+        mkdir -p "$TENSORBOARD_PATH"
         run_training_phase "repeat_${REPEAT}_run_${RUN_NUMBER}"
     fi
 
@@ -326,6 +337,8 @@ for i in $(seq 1 $N_REPEAT); do
         export CHECKPOINT_SAVE_PATH=/tmp/checkpoints/
 
         export RUN_NUMBER=2
+        export TENSORBOARD_PATH="$_REPEAT_TENSORBOARD_PATH/run_${RUN_NUMBER}"
+        mkdir -p "$TENSORBOARD_PATH"
         run_training_phase "repeat_${REPEAT}_run_${RUN_NUMBER}"
 
         export CHECKPOINT_SAVE_PATH=$_CHECKPOINT_SAVE_PATH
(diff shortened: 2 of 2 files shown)


tests/functional_tests/python_test_utils/get_test_results_from_tensorboard_logs.py
```diff
@@ -1,3 +1,4 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
thomasdhc, line 1: do we need the full copyright?

ko3n1g: No, we only include the header not the full file

thomasdhc (approved):

PR #5077, opened 2026-05-31
Fix bugs in re-sharding `DistributedOptimizer` state in `torch_dist` checkpoint

deepakn94:
## Summary

Fixes optimizer state corruption when saving a `torch_dist` checkpoint with `--dist-ckpt-optim-fully-reshardable` and a grad buffer split across more than one bucket.

`get_parameter_state_dp_zero()` packs the coalesced world tensors bucket by bucket with each bucket's end padding stripped, but `sharded_param_state_fully_reshardable()` slices into them using raw `param_index_map` offsets, which still include that padding. Every parameter past the first bucket is read from the wrong offset; the overshoot ran off the end of the world tensor and was silently zero-padded with a warning, so the saved state for those params was partly zeros.

The fix computes the cumulative bucket-end padding stripped before each bucket and subtracts it before indexing. The world tensors stay sized to `buffer_numel_unpadded`, which is exactly what the rebased offsets need, and an explicit bound check on the adjusted end replaces the zero-padding workaround so a future indexing error fails loudly instead of writing zeros.

The bug is invisible with a single bucket, where the adjustment is identically zero, and that is the only configuration existing tests covered.

## Tests

`tests/unit_tests/dist_checkpointing/test_optimizer.py`:

- `test_fully_reshardable_multi_bucket_save_load` - save/load round trip over 2 and 3 buckets, plus one case with `pad_buckets_for_high_nccl_busbw`. The load path reads `state_dict[param_idx]` by key rather than re-slicing `world_tensors`, so a wrong save offset does not cancel out on load.
- `test_coalesced_world_tensors_are_compact` - the world tensors stay at `numel_unpadded` and every rebased param range ends exactly there.

Both assert up front that the buffer really split into N buckets and that the non-final buckets carry nonzero padding, so neither can silently degenerate into the single-bucket case it exists to cover.

Supporting changes: a `MultiBucketModel` fixture, and `ddp_num_buckets` / `ddp_pad_buckets_for_high_nccl_busbw` knobs on `setup_model_and_optimizer`. Both knobs are needed because `resolve_ddp_bucket_size()` forces a single bucket unless `overlap_grad_reduce` is on.

## Test plan

- [ ] `tests/unit_tests/dist_checkpointing/test_optimizer.py -k "multi_bucket or compact"` on 8 GPUs. **Not yet run.**
- [ ] CI: `Run functional tests` label (touches the optimizer code path).

### DIFF
--- a/megatron/core/optimizer/distrib_optimizer.py
+++ b/megatron/core/optimizer/distrib_optimizer.py
@@ -1709,10 +1709,21 @@ class DistributedOptimizer(MixedPrecisionOptimizer):
                 # Note: for NVFP4, param_index_map uses unpacked (full numel)
                 # offsets, which is correct here since optimizer states
                 # (fp32_param, exp_avg, exp_avg_sq) are in unpacked space.
+
+                # Compute cumulative bucket-end padding stripped before each bucket.
+                # world_tensors has bucket-end padding stripped, but param_index_map
+                # indices include bucket-end padding. We need to adjust indices.
+                cumulative_padding_stripped = [0]  # For bucket 0, no prior padding stripped
+                for bucket in buffer.buckets[:-1]:  # All but last bucket
+                    bucket_padding = bucket.grad_data.numel() - bucket.numel_unpadded
+                    cumulative_padding_stripped.append(
+                        cumulative_padding_stripped[-1] + bucket_padding
+                    )
+
                 for model_param, (
                     param_world_start,
                     param_world_end,
-                    _,
+                    bucket_id,
                 ) in buffer.param_index_map.items():
                     try:
                         sharded_metadata = param_to_sharded_metadata[model_param]
@@ -1728,6 +1739,13 @@ class DistributedOptimizer(MixedPrecisionOptimizer):
                     # Note: replica_id is exactly the same as in the model param
                     replica_id = sharded_metadata.replica_id
 
+                    # Adjust indices to account for stripped bucket-end padding.
+                    # param_world_start/end are indices in the buffer (with padding),
+                    # but world_tensors has the bucket-end padding stripped.
+                    padding_adjustment = cumulative_padding_stripped[bucket_id]
+                    adjusted_start = param_world_start - padding_adjustment
+                    adjusted_end = param_world_end - padding_adjustment
+
                     tensors = {}
                     for state_key in world_tensor_keys:
                         if state_key == 'step' or state_key == 'numel' or state_key == 'numel_unpadded':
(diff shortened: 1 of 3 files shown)


kvareddy (approved):

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -1255,16 +1255,16 @@ def get_parameter_state_dp_zero(
...
+                buffer_numel = self.buffers[gbuf_idx].numel
                 buffer_numel_unpadded = self.buffers[gbuf_idx].numel_unpadded
                 # Create coalesced tensors for all state related to parameters in this buffer.
                 world_tensors = {}
                 if data_parallel_rank == 0 or return_on_all_ranks:
                     world_tensors = {
-                        key: torch.zeros(
-                            (buffer_numel_unpadded,), dtype=torch.float32, device="cpu"
-                        )
+                        key: torch.zeros((buffer_numel,), dtype=torch.float32, device="cpu")
```
lmcafee-nvidia: Codex's assessment here, might be worth seeing what Claude thinks:

I think this should stay sized to the compact `buffer_numel_unpadded` layout unless we also change the copy/load paths to use a padded layout consistently. Lines 1343-1349 still write each bucket at `offset_in_world_tensors += gbuf_world_numel_unpadded`, and `load_parameter_state_from_dp_zero()` also reads only the compact unpadded ranges. Allocating `buffer_numel` adds unused zero tails to `dp_zero_gather_scatter` checkpoints and to fully-reshardable load placeholders on every rank; with `pad_buckets_for_high_nccl_busbw`, that can be large. It also masks future bad range math because an adjusted slice can run into the zero tail and still satisfy the length assert. Can we keep the world tensors compact and assert `adjusted_end <= world_tensors[state_key].numel()` after the bucket-padding adjustment?

deepakn94:
/ok to test [0055ac0](https://github.com/NVIDIA/Megatron-LM/pull/5077/commits/0055ac06845b45701879a0454a14fde84ac3643a)

deepakn94:
/ok to test [186a190](https://github.com/NVIDIA/Megatron-LM/pull/5077/commits/186a1903d5873e38b0538b1a1a33bab30b9248e6)

deepakn94:
/ok to test [04bd0c1](https://github.com/NVIDIA/Megatron-LM/pull/5077/commits/04bd0c15d55165e8a3b6b2c26f13c47652262b19)

Phlip79 (approved):

zhongbozhu (approved):

PR #5078, opened 2026-06-01
Add FSDPTensorParallelMuon + 8-rank step perf + numerics tests

wujingyue:
## Summary

- Adds the baseline `FSDPTensorParallelMuon` class to `main` so users have a Muon optimizer that works with Megatron-FSDP ZeRO-1/2/3 out of the box. Implementation derived from prototype PR [#4486](https://github.com/NVIDIA/Megatron-LM/pull/4486) (introducer commit, no perf knobs).
- Adds a pytest-driven test that replays just `optimizer.step()` against fake DTensors reconstructed from a captured per-rank dump of a real 2-node / 8-rank training run. Two tests share the setup: a `pytest-benchmark` microbenchmark, and a numerics correctness check against a fully-replicated reference `emerging_optimizers.Muon`. **The tests and the fixtures are new in this PR.**

> **Note on diff size:** ~95% of the ~37k added lines are the 8 captured-input JSON fixture files (8 × ~4585 lines) — these describe per-rank parameter sharding from a real training run and are data, not hand-written code. The hand-written additions are ~720 lines (optimizer + test).

## What's in this PR

- `megatron/core/optimizer/emerging_optimizers.py` (+175): adds `FSDPTensorParallelMuon(TensorParallelMuon)` and its private helpers (`_needs_boundary_gather`, `_get_boundary_gather_param_indices`, `_copy_dtensor_chunk_metadata`, `_dtensor_from_local_like`, `_reshard_full_update_like`, `_local_muon_update`). `DTensor` is imported at module top (Megatron-LM requires torch>=2.6, where DTensor is stable). The optimizer's boundary-vote `all_gather_object` runs over WORLD — no `dp_group` parameter on the public API.
- `megatron/core/distributed/fsdp/src/megatron_fsdp/uneven_dtensor.py` (+5/-2): minor adjustments needed by the new helpers. Ported from [#4486](https://github.com/NVIDIA/Megatron-LM/pull/4486).
- `tests/unit_tests/distributed/megatron_fsdp/test_muon.py` (+539): **new** — `test_muon_step` (perf) and `test_muon_step_numerics` (correctness).
- `tests/unit_tests/distributed/megatron_fsdp/muon_inputs_fsdp4_tp2/rank{0..7}.json` (+36680): **new** — 8 fixture files emitted by `dump_optimizer_parameters` ([#5307](https://github.com/NVIDIA/Megatron-LM/pull/5307)). Each entry records `global_shape`, `local_shape`, `dtype`, `is_qkv`, `mesh_shape`, `mesh_dim_names`, `mesh_ranks` (flat, row-major), and `placements` per param.
- `pyproject.toml` (+1): adds `pytest-benchmark` as a dev dependency.

The `DistributedSetup` fixture used by both tests lives in the shared `tests/unit_tests/distributed/megatron_fsdp/conftest.py` already on `main`.

## How to run

```
cd Megatron-LM
torchrun --nproc_per_node=8 --standalone -m pytest \
    tests/unit_tests/distributed/megatron_fsdp/test_muon.py -v -s
```

Both tests load the per-rank JSON dump and construct each parameter as a DTensor on the recorded sub-mesh. Each param records its **own** mesh — including the (2,)-rank EP sub-meshes that vary across ranks (e.g. rank 0 sees `[0, 4]`, rank 1 sees `[1, 5]`). Per-rank-divergent `new_group` ordering would hang NCCL, so `_build_mesh_cache` `all_gather_object`s the world-wide union of `(mesh_shape, mesh_dim_names, mesh_ranks)` tuples and constructs every unique sub-mesh in lockstep. The world process group is bootstrapped via `init_device_mesh`'s documented PG-init side effect — no explicit `dist.init_process_group` in the test.

### `test_muon_step` (perf)

Runs `pytest-benchmark`'s `pedantic` mode (3 warmup + 5 timed rounds) over `optimizer.step()` and reports min/median/max/mean/stddev wall time.

### `test_muon_step_numerics` (correctness)

The reference path lives in a small `ReplicatedMuon` class with a single public `step(local_shards, param_specs, dp_group_of)` method that iterates per-param:

1. Build the FSDP DTensor params and grads.
2. Clone each pre-step local shard into `reference_params` (plain tensors), with the grad attached via `.grad`.
3. Run FSDP+TP Muon on the DTensors (mutates in place).
4. `ReplicatedMuon.step(...)` runs the reference for each param: for boundary params (any rank in the param's own DP group has a non-empty local with shape ≠ global), all-gather across that param's `dp_cp` sub-group (mirroring FSDP Phase 2 — TP is left alone since `tp_mode="blockwise"` makes Newton-Schulz purely local), run Muon on the full view (manual Q/K/V split for QKV-fused params), and slice the result back to this rank's shard. For non-boundary params, run Muon directly on the local shard.
5. Assert each FSDP post-step shard is close to its reference local with `torch.testing.assert_close(atol=1e-2, rtol=1e-2)`. NaN/Inf naturally fail this comparison.

Each param is processed in isolation so peak HBM per rank stays bounded by one param's DP-gathered footprint.

**Scope:** the test consumes the 119 ndim=2 wd-group params. The wd group also contains 5 3D Mamba state-space matrices that `FSDPTensorParallelMuon` reshapes internally; plain `emerging_optimizers.Muon` (the reference) rejects them with "Only 2D parameters are supported", so they are filtered. 3D coverage is a follow-up.

## Skip behavior

- Not under torchrun: skips with `Not running under torchrun`.
- WORLD_SIZE != 8: skips with `Dump captured at WORLD_SIZE=8`.
- emerging-optimizers package not importable: skips.
- LTS pytorch container: skips.
- CUDA not available: skips.

## Test plan

- [x] `torchrun --nproc_per_node=8 ... -m pytest test_muon.py -v -s` passes on gb300 (2 × 4 GPUs) — both tests green.
- [x] Microbench: ~836 ms mean per step (5 rounds, ±5 ms StdDev).
- [x] Skip path verified for non-torchrun and WORLD_SIZE=4.
- [x] `black --line-length 100 --skip-string-normalization` clean on changed files.


megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -276,6 +283,179 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+                self._apply_weight_decay_inplace(p_local, local_grad, lr, group["weight_decay"])
+                mom_local.lerp_(local_grad, 1 - group["momentum"])
+                if self.nesterov:
+                    pre_ns_grad = local_grad.lerp(mom_local, group["momentum"])
+                else:
+                    pre_ns_grad = mom_local
+
+                all_updates.append((p, pre_ns_grad, needs_gather, lr, group_kwargs))
+
+        # Phase 2: AG all boundary params — all collectives here, no NS interleaved.
```
Autumn1998, line 364: We currently perform all the momentum updates first and then launch all the gather operations. However, I think we could restructure it like this:

```
for param in params:
    momentum_update(param)
    gather(param) if needed
```

This would allow some degree of overlap between the gather operations and the momentum updates.

As mentioned earlier, we should place the communication-independent ones at both ends so they can overlap with the first gather and the final scatter. We can sort the parameters accordingly; this ordering is not only useful for overlapping NS with gather later, but also applies here for overlapping gather with momentum.

wujingyue: Sure but that's out of the scope of this PR. This PR adds a baseline just so the unit test and the microbench have a target. The optimizer itself will be improved incrementally after this PR.

PR #5080, opened 2026-06-01
[Dev] Support isolated MTP loss

Victarry:
main PR: https://github.com/NVIDIA/Megatron-LM/pull/5098/changes

## Summary
- Add `mtp_isolated_loss` to detach MTP auxiliary loss from the main decoder, shared embeddings, and output layer weights.

For online RL/MTP training, we will rely on the scheme from https://github.com/NVIDIA/Megatron-LM/pull/3457 instead:

keep labels=None on the main model forward so the main LM head returns logits for the external RL loss;
derive MTP auxiliary labels from input_ids inside the MTP loss path, matching the SFT label convention;
use mtp_isolated_loss=True when the auxiliary MTP loss should only update MTP module parameters and stay detached from the main decoder / shared embeddings / output layer.

## RL / Online MTP Background
Online RL training frameworks such as veRL usually run the policy model on sampled responses, read the logits, and compute the policy loss outside the model. The labels for those sampled tokens are still useful for auxiliary objectives such as MTP, but they should not automatically turn the main model output into a supervised LM loss.

Before this change, MCore uses one `labels` argument for both purposes:
- `labels` are consumed by MTP auxiliary loss.
- The same `labels` also make the main GPT/Hybrid forward return LM loss instead of logits.

That coupling is inconvenient for RL. In online RL + MTP, the desired contract is:
- sampled tokens are passed as `labels` so MTP can compute its auxiliary CE loss;
- the main forward still returns logits so the RL framework can compute PPO/GRPO/etc. externally;
- the MTP auxiliary loss can be isolated so it only trains MTP-specific parameters, without changing the base policy through decoder hidden states, shared embeddings, or output layer weights.

This PR keeps the regular pretraining/SFT behavior unchanged by default. For RL-style callers, `return_logits=True` changes only the main LM output contract: labels are still consumed by MTP auxiliary loss, but the forward returns logits instead of main LM loss. `mtp_isolated_loss=True` separately controls the gradient boundary for the auxiliary objective.

Example RL-style use:

```python
logits = model(
    input_ids=input_ids,
    position_ids=position_ids,
    labels=None,
    loss_mask=response_loss_mask,
)
```

## Test plan
- `python -m py_compile megatron/core/models/hybrid/hybrid_model.py megatron/core/models/gpt/gpt_model.py megatron/core/transformer/multi_token_prediction.py megatron/core/transformer/transformer_config.py tests/unit_tests/transformer/test_multi_token_prediction.py megatron/core/models/common/model_chunk_schedule_plan.py megatron/core/models/gpt/fine_grained_callables.py`
- `CHECK_ONLY=true BASE_REF=dev uv run --project /Users/denliu/Projects/repos/agentic-mcore-dev --extra mcore-lint -- bash tools/autoformat.sh`
- `git diff --name-only --diff-filter=d --merge-base autoformatter-remote/dev megatron/core tests/ | grep '\.py$' | xargs uv run --project /Users/denliu/Projects/repos/agentic-mcore-dev --extra mcore-lint -- python tools/check_copyright.py`
- CW manual `salloc` validation on allocation `12394199` / node `pool0-01113`: `bash tests/unit_tests/run_ci_test.sh --tag latest --environment dev --bucket tests/unit_tests/transformer/test_multi_token_prediction.py --unit-test-repeat 1 --log-dir /results/logs/manual-mtp-260601_035257` (`39 passed`, experimental subset `39 deselected`).

### DIFF
--- a/megatron/core/models/common/model_chunk_schedule_plan.py
+++ b/megatron/core/models/common/model_chunk_schedule_plan.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
 from contextlib import nullcontext
 from typing import Optional
@@ -325,6 +325,7 @@ class TransformerModelChunkSchedulePlan(AbstractSchedulePlan):
         runtime_gather_output: Optional[bool] = None,
         loss_mask: Optional[Tensor] = None,
         padding_mask=None,
+        return_logits: bool = False,
     ):
         """Initialize the schedule plan of all Transformer layers' sub-modules.
 
@@ -342,6 +343,8 @@ class TransformerModelChunkSchedulePlan(AbstractSchedulePlan):
             extra_block_kwargs: Additional keyword arguments for blocks.
             runtime_gather_output: Whether to gather output at runtime.
             loss_mask (torch.Tensor): Used to mask out some portions of the loss
+            return_logits (bool): Return logits instead of main LM loss when labels
+                are provided. MTP auxiliary loss still consumes labels.
 
         Returns:
             The model chunk schedule plan.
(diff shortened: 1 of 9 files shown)


Victarry:
Updated this PR to remove the explicit `return_logits` API path.

For online RL/MTP training, we will use the scheme from #3457 instead:
- keep `labels=None` on the main model forward so the main LM head returns logits for the external RL loss;
- derive MTP auxiliary labels from `input_ids` inside the MTP loss path, matching the SFT label convention;
- use `mtp_isolated_loss=True` when the auxiliary MTP loss should only update MTP module parameters and stay detached from the main decoder / shared embeddings / output layer.

I also updated the `TransformerConfig.mtp_isolated_loss` comment to describe this online MTP usage model.

ISEEKYAN (approved):

yaox12 (approved):

PR #5087, opened 2026-06-01
[dev] [DeepSeek-v4] Context Parallel support

kunlunl:

# What does this PR do ?

Add CP support for DS v4

Depends on https://github.com/NVIDIA/Megatron-LM/pull/5011 (Merged)

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

### DIFF
--- a/docker/Dockerfile.ci.dev
+++ b/docker/Dockerfile.ci.dev
@@ -40,8 +40,6 @@ COPY README.md pyproject.toml uv.lock /workspace/
 COPY megatron/core/__init__.py /workspace/megatron/core/
 COPY megatron/core/package_info.py /workspace/megatron/core/
 ARG IMAGE_TYPE=dev
-ARG CUDNN_FRONTEND_COMMIT=0a14b7181d129d30e7bad34b8c3ed0a0c995e23d
-ARG FLASH_MLA_COMMIT=b7643bd54521f563b839b98289b5cd048c062ba2
 RUN --mount=type=cache,target=/root/.cache/uv \
     bash -ex <<"EOF"
     export NVTE_CUDA_ARCHS="80;90;100"
@@ -63,12 +61,7 @@ RUN --mount=type=cache,target=/root/.cache/uv \
         --no-install-package nvidia-cusolver-cu12 \
         --no-install-package nvidia-cusparse-cu12 \
         --no-install-package nvidia-cusparselt-cu12 \
-        --no-install-package nvidia-nccl-cu12 \
-        --no-install-package nvidia-cudnn-frontend
-    uv pip install --no-build-isolation \
-        "flash-mla @ git+https://github.com/deepseek-ai/FlashMLA.git@${FLASH_MLA_COMMIT}" \
-        "nvidia-cudnn-frontend[cutedsl] @ git+https://github.com/NVIDIA/cudnn-frontend.git@${CUDNN_FRONTEND_COMMIT}" \
-        "torch-c-dlpack-ext==0.1.5"
+        --no-install-package nvidia-nccl-cu12
 EOF
 
 # Install DeepEP
@@ -90,11 +83,6 @@ RUN bash -ex <<"EOF"
 EOF
 
 COPY assets/ /opt/data/
-# Reinstall CuTe DSL cleanly after upgrading the version provided by the base image.
-RUN uv pip install --reinstall --no-deps \
-    nvidia-cutlass-dsl==4.5.0 \
-    nvidia-cutlass-dsl-libs-base==4.5.0 \
-    nvidia-cutlass-dsl-libs-cu13==4.5.0
 ENV UV_PYTHON=$UV_PROJECT_ENVIRONMENT/bin/python
 
 ##### For NVIDIANS only #####
--- a/docker/Dockerfile.ci.dev
+++ b/docker/Dockerfile.ci.dev
@@ -31,7 +31,7 @@ RUN bash -ex <<"EOF"
     curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | sh
 EOF
 
-RUN ln -sfn /usr/local/cuda/targets/x86_64-linux/include/cccl/cuda \
+RUN ln -sf /usr/local/cuda/include/cccl/cuda \
     /usr/local/include/cuda
 RUN find /usr/local/cuda -name "utility" 2>/dev/null | head -5 && \
     ls /usr/local/cuda/targets/x86_64-linux/include/ | head -20
--- a/docker/Dockerfile.ci.dev
+++ b/docker/Dockerfile.ci.dev
@@ -31,7 +31,7 @@ RUN bash -ex <<"EOF"
     curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | sh
 EOF
 
-RUN ln -sf /usr/local/cuda/targets/x86_64-linux/include/cuda \
+RUN ln -sfn /usr/local/cuda/targets/x86_64-linux/include/cccl/cuda \
     /usr/local/include/cuda
 RUN find /usr/local/cuda -name "utility" 2>/dev/null | head -5 && \
     ls /usr/local/cuda/targets/x86_64-linux/include/ | head -20
@@ -40,6 +40,8 @@ COPY README.md pyproject.toml uv.lock /workspace/
 COPY megatron/core/__init__.py /workspace/megatron/core/
 COPY megatron/core/package_info.py /workspace/megatron/core/
 ARG IMAGE_TYPE=dev
+ARG CUDNN_FRONTEND_COMMIT=0a14b7181d129d30e7bad34b8c3ed0a0c995e23d
+ARG FLASH_MLA_COMMIT=b7643bd54521f563b839b98289b5cd048c062ba2
 RUN --mount=type=cache,target=/root/.cache/uv \
     bash -ex <<"EOF"
     export NVTE_CUDA_ARCHS="80;90;100"
@@ -61,7 +63,12 @@ RUN --mount=type=cache,target=/root/.cache/uv \
         --no-install-package nvidia-cusolver-cu12 \
         --no-install-package nvidia-cusparse-cu12 \
         --no-install-package nvidia-cusparselt-cu12 \
-        --no-install-package nvidia-nccl-cu12
+        --no-install-package nvidia-nccl-cu12 \
+        --no-install-package nvidia-cudnn-frontend
+    uv pip install --no-build-isolation \
+        "flash-mla @ git+https://github.com/deepseek-ai/FlashMLA.git@${FLASH_MLA_COMMIT}" \
+        "nvidia-cudnn-frontend[cutedsl] @ git+https://github.com/NVIDIA/cudnn-frontend.git@${CUDNN_FRONTEND_COMMIT}" \
+        "torch-c-dlpack-ext==0.1.5"
 EOF
 
 # Install DeepEP
@@ -83,6 +90,11 @@ RUN bash -ex <<"EOF"
 EOF
 
 COPY assets/ /opt/data/
+# Reinstall CuTe DSL cleanly after upgrading the version provided by the base image.
+RUN uv pip install --reinstall --no-deps \
+    nvidia-cutlass-dsl==4.5.0 \
+    nvidia-cutlass-dsl-libs-base==4.5.0 \
+    nvidia-cutlass-dsl-libs-cu13==4.5.0
 ENV UV_PYTHON=$UV_PROJECT_ENVIRONMENT/bin/python
 
 ##### For NVIDIANS only #####
--- a/megatron/core/datasets/data_schedule.py
+++ b/megatron/core/datasets/data_schedule.py
@@ -613,15 +613,6 @@ def get_batch_on_this_rank_for_sequence_packing(
 
     is_first_or_last_stage = is_first_stage or is_last_stage
     dev = torch.cuda.current_device()
-    static_cp_target_len = None
-    if (
-        config is not None
-        and getattr(config, 'experimental_attention_variant', None) == "dsv4_hybrid"
-        and cp_group.size() > 1
-        and getattr(config, 'pad_packed_seq_alignment', None) == "max"
-        and getattr(config, 'max_seqlen_per_dp_cp_rank', None) is not None
-    ):
-        static_cp_target_len = int(config.max_seqlen_per_dp_cp_rank)
 
     # data_iterator should return a batch including the following keys.
     batch_keys = ['cu_seqlens', 'cu_seqlens_padded', 'max_seqlen']
@@ -653,6 +644,32 @@ def get_batch_on_this_rank_for_sequence_packing(
             group_size=local_cp_size_val
         )
 
+    use_dsv4_cp_slice = (
+        config is not None
+        and getattr(config, 'experimental_attention_variant', None) == "dsv4_hybrid"
+    )
+    dsv4_cp_local_target_len = None
+    pad_alignment = (
+        getattr(config, 'pad_packed_seq_alignment', None) if config is not None else None
+    )
+    alignment = target_len = max_num_seqs = None
+    if pad_alignment is not None:
+        alignment, target_len, max_num_seqs = get_thd_padding_kwargs(
+            pad_alignment,
+            getattr(config, 'max_seqlen_per_dp_cp_rank', None),
+            getattr(config, 'thd_max_packed_sequences', None),
+            getattr(config, 'cuda_graph_impl', 'none') != 'none',
+        )
+    if is_tp_rank_0 and use_dsv4_cp_slice and pad_alignment is not None:
+        if target_len is not None:
+            dsv4_cp_local_target_len = target_len
+        else:
+            # Fix the local width before slicing so later padding cannot shift rank origins.
+            assert alignment is not None
+            total_rows = int(batch['cu_seqlens_padded'][-1].item())
+            local_rows = (total_rows + cp_group.size() - 1) // cp_group.size()
+            dsv4_cp_local_target_len = ((local_rows + alignment - 1) // alignment) * alignment
+
     # Build padding_mask before CP slicing while tensors still have the full
     # packed length represented by cu_seqlens_padded[-1].
     if is_tp_rank_0:
@@ -667,20 +684,16 @@ def get_batch_on_this_rank_for_sequence_packing(
         cp_slice_keys = ['padding_mask']
         if is_first_or_last_stage or mtp_on_this_rank:
             cp_slice_keys.extend(['tokens', 'position_ids', 'labels', 'loss_mask'])
-        csa_cp_partition_mode = None
-        if (
-            config is not None
-            and getattr(config, 'experimental_attention_variant', None) == "dsv4_hybrid"
-        ):
-            csa_cp_partition_mode = config.csa_cp_partition_mode
         partition_total_tokens = (
-            static_cp_target_len * cp_group.size() if static_cp_target_len is not None else None
+            dsv4_cp_local_target_len * cp_group.size()
+            if dsv4_cp_local_target_len is not None
+            else None
         )
         get_cp_slice_for_thd(
             batch,
             cp_group,
             keys=cp_slice_keys,
-            csa_cp_partition_mode=csa_cp_partition_mode,
+            use_dsv4_cp_slice=use_dsv4_cp_slice,
             partition_total_tokens=partition_total_tokens,
         )
 
@@ -696,8 +709,8 @@ def get_batch_on_this_rank_for_sequence_packing(
     # Broadcast total_tokens because padding_mask is prepared on every PP stage.
     # Tokens/labels/loss_mask/position_ids use the same length on stages that own them.
     if is_tp_rank_0:
-        if static_cp_target_len is not None:
-            total_tokens = torch.tensor([static_cp_target_len], dtype=torch.int32, device=dev)
+        if dsv4_cp_local_target_len is not None:
+            total_tokens = torch.tensor([dsv4_cp_local_target_len], dtype=torch.int32, device=dev)
         else:
             # Under VPP, the last PP stage has labels but no tokens, so derive
             # total_tokens from cu_seqlens_padded, which is present on every
@@ -824,16 +837,7 @@ def get_batch_on_this_rank_for_sequence_packing(
 
     # Pad the already-packed THD tensors at the end when requested. CUDA Graph
     # additionally pads cu_seqlens tensors to thd_max_packed_sequences + 1 entries.
-    pad_alignment = (
-        getattr(config, 'pad_packed_seq_alignment', None) if config is not None else None
-    )
     if pad_alignment is not None and packed_seq_params is not None:
-        alignment, target_len, max_num_seqs = get_thd_padding_kwargs(
-            pad_alignment,
-            getattr(config, 'max_seqlen_per_dp_cp_rank', None),
-            getattr(config, 'thd_max_packed_sequences', None),
-            getattr(config, 'cuda_graph_impl', 'none') != 'none',
-        )
         tokens, labels, loss_mask, position_ids, packed_seq_params, padding_mask = (
             pad_sequence_for_thd(
                 tokens,
--- a/megatron/core/datasets/data_schedule.py
+++ b/megatron/core/datasets/data_schedule.py
@@ -630,7 +630,7 @@ def get_batch_on_this_rank_for_sequence_packing(
             group_size=local_cp_size_val
         )
 
-    cp_partition_mode = config.cp_partition_mode if config is not None else "zigzag"
+    cp_partition_mode = getattr(config, "cp_partition_mode", "zigzag")
     contiguous_cp_local_target_len = None
     pad_alignment = (
         getattr(config, 'pad_packed_seq_alignment', None) if config is not None else None
--- a/megatron/core/datasets/data_schedule.py
+++ b/megatron/core/datasets/data_schedule.py
@@ -647,6 +647,7 @@ def get_batch_on_this_rank_for_sequence_packing(
     use_dsv4_cp_slice = (
         config is not None
         and getattr(config, 'experimental_attention_variant', None) == "dsv4_hybrid"
+        and cp_group.size() > 1
     )
     dsv4_cp_local_target_len = None
     pad_alignment = (
--- a/megatron/core/datasets/data_schedule.py
+++ b/megatron/core/datasets/data_schedule.py
@@ -613,6 +613,15 @@ def get_batch_on_this_rank_for_sequence_packing(
 
     is_first_or_last_stage = is_first_stage or is_last_stage
     dev = torch.cuda.current_device()
+    static_cp_target_len = None
+    if (
+        config is not None
+        and getattr(config, 'experimental_attention_variant', None) == "dsv4_hybrid"
+        and cp_group.size() > 1
+        and getattr(config, 'pad_packed_seq_alignment', None) == "max"
+        and getattr(config, 'max_seqlen_per_dp_cp_rank', None) is not None
+    ):
+        static_cp_target_len = int(config.max_seqlen_per_dp_cp_rank)
 
     # data_iterator should return a batch including the following keys.
     batch_keys = ['cu_seqlens', 'cu_seqlens_padded', 'max_seqlen']
@@ -658,7 +667,22 @@ def get_batch_on_this_rank_for_sequence_packing(
         cp_slice_keys = ['padding_mask']
         if is_first_or_last_stage or mtp_on_this_rank:
             cp_slice_keys.extend(['tokens', 'position_ids', 'labels', 'loss_mask'])
-        get_cp_slice_for_thd(batch, cp_group, keys=cp_slice_keys)
+        csa_cp_partition_mode = None
+        if (
+            config is not None
+            and getattr(config, 'experimental_attention_variant', None) == "dsv4_hybrid"
+        ):
+            csa_cp_partition_mode = config.csa_cp_partition_mode
+        partition_total_tokens = (
+            static_cp_target_len * cp_group.size() if static_cp_target_len is not None else None
+        )
+        get_cp_slice_for_thd(
+            batch,
+            cp_group,
+            keys=cp_slice_keys,
+            csa_cp_partition_mode=csa_cp_partition_mode,
+            partition_total_tokens=partition_total_tokens,
+        )
 
     # Broadcast cu_seqlens_size because we need it to create placeholder for cu_seqlens and
     # cu_seqlens_padded for non TP 0 ranks.
@@ -672,12 +696,15 @@ def get_batch_on_this_rank_for_sequence_packing(
     # Broadcast total_tokens because padding_mask is prepared on every PP stage.
     # Tokens/labels/loss_mask/position_ids use the same length on stages that own them.
     if is_tp_rank_0:
-        # Under VPP, the last PP stage has labels but no tokens, so derive
-        # total_tokens from cu_seqlens_padded, which is present on every
-        # stage. cu_seqlens_padded keeps the pre-CP packed length; divide
-        # by cp_size to match the already CP-sliced sequence tensors.
-        cp_world = cp_group.size()
-        total_tokens = (batch['cu_seqlens_padded'][-1].to(torch.int32) // cp_world).reshape(1)
+        if static_cp_target_len is not None:
+            total_tokens = torch.tensor([static_cp_target_len], dtype=torch.int32, device=dev)
+        else:
+            # Under VPP, the last PP stage has labels but no tokens, so derive
+            # total_tokens from cu_seqlens_padded, which is present on every
+            # stage. cu_seqlens_padded keeps the pre-CP packed length; divide
+            # by cp_size to match the already CP-sliced sequence tensors.
+            cp_world = cp_group.size()
+            total_tokens = (batch['cu_seqlens_padded'][-1].to(torch.int32) // cp_world).reshape(1)
     else:
         total_tokens = torch.empty(1, dtype=torch.int32, device=dev)
     broadcast_tensor(total_tokens, tp_src_rank, tp_group)
--- a/megatron/core/datasets/data_schedule.py
+++ b/megatron/core/datasets/data_schedule.py
@@ -587,8 +587,8 @@ def get_batch_on_this_rank_for_sequence_packing(
         data_iterator (Iterator): The data iterator to get the batch from.
         mtp_on_this_rank (bool): Whether to use multi-token prediction.
         vp_stage (Optional[int]): The stage of the pipeline.
-        config: Model parallel config used for optional THD packed-sequence padding.
-            When None or config.pad_packed_seq_alignment is None, no padding is applied.
+        config: Model config used for CP partitioning and optional THD packed-sequence padding.
+            When None, CP partitioning defaults to zigzag and no padding is applied.
     Returns:
         tuple of (tokens, labels, loss_mask, attention_mask, position_ids,
         packed_seq_params, padding_mask)
@@ -644,11 +644,7 @@ def get_batch_on_this_rank_for_sequence_packing(
             group_size=local_cp_size_val
         )
 
-    use_contiguous_cp_slice = (
-        config is not None
-        and getattr(config, 'experimental_attention_variant', None) == "dsv4_hybrid"
-        and cp_group.size() > 1
-    )
+    cp_partition_mode = config.cp_partition_mode if config is not None else "zigzag"
     contiguous_cp_local_target_len = None
     pad_alignment = (
         getattr(config, 'pad_packed_seq_alignment', None) if config is not None else None
@@ -661,7 +657,12 @@ def get_batch_on_this_rank_for_sequence_packing(
             getattr(config, 'thd_max_packed_sequences', None),
             getattr(config, 'cuda_graph_impl', 'none') != 'none',
         )
-    if is_tp_rank_0 and use_contiguous_cp_slice and pad_alignment is not None:
+    if (
+        is_tp_rank_0
+        and cp_group.size() > 1
+        and cp_partition_mode == "contiguous"
+        and pad_alignment is not None
+    ):
         if target_len is not None:
             contiguous_cp_local_target_len = target_len
         else:
@@ -694,7 +695,7 @@ def get_batch_on_this_rank_for_sequence_packing(
             batch,
             cp_group,
             keys=cp_slice_keys,
-            use_contiguous_cp_slice=use_contiguous_cp_slice,
+            cp_partition_mode=cp_partition_mode,
             partition_total_tokens=partition_total_tokens,
         )
 
@@ -835,6 +836,7 @@ def get_batch_on_this_rank_for_sequence_packing(
         max_seqlen_kv=max_seqlen,
         local_cp_size=local_cp_size,
         cp_group=cp_group,
+        cp_partition_mode=cp_partition_mode,
         pad_between_seqs=False,
     )
 
--- a/megatron/core/datasets/data_schedule.py
+++ b/megatron/core/datasets/data_schedule.py
@@ -644,12 +644,12 @@ def get_batch_on_this_rank_for_sequence_packing(
             group_size=local_cp_size_val
         )
 
-    use_dsv4_cp_slice = (
+    use_contiguous_cp_slice = (
         config is not None
         and getattr(config, 'experimental_attention_variant', None) == "dsv4_hybrid"
         and cp_group.size() > 1
     )
-    dsv4_cp_local_target_len = None
+    contiguous_cp_local_target_len = None
     pad_alignment = (
         getattr(config, 'pad_packed_seq_alignment', None) if config is not None else None
     )
@@ -661,15 +661,15 @@ def get_batch_on_this_rank_for_sequence_packing(
             getattr(config, 'thd_max_packed_sequences', None),
             getattr(config, 'cuda_graph_impl', 'none') != 'none',
         )
-    if is_tp_rank_0 and use_dsv4_cp_slice and pad_alignment is not None:
+    if is_tp_rank_0 and use_contiguous_cp_slice and pad_alignment is not None:
         if target_len is not None:
-            dsv4_cp_local_target_len = target_len
+            contiguous_cp_local_target_len = target_len
         else:
             # Fix the local width before slicing so later padding cannot shift rank origins.
             assert alignment is not None
             total_rows = int(batch['cu_seqlens_padded'][-1].item())
             local_rows = (total_rows + cp_group.size() - 1) // cp_group.size()
-            dsv4_cp_local_target_len = ((local_rows + alignment - 1) // alignment) * alignment
+            contiguous_cp_local_target_len = ((local_rows + alignment - 1) // alignment) * alignment
 
     # Build padding_mask before CP slicing while tensors still have the full
     # packed length represented by cu_seqlens_padded[-1].
@@ -686,15 +686,15 @@ def get_batch_on_this_rank_for_sequence_packing(
         if is_first_or_last_stage or mtp_on_this_rank:
             cp_slice_keys.extend(['tokens', 'position_ids', 'labels', 'loss_mask'])
         partition_total_tokens = (
-            dsv4_cp_local_target_len * cp_group.size()
-            if dsv4_cp_local_target_len is not None
+            contiguous_cp_local_target_len * cp_group.size()
+            if contiguous_cp_local_target_len is not None
             else None
         )
         get_cp_slice_for_thd(
             batch,
             cp_group,
             keys=cp_slice_keys,
-            use_dsv4_cp_slice=use_dsv4_cp_slice,
+            use_contiguous_cp_slice=use_contiguous_cp_slice,
             partition_total_tokens=partition_total_tokens,
         )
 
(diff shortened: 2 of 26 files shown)


megatron/core/transformer/experimental_attention_variant/dsa_kernels.py
```diff
@@ -1459,6 +1495,167 @@ def backward(ctx, grad_output, grad_loss):
...
+        k_indexer: Tensor,
+        weights: Tensor,
+        indexer_topk_idxs: Tensor,
+        compressed_kv: Tensor,
+        softmax_scale: float,
+        indexer_softmax_scale: float,
+        loss_coeff: float,
+        calculate_per_token_loss: bool,
+        global_query_rows: int,
+    ) -> Tuple[Tensor, Tensor]:
```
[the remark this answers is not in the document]
kunlunl: Fixed.

megatron/core/datasets/data_schedule.py
```diff
@@ -658,7 +667,19 @@ def get_batch_on_this_rank_for_sequence_packing(
         cp_slice_keys = ['padding_mask']
         if is_first_or_last_stage or mtp_on_this_rank:
             cp_slice_keys.extend(['tokens', 'position_ids', 'labels', 'loss_mask'])
-        get_cp_slice_for_thd(batch, cp_group, keys=cp_slice_keys)
+        csa_cp_partition_mode = None
+        if config is not None and config.experimental_attention_variant == "dsv4_hybrid":
+            csa_cp_partition_mode = config.csa_cp_partition_mode
```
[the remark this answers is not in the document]
kunlunl: Done: changed this access to use the same defensive `getattr(config, 'experimental_attention_variant', None)` pattern.

megatron/core/transformer/experimental_attention_variant/dsa_kernels.py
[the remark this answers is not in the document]
kunlunl: Done: CP now builds a local `q_padding_mask` and excludes those rows from the from-top-k indexer loss/backward path.

megatron/core/transformer/experimental_attention_variant/csa.py
```diff
@@ -617,6 +619,84 @@ def unfused_compressed_sparse_attn(
...
+    if calculate_per_token_loss:
+        indexer_loss = raw_local_loss * loss_coeff
+    else:
+        if global_query_rows <= 0:
+            raise RuntimeError(f"global_query_rows must be positive, got {global_query_rows}.")
+        indexer_loss = raw_local_loss * (float(loss_coeff) / float(global_query_rows))
+    return output, indexer_loss
+
+
 # ---------------------------------------------------------------------------
```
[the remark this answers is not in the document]
kunlunl: Done: added a comment explaining that the unfused fallback recomputes the target because it does not run FlashMLA and does not have `lse_indexer`.

megatron/core/transformer/experimental_attention_variant/csa_cp_utils.py
```diff
@@ -0,0 +1,926 @@
...
+        comp_parts = comp_parts + (
+            torch.full((pad_groups,), -1, dtype=torch.int32, device=hidden_local.device),
+        )
+    cu_deltas = sum(cu[1:] - cu[:-1] for cu in cu_parts)
+    cu_compact = torch.cat(
+        (
+            torch.zeros((1,), dtype=cu_seqlens.dtype, device=cu_seqlens.device),
+            torch.cumsum(cu_deltas, dim=0),
+        )
+    )
```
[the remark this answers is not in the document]
kunlunl: Done: replaced `sum(...)` with explicit first-element initialization and a loop.

megatron/core/transformer/experimental_attention_variant/csa_cp_utils.py
```diff
@@ -0,0 +1,926 @@
...
+    cp_group: torch.distributed.ProcessGroup,
+) -> torch.Tensor:
+    """Exchange hidden-state boundary rows for the selected CSA CP partition mode."""
+    ratios = (ratio for ratio in csa_compress_ratios if ratio and ratio > 1)
+    d_comp = max((8 if ratio == 4 else ratio for ratio in ratios), default=0)
+    d_window = max(int(csa_window_size), d_comp)
+    hidden_flat = hidden_states.view(hidden_states.shape[0], -1)
+    cp_size = 1 if cp_group is None else cp_group.size()
+    cp_rank = 0 if cp_group is None else cp_group.rank()
+    local_chunks = len(
```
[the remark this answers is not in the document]
kunlunl: Done: removed the intermediate `ratios` generator and inlined the expression.

megatron/core/transformer/experimental_attention_variant/csa_cp_utils.py
```diff
@@ -0,0 +1,926 @@
...
+    d_window = boundary_rows // len(local_ranges)
+    boundary_ranges = tuple((int(start) - d_window, int(start)) for start, _ in local_ranges)
+    return boundary_ranges + local_ranges
+
+
+def _compressed_group_capacity(rows: int, ratio: int, d_comp: int) -> int:
+    """Return compressed group capacity with row alignment for TE FP8 linear inputs."""
+    c_cap = (int(rows) + int(d_comp)) // int(ratio)
+    group_alignment = 32 // math.gcd(32, int(ratio))
+    return ((c_cap + group_alignment - 1) // group_alignment) * group_alignment
```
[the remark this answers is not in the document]
kunlunl: Done: added a `max(1, ...)` guard before applying the alignment.

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
```diff
@@ -730,21 +836,28 @@ def qkv_up_proj_and_rope_apply(q_compressed, kv_compressed, k_pos_emb, rotary_po
...
             )
         else:
-            query, key, value = qkv_up_proj_and_rope_apply(
-                q_compressed, kv_compressed, k_pos_emb, rotary_pos_emb
+            query, key, value, boundary_kv = qkv_up_proj_and_rope_apply(
+                q_compressed, kv_compressed, k_pos_emb, rotary_pos_emb, boundary_kv_compressed
             )
 
-        return query, key, value, q_compressed, kv_compressed
+        return query, key, value, q_compressed, kv_compressed, boundary_kv
```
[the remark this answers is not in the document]
kunlunl: Done: added a return docstring documenting the 6-tuple and `boundary_kv`.

megatron/core/transformer/experimental_attention_variant/csa_cp_layout_kernels.py
```diff
@@ -0,0 +1,3251 @@
...
+                                                source_index = rank_major
+                                else:
+                                    for comp_row in range(compressed_rows):
+                                        if (
+                                            valid_rank_major[comp_row]
+                                            and seq_ids_rank_major[comp_row] == seq
+                                            and comp_ids_rank_major[comp_row] == comp_id
+                                        ):
+                                            source_kind = 2
+                                            source_index = comp_row
```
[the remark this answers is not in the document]
kunlunl: Done: added a comment noting this is a slow fallback for no-indexer paths and not the hot training path.

kunlunl:
/ok to test [b83d57e](https://github.com/NVIDIA/Megatron-LM/pull/5087/commits/b83d57e9502e72b9eca73ee34835e58c2f968a3b)

kunlunl:
/ok to test [8a271be](https://github.com/NVIDIA/Megatron-LM/pull/5087/commits/8a271beb67cadd9de9d9eebd329ec94db52afb10)

kunlunl:
/ok to test [3aeb5f5](https://github.com/NVIDIA/Megatron-LM/pull/5087/commits/3aeb5f5bcbf382ef43d10ac9797e0698b1ab8852)

kunlunl:
/ok to test [2acda35](https://github.com/NVIDIA/Megatron-LM/pull/5087/commits/2acda35b90739f1d9dd4aa59a69bcf4524593bb6)

kunlunl:
/ok to test [2acda35](https://github.com/NVIDIA/Megatron-LM/pull/5087/commits/2acda35b90739f1d9dd4aa59a69bcf4524593bb6)

kunlunl:
/ok to test [867d4c8](https://github.com/NVIDIA/Megatron-LM/pull/5087/commits/867d4c80458a249ccbb9626cf4efb50c8e06c38d)

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
```diff
@@ -262,18 +263,37 @@ def forward(
             assert packed_seq_params.cp_group is not None, "cp_group must be set in dynamic-cp mode"
             self.pg_collection.cp = packed_seq_params.cp_group
 
+        cp_size = self.pg_collection.cp.size()
+        qkv_format = packed_seq_params.qkv_format if packed_seq_params is not None else None
+        use_thd_cp = cp_size > 1 and qkv_format == 'thd'
```
Victarry, line 268: [CRITICAL Correctness] qkv_format == "thd" does not guarantee the contiguous row layout assumed by this branch. The non-scheduler packed-data path still uses tex.thd_get_partitioned_indices(), while this code later derives positions as cp_rank * local_rows + local_row. That silently mis-maps RoPE positions, boundary rows, windows, and compressed-group ownership. Please carry an explicit CP-layout marker in PackedSeqParams and either produce the contiguous layout in every supported data path or reject incompatible THD/non-THD CP inputs.

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
```diff
@@ -636,115 +696,195 @@ def qkv_up_proj_and_rope_apply(q_compressed, kv_compressed, k_pos_emb, rotary_po
...
+
+            kv, _ = self.linear_kv_proj(kv_projection_input)
             kv = self.kv_layernorm(kv)
+            boundary_kv = None
 
             # [num_tokens, qk_pos_emb_head_dim] -> [num_tokens, 1, qk_pos_emb_head_dim]
             if k_pos_emb is not None:
                 k_pos_emb = torch.unsqueeze(k_pos_emb, -2)
 
+            cp_size = self.pg_collection.cp.size()
```
Victarry: [CRITICAL Correctness] This checkpointed closure re-reads mutable self.pg_collection.cp during backward recomputation. forward() restores the original group before returning, so dynamic-CP recomputation can use a different rank/size than the forward pass and derive the wrong global_start and RoPE positions. Capture the active microbatch CP group explicitly in the checkpoint arguments instead of reading module state, and add dynamic-CP + selective mla_up_proj recompute parity coverage.

megatron/core/transformer/experimental_attention_variant/csa.py
```diff
@@ -1569,6 +1703,8 @@ def forward(
         attention_mask: torch.Tensor,
         x: torch.Tensor = None,
         qr: torch.Tensor = None,
+        boundary_hidden: Optional[torch.Tensor] = None,
```
Victarry: [IMPORTANT Compatibility] Inserting boundary_hidden and boundary_kv before the existing optional arguments changes positional argument binding for this public forward() method. Existing callers can silently pass attn_mask_type into boundary_hidden, etc. Please append the new parameters or make them keyword-only so the existing positional contract remains intact.

megatron/core/transformer/experimental_attention_variant/csa_cp_layout_kernels.py
```diff
@@ -0,0 +1,821 @@
...
+                col = linear - row * total_width
+            else:
+                row = l_local
+
+        if row < l_local:
+            global_q = global_start + row
+            seq_start_found = cutlass.Int32(-1)
+            seq_comp_start = cutlass.Int32(0)
+            seq_comp_len = cutlass.Int32(0)
+            for seq in range(n_seq):
```
Victarry: [IMPORTANT Performance] In modes 1/2, every (query, output-column) thread repeats the full for seq in range(n_seq) lookup. This makes index construction scale as local_rows * total_width * n_seq and can dominate for batches containing many short packed sequences. Please resolve sequence metadata once per query/block and broadcast it to the column workers, or precompute per-row metadata; a benchmark varying n_seq would catch this scaling regression.

megatron/core/transformer/experimental_attention_variant/csa.py
```diff
@@ -2026,6 +2167,282 @@ def _forward_fused_indexer_training_thd(
...
+                    cu_seqlens_compressed,
+                    global_start,
+                    cp_size,
+                    ratio,
+                )
+            )
+
+            if indexer is not None:
+                # ---- Step 4: optional indexer compressed path -----------------
+                if (
```
Victarry: [IMPORTANT Compatibility] Enabling CP breaks the existing dense indexer-loss configuration: dsa_indexer_use_sparse_loss defaults to False, so a valid CP1 setup with nonzero loss coefficient fails only on its first forward. Please either implement the dense CP loss path or reject/document this combination during configuration validation so users do not discover the limitation after launch.

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
yuzhongw-nvidia, line 268: Could we fail fast when `cp_size > 1` but `qkv_format != "thd"`?

The new DSv4 CP implementation appears to be THD-specific: boundary hidden exchange, CP-local RoPE, compressor compaction, and final physical index lowering are all wired through the THD CP path. With the current predicate, a DSv4 config with `context_parallel_size > 1` but missing/non-THD `packed_seq_params` silently sets `use_thd_cp = False`; later `CompressedSparseAttention.forward()` falls through to the SBHD CSA path instead of `_forward_thd_cp()`.

That seems risky because it can run with local-only sparse-attention semantics rather than explicit DSv4 CP semantics. I suggest raising a clear error here or in config validation when DSv4 Hybrid sees `cp_size > 1` without THD packed sequence input, unless SBHD CP is intentionally supported elsewhere. A small unit test for CP>1 + missing/non-THD `PackedSeqParams` would also protect this contract.

tests/unit_tests/transformer/experimental_attention_variant/test_dsv4_hybrid_attention_cp.py
```diff
@@ -0,0 +1,1000 @@
...
+    max_len = int(max(padded_seg_lens)) if padded_seg_lens else 0
+    return PackedSeqParams(
+        cu_seqlens_q=cu,
+        cu_seqlens_q_padded=cu_padded,
+        cu_seqlens_kv=cu,
+        cu_seqlens_kv_padded=cu_padded,
+        max_seqlen_q=max_len,
+        max_seqlen_kv=max_len,
+        qkv_format='thd',
+    )
```
yuzhongw-nvidia, line 129: The DSv4 CP attention parity tests build `PackedSeqParams(qkv_format="thd", ...)` without `local_cp_size` or `cp_group`, so they exercise static CP groups only. The data schedule test separately checks that dynamic CP propagates `local_cp_size`, but I don't see an end-to-end DSv4 attention test that verifies the dynamic CP group handoff.

This handoff is important because `DSv4HybridAttention.forward()` temporarily replaces `self.pg_collection.cp` from `packed_seq_params.cp_group` before computing `use_thd_cp`, RoPE global starts, and boundary exchange. If that integration breaks, the static CP parity tests and the data-schedule test could still pass independently.

Could we add one focused integration test where the static CP group is larger, `PackedSeqParams.local_cp_size/cp_group` points to a smaller dynamic CP group, and the DSv4 attention forward verifies the dynamic group is the one used? One ratio-4 or ratio-128 case should be enough; no need to duplicate the whole parity matrix.

tests/unit_tests/transformer/experimental_attention_variant/test_dsv4_hybrid_attention_cp.py
```diff
@@ -0,0 +1,1000 @@
...
+_DSV4_CP_RAGGED_PADDED_SEG_LENS = (1, 127, 1000, 23, 129, 900, 55, 257, 800, 95, 509, 200)
+# Same shape, padded total, and max padded sequence length as
+# _DSV4_CP_RAGGED_PADDED_SEG_LENS, but the padding is distributed across many
+# sequences instead of only the tail. CUDA graph replay must tolerate this value
+# change because the CP path rebuilds compressed-row metadata from device
+# cu_seqlens_padded rather than from capture-time host sizes.
+_DSV4_CP_REPLAY_PADDED_SEG_LENS = (8, 128, 1000, 32, 132, 904, 64, 260, 804, 96, 512, 156)
+
+
+def _dsv4_cp_fused_kernels_available():
```
yuzhongw-nvidia, line 74: This helper is useful, but it also means the fused DSv4 CP production path can be skipped entirely unless the test machine has SM100+, `flash_mla`, and `cudnn.DSA`. The CUDA graph metadata replay, memory scale, and graph-time scale guards all depend on this availability, and the CuTe layout tests have their own CUDA/CuTe skip.

Could we make sure the PR test plan names at least one CI lane or manual validation environment where these fused tests actually run? Otherwise a normal CI environment can be green while the production fused path is untested.

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
```diff
@@ -601,13 +656,16 @@ def get_query_key_value_tensors(
 
         kv_compressed = hidden_states
```
yuzhongw-nvidia, line 665: This naming is confusing in the DSv4 path. At this point `kv_compressed = hidden_states` and `boundary_kv_compressed = boundary_hidden` are not compressed KV tensors yet; they are pre-projection hidden-state inputs that will later feed `linear_kv_proj` / the compressor path.

Could we rename these to something like `kv_projection_input_local` and `boundary_kv_projection_input`, or add a direct comment explaining that they are pre-projection inputs rather than compressed KV outputs? This path is already carrying several logical/physical layouts, so avoiding misleading names would make future reviews safer.

megatron/core/transformer/experimental_attention_variant/csa_cp_utils.py
```diff
@@ -0,0 +1,401 @@
...
+                )
+            )
+        if cp_rank + 1 < cp_size:
+            send_tail = tensor[-d_window:].contiguous()
+            ops.append(
+                dist.P2POp(
+                    dist.isend, send_tail, dist.get_global_rank(cp_group, cp_rank + 1), cp_group
+                )
+            )
+        for req in dist.batch_isend_irecv(ops):
```
yuzhongw-nvidia, line 155: How about overlapping boundary communications with the Q projection? Or we can leave it as our future work.

megatron/core/transformer/experimental_attention_variant/csa_cp_utils.py
```diff
@@ -0,0 +1,401 @@
...
+    """Build fixed-capacity compressor input for this rank's token block.
+
+    Returns:
+        ``hidden_compact``: rank-local compressor input, shape
+            ``(compact_group_capacity * ratio, ...)``.
+        ``compressed_group_ids``: original per-sequence compressed group id for each
+            compact group, shape ``(compact_group_capacity,)``. For example,
+            with ``ratio=4``, ``comp_id=3`` maps to RoPE position ``12``.
+        ``seq_to_rank_row``: map from global sequence-major compressed rows
+            to their canonical rank-major all-gather rows.
```
yuzhongw-nvidia, line 227: The current comments are helpful, but I think this mapping needs one concrete example. `compressed_group_ids` is a per-sequence local compressed group id, while `seq_to_rank_row` maps a logical sequence-major compressed row to a physical row in the rank-major all-gathered buffer. That distinction is subtle and easy to break during maintenance.

Could we add a compact example with two packed sequences and two CP ranks, showing something like sequence-major compressed rows `[Acomp0, Acomp1, Bcomp0, ...]`, rank-major physical rows after all-gather, and the resulting `seq_to_rank_row` values? This would make the ownership rule and physical/logical row spaces much easier to audit.

megatron/core/transformer/experimental_attention_variant/csa.py
```diff
@@ -2026,6 +2167,282 @@ def _forward_fused_indexer_training_thd(
...
+                k_indexer_rank_major = gather_from_sequence_parallel_region(
+                    indexer_compressed_local.squeeze(1), group=cp_group
+                )
+
+                # Indexer top-k consumes sequence-major K rows. Capacity-tail
+                # map entries are unused by cu_seqlens_compressed.
+                k_indexer_seq_major = torch.index_select(
+                    k_indexer_rank_major, 0, seq_to_rank_row.clamp_min(0)
+                )
+                # Top-k is still in logical compressed coordinates here.
```
yuzhongw-nvidia: Could we expand this comment slightly to spell out the two compressed index spaces used below?

`compressed_topk` is still a logical per-sequence compressed id, e.g. "select compressed block 2 inside this packed segment". `indexer_topk_rank_major`, returned later by `build_attention_indices(..., for_indexer_loss=True)`, is a physical row id into the rank-major compressed K/KV buffers and does not include the `kv_full_thd` compressed-base offset. `topk_idxs` is a third space: physical rows in `kv_full_thd`.

These names are close enough that future contributors can easily mix them up. A short comment here would reduce the chance of breaking the indexer-loss path.

yuzhongw-nvidia:
Thanks for your incredible work! It overall LGTM, and I left several minor comments, fyi.

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
kunlunl: Fixed. CP layout is now an explicit `cp_partition_mode` carried by the config and packed-sequence metadata. CP configurations validate DSv4 as `contiguous` and other attention variants as `zigzag`.

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
kunlunl: Fixed. The current microbatch CP group is now captured as a checkpoint argument, so backward recomputation no longer reads the restored mutable module state. Added selective-recompute forward/backward unit test also.

megatron/core/transformer/experimental_attention_variant/csa.py
kunlunl: Fixed. The new boundary tensors were moved after the existing optional parameters, preserving positional-call compatibility.

megatron/core/transformer/experimental_attention_variant/csa_cp_layout_kernels.py
kunlunl: Fixed inside the kernel without adding metadata. Modes 1 and 2 now scan sequence boundaries once per query row and share the result across the block; many-short-sequence correctness unit test was added.

megatron/core/transformer/experimental_attention_variant/csa.py
kunlunl: Fixed. Configuration validation now rejects dense ratio-4 indexer loss with DSv4 CP because the current cuDNN fused attention API cannot express the required CP-local query offset.

megatron/core/transformer/experimental_attention_variant/dsa_kernels.py
```diff
@@ -584,7 +602,16 @@ def _indexer_topk_core(
...
-    topk_length = (topk_indices >= 0).sum(dim=-1).int()  # (total_q,)
+    if is_thd:
+        row_valid = (topk_indices >= 0) & (topk_indices < seq_lens.unsqueeze(1))
+        topk_indices = topk_indices.masked_fill(~row_valid, -1)
+        safe_topk = topk_indices.clamp(min=0, max=max(sk - 1, 0)).to(torch.long)
+        selected_scores = torch.gather(scores_flat, dim=-1, index=safe_topk)
+        selected_valid = (topk_indices >= 0) & (topk_indices < sk) & torch.isfinite(selected_scores)
+        topk_indices = topk_indices.masked_fill(~selected_valid, -1)
+        topk_length = (topk_indices >= 0).sum(dim=-1).int()
+    else:
```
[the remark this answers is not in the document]
kunlunl: Fixed. Empty key sequences are rejected before entering fused indexer/top-k code.

megatron/core/transformer/experimental_attention_variant/csa_cp_layout_kernels.py
```diff
@@ -0,0 +1,821 @@
...
+        launch_fn.__name__,
+        tuple(
+            (tensor.dtype, tuple(tensor.shape), tuple(tensor.stride())) for tensor in tensor_args
+        ),
+        tuple((i, scalar_args[i]) for i in static_arg_indices),
+    )
+    compiled = _COMPILED_LAUNCH_CACHE.get(key)
+    if compiled is None:
+        cap = torch.cuda.get_device_capability()
+        arch = {(9, 0): "sm_90a", (10, 0): "sm_100a", (10, 3): "sm_103a"}.get(cap)
```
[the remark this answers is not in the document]
kunlunl: Fixed. The unsupported-architecture error now lists all architectures implemented by this kernel: `sm_90a`, `sm_100a`, and `sm_103a`.

megatron/core/transformer/experimental_attention_variant/csa.py
```diff
@@ -2026,6 +2167,282 @@ def _forward_fused_indexer_training_thd(
...
+        cp_rank = cp_group.rank()
+
+        l_local = query.shape[0]
+        if l_local != key.shape[0]:
+            raise RuntimeError("DSv4 THD CP path currently supports self-attention only.")
+        cu_seqlens = (
+            packed_seq_params.cu_seqlens_q_padded
+            if packed_seq_params.cu_seqlens_q_padded is not None
+            else packed_seq_params.cu_seqlens_q
+        )
```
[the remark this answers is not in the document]
kunlunl: Fixed. The CP boundary arguments are now typed as optional, matching their existing validation and call sites.

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
```diff
@@ -636,115 +696,195 @@ def qkv_up_proj_and_rope_apply(q_compressed, kv_compressed, k_pos_emb, rotary_po
...
+            else:
+                query, key, value, boundary_kv = qkv_up_proj_and_rope_apply(
+                    q_compressed, kv_compressed, k_pos_emb, rotary_pos_emb, boundary_kv_compressed
+                )
 
-        return query, key, value, q_compressed, kv_compressed
+        result = (query, key, value, q_compressed, kv_compressed)
+        if boundary_kv is not None:
+            return result + (boundary_kv,)
+        return result
```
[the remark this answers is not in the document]
kunlunl: `CheckpointWithoutOutput` requires tensor outputs because it releases and later restores their storage, so it cannot safely return `None`. A fixed outer tuple would also change the existing non-CP return contract; I kept the current changes.

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
```diff
@@ -636,115 +696,195 @@ def qkv_up_proj_and_rope_apply(q_compressed, kv_compressed, k_pos_emb, rotary_po
...
-            )
+            if boundary_kv_compressed is None:
+                query, key, value = qkv_up_proj_and_rope_apply(
+                    q_compressed, kv_compressed, k_pos_emb, rotary_pos_emb
+                )
+                boundary_kv = None
+            else:
+                query, key, value, boundary_kv = qkv_up_proj_and_rope_apply(
+                    q_compressed, kv_compressed, k_pos_emb, rotary_pos_emb, boundary_kv_compressed
+                )
```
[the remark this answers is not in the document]
kunlunl: `CheckpointWithoutOutput` requires tensor outputs because it releases and later restores their storage, so it cannot safely return `None`. A fixed outer tuple would also change the existing non-CP return contract; I kept the current changes.

megatron/core/datasets/data_schedule.py
```diff
@@ -644,6 +644,33 @@ def get_batch_on_this_rank_for_sequence_packing(
...
+            getattr(config, 'thd_max_packed_sequences', None),
+            getattr(config, 'cuda_graph_impl', 'none') != 'none',
+        )
+    if is_tp_rank_0 and use_contiguous_cp_slice and pad_alignment is not None:
+        if target_len is not None:
+            contiguous_cp_local_target_len = target_len
+        else:
+            # Fix the local width before slicing so later padding cannot shift rank origins.
+            assert alignment is not None
+            total_rows = int(batch['cu_seqlens_padded'][-1].item())
```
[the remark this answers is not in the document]
kunlunl: Removing this `.item()` alone would not remove the synchronization: the same host metadata is required immediately afterward to build the padding mask and broadcast shapes. Dynamic varlen batches also make it unsafe to cache, so I kept it local and explicit.

megatron/core/transformer/experimental_attention_variant/csa_cp_utils.py
```diff
@@ -0,0 +1,401 @@
...
+            )
+        if cp_rank + 1 < cp_size:
+            send_tail = tensor[-d_window:].contiguous()
+            ops.append(
+                dist.P2POp(
+                    dist.isend, send_tail, dist.get_global_rank(cp_group, cp_rank + 1), cp_group
+                )
+            )
+        for req in dist.batch_isend_irecv(ops):
+            req.wait()
```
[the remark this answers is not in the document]
kunlunl: PyTorch already returns the original tensor when `.contiguous()` receives contiguous input, without allocating. An explicit conditional would duplicate that check, so I kept the direct call.

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
kunlunl: Done. `TransformerConfig` now rejects static or dynamic DSv4 CP without a sequence-packing scheduler, and `DSv4HybridAttention.forward()` checks the actual microbatch CP group and raises when `cp_size > 1` but `qkv_format != "thd"`. The runtime check remains necessary because dynamic CP can differ from the static configuration.

tests/unit_tests/transformer/experimental_attention_variant/test_dsv4_hybrid_attention_cp.py
kunlunl: Added ratio-4 forward/backward unit test to the selective recompute test. The eager and recompute modules both start from a CP1 group, then `PackedSeqParams.local_cp_size/cp_group` selects CP2 or CP4 for the microbatch. Fused and unfused RoPE both compare the output, hidden-state gradient, and parameter gradients. This specifically validates that recompute preserves the microbatch group relative to eager.

tests/unit_tests/transformer/experimental_attention_variant/test_dsv4_hybrid_attention_cp.py
kunlunl: I ran the test locally on 4xGB200, all test cases passed. And the memory and perf scale are:
* Peak allocated-memory scale relative to CP1 (ratio-0, ratio-4, ratio-128):
  * CP2: [0.578741, 0.563600, 0.590874] (limit in unit test: <= 0.65)
  * CP4: [0.314517, 0.304487, 0.321605] (limit in unit test: <= 0.35)
* CUDA graph replay-time scale relative to CP1 (ratio-0, ratio-4, ratio-128):
  * CP2: [0.598083, 0.595602, 0.590579] (limit in unit test: <= 0.70)
  * CP4: [0.365788, 0.382846, 0.383631] (limit in unit test: <= 0.50)

megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py
kunlunl: Added a comment instead of renaming these pre-existing variables. It clarifies that `kv_compressed` and `boundary_kv_compressed` are hidden-state inputs to `linear_kv_proj`, while the actual DSv4 compressed KV is produced later by the CSA compressor.

megatron/core/transformer/experimental_attention_variant/csa_cp_utils.py
kunlunl: The opportunity is broader than overlapping the boundary P2P exchange with the Q projection, the indexer K all-gather may overlap with attention pre-attention computation, and the attention compressed-KV all-gather may overlap with the indexer forward. We can explore them as follow-up work rather than in this PR, otherwise this PR is too complicated.

megatron/core/transformer/experimental_attention_variant/csa_cp_utils.py
kunlunl: Added a two-sequence, two-rank example. It shows logical rows `[A0, A1, B0, B1]`, their capacity-padded rank-major storage `[A0, A1, pad, pad | B0, B1, pad, pad]`, and the resulting `seq_to_rank_row` mapping `[0, 1, 4, 5]`.

megatron/core/transformer/experimental_attention_variant/csa.py
kunlunl: Expanded the comments to distinguish all three spaces: `compressed_topk` contains logical per-sequence compressed ids, `indexer_topk_rank_major` addresses the rank-major compressed buffers without the `kv_full_thd` base, and `topk_idxs` addresses the final physical rows in `kv_full_thd`.

kunlunl:
/ok to test [cad0ac5](https://github.com/NVIDIA/Megatron-LM/pull/5087/commits/cad0ac5dd1a0212a319f38e385c8530743b5da38)

Victarry (approved):

megatron/core/transformer/transformer_config.py
```diff
@@ -1457,6 +1460,28 @@ def __post_init__(self):
...
+            if (
+                self.experimental_attention_variant == "dsv4_hybrid"
+                and self.cp_partition_mode != "contiguous"
+            ):
+                raise ValueError("DSv4 Hybrid with CP requires cp_partition_mode='contiguous'.")
+            if (
+                self.experimental_attention_variant != "dsv4_hybrid"
+                and self.cp_partition_mode != "zigzag"
+            ):
+                raise ValueError("cp_partition_mode='contiguous' is only supported with DSv4.")
```
yuzhongw-nvidia:
```suggestion
                raise ValueError("cp_partition_mode='contiguous' currently is only supported with dsv4_hybrid.")
```

yuzhongw-nvidia: Could we make this validation capability-oriented rather than DSv4-only? The current behavior is fine for this PR, but `cp_partition_mode` is a general layout concept and is carried in `PackedSeqParams`, so saying `"contiguous" is only supported with DSv4` may become a small extensibility burden for future mixed stacks.

xlsunstar:
indexer_top_k_wrapper compiles the kernel with bucketed_num_cols = next_pow2(num_cols), but scores_flat is still allocated with the original num_cols. When num_cols is not a power of two, the kernel may access beyond the allocated buffer. Padding scores_flat to the next power of two (using -inf) keeps the runtime tensor shape consistent with the compiled kernel and avoids potential out-of-bounds accesses.
```
--- a/megatron/core/transformer/experimental_attention_variant/dsa_kernels.py
+++ b/megatron/core/transformer/experimental_attention_variant/dsa_kernels.py
@@ -599,6 +599,15 @@ def _indexer_topk_core(
 
     # ---------------- Shared: radix top-K + pad-to-topk -----------------
     topk_k = min(topk, sk)
+
+    # indexer_top_k_wrapper compiles the kernel with bucketed_num_cols=next_pow2(num_cols),
+    # but allocates buffer_torch with actual num_cols — causing out-of-bounds when num_cols
+    # is not a power of 2. Pad scores_flat to the next power of 2 to keep them consistent.
+    _ncols = scores_flat.shape[-1]
+    _ncols_p2 = 1 << (_ncols - 1).bit_length() if _ncols & (_ncols - 1) else _ncols
+    if _ncols_p2 != _ncols:
+        scores_flat = torch.nn.functional.pad(scores_flat, (0, _ncols_p2 - _ncols), value=float('-inf'))
+
     tk_result = _DSA.indexer_top_k_wrapper(
         scores_flat, seq_lens, top_k=topk_k, next_n=1, return_val=False
```

kunlunl:
/ok to test [1ada640](https://github.com/NVIDIA/Megatron-LM/pull/5087/commits/1ada640c9ec312cd2e4aa17d8bc3d0742bfe7c10)

kunlunl:
/ok to test [e1198f6](https://github.com/NVIDIA/Megatron-LM/pull/5087/commits/e1198f6e9fd2ce3e4c8bb58c3fc3fa599d836201)

kunlunl:
> indexer_top_k_wrapper compiles the kernel with bucketed_num_cols = next_pow2(num_cols), but scores_flat is still allocated with the original num_cols. When num_cols is not a power of two, the kernel may access beyond the allocated buffer. Padding scores_flat to the next power of two (using -inf) keeps the runtime tensor shape consistent with the compiled kernel and avoids potential out-of-bounds accesses.
> 
> ```
> --- a/megatron/core/transformer/experimental_attention_variant/dsa_kernels.py
> +++ b/megatron/core/transformer/experimental_attention_variant/dsa_kernels.py
> @@ -599,6 +599,15 @@ def _indexer_topk_core(
>  
>      # ---------------- Shared: radix top-K + pad-to-topk -----------------
>      topk_k = min(topk, sk)
> +
> +    # indexer_top_k_wrapper compiles the kernel with bucketed_num_cols=next_pow2(num_cols),
> +    # but allocates buffer_torch with actual num_cols — causing out-of-bounds when num_cols
> +    # is not a power of 2. Pad scores_flat to the next power of 2 to keep them consistent.
> +    _ncols = scores_flat.shape[-1]
> +    _ncols_p2 = 1 << (_ncols - 1).bit_length() if _ncols & (_ncols - 1) else _ncols
> +    if _ncols_p2 != _ncols:
> +        scores_flat = torch.nn.functional.pad(scores_flat, (0, _ncols_p2 - _ncols), value=float('-inf'))
> +
>      tk_result = _DSA.indexer_top_k_wrapper(
>          scores_flat, seq_lens, top_k=topk_k, next_n=1, return_val=False
> ```

@xlsunstar Thanks for flagging this and for suggesting a fix. `bucketed_num_cols` is used as part of the compilation-cache key and the kernel specialization capacity. In my understanding its purpose is to let nearby runtime widths reuse the same compiled kernel and avoid unnecessary recompilation as `num_cols` changes; it is not intended to define the valid memory-access range.
The actual per-row access range is defined by `seq_lens`, and the kernel is expected to predicate its accesses using that runtime length. MCore sets it accordingly and guarantees `seq_lens <= scores_flat.shape[-1]` in both the THD and BSHD paths.
Have you observed an actual out-of-bounds access or correctness issue here? If so, it may indicate that the cuDNN top-k kernel is not fully honoring the `seq_lens` access-range contract and is instead using the bucketed compilation capacity as the runtime bound. In that case, we would prefer to fix the issue in cuDNN Frontend rather than pad the dense score tensor in MCore.

docker/Dockerfile.ci.dev
```diff
@@ -40,6 +40,8 @@ COPY README.md pyproject.toml uv.lock /workspace/
 COPY megatron/core/__init__.py /workspace/megatron/core/
 COPY megatron/core/package_info.py /workspace/megatron/core/
 ARG IMAGE_TYPE=dev
+ARG CUDNN_FRONTEND_COMMIT=0a14b7181d129d30e7bad34b8c3ed0a0c995e23d
+ARG FLASH_MLA_COMMIT=b7643bd54521f563b839b98289b5cd048c062ba2
```
chtruong814: I think flash_mla is already using the commit you want
https://github.com/NVIDIA/Megatron-LM/blob/main/pyproject.toml#L228

docker/Dockerfile.ci.dev
```diff
@@ -61,7 +63,12 @@ RUN --mount=type=cache,target=/root/.cache/uv \
         --no-install-package nvidia-cusolver-cu12 \
         --no-install-package nvidia-cusparse-cu12 \
         --no-install-package nvidia-cusparselt-cu12 \
-        --no-install-package nvidia-nccl-cu12
+        --no-install-package nvidia-nccl-cu12 \
+        --no-install-package nvidia-cudnn-frontend
+    uv pip install --no-build-isolation \
+        "flash-mla @ git+https://github.com/deepseek-ai/FlashMLA.git@${FLASH_MLA_COMMIT}" \
+        "nvidia-cudnn-frontend[cutedsl] @ git+https://github.com/NVIDIA/cudnn-frontend.git@${CUDNN_FRONTEND_COMMIT}" \
+        "torch-c-dlpack-ext==0.1.5"
```
chtruong814: Please use the pyproject.toml and uv.lock to manage dependencies if possible

docker/Dockerfile.ci.dev
```diff
@@ -40,6 +40,8 @@ COPY README.md pyproject.toml uv.lock /workspace/
 COPY megatron/core/__init__.py /workspace/megatron/core/
 COPY megatron/core/package_info.py /workspace/megatron/core/
 ARG IMAGE_TYPE=dev
+ARG CUDNN_FRONTEND_COMMIT=0a14b7181d129d30e7bad34b8c3ed0a0c995e23d
```
chtruong814: @kunlunl Do we know when cudnn-frontend will include this in a release? Does this need to be included in next 26.08 Megatron-Bridge container release? Is the code backwards compatible to handle if cudnn frontend is not at this commit?

@balasaajay @ko3n1g any concerns?

docker/Dockerfile.ci.dev
```diff
@@ -83,6 +90,11 @@ RUN bash -ex <<"EOF"
 EOF
 
 COPY assets/ /opt/data/
+# Reinstall CuTe DSL cleanly after upgrading the version provided by the base image.
+RUN uv pip install --reinstall --no-deps \
+    nvidia-cutlass-dsl==4.5.0 \
+    nvidia-cutlass-dsl-libs-base==4.5.0 \
+    nvidia-cutlass-dsl-libs-cu13==4.5.0
```
chtruong814: Why do we need to reinstall?  Can we handle this in pyproject toml and uv lock?

chtruong814: @balasaajay @ko3n1g btw, looks like we had to do this in MBridge container though I can't recall why.
https://github.com/NVIDIA-NeMo/Megatron-Bridge/commit/9e06ad9b0592eb4b532b931b034ad33571f727fe

chtruong814 (commented):
If possible, some of the dependency updates should be handled in pyproject toml and uv lock.  I'd like @balasaajay  and @ko3n1g  to weigh in on bumping to the cudnn frontend commit.

docker/Dockerfile.ci.dev
chtruong814: Oh this PR is going into dev branch.  I would just cherry pick this change from main branch 
https://github.com/NVIDIA/Megatron-LM/pull/5448/changes

docker/Dockerfile.ci.dev
chtruong814: @balasaajay @ko3n1g I didn't realize this was going into dev branch.  I don't think much of a concern on the specific cudnn frontend used for dev branch, right?  Main recommendation would still be to move what we can to pyproject toml and uv lock.

kunlunl:
/ok to test [3599a37](https://github.com/NVIDIA/Megatron-LM/pull/5087/commits/3599a37c9b25860da6f49945ad4b65ef1eddf379)

megatron/core/transformer/transformer_config.py
kunlunl: Done.

yuzhongw-nvidia (approved):
LGTM! Thanks for your incredible work!

docker/Dockerfile.ci.dev
kunlunl: Thanks for pointing this out. I mirrored the change from #5448 on the dev branch, pyproject.toml now tracks FlashMLA `nv_dev`, and uv.lock resolves it to b7643bd. I also removed the manual FlashMLA installation from the Dockerfile.

docker/Dockerfile.ci.dev
kunlunl: Done. FlashMLA and cuDNN Frontend are now declared in pyproject.toml, with the exact revisions and the CuTe DSL / DLPack transitive dependencies captured in uv.lock. The Dockerfile now installs this dependency chain through uv sync --locked; the separate manual installation has been removed.

docker/Dockerfile.ci.dev
kunlunl: Thanks for checking. Since this PR targets the dev branch, I pinned the required cuDNN Frontend commit in pyproject.toml / uv.lock for the dev environment.

Older cuDNN Frontend versions remain usable for paths that do not require the new CP offset API. When ratio-4 fused CP is enabled, TransformerConfig now checks that the required wrappers expose q_causal_offsets and raises a clear initialization error if they do not, with the option to disable DSA kernel fusion.

docker/Dockerfile.ci.dev
kunlunl: The explicit reinstall is no longer needed and has been removed. nvidia-cudnn-frontend[cutedsl] now resolves the required CuTe DSL 4.5.0 packages through uv.lock.

chtruong814 (approved):

Issue #5089, opened 2026-06-01
[Feature-request] Add Megatron Core support Gemma 4 dense model (e.g. E4B)


DOGEUNNKIM:
**Is your feature request related to a problem? Please describe.**

Megatron-LM does not currently support Gemma 4 dense models such as Gemma 4 E4B. This makes it difficult to load, validate, and train Gemma 4 dense checkpoints in Megatron Core while preserving Hugging Face parity.

**Describe the solution you'd like**

Based on the scope discussion in this issue, the implementation is split across two layers:

- Megatron Core (generic): GEGLU support and any generic hooks needed for Gemma 4 architectures, kept default-off so existing GPT models are unaffected. - tracked in #5090 
- Megatron-Bridge (Gemma-specific): Dual RoPE, per-layer embeddings, shared KV wiring, HF checkpoint conversion, parity checks, and example scripts — tracked in NVIDIA-NeMo/Megatron-Bridge#4148.

**Describe alternatives you've considered**

One alternative is to keep Gemma 4 support outside Megatron-LM as a downstream fork. However, that makes checkpoint conversion, validation, and maintenance harder, especially as Megatron Core APIs evolve.

**Additional context**

A working implementation covering all Gemma-specific components is available in NVIDIA-NeMo/Megatron-Bridge#4148, including:
- Gemma 4 unit tests and HF layer/block parity checks
- Converted Megatron checkpoint parity against the original HF Gemma 4 E4B model
- TP1, TP2 parity checks for converted checkpoints

The concrete implementation review is tracked in #5090. 

This issue serves as the feature tracking item.

DOGEUNNKIM:
I opened a draft PR with an initial implementation: #5090 

The PR focuses on Gemma 4 dense model support, with Gemma 4 E4B as the initial validated target.

Connor-XY:
Hi @DOGEUNNKIM, thanks for filing this and for already opening #5090 with validation.

I looked at the linked PR. The main MCore issue is not whether Gemma 4 dense support is useful, but where the boundary should sit: MCore should keep generic hooks / GEGLU support, while Gemma-specific layer specs, dual RoPE, PLE/shared-KV wiring, conversion schema, and examples likely belong in Megatron-Bridge. The PR discussion already moved in that direction with NVIDIA-NeMo/Megatron-Bridge#4148, leaving the generic MCore pieces under review.

[...]

PR #5090, opened 2026-06-01
[Feature-request] Add Megatron Core support geglu for Gemma 4 dense model (e.g. E4B)

DOGEUNNKIM:
## Related Issue

Related to #5089

## Summary

This PR adds Gemma 4 dense model support, with Gemma 4 E4B as the initial validated target.

The implementation includes at https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4148.

- Gemma 4 dense model support in Megatron Core
- Gemma 4-specific layer spec
- Dual RoPE support for sliding-window and full-attention layers
- Per-layer embeddings
- Shared-KV runtime wiring
- Hugging Face checkpoint conversion support
- Gemma 4 E4B example scripts and documentation
- Unit tests and HF parity tests

The feature is default-off and only enabled through Gemma 4-specific model configuration/spec usage.

#### Tested with:
```bash
python -m pytest tests/unit_tests/models/test_gemma4.py -v
```

#### Result:
```bash
85 passed
```

#### Converted Gemma 4 E4B checkpoint and ran HF parity check (TP=1 BF16):
```bash
bash examples/gemma4/train_gemma4_e4b_parity.sh
```

#### Result:
```bash
max |diff|  : 2.730531  (atol=3.0)
mean |diff| : 0.110303
--> PASSED
```
#### Converted Gemma 4 E4B checkpoint and ran HF parity check (TP=2 FP32):
```bash
TP_SIZE=2 \
GPUS_PER_NODE=2 \
NVIDIA_VISIBLE_DEVICES=0,1 \ 
BF16=0 \
bash examples/gemma4/train_gemma4_e4b_parity.sh
```

#### Result:
```bash
max |diff|  : 0.151589 (atol=0.3)
mean |diff| : 0.004229
--> PASSED
```

## Notes

Existing GPT model behavior should be unchanged unless Gemma 4-specific config fields are set.

The checkpoint converter supports loading Gemma 4 dense Hugging Face checkpoints into Megatron format.

The TP=2 FP32 parity check validates that the Gemma 4 E4B implementation remains aligned with Hugging Face under tensor-parallel execution.

For setup instructions, checkpoint conversion, and parity validation details, see `examples/gemma4/README.md`.


megatron/core/models/gpt/gpt_model.py
wujingyue, line 1: @Phlip79 are you concerned about changing to gpt_model not hybrid_model?

wujingyue:
Thanks for the contribution! cc @DOGEUNNKIM 

Are you aware of https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/examples/models/gemma/gemma4_vl? Does it support what you need?

DOGEUNNKIM:
> Thanks for the contribution! cc @DOGEUNNKIM
> 
> Are you aware of https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/examples/models/gemma/gemma4_vl? Does it support what you need?

Thanks for pointing this out, and for sharing the Megatron-Bridge Gemma 4 VL example.

Yes, I am aware of that example. My understanding is that it primarily targets the Gemma 4 26B-A4B VLM/MoE workflow in Megatron-Bridge, including checkpoint conversion, inference, and SFT/PEFT recipes.

The scope of this PR is slightly different. This PR focuses on adding native Megatron Core / Megatron-LM support for Gemma 4 E4B-style dense text checkpoints. In particular, it includes the model components needed for HF parity, checkpoint conversion, and Megatron-LM example/parity tests.

As far as I understand, the existing Gemma 4 VL Bridge example does not fully cover the dense text checkpoint path that I need here, especially for directly loading, validating, and training Gemma 4 dense checkpoints in Megatron Core / Megatron-LM.

I would be happy to adjust the scope if you think the boundary between Megatron-LM and Megatron-Bridge should be different. My goal is to make sure the dense Gemma 4 text checkpoint path is supported without unnecessarily duplicating Bridge-level workflows.

One possible split could be to keep this PR focused on the core Gemma 4 dense model/layer support and parity tests in Megatron-LM, while moving higher-level checkpoint loading/conversion workflows and SFT/PEFT recipes to Megatron-Bridge.

yaoyu-33:
Thanks for working on this. I took a closer look and I think we should split this PR a bit.

From the Megatron Bridge side, our usual rule is that MCore should only take changes that are model-agnostic and broadly reusable across model families. Model-family-specific config fields, wiring, checkpoint naming, and HF compatibility logic are usually better kept in Bridge. Otherwise MCore can quickly accumulate one-off args and special cases for each model family, which makes the core API harder to maintain and harder for downstream projects to reason about.

For this PR, several pieces look Gemma4-specific rather than generally reusable:

- `per_layer_inputs` threaded through `TransformerBlock` / recompute looks specific to Gemma4 PLE.
- `GPTModel` directly importing Gemma4 rotary / RMSNorm / KV-sharing logic couples the generic GPT path to Gemma4.
- Config fields like dual RoPE bases, PLE vocab/dim, `attention_k_eq_v`, KV-sharing layers, etc. are documented as Gemma4 behavior.
- The checkpoint saver/schema additions also look Gemma4-specific.

I’d suggest splitting this as:

1. **MCore PR**: keep only the generic pieces, for example:
   - generic `geglu` / `geglu_tanh` activation support, if useful beyond Gemma4;
   - generic GEGLU tensor-parallel checkpoint splitting, if it applies to any GEGLU model;
   - any truly generic hook needed by Bridge, but preferably not named around Gemma4 or PLE specifically.

2. **Megatron Bridge PR**: move the Gemma4 model implementation there:
   - Gemma4 layer specs / attention / RMSNorm / rotary;
   - PLE and shared-KV wiring;
   - Gemma4 config mapping from HF;
   - Gemma4-specific conversion/checkpoint schema;
   - examples and parity tests.

If MCore really needs an extension point for this, I’d prefer a generic hook-style interface rather than adding Gemma4-specific tensors/fields through the core GPT/recompute path.

One concrete issue I noticed: `gpt_builders.py` appears to call `wire_gemma4_kv_sharing(model)` before `model = GPTModel(...)` is assigned, so the current branch may hit an `UnboundLocalError` when that path is enabled.

DOGEUNNKIM:
> Thanks for working on this. I took a closer look and I think we should split this PR a bit.
> 
> From the Megatron Bridge side, our usual rule is that MCore should only take changes that are model-agnostic and broadly reusable across model families. Model-family-specific config fields, wiring, checkpoint naming, and HF compatibility logic are usually better kept in Bridge. Otherwise MCore can quickly accumulate one-off args and special cases for each model family, which makes the core API harder to maintain and harder for downstream projects to reason about.
> 
> For this PR, several pieces look Gemma4-specific rather than generally reusable:
> 
> * `per_layer_inputs` threaded through `TransformerBlock` / recompute looks specific to Gemma4 PLE.
> * `GPTModel` directly importing Gemma4 rotary / RMSNorm / KV-sharing logic couples the generic GPT path to Gemma4.
> * Config fields like dual RoPE bases, PLE vocab/dim, `attention_k_eq_v`, KV-sharing layers, etc. are documented as Gemma4 behavior.
> * The checkpoint saver/schema additions also look Gemma4-specific.
> 
> I’d suggest splitting this as:
> 
> 1. **MCore PR**: keep only the generic pieces, for example:
>    
>    * generic `geglu` / `geglu_tanh` activation support, if useful beyond Gemma4;
>    * generic GEGLU tensor-parallel checkpoint splitting, if it applies to any GEGLU model;
>    * any truly generic hook needed by Bridge, but preferably not named around Gemma4 or PLE specifically.
> 2. **Megatron Bridge PR**: move the Gemma4 model implementation there:
>    
>    * Gemma4 layer specs / attention / RMSNorm / rotary;
>    * PLE and shared-KV wiring;
>    * Gemma4 config mapping from HF;
>    * Gemma4-specific conversion/checkpoint schema;
>    * examples and parity tests.
> 
> If MCore really needs an extension point for this, I’d prefer a generic hook-style interface rather than adding Gemma4-specific tensors/fields through the core GPT/recompute path.
> 
> One concrete issue I noticed: `gpt_builders.py` appears to call `wire_gemma4_kv_sharing(model)` before `model = GPTModel(...)` is assigned, so the current branch may hit an `UnboundLocalError` when that path is enabled.

Thanks for the detailed feedback — the split makes sense and I agree with the architectural principle.

Here's my proposed split:

#### MCore PR (this PR, trimmed down):

- Generic geglu / geglu_tanh activation support in TransformerConfig / MLP — useful for any GEGLU-based model
- Generic GEGLU tensor-parallel checkpoint splitting
- Any generic hook needed by Bridge (named without Gemma4 references)

#### Bridge PR:

- Everything in gemma4_layer_specs.py (layer spec, attention, RMSNorm, rotary, MoE block)
- PLE / shared-KV wiring
- Gemma4 config mapping from HF
- Conversion / checkpoint schema
- Parity tests and examples

I'll also fix the wire_gemma4_kv_sharing ordering bug you caught in gpt_builders.py.

#### One question: 

for per_layer_inputs threading through TransformerBlock — this is generically named and model-agnostic at the MCore level. Moving it to Bridge would require overriding all of TransformerBlock.forward() including pipeline parallel and recompute paths, which is fragile. Would keeping this one generic hook in MCore be acceptable?

DOGEUNNKIM:
We've completed the implementation:

## Megatron-Bridge PR
All Gemma4-specific code has been moved to Bridge : [NVIDIA-NeMo/Megatron-Bridge#4148](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4148)

- Gemma4 layer spec, attention, dual-RoPE, PLE, shared-KV wiring
- HF → Megatron checkpoint conversion and parity check
- `Gemma4E4BProvider` — a Bridge-native provider that works against clean MCore with no Gemma4-specific CLI args or `TransformerConfig` fields

## This MCore PR
Only the minimal MCore-side changes remain:

- Generic `geglu` / `geglu_tanh` activation support
- Generic GEGLU tensor-parallel checkpoint splitting
- Generic `per_layer_inputs` hook threaded through `TransformerBlock` (kept in MCore as requested, since overriding PP + recompute paths from Bridge would be fragile)

Also fixed the `wire_gemma4_kv_sharing` ordering bug you caught in `gpt_builders.py`.

yaoyu-33:
> for per_layer_inputs threading through TransformerBlock — this is generically named and model-agnostic at the MCore level. Moving it to Bridge would require overriding all of TransformerBlock.forward() including pipeline parallel and recompute paths, which is fragile. Would keeping this one generic hook in MCore be acceptable?

In past cases, we usually patch the change in bridge side. MCore has a more strict standard for signature change. Lemme check internally for more guidance.

yaoyu-33:
@DOGEUNNKIM  after some internal discussion, we feel better just to patch `per_layer_inputs` on bridge side as well, with ci guarding for now.

DOGEUNNKIM:
> @DOGEUNNKIM after some internal discussion, we feel better just to patch `per_layer_inputs` on bridge side as well, with ci guarding for now.

@yaoyu-33 Thanks for the direction. I've moved the Gemma4 PLE per_layer_inputs threading out of MCore and implemented it as a Bridge-side patch in NVIDIA-NeMo/Megatron-Bridge#4148.

#### What changed in MCore (#5090):

- Removed per_layer_inputs parameter from TransformerBlock.forward() and _checkpointed_forward()
- Removed the per-layer slicing and distribution logic from the recompute path
- GEGLU / tanh-GEGLU support remains as a generic MCore piece

#### What Bridge now handles (modeling_gemma4.py):

- _patch_ple_block_threading(decoder) is installed alongside the existing _install_ple_forward() hook
- The decoder instance's forward() intercepts per_layer_inputs, stores it locally, and passes clean kwargs to MCore
- Each transformer layer's forward() is patched to inject the per-layer slice ([:, :, layer_idx, :]) from the stored tensor
- _checkpointed_forward() is patched to carry per_layer_inputs through the recompute checkpoint args

All patches are instance-scoped (only the Gemma4 decoder is affected) and idempotent. CI guards are in place via the existing Bridge unit tests.

megatron/training/argument_utils.py
```diff
@@ -315,6 +315,18 @@ def core_transformer_config_from_args(args, config_class=None):
     if args.squared_relu:
         assert not args.swiglu
         kw_args['activation_func'] = squared_relu
+    elif getattr(args, 'geglu_tanh', False):
```
wujingyue:
```suggestion
    elif args.geglu_tanh:
```

megatron/training/argument_utils.py
```diff
@@ -315,6 +315,18 @@ def core_transformer_config_from_args(args, config_class=None):
...
+        assert not args.swiglu
+        assert not getattr(args, 'geglu', False)
+        assert not args.quick_geglu
```
wujingyue: These mutual-exclusiveness checks should go after the if-then-else. Otherwise, as is, we'd have to check in every branch, leading to O(N^2) checks.

E.g.
```
def _check_mutual_exclusive(args, arg_names: list[str]):
  # Check only one of them is set to true.
  ...
```

wujingyue (commented):
LGTM otherwise

megatron/training/argument_utils.py
DOGEUNNKIM: @wujingyue 

Thanks, that makes sense. I added a `_check_mutual_exclusive()` helper and centralized the activation flag validation for `swiglu`, `squared_relu`, `geglu_tanh`, `geglu`, and `quick_geglu`.

I also removed the duplicated branch-local `assert not ...` checks, so adding a new activation option should only require updating the helper call instead of adding pairwise checks in each branch.

megatron/training/argument_utils.py
DOGEUNNKIM: Thanks. I applied this and changed it to direct `args.geglu_tanh` access.

megatron/training/argument_utils.py
```diff
@@ -269,6 +269,14 @@ def _get_field_docstrings(self, src_cfg_class: type) -> dict[str, str]:
         return field_docstrings
 
 
+def _check_mutual_exclusive(args, arg_names: list[str]) -> None:
+    enabled_args = [arg_name for arg_name in arg_names if getattr(args, arg_name)]
+    assert len(enabled_args) <= 1, (
+        f"{', '.join('--' + arg_name.replace('_', '-') for arg_name in enabled_args)} "
```
wujingyue:
```suggestion
        f"{', '.join(enabled_args)} "
```
I understood your intention to show the args as they are in the command line, but the code complexity doesn't seem to worth it.

wujingyue (approved):

megatron/training/argument_utils.py
DOGEUNNKIM: @wujingyue 
Thanks for the suggestion! 
I agree, and I’ve applied the change.

wujingyue:
@yaoyu-33 does this look good to you? If so, I'll ping the owners for final review.

PR #5101, opened 2026-06-02
fix: preserve reload cadence for partial activation offload

lhb8125:
## What does this PR do?

Fixes fine-grained activation offloading when `activation_offload_fraction < 1.0`.

Partial activation offload keeps some offloadable groups on GPU. The reload queue still needs to preserve those groups as no-op slots so backward preloading keeps the same cadence as full offload. Without this, reload timing can drift and the next group may not be preloaded at the expected backward boundary.

This PR:
- Keeps all forward groups in the reload queue, including groups that were intentionally left on GPU.
- Consumes no-op reload slots during backward before preloading the next offloaded group.
- Tracks the forward `OffloadTensorGroup` through the autograd group-start marker so backward can identify the current no-op slot.
- Adds unit coverage for the no-op slot behavior and GPU coverage for `activation_offload_fraction=0.5` numerics and memory savings.

## Validation

- Passed on EOS, H100, PR head `091c2046b2d027281530885fe1bece7a1b093b31`:

```text
tests/unit_tests/pipeline_parallel/test_fine_grained_activation_offloading.py::test_gpt_fine_grained_activation_offload_fraction_correctness_and_memory

1 passed, 34 warnings in 45.35s
```

SLURM job: `5409716`
Log: `/lustre/fsw/coreai_devtech_all/hongbinl/agentic-mcore-dev/runtime/slurm/coreai_devtech_china-pr5101.fgao-fraction-260608_082124_5409716.log`


lhb8125:
/ok to test [efbb14b](https://github.com/NVIDIA/Megatron-LM/pull/5101/commits/efbb14b85e83f7485e554af8240bb60a6db31651)

lhb8125:
/ok to test [61deed1](https://github.com/NVIDIA/Megatron-LM/pull/5101/commits/61deed1297ca0218267df1101e4de3a3bb466e61)

PR #5103, opened 2026-06-02
enable tms no cpu backup region for param buffers

hanwen-sun:
# What does this PR do ?
Enable torch_memory_saver no cpu backup for both grad and param buffer, which can save 4 * param_size and 2 * param_size host memory each(fp32 and bf16). This pr refers to the implementation in miles: https://github.com/radixark/miles/pull/719  @yueming-yuan


HaochenYuan:
Good solution for relieving host memory pressure.  
The feature is currently threaded through hidden `args` attributes and extra DDP constructor kwargs, rather than through `DistributedDataParallelConfig` and normal argument validation. I suggest adding fields to DistributedDataParallelConfig & adding parser flags in `arguments.py` under the DDP/data-parallel argument group if users should enable this from CLI.
```
use_torch_memory_saver_for_grad_buffer: bool = False
use_torch_memory_saver_for_param_buffer: bool = False
torch_memory_saver_buffer_cpu_backup: bool = False
```

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1029,22 +1047,50 @@ def __init__(
...
             ):
-                self.shared_buffer = torch.zeros(
-                    self.numel,
-                    dtype=self.grad_dtype,
-                    device=torch.cuda.current_device(),
-                    requires_grad=False,
+                shared_mem_alloc_context = (
+                    param_mem_alloc_context
+                    if disable_param_buffers_cpu_backup
+                    else grad_mem_alloc_context
```
HaochenYuan: Current API exposes two independent flags but does not enforce that contract. Either collapse this to one DDP-buffer TMS flag, or validate that the param/grad choices match when the buffer is shared. A combined tag such as `shared_param_grad_buffer` would also make the ownership explicit.

FDecaYed (approved):

FDecaYed:
> Good solution for relieving host memory pressure. The feature is currently threaded through hidden `args` attributes and extra DDP constructor kwargs, rather than through `DistributedDataParallelConfig` and normal argument validation. I suggest adding fields to DistributedDataParallelConfig & adding parser flags in `arguments.py` under the DDP/data-parallel argument group if users should enable this from CLI.
> 
> ```
> use_torch_memory_saver_for_grad_buffer: bool = False
> use_torch_memory_saver_for_param_buffer: bool = False
> torch_memory_saver_buffer_cpu_backup: bool = False
> ```

I think this might be more confusing since what we really want is to disable, but the flag name you proposed feels like enable tms. I'm keeping the original flags in this dev PR for now, we can revisit on main PR

FDecaYed:
Addressed the DDP buffer CPU-backup feedback:

- Moved controls into `DistributedDataParallelConfig` instead of hidden DDP constructor kwargs / training-side plumbing.
- Kept the intended semantics: when `torch_memory_saver` is installed, the selected buffer allocations use `enable_cpu_backup=False`.
- Made the flags best-effort when TMS is not installed: they are ignored with a warning, since there is no TMS CPU backup to disable.
- Added tests for CLI/args-to-DDP-config mapping, MXFP8 shared-buffer mismatch validation, and missing-TMS no-op behavior.

FDecaYed (approved):

FDecaYed:
/ok to test [b06400c](https://github.com/NVIDIA/Megatron-LM/commit/b06400cbfcdbcecbfae40c1fd1cc581d873765a3)

Victarry (approved):

PR #5108, opened 2026-06-02
Add int4 qat training support

hanwen-sun:
# What does this PR do ?

Add int4 qat training support for rl training. This pr refers to https://github.com/THUDM/slime/pull/1362


FDecaYed (changes requested):
Let's add some tests so we can catch the issue when it breaks(for example TE internal changes may break this). with that said, I feel this is pretty self-contained and probably ok to merge as it is now.
When later merging this to main, we can take the opportunity to discuss higher level design. I think in the long term it might belong to TE as a feature and not just controlled through env var

hanwen-sun:
> Let's add some tests so we can catch the issue when it breaks(for example TE internal changes may break this). with that said, I feel this is pretty self-contained and probably ok to merge as it is now. When later merging this to main, we can take the opportunity to discuss higher level design. I think in the long term it might belong to TE as a feature and not just controlled through env var

Sounds great, Is there anything I can help?

HaochenYuan:
The fake quant itself is straightforward, but replacing TE's weight tensors with new STE tensors affects more than numerics: TE/Megatron fused wgrad paths use Python attributes on the original weight object (`main_grad`, `grad_added_to_main_grad`, FSDP `get_main_grad`, etc.). Please either add coverage for `gradient_accumulation_fusion` / delayed wgrad / FSDP, or explicitly guard unsupported combinations.

HaochenYuan:
For main branch, mayeb express int4 fake-QAT as a recipe in `fp8/fp4_quantizer_factory` — or ultimately as a TE recipe — rather than an env-var-gated override of _get_weight_tensors.

FDecaYed (approved):

FDecaYed:
Addressed the review feedback by keeping the `transformer_engine.py` change minimal and moving the int4 fake-QAT implementation into a dedicated helper module.

Changes:
- `transformer_engine.py` now only delegates from `TEGroupedLinear._get_weight_tensors()` to `maybe_fake_quantize_int4_weight_tensors()`.
- Added `transformer_engine_int4_fake_qat.py` for the env-var-gated int4 STE implementation.
- Added explicit guards for unsupported combinations where replacing TE weight tensors can break Python attributes used by wgrad/FSDP paths:
  - `gradient_accumulation_fusion`
  - delayed wgrad compute
  - Megatron FSDP-patched weights
- Added focused unit coverage for those guard paths.

FDecaYed (approved):

PR #5114, opened 2026-06-02
Weighted Checkpoint Merging & WSM Support

eric-tramel:
# Summary

This PR adds first-class support in Megatron-LM for **offline weighted checkpoint merging** — the technique used to evaluate and select NVIDIA's Nemotron 3 Super model during pretraining [1].

During the stable phase of a warmup-stable-decay (WSD) learning-rate schedule, individual trained checkpoints show noisy step-to-step benchmark performance, and getting a clean readout of model quality conventionally requires a dedicated learning-rate **decay run** for each evaluation point — expensive. **Weighted checkpoint merging** (also called warmup-stable-merge / WSM [2], or weight-space "model soup" averaging [3]) sidesteps this: instead of decaying, you take a weighted average of recent checkpoints over a sliding window and evaluate the *merged* weights. This produces a stronger, lower-variance quality signal at negligible cost, and the merged checkpoint can itself be the model you ship.

The payoff is large. For Nemotron 3 Super, offline merging during the stable phase delivered a consistent **+2–4 point** improvement across a 12-benchmark suite versus the corresponding trained checkpoint, while avoiding the dedicated decay runs — on the order of **~16% of the total pretraining FLOP budget** [1]. The final base model selected for downstream alignment was itself a 500B-token merge. This workflow needs a merge tool that is exact, cheap, and robust at production scale (the Nemotron 3 Super merges ran over 120B-parameter MoE hybrid Mamba-Transformer checkpoints); that is what this PR provides.

`tools/checkpoint/weighted_merge.py` selects a set of compatible `torch_dist` checkpoints (manually, or by iteration window with deterministic WSM coefficient schedules such as the **minus-sqrt** emulation used in [1, 2]), averages their model tensors with those coefficients, and writes a new `torch_dist` checkpoint that loads directly for evaluation or resume.

The merge is derived **entirely from each source checkpoint's public PyTorch DCP metadata** — tensor FQNs, global shapes, dtypes, chunk/sharding layout, and byte/object `_extra_state`. It **never constructs a model**, so it imports none of the model code (and none of the CUDA-only kernels that Mamba / hybrid / Transformer-Engine model definitions pull in). Consequently it runs **CPU-only for every model family** on a GPU-less node, never gathers a full model on rank 0, and merges 120B-scale checkpoints in minutes on a single CPU node.

**Zero `megatron/core` changes.**

### References

1. NVIDIA. *Nemotron 3 Super: Open, Efficient Mixture-of-Experts Hybrid Mamba-Transformer Model for Agentic Reasoning.* 2025. https://research.nvidia.com/labs/nemotron/files/NVIDIA-Nemotron-3-Super-Technical-Report.pdf (see §2.5, "Tracking Merge Evaluation").
2. C. Tian, J. Wang, Q. Zhao, et al. *WSM: Decay-free learning rate schedule via checkpoint merging for LLM pre-training.* 2025. arXiv:2507.17634.
3. M. Wortsman, G. Ilharco, S. Y. Gadre, et al. *Model soups: averaging weights of multiple fine-tuned models improves accuracy without increasing inference time.* 2022. arXiv:2203.05482.

## Why metadata-driven / CPU-only

Weighted checkpoint merging is **I/O-bound**: the work is reading each input once, accumulating in fp32, and writing one output. There is no model forward pass, so GPU compute buys nothing. The *only* thing that would force a GPU is building a model to obtain the tensor inventory (Mamba/TE model code imports CUDA kernels at definition time) — so this tool reads that inventory from the checkpoint metadata instead and stays on the CPU.

## Placement

This lives in `tools/checkpoint/` because it is a standalone, post-hoc operation over already-saved checkpoints — run as a CLI (or imported on demand), never invoked from the training loop — depending only on public DCP / `dist_checkpointing` metadata APIs. It is intentionally *not* under `megatron/training/`: importing any `megatron.training.*` module first runs `megatron/training/__init__.py`, which eagerly imports the full training stack (`pretrain`, `get_model`, `initialize_megatron`, tokenizers, …). Since the merge constructs no model and needs none of that, coupling its import to the training package would contradict the metadata-only design and add a large, fragile import surface for no benefit.

### Command line

```bash
# Manual weighted merge of two checkpoints
python tools/checkpoint/weighted_merge.py \
  --merge-inputs /ckpts/iter_0000840:0.25 /ckpts/iter_0000880:0.75 \
  --merge-output /ckpts/merged --output-iteration 880000 --merge-save-dtype same

# Iteration-window WSM merge (minus-sqrt schedule)
python tools/checkpoint/weighted_merge.py \
  --merge-inputs /ckpts --start-checkpoint 840000 --end-checkpoint 880000 \
  --merge-style minus-sqrt --merge-output /ckpts/merged --output-iteration 880000

# Parallelize reads across ranks on one node (torchrun)
torchrun --standalone --nproc-per-node 8 tools/checkpoint/weighted_merge.py \
  --merge-inputs /ckpts --start-checkpoint 840000 --end-checkpoint 880000 \
  --merge-style minus-sqrt --merge-output /ckpts/merged --output-iteration 880000
```

### Slurm

The merge is CPU-only and I/O-bound, so allocate CPU nodes and scale the number of tasks (one task per reader rank) to parallelize reads across the source checkpoints. Each rank only needs the standard torch-distributed env vars that `ensure_process_group` reads, so map the `SLURM_*` variables to them — no torchrun required. The example sbatch below merges an iteration window across 4 nodes (32 reader ranks):

```bash
#!/bin/bash
#SBATCH --job-name=wsm-merge
#SBATCH --partition=cpu            # CPU-only; no GPUs requested
#SBATCH --nodes=4
#SBATCH --ntasks-per-node=8        # reader ranks per node
#SBATCH --cpus-per-task=12
#SBATCH --time=01:00:00
#SBATCH --output=wsm-merge-%j.out

# Rendezvous address (set once; inherited by every task)
export MASTER_ADDR=$(scontrol show hostnames "$SLURM_JOB_NODELIST" | head -n1)
export MASTER_PORT=29500

# One task per rank; map SLURM_* to the torch-distributed env vars
srun bash -lc '
  export RANK=$SLURM_PROCID WORLD_SIZE=$SLURM_NTASKS LOCAL_RANK=$SLURM_LOCALID
  python tools/checkpoint/weighted_merge.py \
    --merge-inputs /ckpts --start-checkpoint 840000 --end-checkpoint 880000 \
    --merge-style minus-sqrt --merge-output /ckpts/merged --output-iteration 880000'
```

Scale `--nodes` / `--ntasks-per-node` to trade wall-clock for nodes (the merge saturates a single node's storage-read bandwidth, so spreading ranks across nodes is what pushes throughput higher). For a single-node merge set `--nodes=1`; for a one-rank merge drop `srun` and run `python ...` directly. (Alternatively, launch one `torchrun` agent per node under `srun --ntasks-per-node 1` with `--rdzv-backend c10d`; torchrun then sets the env vars itself.)

### Library

The merge is also callable directly. `merge_same_layout_dcp_metadata_checkpoints` initializes a gloo process group if one is not already active, so it works in a plain single-process script; launch the script under `torchrun` (single node) or Slurm `srun` (across nodes) to parallelize reads across ranks.

```python
from tools.checkpoint.weighted_merge import (
    merge_same_layout_dcp_metadata_checkpoints,
    checkpoint_coefficients,
)

# Manual weights
result = merge_same_layout_dcp_metadata_checkpoints(
    ["/ckpts/iter_0000840", "/ckpts/iter_0000880"],
    [0.25, 0.75],
    "/ckpts/merged",
    output_iteration=880,
    save_dtype="same",
)
print(result.output_dir, result.averaged_tensors, result.timings.total)

# WSM iteration window with minus-sqrt coefficients
iters = [840, 860, 880]
ckpts = [f"/ckpts/iter_{it:07d}" for it in iters]
weights = list(checkpoint_coefficients(iters, "minus-sqrt").values())
result = merge_same_layout_dcp_metadata_checkpoints(
    ckpts, weights, "/ckpts/merged",
    output_iteration=880, save_dtype="same", merge_style="minus-sqrt",
)
```

## Behavior

- Manual `PATH:WEIGHT` inputs (including unnormalized and negative weights, plus optional `--normalize`).
- Range/iteration selection (`--start-checkpoint` / `--end-checkpoint`) with deterministic WSM coefficient schedules (`linear`, `minus-sqrt`) and the `__reverse` / `__scramble` modifiers; `--min-iteration-interval`, `--min-checkpoints`, `--coefficient-seed`.
- fp32 CPU accumulation; `same`, `float32`, `float16`, `bfloat16` save-dtype policy.
- Transformer Engine `_extra_state` copied (not averaged) from a single selected source (`--extra-state-source-index`), including byte/object entries.
- Correct `metadata.json` + `latest_checkpointed_iteration.txt` so the output loads in Megatron with `--no-load-optim --no-load-rng`.
- Fail-closed validation: inputs must share an identical layout (automatically true for WSM — averaging checkpoints of the *same* model at different iterations); mismatched key sets, shapes, chunk layouts, dtypes (under `same`), non-floating tensors, unsupported formats, missing metadata, and byte-objects outside recognized model roots are all rejected.
- Output written via public PyTorch DCP and published atomically (write to a hidden temp dir, fail-closed if the target exists, rename, then the latest marker; best-effort fsync of sidecars and parents).

## Execution

Pure-Python: no Megatron model bootstrap, no model-parallel state, no model construction — a gloo process group only. Parallelism is the launched world size; more ranks (launched via `torchrun` on one node or Slurm `srun` across nodes) give more parallel reads, saturating at storage bandwidth. A single process is the simplest invocation; multi-rank reshards via DCP.

## Correctness evidence

- Unit suite (CPU): coefficient math and modifiers, fp32 accumulation, dtype policy, `_extra_state` (incl. byte/object) copy and source selection, same-layout fail-closed validation (key / shape / layout / dtype / format / missing-metadata and out-of-root byte-objects), iteration/window selection, atomic publication (reject-existing), and unprefixed-GPT / MoE / MTP roots.
- End-to-end CPU merge of generated `torch_dist` checkpoints on a GPU-less node, with `_extra_state` copied and a loadable output checkpoint.

### Practical rollout demonstration

A PDX rollout reproduction also exercises the PR output on a real Nano stable-phase checkpoint window, not just synthetic/unit fixtures:

- Built a **250B-token WSM merge** from Nano stable-phase checkpoints at `iter_0610000` using this PR's CPU-only metadata-driven merge path.
- Evaluated the stable checkpoint and the merged checkpoint with the Nano eval harness on ARC Challenge (`arc_challenge_llama`). Slurm job `5649497` completed successfully.
- Stable checkpoint: `acc_norm = 0.8694539249146758`.
- PR-merged checkpoint: `acc_norm = 0.9027303754266212`.
- Absolute improvement: **+0.0332764505119454** (**+3.33 percentage points**).
- Historical Nano 250B-window reference for the same ARC Challenge slice was `87.201 -> 89.676` (**+2.475 pp**), so this rollout reproduces the expected improvement direction and lands within the practical variance envelope, with a slightly stronger observed gain.

This rollout used GPUs only for downstream evaluation. The checkpoint merge itself remained CPU-only. The downstream eval required local PDX server/eval-harness compatibility wiring for the Nano hybrid model stack; the merged checkpoint under evaluation was produced by the PR's `weighted_merge.py` implementation.

## Performance

The merge approaches the I/O-volume lower bound (read each input once, write one output). Walltime is dominated by checkpoint load (storage bandwidth); fp32 accumulation is a small fraction. Throughput scales with parallel ranks until storage bandwidth saturates; GPUs provide no speedup (at equal world size, CPU and GPU runs are within noise).

### Benchmark

Swept merge wall-clock over **checkpoint count** and **parallel readers (ranks)** for four representative models at production scale — ~30B and ~120B total parameters, in dense-GPT and Mamba-MoE (Nemotron-H-style) topologies — on a single CPU node, merging in bf16:

| model | params | GiB/ckpt | tensors | merge 8 ckpts @ 1 rank | @ 8 ranks | speedup |
|---|---|---|---|---|---|---|
| GPT, ~30B | 29.6B | 55 | 291 | 658 s | 118 s | 5.6× |
| GPT, ~120B | 121.8B | 227 | 435 | 2574 s | 572 s | 4.5× |
| Mamba-MoE, ~30B | 29.2B | 54 | 1,317 | 799 s | 143 s | 5.6× |
| Mamba-MoE, ~120B | 114.2B | 213 | 3,291 | 3175 s | 589 s | 5.4× |

- **Linear in checkpoint count** — each additional checkpoint is one more read pass (e.g. the 120B GPT at 8 ranks: 185 → 272 → 572 s for 2 → 4 → 8 checkpoints).
- **Near-linear rank speedup, then saturating** — ~4.5–5.6× at 8 parallel readers, plateauing at a single node's read ceiling (~3.3–4.9 GiB/s). Reads distribute across ranks by chunk, so more ranks = more parallel reads.
- **Cost tracks bytes, not architecture** — the 120B models cost ~4× the 30B at equal settings (the byte ratio), and dense-GPT vs Mamba-MoE land within ~25% of each other at equal bytes despite the MoE models carrying ~10× the tensor count. Total bytes dominate; per-tensor overhead is small.

## Limitations / non-claims

- Same-layout inputs only (the WSM case). Merging checkpoints of *different* architectures or sharding is out of scope.
- No model-template verify-load step (no model is built); verification of merged values is done separately.
- Token-window selection (`--merge-window-btoks`) is unsupported (no model ⇒ no `seq_length` / `global_batch_size` to derive a token window from).
- Atomic publication covers temp-dir + rename; it does not claim crash atomicity across a kill during the final rename or a filesystem power-loss event.


dimapihtar:
Hi @eric-tramel, thanks a lot for the contribution.

Could you fix code style, please? Thanks.

eric-tramel:
Thanks @dimapihtar 🙏  Missed the black pass. Should be passing on lint/format now. 

Should I try to republish this as a branch instead of a fork if it would be easier for CI?

dimapihtar:
> Thanks @dimapihtar 🙏 Missed the black pass. Should be passing on lint/format now.
> 
> Should I try to republish this as a branch instead of a fork if it would be easier for CI?

Thanks. No, fork is fine.

eric-tramel:
Updated to try to address world-size handling in tests (guard single v multi-rank tests appropriately).

eric-tramel:
Format, greptile etc was all met. Not sure why this PR isn't playing well with CI, but would be great to get some feedback on if this can come into the library or not.

Connor-XY:
/ok to test [6960fae](https://github.com/NVIDIA/Megatron-LM/pull/5114/commits/6960faecfc5a3058d6ec8da2691d3863ed9a63c0)

tools/checkpoint/weighted_merge.py
```diff
@@ -0,0 +1,2006 @@
...
+            "Rank-work balanced plan: "
+            f"tensor_bytes_by_rank={list(work_plan.tensor_bytes_by_rank)}, "
+            f"tensor_chunks_by_rank={list(work_plan.tensor_chunks_by_rank)}",
+            flush=True,
+        )
+
+    temporary_output_dir = _prepare_temporary_output_dir(output_dir)
+
+    base_common_state = dist_checkpointing.load_common_state_dict(str(resolved_input_dirs[0]))
+    common_state = _prepare_common_state(base_common_state, output_iteration)
```
Connor-XY: The output iteration is updated here, but the sample counters still come from the first checkpoint. In range mode, we end up using the last iteration in the window with the data position from the first one.

For example, merging iterations 1000–2000 with a global batch size of 128 leaves `consumed_train_samples` at 128,000 instead of 256,000. Training restores that old counter even with `--no-load-optim --no-load-rng`, so resuming repeats data from the merge window.

Could we take the progress counters from the checkpoint matching the output iteration? If the requested iteration doesn't match an input checkpoint, we should define how to handle that case too.

tools/checkpoint/weighted_merge.py
```diff
@@ -0,0 +1,2006 @@
...
+            resolved_input_dirs[1:], tensor_metadata_by_checkpoint[1:]
+        ):
+            actual = _metadata_same_layout_tensor_layout(fqn, tensor_metadata[fqn])
+            if actual.global_shape != expected.global_shape:
+                raise WeightedMergeError(
+                    f"Shape mismatch for '{fqn}' in metadata-same-layout: "
+                    f"expected {expected.global_shape}, got {actual.global_shape} "
+                    f"in {checkpoint_dir}."
+                )
+            if require_matching_chunks and actual.chunks != expected.chunks:
```
Connor-XY, line 1420: This comparison checks the order of the chunks as well as their offsets and sizes. Two checkpoints can have the same layout but list the chunks in a different order.

The new balancing option can produce exactly that case: with chunks at offsets `(0, 0)` and `(2, 0)`, a single-rank balanced plan puts the second chunk first. Trying to merge that output with its source then raises `Chunk layout mismatch`, even when the first merge used only that source with weight 1.

Could we sort the `(offsets, sizes)` pairs before comparing them? A test with multiple chunks for one tensor would cover this.

tools/checkpoint/weighted_merge.py
```diff
@@ -0,0 +1,2006 @@
...
+            self.resolved_input_dirs[self.extra_state_source_index],
+            [(spec.path, spec.template_leaf)],
+            self.load_strategies[self.resolved_input_dirs[self.extra_state_source_index]],
+        )
+        self.load_time += time.perf_counter() - load_start
+        value = loaded[spec.path]
+        if isinstance(value, io.BytesIO):
+            value = io.BytesIO(value.getvalue())
+        elif isinstance(value, (bytes, bytearray)):
+            value = io.BytesIO(value)
```
Connor-XY: This turns a `bytes` or `bytearray` extra-state value into a `BytesIO` object. The contents survive, but loading the merged checkpoint returns a different type. That can break a model's `set_extra_state` method if it expects raw bytes, for example when calling `pickle.loads`.

Could we keep the original value type here? `torch.save([value], serialized)` already handles serializing it, so the extra `BytesIO` wrapper shouldn't be needed for these values.

Connor-XY (dismissed):

tools/checkpoint/weighted_merge.py
kevjshih: I agree, we should avoid repeating data in the case of a resume.

kevjshih:
Thanks for the fixes here. This PR also needs a few additional changes to handle recent MiMo and VLM checkpoints:

- **Explicit model-prefix selection.** Our MiMo/VLM checkpoints store model weights under `language_model.` and `modality_submodules.`. The current implementation only recognizes `model.`, so it finds no mergeable model tensors. We need a repeatable `--merge-model-prefix` option that requires every specified prefix to exist in every input checkpoint.

- **Opt-in handling of non-model state.** These checkpoints contain optimizer, RNG, and other common state alongside the model weights. WSM should merge only the selected model tensors while copying the appropriate common state. We need an explicit `--merge-ignore-non-model-state` option for this case, while keeping the default fail-closed so unexpected state is not silently discarded.

- **Dry-run mode.** Before writing a large merged checkpoint, we need to verify that all inputs have compatible keys, shapes, dtypes, and distributed work planning. A `--dry-run` option would run those checks without creating an output.

- **Merge provenance.** The merged checkpoint should record the requested model prefixes and whether non-model state was ignored, so it is clear exactly what was included in the merge.

I'll add these changes as focused commits with tests.

kevjshih:
/ok to test [957b0e0](https://github.com/NVIDIA/Megatron-LM/pull/5114/commits/957b0e01152a0a06ec979e6f7050d1c039b04cd7)

kevjshih:
/ok to test [e2503d4](https://github.com/NVIDIA/Megatron-LM/pull/5114/commits/e2503d4e0591039ca654366a9af4b4bc8b9723a4)

kevjshih:
/ok to test [9bad76b](https://github.com/NVIDIA/Megatron-LM/pull/5114/commits/9bad76bc1220abeeef312503f27adec3833f88c8)

tools/checkpoint/weighted_merge.py
```diff
@@ -0,0 +1,2150 @@
...
+        iterations.append(common_state["iteration"])
+    checkpoint_args = common_state.get("args")
+    if checkpoint_args is not None and hasattr(checkpoint_args, "iteration"):
+        iterations.append(checkpoint_args.iteration)
+
+    if not iterations:
+        return None
+    if any(not isinstance(iteration, int) for iteration in iterations):
+        raise WeightedMergeError("Checkpoint common state contains a non-integer iteration.")
+    if len(set(iterations)) != 1:
```
Connor-XY: These two iteration fields can differ in a normal training checkpoint. `args.iteration` stays at the step where training started or resumed, while the training loop advances a separate counter that gets saved as the top-level `iteration`.

For example, a valid checkpoint can have `iteration=2000` and `args.iteration=0`. This check rejects it, breaking both range merges and `--common-state-checkpoint`. I reproduced this using `generate_state_dict`.

Could we use the top-level `iteration` when present, fall back to `args.iteration` only when it’s missing, and add a test covering this case?

kevjshih:
/ok to test [fef31ad](https://github.com/NVIDIA/Megatron-LM/pull/5114/commits/fef31adef770f6ee124001623b8bf79260bfd0bf)

kevjshih:
/ok to test [7a1fc9e](https://github.com/NVIDIA/Megatron-LM/pull/5114/commits/7a1fc9e5042a690eb0fa99c6a2a4b4926645009e)

kevjshih:
/ok to test [ca4826e](https://github.com/NVIDIA/Megatron-LM/pull/5114/commits/ca4826e9555032b8710dfb4a4e5693c46c1613dd)

Connor-XY (approved):

sanyalsunny111:
This technique is very similar to the approach proposed in our paper (https://arxiv.org/abs/2306.03241) back in 2023, except that it uses weighted tail averaging. It might be worth acknowledging this connection.

PR #5118, opened 2026-06-02
ci: make CI resilient to pip/uv network timeouts

ko3n1g:

## Background / motivation

Several CI jobs fail on transient network timeouts during dependency / image fetches, with no retry cushion:

- **Host `setup-python` pip upgrade** — on a fresh toolcache, `actions/setup-python@v5` fetches the latest pip (26.1.2) from PyPI; the download hit a read timeout and aborted the whole job *before any test ran*:
  ```
  TimeoutError: The read operation timed out
  Error: The process '/usr/bin/bash' failed with exit code 2
  ```
- **manylinux wheel build** — the `megatron-fsdp` build's `docker run` implicitly pulled `quay.io/pypa/manylinux_2_28_x86_64` and the registry request timed out:
  ```
  docker: Error response from daemon: Get "https://quay.io/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
  ##[error]Process completed with exit code 125.
  ```

pip's default per-read socket timeout is ~15s with no retries; the docker pull had no retry at all. The same flakiness hits in-container `uv sync` / `pip`.

## What changed

- **`cicd-main.yml`** — add `PIP_DEFAULT_TIMEOUT=120` + `PIP_RETRIES=5` to the three job-level `env:` blocks, inherited by the host `setup-python` pip upgrade so it retries instead of hard-failing.
- **`action.yml`** — export `PIP_DEFAULT_TIMEOUT` / `PIP_RETRIES` / `UV_HTTP_TIMEOUT=120` in both generated run-scripts so in-container `uv sync` / `pip` get the same cushion.
- **`_build_test_publish_wheel.yml`** — pre-pull the manylinux image with the repo's standard 3× retry loop before `docker run`, so a transient quay.io timeout no longer fails the job (later runs in the job reuse the cached image).
- **`launch_nemo_run_workload.py`** — classify read/connection timeouts (`The read operation timed out`, `Read timed out`, `TimeoutError`, `Connection broken`, `Temporary failure in name resolution`) as flaky so the launcher auto-restarts; and collect harness/task logs (`**/*.log`), not just per-rank `attempt_0/*/std*.log`, so those signatures are actually visible to `is_flaky_failure`.

## Details

- Host pip env vars sit beside the existing `PIP_DISABLE_PIP_VERSION_CHECK` / `PIP_NO_PYTHON_VERSION_WARNING` / `PIP_ROOT_USER_ACTION` block — three identical job-level `env:` maps (latest, h100, gb200).
- The docker pre-pull mirrors the existing `apt-get` / `uv install` retry loops in `action.yml` (`for i in 1 2 3; do … && break; sleep 10; done`).
- `_collect_failure_logs()` dedups by resolved path and is `OSError`-tolerant; it widens the glob so the golden-value assertion (written to the nemo-run task log, outside the per-rank tree) is no longer missed.

## Tested

- `ruff check`, `isort --check`, `black==24.4.2 --check` — all pass on the changed Python file.
- YAML parse of all three workflow files — valid.
- `py_compile` — OK.


balasaajay (approved):
looks good.

just a heads up: there was a requirement earlier today to surface the actual error in a different step so its easier to jump to.

PR #5129, opened 2026-06-03
NCCL EP support

### LINKED ISSUE #4644: [QUESTION] Do we have any plan to introduce NCCL EP?
Hello,

I don't find any discussion about introducing of the new NCCL feature, NCCL EP.  Do we have any plan about this? Is the issue https://github.com/NVIDIA/nccl/issues/2104 the main concern?

Thanks

YangFei1990:

Fixes https://github.com/NVIDIA/Megatron-LM/issues/4644

# What does this PR do ?
This PR adds support for NCCL EP in megatron core. It is building on top of TE PR https://github.com/NVIDIA/TransformerEngine/pull/3035. 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


kvareddy:
@jiemingz  @fanshiqing can you please take a look at this MR?

megatron/core/transformer/transformer_config.py
```diff
@@ -865,6 +866,25 @@ class TransformerConfig(ModelParallelConfig):
     moe_hybridep_num_sms_preprocessing: int = 108
     """Number of SMs to use for HybridEP preprocessing (metadata scan kernel)."""
 
+    moe_ncclep_num_sms: int = 0
```
jiemingz: The corresponding arg for the other backends are: 
```
moe_deepep_num_sms
moe_hybridep_num_sms
```
Rather than adding yet another and bloating the config could we please just have a single
`moe_flex_dispatcher_num_sms`?

The existing configs can stay but are put behind a deprecation flag, and just get routed to `moe_flex_dispatcher_num_sms` in the post_init.

megatron/core/transformer/transformer_config.py
```diff
@@ -865,6 +866,25 @@ class TransformerConfig(ModelParallelConfig):
...
+    """Number of SMs to use for NCCL EP (the max_num_sms passed to TransformerEngine's
+    ep_bootstrap). 0 lets TransformerEngine/NCCL choose the default."""
+
+    moe_ncclep_static_shape: bool = False
```
jiemingz, line 925: Was this a specific request? Previously we haven't supported this mode since allocating a fully sized receive buffer allocated too much memory to be useful.

For cudagraphing the MoE layer the strategy is currently to use paged stashing: https://github.com/NVIDIA/Megatron-LM/pull/2690

yaox12 (approved):

megatron/core/transformer/transformer_config.py
YangFei1990: This is specific for NCCL EP. Unlike Hybrid EP which supports dynamic shapes, NCCL EP always requires static shapes for dispatch and combine. The static shape size is managed by `alignment` and  `moe_expert_rank_capacity_factor`. As a result, the output of dispatch will contain padded tokens, the cuteDSL groupgemm kernel can handle this padding properly, but it is only available on blackwell. I think the other code paths requires the tokens to be all valid. `moe_ncclep_static_shape=False` provide a feasible way for this code path, where we remove the padded tokens so the groupgemm/mlp will only process the valid tokens. And in this way the shape is not static.

For full CG, we still use paged stashing with static allocated buffer, nothing changed there.

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -600,3 +600,282 @@ def hybrid_ep_combine(x, handle, num_permuted_tokens, pad_multiple, fused=False)
...
+    before the first dispatch. Reuses TransformerEngine's own one-time flag, so repeated
+    calls (e.g. once per MoE layer) are no-ops.
+
+    Args:
+        ep_group (torch.distributed.ProcessGroup): The expert-parallel process group.
+        num_experts (int): Total experts across ``ep_group`` (global, not per-rank).
+        max_tokens_per_rank (int): Upper bound on local input tokens per forward. Must be
+            even (NCCL EP requires ``num_tokens_per_rank * inner_dim % 4 == 0``).
+        recv_capacity_per_rank (int): Per-rank receive-buffer capacity in tokens. Must be
+            ``>= max_tokens_per_rank``; runtime overflow hard-traps (no soft drop).
```
nanz-nv, line 635: Can this be improved? Hard failure when overflow happens is not great.

megatron/core/transformer/transformer_config.py
```diff
@@ -1525,15 +1556,18 @@ def __post_init__(self):
                 )
 
         if self.moe_expert_rank_capacity_factor is not None:
```
nanz-nv, line 1656: This looks like one of the major differences between NCCL-EP and HybridEP. In eager mode, HybridEP could allow a sync and on-demand sizing of the buffer. Do we have a plan to support that? This plus the hard failure when the token exceeds the value might make this feature much less useful in practice.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1392,6 +1400,271 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
...
+            return NcclEpContext(
+                top_k=self.router_topk,
+                max_tokens_per_rank=max_tokens_per_rank,
+                recv_capacity_per_rank=self._recv_capacity,
+                hidden_dim=self.hidden_dim,
+                num_local_experts=self.num_local_experts,
+                alignment=self.alignment,
+            )
+
+        self._pool = NcclEpContextPool(
```
Autumn1998: I don’t understand why we need a pool of buffers. In my mental model, the buffer should be shared by all communications. Even if we have many microbatches, there shouldn’t be two EP communications happening at the same time.

If the buffer has an internal buffer whose size is controlled by num_tokens, then we should actually share a single global buffer, because every additional buffer consumes extra GPU memory.

If the buffer doesn’t have such an internal buffer and is just a set of functions, then there’s even less reason to create multiple buffers.

Do we have a strong reason why we must use a pool to manage multiple buffers?

megatron/core/transformer/moe/fused_a2a.py
phu0ngng: This will be improved. It requires changes in both NCCL EP and TE, which we will work on after the first integration is merged.

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -600,3 +600,282 @@ def hybrid_ep_combine(x, handle, num_permuted_tokens, pad_multiple, fused=False)
...
+
+
+def nccl_ep_finalize():
+    """Tear down the NCCL EP context. Idempotent; safe when never bootstrapped.
+
+    Releases the borrowed NCCL communicator and must run before the process group is
+    destroyed.
+    """
+    if HAVE_TE_EP:
+        te_ep.ep_finalize()
```
phu0ngng, line 664: I think you don't need to call this `ep_finalize()` since we already registered this `atexit`.

The API is there for two special use cases:
1. For users who want to re-bootstrap. Let me know if you intend to re-bootstrap in any cases in MCore.
2. When users call `dist.destroy_process_group` explicitly.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1392,6 +1400,271 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
...
+        self.token_indices: Optional[torch.Tensor] = None
+        self.dispatched_probs: Optional[torch.Tensor] = None
+        self.tokens_per_expert: Optional[torch.Tensor] = None
+        self.num_local_tokens: Optional[int] = None
+
+    def setup_metadata(self, routing_map: torch.Tensor, probs: torch.Tensor):
+        num_tokens = routing_map.shape[0]
+        probs = probs.reshape(num_tokens, self.num_experts)
+        # Convert the multihot routing map to (topk weights, topk indices), like DeepEP.
+        self.token_probs, self.token_indices = torch.topk(probs, self.router_topk, dim=-1)
```
phu0ngng, line 1514: Note: to remove after NCCL EP adds support for routing map topk + dense prob (i.e. in the shape of `[num_tokens, num_expert]` instead of `[num_tokens, num_topk]`).

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1392,6 +1400,271 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
...
+        self.num_local_tokens = num_tokens
+
+    def _ensure_pool(self):
+        """Bootstrap NCCL EP and build the per-layer context pool on first use (static shapes)."""
+        if self._pool is not None:
+            return
+        # NCCL EP's HT backend requires max_dispatch_tokens_per_rank to be a multiple of the HT
+        # chunk size (64); ncclEpCreateGroup otherwise fails with "invalid usage".
```
phu0ngng, line 1522: Did we have an nvbug for tracking this?

megatron/core/transformer/moe/token_dispatcher.py
phu0ngng: Right now, for non-zc mode, the TE `EpBuffer` only owns `handle_mem` which carries the routing maps, and `token_counts` which is not exactly the same as the `HybridEpBuffer`. 
We can move these tensors out of `EpBuffer` and turn this into a lightweight EpConfig, and let the `ep_prepare` allocate and return those tensors, then we call `ep_dispatch/combine` with this `EpConfig` and `handle_mem` as input args. 

However, for the zc copy mode, we may need to allocate some symmem tensors (if users pass the dispatch recv_token, we still allocate the dispatch recv_topk_weight), thus we need to have a module to be able to allocate those symmem tensors in the `__init__`. So I think keeping the `EpBuffer` may allow us to do more customization later.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1392,6 +1400,271 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
...
+    ) -> torch.Tensor:
+        # Note: this needs to stay out of the torch.compile region because TE's ep_bootstrap does
+        # opaque ProcessGroup._get_backend()._comm_ptr() access that dynamo cannot trace.
+        self._ensure_pool()
+        lease = self._pool.lease()
+        self._active_lease = lease
+        # TE requires int64 indices and float32 weights.
+        # token_indices/token_probs: [num_local_tokens, router_topk]
+        topk_idx = self.token_indices.to(torch.int64)
+        topk_weights = self.token_probs.float()
```
phu0ngng: Does it mean you actually have topk_idx in int32 and weight in FP64? 

We have pending work on supporting topk_idx in int32. cc @artpol84. 

For now, let me make TE accept `topk_idx` in both I32 and I64 and have a conversion WAR in TE until NCCL EP supports it. Same for the `topk_weights`.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1392,6 +1400,271 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
...
+        #   dispatched_probs: [recv_capacity_per_rank]
+        if torch.is_grad_enabled() and hidden_states.requires_grad:
+            hidden_states = _NcclEpLeaseRelease.apply(hidden_states, self._pool, lease)
+        else:
+            # we are inside of activation checkpointing, do not need to save for backward
+            lease.release_at_combine = True
+        recv_tokens, tokens_per_expert, dispatched_probs = nccl_ep_dispatch(
+            lease.ctx, hidden_states, topk_idx, topk_weights
+        )
+        self.tokens_per_expert = tokens_per_expert.to(torch.int64)
```
phu0ngng, line 1578: Why do you have `topk_idx` in int32 but `tokens_per_expert` in int64? 

I can make TE return `tokens_per_expert` in int64, but I need a way to know users want it to be in int64, and infer from `topk_idx` seems to be the cleanest way.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1392,6 +1400,271 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
...
+        self.dispatched_probs = dispatched_probs
+        return recv_tokens
+
+    def get_permuted_hidden_states_by_experts(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        if self.static_shape:
+            return hidden_states, self.dispatched_probs
+        # narrow to the sum(counts) valid (alignment-padded) rows the experts consume.
+        num_valid = int(self.tokens_per_expert.sum().item())  # sum(counts) = Σ
```
phu0ngng, line 1586: This will trigger a D2H, right? 
Is this function only used in Eager mode? Would it be useful if I make `ep_dispatch` return `num_recv_tokens` tensor directly?

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1392,6 +1400,271 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
...
+        permuted_hidden = hidden_states[:num_valid]  # [recv_capacity_per_rank, H] -> [Σ, H]
+        permuted_probs = self.dispatched_probs[:num_valid]  # [recv_capacity_per_rank] -> [Σ]
+        return permuted_hidden, permuted_probs
```
phu0ngng, line 1589: But why do you need to strip out the over-allocated tail? Don't the GroupedGEMM and other ops read the `tokens_per_expert` and not read the overallocated-tails anyway?

megatron/core/transformer/moe/token_dispatcher.py
artpol84: We do now

megatron/core/transformer/moe/token_dispatcher.py
phu0ngng: A follow-up question, if the topk_weight is in FP64, does it mean we should dispatch it in FP64 and do the accumulation in Combine in FP64 as well?
Note that this may make things perhaps slightly slower (!?), mainly because of the accumulation in FP64.

megatron/core/transformer/transformer_config.py
phu0ngng: Adding support for on-demand sizing of the buffer (which involves a D2H) is on my plan after the first integration is done.

megatron/core/transformer/moe/token_dispatcher.py
YangFei1990: @phu0ngng for zero-cp we can let user manage all the buffers, recv/topk_weights etc, I think that should be okay. But my understanding is that the routing maps stored in the `handle_mem` is per-layer/per-mb specific, which will be reused during backward, so that is why we need this pool. Is that the right understanding?

megatron/core/transformer/moe/fused_a2a.py
YangFei1990: Mcore will call `dist.destroy_process_group` explicitly, you can check changes in `megatron/core/parallel_state.py`. I remember we need to call this before destroy the pg otherwise there are some errors.

megatron/core/transformer/moe/token_dispatcher.py
YangFei1990: From framework perspective, `token_probs` should be either bf16 or fp32 (controlled by `--moe-router-dtype=fp32`), `topk_idx` will be int64

megatron/core/transformer/moe/token_dispatcher.py
YangFei1990: Can you elaborate? I do not see an int32 tensor. As mentioned above probs are fp32 and indices are int64

megatron/core/transformer/moe/token_dispatcher.py
YangFei1990: Yes this is eager mode, the graph mode goes with `static_shape=True`. I think we need to consider the full path to support dynamic, having this narrow inside TE/mcore does not make too much difference.

megatron/core/transformer/moe/token_dispatcher.py
YangFei1990: Only the cuteDSL groupgemm can support that, the rest groupgemm implementation expects only valid tokens. This path is to make sure these groupgemms won't break

YangFei1990:
/ok to test [cadf32c](https://github.com/NVIDIA/Megatron-LM/pull/5129/commits/cadf32cbb7f755a8c2b08d8947d694f68144bc07)

megatron/core/transformer/moe/token_dispatcher.py
Autumn1998: I think the key point here is that although we do have per-layer / per-microbatch data, their lifetime is strictly bounded by the forward–backward pair, and they can be released immediately after the corresponding backward finishes.

In previous HEP/DEP implementations, we handled this with simple handles rather than a buffer pool. The lifecycle looks like:

fwd layer 0 → create and store handle for layer 0
fwd layer 1 → create and store handle for layer 1
...
bwd layer 1 → release handle for layer 0
bwd layer 0 → release handle for layer 1

In this pattern, there is effectively no opportunity for memory reuse across layers or microbatches, since each piece of data must be preserved until its backward is done. As a result, introducing a buffer pool does not actually improve reuse, but instead increases memory overhead by keeping multiple buffers alive.

Given that, would it make more sense to follow the HEP/DEP approach and manage these tensors via lightweight handles with explicit lifecycle, rather than introducing a pool abstraction?

Autumn1998: A simple mental model we have been using is the following:

output, handle = dispatch(input, EpBuffer)
output = combine(input, EpBuffer, handle)
output = dispatch_bwd(input, EpBuffer, handle) 
output = combine_bwd(input, EpBuffer, handle)

In this model, the EpBuffer mainly serves as the execution context for dispatch/combine. When zero-copy is not enabled, it may internally hold some communication buffers, but it does not need to own any per-layer / per-microbatch persistent state. All such state is carried by the handle and follows the forward–backward lifecycle.

Given this, is there anything fundamentally different in NCCL EP that requires a different design, such as introducing a buffer pool?

YangFei1990: @Autumn1998 Can you elaborate how that is handled by HEP/DEP? I do not see anything on framework/mcore side so I assume this buffer is managed internally in HEP/DEP?

megatron/core/transformer/moe/fused_a2a.py
phu0ngng: If you call `dist.destroy_process_group` explicitly, then you need to call `ep_finalize()` before, which is correct.

megatron/core/transformer/moe/token_dispatcher.py
phu0ngng: > topk_idx will be int64

Then why do you need to apply a conversion here i.e. `to(torch.int64)`?

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -600,3 +600,141 @@ def hybrid_ep_combine(x, handle, num_permuted_tokens, pad_multiple, fused=False)
...
+                grouped by local expert (no separate compaction step).
+              * ``tokens_per_expert``: ``int32`` ``[num_local_experts]`` device tensor of
+                received counts per local expert (feeds grouped GEMM as group sizes;
+                alignment-padded, == actual when ``alignment=0``).
+              * ``dispatched_probs``: ``float32`` ``[recv_capacity_per_rank]`` per-slot
+                weights; apply them in the expert MLP (combine is called unweighted).
+
+            ``tokens_per_expert`` is non-differentiable.
+        """
+        recv_tokens, dispatched_probs, tokens_per_expert = te_ep.ep_dispatch(
```
Autumn1998, line 714: We directly output the tokens-per-expert right after dispatch, without distinguishing between the preprocess and dispatch stages. This looks simple, but when we build the output buffer inside dispatch, we don’t actually use the tokens-per-expert information.

I notice that in the non-static-buffer mode, we need an extra split at the end of dispatch, and a zero-padding step at the beginning of combine. However, if we already have tokens-per-expert and can precisely compute the buffer sizes, we wouldn’t need these extra split and pad operations, nor the additional static-buffer parameter. Then we could follow the design of our grouped-GEMM backend and automatically select an appropriate buffer layout.

I think we can do this inside TE？ @YangFei1990 @phu0ngng 
My idea is that since we already have sync, we can avoid these overheads.

phu0ngng (approved):

YangFei1990:
/ok to test [b38a5b0](https://github.com/NVIDIA/Megatron-LM/pull/5129/commits/b38a5b0dfe1c72d9932366a1a355328057046f8a)

YangFei1990:
/ok to test [8a62521](https://github.com/NVIDIA/Megatron-LM/pull/5129/commits/8a62521291c89681e63a30bbf1f53239b76d394b)

Autumn1998 (approved):

fanshiqing (approved):

ericharper (approved):

megatron/core/transformer/transformer_config.py
jiemingz: makes sense thanks; is there any plan to support dynamic output shape for non cudagraph modes?

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -600,3 +600,141 @@ def hybrid_ep_combine(x, handle, num_permuted_tokens, pad_multiple, fused=False)
...
+
+
+try:
+    from transformer_engine.pytorch import ep as te_ep
+
+    HAVE_TE_EP = True
+except ImportError:
+    HAVE_TE_EP = False
+
+_TE_EP_MISSING_MSG = (
```
jiemingz: nit: whats the point of putting this here? as I understand its only called in one place?

megatron/core/transformer/transformer_config.py
YangFei1990: Yes eager mode with dynamic shape is on our roadmap

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1394,6 +1404,227 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
...
+        self,
+        hidden_states: torch.Tensor,
+        async_finish: bool = True,
+        allocate_on_comm_stream: bool = True,
+    ) -> torch.Tensor:
+        # Note: this needs to stay out of the torch.compile region because TE's ep_bootstrap does
+        # opaque ProcessGroup._get_backend()._comm_ptr() access that dynamo cannot trace.
+        self._ensure_bootstrap()
+        # Fresh buffer per dispatch; held until the matching combine consumes it.
+        self._buffer = new_nccl_ep_buffer(
```
jiemingz, line 1560: nit: why put this inside the runtime dispatch? it seems `new_nccl_ep_buffer` could be called at init since it only takes `self` arguments?

YangFei1990: Several of its inputs are ready after running `setup_metadata` and `_ensure_bootstrap`, i.e. `self._max_tokens_per_rank` and `self._recv_capacity`

jiemingz (approved):
lgtm!

YangFei1990:
/ok to test [3483c83](https://github.com/NVIDIA/Megatron-LM/pull/5129/commits/3483c83a3f20c187afc6218df1f3575cd284a9b1)

PR #5135, opened 2026-06-03
Support latent MoE shared expert overlap

Victarry:
## Summary
- Enable latent MoE to use shared expert overlap with A2A/Flex dispatchers by running shared-expert pre-comm on full hidden states in `MoELayer`, while keeping shared-expert FC launches inside dispatcher overlap windows.
- Keep dispatcher-owned shared-expert pre/post handling for the non-latent path only, and add the latent shared-expert output after `fc2_latent_proj`.
- Add forward/backward parity coverage comparing shared-expert overlap vs no-overlap for latent A2A and HybridEP.

## Test plan
- PASS: `uv run unit-test run tests/unit_tests/transformer/moe/test_shared_experts.py --cluster cw --platform h100 --environment dev --time-limit 0:45:00` (SLURM `12467070`, 11 passed; includes latent A2A parity and `test_latent_shared_expert_hybridep_forward_backward`).
- PASS: `CHECK_ONLY=true BASE_REF=main UV_CACHE_DIR=/project/agentic-mcore-dev/.uv_cache uv run --project /project/agentic-mcore-dev/.claude/worktrees/latent-moe-shared-expert-overlap --extra mcore-lint -- bash tools/autoformat.sh` (Black/isort/pylint/ruff clean; wrapper exits 0; mypy still prints existing missing-dependency/type issues from this local lint env).
- PASS: `tools/check_copyright.py` on changed Python files.

## Notes
- A separate main-base replay in a temporary CW worktree reached import before test collection but was blocked by the local `mcore-ci-dev` image having `nvidia-resiliency-ext < 0.6.0`, while current `origin/main` requires `nvidia-resiliency-ext==0.6.0` (SLURM `12467214`).


kvareddy:
@sidsingh-nvidia  can you take a look at this MR. You added this support on inference side.

Victarry:
Folded the FSDP shared-expert stream synchronization fix into this latent MoE shared expert overlap PR. I also adjusted the implementation to use an opt-in parameter marker from shared expert overlap, so Megatron FSDP handles a generic side-stream grad processing requirement instead of importing MoE shared expert classes or matching parameter names.

yaox12 (approved):

PR #5148, opened 2026-06-04
Add named layouts to HyperCommGrid for heterogeneous parallelism

yashaswikarnati:
## What

Add named **views** to `HyperCommGrid`: one grid (a single rank span) can register extra factorizations beyond the implicit `base` view, then create / retrieve / enumerate process groups against any of them via a keyword-only `view=`.

- `register_view(name, shape, dim_names, shared_dims=None)` — register and validate a named factorization over the same ranks.
- `create_pg` / `get_pg` / `get_rank_enum` accept `view=` (defaults to `base`, so the single-view path is unchanged).
- Base-view group keys are byte-for-byte unchanged; view-private groups use namespaced keys, and dims listed in `shared_dims` reuse the base group instead of duplicating it.

## Why

Foundational for heterogeneous / non-colocated parallelism, where a dense (`tp/cp/dp/pp`) and an expert (`expt_tp/ep/expt_dp/pp`) factorization span the **same** ranks with different shapes. These are alternate tilings of one rank set — not orthogonal axes, so they can't be a single cube. `register_view` models each as a separate factorization that must agree on any `shared_dims`.

## How

- `_RankViewSpec` holds each factorization; the base view is auto-registered from the constructor args.
- Rank enumeration is generalized to any view's `shape`/`dim_names` via `np.moveaxis` + reshape (drops the einops dependency).
- `register_view` proves each `shared_dim` enumerates identically to the base view, so shared groups are reused rather than rebuilt.
- Robustness: `destroy()` skips groups this rank isn't a member of (`NON_GROUP_MEMBER` sentinel) and frees each shared group once; the rank-0 log is guarded by `is_initialized()`.

Fully backward compatible — all new params are optional and keyword-only. Covered by `tests/unit_tests/test_hyper_comm_grid.py`, including a real-distributed 8-GPU view test.

### DIFF
--- a/megatron/core/hyper_comm_grid.py
+++ b/megatron/core/hyper_comm_grid.py
@@ -1,18 +1,11 @@
 # Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
 
 import os
-from operator import itemgetter
-from typing import Any, Optional, Tuple, Union
+from dataclasses import dataclass
+from typing import Any, Optional, Union
 
 import numpy as np
 import torch.distributed as dist
-
-try:
-    import einops
-
-    HAVE_EINOPS = True
-except ImportError:
-    HAVE_EINOPS = False
 
 try:
     from absl import logging
@@ -40,90 +33,17 @@ def _is_process_group_member(pg: Optional[dist.ProcessGroup]) -> bool:
     return pg is not None and pg is not non_member
 
 
-class GridLayout:
-    r"""An additional named factorization of a :class:`HyperCommGrid`'s ranks.
+_BASE_VIEW_NAME = "base"
 
-    Lets one rank span carry a second factorization (e.g. an expert
-    ``[expt_tp, ep, expt_dp, pp]`` alongside the dense base ``[tp, cp, dp, pp]``). Obtained from
-    :meth:`HyperCommGrid.register_layout`, retrieved with :meth:`HyperCommGrid.get_layout`; its
-    groups are reached only through this handle (the grid's own methods use the base layout). Dims
-    in ``shared_dims`` (e.g. ``pp``, which must match between dense and expert) reuse the base
-    grid's group rather than creating a duplicate.
-    """
 
-    def __init__(
-        self,
-        grid: "HyperCommGrid",
-        name: str,
-        shape: list[int],
-        dim_names: list[str],
-        shared_dims: list[str],
-    ) -> None:
-        # [:] insures a copy.
-        self._grid = grid
-        self.name = name
-        self.shape = shape[:]
-        self.dim_names = dim_names[:]
-        self.shared_dims = shared_dims[:]
+@dataclass
+class _RankViewSpec:
+    """A named rank factorization over the same rank span as the base grid."""
 
-    def _is_shared(self, ordered_dims: list[str]) -> bool:
-        """Whether ``ordered_dims`` spans only shared dims (so the group reuses the base grid's)."""
-        return all(d in self.shared_dims for d in ordered_dims)
-
-    def create_pg(self, dims: Union[str, list[str]], **kwargs: Any) -> dist.ProcessGroup | None:
-        """Create this layout's process group for ``dims`` (collective -- call on all ranks).
-
-        A group spanning only shared dims reuses the base grid's group; otherwise it is
-        layout-private. Not idempotent (re-creating raises ``KeyError``); retrieve with
-        :meth:`get_pg`. ``kwargs`` forward to ``dist.new_subgroups_by_enumeration``.
-        """
-        ordered_dims, _ = self._grid._order_dims_for(self.dim_names, dims)
-        if self._is_shared(ordered_dims):
-            # Shared dims must resolve to the *same* group as the base layout. Delegate to the
-            # base grid so the object is shared rather than duplicated.
-            return self._grid.create_pg(dims, **kwargs)
-
-        key = (self.name, tuple(ordered_dims))
-        if key in self._grid._pgs:
-            raise KeyError(
-                f"Process group {dims} for layout {self.name!r} has already been created. Because "
-                f"there is no way to check whether options to create process group matches the "
-                f"first, we error out instead of returning the process group that has already "
-                f"been created before."
-            )
-
-        rank_enum = self._grid._gen_rank_enum_for(self.shape, self.dim_names, ordered_dims)
-        pg, _ = dist.new_subgroups_by_enumeration(rank_enum, backend=self._grid.backend, **kwargs)
-
-        if dist.is_initialized() and dist.get_rank() == 0:
-            logging.info(
-                f"Generated process group for layout {self.name!r} {ordered_dims} with "
-                f"enumeration {rank_enum}"
-            )
-        self._grid._pgs[key] = pg
-        return pg
-
-    def get_pg(self, dims: Union[str, list[str]]) -> dist.ProcessGroup:
-        """Get this layout's previously-created group for ``dims`` (shared dims return the base group).
-
-        Raises ``KeyError`` if it has not been created yet.
-        """
-        ordered_dims, _ = self._grid._order_dims_for(self.dim_names, dims)
-        if self._is_shared(ordered_dims):
-            return self._grid.get_pg(dims)
-
-        key = (self.name, tuple(ordered_dims))
-        if key not in self._grid._pgs:
-            raise KeyError(
-                f"Process group {dims} for layout {self.name!r} hasn't been created. Call "
-                f"create_pg first."
-            )
-        return self._grid._pgs[key]
-
-    def get_rank_enum(self, dims: Union[str, list[str]]) -> list[list[int]]:
-        """Rank enumeration for ``dims`` under this layout (matches the base grid's for shared dims)."""
-        ordered_dims, _ = self._grid._order_dims_for(self.dim_names, dims)
-        return self._grid._gen_rank_enum_for(self.shape, self.dim_names, ordered_dims)
+    name: str
+    shape: list[int]
+    dim_names: list[str]
+    shared_dims: list[str]
 
 
 class HyperCommGrid:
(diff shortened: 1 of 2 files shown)


yashaswikarnati:
Reworked per review — thanks for the careful read. Summary of changes:

- **Shared dims are reused, not duplicated.** `register_layout(..., shared_dims=["pp"])` validates that a shared dimension's rank enumeration matches the base layout's, and a group spanning only shared dims now returns the **base grid's group object** (same ranks). This honors the invariant that dense and expert pipeline groups must be identical (`parallel_state.py` `decoder_rank_generator.get_ranks("pp") == expert_decoder_rank_generator.get_ranks("pp")`). The previous "distinct `expert:pp`" behavior is gone.
- **No implicit layout inference.** Removed base-precedence resolution. Layout-private groups are reachable only through an explicit `GridLayout` handle (`grid.get_layout("expert").get_pg(...)`); the base grid's `create_pg`/`get_pg`/`get_rank_enum` are unchanged and operate on the base layout only.
- **Smaller surface / internal keys.** Dropped `has_layout` and the `"<layout>:<dims>"` string namespacing; layout-private groups are keyed by `(layout_name, ordered_dims)` tuples. Base-grid behavior is byte-for-byte unchanged for existing callers.
- **Real distributed coverage.** Replaced the monkey-patched `create_pg` tests with a real 8-rank integration test (`test_real_distributed_registered_layout`) that registers base + expert layouts, creates groups in both, asserts actual rank membership of an expert-private group (with a real `all_reduce`), asserts the shared `pp` group is the same object/ranks as base, and that `destroy()` frees reused groups exactly once. Pure-Python `register_layout` validation tests are kept.

Validated on 1 node × 8 GPUs (`torch.distributed.run --nproc-per-node 8`): all green.

megatron/core/hyper_comm_grid.py
```diff
@@ -30,6 +30,157 @@
     HAVE_ABSL = False
 
 
+def _is_process_group_member(pg: Optional[dist.ProcessGroup]) -> bool:
+    """Return whether ``pg`` is a process group the current rank belongs to.
+
+    ``new_subgroups_by_enumeration`` returns ``GroupMember.NON_GROUP_MEMBER`` (and
```
yashaswikarnati: remove verbosity in doc strings

megatron/core/hyper_comm_grid.py
```diff
@@ -30,6 +30,157 @@
...
+    r"""An additional named factorization of a :class:`HyperCommGrid`'s rank span.
+
+    A grid is constructed with one *base* factorization (its ``shape``/``dim_names``). Some
+    rank universes admit more than one valid factorization simultaneously, for example a dense
+    ``["tp", "cp", "dp", "pp"]`` factorization alongside an expert
+    ``["expt_tp", "ep", "expt_dp", "pp"]`` factorization that re-partitions the very same ranks.
+    A :class:`GridLayout` is the explicit handle through which the second (and further)
+    factorization is materialized into process groups.
+
+    The handle is obtained from :meth:`HyperCommGrid.register_layout` (by the module that owns
```
yashaswikarnati: from below here, we dont need ? doc string verbose.

megatron/core/hyper_comm_grid.py
```diff
@@ -30,6 +30,157 @@
...
+
+    def _is_shared(self, ordered_dims: list[str]) -> bool:
+        """Return whether ``ordered_dims`` is exactly (a subset spanning only) shared dims.
+
+        A group is reused from the base grid only when every dim it spans is a shared dim. Any
+        layout-private dim makes the whole group layout-private.
+        """
+        return all(d in self.shared_dims for d in ordered_dims)
+
+    def create_pg(self, dims: Union[str, list[str]], **kwargs: Any) -> dist.ProcessGroup | None:
```
yashaswikarnati: do we also need create pg for this grid layout ?

megatron/core/hyper_comm_grid.py
```diff
@@ -30,6 +30,157 @@
...
+        r"""Create a process group for ``dims`` under this layout.
+
+        If ``dims`` spans only shared dims, creation is delegated to the base grid so the shared
+        ranks map to a single process group object across layouts (it is the base group, under the
+        base key). Otherwise a layout-private group is created and stored under this layout.
+
+        Like :meth:`HyperCommGrid.create_pg`, this is not idempotent: creating a group that already
```
yashaswikarnati: verbose doc strings

megatron/core/hyper_comm_grid.py
yashaswikarnati: Done — trimmed to two lines (kept only the NON_GROUP_MEMBER sentinel rationale, since that's the non-obvious bit).

megatron/core/hyper_comm_grid.py
yashaswikarnati: Done — tightened all the GridLayout docstrings (class + create_pg/get_pg/get_rank_enum/_is_shared) to one or two lines each.

megatron/core/hyper_comm_grid.py
yashaswikarnati: Yes — `create_pg` on the layout is needed: the layout-private expert-factorization groups (e.g. `expt_tp`, `ep`, `expt_tp×ep`) are distinct NCCL subgroups that the base grid's `create_pg` can't materialize (it only knows the base factorization). It mirrors the base grid's contract: `create_pg` is the collective creation step (all ranks must call it), `get_pg` is a local lookup. We deliberately don't create lazily inside `get_pg` — that would hide a collective behind a local-looking call and risk rank desync. A group spanning only shared dims (e.g. `pp`) delegates to the base grid so it's reused, not duplicated. Kept it; trimmed its docstring.

megatron/core/hyper_comm_grid.py
yashaswikarnati: Done — trimmed.

megatron/core/hyper_comm_grid.py
```diff
@@ -30,6 +30,102 @@
...
+    """
+    non_member = getattr(getattr(dist, "GroupMember", None), "NON_GROUP_MEMBER", None)
+    return pg is not None and pg is not non_member
+
+
+class GridLayout:
+    r"""An additional named factorization of a :class:`HyperCommGrid`'s ranks.
+
+    Lets one rank span carry a second factorization (e.g. an expert
+    ``[expt_tp, ep, expt_dp, pp]`` alongside the dense base ``[tp, cp, dp, pp]``). Obtained from
```
yashaswikarnati: remove everything from here. Obtained from
    :meth:`HyperCommGrid.register_layout`, retrieved with :meth:`HyperCommGrid.get_layout`; its
    groups are reached only through this handle (the grid's own methods use the base layout). Dims
    in ``shared_dims`` (e.g. ``pp``, which must match between dense and expert) reuse the base
    grid's group rather than creating a duplicate. (verbose comments are not needed)

megatron/core/hyper_comm_grid.py
```diff
@@ -115,7 +218,103 @@ def __init__(
...
+        name: str,
+        shape: list[int],
+        dim_names: list[str],
+        shared_dims: Optional[list[str]] = None,
+    ) -> GridLayout:
+        r"""Register an additional factorization over this grid's rank span.
+
+        Returns a :class:`GridLayout` handle through which the layout's process groups are
+        created and retrieved. The base layout (the constructor's ``shape``/``dim_names``) is
+        unaffected.
```
yashaswikarnati: very verbose doc string

megatron/core/hyper_comm_grid.py
```diff
@@ -224,14 +429,24 @@ def _gen_rank_enum(self, dims: list[str]) -> list[list[int]]:
...
         easier to test against MCore's RankGenerator
         """
+        return self._gen_rank_enum_for(self.shape, self.dim_names, dims)
+
+    def _gen_rank_enum_for(
+        self, shape: list[int], dim_names: list[str], dims: list[str]
+    ) -> list[list[int]]:
+        r"""Generate rank enumeration for ``dims`` under an explicit ``shape``/``dim_names``.
 
+        Identical logic to :meth:`_gen_rank_enum` but parameterized by the factorization, so it
```
yashaswikarnati: rermove verbose comment starting from Identical logic to :meth:`_gen_rank_enum` but parameterized by

megatron/core/hyper_comm_grid.py
```diff
@@ -243,17 +458,30 @@ def _gen_rank_enum(self, dims: list[str]) -> list[list[int]]:
...
+        ordered_dims, _ = self._order_dims_for(self.dim_names, dims)
+        unique_group_key = "-".join(ordered_dims)
+        return ordered_dims, unique_group_key
+
+    def _order_dims_for(
+        self, dim_names: list[str], dims: Union[str, list[str]]
+    ) -> Tuple[list[str], str]:
+        r"""Reorder ``dims`` against an explicit ``dim_names``.
+
+        Identical ordering logic to :meth:`_order_dims` but parameterized by the factorization's
```
yashaswikarnati: remove verbose comment starting from - Identical ordering logic to :meth:`_order_

tests/unit_tests/test_hyper_comm_grid.py
```diff
@@ -309,6 +309,164 @@ def test_rank_enumeration_correctness(self):
         expected_ab = [[0, 2, 1, 3], [4, 6, 5, 7]]
         assert rank_enum_ab == expected_ab
 
+    # ------------------------------------------------------------------
+    # register_layout / GridLayout validation (mock-based unit tests)
+    # ------------------------------------------------------------------
+
+    def test_register_layout_returns_handle(self):
```
yashaswikarnati: all these pytest raise error test, trim them to actual valued tests and which actually tests something, i see lot of trivial tests

megatron/core/hyper_comm_grid.py
```diff
@@ -30,6 +24,29 @@
     HAVE_ABSL = False
 
 
+def _is_process_group_member(pg: Optional[dist.ProcessGroup]) -> bool:
+    """Whether the current rank belongs to ``pg``.
+
+    ``new_subgroups_by_enumeration`` returns the ``GroupMember.NON_GROUP_MEMBER``
```
yashaswikarnati: remove - new_subgroups_by_enumeration`` returns the ``GroupMember.NON_GROUP_MEMBER``
    sentinel (not ``None``) for non-member ranks, which must not be destroyed.

megatron/core/hyper_comm_grid.py
```diff
@@ -41,6 +58,13 @@ class HyperCommGrid:
     For any combination of dimensions, a process group can only be created once.
     Creating process groups for the same combination with different options is not supported.
 
+    The grid's own :meth:`create_pg` / :meth:`get_pg` / :meth:`get_rank_enum` operate on the base
```
yashaswikarnati: remove verbise dic string, can be concisely said >

deepakn94 (approved):

yaoyu-33 (approved):

yashaswikarnati:
/ok to test [644aab2](https://github.com/NVIDIA/Megatron-LM/pull/5148/commits/644aab25ef52db5f5882d9104038885554fa0fc6)

yashaswikarnati:
/ok to test [dd90fc6](https://github.com/NVIDIA/Megatron-LM/pull/5148/commits/dd90fc6b0df8d8be74657a6cf97194a2078b9260)

PR #5150, opened 2026-06-04
Apply MIMO SP/CP sharding with explicit groups and enable THD in non-colocated path

yashaswikarnati:
## What

Make MIMO language-model input sharding correct for **non-colocated / heterogeneous** parallelism, where the LM owns its own process-group grid.

`PartitionAdapter.shard()` owns the layout contract — sequence-first `(S, B, H)` in and out:

- **Explicit TP group** for the SP scatter (was the global `parallel_state` group).
- **Correct SP dim**: scatters the *sequence*; CP transposes to batch-first only internally (for `get_batch_on_this_cp_rank`), and SP-only does no transpose. Previously the SP path scattered `[B, S, H]` along the batch dim.
- **`loss_mask` and `packed_seq_params` threaded** through the LM forward and CP-sharded, so per-token loss and packed/THD sequences line up with the CP-local output on the non-colocated path.
- **No dense `attention_mask` in the adapter**; fast-fail if a dense mask is combined with CP (mask via a causal `attn_mask_type` / `packed_seq_params`).

`shard()` returns `(embeddings[S/(cp*tp),B,H], labels[B,S/cp], loss_mask[B,S/cp], packed_seq_params)`. Single in-core caller (`MimoModel`).

## Backward compatibility

Colocated path unchanged (still returns `(output, loss_mask)`). Under CP, `position_ids` are passed full-length — the rotary module shards them internally, so MIMO must not.

## Testing

Real 8-GPU tests assert per-rank shape **and** content for SP-only / CP-only / CP+SP, plus loss-mask CP-sharding and the attention-mask-under-CP guard. **122 passed, 2 skipped** on 1 node × 8 GPUs.

No new direct `parallel_state.get_*_group()` reads.

### DIFF
--- a/megatron/core/models/mimo/model/base.py
+++ b/megatron/core/models/mimo/model/base.py
@@ -2,7 +2,7 @@
 
 import logging
 import warnings
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Tuple
 
 import torch
 
@@ -77,8 +77,11 @@ class MimoModel(MegatronModule):
         max_seq_len = mimo_config.language_model_spec.params.get('max_sequence_length', 4096)
 
         self.partition_adapter: Optional[PartitionAdapter] = None
-        # Create partition adapter only if parallelism is enabled
-        if language_config.context_parallel_size > 1 or language_config.sequence_parallel:
+        # Only on language-module ranks: encoder-only ranks never shard and would read
+        # process groups they do not own.
+        if self.role.has_language_module and (
+            language_config.context_parallel_size > 1 or language_config.sequence_parallel
+        ):
             partition_config = PartitionConfig.from_mp_config(
                 mp=language_config,
                 max_seq_len=max_seq_len,
@@ -307,10 +310,24 @@ class MimoModel(MegatronModule):
         else:
             position_ids_text = position_ids[batch_idx, seq_idx].unsqueeze(0)
 
-        text_embeddings = (
-            unwrap_model(self.language_model)
-            .embedding(input_ids=input_ids_text, position_ids=position_ids_text)
-            .squeeze(1)
+        embedding_layer = unwrap_model(self.language_model).embedding
+        # Combined embeddings are SP-scattered later in PartitionAdapter; a second scatter
+        # here would split the flat text tokens across TP ranks before alignment.
+        if (
+            self.partition_adapter is not None
+            and self.partition_adapter.cfg.seq_parallel
+            and getattr(embedding_layer, 'scatter_to_sequence_parallel', False)
+        ):
+            raise RuntimeError(
+                "MIMO sequence parallelism requires the language embedding scatter to be "
+                "disabled; pass scatter_embedding_sequence_parallel=False when constructing "
+                "the language model."
+            )
+
+        text_embeddings = embedding_layer(
+            input_ids=input_ids_text, position_ids=position_ids_text
+        ).squeeze(
+            1
         )  # Shape: [num_text_tokens, hidden_dim]
         return text_embeddings
 
@@ -382,11 +399,14 @@ class MimoModel(MegatronModule):
                 return self._forward_encoders(input_ids, modality_inputs, input_tensors), loss_mask
 
             if self.role.has_language_module:
-                return (
-                    self._forward_language_module(
-                        input_ids, position_ids, attention_mask, labels, input_tensors
-                    ),
+                return self._forward_language_module(
+                    input_ids,
+                    position_ids,
+                    attention_mask,
                     loss_mask,
+                    labels,
+                    input_tensors,
+                    packing_kwargs,
                 )
 
             raise RuntimeError(f"Rank has no modules assigned in role: {self.role}")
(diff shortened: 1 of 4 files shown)


megatron/core/models/mimo/model/base.py
```diff
@@ -77,8 +77,12 @@ def __init__(self, mimo_config: MimoModelConfig, cp_group=None, tp_group=None) -
         max_seq_len = mimo_config.language_model_spec.params.get('max_sequence_length', 4096)
 
         self.partition_adapter: Optional[PartitionAdapter] = None
-        # Create partition adapter only if parallelism is enabled
-        if language_config.context_parallel_size > 1 or language_config.sequence_parallel:
+        # Create the partition adapter only on ranks that own the language module and
```
yashaswikarnati: remove verbose comment

megatron/core/models/mimo/model/base.py
```diff
@@ -307,10 +311,26 @@ def get_text_embeddings(
         else:
             position_ids_text = position_ids[batch_idx, seq_idx].unsqueeze(0)
 
-        text_embeddings = (
-            unwrap_model(self.language_model)
-            .embedding(input_ids=input_ids_text, position_ids=position_ids_text)
-            .squeeze(1)
+        embedding_layer = unwrap_model(self.language_model).embedding
+        # MIMO scatters the *combined* embeddings for sequence parallelism in
```
yashaswikarnati: remove verbose comment

yashaswikarnati:
Updated per review:

- **Removed the dense-`attention_mask` CP guard.** It was inconsistent (only `_shard_language_inputs` had it; `_forward_all_modules` didn't pass the mask in) and dense masks aren't a supported MIMO path. The adapter now simply does not handle a dense `attention_mask` — documented that masking goes through causal attention / `packed_seq_params`. Dropped the two guard tests.
- **Narrowed the docstring.** Scoped the caveat to CP-local hidden states (not "whenever CP/SP is active"); clarified SP scatters only the embeddings, and labels/loss_mask are CP-sharded but never SP-scattered.
- **Real 8-GPU tests instead of mocks.** Replaced the mock-collective sharding tests with `TestPartitionAdapterShardRealDistributed`, which builds real tensor/context-parallel groups on 8 ranks and verifies per-rank sharded **shape and content** (positionally) for SP-only (`[S/tp,B,H]`), CP-only (`[S/cp,B,H]` zigzag), and combined CP+SP (`[S/(cp*tp),B,H]`), plus that labels/loss_mask are CP-sharded but not SP-scattered. Renamed the artificial "both disabled" test to reflect the transpose-only contract. Kept the pure-Python validation tests.

Validated on 1 node × 8 GPUs (`--experimental`): all green, including the three real-distributed sharding tests.

megatron/core/models/mimo/model/base.py
```diff
@@ -497,25 +514,83 @@ def _empty_encoder_output(self, encoder_name: str) -> torch.Tensor:
...
+        packed_seq_params: Optional[PackedSeqParams] = None,
+    ) -> Tuple[
+        Optional[torch.Tensor],
+        Optional[torch.Tensor],
+        Optional[torch.Tensor],
+        Optional[PackedSeqParams],
+    ]:
+        """Apply CP/SP sharding and return language-model-layout embeddings.
+
+        ``embeddings`` arrive in (S, B, H) layout (the layout produced by
```
yashaswikarnati: very verbose doc string, we dont need such verbose doc strings

megatron/core/models/mimo/model/base.py
```diff
@@ -497,25 +514,83 @@ def _empty_encoder_output(self, encoder_name: str) -> torch.Tensor:
             requires_grad=True,
         )
 
+    def _shard_language_inputs(
```
yashaswikarnati, line 531: do we even need this function? if all it does is calling partition adapter shard ?

megatron/core/models/mimo/model/base.py
```diff
@@ -679,25 +765,15 @@ def _forward_all_modules(
...
-                    loss_mask=loss_mask,
-                    attention_mask=attention_mask,
-                    packed_seq_params=packed_seq_params,
-                )
-            )
-            # shard() returns embeddings in [B, S/cp, H]; transpose to [S/cp, B, H]
-            # which is what the language model expects.
-            if combined_embeddings is not None:
-                combined_embeddings = combined_embeddings.transpose(0, 1).contiguous()
+        # 3. If sharding is needed, apply PartitionAdapter. combined_embeddings is
```
yashaswikarnati: how many total transposes are happening in this back and forth ?

megatron/core/models/mimo/partition/utils.py
```diff
@@ -100,100 +100,108 @@ def __init__(self, cfg: PartitionConfig):
...
+        Tensor inputs use the batch-first dataloader layout:
             - embeddings: (B, S, H)
-            - labels / loss_mask / attention_mask: (B, S)
-
-        After this call embeddings are still in (B, S/cp, H) batch-first layout.
-        The caller is responsible for transposing to (S/cp, B, H) if the language model
-        requires sequence-first tensors.
+            - labels / loss_mask: (B, S)
+
+        Embeddings are returned in language-model layout (S, B, H) with CP and SP applied
```
yashaswikarnati: verbose doc string

megatron/core/models/mimo/partition/utils.py
```diff
@@ -100,100 +100,108 @@ def __init__(self, cfg: PartitionConfig):
 
     def shard(
         self,
-        embeddings: torch.Tensor,
-        labels: torch.Tensor,
-        loss_mask: torch.Tensor,
-        attention_mask: torch.Tensor,
+        embeddings: Optional[torch.Tensor],
```
yashaswikarnati, line 98: why did we change type hints? did we make them optional ?

megatron/core/models/mimo/partition/utils.py
```diff
@@ -100,100 +100,108 @@ def __init__(self, cfg: PartitionConfig):
...
+        # which splits along the leading (sequence) dimension.
+        if embeddings is not None:
+            embeddings = embeddings.transpose(0, 1).contiguous()
+            if self.cfg.seq_parallel:
+                embeddings = tensor_parallel.scatter_to_sequence_parallel_region(
+                    embeddings, group=self.cfg.tp_group
+                )
 
-        return embeddings, labels, loss_mask, attention_mask, packed_seq_params
+        return embeddings, labels, loss_mask, packed_seq_params
```
yashaswikarnati, line 166: what happened to attention mask ?

megatron/core/models/mimo/partition/utils.py
```diff
@@ -216,12 +222,10 @@ def _apply_context_parallel(
                 - embeddings (Optional[torch.Tensor]): Sharded embeddings. Shape: (B, S/cp, H)
                 - labels (Optional[torch.Tensor]): Possibly sharded labels. Shape: (B, S/cp)
                 - loss_mask (Optional[torch.Tensor]): Possibly sharded loss mask. Shape: (B, S/cp)
-                - attention_mask (Optional[torch.Tensor]): Possibly sharded attention mask.
-                                                           Shape: (B, S/cp)
                 - packed_seq_params (PackedSeqParams, optional): Updated packed sequence parameters.
         """
         if not self.cfg.use_cp:
-            return embeddings, labels, loss_mask, attention_mask, packed_seq_params
+            return embeddings, labels, loss_mask, packed_seq_params
```
yashaswikarnati, line 201: why we kept removing attention mask

yashaswikarnati:
Revised per review. All inline comments addressed; verified on 1 node × 8 GPUs (`--experimental`): **122 passed, 2 skipped** (the 2 skips are pre-existing 1f1b tests needing 2/4 GPUs).

### Design changes
- **`shard()` is now sequence-first.** It consumes `(S, B, H)` (what `align_embeddings_by_token_positions` produces and the LM consumes) and transposes to batch-first **only inside the CP block** (`get_batch_on_this_cp_rank` needs batch-first). The **SP-only path now scatters dim 0 directly with zero transposes** — the previous `[S,B,H]→[B,S,H]→[S,B,H]` round-trip is gone. (re: "how many transposes…")
- **Dense `attention_mask` under CP is now rejected** in `_forward_language_module` (fast-fail) — a dense `[B,S]` mask cannot line up with the CP-sharded sequence; masking goes through a causal `attn_mask_type` or `packed_seq_params`. This closes the gap where the non-colocated path forwarded the mask unguarded while the colocated path forces `None`. (re: "what happened to attention mask / why remove it")
- **`packed_seq_params` is now threaded through the non-colocated path** (`packing_kwargs → _forward_language_module → shard() → LM`) via a shared `_build_packed_seq_params` helper, so packed/THD sequences are CP-sharded and reach the LM — matching the colocated path (previously unsupported there).

### Inline comments
- **`_shard_language_inputs` "do we even need this?"** — kept, but it's now just the `None`-guard + `shard()` call (the transpose moved into `shard()`); the guard is shared across its 3 call sites. Docstring trimmed.
- **Verbose comments/docstrings** (`:80`, `:315`, `:531`, `:768`, `shard()` docstring) — trimmed to the load-bearing rationale.
- **`Optional` type hints** — kept; `None` is reachable on non-first PP stages (`embeddings=None`, labels-only sharding).
- **Dead `is_partitioning_enabled`** — removed (was test-only) along with its 4 tests.
- **position_ids / mRoPE** — left full-length **on purpose**: the rotary module (`RotaryEmbedding`/`MultimodalRotaryEmbedding`) CP-shards positions internally with the same zigzag as the hidden states, so pre-sharding would double-shard. No change.

### Tests
Real 8-GPU sharding tests (SP-only, CP-only, CP+SP) assert per-rank **shape and content**; added the attention-mask-under-CP guard test; kept loss-mask CP-sharding coverage.

tests/unit_tests/models/mimo/test_mimo_partition.py
```diff
@@ -151,134 +130,56 @@ def _make_cfg(
...
-        out = adapter.shard(embeddings, labels, loss_mask, attention_mask)
+        embeddings, labels, loss_mask = self._make_tensors(B=2, S=8, H=16)
+        out = adapter.shard(embeddings, labels, loss_mask)
         assert out[0] is embeddings
         assert out[1] is labels
         assert out[2] is loss_mask
-        assert out[3] is attention_mask
-        assert out[4] is None
-
-    def test_cp_only_shards_sequence(self):
```
yashaswikarnati, line 172: why did we remove these tests? are these not adding value?

yaoyu-33 (approved):

yashaswikarnati:
/ok to test [383cd54](https://github.com/NVIDIA/Megatron-LM/pull/5150/commits/383cd542635fda182233351b494160a1a04a78be)

kvareddy (approved):

PR #5183, opened 2026-06-05
test(elastification): make Flextron gumbel test deterministic

### LINKED ISSUE #5155: 🐛 Flaky CI: test_gumbel_determinism (FlextronRouter hard_sample uses un-seeded random)
**Describe the bug**

CI test `tests/unit_tests/elastification/test_hybrid_flex_router.py::TestFlextronRouter::test_gumbel_determinism` is **flaky**. It passes on most ranks/runs but intermittently fails with:

```
AssertionError: Tensor-likes are not equal!
Mismatched elements: 2 / 2 (100.0%)
Greatest absolute difference: 0.4191093146800995 at index (1,)
Greatest relative difference: inf at index (1,)
tests/unit_tests/elastification/test_hybrid_flex_router.py:180: AssertionError
```

First observed in job [`tests/unit_tests/**/*.py - latest`](https://github.com/NVIDIA/Megatron-LM/actions/runs/26937673156/job/79486685125) (rank 3 failed while rank 0 passed all 1460 tests). The triggering PR (#5152) is a CI-notification config change and is unrelated to the failure.

Tag @NVIDIA/mcore-oncall to get oncall's attention to this issue.

**Failing run**

| Field | Value |
|-------|-------|
| PR    | [#5152: ci(notify): mention mcore-oncall and philipp on critical CI events](https://github.com/NVIDIA/Megatron-LM/pull/5152) |
| Run   | [26937673156](https://github.com/NVIDIA/Megatron-LM/actions/runs/26937673156) |
| Job   | [tests/unit_tests/**/*.py - latest](https://github.com/NVIDIA/Megatron-LM/actions/runs/26937673156/job/79486685125) |

**Root cause**

The test asserts two routers produce **bit-identical** Gumbel-softmax logits (`torch.testing.assert_close(..., atol=0, rtol=0)`). The Gumbel sample itself is deterministic — `_dp_gumbel_softmax` wraps `F.gumbel_softmax` in `torch.manual_seed(seed)` with save/restore of the RNG state.

However, `FlextronRouter.forward` derives a separate stochastic flag from Python's **global, un-seeded** RNG:

```python
# megatron/elastification/router/hybrid_flex_router.py:545
hard_sample = random.random() > self.hard_sample_th   # hard_sample_th = 0.996
```

This `random.random()` draw is **not** covered by the torch seeding, so `router_a(1.0)` and `router_b(1.0)` each draw independently:

- ~99.6% of the time both get `hard_sample=False` (soft Gumbel) → identical logits → pass.
- ~0.4% of the time one router draws `> 0.996` (`hard=True`, one-hot) while the other stays soft → divergent logits → the `atol=0, rtol=0` assert fails. The `inf` relative difference is the zero entry of the one-hot vector.

It surfaces on an arbitrary rank because each rank's global Python-RNG state differs by the time this test runs.

**Proposed fix**

Make the test deterministic by seeding Python's `random` identically before each router forward, so both routers observe the same `hard_sample`:

```python
random.seed(0)
out_a = router_a(1.0)
random.seed(0)
out_b = router_b(1.0)
```

This preserves the test's intent (verifying Gumbel **RNG** determinism) without touching production behavior. A separate, deeper question for the elastification owner: whether `hard_sample` should be synchronized across DP ranks during real training (currently each process draws independently from `random.random()`).

**Mitigation (already applied)**

The test has been quarantined with `@pytest.mark.flaky_in_dev` to keep the default CI pipeline green until the fix above lands.

```bash
uv run python -m torch.distributed.run --nproc-per-node 8 -m pytest -q \
  tests/unit_tests/elastification/test_hybrid_flex_router.py::TestFlextronRouter::test_gumbel_determinism
```

Repeat across several runs / ranks to observe the intermittent failure.

**Additional context**

Triaged automatically via `/create-issue`.

Williams500:

# What does this PR do ?

Make `TestFlextronRouter::test_gumbel_determinism` deterministic by controlling the Python `random` draw that selects the router hard/soft branch.

## Issue tracking

Linked issue: Fixes #5155

### Pre-checks

  - N/A: this is a test-only fix that updates an existing unit test and removes the `flaky_in_dev` quarantine once the test controls the extra Python RNG source.
  - N/A: this PR only fixes a unit-test flake and does not change production training behavior or numerics.
  - N/A: no public API, production code, or type signatures changed.
  - N/A: no user-facing behavior or docs surface changed.

## Summary

`test_gumbel_determinism` compares two `FlextronRouter` instances bit-for-bit after copying weights. The Gumbel path is torch-seeded, but `FlextronRouter.forward()` also samples `hard_sample` from Python `random.random()`. If the two forwards draw different hard/soft branches, the bit-exact assertion can fail even though the Gumbel seed path is deterministic.

This change seeds Python `random` identically before each router forward in the test, restores the previous Python RNG state afterward, and removes the `flaky_in_dev` quarantine. Production router behavior is unchanged.

## Test plan

- [x] `uv run isort --check-only tests/unit_tests/elastification/test_hybrid_flex_router.py`
- [x] `uv run python -m py_compile tests/unit_tests/elastification/test_hybrid_flex_router.py`
- [x] `git diff --check`
- [x] `PATH="$PWD/.venv/bin:$PATH" BASE_REF=main CHECK_ONLY=true SKIP_DOCS=false bash tools/autoformat.sh`
  - `black`, `isort`, `pylint`, and `ruff` passed for the changed file.
  - The script reached `mypy`, but local `.venv` does not contain `mypy`; the repository script handles this with `|| true` and exits 0.
- [x] `uv run python - <<'PY' ...` verified `random.seed(0)` produces the same `hard_sample` draw (`0.84442185152504812`, below `hard_sample_th=0.996`) for both forwards.
- [ ] `uv run python -m pytest -q tests/unit_tests/elastification/test_hybrid_flex_router.py::TestFlextronRouter::test_gumbel_determinism` did not run locally: the local uv environment fails while loading `tests/unit_tests/conftest.py` with `ModuleNotFoundError: No module named 'torch'`. This machine also has no local NVIDIA GPU, so the CUDA-backed target test needs CI validation.


guihong-nv:
/ok to test [091e02e](https://github.com/NVIDIA/Megatron-LM/pull/5183/commits/091e02e33fd6432b9707a88c1b9a209318e1e0a4)

guihong-nv (approved):
LGTM

ericharper (approved):

PR #5189, opened 2026-06-05
Fix Emerging-Optimizers install instructions

### LINKED ISSUE #5179: Muon optimizer is unusable: `emerging_optimizers` not installable from PyPI
### Description
`megatron/core/optimizer/muon.py` requires the `emerging_optimizers` package:

```python
# muon.py
from emerging_optimizers.orthogonalized_optimizers import (
    OrthogonalizedOptimizer, get_muon_scale_factor,
)
from emerging_optimizers.orthogonalized_optimizers.muon_utils import newton_schulz_tp
HAVE_EMERGING_OPTIMIZERS = True
```

If the import fails, `HAVE_EMERGING_OPTIMIZERS=False` and any attempt to build the optimizer hits:

```python
# muon.py:187
assert HAVE_EMERGING_OPTIMIZERS, "Emerging Optimizers is not installed."
```

i.e. **`--optimizer muon` is completely unusable without this package.**

### Problem
`pip install emerging_optimizers` does **not** install a working package — the name on PyPI resolves to a placeholder/stub (version `999.9.9`), not the module `muon.py` imports. I could not find the real package published anywhere public, nor a pin/source in `requirements*.txt`, `pyproject.toml`, or the docs.

### Questions
1. Where is the real `emerging_optimizers` package meant to be obtained (PyPI? a public git repo? NGC)?
2. Could it be published to PyPI, or pinned via a git URL in the requirements, or vendored?
3. Could `muon.py` print an actionable install hint at the `HAVE_EMERGING_OPTIMIZERS` assert (where to get it) instead of just "not installed"?

### Repro
```
git checkout d942260dd
pip install emerging_optimizers        # installs the stub, not the real module
python -c "from emerging_optimizers.orthogonalized_optimizers import OrthogonalizedOptimizer"
# ImportError -> HAVE_EMERGING_OPTIMIZERS=False -> Muon unusable
```

### Env
Megatron-LM @ `d942260dd`, Python 3.12, CUDA/torch (B200). Single-node.

janEbert:

# What does this PR do ?
Fix Emerging-Optimizers install instructions, which were previously incorrectly pointing at PyPI.

## Issue tracking

Fix https://github.com/NVIDIA/Megatron-LM/issues/5179

### DIFF
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -756,7 +756,8 @@ def _get_megatron_emerging_optimizer(
     if not HAVE_EMERGING_OPTIMIZERS:
         raise ImportError(
             f"emerging-optimizers package is required for optimizer='{eopt_name}'. "
-            "Install it with: pip install emerging-optimizers"
+            "Install it with: "
+            "pip install git+https://github.com/NVIDIA-NeMo/Emerging-Optimizers.git"
         )
     assert not (use_layer_wise and config.overlap_param_gather_with_optimizer_step), (
         "overlap_param_gather_with_optimizer_step is not supported with "
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -756,7 +756,7 @@ def _get_megatron_emerging_optimizer(
     if not HAVE_EMERGING_OPTIMIZERS:
         raise ImportError(
             f"emerging-optimizers package is required for optimizer='{eopt_name}'. "
-            "Install it with: "
+            "Install it with:\n"
             "pip install git+https://github.com/NVIDIA-NeMo/Emerging-Optimizers.git"
         )
     assert not (use_layer_wise and config.overlap_param_gather_with_optimizer_step), (


ericharper (approved):

megatron/core/optimizer/__init__.py
```diff
@@ -756,7 +756,8 @@ def _get_megatron_emerging_optimizer(
     if not HAVE_EMERGING_OPTIMIZERS:
         raise ImportError(
             f"emerging-optimizers package is required for optimizer='{eopt_name}'. "
-            "Install it with: pip install emerging-optimizers"
+            "Install it with:\n"
+            "pip install git+https://github.com/NVIDIA-NeMo/Emerging-Optimizers.git"
```
skyw, line 760: Nit: may be include versions in the instruction?

skyw (approved):

megatron/core/optimizer/__init__.py
janEbert: Then we'd have to continuously update it :/

janEbert: So I've started to look into how to properly do this. One option would be to add an extra `setuptools.package-data` file into the package (e.g., `megatron/core/_emerging-optimizers_rev.toml`, created during `setup.py`. The file would contain the EO version/revision string, simply parsed from `pyproject.toml`, and we could parse that file to improve this message at runtime.

The caveat of this approach is that the package gets "dirtied" with this extra file and the required changes become a bit broader. I suppose especially the modification of `setup.py` would be critical/undesired.

janEbert:
This PR has become redundant since we now own the PyPI package, making the instructions correct.

PR #5195, opened 2026-06-05
fix(moe): align aux-loss routing map with dispatch for grouped routing and expert bias

### LINKED ISSUE #2767: [QUESTION] MoE score for aux loss
I noticed that compute_routing_scores_for_aux_loss in megatron/core/transformer/moe/moe_utils.py does not handle grouped topk routing, while topk_routing_with_score_function does. This may cause the auxiliary loss to optimize for a different routing distribution than what is actually used for token dispatching.

Is this intentional behavior, or is it a bug?
If intentional, what is the reasoning behind using different routing logic for aux loss vs. actual token dispatching?

Connor-XY:
## What

Fixes #2767. The MoE auxiliary-loss routing map is built by `compute_routing_scores_for_aux_loss`, but it selected experts differently from the actual dispatch path (`topk_routing_with_score_function`), so the aux loss could optimize for a routing distribution that differs from what is dispatched.

Two sources of divergence are addressed:

1. **Group-limited routing** — the aux-loss map used a plain `torch.topk`, ignoring `moe_router_num_groups`/`moe_router_group_topk`. It now uses the same `group_limited_topk` selection as dispatch.
2. **Expert bias** — for `sigmoid`/`sqrtsoftplus`, dispatch adds `expert_bias` to the *un-normalized* scores before top-k. The aux-loss path now threads `expert_bias` through and selects on `raw_scores + expert_bias`, while still returning the normalized scores used by the loss.

The fused aux-loss kernel is bypassed when group routing or expert bias is in use (it cannot account for either). `softmax` is unaffected — the dispatch path never applies expert bias there, and the pre-/post-softmax selection input is preserved.

## Why

With grouping and/or aux-loss-free expert bias enabled (e.g. DeepSeek-V3-style configs), the routing map fed to the aux loss did not match the tokens actually dispatched, so the load-balancing signal was computed against the wrong distribution.

## Testing

Added `test_aux_loss_*_routing_map_matches_dispatch` cases asserting the aux map equals the dispatch map (and diverges when the relevant feature is ignored), parametrized across `softmax`/`sigmoid`/`sqrtsoftplus` and grouped/non-grouped routing, with and without expert bias.

Ran on a GPU node via `torch.distributed.run`:

```
10 passed in 2.21s
```


Connor-XY:
/ok to test [8e432e2](https://github.com/NVIDIA/Megatron-LM/pull/5195/commits/8e432e284e2974b2fefd91091d0996a5e97dbf97)

Connor-XY:
/ok to test [101fc2e](https://github.com/NVIDIA/Megatron-LM/pull/5195/commits/101fc2efc58acd7ab2bc39cb37cb71a76e3235ca)

Victarry:
I think this reverses an existing intentional contract rather than fixing a bug.

For expert_bias and node-/group-limited routing, aux-loss routing_map is supposed to be recomputed from the aux-loss scores, not copied/aligned to the actual dispatch route. This was discussed with Abhinav @aklife97  and landed as [ADLR/megatron-lm!3279 (`cfe7b066c`, "Fix MoE Aux loss")](https://github.com/NVIDIA/Megatron-LM/commit/cfe7b066c04299525dcdf85b687ebc09de88cf3a)

Connor-XY:
Closing this after re-checking the intended aux-loss contract. Victarry's review pointed to cfe7b066, which intentionally recomputes the aux-loss routing map from aux-loss scores instead of aligning it to dispatch for expert bias or group-limited routing. After backing out the behavior change locally, the remaining net diff would only document and test that existing contract, so this is not the right PR to keep as a bug fix for #2767.

PR #1662, opened 2025-07-02
Speed up model parallel initialization

alexqdh:
The process groups are created on all ranks regardless of whether the current rank was actually part of that group. This is extremely time-consuming, especially during large-scale distributed training. By default, in torch groups should be created in the same order in all processes. 

Parameter `use_local_synchronization=True` is added to all `create_group()` calls to enable local synchronization optimization, which can further improve performance during group creation. After modification, process groups are now only created when the current rank is actually part of the group `(if rank in ranks:)`.

PR #5214, opened 2026-06-08
Speed up model parallel initialization

alexqdh:
Continues the rebase work from #1662

The process groups are created on all ranks regardless of whether the current rank was actually part of that group. This is extremely time-consuming, especially during large-scale distributed training. By default, in torch groups should be created in the same order in all processes.

Parameter `use_local_synchronization=True` is added to all `create_group()` calls to enable local synchronization optimization, which can further improve performance during group creation. After modification, process groups are now only created when the current rank is actually part of the group (`if rank in ranks:`).


Victarry:
https://docs.pytorch.org/docs/main/distributed.html#torch.distributed.new_group

There is a requirement from pytorch side:
> This function requires that all processes in the main group (i.e. all processes that are part of the distributed job) enter this function, even if they are not going to be members of the group. Additionally, groups should be created in the same order in all processes.

Could you please check if there are any potentials risks when we use `use_local_synchronization`?

alexqdh:
> https://docs.pytorch.org/docs/main/distributed.html#torch.distributed.new_group
> 
> There is a requirement from pytorch side:
> 
> > This function requires that all processes in the main group (i.e. all processes that are part of the distributed job) enter this function, even if they are not going to be members of the group. Additionally, groups should be created in the same order in all processes.
> 
> Could you please check if there are any potentials risks when we use `use_local_synchronization`?

Actually, this requirement is the default behavior when creating a new group (`use_local_synchronization=False`). When we set `use_local_synchronization=True`, it allows new groups to be created without requiring them to be in the same order across ranks. This results in a significant speedup, especially with a large number of ranks.

guihong-nv:
/ok to test [5a0c348](https://github.com/NVIDIA/Megatron-LM/pull/5214/commits/5a0c348ec17068d574546b32ad98aae805c65420)

deepakn94:
What PyTorch version introduced this new kwarg?

alexqdh:
> What PyTorch version introduced this new kwarg?

@deepakn94  It looks like it was first present in the PyTorch 2.1 documentation: https://docs.pytorch.org/docs/2.1/distributed.html#torch.distributed.new_group

alexqdh:
@deepakn94 Could you help review this PR when you have time, thanks

guihong-nv:
@alexqdh Please resolve the failing tests.

alexqdh:
@guihong-nv  The failing tests have been fixed. Could you help retrigger the tests?

alexqdh:
@Phlip79 Fix some code style issues, please help retrigger the tests, thx

YangFei1990:
When reading the PT doc:
> N.B. use_local_synchronization doesn’t work with MPI.

> N.B. While use_local_synchronization=True can be significantly faster with larger clusters and small process groups, care must be taken since it changes cluster behavior as non-member ranks don’t join the group barrier().

> N.B. use_local_synchronization=True can lead to deadlocks when each rank creates multiple overlapping process groups. To avoid that, make sure all ranks follow the same global creation order.

I think there are some implications, we should not suddenly set it to be True as default behavior. Maybe add a config to trigger it?

alexqdh:
> When reading the PT doc:
> 
> > N.B. use_local_synchronization doesn’t work with MPI.
> 
> > N.B. While use_local_synchronization=True can be significantly faster with larger clusters and small process groups, care must be taken since it changes cluster behavior as non-member ranks don’t join the group barrier().
> 
> > N.B. use_local_synchronization=True can lead to deadlocks when each rank creates multiple overlapping process groups. To avoid that, make sure all ranks follow the same global creation order.
> 
> I think there are some implications, we should not suddenly set it to be True as default behavior. Maybe add a config to trigger it?

@YangFei1990  Thanks, I have updated the PR to add `use_local_synchronization` config, please review

PR #5220, opened 2026-06-08
MoE routing analysis and metrics capture

mathemakitten:

# What does this PR do ?
* Adds a `RouterTracer` class that captures per-layer top-K routing decisions via forward hooks on TopKRouter modules, writing one JSONL record per (step, layer, rank). Works identically for training and inference.
* Adds `tools/moe_routing/` with two analysis scripts that work on traces from either training or inference:
  - `analyze_routing_concentration.py` — per-layer expert activation frequency, top-N coverage curves, and a heatmap of hot experts vs layers.
  - `analyze_routing_predictability.py` — applies layer L's router weights to L_prev's hidden states and measures cosine/Spearman similarity between predicted and actual per-expert token-count distributions. High scores here are the affirmative signal that distributional routing is predictable one layer ahead.

**Notes on usability**
- Use --moe-routing-trace-max-iters / --moe-routing-trace-max-steps to bound the capture window; the tracer self-disables and removes its hooks after the limit is reached
- Python forward hooks do not fire during CUDA graph replay. Run with `--cuda-graph-impl none` when collecting traces.
- `analyze_routing_predictability.py` requires `--moe-routing-trace-capture-hidden-states` and `--moe-routing-trace-dump-weights` at collection time.

**Follow-up items**
*  Unify hook registration helpers: `megatron/training/activation_logging.py` has a generic `_register_hooks(model, module_types, hook_factory)` utility that walks model chunks and registers forward hooks. `RouterTracer.register_hooks` does the same thing with a loop. These can be consolidated with a shared helper into `core/`, but requires some more work to unify the hook registration pattern (`_register_hooks` in `activation_logging.py` does not currently live in core).

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


megatron/core/transformer/moe/router_trace.py
```diff
@@ -0,0 +1,379 @@
...
+import os
+import threading
+from typing import List, Optional
+
+import torch
+
+_TRACER: Optional["RouterTracer"] = None
+
+
+def init_tracer(
```
santhnm2: Can we make these function names (`init_tracer` and `get_tracer`) more specific to MoE routing? For example `init_moe_router_tracer` / `get_moe_router_tracer`. Same thing with global `_TRACER`

examples/inference/advanced/gpt_dynamic_inference_with_coordinator.py
```diff
@@ -225,8 +226,23 @@ async def main(
...
+                max_steps=max_steps,
+                rank=rank,
+                capture_hidden_states=getattr(args, 'moe_routing_trace_capture_hidden_states', False),
+                capture_logits=getattr(args, 'moe_routing_trace_capture_logits', False),
+                dump_router_weights=getattr(args, 'moe_routing_trace_dump_weights', False),
+            )
+
         model = get_model_for_inference()
 
+        if get_tracer() is not None:
```
santhnm2: Nit: could use walrus operator here to avoid calling `get_tracer()` twice
```
if (tracer := get_tracer()) is not None:
    tracer.register_hooks(model)
```

megatron/core/transformer/moe/router_trace.py
```diff
@@ -0,0 +1,379 @@
...
+    """
+    if "hs_offset" not in record:
+        raise ValueError("Record does not contain hidden-state metadata.")
+    path = os.path.join(trace_dir, f"hidden_states_rank{record['rank']}.bin")
+    with open(path, "rb") as f:
+        f.seek(record["hs_offset"])
+        data = f.read(record["hs_bytes"])
+    arr = torch.frombuffer(bytearray(data), dtype=torch.int16)
+    arr = arr.view(torch.bfloat16).reshape(record["hs_shape"])
+    return arr.clone()
```
santhnm2: Why is the `clone` necessary? Same thing for `load_logits_for_record`

megatron/core/transformer/moe/router_trace.py
```diff
@@ -0,0 +1,379 @@
...
+def _unwrap_model(model):
+    """Unwrap DDP/FSDP wrappers to get the underlying module."""
+    try:
+        from megatron.core.utils import unwrap_model
+
+        return unwrap_model(model)
+    except Exception:
+        while hasattr(model, "module"):
+            model = model.module
+        return model
```
santhnm2: Why do we need this? Isn't the existing `unwrap_model` sufficient?

megatron/inference/utils.py
```diff
@@ -285,6 +286,15 @@ def add_inference_args(parser: ArgumentParser) -> ArgumentParser:
...
     )
+
+    group.add_argument(
+        "--moe-routing-trace-max-steps",
+        type=int,
+        default=None,
+        help="Maximum number of decode steps to trace (inference). Default is unlimited."
+             "Training uses --moe-routing-trace-max-iters instead.",
+    )
+
```
santhnm2, line 326: Maybe we should make the training / inference argument split more explicit? Otherwise `iters` and `steps` seem pretty close semantically / could be easily confused. So `--moe-routing-trace-max-steps` -> `moe-routing-trace-max-inference-steps` and `--moe-routing-trace-max-iters` -> `--moe-routing-trace-max-training-iters`
Or if `--moe-routing-trace-max-iters` is already in use for training maybe we could just do the inference name change

santhnm2 (approved):

megatron/core/transformer/moe/router_trace.py
```diff
@@ -0,0 +1,503 @@
...
+        dump_router_weights: bool = False,
+    ) -> None:
+        self.output_dir = output_dir
+        self.max_steps = max_steps
+        self.rank = rank
+        self.training_mode = training_mode
+        self.step_id = 0
+        self.layers_seen_this_step: set[int] = set()
+        self.records: list[dict] = []
+        self._lock = threading.Lock()
```
Mellonta: What's the purpose of the lock here, is there any contention between different threads during inference?

megatron/core/transformer/moe/router_trace.py
```diff
@@ -0,0 +1,503 @@
...
+        def hook(module, inputs, outputs):
+            if self._stopped:
+                return
+            if torch.cuda.is_current_stream_capturing():
+                return
+            self._record(module, inputs, outputs, identity)
+
+        return hook
+
+    def _extract_hidden_state(self, inputs, expected_num_tokens):
```
Mellonta, line 274: Could you add an one-line comment here explaining what it does?

tools/moe_routing/analyze_routing.py
```diff
@@ -0,0 +1,93 @@
...
+    # 1. Expert concentration (hot-set size).
+    run(
+        "analyze_routing_concentration.py",
+        [args.trace_dir] + nexpert_args + outdir_args,
+        "Expert concentration  (hot-set size, routing distribution)",
+    )
+
+    # 2. Distribution predictability: how well do L_prev's hidden states predict L's routing?
+    run(
+        "analyze_routing_predictability.py",
```
yaox12, line 86: **[IMPORTANT Correctness]** The dispatcher calls `analyze_routing_predictability.py`, but that file is not added in this PR.

The documented top-level analysis command will always fail or warn on the second analysis, and the README advertises predictability output that users cannot actually run. Please either add the missing script in this PR, or remove the dispatcher call and README references until the predictability analysis lands.

tools/moe_routing/analyze_routing_concentration.py
```diff
@@ -0,0 +1,272 @@
...
+    print(header)
+    print("  " + "-" * (len(header) - 2))
+
+    per_layer_results = []
+    for layer in layers:
+        freqs = per_layer_freq[layer]
+        ntoken = per_layer_tokens[layer]
+        unique_experts = len(freqs)
+        cov = topk_coverage(freqs, n_values)
+        uniform_bl = 22 / args.num_experts  # random baseline at N=22
```
yaox12: **[IMPORTANT Correctness]** The concentration analysis hard-codes router top-k as `22`.

Models with `moe_router_topk != 22` get incorrect baselines and heatmap rates, even though each trace record already includes `topk`. Please read `topk` from the trace records, or add/pass a `--top-k` argument and use it consistently here and in the heatmap normalization below.

tools/moe_routing/analyze_routing_concentration.py
```diff
@@ -0,0 +1,272 @@
...
+    pattern = os.path.join(trace_dir, "router_trace_rank*.jsonl")
+    paths = sorted(glob.glob(pattern))
+    if not paths:
+        raise FileNotFoundError(f"No trace files matching {pattern}")
+    for path in paths:
+        with open(path) as f:
+            for line in f:
+                r = json.loads(line)
+                yield (
+                    r["rank"], r["step"], r["layer"],
```
yaox12: **[IMPORTANT Correctness]** The loader drops `block` and `mtp_idx`, then the analysis aggregates only by integer `layer`.

Decoder layer 0 and MTP layer 0 can collide, even though `router_trace.py` writes `mtp_idx` specifically to avoid that ambiguity. That makes the reported per-layer concentration silently merge distinct routers. Please preserve a full identity such as `(block, mtp_idx, layer)` throughout the counters and output labels.

megatron/core/transformer/moe/router_trace.py
```diff
@@ -0,0 +1,503 @@
...
+        """
+        with self._lock:
+            if step_id is not None:
+                for rec in self.records:
+                    rec["step"] = step_id
+                self.step_id = step_id
+            self._flush_records_to_disk()
+            self.step_id += 1
+            self.layers_seen_this_step.clear()
+            if self.step_id >= self.max_steps:
```
yaox12: **[IMPORTANT Correctness]** `max_steps` is compared against `step_id`, but training passes the absolute iteration number into `advance_step(iteration)`.

On checkpoint resume, `--moe-routing-trace-max-training-iters 500` can stop immediately if the resumed iteration is already >= 500, instead of tracing 500 iterations as documented. Please track a separate captured-step counter, or compute an absolute stop iteration from `args.iteration + moe_routing_trace_max_training_iters`.

megatron/core/transformer/moe/router_trace.py
```diff
@@ -0,0 +1,503 @@
...
+                    if self.step_id >= self.max_steps:
+                        self._stopped = True
+                        return
+                self.layers_seen_this_step.add(layer_key)
+
+            # The router-weight dump is keyed by the integer `layer` to match the
+            # JSONL records' `layer` field, which is how downstream analysis joins
+            # weights to traces. (`layers_seen_this_step` above uses the full
+            # `layer_key` only for collision-free step-boundary detection.)
+            if self.dump_router_weights and layer not in self._router_state:
```
yaox12: **[IMPORTANT Correctness]** Router weight sidecars are keyed only by integer `layer`.

Decoder/MTP routers, or multiple MTP heads with the same inner layer, can overwrite or suppress each other here. A predictability analysis would then join records to the wrong router weights. Please key router state with the same full identity used in records, e.g. `(block, mtp_idx, layer)`, or an equivalent string key.

yaox12 (commented):
Post some comments from codex.

yaox12 (approved):

Mellonta (approved):

Phlip79 (changes requested):
Can you please add tests?

Phlip79 (approved):

cuichenx (approved):

megatron/training/config/training_config.py
maanug-nv, line 1: wouldn't `LoggerConfig` make more sense for these settings?

kvareddy (approved):

megatron/training/config/training_config.py
mathemakitten: Yes, moved.

deepakn94 (approved):

PR #5227, opened 2026-06-09
Fix avoid memory leak when open recompute

yuyang66:
- [✅] I, the PR author, have personally reviewed every line of this PR.

# What does this PR do ?

I ran into an issue where gradients kept growing during the backward pass after enabling recompute. 
After debugging, I traced the root cause to `DeepepManager.token_probs` not being properly released. This caused the corresponding leaf tensor `hidden_states` (with `retain_grad()` applied) and its `hidden_states.grad` to remain unfreed. 
The issue was resolved by assigning `None` to `self.token_probs` after the `combine` step.

Clipboard_Screenshot_1780988475

After fixed:
Clipboard_Screenshot_1780988626


yaox12 (approved):

YangFei1990:
> LGTM — the fix correctly releases `token_indices` and `token_probs` in `_DeepepManager.combine()`, matching the existing pattern for `dispatched_indices`/`dispatched_probs`.
> 
> **Non-blocking note:** `_HybridEPManager.combine()` (line ~1120) has a similar pattern — it stores `self.token_probs` and `self.dispatched_probs` during `setup_metadata`/`dispatch` but does not release them in `combine()`. Might be worth a follow-up to apply the same cleanup there.

@yuyang66 Thanks for the PR. Could you also address the same pattern for the HybridEP combine?

kvareddy (approved):

YangFei1990:
@yuyang66 Are you running with cuda graphs? Actually after checking the code these are registered as `cudagraph_attrs `: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/moe/token_dispatcher.py#L1430, cuda graph will use them as outputs and always re-register them to the dispatcher after mlp post-process. Technically speaking deallocate is not harmful, but it would be save to keep them when partial cuda graph is enabled.

guihong-nv:
@yuyang66 May you please reply to the comments above from @YangFei1990 ?

PR #5244, opened 2026-06-09
[split 2/4] Scale DSA indexer loss in pipeline schedules

HollowMan6:
Split out from #3674 to reduce review scope.

Original changes by @HollowMan6 in #3674.

## Scope
- Generalize auxiliary loss-scale handling in pipeline schedules.
- Wire DSA indexer loss scaling through the experimental attention variant hook.
- Update hybrid CP schedule calls after the `backward_step` signature change.
- Add unit coverage for CP scaling, dict outputs, and non-mutating default DSA hook behavior.

## Dependencies
None.

## Validation
- `git diff --check upstream/main..split-3674-dsa-loss-scale`

### DIFF
--- a/megatron/core/pipeline_parallel/hybrid_cp_schedule.py
+++ b/megatron/core/pipeline_parallel/hybrid_cp_schedule.py
@@ -545,9 +545,15 @@ def hybrid_context_parallel_forward_backward(
             )
             sample["local_cp_size"] = torch.tensor(partner_cp_size, dtype=torch.int32)
             new_data_iterator = RerunDataIterator(iter([sample]))
-            return new_data_iterator
         else:
-            return None
+            partner_cp_size = 0
+            new_data_iterator = None
+
+        partner_cp_size_tensor = torch.tensor(
+            [partner_cp_size], dtype=torch.int32, device=torch.cuda.current_device()
+        )
+        _broadcast(partner_cp_size_tensor)
+        return new_data_iterator, int(partner_cp_size_tensor.item())
 
     # We get data once per global batch and schedule the sub-samples.
     # TODO(pmannan): Should we wrap the data_iterator here instead of the training.py file?
@@ -579,7 +585,7 @@ def hybrid_context_parallel_forward_backward(
             sample_ids_this_group = sample_id_groups[j][hdp_rank] if is_first_tp_rank else None
             for i in range(num_samples_this_group[j]):
                 # Call forward step for each sub-sample
-                new_data_iterator = _get_new_data_iterator(i, j)
+                new_data_iterator, cp_group_size = _get_new_data_iterator(i, j)
                 # TODO: Find the usage of current_microbatch and is_first_microbatch and
                 # how that may affect my usage.
                 output_tensor, num_tokens = forward_step(
@@ -590,7 +596,8 @@ def hybrid_context_parallel_forward_backward(
                     input_tensor,
                     forward_data_store,
                     config,
-                    collect_non_loss_data,
+                    cp_group_size=cp_group_size,
+                    collect_non_loss_data=collect_non_loss_data,
                     is_first_microbatch=check_first_val_step(
                         first_val_step, forward_only, current_microbatch == 0
                     ),
@@ -599,9 +606,7 @@ def hybrid_context_parallel_forward_backward(
                 current_microbatch += 1
                 total_num_tokens += num_tokens.item()
                 if not forward_only:
-                    backward_step(
-                        input_tensor, output_tensor, output_tensor_grad, model_type, config
-                    )
+                    backward_step(input_tensor, output_tensor, output_tensor_grad, config)
 
             # Create a barrier at end of each group.
             # This barrier ensures that all ranks are prepared to change assigned CP group sizes and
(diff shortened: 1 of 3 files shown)


megatron/core/pipeline_parallel/schedules.py
```diff
@@ -341,17 +355,20 @@ def forward_step_calc_loss(
...
+        device = get_tensor_device(output_tensor)
+        loss_scale = _compute_loss_scale(config, device)
         if config.calculate_per_token_loss:
-            DSAIndexerLossAutoScaler.set_loss_scale(loss_scale)
+            experimental_attention_variant_loss_scale_func(loss_scale)
         else:
-            DSAIndexerLossAutoScaler.set_loss_scale(loss_scale / num_microbatches)
+            cp_size_for_scaling = cp_group_size if cp_group_size is not None else 1
+            experimental_attention_variant_loss_scale_func(
+                loss_scale * cp_size_for_scaling / num_microbatches
```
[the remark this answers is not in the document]
HollowMan6: Fixed in `8c9432387`. `hybrid_context_parallel_forward_backward` now passes `cp_group_size` and `collect_non_loss_data` by keyword, and broadcasts the per-sub-sample local CP size from TP rank 0 before calling `forward_step`, so `collect_non_loss_data=False` can no longer be interpreted as CP size 0.

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -344,17 +363,20 @@ def forward_step_calc_loss(
...
+    )
+    if experimental_attention_variant_loss_scale_func is not None:
+        device = get_tensor_device(output_tensor)
+        loss_scale = _compute_loss_scale(config, device)
         if config.calculate_per_token_loss:
-            DSAIndexerLossAutoScaler.set_loss_scale(loss_scale)
+            experimental_attention_variant_loss_scale_func(loss_scale)
         else:
-            DSAIndexerLossAutoScaler.set_loss_scale(loss_scale / num_microbatches)
+            cp_size_for_scaling = cp_group_size if cp_group_size is not None else 1
```
yashaswikarnati, line 380: with dynamic cp - do we change the cp_group_size for each microbatch ? the way we are setting the loss scale might lead to unintended behaviour - like when loss scale computed on one microbatch can be used for the next immediate backward of different microbatch when pipelining.. 
also @cc: @parthmannan

HollowMan6: Good point! Dynamic CP can change `cp_group_size` per sub-sample, but today hybrid CP is not used with pipeline parallelism and requires `calculate_per_token_loss`, so this non-per-token CP-scaled path is only reached with static CP.

So I don’t think this causes an immediate overwrite issue in current supported configs. But agreed the class-level scaler is fragile if dynamic CP + pipelining/non-per-token loss is enabled later; the robust fix would be to store the scale per autograd context/microbatch instead of globally, but that can be left for future dynamic CP work.

I will add a comment in the code for this.

yashaswikarnati: Thanks for the note!  lgtm. Also, i'm hoping we already have an assert for dynamic cp to only use the per token loss scaling.

yashaswikarnati (approved):

megatron/core/pipeline_parallel/schedules.py
HollowMan6: Thank you for reviewing!

> Also, i'm hoping we already have an assert for dynamic cp to only use the per token loss scaling.

Yes, we already have that assert at
https://github.com/NVIDIA/Megatron-LM/blob/a12484bd1961078b27a0d0241639fe78903cea30/megatron/training/arguments.py#L1381

megatron/core/pipeline_parallel/hybrid_cp_schedule.py
```diff
@@ -545,9 +545,15 @@ def _get_new_data_iterator(sample_id_in_group, group_id):
...
             sample["local_cp_size"] = torch.tensor(partner_cp_size, dtype=torch.int32)
             new_data_iterator = RerunDataIterator(iter([sample]))
-            return new_data_iterator
         else:
-            return None
+            partner_cp_size = 0
+            new_data_iterator = None
+
+        partner_cp_size_tensor = torch.tensor(
+            [partner_cp_size], dtype=torch.int32, device=torch.cuda.current_device()
```
Phlip79, line 555: Can you add a comment why this is hardcoded to int32?

Phlip79 (approved):
Approving on condition of adding comment

megatron/core/pipeline_parallel/hybrid_cp_schedule.py
HollowMan6: Thank you for reviewing @Phlip79 ! Comments added. https://github.com/NVIDIA/Megatron-LM/pull/5244/commits/58c5beebda37f14e409c1eb87fd3b295219a41e9

yaoyu-33 (approved):

Mellonta (approved):

kvareddy (approved):

PR #5245, opened 2026-06-09
[split 3/4] Refactor absorbed MLA projection handling

HollowMan6:
Split out from #3674 to reduce review scope.

Original changes by @HollowMan6 in #3674.

## Scope
- Switch absorbed MLA to a combined KV up-projection layout.
- Add helper handling for packed THD output rank restoration and optional core-attention V up-projection consumption.
- Pass CP communication metadata into absorbed MLA core attention.
- Add focused absorbed MLA tests for checkpoint metadata and updated projection-gradient parity.

## Dependencies
None.

## Validation
- `git diff --check upstream/main..split-3674-absorbed-mla`

PR #5049, opened 2026-05-29
[2/2] Add TileLang fused DSA kernels support with THD and CP & Clean up

HollowMan6:
# What does this PR do ?

Closes #5703 

This PR adds the TileLang implementation of the fused DSA operations.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

## Main changes

- Adds TileLang kernels for DSA indexer forward and backward.
- Adds TileLang kernels for absorbed sparse-MLA forward and backward.
- Adds Python wrapper code under `megatron/core/transformer/experimental_attention_variant/ops/`.

## Packed THD and CP behavior

The TileLang path uses the metadata and bounds produced by PR 1. When the current TileLang kernel path cannot safely handle a packed THD or CP case, DSA falls back to the unfused implementation instead of changing semantics. This keeps packed-sequence and context-parallel correctness owned by the common DSA path while allowing supported dense cases to use TileLang.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
        C1[Expert Review] --> C2[Final Review]
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [X] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [X] I have added relevant unit tests
- [X] I have added relevant functional tests
- [X] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [X] I have added relevant documentation
- [X] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

The following process is enforced via the CODEOWNERS file for changes into `megatron/core`. For changes outside of `megatron/core`, it is up to the PR author whether or not to tag the Final Reviewer team.

<details>
<summary>For MRs into `main` branch</summary>

Feel free to message or comment the @mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

#### (Step 2): Collect the expert reviewers reviews

1. Attach the `Expert Review` label when your PR is ready for review.
2. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

:warning: Only proceed to the next step once all reviewers have approved, merge-conflict are resolved and the CI is passing.  
[...]

PR #5246, opened 2026-06-09
[split 4/4] Enable DSA CP and THD hooks

HollowMan6:
Split out from #3674 to reduce review scope.

Original changes by @HollowMan6 in #3674.

## Scope
- Refactor DSA around backend-neutral fused-kernel hooks.
- Add DSA layout and masking helpers for allgather CP and packed THD.
- Add optional DSA kernel backend config and DSA indexer compatibility knobs.
- Remove the DSA-specific MLA call path from generic `multi_latent_attention.py`.
- Wire GPT DSA specs to absorbed MLA and add DSA-focused unit coverage.

## Dependencies
- Depends on #5243 for the packed THD RoPE CP fix.
- #5244 for DSA indexer loss-scale scheduling (can be merged separately).
- Depends on #5245 for the absorbed MLA interface used by the spec/tests.

## CODEOWNERS note
`megatron/training/arguments.py` has an exact CODEOWNERS entry with no owners after the general `megatron/training/` rule, so this PR includes that one-line skip-field update here and calls out the ownership ambiguity explicitly.

## Validation
- `git diff --check upstream/main..split-3674-dsa-cp-thd-core`


HollowMan6:
@codex review

megatron/core/models/gpt/experimental_attention_variant_module_specs.py
```diff
@@ -109,9 +109,9 @@ def get_dsa_module_spec_for_backend(
     )
 
     attention = ModuleSpec(
-        module=MLASelfAttention,
+        module=AbsorbedMLASelfAttention,
         params={"attn_mask_type": AttnMaskType.causal},
```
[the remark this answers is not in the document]
HollowMan6: This is covered by the explicit draft-stack dependency on #5245. The absorbed MLA wrapper in #5245 passes `up_v_weight` into core attention and sets up the consumed-projection path; #5246 is intended to merge after #5245, which is why this branch alone appears to miss the argument.

tests/unit_tests/transformer/experimental_attention_variant/test_attention_variant_dsa.py
```diff
@@ -10,34 +10,51 @@
...
+from megatron.core.packed_seq_params import PackedSeqParams
 from megatron.core.process_groups_config import ProcessGroupCollection
 from megatron.core.tensor_parallel.random import model_parallel_cuda_manual_seed
-from megatron.core.transformer import TransformerConfig
 from megatron.core.transformer.enums import AttnMaskType
+from megatron.core.transformer.experimental_attention_variant import dsa_kernels
+from megatron.core.transformer.experimental_attention_variant.absorbed_mla import (
+    AbsorbedMLASelfAttention,
+    _apply_absorbed_v_up_projection,
+    _restore_packed_thd_batch_dim,
```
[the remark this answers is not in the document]
HollowMan6: This is also a draft-stack dependency on #5245. The imported helpers are introduced in #5245 and #5246 is intended to merge after it. I kept the dependency explicit in the PR body rather than duplicating the absorbed MLA helper changes in this DSA-core split.

megatron/core/models/gpt/experimental_attention_variant_module_specs.py
```diff
@@ -109,9 +109,9 @@ def get_dsa_module_spec_for_backend(
     )
 
     attention = ModuleSpec(
-        module=MLASelfAttention,
+        module=AbsorbedMLASelfAttention,
```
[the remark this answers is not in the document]
HollowMan6: This is covered by the explicit draft-stack dependency on https://github.com/NVIDIA/Megatron-LM/pull/5245. The absorbed MLA wrapper in https://github.com/NVIDIA/Megatron-LM/pull/5245 passes up_v_weight into core attention and sets up the consumed-projection path; https://github.com/NVIDIA/Megatron-LM/pull/5246 is intended to merge after https://github.com/NVIDIA/Megatron-LM/pull/5245, which is why this branch alone appears to miss the argument.

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -27,6 +32,204 @@
...
+            raise RuntimeError(
+                "Absorbed DSAttention expects value=None (latent path). "
+                "Received absorbed layout with explicit value tensor."
+            )
+        output = None
+        if dsa_kernels.use_fused_dsa_kernels(config):
+            output = dsa_kernels.run_fused_absorbed_sparse_attention(
+                config, query, key, topk_indices, softmax_scale, latent_v_channels
+            )
+        if output is None:
```
santhnm2, line 158: To confirm, the reason for this condition being `if output is None` rather than just an `else` is because `run_fused_absorbed_sparse_attention` could return `None`? If so can you add a comment explaining why this could happen?

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -27,6 +32,204 @@
...
+                key,
+                topk_indices,
+                softmax_scale,
+                latent_v_channels,
+                mask=mask,
+                varlen_starts=varlen_starts,
+                varlen_ends=varlen_ends,
+                key_positions=key_positions,
+            )
+        output = torch.einsum("sbhc,hdc->sbhd", output, up_v_weight).contiguous()
```
santhnm2, line 171: Maybe add an assert that `output is not None` before computing the einsum

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -421,20 +776,35 @@ def bwd_fused_indexer_loss_naive(
     # L1 normalize
     attention_scores_normalized = attention_scores_sum / attention_scores_sum.sum(
         dim=-1, keepdim=True
-    )
+    ).clamp_min(1e-10)
```
santhnm2, line 799: Can you add a comment explaining why this clamp is necessary?

megatron/core/transformer/transformer_config.py
santhnm2, line 1: These library check functions should perhaps go in `megatron/core/utils.py` since we have similar helper functions there already, e.g. for the Mamba kernels

megatron/core/models/gpt/experimental_attention_variant_module_specs.py
santhnm2, line 1: Are the `MLA` -> `AbsorbedMLA` changes backwards compatible with existing checkpoints?

megatron/core/transformer/experimental_attention_variant/dsa.py
HollowMan6: Yes, as Fused backends may decline unsupported shapes or layouts by returning None, so we still keep the absorbed PyTorch path as the fallback. Comment added.

megatron/core/transformer/experimental_attention_variant/dsa.py
HollowMan6: added

megatron/core/transformer/experimental_attention_variant/dsa.py
HollowMan6: Since I found that fully masked packed/varlen rows can have zero summed attention mass, so it would be needed to clamp the denominator so those rows stay finite and are later zeroed by the row-valid loss mask. Comment added

megatron/core/transformer/transformer_config.py
HollowMan6: done

megatron/core/models/gpt/experimental_attention_variant_module_specs.py
HollowMan6: Yes, this is backward compatible with existing checkpoints. The checkpoint keys are derived from the parameter-bearing submodule names in the spec, not from the Python attention class name. AbsorbedMLASelfAttention keeps the same MLA projection/norm submodule names such as linear_q_proj, linear_q_down_proj, linear_q_up_proj, linear_kv_down_proj, linear_kv_up_proj, linear_proj, q_layernorm, and kv_layernorm, so existing MLA checkpoint key names are unchanged.

tests/unit_tests/transformer/experimental_attention_variant/test_attention_variant_dsa.py
```diff
@@ -1,43 +1,64 @@
...
```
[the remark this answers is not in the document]
HollowMan6: fixed

megatron/core/models/gpt/experimental_attention_variant_module_specs.py
```diff
@@ -109,9 +111,9 @@ def get_dsa_module_spec_for_backend(
...
```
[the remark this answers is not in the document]
HollowMan6: rebased, this should not be a issue now

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -1041,84 +1663,380 @@ def forward(
...
+            _validate_nonpacked_cp_uniform_length(
+                sq=sq, skv=key.size(0), cp_size=cp_size, cp_group=cp_group, device=query.device
+            )
+            kv_reorder_idx = dsa_layout.build_zigzag_allgather_cp_key_reorder(
+                sq=sq, cp_size=cp_size, device=query.device
+            )
+
+        if cp_size > 1:
+            assert (
+                self.cp_comm_type == "allgather"
```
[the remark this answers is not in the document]
HollowMan6: This will be overly defensive, as:

TransformerLayer -> passes cp_comm_type into AbsorbedMLASelfAttention
AbsorbedMLASelfAttention -> passes that same cp_comm_type into nested DSAttention

megatron/core/transformer/multi_latent_attention.py
```diff
@@ -398,7 +391,6 @@ def forward(
                         attention_mask,
                         packed_seq_params=packed_seq_params,
                         attn_mask_type=attn_mask_type,
-                        **extra_kwargs,
                     )
```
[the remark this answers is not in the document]
HollowMan6: Addressed

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -27,6 +32,216 @@
...
 
+def is_dsa_skip_topk_layer(layer_number: int, skip_topk_offset: int, topk_freq: int) -> bool:
+    """Return whether a 1-indexed layer reuses a previous DSA top-k result."""
+    if layer_number < 1:
+        raise ValueError(f"layer_number must be 1-indexed and positive, got {layer_number}.")
+    if skip_topk_offset < 0:
+        raise ValueError(f"skip_topk_offset must be non-negative, got {skip_topk_offset}.")
+    if topk_freq < 1:
+        raise ValueError(f"topk_freq must be positive, got {topk_freq}.")
+    return (max(layer_number - skip_topk_offset, 0) % topk_freq) != 0
```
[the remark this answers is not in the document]
HollowMan6: Addressed

megatron/core/models/gpt/experimental_attention_variant_module_specs.py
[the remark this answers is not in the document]
HollowMan6: fixed

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -27,6 +32,219 @@
...
+    index_mask = dsa_masking.apply_sparse_validity_to_index_mask(
+        index_mask,
+        row_mask=row_mask,
+        varlen_starts=varlen_starts,
+        varlen_ends=varlen_ends,
+        key_positions=key_positions,
+    )
+
+    attention_scores += index_mask.unsqueeze(1)
+    attention_scores = torch.nn.functional.softmax(attention_scores, dim=-1, dtype=torch.float32)
```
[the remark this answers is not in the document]
HollowMan6: fixed

megatron/core/transformer/multi_latent_attention.py
```diff
@@ -1432,12 +1420,8 @@ def _clone_sharded_object_with_key(obj: ShardedObject, new_key: str) -> ShardedO
...
         return sharded_state_dict
 
-    def _synthetic_state_dict_key_suffixes(self):
-        """Return source checkpoint keys used to locate this module in a state dict."""
-        return ("linear_q_down_proj.weight",)
-
-    def _synthesize_fused_qkv_down_weight(self, state_dict, prefix):
-        """Materialize fused qkv-down weight from old separate q/kv checkpoint keys."""
+    def _load_from_state_dict(self, state_dict, prefix, *args, **kwargs):
+        """Load state dict with automatic unfused->fused conversion."""
```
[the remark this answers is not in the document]
HollowMan6: addressed

megatron/core/transformer/multi_latent_attention.py
```diff
@@ -1384,11 +1375,8 @@ def _clone_sharded_object_with_key(obj: ShardedObject, new_key: str) -> ShardedO
                 sharded_state_dict[q_extra_key] = fused_obj
                 sharded_state_dict[kv_extra_key] = fused_obj
 
-        # Keep fused layernorm params so TransformerLayer's key map can load old
-        # input_layernorm checkpoints into the fused TE down-proj module.
         for key in list(sharded_state_dict.keys()):
-            suffix = key[len(fused_prefix) :] if key.startswith(fused_prefix) else ""
-            if key.startswith(fused_prefix) and not suffix.startswith("layer_norm_"):
+            if key.startswith(fused_prefix):
                 del sharded_state_dict[key]
```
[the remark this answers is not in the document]
HollowMan6: addressed

megatron/core/transformer/multi_latent_attention.py
```diff
@@ -1337,17 +1339,6 @@ def _qkv_down_projection(self, hidden_states):
...
-        self.linear_kv_up_proj.backward_dw()
-        self.linear_qkv_down_proj.backward_dw()
-        self.linear_q_up_proj.backward_dw()
-        self._backward_output_proj()
-
-    def set_for_recompute_input_layernorm(self):
-        """Set the attention layer for recompute input_layernorm. Only needed for fp8/fp4."""
-        set_save_original_input(self.linear_qkv_down_proj)
-
     def sharded_state_dict(self, prefix: str = "", sharded_offsets: tuple = (), metadata=None):
```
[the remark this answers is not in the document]
HollowMan6: addressed

megatron/core/transformer/experimental_attention_variant/dsa_kernels.py
```diff
@@ -0,0 +1,227 @@
...
+
+if TYPE_CHECKING:
+    from megatron.core.packed_seq_params import PackedSeqParams
+    from megatron.core.process_groups_config import ProcessGroupCollection
+    from megatron.core.transformer.transformer_config import TransformerConfig
+
+_BACKEND_MODULE_NAME_BY_BACKEND = {
+    "tilelang": "megatron.core.transformer.experimental_attention_variant.dsa_tilelang_kernels",
+    "cudnn": "megatron.core.transformer.experimental_attention_variant.dsa_cudnn_kernels",
+}
```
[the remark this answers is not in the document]
HollowMan6: Those will be in #5049 and #5099

megatron/core/transformer/attention.py
santhnm2, line 1: Can you explain why the changes in this file are necessary? In particular, why do we need to manually construct this `checkpoint_inputs` list and specifically overwrite the `attn_mask_type` value?

HollowMan6: Added code comments for the explanation. Some more details here:

The change is needed for activation checkpointing (recompute) when the core attention needs extra tensor inputs, specifically the DSA path in this case. In the DSA/MLA flow, core_attention_extra_kwargs can contain tensors like x=hidden_states and qr=q_compressed. Those tensors are used inside core_attention and must be explicit inputs to tensor_parallel.checkpoint; otherwise they are captured from the Python closure, which can either defeat checkpointing by retaining the original forward tensors or give incorrect/backward-fragile behavior because checkpoint recompute only detaches and returns gradients for tensors passed as checkpoint inputs.

attn_mask_type is handled specially because it is metadata semantically, but Megatron’s checkpoint wrapper saves its *args via ctx.save_for_backward(*args), so the checkpoint argument list must contain tensors, not an AttnMaskType enum. We preserve the existing behavior from main by encoding the enum value as a small tensor before calling checkpoint, then converting it back inside custom_forward with AttnMaskType(attn_mask_type.item()).

The overwrite of checkpoint_inputs[5] is just because the list is initially constructed with a placeholder before defaulting attn_mask_type and converting it to the tensor form. It is not changing the mask type; it replaces None/enum metadata with the tensor representation required by checkpointing.

megatron/core/transformer/attention.py
```diff
@@ -443,6 +455,9 @@ def custom_forward(*inputs):
             attention_mask = inputs[3]
             attn_mask_type = inputs[5]
             attn_mask_type = AttnMaskType(attn_mask_type.item())
+            extra_kwargs = dict(core_attention_extra_kwargs)
+            for name, kwarg_value in zip(tensor_kwarg_names, inputs[6:]):
```
kvareddy: Let's not have [6:]. This is error-prone. Can you refactor this code?

HollowMan6: Thank you @kvareddy for reviewing, this is done in https://github.com/NVIDIA/Megatron-LM/pull/5246/commits/8427cf32b641e83a3386394dbc4e07c7c6e1e9d4. Please let me know if you have any other comments!

megatron/core/transformer/experimental_attention_variant/dsa_layout.py
```diff
@@ -0,0 +1,283 @@
...
+    seq_ends = cu_seqlens_i64[1:]
+    seq_lens = seq_ends - seq_starts
+    nonzero = seq_lens > 0
+    seq_starts = seq_starts[nonzero]
+    seq_ends = seq_ends[nonzero]
+    seq_lens = seq_lens[nonzero]
+    if seq_lens.numel() == 0:
+        return torch.empty(0, dtype=torch.int64, device=device)
+
+    if cu_seqlens_i64.device.type == "cpu":
```
Mellonta, line 164: What does this block do? IIUC the sequence lengths are only known at runtime and vary per batch, so divisibility can only really be guaranteed by the data pipeline's padding and this block is a no-op during training, is that right?

megatron/core/transformer/experimental_attention_variant/dsa_masking.py
```diff
@@ -0,0 +1,505 @@
...
+    "normalize_query_valid_rows",
+    "normalize_varlen_bounds",
+    "prepare_additive_mask",
+    "prepare_sparse_mask_context",
+    "scatter_topk_into_index_mask",
+]
+
+
+def build_causal_mask_from_positions(
+    query_pos: torch.Tensor, key_pos: torch.Tensor
```
Mellonta, line 35: Q: it looks like key_pos is always an arange, is there a use case where key_pos is shuffled?

tests/unit_tests/transformer/experimental_attention_variant/test_attention_variant_dsa.py
```diff
@@ -61,6 +323,902 @@ def patch_hadamard_if_needed():
...
+        gathered_key_pos = torch.tensor(
+            [0, 1, 6, 7, 8, 9, 14, 15, 2, 3, 4, 5, 10, 11, 12, 13], dtype=torch.int64
+        )
+        restored = gathered_key_pos.index_select(0, key_reorder_idx)
+        assert restored.tolist() == list(range(16))
+
+    def test_cp_packed_zigzag_varlen_matches_dense_mask(self):
+        """Packed zigzag CP query positions + gathered-KV reorder should match dense masking."""
+        cp_size, cp_rank = 2, 1
+        cu_seqlens = torch.tensor([0, 8, 16], dtype=torch.int32)
```
Mellonta: Could we use uneven lengths like [0,4,16] to actually exercise different per-sequence chunk sizes?

tests/unit_tests/transformer/experimental_attention_variant/test_attention_variant_dsa.py
```diff
@@ -48,6 +67,249 @@ def mock_hadamard_transform(x: torch.Tensor, scale: float = 1.0) -> torch.Tensor
...
+    if calculate_per_token_loss:
+        kl_div = kl_per_row.sum()
+    elif query_valid_rows is None:
+        kl_div = kl_per_row.mean()
+    else:
+        kl_div = kl_per_row.sum() / query_valid_rows.sum().to(dtype=torch.float32).clamp_min(1.0)
+    return kl_div * loss_coeff
+
+
+class _FakeCPGroup:
```
Mellonta, line 302: All tests use `_FakeCPGroup` and a faked gather, so they only check the position math, never real sharding. Could we add a CP-invariance test that derives real per-rank shards from a global tensor and asserts cp_size=2 matches a cp_size=1 full-sequence run?

megatron/core/transformer/experimental_attention_variant/dsa_layout.py
HollowMan6: Yes, this block is a host-side validation guard for CPU/test/debug paths. Packed zigzag CP assumes each packed sequence is padded to a multiple of 2 * cp_size, so the rank-local front/back chunks are integral. In training we avoid checking CUDA runtime lengths here to prevent synchronization; the data pipeline is responsible for that padding guarantee. My preference is to keep it with the comment because it documents and validates the packed zigzag CP invariant without affecting CUDA training, but we can also get this removed if you prefer.

megatron/core/transformer/experimental_attention_variant/dsa_masking.py
HollowMan6: In the current main path, after gathered KV is restored to global order, key_pos is usually arange. I kept it explicit because this helper is a public function that meant to build masks from logical token positions rather than assuming KV storage order. That also covers cases where we build/apply the mask before KV reorder, or if a caller passes subset/reordered KV.

tests/unit_tests/transformer/experimental_attention_variant/test_attention_variant_dsa.py
HollowMan6: changed

tests/unit_tests/transformer/experimental_attention_variant/test_attention_variant_dsa.py
HollowMan6: added

tests/unit_tests/transformer/experimental_attention_variant/test_attention_variant_dsa.py
```diff
@@ -61,6 +324,1006 @@ def patch_hadamard_if_needed():
...
+            sq=sq,
+            skv=skv,
+            cp_size=cp_size,
+            cp_rank=cp_rank,
+            cp_comm_type="allgather",
+            device=torch.device("cpu"),
+        )
+
+        # Build varlen starts/ends for local query rows.
+        starts_all, ends_all = generate_varlen_mask_params(cu_seqlens.to(torch.int64))
```
Mellonta: The packed-THD CP path in production builds its varlen bounds with `generate_varlen_mask_params_for_positions`, but all tests use `generate_varlen_mask_params` + `index_select` (which I believe is equivalent). Could we call `generate_varlen_mask_params_for_positions` directly in some places, so the production path is also covered.

HollowMan6: good catch! addressed

Mellonta (approved):

kvareddy (approved):

santhnm2 (approved):

yaoyu-33 (approved):

PR #5253, opened 2026-06-10
fix(ci): stabilize failing functional test recipes

Connor-XY:
## What
Combine the CI recipe mitigations from:
- #5240: use batched pipeline P2P for `bert_mcore_tp1_pp4_vp2`
- #5241: use synchronous data loading for hanging T5 functional tests
- #5242: pin linear CUDA-graph sizing distribution for GRPO 8B throughput tests

Follow-up from the latest GitLab/JET BERT logs: `bert_mcore_tp1_pp4_vp2` still hung after the batched-P2P change, with one rank blocked in DataLoader `queue.get` while another timed out in batched pipeline P2P. This PR now also sets `--num-workers: 0` for that BERT recipe.

The fixes are kept as separate commits so each mitigation remains easy to inspect or revert independently.

## Why
These are all test-config-only fixes for functional tests that are currently failing or flaky in CI.

### BERT pp4/vp2 hang
`bert_mcore_tp1_pp4_vp2` originally failed as an intermittent pipeline-parallel P2P hang. Disabling P2P overlap routes the interleaved pp4/vp2 recipe to batched P2P (`batch_isend_irecv`) instead of the unbatched per-pair P2P path that can lazily create 2-rank NCCL communicators in inconsistent order.

The follow-up GitLab/JET failure shows the recipe update was applied (`overlap_p2p_comm=False`) and the hang moved to a batched `COALESCED` NCCL operation after iteration 15, while another rank was parked in `get_batch` / DataLoader `queue.get`. Setting `--num-workers: 0` removes the worker subprocess and pin-memory queue from this small functional test.

### T5 dataloader hang
The affected T5 recipes can surface as `finalize_model_grads` all-reduce timeouts, but fault-handler dumps show the stuck rank parked in `get_batch` / DataLoader `queue.get`. Setting `--num-workers: 0` removes the worker subprocess queue from these small functional tests.

### GRPO CUDA-graph memory guardrail
The GRPO throughput recipes fail their memory guardrail after the CUDA graph sizing default changed to `exponential` plus mixed-prefill grid. Pinning `--inference-dynamic-batching-cuda-graph-sizing-distribution: linear` restores the pre-change sizing behavior for these tests. This still needs CI confirmation because mixed-prefill count may also contribute to peak memory.

## Validation
- Parsed the updated BERT YAML locally with `python3` / PyYAML.
- Inspected the failing GitLab/JET BERT logs and confirmed the recipe was using `--no-overlap-p2p-communication`, `overlap_p2p_comm=False`, and still timing out in batched P2P while another rank was blocked in DataLoader `queue.get`.

Needs `Run functional tests` to validate the CI-dependent failure modes.

Supersedes #5240, #5241, and #5242.

### DIFF
--- a/tests/functional_tests/test_cases/bert/bert_mcore_tp1_pp4_vp2/model_config.yaml
+++ b/tests/functional_tests/test_cases/bert/bert_mcore_tp1_pp4_vp2/model_config.yaml
@@ -42,4 +42,9 @@ MODEL_ARGS:
   --bf16: true
   --ckpt-format: torch
   --attention-backend: unfused
+  # Interleaved/VPP schedule defaults to overlap_p2p_comm=True, which forces
+  # batch_p2p_comm=False and uses unbatched per-pair P2P that lazily creates
+  # 2-rank NCCL communicators -> intermittent lazy-comm-init deadlock (RECV hang).
+  # Use batched P2P here (pp=4 > 2 satisfies the interleaved+no-overlap assert).
+  --no-overlap-p2p-communication: true
 TEST_TYPE: regular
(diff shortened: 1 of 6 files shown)


Connor-XY:
/ok to test [a5f57c6](https://github.com/NVIDIA/Megatron-LM/pull/5253/commits/a5f57c64b064635c72948f5efab84f7d2356b181)

PR #5251, opened 2026-06-10
Add optional group= to common_utils model/data-parallel reduction helpers

yashaswikarnati:
Adds an optional `group=` parameter (default `None`) to the three single-group reduction helpers in `common_utils`: `average_losses_across_data_parallel_group`, `reduce_max_stat_across_model_parallel_group`, and `logical_and_across_model_parallel_group`. When `group` is `None`, each helper falls back to the same `mpu` getter as before.

Why: enables hetero-MIMO to reuse the stock training loop by threading an explicit ProcessGroupCollection; default `None` preserves current behavior for homogeneous runs.

Testing: unit test added covering explicit-group routing and default-None fall-through (including divisor correctness). Homogeneous goldens / cog 8-GPU validation is the next gate and has NOT yet been run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5259, opened 2026-06-10
Thread pg_collection into train_step reductions

yashaswikarnati:
Adds an optional `pg_collection: ProcessGroupCollection = None` to `train_step` and threads it through the four per-iteration reductions (logical_and / reduce_max over the model-parallel group, `is_pipeline_last_stage`, and the per-key loss all-reduce over the data-parallel+context-parallel group). When `pg_collection` is `None` (the default), every reduction falls back to the existing mpu globals, so homogeneous / non-MIMO runs are byte-for-byte unchanged.

## Why
Enables MIMO to run on the stock `megatron/training` loop via an explicit `ProcessGroupCollection`, sourcing per-module groups instead of process-global mpu state. Default `None` preserves current behavior for all existing callers.

## Testing
Real cog 8-GPU unit test (`tests/unit_tests/test_utils.py::TestTrainStepReductionThreading::test_threaded_groups_match_default`, tp=2 pp=2, no mocks), run name `tl3-train-step-test` — PASSED on all ranks. The test asserts the threaded reduction sub-logic (logical_and/max over `.mp`, `is_pp_last_stage` over `.pp`, loss all-reduce over `.dp_cp`) matches the default mpu path. Homogeneous goldens remain the next CI gate.

## Stacking
Stacked on #5251 (TL2): the branch contains the TL2 commit, so the GitHub diff shows TL2+TL3 until #5251 merges. Stacked on #5251 — rebase onto main after it merges.

### DIFF
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -2159,8 +2159,12 @@ def dummy_train_step(data_iterator):
             )
 
 
-def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_scheduler, config, forward_backward_func, iteration=None):
-    """Single training step."""
+def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_scheduler, config, forward_backward_func, iteration=None, pg_collection: Optional[ProcessGroupCollection] = None):
+    """Single training step.
+
+    pg_collection: optional per-module :class:`ProcessGroupCollection`; None uses the mpu globals,
+        otherwise it must define mp, pp, and dp_cp.
+    """
     args = get_args()
     timers = get_timers()
 
@@ -2298,14 +2302,25 @@ def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_sch
     if save_params_in_this_iteration:
         _save_state_dict(attr_name="data", label="params")
 
+    if pg_collection is None:
+        pg_collection = ProcessGroupCollection.use_mpu_process_groups()
+    for _required in ("mp", "pp", "dp_cp"):
+        assert getattr(pg_collection, _required, None) is not None, (
+            f"pg_collection passed to train_step must define {_required}"
+        )
+    mp_group = pg_collection.mp
+    dp_cp_group = pg_collection.dp_cp
+    is_last_stage = is_pp_last_stage(pg_collection.pp)
     # when freezing sub-models we may have a mixture of successful and unsucessful ranks,
     # so we must gather across mp ranks
-    update_successful = logical_and_across_model_parallel_group(update_successful)
+    update_successful = logical_and_across_model_parallel_group(update_successful, group=mp_group)
     # grad_norm and num_zeros_in_grad will be None on ranks without trainable params,
     # so we must gather across mp ranks
-    grad_norm = reduce_max_stat_across_model_parallel_group(grad_norm)
+    grad_norm = reduce_max_stat_across_model_parallel_group(grad_norm, group=mp_group)
     if args.log_num_zeros_in_grad:
-        num_zeros_in_grad = reduce_max_stat_across_model_parallel_group(num_zeros_in_grad)
+        num_zeros_in_grad = reduce_max_stat_across_model_parallel_group(
+            num_zeros_in_grad, group=mp_group
+        )
 
     # Vision momentum.
     if args.vision_pretraining and args.vision_pretraining_type == "dino":
@@ -2324,20 +2339,16 @@ def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_sch
     if args.empty_unused_memory_level >= 2:
         torch.cuda.empty_cache()
 
-    if mpu.is_pipeline_last_stage(ignore_virtual=True):
+    if is_last_stage:
         # Average loss across microbatches.
         loss_reduced = {}
-
         for key in losses_reduced[0].keys():
             val = [x[key].view(-1) for x in losses_reduced]
             if val[0].numel() == 2:
                 # there is one dict per microbatch. in new reporting, we average
                 # over the total number of tokens across the global batch.
                 val = torch.vstack(val).sum(dim=0)
-                torch.distributed.all_reduce(
-                    val,
-                    group=mpu.get_data_parallel_group(with_context_parallel=True)
-                )
+                torch.distributed.all_reduce(val, group=dp_cp_group)
                 loss_reduced[key] = val[0] / val[1]
             elif val[0].numel() == 1:
                 # legacy behavior, we average over the number of microbatches
--- a/tests/unit_tests/test_utils.py
+++ b/tests/unit_tests/test_utils.py
@@ -604,3 +604,56 @@ class TestSingleGroupReductionHelpers:
         default = reduce_max_stat_across_model_parallel_group(float(rank))
         assert default == explicit
         Utils.destroy_model_parallel()
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 8, reason="requires 8 GPUs")
+class TestTrainStepReductionThreading:
+    """train_step's reductions over a per-module ProcessGroupCollection's mp/pp/dp_cp groups."""
+
+    @staticmethod
+    def _ranks(group):
+        return torch.distributed.get_process_group_ranks(group)
+
+    def test_grid_threaded_reductions(self):
+        from megatron.core.hyper_comm_grid import HyperCommGrid
+        from megatron.core.pipeline_parallel.utils import is_pp_last_stage
+        from megatron.training.utils.common_utils import (
+            logical_and_across_model_parallel_group,
+            reduce_max_stat_across_model_parallel_group,
+        )
+
+        Utils.initialize_distributed()  # torch.distributed only; NOT initialize_model_parallel
+        rank = torch.distributed.get_rank()
+        # tp=2,pp=2,dp=2 over the world: mp == tp+pp, dp_cp == dp.
+        grid = HyperCommGrid([2, 2, 2], ["tp", "pp", "dp"], backend="nccl")
+        grid.create_pg(["tp", "pp"])
+        grid.create_pg(["pp"])
+        grid.create_pg(["dp"])
+        mp_group = grid.get_pg(["tp", "pp"])
+        pp_group = grid.get_pg(["pp"])
+        dp_cp_group = grid.get_pg(["dp"])
+        mp_ranks = self._ranks(mp_group)
+        pp_ranks = self._ranks(pp_group)
+        dp_cp_ranks = self._ranks(dp_cp_group)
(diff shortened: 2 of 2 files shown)


megatron/training/training.py
```diff
@@ -2300,12 +2304,13 @@ def _save_state_dict(attr_name, label):
 
     # when freezing sub-models we may have a mixture of successful and unsucessful ranks,
     # so we must gather across mp ranks
-    update_successful = logical_and_across_model_parallel_group(update_successful)
+    mp_group = pg_collection.mp if pg_collection is not None else None
+    update_successful = logical_and_across_model_parallel_group(update_successful, group=mp_group)
     # grad_norm and num_zeros_in_grad will be None on ranks without trainable params,
     # so we must gather across mp ranks
-    grad_norm = reduce_max_stat_across_model_parallel_group(grad_norm)
+    grad_norm = reduce_max_stat_across_model_parallel_group(grad_norm, group=mp_group)
```
yashaswikarnati, line 2319: does mimo hetero optimizer already handle these reductions/stats across different ranks and also encoder ranks vs LLM ranks

yashaswikarnati: Short answer: partially, and at a **broader** scope — so these TL3 reductions stay correct and are complementary, not redundant or conflicting.

**What `MimoOptimizer.step()` already reduces (over the WORLD / default group):**
- `found_inf` → `all_reduce(MAX)` over the default group (`megatron/core/models/mimo/optimizer.py:79-81`). On inf it returns `(False, None, None)`, so `update_successful` is already world-consistent (and across the disjoint encoder vs LLM optimizers).
- `grad_norm` → `get_grad_norm()` builds a per-module `norm_sq` vector and `all_reduce(MAX)` over the default group (`optimizer.py:60-71`), returning the global sqrt-sum — identical on every rank, encoder and LLM.
- Underneath, each per-module `DistributedOptimizer` already reduced its own grad stats over its `grad_stats_parallel_group` (= the full per-module world, see `optimizer.py:180-198` and the `intra_dist_opt` group `_get_pg_collection_for_optimizer` at `models/mimo/optimizer.py`).

**What `MimoOptimizer.step()` does NOT reduce across modules/world:**
- `num_zeros` → `count_zeros()` just sums the active local optimizers (`optimizer.py:117-118`); no cross-module all_reduce.

**What the train_step reductions are for (`common_utils.py:246-281`):** they MAX/MIN-gather over the *model-parallel* group so ranks without trainable params (non-last PP stage, frozen submodels) still see the logging value. For MIMO they fire over `pg_collection.mp` (one module’s tp×pp), a **subset** of the world the optimizer already reduced over.

**Determination — complementary / idempotent, leave as-is:**
- `update_successful` and `grad_norm`: already world-uniform out of the optimizer, so re-reducing (MIN / MAX) over the smaller `pg_collection.mp` operates on identical operands → idempotent no-op. Not wrong.
- `num_zeros_in_grad`: NOT pre-reduced across modules, so the `pg_collection.mp` MAX is the meaningful one for propagating it to non-optimizer ranks within the module — genuinely needed.
- The hetero prototype (`examples/mimo/training/hetero/step.py:191-216`) only re-reduces `update_successful` (over the default/world group) and skips the grad_norm/num_zeros gathers, consistent with the optimizer already world-reducing those.

So on the stock train_step path TL3 is enabling, these three reductions are correct: harmless for `update_successful`/`grad_norm`, and required for `num_zeros_in_grad`. No code change needed. The only nuance worth a follow-up: the optimizer reduces `found_inf`/`grad_norm` over the *world* while train_step re-reduces over the *per-module mp* — fine for consensus values, but if a future change makes these per-module-divergent we’d want them scoped consistently.

yashaswikarnati:
Re-implemented per review feedback (force-pushed).

**Hardened per-field group selection in `train_step`.** Each field of an optional `pg_collection` is now read defensively via `getattr(pg_collection, <field>, None)`, so a partially-initialized `ProcessGroupCollection` (a field left unset or explicitly `None`) falls back to the mpu global instead of mis-routing:
- `pp` unset/None: `is_pp_last_stage` is skipped and we use `mpu.is_pipeline_last_stage(ignore_virtual=True)` (previously `is_pp_last_stage(None)` would have returned True on every rank).
- `dp_cp` unset/None: the per-key loss all-reduce uses `mpu.get_data_parallel_group(with_context_parallel=True)` (previously `all_reduce(group=None)` would have silently used the default group).
- `mp` unset/None: passed through to the reduction helpers, which already fall back to `mpu.get_model_parallel_group()`.

Docstring now states that any field left `None` falls back to its mpu global. Default `pg_collection=None` remains byte-for-byte identical to today; no call sites change.

**Tests** (validated on cw-dfw 8-GPU, all PASSED, none skipped):
- `TestTrainStepReductionThreading`: 8-GPU `HyperCommGrid([2,2,2],[tp,pp,dp])` (no `parallel_state`) exercising reduce_max/logical_and over mp, `is_pp_last_stage` over pp, and loss all-reduce over dp_cp against hand-computed expectations.
- `TestTrainStepPGCFallbackSelection`: non-distributed coverage of the per-field selection for unset / explicit-None / set fields and a `None` collection.

megatron/training/training.py
```diff
@@ -2298,14 +2302,17 @@ def _save_state_dict(attr_name, label):
     if save_params_in_this_iteration:
         _save_state_dict(attr_name="data", label="params")
 
+    mp_group = getattr(pg_collection, "mp", None) if pg_collection is not None else None
```
yashaswikarnati: mp group none is dangerous? does it mean whole world group? should the fallback be parallel state ?

megatron/training/training.py
```diff
@@ -2324,20 +2331,26 @@ def _save_state_dict(attr_name, label):
...
+    pp_group = getattr(pg_collection, "pp", None) if pg_collection is not None else None
+    is_last_stage = (
+        is_pp_last_stage(pp_group)
+        if pp_group is not None
+        else mpu.is_pipeline_last_stage(ignore_virtual=True)
+    )
+    if is_last_stage:
         # Average loss across microbatches.
         loss_reduced = {}
+        dp_cp_group = getattr(pg_collection, "dp_cp", None) if pg_collection is not None else None
```
yashaswikarnati: we dont want to have none fallbacks, if a pg collection is passed, it shold be expected to have dp cp and pp groups and whatever groups are required here. we can assert for it, the fallback should be parallel state

tests/unit_tests/test_utils.py
```diff
@@ -604,3 +604,104 @@ def test_default_falls_back_to_mpu(self):
...
+            torch.distributed.all_reduce(val, group=dp_cp_group)
+            expected = torch.tensor(
+                [float(sum(dp_cp_ranks)), float(len(dp_cp_ranks))], dtype=torch.float32
+            )
+            torch.testing.assert_close(val.cpu(), expected)
+        finally:
+            grid.destroy()
+
+
+class TestTrainStepPGCFallbackSelection:
```
yashaswikarnati: looks like this test is not addding any value ?

tests/unit_tests/test_utils.py
```diff
@@ -604,3 +604,104 @@ def test_default_falls_back_to_mpu(self):
         default = reduce_max_stat_across_model_parallel_group(float(rank))
         assert default == explicit
         Utils.destroy_model_parallel()
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 8, reason="requires 8 GPUs")
+class TestTrainStepReductionThreading:
+    """train_step's threaded reduction sub-logic over a per-module ProcessGroupCollection
```
yashaswikarnati: remove verbose explanations in doc stings

megatron/training/training.py
yashaswikarnati: Reworked per this — removed the per-field `None` fallbacks. `train_step` now: if `pg_collection is None` → use `parallel_state` (mpu) globals; otherwise **assert** it defines `mp`/`pp`/`dp_cp` and use them directly. A partially-built collection now fails loudly instead of silently routing through a global. (commit `a5e8ee4da5`, cog 8-GPU green)

megatron/training/training.py
yashaswikarnati: Fixed — `mp_group` is never `None` now: the `pg_collection is None` branch passes `mpu.get_model_parallel_group()` explicitly, and the non-None branch asserts `.mp` is present. No `group=None` reaches the reduction helpers.

tests/unit_tests/test_utils.py
yashaswikarnati: Removed — that test only exercised the per-field fallback logic, which is gone now (replaced by the assert). The real value test is the 8-GPU `TestTrainStepReductionThreading`.

tests/unit_tests/test_utils.py
yashaswikarnati: Trimmed the docstrings to one concise line each.

yashaswikarnati:
/ok to test [e80fe18](https://github.com/NVIDIA/Megatron-LM/pull/5259/commits/e80fe18f2a5ddd88a638f0dd4adb09198d4a49f2)

deepakn94 (approved):

yaoyu-33 (approved):

megatron/training/training.py
```diff
@@ -2298,14 +2302,28 @@ def _save_state_dict(attr_name, label):
     if save_params_in_this_iteration:
         _save_state_dict(attr_name="data", label="params")
 
+    if pg_collection is None:
+        mp_group = mpu.get_model_parallel_group()
```
maanug-nv: would it be equivalent to use `ProcessGroupCollection.use_mpu_process_groups()`? pretty soon i think i'll make PGCollection required and just pass that in as default from `train()`.

yashaswikarnati: Yes, equivalent — switched to it. `train_step` now does `pg_collection = pg_collection or ProcessGroupCollection.use_mpu_process_groups()`, then asserts `mp`/`pp`/`dp_cp` are present and uses them uniformly (dropped the explicit `mpu.*` branch). `use_mpu_process_groups()` populates `mp` (`get_model_parallel_group`), `pp` (`get_pipeline_model_parallel_group`), and `dp_cp` (`get_data_parallel_group(with_context_parallel=True)`) — same groups/semantics as before, so the default path is unchanged. This is forward-compatible with making PGCollection required + passing it from `train()`: that just skips the `is None` line. (commit `bbdd1ee7f1`, cog 8-GPU green)

maanug-nv (approved):

yashaswikarnati:
/ok to test [bbdd1ee](https://github.com/NVIDIA/Megatron-LM/pull/5259/commits/bbdd1ee7f1cd39e6d748d5b0c71b590404fa867e)

PR #5260, opened 2026-06-10
Add MIMO hetero topology + distributed bootstrap (examples/mimo training-loop folder)

yashaswikarnati:
Adds per-module HyperCommGrid topology and a distributed bootstrap for hetero MIMO training under examples/mimo/training/, building each module's process groups (TP/CP/PP/DP, expert views, and language embedding groups) and packaging them into a MultiModuleProcessGroupCollection.

## Why

Enables MIMO to run on the stock megatron/training loop via an explicit ProcessGroupCollection rather than parallel_state globals. Process groups are owned by HyperCommGrid, not parallel_state. The default None preserves current behavior; there are no changes to megatron/core and no homogeneous/non-MIMO behavior changes.

## Testing

Real cog 8-GPU distributed unit test (no mocks), run name mm1-topology-test: 4 passed (test_grids_partition_world, test_pgc_group_sizes, test_validate_rejects_overlapping_not_equal, test_validate_rejects_gap_in_world_coverage). The homogeneous goldens remain the next CI gate.

Validation now enforces that module grids partition the world [0, world_size) with no gaps in addition to the pairwise-disjoint-XOR-fully-shared invariant.

## Stacking

Standalone on origin/main (no stacking).

### DIFF
--- a/examples/mimo/training/distributed.py
+++ b/examples/mimo/training/distributed.py
@@ -1,0 +1,61 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+"""Bring up torch.distributed and the global memory buffer for hetero MIMO without initializing MPU."""
+
+from __future__ import annotations
+
+import os
+
+import torch
+import torch.distributed as dist
+
+from megatron.core import parallel_state
+from megatron.training.utils import print_rank_0
+
+__all__ = ["initialize_distributed", "print_rank_0", "shutdown_distributed"]
+
+
+def initialize_distributed() -> None:
+    """Bring up torch.distributed + the global memory buffer without MPU init."""
+    local_rank = int(os.environ.get("LOCAL_RANK", "0"))
+    torch.cuda.set_device(local_rank)
+    if not dist.is_initialized():
+        dist.init_process_group(backend="nccl", device_id=torch.device(f"cuda:{local_rank}"))
+    assert_parallel_state_uninitialized()
+    try:
+        parallel_state.get_global_memory_buffer()
+    except AssertionError:
+        parallel_state._set_global_memory_buffer()
+    dist.barrier()
+
+
+def assert_parallel_state_uninitialized() -> None:
+    """Assert no Megatron model-parallel globals are set so the hetero path owns setup."""
+    # Defensive parallel_state reads (allowed compatibility point); check each global MIMO
+    # constructs, since model_parallel_is_initialized() omits CP/embedding.
+    checks = (
+        ("data_parallel", parallel_state.is_initialized),
+        ("tensor_model_parallel", _grp(parallel_state.get_tensor_model_parallel_group)),
+        ("pipeline_model_parallel", _grp(parallel_state.get_pipeline_model_parallel_group)),
+        ("context_parallel", _grp(parallel_state.get_context_parallel_group)),
+        ("embedding", _grp(parallel_state.get_embedding_group)),
+        ("position_embedding", _grp(parallel_state.get_position_embedding_group)),
+    )
+    initialized = [label for label, predicate in checks if predicate()]
+    if initialized:
+        raise RuntimeError(
+            "Hetero MIMO bootstrap expects Megatron parallel_state process groups to be "
+            f"uninitialized, but found: {', '.join(initialized)}"
+        )
+
+
+def _grp(getter):
+    """Turn a group getter into a no-arg "is it set?" predicate."""
+    return lambda: getter(check_initialized=False) is not None
+
+
+def shutdown_distributed() -> None:
+    """Tear down the global memory buffer and torch.distributed (thin stock teardown)."""
+    parallel_state.destroy_global_memory_buffer()
+    if dist.is_initialized():
+        dist.destroy_process_group()
--- a/examples/mimo/training/topology.py
+++ b/examples/mimo/training/topology.py
@@ -1,0 +1,247 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+"""Per-module ``HyperCommGrid`` topology and process-group ownership for hetero MIMO training."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Optional
+
+import torch.distributed as dist
+
+from megatron.core.hyper_comm_grid import HyperCommGrid
+from megatron.core.models.mimo.config.role import MIMO_LANGUAGE_MODULE_KEY, ModuleLayout, RankRole
+from megatron.core.parallel_state import default_embedding_ranks, default_position_embedding_ranks
+from megatron.core.process_groups_config import (
+    MultiModuleProcessGroupCollection,
+    ProcessGroupCollection,
+)
+
+_EXPERT_VIEW = "expert"
+
+
+@dataclass
+class ModuleGridSpec:
+    """One module's grid factorization and placement.
+
+    ``num_ranks`` is the ground truth; ``dp`` and ``expt_dp`` are derived in ``__post_init__``.
+    """
+
+    name: str
+    num_ranks: int
+    tp: int = 1
+    cp: int = 1
+    pp: int = 1
+    ep: int = 1
+    rank_offset: int = 0
+    # Experts default to TP=1 (set explicitly for MoE); intentionally not Megatron's etp=tp default.
+    expt_tp: int = 1
+    dp: int = field(init=False)
+    expt_dp: int = field(init=False)
+
+    def __post_init__(self) -> None:
+        dense = self.tp * self.cp * self.pp
+        if self.num_ranks % dense != 0:
+            raise ValueError(
+                f"num_ranks ({self.num_ranks}) must be divisible by tp*cp*pp ({dense})"
+            )
+        self.dp = self.num_ranks // dense
+        expert = self.expt_tp * self.ep * self.pp
+        if self.num_ranks % expert != 0:
+            raise ValueError(
+                f"num_ranks ({self.num_ranks}) must be divisible by expt_tp*ep*pp ({expert})"
+            )
+        self.expt_dp = self.num_ranks // expert
+
+    @property
+    def size(self) -> int:
+        """Total ranks spanned by this module's grid."""
+        return self.num_ranks
+
+
+@dataclass
+class HeteroTopology:
+    """Process groups and rank topology for one hetero MIMO run."""
+
+    grids: dict[str, HyperCommGrid]
+    module_pgs: dict[str, ProcessGroupCollection]
+    schedule_pg_collection: MultiModuleProcessGroupCollection
+
+    def destroy(self) -> None:
+        """Destroy every process group owned by this topology."""
+        destroyed: set[int] = set()
+        for pgc in self.module_pgs.values():
+            for pg in (pgc.embd, pgc.pos_embd):
+                if pg is None or id(pg) in destroyed or not _is_process_group_member(pg):
+                    continue
+                dist.destroy_process_group(pg)
+                destroyed.add(id(pg))
+        for grid in self.grids.values():
+            grid.destroy()
+
+
+def create_topology(specs: list[ModuleGridSpec]) -> HeteroTopology:
+    """Build every module's grid, PGC, and embedding groups.
+
+    Exactly one spec must be named ``MIMO_LANGUAGE_MODULE_KEY`` (the language module); specs must
+    tile ``[0, world_size)`` with no gaps (validated by :func:`_validate_grid_layout`).
+    """
+    if not specs:
+        raise ValueError("create_topology requires at least one ModuleGridSpec")
+    language_specs = [spec for spec in specs if spec.name == MIMO_LANGUAGE_MODULE_KEY]
+    if len(language_specs) != 1:
+        raise ValueError(
+            f"create_topology requires exactly one spec named {MIMO_LANGUAGE_MODULE_KEY!r} "
+            f"(the language module), got {len(language_specs)}"
+        )
+    language_name = MIMO_LANGUAGE_MODULE_KEY
+
+    grids: dict[str, HyperCommGrid] = {}
+    module_pgs: dict[str, ProcessGroupCollection] = {}
+    try:
+        for spec in specs:
+            grids[spec.name] = _build_grid(spec)
+        _validate_grid_layout(grids)
+
+        for name, grid in grids.items():
+            module_pgs[name] = pg_collection_from_grid(grid, is_language=(name == language_name))
+
+        schedule_pg_collection = build_schedule_pg_collection(grids, module_pgs, language_name)
+        return HeteroTopology(
+            grids=grids, module_pgs=module_pgs, schedule_pg_collection=schedule_pg_collection
+        )
+    except Exception:
+        HeteroTopology(grids=grids, module_pgs=module_pgs, schedule_pg_collection=None).destroy()
+        raise
+
+
+def _build_grid(spec: ModuleGridSpec) -> HyperCommGrid:
+    """Create a dense grid plus its expert view and the process groups MIMO needs."""
+    grid = HyperCommGrid(
+        shape=[spec.tp, spec.cp, spec.dp, spec.pp],
+        dim_names=["tp", "cp", "dp", "pp"],
+        rank_offset=spec.rank_offset,
+        backend="nccl",
+    )
+    # Expert factorization over the same rank span; pp is shared with the base view.
+    grid.register_view(
+        _EXPERT_VIEW,
+        shape=[spec.expt_tp, spec.ep, spec.expt_dp, spec.pp],
+        dim_names=["expt_tp", "ep", "expt_dp", "pp"],
+        shared_dims=["pp"],
+    )
+
+    try:
+        for dims in (["tp"], ["cp"], ["pp"], ["dp"], ["dp", "cp"], ["tp", "cp"], ["tp", "pp"]):
+            grid.create_pg(dims)
+        for dims in (["ep"], ["expt_tp"], ["expt_dp"], ["expt_tp", "ep"], ["expt_tp", "ep", "pp"]):
+            grid.create_pg(dims, view=_EXPERT_VIEW)
+    except Exception:
+        grid.destroy()
+        raise
+    return grid
+
+
+def _validate_grid_layout(grids: dict[str, HyperCommGrid]) -> None:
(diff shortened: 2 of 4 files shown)


examples/mimo/training/distributed.py
```diff
@@ -0,0 +1,77 @@
...
+import os
+import sys
+
+import torch
+import torch.distributed as dist
+
+from megatron.core import parallel_state
+
+
+def initialize_distributed() -> None:
```
yashaswikarnati, line 18: do we need seperate function here for this ? wont megatron/training handle this ?

examples/mimo/training/distributed.py
```diff
@@ -0,0 +1,77 @@
...
+        parallel_state._set_global_memory_buffer()
+    dist.barrier()
+
+
+def assert_parallel_state_uninitialized() -> None:
+    """Assert no Megatron model-parallel globals are set so the hetero path owns setup."""
+    # Defensive None-checks against parallel_state globals (allowed compatibility point):
+    # this path owns process-group construction and must not collide with MPU init.
+    initialized = []
+    if parallel_state.is_initialized():
```
yashaswikarnati: checking for all groups is too verbose, can just check parallel state

examples/mimo/training/distributed.py
```diff
@@ -0,0 +1,77 @@
...
+    if parallel_state.get_position_embedding_group(check_initialized=False) is not None:
+        initialized.append("position_embedding")
+    if initialized:
+        raise RuntimeError(
+            "Hetero MIMO bootstrap expects Megatron parallel_state process groups to be "
+            f"uninitialized, but found: {', '.join(initialized)}"
+        )
+
+
+def print_rank_0(message: str) -> None:
```
yashaswikarnati: does megatron train loop already have helper function like this ?

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+@dataclass
+class HeteroTopology:
+    """Process groups and rank topology for one hetero MIMO run."""
+
+    grids: dict[str, HyperCommGrid]
+    module_pgs: dict[str, ProcessGroupCollection]
+    schedule_pg_collection: MultiModuleProcessGroupCollection
+    language_embedding_groups: LanguageEmbeddingGroups
```
yashaswikarnati: do we really need seperate anguage_embedding_groups: LanguageEmbeddingGroups
    language_module_name: str = MIMO_LANGUAGE_MODULE_KEY ?

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,283 @@
...
+def create_topology(specs: list[ModuleGridSpec]) -> HeteroTopology:
+    """Build all grids, their PGCs, and embedding groups in one deterministic order.
+
+    The last spec is the language module; all preceding specs are modality (e.g. encoder) modules.
+    The specs must collectively span every world rank in ``[0, world_size)`` with no gaps: the
+    grids either tile the world disjointly (non-colocated) or fully share every rank (colocated).
+    """
+    if not specs:
+        raise ValueError("create_topology requires at least one ModuleGridSpec")
+    language_spec = specs[-1]
```
yashaswikarnati: strong assumption specs[-1] is language

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,283 @@
...
+        if _is_process_group_member(pg):
+            dist.destroy_process_group(pg)
+        destroyed.add(id(pg))
+    embedding_groups.clear()
+
+
+def build_schedule_pg_collection(
+    grids: dict[str, HyperCommGrid],
+    module_pgs: dict[str, ProcessGroupCollection],
+    language_module_name: str,
```
yashaswikarnati: this is kind of fixed ? language_module_name: str, ?

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,283 @@
...
+) -> MultiModuleProcessGroupCollection:
+    """Build the schedule-facing collection of the modules this rank participates in."""
+    rank_modules = {}
+    language_name = None
+    for name, grid in grids.items():
+        if not grid.is_current_rank_in_grid():
+            continue
+        rank_modules[name] = module_pgs[name]
+        if name == language_module_name:
```
yashaswikarnati: what is this if block even doing ?

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,283 @@
...
+            language_name = name
+    return MultiModuleProcessGroupCollection(
+        module_pgs=rank_modules, language_model_module_name=language_name
+    )
+
+
+def _is_process_group_member(pg: Optional[dist.ProcessGroup]) -> bool:
+    """Whether the current rank belongs to ``pg`` (not the non-member sentinel)."""
+    non_member = getattr(getattr(dist, "GroupMember", None), "NON_GROUP_MEMBER", None)
```
yashaswikarnati: what is this GroupMember and non group number, where is thi even defined ?

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,283 @@
...
+        )
+
+    modality_names = [n for n in names if n != MIMO_LANGUAGE_MODULE_KEY]
+    role = RankRole.build(modality_names, grids)
+    expected = ModuleLayout.COLOCATED if all_same else ModuleLayout.NON_COLOCATED
+    if role.mode is not expected:
+        raise ValueError(f"RankRole reported {role.mode} but rank spans imply {expected}: {spans}")
+
+
+def pg_collection_from_grid(grid: HyperCommGrid) -> ProcessGroupCollection:
```
yashaswikarnati: embedding groups can also be part of pg collection if embedding groups are requested in case of LLM

yashaswikarnati:
## Review comment resolutions

Mapping each review comment to its resolution in the latest revision (`75e2d04`):

1. **Distributed init kept (why stock cannot replace it):** `distributed.py` retains a torch.distributed + global-memory-buffer bootstrap. Stock `mpu.initialize_model_parallel` cannot be used because it materializes the model-parallel globals for a single homogeneous world tiling; the hetero MIMO topology needs per-module grids with disjoint rank offsets, so we bring up `torch.distributed` WITHOUT initializing MPU and assert the MPU globals stay uninitialized.
2. **Assert simplified:** `assert_parallel_state_uninitialized` is now a compact loop over the model-parallel globals (DP/TP/PP/CP/embd/pos_embd) rather than a call to `model_parallel_is_initialized`, so a partial/leaked init is caught.
3. **print_rank_0 reused:** now imported from `megatron.training.utils` instead of redefined locally.
4. **specs[-1] replaced:** the language module is now selected via the explicit `is_language_module` flag on the spec; `create_topology` validates exactly one spec sets it.
5. **language_module_name + side dict deleted:** removed; the layout is driven by `is_language_module` / `RankRole` / `ModuleLayout`.
6. **Embedding groups in PGC:** language embedding groups now live in `PGC.embd` / `PGC.pos_embd`, built collectively via `parallel_state.default_embedding_ranks` / `default_position_embedding_ranks` as real `ProcessGroup`s (encoder modules get `None`).
7. **is_current_rank_in_grid filter kept + explained:** retained to scope per-module construction to ranks actually in the grid; commented to explain why.
8. **GroupMember sentinel replaced:** membership is now checked via `get_rank(group) >= 0` (handles `-1` for non-members) instead of the `GroupMember` sentinel.
9. **Validation preserved:** the world-tiling invariant (grids tile the world disjointly XOR fully share ranks) is preserved and exercised by the test, including rejection of overlapping-but-not-equal and gap-in-coverage layouts.

### Validation
8-GPU real-distributed run on cw-dfw: `tests/unit_tests/test_mimo_hetero_topology.py` — **5 passed** (no skips):
`test_grids_partition_world`, `test_pgc_group_sizes`, `test_embedding_groups`, `test_validate_rejects_overlapping_not_equal`, `test_validate_rejects_gap_in_world_coverage`.

examples/mimo/training/distributed.py
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+"""torch.distributed + global-memory-buffer bootstrap for heterogeneous MIMO training.
+
+This bootstrap deliberately does NOT call ``mpu.initialize_model_parallel``: the hetero
```
yashaswikarnati: remove this verbose top of file doc string, not needed

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,270 @@
...
+    """
+
+    name: str
+    tp: int = 1
+    cp: int = 1
+    pp: int = 1
+    dp: int = 1
+    ep: int = 1
+    rank_offset: int = 0
+    expt_tp: Optional[int] = None
```
yashaswikarnati: why is this expt tp None instead of 1 ?

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,270 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+"""Per-module ``HyperCommGrid`` topology and process-group ownership for hetero MIMO training.
+
+Each module (one or more encoders plus the language model) gets its own ``HyperCommGrid`` placed
+at a configurable ``rank_offset``. The grids must either tile the world disjointly (non-colocated)
+XOR fully share every rank (colocated); this module validates that invariant. The shipped
+configuration is non-colocated (distinct, disjoint offsets), but nothing here precludes a
```
yashaswikarnati: remove -  The shipped
configuration is non-colocated (distinct, disjoint offsets), but nothing here precludes a
colocated layout (offset 0 for every module with distinct shapes).

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,270 @@
...
+            grids=grids, module_pgs=module_pgs, schedule_pg_collection=schedule_pg_collection
+        )
+    except Exception:
+        HeteroTopology(grids=grids, module_pgs=module_pgs, schedule_pg_collection=None).destroy()
+        raise
+
+
+def _build_grid(spec: ModuleGridSpec) -> HyperCommGrid:
+    """Create a dense grid plus its expert view and the process groups MIMO needs."""
+    expt_tp = spec.tp if spec.expt_tp is None else spec.expt_tp
```
yashaswikarnati: lets not make these assumptions, lets explicitly define and have default values for expt tp and expt dp

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,270 @@
...
+    expert_model_size = expt_tp * spec.ep * spec.pp
+    if spec.size % expert_model_size != 0:
+        raise ValueError(
+            f"module size ({spec.size}) must be divisible by expt_tp*ep*pp ({expert_model_size})"
+        )
+    expt_dp = spec.size // expert_model_size if spec.expt_dp is None else spec.expt_dp
```
yashaswikarnati: why do we handle expt_dp and dp differently.. its good that we calculate based on world size and implicitly calcuclate dp for both expert and non expert params.  how do we handle this better ? does this belong in post init of module grid spec and handle it there ?

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,270 @@
...
+    pgc.pos_embd = None
+    if is_language:
+        _build_language_embedding_groups(grid, pgc)
+    return pgc
+
+
+def _build_language_embedding_groups(grid: HyperCommGrid, pgc: ProcessGroupCollection) -> None:
+    """Build this rank's word/position embedding groups for the language grid.
+
+    Mirrors parallel_state: groups are created collectively (every rank in the grid calls
```
yashaswikarnati: verbose doc string, can be concisely said

examples/mimo/training/distributed.py
```diff
@@ -0,0 +1,73 @@
...
+from megatron.core import parallel_state
+from megatron.training.utils import print_rank_0
+
+__all__ = ["initialize_distributed", "print_rank_0", "shutdown_distributed"]
+
+
+def initialize_distributed() -> None:
+    """Bring up torch.distributed and the global memory buffer for torchrun.
+
+    Stock ``initialize_megatron`` cannot replace this: ``skip_mpu_initialization=True`` returns
```
yashaswikarnati: verbose doc string and can be be concisely written better

yashaswikarnati:
## Round-2 review comments addressed

**Docstrings (4 trimmed for concision while preserving architectural meaning):**
- `distributed.py` module docstring and `initialize_distributed` collapsed to a one-line summary plus a one-line justification.
- `topology.py` module docstring and `_build_language_embedding_groups` trimmed; the essential invariants (per-module grids, embedding-group construction) are kept.

**Expert dims (`expt_tp` / `expt_dp`):**
- These are now resolved to concrete `int`s and validated inside `ModuleGridSpec.__post_init__` (no `None` sentinels left to flow downstream). `expt_tp` defaults to `tp`; `expt_dp` is derived as `size // (expt_tp * ep * pp)`, with divisibility and product checks (`expt_tp * ep * expt_dp * pp == size`) enforced at construction.
- `_build_grid` is simplified to read these already-resolved concrete values directly, with no `None`-fallback branching.

**`dp` kept explicit (rationale):**
- The module-view size invariant is `size = tp * cp * pp * dp`, so `dp` is a first-class field of the module grid. `expt_dp` is its expert-view analog and is derived in `__post_init__` from the expert factorization; keeping `dp` explicit mirrors the two distinct views (module vs. expert) rather than overloading one field.

Verified on an 8-GPU real-distributed run: all 8 tests in `tests/unit_tests/test_mimo_hetero_topology.py` pass (including new `TestModuleGridSpecResolution` cases for implicit/explicit expert resolution and invalid factorization).

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,271 @@
...
+                dist.destroy_process_group(pg)
+                destroyed.add(id(pg))
+        for grid in self.grids.values():
+            grid.destroy()
+
+
+def create_topology(specs: list[ModuleGridSpec]) -> HeteroTopology:
+    """Build all grids, their PGCs, and embedding groups in one deterministic order.
+
+    Exactly one spec must set ``is_language_module=True``; all others are modality (e.g. encoder)
```
yashaswikarnati: remove - Exactly one spec must set ``is_language_module=True``; all others are modality (e.g. encoder)
    modules. The specs must collectively span every world rank in ``[0, world_size)`` wi

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,271 @@
...
+    def __post_init__(self) -> None:
+        if self.expt_tp is None:
+            self.expt_tp = self.tp
+        expert_model_size = self.expt_tp * self.ep * self.pp
+        if self.size % expert_model_size != 0:
+            raise ValueError(
+                f"module size ({self.size}) must be divisible by expt_tp*ep*pp "
+                f"({expert_model_size})"
+            )
+        if self.expt_dp is None:
```
yashaswikarnati: what about dp ? should we not derive this as well?

examples/mimo/training/topology.py
yashaswikarnati: Good call. `expt_tp` is now a plain `int` with default `1` (no `None` sentinel anywhere in `ModuleGridSpec`). Worth noting it is an integer dimension of the expert factorization, not a process group handle, so there was never a `group=None` -> "use the whole world" footgun here. Still, an explicit `1` is clearer than `None`, and MoE runs set it explicitly since experts default to TP=1.

examples/mimo/training/topology.py
yashaswikarnati: Yes — `dp` is now derived too. `num_ranks` is the new ground-truth field, and `__post_init__` computes `dp = num_ranks // (tp * cp * pp)` after a divisibility check, symmetric with `expt_dp = num_ranks // (expt_tp * ep * pp)`. Both are `field(init=False)` so callers cannot set them directly. This mirrors `parallel_state`'s `world_size // model_parallel_size` pattern and removes the asymmetry where `dp` was an input but `expt_dp` was derived.

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,253 @@
...
+    """Process groups and rank topology for one hetero MIMO run."""
+
+    grids: dict[str, HyperCommGrid]
+    module_pgs: dict[str, ProcessGroupCollection]
+    schedule_pg_collection: MultiModuleProcessGroupCollection
+
+    @property
+    def module_to_grid_map(self) -> dict[str, HyperCommGrid]:
+        """MIMO module-to-grid mapping consumed by schedules and models."""
+        return dict(self.grids)
```
liding-nv: nit: `grids` is a dict object `grids: dict[str, HyperCommGrid]`

examples/mimo/training/topology.py
```diff
@@ -0,0 +1,253 @@
...
+    """
+    if not specs:
+        raise ValueError("create_topology requires at least one ModuleGridSpec")
+    language_specs = [spec for spec in specs if spec.is_language_module]
+    if len(language_specs) != 1:
+        raise ValueError(
+            f"create_topology requires exactly one spec with is_language_module=True, "
+            f"got {len(language_specs)}"
+        )
+    language_name = language_specs[0].name
```
liding-nv: should we use `MIMO_LANGUAGE_MODULE_KEY`?

examples/mimo/training/topology.py
yashaswikarnati: Done — removed the `module_to_grid_map` property. It was a redundant copy of `grids` (already `dict[str, HyperCommGrid]`) and unused; callers use `topology.grids` directly. (commit `c82f09f7b2`)

examples/mimo/training/topology.py
yashaswikarnati: Good catch — yes. The language module is now identified by `MIMO_LANGUAGE_MODULE_KEY` (dropped the `is_language_module` flag): `create_topology` requires exactly one spec named `MIMO_LANGUAGE_MODULE_KEY`. This also fixes a latent inconsistency — `_validate_grid_layout` already derived modality vs. language by `name != MIMO_LANGUAGE_MODULE_KEY`, so a flag-named language module would have been misclassified as a modality. Now everything (selection, validation, RankRole, MMPGC) keys off the canonical constant. Re-validated on cw-dfw 8-GPU: 9 passed. (commit `c82f09f7b2`)

liding-nv (approved):

yashaswikarnati:
/ok to test [c82f09f](https://github.com/NVIDIA/Megatron-LM/pull/5260/commits/c82f09f7b28b047bfb5077c0cbc788209d3fbc79)

PR #5263, opened 2026-06-10
Support non-colocated/heterogeneous grids in the MIMO optimizer

yashaswikarnati:
## What

Make the MIMO per-module optimizer work when a MIMO model's submodules live on **disjoint rank grids** (e.g. a vision encoder and a (MoE) language model on separate grids stitched by the 1F1B bridge schedule), rather than only the colocated case.

## Changes (`megatron/core/models/mimo/optimizer.py`)

1. **Expert-parallel groups from the grid's registered `"expert"` rank view.** `_get_pg_collection_for_optimizer` sources `tp_ep_pp` / `expt_dp` via `grid.get_pg([...], view="expert")`, falling back to base dims (`["tp","ep","pp"]` / `["dp","ep"]`) when the view is absent (`KeyError`). The dense `intra_dist_opt = grid.get_pg(["tp","cp","dp","pp"])` deliberately **excludes** the expert dim so dense grad statistics are not inflated by `ep_size`.
2. **`count_zeros` via per-module cross-rank `all_reduce(MAX)`** (was a local sum), so the reported count is consistent across each module's grid — mirroring the existing `get_grad_norm` pattern.
3. **`sharded_state_dict` prefixes each submodule with `mimo.{name}.`** via `add_prefix_for_sharding`, so distributed-optimizer checkpoint keys don't collide between submodules sharing an mp-rank-0 slot. Save/load are symmetric; the existing dist-checkpoint machinery is preserved.
4. **Skip building an optimizer for fully frozen submodules**, gated by a cross-rank trainable-parameter check (`all_reduce(MAX)` over the module's own `intra_dist_opt` group, so hetero subsets don't deadlock).

## Tests

- `tests/unit_tests/models/mimo/test_mimo_optimizer.py` (new): expert-view resolution + base-dim fallback, `count_zeros` cross-rank consistency, and sharded-key collision/disambiguation. Real distributed tests on the 1-node x 8-GPU lane.
- `tests/unit_tests/models/mimo/test_mimo_1f1b_schedule.py`: adds `test_moe_llm_ep2_etp2_8gpu`, an 8-GPU MoE e2e exercising the optimizer through the hetero 1F1B bridge schedule (encoder PP1/DP4 -> MoE LLM TP2/PP1/DP2, EP2/ETP2). Asserts finite positive `grad_norm`, finite loss, non-zero expert-param grads, and that expert (`tp_ep_pp`) ranks differ from dense (`mp`) ranks.

Both suites pass on 8 GPUs.

## Scope

Touches only `megatron/core/models/mimo/optimizer.py` and its tests (single CODEOWNERS group). No changes to `megatron/core/optimizer/__init__.py`.

### DIFF
--- a/megatron/core/models/mimo/optimizer.py
+++ b/megatron/core/models/mimo/optimizer.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
 import torch
 
 from megatron.core.dist_checkpointing.mapping import ShardedObject
+from megatron.core.dist_checkpointing.utils import add_prefix_for_sharding
 from megatron.core.optimizer.clip_grads import clip_grad_by_total_norm_fp32
 from megatron.core.optimizer.optimizer import MegatronOptimizer
 from megatron.core.optimizer.optimizer_config import OptimizerConfig
@@ -119,7 +120,26 @@ class MimoOptimizer(MegatronOptimizer):
         return torch.tensor([1.0], dtype=torch.float32, device="cuda")
 
     def count_zeros(self) -> int:
-        return sum(opt.count_zeros() for opt in self._active_optimizers)
+        """Count zero gradients globally across all modules via all_reduce MAX.
+
+        Mirrors ``get_grad_norm``: each module optimizer's ``count_zeros`` already
+        SUM-reduces over its own grad-stats group, so the count is identical across
+        all ranks that own the module but is ``0`` on ranks where the module is
+        inactive. A plain local ``sum`` over ``self._active_optimizers`` therefore
+        returns a *different* value on encoder ranks vs. language-model ranks in
+        non-colocated mode. Indexing each module into its own slot and taking a
+        cross-rank MAX recovers every module's true count on every rank, so the
+        summed total is consistent everywhere.
+        """
+        num_modules = len(self.module_infos)
+        zeros_by_module = torch.zeros(num_modules, device="cuda", dtype=torch.float32)
+
+        for i, (name, info) in enumerate(sorted(self.module_infos.items())):
+            if info.is_active and info.optimizer:
+                zeros_by_module[i] = float(info.optimizer.count_zeros())
+
+        torch.distributed.all_reduce(zeros_by_module, op=torch.distributed.ReduceOp.MAX)
+        return int(zeros_by_module.sum().item())
 
     @property
     def param_groups(self) -> List[dict]:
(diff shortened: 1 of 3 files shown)


megatron/core/models/mimo/optimizer.py
```diff
@@ -119,7 +120,26 @@ def get_loss_scale(self) -> torch.Tensor:
         return torch.tensor([1.0], dtype=torch.float32, device="cuda")
 
     def count_zeros(self) -> int:
-        return sum(opt.count_zeros() for opt in self._active_optimizers)
+        """Count zero gradients globally across all modules via all_reduce MAX.
+
+        Mirrors ``get_grad_norm``: each module optimizer's ``count_zeros`` already
```
yashaswikarnati, line 125: very long verbose doc string

megatron/core/models/mimo/optimizer.py
```diff
@@ -314,53 +349,126 @@ def _get_replica_id(pg_collection: Optional[ProcessGroupCollection]) -> tuple:
...
+
+
 def _get_pg_collection_for_optimizer(grid) -> ProcessGroupCollection:
     """Create ProcessGroupCollection from HyperCommGrid for optimizer use.
 
-    Only fetches process groups required by the optimizer. Assumes all groups
-    are pre-created in the grid via grid.create_pg() - does not create any new groups.
-
-    The following groups must be pre-created in the grid before calling this function:
+    Only fetches process groups required by the optimizer. Assumes all groups are
```
yashaswikarnati, line 384: very verbose doc strings throughout not needed

megatron/core/models/mimo/optimizer.py
```diff
@@ -314,53 +349,126 @@ def _get_replica_id(pg_collection: Optional[ProcessGroupCollection]) -> tuple:
...
+        - intra_dist_opt: distributed-optimizer grad-stats group (dense dims only)
     """
     pg = ProcessGroupCollection()
 
-    # Core groups needed by optimizer and checkpointing
+    # Dense groups (base view).
     pg.dp = grid.get_pg("dp")
     pg.dp_cp = grid.get_pg(["dp", "cp"])
+    # With a single distributed-optimizer instance the intra/full DP groups coincide.
+    pg.intra_dp_cp = pg.dp_cp
```
yashaswikarnati, line 444: intra dp cp is not same as dp cp. this is wrong. can we check parallel state for how intra dp cp is created. where is intra dp cp even needed? is not not intra dist opt ?

megatron/core/models/mimo/optimizer.py
```diff
@@ -314,53 +349,126 @@ def _get_replica_id(pg_collection: Optional[ProcessGroupCollection]) -> tuple:
...
+
+    # Distributed-optimizer grad-stats group. ``get_grad_norm_fp32`` /
+    # ``count_zeros_fp32`` SUM-reduce over this group, so it must contain exactly the
+    # ranks that hold *unique* dense gradient shards: tp/cp/pp partition the
+    # parameters and dp holds distinct optimizer shards after reduce-scatter. It must
+    # NOT include the expert ``ep`` axis: dense parameters are replicated across ``ep``
+    # in a heterogeneous grid (where ``ep`` is an independent dimension rather than a
+    # sub-partition of ``dp``), so including it would SUM each dense contribution
+    # ``ep_size`` times and inflate the grad norm / zero count.
+    pg.intra_dist_opt = grid.get_pg(["tp", "cp", "dp", "pp"])
```
yashaswikarnati, line 471: intra dist opt is also not tp cp dp pp. check the logic for intra and inter dist opt groups, generic to support num distributed optimzier instances > 1

PR #5285, opened 2026-06-11
Add MIMO runtime setup: per-role RNG seeding and DDP wrapping

yashaswikarnati:
Introduce `configure_module_rng` to seed the CUDA RNG tracker from a module's tp/pp coordinates plus a role offset (so disjoint modules and PP stages get independent RNG state while DP/CP replicas stay in sync), 

## Why
Enables MIMO to run on the `megatron/training` loop: per-role RNG seeding.

## Testing
Real cog 8-GPU unit test `mm2-rng-data-test` (`tests/unit_tests/test_mimo_hetero_data_rng.py`): 3 passed on all ranks (no skips). Covers RNG distinctness across role offsets/PP stages, encoder+language PP-edge iterator selection, and language-only middle-stage `None` selection. Functional goldens are the next CI gate.

### DIFF
--- a/examples/mimo/training/runtime.py
+++ b/examples/mimo/training/runtime.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import argparse
+from types import SimpleNamespace
 from typing import Optional
 
 import torch
@@ -16,7 +17,7 @@ from megatron.core.models.mimo.model.base import MimoModel
 from megatron.core.process_groups_config import ProcessGroupCollection
 from megatron.core.tensor_parallel.random import model_parallel_cuda_manual_seed
 from megatron.core.utils import get_pg_rank, get_pg_size
-from megatron.training.training import wrap_model_chunks_with_ddp
+from megatron.training.training import resolve_ddp_bucket_size, wrap_model_chunks_with_ddp
 from megatron.training.utils import print_rank_0
 
 
@@ -47,18 +48,27 @@ def configure_module_rng(
 def _resolve_bucket_size(
     args: argparse.Namespace, module: torch.nn.Module, dp_cp_group, overlap_grad_reduce: bool
 ) -> Optional[int]:
-    """Mirror megatron.training.get_model's DDP bucket-size logic (training.py:1764-1778)."""
-    if not overlap_grad_reduce:
-        return None  # get_model sets bucket_size=None when overlap is off
+    """Resolve a module's DDP bucket size via the shared get_model helper.
+
+    Maps the MIMO ``args`` (``ddp_num_buckets`` / ``ddp_bucket_size``) onto the
+    fields ``resolve_ddp_bucket_size`` reads from a DDP config, then delegates so the
+    3-branch policy stays single-sourced with ``get_model``. ``ddp_bucket_size <= 0``
+    is normalized to ``None`` (the default trigger) and an empty module yields ``None``.
+    """
     num_buckets = getattr(args, "ddp_num_buckets", None)
     if num_buckets is not None:
         assert num_buckets > 0
-        num_params = sum(p.numel() for p in module.parameters())
-        return max(1, num_params // num_buckets) if num_params > 0 else None
     bucket_size = getattr(args, "ddp_bucket_size", 0)
-    if bucket_size and bucket_size > 0:
-        return bucket_size
-    return max(40_000_000, 1_000_000 * get_pg_size(dp_cp_group))  # get_model's sane default
+    if not bucket_size or bucket_size <= 0:
+        bucket_size = None
+    num_params = sum(p.numel() for p in module.parameters())
+    if num_buckets is not None and num_params == 0:
+        return None  # empty module: no params to bucket
+    config = SimpleNamespace(num_buckets=num_buckets, bucket_size=bucket_size)
+    resolved = resolve_ddp_bucket_size(config, dp_cp_group, overlap_grad_reduce, num_params)
+    if num_buckets is not None and resolved is not None:
+        return max(1, resolved)  # never hand DDP a zero-size bucket
+    return resolved
 
 
 def set_module_requires_grad(module: Optional[torch.nn.Module], requires_grad: bool) -> None:
--- a/examples/mimo/training/runtime.py
+++ b/examples/mimo/training/runtime.py
@@ -16,9 +16,29 @@ from megatron.core.models.mimo.config.role import MIMO_LANGUAGE_MODULE_KEY
 from megatron.core.models.mimo.model.base import MimoModel
 from megatron.core.process_groups_config import ProcessGroupCollection
 from megatron.core.tensor_parallel.random import model_parallel_cuda_manual_seed
+from megatron.core.transformer.module import Float16Module
 from megatron.core.utils import get_pg_rank, get_pg_size
 from megatron.training.training import resolve_ddp_bucket_size, wrap_model_chunks_with_ddp
 from megatron.training.utils import print_rank_0
+
+
+class _EncoderFloat16Module(Float16Module):
+    """Float16Module variant whose outputs stay in model precision (bf16/fp16).
+
+    The stock :class:`Float16Module` upcasts last-PP-stage outputs to fp32 (its
+    ``forward`` default ``fp32_output=True``). For a MIMO encoder that is wrong: the
+    encoder's last-stage output is the projected modality activation that flows over the
+    cross-grid bridge into the language model, and the bridge expects model-precision
+    (bf16) activations. ``MimoModel._forward_encoders`` calls the submodule as
+    ``submodule.forward(encoder_inputs=..., hidden_states=...)`` and does not thread
+    ``fp32_output`` through, and the flag is forward-only (not a constructor argument), so
+    this thin subclass pins the default to ``False`` for encoder submodules. The language
+    model keeps the stock ``fp32_output=True`` (its label-mode output is the loss, already
+    fp32 from the cross-entropy upcast, so the upcast is a no-op).
+    """
+
+    def forward(self, *inputs, fp32_output=False, **kwargs):  # noqa: D102
+        return super().forward(*inputs, fp32_output=fp32_output, **kwargs)
 
 
 def configure_module_rng(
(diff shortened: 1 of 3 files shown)


examples/mimo/training/data.py
```diff
@@ -0,0 +1,152 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+"""Role-aware data iterator selection for heterogeneous MIMO training."""
```
yashaswikarnati: why is data part of PR cuda rng tracker and random seed handling PR

examples/mimo/training/runtime.py
```diff
@@ -0,0 +1,33 @@
...
+import argparse
+
+import torch
+
+from megatron.core.process_groups_config import ProcessGroupCollection
+from megatron.core.tensor_parallel.random import model_parallel_cuda_manual_seed
+from megatron.core.utils import get_pg_rank
+
+
+def configure_module_rng(
```
yashaswikarnati, line 30: do we need this in mimo train loop ? i wonder where megatron/training loop handles this and update this. i wonder if we can reuse this as part of megatron training loop or any need for having this here?

examples/mimo/training/data.py
yashaswikarnati: Split out — role-aware data selection (data.py) now lives in its own PR https://github.com/NVIDIA/Megatron-LM/pull/5287; #5285 is RNG-only now.

examples/mimo/training/runtime.py
yashaswikarnati: Kept here, and confirmed the mimo folder is the right home. The stock seeder _set_random_seed is called once at init (megatron/training/initialize.py:100, inside finish_mpu_init — the parallel_state-init path MIMO skips), not in the train loop, and it offsets the seed from mpu.get_pipeline_model_parallel_rank() — so MIMO can't reuse it. Per-module seeding (distinct state per module, ranks injected from the pg_collection, force_reset_rng to re-seed) is a different shape. The reusable piece is the core model_parallel_cuda_manual_seed (already accepts injected tp/ep/etp ranks), which configure_module_rng wraps with pgc-derived ranks + the per-role offset. So a thin MIMO-specific helper here is correct.

yashaswikarnati:
Expanded scope: this PR now adds the per-rank hetero MIMO runtime helpers in `examples/mimo/training/runtime.py` — `configure_module_rng` (per-role CUDA RNG seeding), `_resolve_bucket_size`, `set_module_requires_grad`, `wrap_active_modules_with_ddp` (freeze + DDP-wrap active modules over their per-module process groups), and `iter_active_ddp_modules`. Model construction (`build_mimo_runtime`) is deferred to the integration PR. Validated on an 8-GPU cw-dfw run: 5 passed (distributed tests run, not skipped).

examples/mimo/training/runtime.py
```diff
@@ -0,0 +1,137 @@
...
+    if mimo_model.language_model is not None:
+        if getattr(args, "freeze_lm", False):
+            set_module_requires_grad(mimo_model.language_model, False)
+        ddp_config = DistributedDataParallelConfig(
+            overlap_grad_reduce=getattr(args, "overlap_grad_reduce", False),
+            overlap_param_gather=getattr(args, "overlap_param_gather", False),
+            bucket_size=_resolve_bucket_size(args, mimo_model.language_model),
+            pad_buckets_for_high_nccl_busbw=pad_buckets,
+            use_distributed_optimizer=True,
+            # Keep main_grad in fp32; bf16 main_grad causes step-2 weight drift after Adam.
```
yashaswikarnati: remove comment - # Keep main_grad in fp32; bf16 main_grad causes step-2 weight drift after Adam.

examples/mimo/training/runtime.py
```diff
@@ -0,0 +1,137 @@
...
+        print_rank_0(f"wrapping modality submodule {name!r} in DDP")
+        mimo_model.modality_submodules[name] = DistributedDataParallel(
+            config=_module_config(submodule),
+            ddp_config=ddp_config,
+            module=submodule,
+            pg_collection=topology.module_pgs[name],
+        )
+
+
+def iter_active_ddp_modules(mimo_model: MimoModel) -> Iterator[DistributedDataParallel]:
```
yashaswikarnati: where and why this function is used ?

examples/mimo/training/runtime.py
```diff
@@ -0,0 +1,137 @@
...
+            module=mimo_model.language_model,
+            pg_collection=topology.module_pgs[MIMO_LANGUAGE_MODULE_KEY],
+        )
+
+    for name, submodule in mimo_model.modality_submodules.items():
+        if submodule is None or name not in topology.module_pgs:
+            continue
+        if getattr(args, "freeze_encoders", False):
+            set_module_requires_grad(submodule, False)
+        # Modality DDP keeps overlap off: actual-data batches may be text-only, so some encoder
```
yashaswikarnati: remove comment -  Modality DDP keeps overlap off: actual-data batches may be text-only, so some encoder
        # DP ranks see zero grads/params per step; overlap'd collectives are unsafe under that. are we sure about this claim? how did we verify ?

examples/mimo/training/runtime.py
```diff
@@ -0,0 +1,137 @@
...
+def configure_module_rng(
+    args: argparse.Namespace, pg_collection: ProcessGroupCollection, role_seed_offset: int
+) -> None:
+    """Seed the CUDA RNG tracker for one module role from its tp/pp coordinates plus the offset.
+
+    The seed is shared across a module's DP/CP replicas but distinct across PP stages and roles,
+    so disjoint modules (and stages) get independent RNG state. Caller invokes once per active
+    module on this rank.
+    """
+    pp_rank = get_pg_rank(getattr(pg_collection, "pp", None))
```
yashaswikarnati: none fallback is dangerous, pg collection should defn have these attributes and we can even assert for that

examples/mimo/training/runtime.py
```diff
@@ -0,0 +1,137 @@
...
+def _resolve_bucket_size(
+    args: argparse.Namespace, module: Optional[torch.nn.Module]
+) -> Optional[int]:
+    """Resolve DDP bucket_size: num_params // ddp_num_buckets, else ddp_bucket_size, else None."""
+    num_buckets = getattr(args, "ddp_num_buckets", None)
+    if num_buckets is not None:
+        if num_buckets <= 0:
+            raise ValueError("--ddp-num-buckets must be > 0 when set")
+        if getattr(args, "ddp_bucket_size", 0):
+            raise ValueError("--ddp-num-buckets and --ddp-bucket-size are mutually exclusive")
```
yashaswikarnati: also check megatron training loop on how it calculates default bucker sizes if these args are not specified? also can we reuse the function from megatron/training?

examples/mimo/training/runtime.py
```diff
@@ -0,0 +1,137 @@
...
+        if module is None:
+            return None
+        num_params = sum(p.numel() for p in module.parameters())
+        return max(1, num_params // num_buckets) if num_params > 0 else None
+    bucket_size = getattr(args, "ddp_bucket_size", 0)
+    return bucket_size if bucket_size and bucket_size > 0 else None
+
+
+def set_module_requires_grad(module: Optional[torch.nn.Module], requires_grad: bool) -> None:
```
yashaswikarnati: do we need to explicitly do this ? how does megatron training loop handle this ?

examples/mimo/training/runtime.py
yashaswikarnati: Removed — it's only consumed by the no_sync / zero_grad_buffer path (build_no_sync_func / zero_active_grad_buffers), which lands in MM4. Nothing in this PR used it, so it'll come with MM4 (its first consumer).

examples/mimo/training/runtime.py
yashaswikarnati: Trimmed the comment, kept overlap off. Verified: the encoder's per-step DP participation varies — mark_modality_participation tags each step on whether the rank had image input and _vision_participation_count all-reduces it (both in #5286). On a text-only microbatch some encoder DP ranks skip the encoder backward, so DDP's overlap_grad_reduce backward hooks never fire on those ranks while firing on others -> the per-bucket all-reduce desyncs across the DP group (hang). overlap-off makes the reduce happen once in finalize_model_grads where #5286's participation scaling corrects it.

examples/mimo/training/runtime.py
yashaswikarnati: Done — now asserts pg_collection defines pp/tp/ep/expt_tp (MM1's topology always builds the expert view, so all four are present) and accesses them directly; no getattr-None fallback.

examples/mimo/training/runtime.py
yashaswikarnati: Checked get_model (training.py:1764-1778): with --ddp-num-buckets it does num_parameters // num_buckets; else if no explicit bucket_size it defaults to max(40_000_000, 1_000_000 * dp_cp_size); and sets bucket_size=None when overlap_grad_reduce is off. _resolve_bucket_size now mirrors that exactly. The logic is inline in get_model (no importable helper), so I matched it rather than imported.

examples/mimo/training/runtime.py
yashaswikarnati: Yes — get_model does exactly this (training.py:1765: ddp_config.bucket_size = num_parameters // ddp_config.num_buckets). DistributedDataParallelConfig has both bucket_size and num_buckets and asserts mutual-exclusivity, but DDP needs an absolute bucket_size, so the per-module num_params//num_buckets conversion is done explicitly, mirroring get_model.

yashaswikarnati:
Refactored \`wrap_active_modules_with_ddp\` to reuse \`wrap_model_chunks_with_ddp\` (dist-opt full_param_layout + DDP construction) inside the \`get_model\`-style CUDA side-stream, instead of constructing \`DistributedDataParallel\` directly. \`_resolve_bucket_size\` is kept since it mirrors \`get_model\`'s pre-resolution of the bucket size that the layout computation needs.

STACKED on #5328 (the pgc-aware-wrap PR that threads \`pg_collection\` through \`wrap_model_chunks_with_ddp\` for both DDP and FSDP, and sources the dist-opt DP/expert-DP world sizes from \`pg_collection\` when provided). Will rebase onto main after #5328 merges.

Validated on cw-dfw 8-GPU: all 8 ranks pass \`tests/unit_tests/models/mimo/test_mimo_hetero_runtime.py\` (6 passed per rank, zero failures).

maanug-nv (approved):

deepakn94 (approved):

yashaswikarnati:
/ok to test [5d5f7d9](https://github.com/NVIDIA/Megatron-LM/pull/5285/commits/5d5f7d98d098ced11e45815edac8d543b760c316)

PR #5286, opened 2026-06-11
Add MIMO dual gradient finalization (colocated + non-colocated)

yashaswikarnati:
Adds `examples/mimo/training/grad_sync.py :: configure_grad_sync`, which installs a `finalize_model_grads_func` that finalizes the LLM and each modality submodule over its own per-module process-group collection and applies the per-token mean (1/N_global) uniformly, with optional partial-participation correction for vision DP.

Faithful to the prototype (sasatheesh/pre-vlm-07): `N_global` is sourced only from the LLM token-source ranks (`is_token_source_rank`: last PP stage, TP rank 0), reduced over the LLM DP/CP group, then broadcast world-wide via `ReduceOp.MAX` so the disjoint encoder grid divides by the same per-token mean. Works for both colocated and non-colocated layouts.

## Why
The dual gradient finalization lets heterogeneous (vision + language) modules on their own DP groups share a single per-token loss normalization on the stock `megatron/training` loop.

## Testing (cog 8-GPU, per-rank verified)
- **Non-colocated e2e**: `test_mimo_1f1b_schedule` now drives every config through `configure_grad_sync` (calculate_per_token_loss=True). The TP=2 / LLM PP=3 config exercises the cross-grid token sourcing + N_global broadcast.
- **Colocated numeric oracle**: `test_mimo_colocated_correctness` wires `configure_grad_sync` and checks post-step weights match a DP=1 reference.
- **Participation**: `test_mimo_grad_sync` covers the partial-participation correction helper.
All green: 14 passed, 2 skipped (lower-GPU configs), 0 failures.

Part of the NMFW-516 MIMO-on-stock-trainloop series.

### DIFF
--- a/examples/mimo/training/grad_sync.py
+++ b/examples/mimo/training/grad_sync.py
@@ -1,0 +1,170 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+"""Dual gradient finalization for MIMO training on the stock Megatron loop."""
+
+from __future__ import annotations
+
+import torch
+import torch.distributed as dist
+
+from examples.mimo.training.topology import HeteroTopology
+from megatron.core.distributed.finalize_model_grads import finalize_model_grads
+from megatron.core.models.mimo.config.role import MIMO_LANGUAGE_MODULE_KEY
+from megatron.core.models.mimo.model.base import MimoModel
+from megatron.core.pipeline_parallel.utils import is_pp_last_stage
+
+# Sentinel set per modality submodule when this rank had that modality's input this step.
+_PARTICIPATED_ATTR = "_mimo_rank_processed_input"
+
+
+def _has_modality_input(value) -> bool:
+    """Whether this rank received this modality's input this step.
+
+    The batch omits a modality's key when absent, so ``value`` is None (not present) or a
+    non-empty nested dict (present); an empty tensor also counts as absent.
+    """
+    if isinstance(value, torch.Tensor):
+        return value.numel() > 0
+    return bool(value)
+
+
+def mark_modality_participation(mimo_model: MimoModel, batch) -> None:
+    """Tag each modality submodule with whether this rank had that modality's input this step.
+
+    Reads ``batch["modality_inputs"]`` (keyed by modality name) so the flag is per modality
+    rather than vision-specific.
+    """
+    modality_inputs = batch.get("modality_inputs", {}) if isinstance(batch, dict) else {}
+    for name, submodule in mimo_model.modality_submodules.items():
+        if submodule is not None:
+            setattr(submodule, _PARTICIPATED_ATTR, _has_modality_input(modality_inputs.get(name)))
+
+
+def reset_modality_participation(mimo_model: MimoModel) -> None:
+    """Clear per-step participation flags at the top of each train step."""
+    for submodule in mimo_model.modality_submodules.values():
+        if submodule is not None:
+            setattr(submodule, _PARTICIPATED_ATTR, False)
+
+
+def _vision_participation_count(submodule, vision_dp_group) -> float:
+    """Number of vision-DP ranks that processed image input this step."""
+    val = 1.0 if getattr(submodule, _PARTICIPATED_ATTR, False) else 0.0
+    indicator = torch.tensor([val], dtype=torch.float32, device="cuda")
+    dist.all_reduce(indicator, op=dist.ReduceOp.SUM, group=vision_dp_group)
+    return float(indicator.item())
+
+
+def _is_pg_member(pg) -> bool:
+    """Whether the current rank belongs to ``pg`` (defensive; -1 for non-members)."""
+    return pg is not None and dist.get_rank(group=pg) >= 0
+
+
+def _is_token_source_rank(language_pg) -> bool:
+    """Whether this rank is the single LLM coordinate that owns the global token count.
+
+    Sourcing the count from one coordinate (last PP stage, TP rank 0) avoids
+    double-counting it across TP/PP replicas. In non-colocated, language_pg is the LLM
+    collection seen on every rank, so encoder-grid ranks reach here with non-member
+    (None) pp/tp groups — the _is_pg_member guards short-circuit so they are never the
+    source.
+    """
+    if language_pg is None:
+        return False
+    pp = getattr(language_pg, "pp", None)
+    tp = getattr(language_pg, "tp", None)
+    return (
+        _is_pg_member(pp)
+        and _is_pg_member(tp)
+        and is_pp_last_stage(pp)
+        and dist.get_rank(group=tp) == 0
+    )
+
+
+def _global_token_count(num_tokens, language_pg) -> float:
+    """Total non-padded tokens in the global batch, visible on every rank.
+
+    Only the LLM token-source ranks contribute: they sum over the LLM DP/CP group;
+    a world MAX then publishes that N_global to every rank, including the
+    non-colocated encoder grid (where ``language_pg`` is None and the count is 0).
+    """
+    global_num_tokens = torch.zeros(1, dtype=torch.float32, device="cuda")
+    if _is_token_source_rank(language_pg):
+        token_count = num_tokens.to(dtype=torch.float32).sum().view(1)
+        dist.all_reduce(token_count, group=language_pg.dp_cp, op=dist.ReduceOp.SUM)
+        if dist.get_rank(group=language_pg.dp_cp) == 0:
+            global_num_tokens.copy_(token_count)
+    dist.all_reduce(global_num_tokens, op=dist.ReduceOp.MAX)
+    return float(global_num_tokens.item())
+
+
+def configure_grad_sync(args, mimo_model: MimoModel, topology: HeteroTopology) -> None:
+    """Configure per-module gradient finalization: each module finalizes over its own groups.
+
+    The encoder and LLM have decoupled parallelism (separate grids), so each reduces its
+    gradients over its own process-group collection; both then divide by one shared
+    per-token mean (N_global).
+
+    MimoModel structure (each a separately DDP-wrapped module on its own grid)::
+
+        MimoModel
+        ├─ language_model          (LLM)      -> own process groups
+        └─ modality_submodules[*]  (encoders) -> own process groups
+    """
+    module_pgs = topology.module_pgs
+    language_pg = module_pgs.get(MIMO_LANGUAGE_MODULE_KEY)
+    correct_vision_grad = bool(
+        getattr(args, "correct_encoder_grad_for_partial_participation", False)
+    )
+
+    def finalize_grads_func(_model_list, num_tokens, force_all_reduce=False, **_kwargs):
+        # calculate_per_token_loss=True => DDP gradient_scaling_factor 1.0 (pure SUM),
+        # so the per-token mean is applied here by dividing every shard by N_global.
+        assert num_tokens is not None, (
+            "MIMO grad sync expects calculate_per_token_loss=True so the schedule "
+            "forwards total_num_tokens; got None."
+        )
+
+        # N_global is the global token count, published to every rank (including the
+        # non-colocated encoder grid) so both modules divide by the same per-token mean.
+        n_global = _global_token_count(num_tokens, language_pg)
+        inv = 1.0 / n_global if n_global > 0 else 0.0
+
+        if mimo_model.language_model is not None:
+            finalize_model_grads(
+                [mimo_model.language_model],
+                num_tokens=None,
+                pg_collection=language_pg,
+                force_all_reduce=force_all_reduce,
+            )
+            if inv != 0.0:
+                mimo_model.language_model.scale_gradients(inv)
+
+        for name, submodule in mimo_model.modality_submodules.items():
+            if submodule is None:
+                continue
+            vision_pg = module_pgs.get(name)
+            finalize_model_grads(
+                [submodule],
+                num_tokens=None,
+                pg_collection=vision_pg,
+                force_all_reduce=force_all_reduce,
+            )
+
+            vision_scale = inv
+            if correct_vision_grad and vision_pg is not None and vision_pg.dp is not None:
+                vision_dp_group = vision_pg.dp
+                if _is_pg_member(vision_dp_group):
+                    vision_dp_size = dist.get_world_size(vision_dp_group)
+                    if vision_dp_size > 1:
+                        participation = _vision_participation_count(submodule, vision_dp_group)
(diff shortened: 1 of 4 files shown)


examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,136 @@
...
+                    vision_dp_size = dist.get_world_size(vision_dp_group)
+                    if vision_dp_size > 1:
+                        participation = _vision_participation_count(submodule, vision_dp_group)
+                        if 0.0 < participation < vision_dp_size:
+                            vision_scale *= vision_dp_size / participation
+
+            if vision_scale != 0.0:
+                submodule.scale_gradients(vision_scale)
+
+        # End-of-iter barrier across vision-DP ranks guards a cross-grid collective
```
yashaswikarnati: this barrier is not requried

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,136 @@
...
+            vision_pg = module_pgs.get(name)
+            finalize_model_grads(
+                [submodule],
+                num_tokens=None,
+                pg_collection=vision_pg,
+                force_all_reduce=force_all_reduce,
+            )
+
+            vision_scale = inv
+            # Partial-participation correction only applies when vision ranks are
```
yashaswikarnati: remove comment - also its not true that this is only for non colocated. colocated can also have differentr participation >

examples/mimo/training/grad_sync.py
yashaswikarnati: Removed the end-of-iter vision-DP barrier entirely (commit 623bc56e). It was a stale deadlock workaround; the authoritative pre-vlm-07 prototype had already dropped it.

examples/mimo/training/grad_sync.py
yashaswikarnati: Fixed — removed the comment and dropped the `non_colocated` gate from the participation correction. You are right that colocated ranks can also have differing participation (a colocated rank may process a text-only microbatch), so the correction now applies regardless of layout, gated only by `--correct-encoder-grad-for-partial-participation` plus the DP-size checks. Also cleaned up the now-dead `non_colocated` local + `ModuleLayout` import and fixed two test docstrings that repeated the same inaccurate claim. cog re-verified green per-rank.

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,148 @@
...
+
+def _is_token_source_rank(language_pg) -> bool:
+    """The single LLM coordinate (last PP stage, TP rank 0) that owns the token count.
+
+    The schedule's ``total_num_tokens`` is only meaningful where the loss is computed
+    (the LLM last pipeline stage); other PP/TP coordinates hold stale values.
+    """
+    if language_pg is None:
+        return False
+    pp, tp = getattr(language_pg, "pp", None), getattr(language_pg, "tp", None)
```
yashaswikarnati: none fallback is dangerous, assert that pg has pp an tp..

tests/unit_tests/models/mimo/test_mimo_1f1b_schedule.py
```diff
@@ -599,26 +607,37 @@ def run_mimo_1f1b_test(
...
-                finalize_model_grads([submodule], num_tokens=None, pg_collection=vision_pg)
-
-    mimo_model.config.no_sync_func = no_sync_func
-    mimo_model.config.finalize_model_grads_func = finalize_grads_func
-    mimo_model.config.grad_scale_func = lambda loss: (
-        torch.tensor(loss, dtype=torch.float32, device='cuda', requires_grad=True)
-        if isinstance(loss, (int, float))
-        else loss
-    )
+    if grad_sync:
```
yashaswikarnati: why this is gated on if grad sync ?

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,148 @@
...
+def _is_pg_member(pg) -> bool:
+    """Whether the current rank belongs to ``pg`` (defensive; -1 for non-members)."""
+    return pg is not None and dist.get_rank(group=pg) >= 0
+
+
+def _is_token_source_rank(language_pg) -> bool:
+    """The single LLM coordinate (last PP stage, TP rank 0) that owns the token count.
+
+    The schedule's ``total_num_tokens`` is only meaningful where the loss is computed
+    (the LLM last pipeline stage); other PP/TP coordinates hold stale values.
```
yashaswikarnati: remove - other PP/TP coordinates hold stale values. - this is wrong statement

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,148 @@
...
+
+# Sentinel set on a modality submodule by forward_step when this rank had image input.
+_PARTICIPATED_ATTR = "_mimo_rank_processed_input"
+
+
+def mark_modality_participation(mimo_model: MimoModel, batch) -> None:
+    """Tag each modality submodule with whether this rank had image input this step."""
+    images = batch.get("images") if isinstance(batch, dict) else None
+    if isinstance(images, torch.Tensor):
+        had_input = images.numel() > 0
```
deepakn94: `has_images` might be a better name?

deepakn94: Should we make this more generic too?

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,148 @@
...
+    a world MAX then publishes that N_global to every rank, including the
+    non-colocated encoder grid (where ``language_pg`` is None and the count is 0).
+    """
+    global_num_tokens = torch.zeros(1, dtype=torch.float32, device="cuda")
+    if _is_token_source_rank(language_pg):
+        token_count = num_tokens.to(dtype=torch.float32).sum().view(1)
+        dist.all_reduce(token_count, group=language_pg.dp_cp, op=dist.ReduceOp.SUM)
+        if dist.get_rank(group=language_pg.dp_cp) == 0:
+            global_num_tokens.copy_(token_count)
+    dist.all_reduce(global_num_tokens, op=dist.ReduceOp.MAX)
```
deepakn94: Why do we need the world max?

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,148 @@
...
+    """Install a finalize_model_grads_func that finalizes the LLM and each modality
+    submodule over its own per-module process-group collection.
+    """
+    module_pgs = topology.module_pgs
+    language_pg = module_pgs.get(MIMO_LANGUAGE_MODULE_KEY)
+    correct_vision_grad = bool(
+        getattr(args, "correct_encoder_grad_for_partial_participation", False)
+    )
+
+    def finalize_grads_func(_model_list, num_tokens, force_all_reduce=False, **_kwargs):
```
deepakn94, line 143: Comment on why we need this?

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,148 @@
...
+            "MIMO grad sync expects calculate_per_token_loss=True so the schedule "
+            "forwards total_num_tokens; got None."
+        )
+
+        # N_global is the global token count, published to every rank (including the
+        # non-colocated encoder grid) so both modules divide by the same per-token mean.
+        n_global = _global_token_count(num_tokens, language_pg)
+        inv = 1.0 / n_global if n_global > 0 else 0.0
+
+        if mimo_model.language_model is not None:
```
deepakn94, line 156: Maybe add quick comment here as to what the structure of a mimo_model is? Some ASCII art might be useful

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,161 @@
...
+_PARTICIPATED_ATTR = "_mimo_rank_processed_input"
+
+
+def _has_modality_input(value) -> bool:
+    """Whether a per-modality batch entry carries actual input on this rank this step."""
+    if value is None:
+        return False
+    if isinstance(value, torch.Tensor):
+        return value.numel() > 0
+    if isinstance(value, (list, tuple, dict)):
```
yashaswikarnati: why if isinstance(value, (list, tuple, dict)): ? can it be a dict with none value ?

examples/mimo/training/grad_sync.py
yashaswikarnati: here were are computing the normalization factor for grads - global num  tokens in the entire global batch. all the ranks (including encoder) gets initialized with value 0 and only dp_cp=0 rank has the actual count. we could also do broadcast instead of max, tried out to avoid book keeping of src ranks and dest ranks if we wanna do broadcast. can also modify it that way. lmk what you think

examples/mimo/training/grad_sync.py
yashaswikarnati: added more detai in the doc string of top level function

examples/mimo/training/grad_sync.py
yashaswikarnati: yeah, made this mode generic to accound for all modalities/encoders present in the model.

examples/mimo/training/grad_sync.py
yashaswikarnati: added in the outer function doc string

yashaswikarnati:
/ok to test [19364db](https://github.com/NVIDIA/Megatron-LM/pull/5286/commits/19364dbe76d19703e872d84bd8cd89708b5fd008)

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,170 @@
...
+    tp = getattr(language_pg, "tp", None)
+    return (
+        _is_pg_member(pp)
+        and _is_pg_member(tp)
+        and is_pp_last_stage(pp)
+        and dist.get_rank(group=tp) == 0
+    )
+
+
+def _global_token_count(num_tokens, language_pg) -> float:
```
liding-nv: Q: `train.py` currently all-reduces loss and tokens over CP before returning `local_num_tokens`. Should `global_token_count` expect CP_local tokens or already CP-reduced tokens?
https://github.com/NVIDIA/Megatron-LM/blob/1cfa834a8dfa0ac9826888e80327d2bae23540a6/examples/mimo/train.py#L157-L173

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,170 @@
...
+    return bool(value)
+
+
+def mark_modality_participation(mimo_model: MimoModel, batch) -> None:
+    """Tag each modality submodule with whether this rank had that modality's input this step.
+
+    Reads ``batch["modality_inputs"]`` (keyed by modality name) so the flag is per modality
+    rather than vision-specific.
+    """
+    modality_inputs = batch.get("modality_inputs", {}) if isinstance(batch, dict) else {}
```
deepakn94, line 37: Maybe "multimodal_inputs" is better?

examples/mimo/training/grad_sync.py
deepakn94: Broadcast is better / more intuitive, I think.

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,170 @@
...
+    if _is_token_source_rank(language_pg):
+        token_count = num_tokens.to(dtype=torch.float32).sum().view(1)
+        dist.all_reduce(token_count, group=language_pg.dp_cp, op=dist.ReduceOp.SUM)
+        if dist.get_rank(group=language_pg.dp_cp) == 0:
+            global_num_tokens.copy_(token_count)
+    dist.all_reduce(global_num_tokens, op=dist.ReduceOp.MAX)
+    return float(global_num_tokens.item())
+
+
+def configure_grad_sync(args, mimo_model: MimoModel, topology: HeteroTopology) -> None:
```
liding-nv, line 121: where do we call this function? is the intent to be installed by a later PR? Just to understand where this hook will become active

yashaswikarnati: yeah all of this will wire up once i raise the subsequent PRs which completes the whole train loop. for now just setting up the building blocks

examples/mimo/training/grad_sync.py
yashaswikarnati: Could we do it in a follow up PR. i basically have to update the naming at data loader, here and model fwd pass interface. can stage a different PR for rename across the board

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,203 @@
...
+def _token_source_global_rank(language_grid) -> int:
+    """Global (default-group) rank of the single LLM token-source coordinate.
+
+    The source is the language coordinate (tp=0, cp=0, dp=0, pp=last). It is derived
+    statically from grid metadata (shape/dim_names/rank_offset) that is identical on
+    every rank, so even the non-colocated encoder grid — which is not a member of any
+    LLM group — can name this global rank as the broadcast root.
+
+    ``get_rank_enum("pp")`` yields the exact PP rank lists ``create_pg`` uses. The grid's
+    ``rank_offset`` is always the all-zero coordinate (tp=0, cp=0, dp=0, pp=0), so the PP
```
yashaswikarnati: is there a more direct way than assuming - ``rank_offset`` is always the all-zero coordinate (tp=0, cp=0, dp=0, pp=0),?

examples/mimo/training/grad_sync.py
```diff
@@ -0,0 +1,203 @@
...
+
+    MimoModel structure (each a separately DDP-wrapped module on its own grid)::
+
+        MimoModel
+        ├─ language_model          (LLM)      -> own process groups
+        └─ modality_submodules[*]  (encoders) -> own process groups
+    """
+    module_pgs = topology.module_pgs
+    language_pg = module_pgs.get(MIMO_LANGUAGE_MODULE_KEY)
+    # Static (collective-free) derivation of the token source's global rank, used as the
```
yashaswikarnati: very verbose comment

examples/mimo/training/grad_sync.py
yashaswikarnati: updated to broadcast

deepakn94 (approved):

yashaswikarnati:
/ok to test [75872f2](https://github.com/NVIDIA/Megatron-LM/pull/5286/commits/75872f2e8718c8967501340a158d9150b400bc65)

yashaswikarnati:
/ok to test [7457d08](https://github.com/NVIDIA/Megatron-LM/pull/5286/commits/7457d0824427571f6aa112a365c4f8e048ffbc67)

yashaswikarnati:
/ok to test [cadd331](https://github.com/NVIDIA/Megatron-LM/pull/5286/commits/cadd331400dac574f2e079959276646efe34c0e0)

PR #5288, opened 2026-06-11
[Dev] Stabilize mFSDP HybridEP full-CG and paged stash

lhb8125:
## Status

Closed with zero diff. The source branch was reset to `dev@5139086e7` after review showed that the experimental changes were not justified as requirements for full-iteration CUDA graph support.

## Why the changes were removed

- The merge-base A/B run completed full-CG capture and stable replay without the original four-file PR delta on the paged-stash path.
- `capture_error_mode="relaxed"` and HybridEP handle retention had no isolated failure demonstrating that they were necessary.
- The implicit HybridEP `1.2` static budget could silently drop tokens and did not handle the overflow flag, so a throughput-only run was not sufficient evidence to merge it.
- The paged-stash/FSDP parameter-lifecycle fixes added during later investigation address a separate issue and do not belong in this PR.
- The no-paged-stash MBS=2 run with the experimental static-budget path completed capture, but both it and the historical baseline reported `grad norm: nan`; that result validates throughput only, not correctness or necessity.

The removed investigation head remains preserved locally as `codex/pr5288-investigation-backup-e62fc50` for reference. Any independently reproducible lifecycle bug should be proposed in a separate, focused PR with its own correctness test.


lhb8125:
/ok to test [c32874a](https://github.com/NVIDIA/Megatron-LM/pull/5288/commits/c32874a4d143851f5d1e33700abb4223531d07c6)

lhb8125:
/ok to test [3d121d5](https://github.com/NVIDIA/Megatron-LM/pull/5288/commits/3d121d58fa5b95eeca06b9a144803e94cf027b00)

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1121,6 +1125,15 @@ def setup_metadata(self, routing_map: torch.Tensor, probs: torch.Tensor):
             # Round budget up to pad_multiple (FP8/FP4/CUTLASS alignment for permute buffers).
             budget += -budget % pad_multiple
             self.num_permuted_tokens = budget
+        elif self.config.cuda_graph_impl == "full_iteration":
```
nanz-nv: Do you think it is better to add an assertion to error out explicitly instead of setting the budget silently?

megatron/core/full_cuda_graph.py
```diff
@@ -18,6 +19,31 @@
 _shared_capture_stream = None
 
 
+def _synchronize_fsdp_param_gathers(model):
```
nanz-nv: My preference is that to make this file generic enough and add mode-specific information in other palces. Similar comments for the code change below. @vasunvidia what's your opinion on this?

megatron/core/full_cuda_graph.py
```diff
@@ -199,39 +247,53 @@ def __call__(self, *args, **kwargs):
...
+            captured_result = None
+            # FSDP/DTensor backward can run cleanup from autograd worker threads; relaxed mode
+            # keeps those releases from invalidating PyTorch allocator state during capture.
+            with _override_stale_capture_stream():
+                with torch.autograd.set_multithreading_enabled(False):
+                    with torch.cuda.graph(
+                        FullCudaGraphWrapper.cuda_graph[training_str],
+                        stream=capture_stream,
+                        pool=get_graph_pool(self.use_single_mempool),
+                        capture_error_mode="relaxed",
```
nanz-nv: Is this necessary? What is the case that triggers capture error without changing it to relax? I was under the impression that FSDP works with full-iter cuda graph.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1214,6 +1238,10 @@ def combine(
         # Release the used handle/num_permuted_tokens which could change in each iteration.
         # For drop_and_pad mode, we don't need to reset the num_permuted_tokens and
         # num_dispatched_tokens, because their values never change.
+        is_current_stream_capturing = getattr(torch.cuda, "is_current_stream_capturing", None)
+        if is_current_stream_capturing is not None and is_current_stream_capturing():
+            # Captured graph replays may still reference backend resources stored in the handle.
+            self._cuda_graph_handles.append(self.handle)
```
nanz-nv: Why do we need this? Do you encounter a specific failure that you suspect might be triggered by this?

megatron/core/transformer/moe/experts.py
```diff
@@ -670,6 +692,78 @@ def _fused_forward(
...
+        )
+        if (
+            not is_hybridep_full_cg
+            or permuted_probs.shape[0] < permuted_local_hidden_states.shape[0]
+        ):
+            raise RuntimeError(
+                "Mismatched MoE dispatched token/probability rows: "
+                f"hidden={permuted_local_hidden_states.shape[0]}, probs={permuted_probs.shape[0]}"
+            )
+        return permuted_probs[: permuted_local_hidden_states.shape[0]]
```
vasunvidia: What scenario causes permuted_probs.shape[0] to be greater than permuted_local_hidden_states.shape[0] to require this line?

megatron/core/transformer/moe/experts.py
```diff
@@ -670,6 +692,78 @@ def _fused_forward(
...
+        is_hybridep_full_cg = (
+            self.config.cuda_graph_impl == "full_iteration"
+            and self.config.moe_token_dispatcher_type == "flex"
+            and self.config.moe_flex_dispatcher_backend == "hybridep"
+        )
+        if not is_hybridep_full_cg:
+            return tokens_per_expert_list
+        padded_tokens_per_expert = list(tokens_per_expert_list)
+        token_padding = permuted_local_hidden_states.shape[0] - sum(padded_tokens_per_expert)
+        padded_tokens_per_expert[-1] += token_padding
```
vasunvidia: Why do we need to modify tokens_per_expert to include padding?
Won't this cause the CuteDSL grouped GEMM kernel to treat the padded region as valid tokens for the last local expert ?

megatron/core/transformer/moe/token_dispatcher.py
vasunvidia: I also agree that we should add `assert moe_expert_rank_capacity_factor is not None` for Full iteration CG instead of silently using a default value which might not be sufficient for a given model.

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -733,6 +748,12 @@ def backward(ctx, grad_x):
             num_permuted_tokens=ctx.num_permuted_tokens,
             **({"fuse_permute_dispatch": ctx.fused} if ctx.fused else {}),
         )
+        dispatched_hidden = trim_hybridep_static_budget_padding(
```
vasunvidia: Could you add a comment indicating why this is required? When will dispatched_hidden.shape[0] be greater than ctx.input_num_tokens?

megatron/core/full_cuda_graph.py
vasunvidia: Yes. Full CG should be agnostic to FSDP implementation.

PR #5017, opened 2026-05-27
Add new `SFTDataset` to `megatron.core.datasets`

asolergi-nv:
- [X] I, the PR author, have personally reviewed every line of this PR.

## Summary

This PR replaces the JSONL-backed `SFTDataset` in `megatron/training/datasets/sft_dataset.py` with a new pre-tokenized, packed implementation under `megatron/core/datasets/sft_dataset.py`. The new dataset:

- Reads pre-tokenized chat conversations from Megatron's file prefixes (`.bin` / `.idx`) like pretraining. It's up to the user to apply the chat template and later on pretokenize the data with the `tools/preprocess_data.py` script.
- Packs variable-length conversations into fixed-capacity samples using the **Modified First-Fit Decreasing (MFFD)** bin-packing algorithm and caches the resulting indices on disk.
- Builds per-token **loss masks** by parsing each conversation against a configurable chat template (`ChatTemplateConfig` — Nemotron-3 default). Three flags control what counts toward the loss:
  - `train_on_assistant_responses_only`
  - `train_on_thinking_traces` (mask `<think>…</think>` regions)
  - `train_on_tool_calls` (mask `<tool_call>…</tool_call>` regions). Tool responses (environment output inside user turns) are always masked.
- Emits THD-packed outputs (`tokens`, `labels`, `loss_mask`, `cu_seqlens`) ready for Flash Attention varlen kernels.

A new helper, `preprocess_sft_batch`, is added to `megatron/core/utils.py` and called from both `pretrain_gpt.py` and `pretrain_hybrid.py`. It handles the runtime steps that used to happen inside the old `__getitem__`: padding each sub-sequence to a CP-divisible length, padding or truncating the packed sample to `seq_length`, and producing `position_ids`, `cu_seqlens_padded`, and `max_seqlen`.

A standalone debug script, `inspect_sft_file_prefix.py`, is added at the repo root for inspecting a `.bin`/`.idx` pair against the Nemotron-3 template. Will remove before merge.

User-facing reference docs for the dataset live in `megatron/core/datasets/sft.MD`.

---

### New: dataset and helpers

| File | What it does |
|------|--------------|
[...]

PR #5298, opened 2026-06-11
Add inter-document attention masking to GPTDataset

deepakn94:
## Summary
- Add `inter_document_masking` config option to `GPTDatasetConfig` that computes `cu_seqlens` and `max_seqlen` from document boundaries within each sample, and resets position IDs per document.
- Enable FlashAttention / TransformerEngine (and linear attention) to restrict attention to individual documents via the varlen (`thd`) path, replacing the legacy dense attention mask approach used by `--reset-attention-mask`.
- Wire `cu_seqlens` / `max_seqlen` through TP broadcast (`get_batch_on_this_tp_rank`) so non-zero TP ranks receive the metadata.
- Add `--dataloader-inter-document-masking` CLI argument.
- Update `FIMDataset._query_document_sample_shuffle_indices` to return the 3-tuple expected by the new code path.
- Enable context parallelism for inter-document masking via per-document CP balancing (`_get_batch_on_this_cp_rank_per_document_balancing`).
- Pad `cu_seqlens` to a fixed length in `GPTDataset` so `default_collate` can stack samples with different numbers of documents when `micro_batch_size > 1`.

## Results

8b_inter_document_masking_raw_metrics

8B-scale training runs (300B tokens) comparing baseline vs inter-document masking for both Transformer and Hybrid (Mamba+attention) architectures. Inter-document masking produces lower training and validation loss for both architectures, with no throughput regression. Gradient norms and parameter norms track closely between baseline and masking variants.

## Test plan
- [x] Unit test (`test_inter_document_masking`) verifying `cu_seqlens` values, position ID resets, and `max_seqlen` across 20 samples.
- [x] Unit test (`test_inter_document_masking_batch`) verifying end-to-end `get_batch` with inter-document masking across TP and PP ranks (parametrized over TP={1,2,4} x PP={1,2,4}).
- [x] Unit test (`test_hybrid_cp_batch`) verifying CP partitioning with inter-document masking across TP, PP, and CP ranks.
- [x] Pad/strip round-trip tested via `test_flatten_batch_for_packed_sequences_padded_cu_seqlens` (on main via #5454).
- [x] 8B training runs with `--dataloader-inter-document-masking` comparing loss curves against baseline (see Results above).

### DIFF
--- a/megatron/core/datasets/gpt_dataset.py
+++ b/megatron/core/datasets/gpt_dataset.py
@@ -76,6 +76,10 @@ class GPTDatasetConfig(BlendedMegatronDatasetConfig):
     context_parallel_size: Optional[int] = None
     """The size of the context parallel group. Needed for padding in packed sequences."""
 
+    inter_document_masking: bool = False
+    """When True, return cu_seqlens marking document boundaries within each sample so
+    that attention is restricted to individual documents."""
+
     def __post_init__(self) -> None:
         """Do asserts and set fields post init"""
         super().__post_init__()
@@ -233,9 +237,9 @@ class GPTDataset(MegatronDataset):
         """
         if idx is None:
             # Batch padding sequence so the index does not matter
-            text, _ = self._query_document_sample_shuffle_indices(0)
+            text, _, document_lengths = self._query_document_sample_shuffle_indices(0)
         else:
-            text, _ = self._query_document_sample_shuffle_indices(idx)
+            text, _, document_lengths = self._query_document_sample_shuffle_indices(idx)
 
         text = torch.from_numpy(text).long()
         if self.config.add_extra_token_to_sequence:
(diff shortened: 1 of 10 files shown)


megatron/training/arguments.py
```diff
@@ -1275,6 +1275,15 @@ def validate_args(args, defaults={}):
             'seq-length should be a multiple of 2 * context-parallel-size ' \
             'if context-parallel-size > 1.'
 
+    if getattr(args, 'dataloader_inter_document_masking', False):
+        if args.context_parallel_size > 1:
+            assert False, \
+                '--dataloader-inter-document-masking does not yet support context parallelism > 1.'
+        # The dataset omits attention_mask when inter-document masking is
+        # enabled; disable the flag to avoid a TP broadcast mismatch.
+        if args.create_attention_mask_in_dataloader:
```
asolergi-nv, line 1258: This attention mask functionality is quite old & is only taking effect when using the local spec, not the TE one. Should we drop it at least from GPTDataset?

megatron/core/datasets/gpt_dataset.py
```diff
@@ -279,32 +283,74 @@ def __getitem__(self, idx: Optional[int]) -> Dict[str, torch.Tensor]:
...
+            # equals sequence_length.
+            shortfall = self.config.sequence_length - sum(document_lengths)
+            if shortfall > 0:
+                if document_lengths:
+                    document_lengths[-1] += shortfall
+                else:
+                    document_lengths.append(shortfall)
+            cu_seqlens = torch.tensor(numpy.cumsum([0] + document_lengths), dtype=torch.int32)
+
+            max_seqlen = (cu_seqlens[1:] - cu_seqlens[:-1]).max()
```
asolergi-nv, line 308: As you point out in megatron/training/arguments.py, since we don't yet support CP we can compute this here, but we should move it outside to a helper. In #5017 I'm doing so for [SFT samples](https://github.com/NVIDIA/Megatron-LM/pull/5017/changes#diff-3f0ab2a1c7eafe7bbccfa7c0d065383b30a2232debe6b74251564f775a4ed9bfR2208-R2219), will have to update it to support this new feature

asolergi-nv (approved):
Looks good to me, I resolved some CC comments that you already addressed!

megatron/training/arguments.py
jkamalu: The only hesitation I would have is if people use local and weird masks in their research outside of things TE supports. But I would lean toward dropping it too.

jkamalu (approved):

jaredcasper (approved):

ericharper (approved):

deepakn94:
Note: context parallelism (`--context-parallel-size > 1`) is not yet supported with inter-document masking in pre-training. TE's THD ring attention assumes per-document load balancing, which requires each document length to be divisible by `2 * cp_size`. #5531 adds a guard to assert `context_parallel_size == 1` when inter-document masking is enabled.

PR #5299, opened 2026-06-11
Avoid X11 master port default

### LINKED ISSUE #5232: Please avoid default MASTER_PORT=6000 in example startup scripts
Disclaimer: This issue is created with the help of AI tools. The debugging process described below is true and painful.

**Describe the bug**

Many launch scripts in this repo default to `MASTER_PORT=6000` (numerous `examples/` and `*.sh` wrappers hardcode it). Port `6000` is the well-known TCP port for **X11 display `:0`** — X11 listens on `6000 + display_number`. Defaulting the PyTorch distributed rendezvous port (the rank-0 `TCPStore`) to 6000 creates a class of confusing, hard-to-attribute collisions on any host where X11-over-TCP is in play.

The most common symptom: on a headless GPU node where `/tmp/.X11-unix/X0` does not exist, any X11 client that resolves display `:0` falls back to TCP `localhost:6000`, connects to the running Megatron `TCPStore`, sends the 12-byte X11 connection-setup request, and then hangs forever in `poll()` because the `TCPStore` speaks its own protocol and never replies.

Minimal, no GPUs required.

1. Start a listener on 6000 that accepts but never responds — this is exactly what a live `TCPStore` looks like to a non-c10d client. (Equivalently: launch any Megatron script that uses the default `MASTER_PORT=6000`.)

```bash
python3 -c '
import socket
s = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
s.bind(("::", 6000)); s.listen()
while True:
    c, _ = s.accept()   # accept the connection, then do nothing
'
```

2. On the same host, make any X11 client resolve display `:0`. On a headless node (no `/tmp/.X11-unix/X0`) xcb falls back to TCP automatically:

```bash
DISPLAY=localhost:0 timeout 10 xset q ; echo "exit=$?"   # hangs -> exit=124
```

The client connects to `[::1]:6000`, sends the X11 handshake, and blocks waiting for a server reply that never comes.

**Expected behavior**

The default rendezvous port should not collide with a well-known/reserved port. PyTorch's common convention is `29500` (and `torchrun` uses `29400` for the c10d rendezvous backend). Suggested fixes:

- Change the default `MASTER_PORT` in example/launch scripts to a more random one (or another port outside reserved ranges).
- Avoid the entire `6000–6063` range, which maps to X11 displays `:0`–`:63`.
- At minimum, document prominently that users on shared or X11-capable nodes must override it.

**Additional context**

This cost me a long debugging session before I traced it back here. On a Slurm + headless GPU node with `MpiDefault=pmix`, submitting *any* job would hang in `slurmstepd` (and the step couldn't be killed cleanly); `slurmctld` logged `slurm_send_node_msg ... Unexpected missing socket` and `srun` reported `Zero Bytes were transmitted or received`. A gdb backtrace of the stuck `slurmstepd` showed:

```
poll()
xcb_connect_to_fd()
XOpenDisplay()
hwloc_gl.so                 # hwloc GL discovery plugin
hwloc_topology_load()
pmixp_libpmix_job_set()     # PMIx does its own (unfiltered) hwloc topology load
mpi_p_slurmstepd_prefork()  # mpi_pmix_v5.so
job_manager() / main()
```

Chain of events:

1. A Megatron job launched by another user on the shared node was running with `MASTER_PORT=6000`; its rank-0 `TCPStore` was listening on `[::]:6000`.
2. PMIx init in `slurmstepd` calls `hwloc_topology_load()`. Unlike Megatron/Slurm's own hwloc usage (which filters out I/O devices), PMIx does an unfiltered topology load, so the `hwloc_gl` plugin runs.
3. `hwloc_gl` probes display `:0` via `XOpenDisplay(":0")`. The node has no `/tmp/.X11-unix/X0`, so xcb falls back to TCP `localhost:6000`.
4. It connects to the Megatron `TCPStore`, sends the X11 setup request, gets no reply, and hangs — taking the entire job launch with it.

Because it only reproduces while a `MASTER_PORT=6000` job happens to be live, it's intermittent and very hard to attribute back to Megatron.

Other potential collisions from defaulting to 6000:

- **Multi-tenant bind conflicts:** two jobs on the same node both defaulting to 6000 → `Address already in use`, or — worse — a worker rank silently connecting to the *wrong* job's store.
- **Real X server present:** on workstations, nodes with a display manager, or headless X kept around for GLX/CUDA-GL interop, the `TCPStore` `bind()` fails or X clients misbehave.
- **Firewalls:** the `6000–6063` range is frequently blocked or specially handled by default firewall policies (X11 is a classic remote-attack surface), which can silently break rendezvous on locked-down clusters.

Workaround: set `HWLOC_COMPONENTS=-gl` in the `slurmd` service environment so PMIx/hwloc never probes display `:0`.

Relevant blog: https://harrychen.xyz/2026/06/10/slurm-hung-diagnosis/

guihong-nv:
## Summary
- update Megatron examples, docs, and test helpers to default MASTER_PORT / --master_port to 29500 instead of 6000
- avoid colliding with the X11 port range (6000-6063)

Fixes #5232

## Tests
- git diff --check
- python3 -m py_compile tests/unit_tests/test_utilities.py
- bash -n on edited shell scripts except existing msdp template scripts with placeholder syntax


guihong-nv:
/ok to test [f4c30b5](https://github.com/NVIDIA/Megatron-LM/pull/5299/commits/f4c30b53eaa4989b806bdceaaee5e5a352ab80f2)

yaoyu-33 (approved):

yaox12 (approved):

guihong-nv:
/ok to test [33d4355](https://github.com/NVIDIA/Megatron-LM/pull/5299/commits/33d43550aa9b03b4982660581a613ea799599448)

kvareddy (approved):

balasaajay (approved):

guihong-nv:
/ok to test [42d01b1](https://github.com/NVIDIA/Megatron-LM/pull/5299/commits/42d01b16395b97bd8e9731198bd7539acef7322c)

PR #5306, opened 2026-06-11
Add RL rollout submission and consumption granularity controls

lauradang:

# What does this PR do?

Adds explicit submit/consume granularity controls for Megatron RL partial rollouts. The goal is to let rollout generation stream more eagerly while keeping the trainer-side ordering guarantees required by each consumption mode.

This PR supports three new lag-shape modes:

- `G-submit / B-consume`: submit one rollout group at a time, then consume only complete prompt batches in original submission order.
- `R-submit / B-consume`: submit individual rollouts independently, re-form complete groups, then consume complete prompt batches in original submission order.
- `R-submit / G-consume`: submit individual rollouts independently, re-form complete groups, then consume groups as they become available.

The CLI surface is:

- `--rl-submission-granularity {R,G,B}`
- `--rl-consumption-granularity {G,B}`

`R` consumption is intentionally rejected for now because training still consumes complete rollout groups. `B-submit / G-consume` is also rejected because batch submission cannot provide useful group-level consumption relaxation.

Implementation details:

- Adds `RLRolloutGranularity` for rollout, group, and batch granularity values.
- Updates RL argument validation to derive `rl_generation_batch_size`, `rl_parallel_generation_tasks`, and `rl_enforce_generation_order` from the explicit granularity flags.
- Extends `GroupedRolloutGenerator.get_grouped_rollouts()` so rollout-level submission uses per-rollout semaphore slots while still yielding complete groups to training.
- Keeps batch consumption ordered by buffering completed groups by submission batch and yielding them in original order.
- Threads the new request granularity through the reward-only agent, weighted multi-task agent, FastAPI environment server, and RL rollout utility path.
- Adds unit coverage for argument validation, unsupported combinations, ordered batch consumption, group consumption, and rollout-level submission behavior.

## Issue tracking

Linked issue: N/A

### Pre-checks

- [x] I have added relevant functional/manual validation
- [x] I have added relevant documentation through CLI help text

### Testing

Static/unit validation:

- `uv run --no-sync ruff check megatron/rl/agent/api.py tests/unit_tests/rl/test_grouped_rollouts.py tests/unit_tests/rl/test_rl_utils.py megatron/training/arguments.py`

Completed manual AWS1 validation (AWS1, 2026-06-12):

All five runs used the same Qwen2.5-3B / GSM8K setup with TP=2, PP=1, `grpo_group_size=2`, `grpo_prompts_per_step=2`, `--rl-partial-rollouts`, `--rl-generation-lag 4`, detailed timing enabled (`--timing-log-level 1`, `--timing-log-option minmax`, `--log-timers-to-tensorboard`), and no artificial step cap beyond the 2h SLURM wall time (`--train-samples 48828125`). Runs were submitted from the AWS1 worktree at commit `ad28094afcd6a3296fd31c57b68b97200c1c113d`.

Run matrix:

| Mode | SLURM | W&B | Final state | Iterations | Samples |
|---|---:|---|---|---:|---:|
| B-submit / B-consume | `530913` | https://wandb.ai/adlr/nano-v3-megatron-inference/runs/on4lth88 | SLURM `TIMEOUT` at 2h | 548 | 2192 |
| G-submit / B-consume | `530914` | https://wandb.ai/adlr/nano-v3-megatron-inference/runs/uv3p568f | SLURM `TIMEOUT` at 2h | 354 | 1416 |
| G-submit / G-consume | `530915` | https://wandb.ai/adlr/nano-v3-megatron-inference/runs/cs8b5fc4 | SLURM `TIMEOUT` at 2h | 632 | 2528 |
| R-submit / B-consume | `530916` | https://wandb.ai/adlr/nano-v3-megatron-inference/runs/2ora6o0f | SLURM `TIMEOUT` at 2h | 543 | 2172 |
| R-submit / G-consume | `530917` | https://wandb.ai/adlr/nano-v3-megatron-inference/runs/tvf7gkc1 | Failed after 1:32:55 due W&B artifact logging timeout | 653 | 2612 |

Performance summary, parsed from training logs after warmup iterations:

| Mode | Median step | Mean step | p90 step | p99 step | Max step | Median toks/s | Mean toks/s |
|---|---:|---:|---:|---:|---:|---:|---:|
| B/B | 6.34s | 12.80s | 14.69s | 150.14s | 451.62s | 2584 | 2197 |
| G/B | 6.64s | 19.92s | 20.97s | 353.97s | 495.91s | 2468 | 2052 |
| G/G | 9.17s | 11.11s | 15.02s | 22.78s | 289.97s | 1786 | 1796 |
| R/B | 5.98s | 12.91s | 6.64s | 236.89s | 437.17s | 2740 | 2734 |
| R/G | 6.46s | 7.14s | 9.04s | 11.82s | 18.80s | 2537 | 2377 |

Staleness summary from W&B scalar history:

| Mode | Latest mean/min/max policy staleness | Run mean policy staleness | Run median policy staleness | Max observed policy staleness | Max observed last-token mean staleness | Readout |
|---|---:|---:|---:|---:|---:|---|
| B/B | 4 / 4 / 4 | 3.98 | 4.00 | 4 | 4.00 | bounded exactly at configured lag |
| G/B | 4.5 / 4 / 5 | 4.46 | 4.50 | 5 | 4.50 | bounded around lag 4 plus group boundary |
| G/G | 3 / 2 / 4 | 4.45 | 4.00 | 87 | 22.25 | usually low, with transient stale spikes |
| R/B | 12 / 12 / 12 | 62.39 | 56.00 | 166 | 161.25 | fast median path but builds/drains stale backlog |
| R/G | 5.5 / 4 / 7 | 6.76 | 5.75 | 180 | 45.25 | best throughput/progress before W&B artifact timeout; staleness mostly recovers after spikes |

Validation notes:

- `R-submit / G-consume` had the best wall-clock progress and the tightest step-time distribution, but failed when `wandb_writer.log(...)` attempted to log table/histogram artifacts and W&B timed out waiting for its internal service. The other ranks were terminated by `torchrun` after rank 3 exited.
- `R-submit / B-consume` had the fastest median step time, but the strict batch-order consumer produced large stale backlogs and large tail stalls.
- `B/B` and `G/B` kept staleness bounded, but `G/B` had the weakest wall-clock progress because tail stalls dominated.
- `G/G` was the stable middle option: slower median step than R-level submission, but substantially less pathological than `R/B`.


megatron/training/arguments.py
```diff
@@ -2393,9 +2471,27 @@ def _add_rl_args(parser):
...
+                            'G submits one rollout group at a time. '
+                            'B submits grpo_prompts_per_step rollout groups together. '
+                            'Requires --rl-partial-rollouts. Overrides '
+                            '--rl-generation-batch-size when set to G or B.')
+    group.add_argument('--rl-consumption-granularity', type=RLRolloutGranularity, default=None,
+                       choices=[RLRolloutGranularity.GROUP, RLRolloutGranularity.BATCH],
+                       help='Granularity for consuming generated rollout groups. '
+                            'G consumes groups as they complete. '
+                            'B consumes groups in submission/batch order. '
+                            'Requires --rl-partial-rollouts.')
```
tdene: nit: I would accept the entire `choices=list(RLRolloutGranularity)` set as an argument.

And then in your asserts - where you also assert against B/G - also assert against X / R.

megatron/training/arguments.py
```diff
@@ -548,13 +604,34 @@ def validate_args(args, defaults={}):
...
-        args.rl_enforce_generation_order = (args.rl_generation_batch_size > 1)
+        if args.rl_consumption_granularity == RLRolloutGranularity.BATCH:
+            args.rl_enforce_generation_order = True
+        elif args.rl_consumption_granularity == RLRolloutGranularity.GROUP:
+            args.rl_enforce_generation_order = False
+        else:
+            args.rl_enforce_generation_order = (
+                args.rl_generation_batch_size > 1
+                or (args.rl_partial_rollouts and args.grpo_prompts_per_step > 1)
+            )
```
tdene: We should get rid of `rl_enforce_generation_order` altogether and just pass through `rl_consumption_granularity`.

megatron/training/arguments.py
```diff
@@ -526,6 +534,51 @@ def validate_args(args, defaults={}):
...
+                )
+            args.rl_generation_batch_size = 1
+        elif args.rl_submission_granularity == RLRolloutGranularity.BATCH:
+            if args.rl_generation_batch_size not in (None, args.grpo_prompts_per_step):
+                warn_rank_0(
+                    f"--rl-submission-granularity B overrides --rl-generation-batch-size "
+                    f"({args.rl_generation_batch_size}) with --grpo-prompts-per-step "
+                    f"({args.grpo_prompts_per_step})."
+                )
+            args.rl_generation_batch_size = args.grpo_prompts_per_step
```
tdene: In general, I do not understand why `rl_generation_batch_size` is still an argument. I would not expect it to play a useful role anymore. And it seems like it's being overridden or not supported in most cases.

Can we just drop it? iirc the only thing it was useful for was to support fractional and negative lag. We can re-add support for that in a later PR if we want.

megatron/training/arguments.py
```diff
@@ -54,6 +55,13 @@
 
 from megatron.training.argument_utils import ArgumentGroupFactory, core_transformer_config_from_args
 
+
+def _normalize_rl_rollout_granularity(value):
+    if value is None or isinstance(value, RLRolloutGranularity):
+        return value
+    return RLRolloutGranularity(value)
```
tdene: I don't understand why we would ever want `None` for these arguments. It seems to me like the default must specifically not be `None`, because some choice is needed for the granularity values.

megatron/training/arguments.py
```diff
@@ -526,6 +534,51 @@ def validate_args(args, defaults={}):
...
+        )
+        if submit_rollouts_at_rollout_granularity:
+            assert (
+                args.rl_partial_rollouts
+            ), "Rollout submission granularity requires streaming grouped rollouts."
+
+        explicit_granularity = (
+            args.rl_submission_granularity is not None
+            or args.rl_consumption_granularity is not None
+        )
```
tdene: As mentioned in another comment, I think we should always be explicit? Unless I'm missing a case?

megatron/training/arguments.py
```diff
@@ -534,11 +587,14 @@ def validate_args(args, defaults={}):
         if args.rl_num_parallel_generations is not None:
             assert args.rl_partial_rollouts, \
                 "--rl-num-parallel-generations requires --rl-partial-rollouts."
-            assert args.rl_num_parallel_generations % args.grpo_group_size == 0, \
-                f"--rl-num-parallel-generations ({args.rl_num_parallel_generations}) " \
-                f"must be divisible by --grpo-group-size ({args.grpo_group_size})."
-            args.rl_parallel_generation_tasks = (
-                args.rl_num_parallel_generations // args.grpo_group_size)
+            if submit_rollouts_at_rollout_granularity:
+                args.rl_parallel_generation_tasks = args.rl_num_parallel_generations
```
tdene: Can we please clean up `--rl-num-parallel-generations`, `--rl-num-parallel-generation-batches`, `--rl-parallel-generation-tasks`? We no longer need these separate arguments. They are in part being determined by the granularity arguments? Can we just delete all 3 arguments and have a single argument to specify the lag (in units of batches)?

megatron/rl/agent/api.py
```diff
@@ -200,12 +208,20 @@ def __init__(self, *, parallel_generation_tasks: int | None = None, **kwargs):
...
+    ) -> list[Rollout]:
+        ...
 
     async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
         assert isinstance(
             request.inference_interface, ReturnsRaw
         ), "InferenceInterface must support raw_text return to provide rollouts."
+        submit_at_rollout_granularity = (
+            request.submission_granularity == RLRolloutGranularity.ROLLOUT
+        )
```
tdene, line 215: Instead of having `if/then` branching based on `submit_at_rollout_granularity`, would it make more sense to add a `get_rollout` under `RolloutGenerator` and have `get_grouped_rollouts` call on `get_rollout`?

megatron/rl/agent/api.py
```diff
@@ -33,7 +34,13 @@ class RolloutRequest(Request):
 
 
 class GroupedRolloutRequest(Request):
-    """Request to agent to generate grouped Rollouts."""
+    """Request to agent to generate grouped Rollouts.
+
+    num_groups controls how many RolloutGroup objects are submitted together.
+    The trainer consumes grpo_prompts_per_step yielded groups per rollout
+    collection. When enforce_order is true, yielded groups preserve submission
+    order even if generation completes out of order.
```
tdene: Should we get rid of `num_groups` as well and just make decisions off the granularity arguments alone?

lauradang: What about if we changed the name to groups_per_batch for clarity? The user should still be able to control the number of groups when we are on batch submission mode

tdene: Isn't `grpo_group_size` sufficient?

tdene: I see now that `num_groups` is just set `grpo_group_size`. And yeah, that information is necessary.

This comment is irrelevant. I was thinking of some old logic that gated specifically on `num_groups==1`. That no longer exists.

megatron/rl/agent/api.py
tdene: Discussed offline; this does not apply

megatron/rl/agent/api.py
```diff
@@ -257,11 +266,18 @@ async def generate_task():
                         for i in range(groups_per_worker)
                     ])
                 else:
-                    if not await generate_and_enqueue(batch_id, 0):
+                    if consume_at_batch_granularity:
+                        while not await generate_and_enqueue(batch_id, 0):
+                            pass
+                    elif not await generate_and_enqueue(batch_id, 0):
```
tdene, line 281: If I'm understanding this code correctly, we are still generating whole groups at a time; we await the generation of an entire group.

So if we're in rollout submission granularity, it looks like we still need to wait for the entire group to complete before we resubmit new requests. Because this method is the only one that submits new requests to the inference engine, and it only does so after `await generate_and_enqueue`.

lauradang: get_grouped_rollouts() passes the semaphore into group_rollout() which acquires the slot around each individual agenerate() call. So new rollout requests can be submitted as individual rollout slots free up. we do not wait for the whole group to finish before submitting more inference work.

-- All in all, the downstream trainer still sees complete groups, even though R submission granularity lets the individual agenerate() calls inside that group be submitted independently

tdene (approved):

PR #5307, opened 2026-06-11
Add optimizer parameter dump helper for DTensor state

wujingyue:
## Summary

Adds `megatron/core/optimizer/utils.py` with `dump_optimizer_parameters(optimizer, dump_dir)`, a small helper that writes per-rank JSON snapshots of DTensor optimizer parameter sharding.

The dump is intended for reproducing optimizer layouts in tests and microbenchmarks without re-running full model setup. It captures the data needed to reconstruct the parameter layout: global/local shape, dtype, QKV flag, device mesh shape/names/ranks, placements, and optimizer param-group structure.

Training scripts can opt in with:

```bash
--optimizer-parameters-dump-dir /path/to/dump
```

Each rank writes one file under that directory:

```text
/path/to/dump/rank0.json
/path/to/dump/rank1.json
...
```

## Design points

- The helper lives under `megatron/core/optimizer/` because it operates on optimizer param groups, not on distributed wrappers.
- It is optimizer-agnostic: it works with any PyTorch optimizer over DTensor params.
- It is DTensor-only; non-DTensor params fail naturally at `to_local()`.
- Param-group structure is preserved under top-level `groups`, but group hyperparameters such as `lr`, `betas`, and `weight_decay` are intentionally not recorded.
- `mesh_ranks` is omitted for canonical full-world meshes and recorded only when needed to reconstruct non-canonical device assignment.
- The CLI takes a directory, not a file path, so rank filenames are deterministic and the user-provided path is not rewritten.

## Output schema

```json
{
  "world_size": 8,
  "rank": 0,
  "groups": [
    {
      "params": [
        {
          "global_shape": [35072, 8192],
          "dtype": "torch.bfloat16",
          "is_qkv": false,
          "local_shape": [6432, 8192],
          "mesh_shape": [4, 2],
          "mesh_dim_names": ["dp_cp", "tp"],
          "mesh_ranks": [0, 1, 2, 3, 4, 5, 6, 7],
          "placements": ["Shard(dim=0)", "Shard(dim=0)"]
        }
      ]
    }
  ]
}
```

## Test plan

- [x] `git diff --check`
- [x] `uv run --no-sync python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/megatron_fsdp/test_dump_parameters.py`
- [x] `uv run --no-sync python -m py_compile megatron/core/optimizer/utils.py megatron/training/arguments.py megatron/training/training.py tests/unit_tests/distributed/megatron_fsdp/test_dump_parameters.py`

All commits are signed off.


megatron/core/optimizer/utils.py
```diff
@@ -0,0 +1,96 @@
...
+from typing import Any
+
+import torch
+import torch.distributed as dist
+from torch.distributed.tensor import DTensor
+
+logger = logging.getLogger(__name__)
+
+
+def dump_optimizer_parameters(
```
yashaswikarnati, line 38: seems like helper function, would these need to go into core ?

wujingyue: I guess "need" is a strong word :)

My other PR https://github.com/NVIDIA/Megatron-LM/pull/5078 currently relies on checked-in dumped files. Given how simple the data collection/generation process is, I think it would be much better to check in not only the test data itself, but also the script or procedure used to generate it. That would make the test artifacts easier to understand, reproduce, and update in the future.

What do you think? cc @yashaswikarnati

megatron/training/training.py
```diff
@@ -2060,6 +2061,17 @@ def setup_model_and_optimizer(
         )
         opt_param_scheduler = get_optimizer_param_scheduler(optimizer)
 
+        # Optional: dump per-rank optimizer parameter sharding for fixture capture.
+        # Gated by `DUMP_OPTIMIZER_PARAMETERS_OUTPUT` env var; each rank writes a
+        # JSON snapshot under that path. Optimizer-agnostic — fires regardless
+        # of optimizer class. See `megatron.core.optimizer.utils` for the schema.
+        dump_path = os.environ.get("DUMP_OPTIMIZER_PARAMETERS_OUTPUT")
```
deepakn94: Can you make this a command-line argument?

yaoyu-33 (approved):

PR #5312, opened 2026-06-12
Add support for non-Gym multi-turn environments

tdene:

# What does this PR do ?

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


megatron/rl/agent/reward_only_agent.py
```diff
@@ -86,46 +88,132 @@ def _get_rank_subset(
...
+        responses: list[InferenceResponse] = []
+
+        for turn_idx in range(self.max_turns):
+            turn_request = request.inference_interface.prepare_request(
+                conversation, request.generation_args
+            )
+            # Adopt the request's prompt as the conversation: turn 0 may start from a bare
+            # string, which prepare_request normalizes into a single user message.
+            conversation = list(turn_request.prompt)
+            responses.append(await self.get_rollout_response(request, turn_request))
```
lauradang: Since _run_episode now launches multiple rollouts, wouldn't that cause the num_inference_workers that we promise to the user to be different? It's as if we are now hiding more rollouts within one worker

megatron/rl/agent/reward_only_agent.py
```diff
@@ -86,46 +88,132 @@ def _get_rank_subset(
...
+
+            observation, done = await self.get_observation(
+                turn_idx, responses[-1], conversation, golden
+            )
+            if done or observation is None:
+                break
+            assert observation, (
+                "get_observation must return a non-empty observation when continuing (done=False)."
+            )
+            conversation = conversation + [
```
lauradang: I think conversation += ... works here

lauradang: in extension, this might be easier to read:


```suggestion
for turn_idx in range(self.max_turns):
    turn_request = request.inference_interface.prepare_request(
        conversation, request.generation_args
    )
    conversation = list(turn_request.prompt)

    response = await self.get_rollout_response(request, turn_request)
    responses.append(response)

    if turn_idx + 1 < self.max_turns:
        observation, done = await self.get_observation(
            turn_idx, response, conversation, golden
        )
        if done:
            break
        if not observation:
            raise ValueError(
                "get_observation must return a non-empty observation"
            )

        conversation += [
            response.response,
            LLMChatMessage(role="user", content=observation),
        ]
```

lauradang (changes requested):

lauradang:
Looks like some of your unit tests are failing

megatron/rl/rl_utils.py
```diff
@@ -1285,17 +1309,27 @@ def maybe_log_training_metrics(
     wandb_writer.log(metrics, step=current_iteration)
 
 
+PAD_TURN_UNIT = (None, -1)
```
lauradang, line 1312: Just a q: "Used to equalize per-rank trajectory counts without contributing to the loss" - Can I have more context why this is needed? What was the before/after effect of adding this

megatron/rl/agent/reward_only_agent.py
tdene: That is true, and a serious concern. `_run_episode` will launch multiple sequential rollouts, meaning that when a rollout reserves an inference slot and acquires the semaphore, it doesn't necessarily use the inference slot the whole time. This drives down utilization.

It does not lead to unexpected overuse of the inference engine because all of the inference calls in `_run_episode` are sequential.

However, this design choice is already the current way that we are running RL through NeMo Gym. This change here just makes it explicit, and enables it for non-Gym runs (which are not actually used in practice).

So I'm not making a new design choice, I'm just preserving an existing one. Should this be changed? Maybe, we need to discuss in-depth. Seems out-of-scope for this PR though.

megatron/rl/agent/reward_only_agent.py
tdene: Suggestion applied!

megatron/rl/rl_utils.py
tdene: Training is split across DP groups, and all DP groups must process an identical amount of microbatches to make sure the torch collectives don't hang.

Before this PR, it was easy to split the data up across DP groups, since the number of rollouts was divisible by design.

After this PR, the data consists of turns, and the number of turns is non-deterministic and arbitrary. Trying to divide the turns among DP groups may leave us with unfilled slots. We need to pad those or torch collectives hang. So we choose to pad them with a value that will not affect training.

megatron/rl/agent/reward_only_agent.py
lauradang: Yeah that's a good point that this issue already existed in Nemo gym. agreed out of scope.. maybe the env balancing logic can help with this a bit

lauradang (approved):

PR #5328, opened 2026-06-12
Thread pg_collection through wrap_model_chunks_with_ddp

yashaswikarnati:
Lets a parallel_state-free caller (hetero MIMO) reuse the stock DDP-wrap helper — forwards pg_collection to standard DDP and sources the dist-opt layout DP world sizes from the pgc; default None preserves mpu behavior byte-for-byte. Testing: real cog 8-GPU. Enables MM2 (#5285) to drop its hand-rolled DDP wrap.

### DIFF
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -1584,17 +1584,17 @@ def wrap_model_chunks_with_ddp(
         else:
             compute_layout = None
         if compute_layout is not None:
-            if pg_collection is not None:
-                # Source DP world sizes from the explicit pg_collection.
-                data_parallel_world_size = get_pg_size(pg_collection.dp_cp)
-                expert_data_parallel_world_size = get_pg_size(
-                    getattr(pg_collection, "expt_dp", None)
-                )
-            else:
-                data_parallel_world_size = mpu.get_data_parallel_world_size(
-                    with_context_parallel=True
-                )
-                expert_data_parallel_world_size = mpu.get_expert_data_parallel_world_size()
+            layout_pgs = (
+                pg_collection
+                if pg_collection is not None
+                else ProcessGroupCollection.use_mpu_process_groups()
+            )
+            assert layout_pgs.dp_cp is not None, (
+                "wrap_model_chunks_with_ddp requires a dp_cp process group to size "
+                "the distributed-optimizer parameter layout"
+            )
+            data_parallel_world_size = get_pg_size(layout_pgs.dp_cp)
+            expert_data_parallel_world_size = get_pg_size(getattr(layout_pgs, "expt_dp", None))
             for i, (chunk, bucket_size) in enumerate(zip(model_chunks, bucket_sizes)):
                 all_params = [p for p in chunk.parameters() if p.requires_grad]
                 per_chunk_layouts[i] = compute_layout(
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -1551,8 +1551,9 @@ def wrap_model_chunks_with_ddp(
             to DDP. ``False`` keeps LayerWise on its legacy sync path.
         DP: The DDP class to construct (``DistributedDataParallel`` or an FSDP
             variant).
-        pg_collection: Optional :class:`ProcessGroupCollection`. Forwarded to
-            FSDP-style DPs only.
+        pg_collection: Optional :class:`ProcessGroupCollection`. When provided,
+            forwarded to both standard DDP and FSDP variants, and used to source
+            DP world sizes for layout computation.
         bucket_sizes: Optional per-chunk bucket size override; defaults to
             ``[ddp_config.bucket_size] * len(model_chunks)``.
         disable_bucketing_per_chunk: Optional per-chunk disable_bucketing flag;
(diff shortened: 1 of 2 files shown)


tests/unit_tests/test_wrap_model_chunks_with_ddp.py
```diff
@@ -0,0 +1,79 @@
...
+
+class TestWrapModelChunksWithDDP:
+    def teardown_method(self, method):
+        Utils.destroy_model_parallel()
+
+    def test_explicit_pg_collection_forwarded_to_ddp(self):
+        """An explicit pg_collection is forwarded to standard DDP."""
+        from megatron.training.training import wrap_model_chunks_with_ddp
+
+        Utils.initialize_model_parallel(
```
yashaswikarnati: still initializing parallel state ? not custom process group with hyper comm grid ? if the test can be concisely writtern we should prefer that

megatron/training/training.py
```diff
@@ -1603,7 +1611,8 @@ def wrap_model_chunks_with_ddp(
         model_chunks, per_chunk_layouts, disable_bucketing_per_chunk
     ):
         chunk_kwargs = {}
-        if pg_collection is not None and DP is not DDP:
+        if pg_collection is not None:
```
yashaswikarnati: does this break exisitng FSDP paths ?

megatron/training/training.py
```diff
@@ -1583,10 +1584,17 @@ def wrap_model_chunks_with_ddp(
         else:
             compute_layout = None
         if compute_layout is not None:
-            data_parallel_world_size = mpu.get_data_parallel_world_size(
-                with_context_parallel=True
-            )
-            expert_data_parallel_world_size = mpu.get_expert_data_parallel_world_size()
+            if pg_collection is not None:
```
yashaswikarnati: none fallback is dangerous, the pattern thosuld be if pg collecton is none then pg collectin initialize from use mpu process grpoups, if not assert we need dp cp and expt dp and use it

yashaswikarnati: Fixed in 75a75f1. Replaced the None-vs-explicit branch with a single pg_collection path: when the caller passes None we materialize from `ProcessGroupCollection.use_mpu_process_groups()`, then assert `dp_cp` before use.

```python
layout_pgs = pg_collection if pg_collection is not None else ProcessGroupCollection.use_mpu_process_groups()
assert layout_pgs.dp_cp is not None, "...requires a dp_cp process group..."
data_parallel_world_size = get_pg_size(layout_pgs.dp_cp)
expert_data_parallel_world_size = get_pg_size(getattr(layout_pgs, "expt_dp", None))
```

I kept `expt_dp` optional rather than asserted: dense (non-MoE) models have no expert-DP group and `get_pg_size(None)` returns 1, so asserting it would break every dense model. I also scoped this to a local `layout_pgs` (did not reassign the outer `pg_collection`) so the default path still forwards nothing to DDP — see the FSDP thread.

megatron/training/training.py
yashaswikarnati: No — FSDP is unaffected. Both standard DDP and the FSDP variants resolve their groups via `ProcessGroupCollection.setup_process_groups_for_ddp(pg_collection, ...)` inside `__init__`, which already handles `pg_collection=None` by falling back to `parallel_state`. This forward is gated on `is not None`, so every existing caller (who passes None) forwards nothing and DDP/FSDP take their unchanged internal None path. FSDP previously received the pgc whenever provided and still does — byte-identical. The only new behavior is forwarding an *explicitly provided* pgc to standard DDP, which it accepts. That is also why I kept the layout sizing on a local `layout_pgs` and did not reassign the outer variable.

tests/unit_tests/test_wrap_model_chunks_with_ddp.py
yashaswikarnati: Rewritten in 75a75f1. The forwarding test now builds the pgc from a `HyperCommGrid` with `Utils.initialize_distributed()` only (no `initialize_model_parallel`), and asserts `chunk.dp_cp_group is dp` against the grid-derived group:

```python
grid = HyperCommGrid([1, 1, 1, 8], ["tp", "pp", "ep", "dp"], backend="nccl")
dp = grid.create_pg("dp")
pg_collection = ProcessGroupCollection(tp=grid.create_pg("tp"), pp=grid.create_pg("pp"), ep=grid.create_pg("ep"), dp=dp, dp_cp=dp, expt_dp=dp)
```

(tp/pp/ep are required by `setup_process_groups_for_ddp`, so they are built at size 1.) The `pg_collection=None -> MPU` back-compat case is isolated in its own small `TestDefaultPath`. Verified per-rank on 8 GPUs: 2 passed on every rank.

megatron/training/training.py
maanug-nv: TorchFSDP doesn't take pg collection afaik? i guess that's a pre-existing bug

maanug-nv (approved):

deepakn94 (approved):

yaoyu-33 (approved):

megatron/training/training.py
yashaswikarnati: Good catch — confirmed. `TorchFullyShardedDataParallel.__init__` takes a single `process_group`, not a `ProcessGroupCollection` (0 `pg_collection` references in `torch_fully_sharded_data_parallel.py`, vs the DDP and Megatron-FSDP adapter which both accept it). It is pre-existing and latent: the original `DP is not DDP` guard would forward `pg_collection` to TorchFSDP, but `get_model` passes `pg_collection=pg_collection if args.use_megatron_fsdp else None` (training.py:1828), so TorchFSDP never actually receives a non-None value from any caller.

Fixed in 04f6fc4 by restricting the forward to the two DPs that accept it:

```python
# TorchFSDP takes a single process_group, not a ProcessGroupCollection;
# forward pg_collection only to DDP and the Megatron-FSDP adapter.
if pg_collection is not None and not (HAVE_FSDP2 and DP is torch_FSDP):
    chunk_kwargs["pg_collection"] = pg_collection
```

yashaswikarnati:
/ok to test [01df48b](https://github.com/NVIDIA/Megatron-LM/pull/5328/commits/01df48b9ad7f447fbca60ac09f4d60b02f3821af)

PR #5332, opened 2026-06-13
Add optimizer-provider and pg_collection hooks to setup_model_and_optimizer

yashaswikarnati:
Adds two optional hooks to setup_model_and_optimizer:
- **optimizer_provider_func** (defaults to get_megatron_optimizer) so a caller can inject a custom optimizer factory (e.g. MIMO per-module MimoOptimizer).
- **pg_collection** forwarded to get_model.

Both default to the current behavior => byte-for-byte unchanged for existing callers. Enables reusing stock pretrain()/setup_model_and_optimizer for heterogeneous MIMO.

Part of the NMFW-516 MIMO-on-stock-trainloop series.

cog: per-rank verified green on cw-dfw 8-GPU.

### DIFF
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -1957,6 +1957,7 @@ def setup_model_and_optimizer(
     model_provider_func,
     model_type,
     checkpointing_context=None,
+    pg_collection=None,
 ):
     """Setup model and optimizer."""
     args = get_args()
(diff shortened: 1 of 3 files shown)


megatron/training/training.py
```diff
@@ -1957,19 +1957,28 @@ def setup_model_and_optimizer(
     model_provider_func,
     model_type,
     checkpointing_context=None,
+    optimizer_provider_func=None,
```
yashaswikarnati: instead of having a provider function, can we instead change get_megatron_optimizer to handle mimo training ?

yashaswikarnati: Done — dropped the provider hook. `get_megatron_optimizer` now dispatches a single `MimoModel` chunk to `get_mimo_optimizer` directly (lazy import to avoid the circular dep, since get_mimo_optimizer already calls get_megatron_optimizer per module). `setup_model_and_optimizer` keeps only the `pg_collection`→`get_model` forwarding. (b3913cc9)

yashaswikarnati:
Folded into #5333 (combined the optimizer dispatch + setup pg_collection into the schedule-plumbing PR, and slimmed the test). Closing in favor of #5333.

PR #5333, opened 2026-06-13
Thread MIMO support through the stock training loop (schedule + optimizer)

yashaswikarnati:
Threads heterogeneous (cross-grid) MIMO support through the stock training loop, default-None so non-MIMO callers are byte-for-byte unchanged. Part of the NMFW-516 MIMO-on-stock-trainloop series. **Folds in #5332.**

Two atomic commits:
1. **Schedule plumbing** — `train_step`/`train` thread an optional `p2p_communicator` and `schedule_pg_collection` (a `MultiModuleProcessGroupCollection`) into `forward_backward_func` (the core schedule already accepts both). `train()` also passes its per-module `model_pg_collection` to `train_step` for the post-step reductions.
2. **Optimizer dispatch** — `get_megatron_optimizer` routes a single `MimoModel` chunk to `get_mimo_optimizer` (lazy import, no circular dep); `setup_model_and_optimizer` forwards an optional `pg_collection` to `get_model`.

cog: per-rank verified green on cw-dfw 8-GPU (3 passed x 8 ranks).

### DIFF
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -1008,9 +1008,9 @@ def get_megatron_optimizer(
     if isinstance(model_chunks[0], MimoModel):
         from megatron.core.models.mimo.optimizer import get_mimo_optimizer
 
-        assert len(model_chunks) == 1, (
-            "MimoModel does not support virtual pipeline parallelism (multiple model chunks)"
-        )
+        assert (
+            len(model_chunks) == 1
+        ), "MimoModel does not support virtual pipeline parallelism (multiple model chunks)"
         return get_mimo_optimizer(model_chunks[0], config)
 
     # None → apply standard defaults. To extend defaults with custom overrides,
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -1002,6 +1002,17 @@ def get_megatron_optimizer(
         Instance of MegatronOptimizer.
     """
 
+    # A MimoModel routes to the heterogeneous per-module optimizer builder.
+    from megatron.core.models.mimo.model.base import MimoModel
+
+    if isinstance(model_chunks[0], MimoModel):
+        from megatron.core.models.mimo.optimizer import get_mimo_optimizer
+
+        assert len(model_chunks) == 1, (
+            "MimoModel does not support virtual pipeline parallelism (multiple model chunks)"
+        )
+        return get_mimo_optimizer(model_chunks[0], config)
+
     # None → apply standard defaults. To extend defaults with custom overrides,
     # start from get_standard_config_overrides(config) and merge yours in.
     if config_overrides is None:
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -1961,6 +1961,7 @@ def setup_model_and_optimizer(
     model_provider_func,
     model_type,
     checkpointing_context=None,
+    pg_collection=None,
 ):
     """Setup model and optimizer."""
     args = get_args()
@@ -1973,7 +1974,9 @@ def setup_model_and_optimizer(
     has_rl_optimizer = args.perform_rl_step and not args.no_load_optim
     skip_optimizer = not (has_normal_optimizer or has_rl_optimizer)
     wrap_with_ddp = not skip_optimizer
-    model = get_model(model_provider_func, model_type, wrap_with_ddp=wrap_with_ddp)
+    model = get_model(
+        model_provider_func, model_type, wrap_with_ddp=wrap_with_ddp, pg_collection=pg_collection
+    )
     unwrapped_model = unwrap_model(model)
 
     if args.logits_save_dir is not None:
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -156,7 +156,11 @@ from megatron.core.pipeline_parallel.utils import (
     is_vp_first_stage,
     is_vp_last_stage,
 )
-from megatron.core.process_groups_config import ProcessGroupCollection
+from megatron.core.pipeline_parallel.p2p_communication import P2PCommunicator
+from megatron.core.process_groups_config import (
+    MultiModuleProcessGroupCollection,
+    ProcessGroupCollection,
+)
 from megatron.core.transformer.cuda_graphs import TECudaGraphHelper
 from megatron.core.transformer.module import Float16Module
 from megatron.core.transformer.moe.paged_stash import PagedStashRunner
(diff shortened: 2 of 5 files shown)


megatron/training/training.py
```diff
@@ -3238,6 +3246,11 @@ def train(
 
     model_pg_collection = get_attr_wrapped_model(model[0], "pg_collection")
 
+    # Optional heterogeneous (cross-grid) schedule plumbing. For ordinary models these
+    # attributes do not exist, so both resolve to None and behavior is byte-identical.
+    schedule_pg_collection = getattr(unwrap_model(model[0]), "schedule_pg_collection", None)
```
yashaswikarnati: why are we setting these as attributes of the model ?

tests/unit_tests/training/test_train_step_schedule_plumbing.py
```diff
@@ -0,0 +1,128 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
yashaswikarnati, line 1: do we have an existring subfolder for training related unit tests ?

megatron/training/training.py
yashaswikarnati: Good call — removed. Instead of reading these off the model, `train()` now takes explicit `p2p_communicator=None, schedule_pg_collection=None` params and passes them to `train_step` (both default None → byte-identical for non-MIMO). The MIMO entry calls `train()` directly with them. (a754b148)

tests/unit_tests/training/test_train_step_schedule_plumbing.py
yashaswikarnati: Yes — moved to `tests/unit_tests/training/test_train_step_schedule_plumbing.py`.

megatron/training/training.py
```diff
@@ -2195,11 +2195,17 @@ def dummy_train_step(data_iterator):
             )
 
 
-def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_scheduler, config, forward_backward_func, iteration=None, pg_collection: Optional[ProcessGroupCollection] = None):
+def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_scheduler, config, forward_backward_func, iteration=None, pg_collection: Optional[ProcessGroupCollection] = None, p2p_communicator=None, schedule_pg_collection=None):
```
yashaswikarnati: is schedule pg collection has process groups that are subset of pg_colelction? if so can we derive ?

megatron/training/training.py
```diff
@@ -3572,7 +3586,8 @@ def trace_handler(p):
                 num_zeros_in_grad,
                 max_attention_logit,
             ) = train_step(
-                forward_step_func, train_data_iterator, model, optimizer, opt_param_scheduler, config, forward_backward_func, iteration=iteration
+                forward_step_func, train_data_iterator, model, optimizer, opt_param_scheduler, config, forward_backward_func, iteration=iteration,
```
yashaswikarnati, line 3596: we are not passing pg collection to train step ? how is this even working ?

megatron/training/training.py
yashaswikarnati: Good question — I checked the schedule, and they cannot be unified. When the `p2p_communicator` is a `MultiModulePipelineCommunicator` (the MIMO cross-grid bridge), `forward_backward_pipelining_without_interleaving` takes the `is_multimodule` path and calls `pg_collection.has_language_model()` / `get_language_model_cp_size()` (schedules.py:2164-2174) — methods that exist only on `MultiModuleProcessGroupCollection`, not on a plain `ProcessGroupCollection`. So the schedule genuinely needs the MMPGC.

The relationship is the reverse of a subset: `schedule_pg_collection` (MMPGC, per-module dict) is the *richer* object; the per-module plain `pg_collection` the reductions use (`.mp/.dp_cp/.pp`) could be derived *from* the MMPGC, but not the other way around. Deriving it inside train_step would couple the stock reduction code to the MMPGC API and is ambiguous in the colocated case (two modules per rank), so I kept them as two explicit args.

megatron/training/training.py
yashaswikarnati: You are right — that was a real gap. `train()` computes `model_pg_collection` at the top (line ~3197) but was not passing it to `train_step`, so the post-step reductions fell back to `use_mpu_process_groups()` — fine for standard runs but broken for MIMO (no mpu globals). Fixed: `train()` now passes `pg_collection=model_pg_collection` to `train_step` (commit 2e35676ce3). For standard runs `model_pg_collection` is the mpu-derived collection, so behavior is unchanged.

yashaswikarnati: Added type hints (commit 44a24813): `p2p_communicator: Optional[P2PCommunicator]` and — you are right — `schedule_pg_collection: Optional[MultiModuleProcessGroupCollection]` (not a plain `ProcessGroupCollection`; the cross-grid schedule path needs the per-module groups the MMPGC exposes). Fixed the docstring accordingly and imported both types. Imports added manually rather than via isort, since this file's import block is already isort-dirty upstream and a full isort run would reformat unrelated lines. cog re-verified green per-rank.

megatron/core/optimizer/__init__.py
```diff
@@ -1002,6 +1002,14 @@ def get_megatron_optimizer(
         Instance of MegatronOptimizer.
     """
 
+    # A MimoModel routes to the heterogeneous per-module optimizer builder.
+    from megatron.core.models.mimo.model.base import MimoModel
+
+    if len(model_chunks) == 1 and isinstance(model_chunks[0], MimoModel):
```
yashaswikarnati: why we have  if len(model_chunks) == 1

yashaswikarnati: Removed it — it was redundant defensiveness. `get_megatron_optimizer` already indexes `model_chunks[0]` unguarded throughout (e.g. lines 819, 826, 1026, 1056), and a `MimoModel` is a single composite model that is never pipeline-chunked, so it always arrives as the sole chunk. Simplified to `if isinstance(model_chunks[0], MimoModel):` (commit 19217f9f), consistent with how the rest of the function treats the chunk list.

yaoyu-33 (approved):

megatron/core/optimizer/__init__.py
```diff
@@ -1002,6 +1002,14 @@ def get_megatron_optimizer(
         Instance of MegatronOptimizer.
     """
 
+    # A MimoModel routes to the heterogeneous per-module optimizer builder.
+    from megatron.core.models.mimo.model.base import MimoModel
+
+    if isinstance(model_chunks[0], MimoModel):
+        from megatron.core.models.mimo.optimizer import get_mimo_optimizer
+
+        return get_mimo_optimizer(model_chunks[0], config)
```
deepakn94, line 1014: Why pass in only the first chunk?

yashaswikarnati: for now VPP is not supported yet, added an assert

deepakn94 (approved):

yashaswikarnati:
/ok to test [95de128](https://github.com/NVIDIA/Megatron-LM/pull/5333/commits/95de128a0f45f0dc56fa55fee4d95156508cb5f6)

maanug-nv (approved):

Issue #5174, opened 2026-06-05
[BUG] `torch_dist` checkpoint save fails for NVFP4 + distributed Adam optimizer with out-of-bounds optimizer param shard


fushigu:
## Describe the bug

Saving a `torch_dist` distributed checkpoint with optimizer state fails when using NVFP4 mixed precision together with the standard distributed optimizer (`--use-distributed-optimizer`) and Adam/AdamW.

The failure happens during PyTorch Distributed Checkpoint global-plan validation. The optimizer param shard for an NVFP4 parameter is rejected as an out-of-bounds chunk:

```text
ValueError: Failed to validate global plan:
key: optimizer.distributed.dp_group_idx_0.gbuf_idx_1.dtype_(torch.uint8, torch.float32).bucket_idx_0.param
has out of bounds chunk:
tensor-size: torch.Size([45613056])
chunk: ChunkStorageMetadata(
    offsets=torch.Size([45613056]),
    sizes=torch.Size([11403264])
)
...
```

This appears to be specific to NVFP4-packed parameters and the standard distributed optimizer checkpoint path. I do not see the same failure with BF16/FP8, and I was able to save/resume NVFP4 checkpoints when using `dist_muon`, which appears to use a different sharded optimizer-state path.

## Suspected root cause

The failure appears to come from a unit mismatch in the `dp_reshardable` optimizer-state sharding path, likely around:

```text
megatron/core/optimizer/distrib_optimizer.py
DistributedOptimizer.sharded_param_state_dp_reshardable
```

For NVFP4 parameters, the saved optimizer param tensor is represented as packed `uint8` storage. Since NVFP4 stores two FP4 values per byte, the packed parameter element count is roughly half of the corresponding unpacked element count.

However, the `ShardedTensor` metadata appears to mix two different unit systems:

* the global tensor size is derived from the packed `uint8` parameter storage size, e.g. `bucket.numel_unpadded`;
* the per-DP-rank offset/stride appears to be derived from the unpacked grad/main-param element count, e.g. `grad_data.numel() // data_parallel_world_size`.

As a result, for DP ranks in the latter half of the data-parallel group, the computed chunk offset can exceed the declared packed global tensor size. In the observed failure:

```text
tensor-size = 45613056
chunk offset = 45613056
chunk size   = 11403264
```

The chunk starts exactly at the declared tensor end, so `offset + size > tensor_size`, and PyTorch DCP correctly rejects the global plan.

## Steps to reproduce

[...]

PR #5334, opened 2026-06-14
Fix NVFP4 out-of-bounds optimizer shard in dp_reshardable checkpoint

CodersAcademy006:
## What

Fixes the `torch_dist` checkpoint save failure in #5174. Saving optimizer state with NVFP4 mixed precision plus the distributed optimizer fails PyTorch DCP global-plan validation with an out-of-bounds chunk.

## Root cause

In `DistributedOptimizer.sharded_param_state_dp_reshardable`, every optimizer-state `ShardedTensor` is declared with a packed (uint8) global size and packed per-tensor local offsets (`gbuf_local_start`/`gbuf_local_end`). The per-rank shard stride `gbuf_local_numel` is instead derived from the unpacked grad buffer size (`bucket.grad_data.numel()`). NVFP4 packs two FP4 values per byte, so the unpacked stride ends up twice the packed one. DP ranks past the midpoint then compute `data_parallel_rank * gbuf_local_numel + gbuf_local_start` offsets that run past the declared packed global size:

```
tensor-size: torch.Size([45613056])
chunk: offsets=[45613056], sizes=[11403264]
```

## Fix

When the buffer holds NVFP4 parameters, compute the shard stride from the packed bucket size (`nvfp4_packed_bucket_indices`) so it lines up with the packed global size and the local offsets. Buffers without NVFP4 params keep the old path unchanged (guarded by `has_nvfp4_params`).

## Testing

Non-NVFP4 paths are guarded out, so existing checkpoint tests are unaffected. I don't have NVFP4 (Blackwell) hardware, so I haven't been able to run the save/resume repro myself. @fushigu @dimapihtar if either of you can confirm this clears the DCP error and that load round-trips, I'll follow up on anything else needed.

Opened as draft per CONTRIBUTING.


CodersAcademy006:
Following up on this one, which has had no human comment since it was opened on 14 June and is still sitting behind the copy-pr-bot gate, so NVIDIA CI has never run on it. It is a narrow fix: with NVFP4, `dp_reshardable` checkpoint save computes an optimizer shard range that can run past the end of the local parameter buffer, so a resume at a different data-parallel size reads out of bounds rather than failing.

@balasaajay @maanug-nv apologies for the cold ping, you are the two I see closest to the checkpoint and CI surface at the moment. If this needs a vetter `/ok to test`, or if the NVFP4 sharding path has been reworked enough in the last three months that the fix no longer applies, I am happy to rebase it onto current main or close it, whichever is more useful to you.

PR #5338, opened 2026-06-14
fix(tensor_parallel): _reduce returns unreduced tensor for non-contig…

### LINKED ISSUE #5337: _reduce silently returns unreduced tensor when input is non-contiguous
**Describe the bug**

@NVIDIA/mcore-oncall

In `megatron/core/tensor_parallel/mappings.py`, function `_reduce` silently returns the original unreduced tensor when the input is non-contiguous. 

Current code:

```python
def _reduce(input_, group):
    """All-reduce the input tensor across model parallel group."""
    assert group is not None, "group should not be None"

    # Bypass the function if we are using only 1 GPU.
    if group.size() == 1:
        return input_

    # All-reduce.
    torch.distributed.all_reduce(input_.contiguous(), group=group)

    return input_
```

`tensor.contiguous()` returns `self` if the tensor is already contiguous (same storage, no copy), but returns a **new tensor with independent storage** if it is not. When `input_` is non-contiguous:

1. `input_.contiguous()` creates `temp` — a new contiguous tensor
2. `all_reduce(temp)` all-reduces `temp` in-place ✓
3. `return input_` returns the **original, never-reduced tensor** ✗

The presence of `.contiguous()` implies the function is designed to handle non-contiguous inputs gracefully, but the actual behavior is to silently return unreduced data in that case.

While this bug is not triggered by any current call path (**see Additional context**), it can be demonstrated with the following test:

```python
import torch
import torch.distributed as dist

def _reduce(input_, group):
    """All-reduce the input tensor across model parallel group."""
    assert group is not None, "group should not be None"

    # Bypass the function if we are using only 1 GPU.
    if group.size() == 1:
        return input_

    # All-reduce.
    torch.distributed.all_reduce(input_.contiguous(), group=group)

    return input_

def main():
    dist.init_process_group(backend="gloo")
    rank = dist.get_rank()
    group = dist.group.WORLD

    if rank == 0:
        input_tensor = torch.tensor([[1, 2, 3, 4], 
                              [5, 6, 7, 8]], dtype=torch.float32)
    else:
        input_tensor = torch.tensor([[10, 20, 30, 40], 
                              [50, 60, 70, 80]], dtype=torch.float32)

    input_noncontig = input_tensor.t() # transpose to create a non-contiguous tensor
    assert not input_noncontig.is_contiguous()

    reference = torch.tensor([[11, 55],
                              [22, 66],
                              [33, 77],
                              [44, 88]], dtype=torch.float32)

    result = _reduce(input_noncontig, group)
    dist.barrier()

    if rank == 0:
        correct_match = torch.allclose(result, reference)
        print(f"correct_match = {correct_match}")
        print("result:")
        print(result)

    dist.destroy_process_group()

if __name__ == "__main__":
    main()
```

To run the code: 

`torchrun --nproc_per_node=2 reproduce_example.py`

When running the code above, it prints `correct_match = False`. The `result` tensor remains unreduced (equal to the original local input tensor), instead of matching the `reference` tensor.

**Expected behavior**

`_reduce` should return the reduced tensor for all inputs, regardless of contiguity.

* Suggested fix:

  ```python
  def _reduce(input_, group):
      """All-reduce the input tensor across model parallel group."""
      assert group is not None, "group should not be None"
  
      # Bypass the function if we are using only 1 GPU.
      if group.size() == 1:
          return input_
  
      # All-reduce.
      input_ = input_.contiguous()  # reassign so all_reduce target == returned tensor
      torch.distributed.all_reduce(input_, group=group)
  
      return input_
  ```

  This is a one-line change. It makes the function correct for all inputs and removes the false sense of safety created by the current `.contiguous()` placement.

* Alternatively, if the contract is that callers must provide contiguous input, the `.contiguous()` call should be removed entirely and replaced with an assertion:

  ```python
  def _reduce(input_, group):
      """All-reduce the input tensor across model parallel group."""
      assert group is not None, "group should not be None"
  
      # Bypass the function if we are using only 1 GPU.
      if group.size() == 1:
          return input_
  
      # All-reduce.
      assert input_.is_contiguous(), "_reduce requires contiguous input"
      torch.distributed.all_reduce(input_, group=group)
  
      return input_
  ```

**Additional context**

**Why it hasn't caused problems in practice:**

All current callers pass contiguous tensors (`_ReduceFromModelParallelRegion.forward` receives output from a linear layer; `_CopyToModelParallelRegion.backward` receives a gradient tensor from autograd — both are guaranteed contiguous). So `.contiguous()` is always a no-op and the bug is never triggered.

I have already prepared a fix and will submit a PR shortly.

Pearblossom-M:

# What does this PR do ?
Fix silent correctness issue in `_reduce` where non-contiguous input causes all-reduce result to be discarded.

## Issue tracking
Fixes #5337 

### Pre-checks

### DIFF
--- a/megatron/core/tensor_parallel/mappings.py
+++ b/megatron/core/tensor_parallel/mappings.py
@@ -28,7 +28,11 @@ def _reduce(input_, group):
         return input_
 
     # All-reduce.
-    torch.distributed.all_reduce(input_.contiguous(), group=group)
+    # Note: If input_ is contiguous, it is mutated in-place.  
+    # If not, a new contiguous tensor is created and returned;  
+    # callers must use the returned value to ensure the reduced result is captured.
+    input_ = input_.contiguous()
+    torch.distributed.all_reduce(input_, group=group)
 
     return input_
 
--- a/megatron/core/tensor_parallel/mappings.py
+++ b/megatron/core/tensor_parallel/mappings.py
@@ -28,8 +28,8 @@ def _reduce(input_, group):
         return input_
 
     # All-reduce.
-    # Note: If input_ is contiguous, it is mutated in-place.  
-    # If not, a new contiguous tensor is created and returned;  
+    # Note: If input_ is contiguous, it is mutated in-place.
+    # If not, a new contiguous tensor is created and returned;
     # callers must use the returned value to ensure the reduced result is captured.
     input_ = input_.contiguous()
     torch.distributed.all_reduce(input_, group=group)


guihong-nv:
Thanks for catching this; the diagnosis is correct. One things before this can merge:
Can you add a one-line comment on the semantics? It's worth noting that the contiguous path still mutates the caller's tensor in place, while the non-contiguous path returns a new tensor and does not; so callers must use the return value, not the side effect.

guihong-nv:
/ok to test [07e8d1e](https://github.com/NVIDIA/Megatron-LM/pull/5338/commits/07e8d1e3694b65f78c4dba5db56b8b2faab0d2cc)

Pearblossom-M:
@guihong-nv Thanks for the feedback! I've added the requested comment on the return value semantics. Let me know if anything else is needed!

Pearblossom-M:
@guihong-nv Thanks again for the review! I've pushed the updated commit addressing your feedback. Could you trigger /ok to test on it when you have a moment? Thanks!

guihong-nv:
/ok to test [526617e](https://github.com/NVIDIA/Megatron-LM/pull/5338/commits/526617e831a61831f5abdd15f567b7b1020c8d16)

wujingyue (approved):

wujingyue:
https://github.com/NVIDIA/Megatron-LM/pull/5338/checks?check_run_id=83792470015. Can you sign off your commits?

deepakn94 (approved):

Pearblossom-M:
Hi @wujingyue, thanks for the ping! I squashed the commits and added the sign-off, but caught that the black formatting fix had gotten dropped in the process — it's back in now. Could you trigger /ok to test on it when you have a moment? Thanks again!

yaoyu-33 (approved):

Pearblossom-M:
Hi @wujingyue, it looks like @yaoyu-33's approval might satisfy the core-nemo requirement, but the codeowners-approval check hasn't picked up the new approval. Could you take a look when you have a chance? Thanks!

Pearblossom-M:
Hi @wujingyue, it looks like tests/unit_tests/**/*.py - latest got cancelled after being queued for about 25 hours (looks like it hit GitHub's auto-cancel timeout waiting for a runner), which then caused Nemo_CICD_Test to fail as well. No code changes since last time — just flagging that it seems to be a runner availability issue.

Pearblossom-M:
Hi @wujingyue, sorry for the repeated pings — two more checks failed this round, but both look infra-related rather than code issues:
- gpt/gpt3_mcore_te_tp2_pp1_gdn_no_nvrx_async_mcore - latest was cancelled (similar to the runner timeout from before)
- moe/gpt3_moe_mcore_te_tp4_ep2_etp2_pp2_resume_torch_dist_dist_optimizer failed — from the log it looks like a GPU hardware issue on the runner, not related to this PR's changes

No code changes since the approved version. Hoping a re-run would clear both up — thanks for bearing with the CI flakiness!

PR #5257, opened 2026-06-10
Add test for layer-wise optimizer bucket params on single DP rank

Rohithmatham12:
## Summary
- add a CPU-only regression that exercises `set_bucket_layerwise_params_list()` in the single-DP-rank state reported in #5203

## Testing
- `python3 -m py_compile megatron/core/optimizer/layer_wise_optimizer.py tests/unit_tests/test_layer_wise_optimizer.py`
- `git diff --check`

Not run locally: pytest/runtime regression, because this local checkout does not have `torch`/`pytest` installed. The added regression is CPU-only and should run in the project test environment.

Fixes #5203

PR #5339, opened 2026-06-14
Fix layer-wise optimizer crash on single-GPU bucket param assignment

### LINKED ISSUE #5203: On single-GPU training, self.dp_cp_params_list is None, so iterating over it in a for loop throws an error
In file: /opt/Megatron-Bridge/3rdparty/Megatron-LM/megatron/core/optimizer/layer_wise_optimizer.py, 
at LayerWiseDistributedOptimizer::shard_params()
```
  dp_cp_size = get_pg_size(self.pg_collection.dp_cp)
  if dp_cp_size == 1:
      self.dp_cp_params_list = None                    # it is none!
      self.expt_dp_params_list = None
      return
```

but at LayerWiseDistributedOptimizer::set_bucket_layerwise_params_list()

```
for model_chunk in model_chunks:
            for group in model_chunk.bucket_groups:
                for bucket in group.buckets:
                    if not _bucket_is_managed_by_layer_wise_optimizer(bucket):
                        continue
                    bucket_params_list = [[] for _ in range(get_pg_size(self.pg_collection.dp_cp))]
                    for bucket_list, full_params_list in zip(
                        bucket_params_list, self.dp_cp_params_list             #  self.dp_cp_params_list is none, so raise a error.
                    ):
                        for param in full_params_list:
                            if param in bucket.params:
                                bucket_list.append(param)
                    bucket.set_layerwise_params_list(bucket_params_list)
```

CodersAcademy006:
## What

Fixes #5203: single-GPU training with the layer-wise distributed optimizer crashes with `TypeError: 'NoneType' object is not iterable`.

## Why

When the dp_cp group has size 1, `_shard_params` sets `self.dp_cp_params_list = None` (layer_wise_optimizer.py:453). `set_bucket_layerwise_params_list` then does:

```python
bucket_params_list = [[] for _ in range(get_pg_size(self.pg_collection.dp_cp))]  # length 1
for bucket_list, full_params_list in zip(bucket_params_list, self.dp_cp_params_list):  # zip(..., None)
```

so it iterates over `None` and raises. The expert-parallel branch a few lines below already handles its size-1 case explicitly; the dp_cp branch was missing the equivalent guard (`allgather_params` and `broadcast_params` both guard it too).

## Fix

When `dp_cp_params_list is None`, the single rank owns every param in the bucket, so initialize `bucket_params_list = [list(bucket.params_list)]` — the same shape the expert branch uses for `expt_dp_size == 1`. The multi-rank path is untouched.

## Test

Added `tests/unit_tests/optimizer/test_layer_wise_optimizer_single_gpu.py`. The standard unit-test harness runs at a fixed world size > 1, so the `world_size == 1` condition can't be expressed there; the test drives `set_bucket_layerwise_params_list` directly with a single-rank stand-in and asserts the bucket's params all land on rank 0 without raising. The failure mode is pure Python (iterating `None` before any device op), so this reproduces and guards the bug without a GPU.


guihong-nv:
@CodersAcademy006 do you have any plan to make this PR ready for review?

CodersAcademy006:
Hey @guihong-nv i have made this pr ready for review, please provide your valuable reviews on this. Thank You.

Connor-XY:
Thanks for making this ready for review and for adding regression coverage. This overlaps exactly with #5257: both make `set_bucket_layerwise_params_list()` handle the single-DP-rank case where `dp_cp_params_list` is `None`.

#5257 is already approved and has gone through CI, so we’re consolidating on that PR rather than reviewing multiple implementations of the same fix. I’m closing this in favor of #5257. If #5257 exposes a gap, we can revisit the additional test coverage here.

PR #5349, opened 2026-06-15
Implement Quantile Balancing in MoE

Mellonta:
Adding [Quantile Balancing (QB) MoE](https://kexue.fm/archives/11619).

Standard MoE load balancing adds an auxiliary loss whose gradient nudges the router toward uniform expert usage. QB instead enforces balance directly through a per-expert routing bias, with no auxiliary loss and no gradient contribution, load balancing is handled entirely by a bias that is updated once per global batch from a quantile estimate. From marin's report: https://openathena.ai/blog/quantile-balancing/

> Conceptually, it can be understood as a two-part process run after each step, based on the scores that each token generates for each expert.
> 1. For each token, compute the threshold score required to become activated.
> 2. For each expert, using the result from step 1, determine the bias that would activate a balanced number of tokens.

Usage
```
--moe-router-load-balancing-type quantile_balancing
--moe-router-quantile-balancing-ema 0.0   # optional; 0.0 (default) = no memory, e.g. 0.9 to smooth
--moe-aux-loss-coeff 0                     # aux loss must be off
```

Verified on 2B models

### DIFF
--- a/megatron/core/distributed/finalize_model_grads.py
+++ b/megatron/core/distributed/finalize_model_grads.py
@@ -326,6 +326,9 @@ def reset_model_temporary_tensors(config: TransformerConfig, model: List[torch.n
                 or "global_aux_loss" in config.moe_router_load_balancing_type
             ) and hasattr(module, 'reset_global_aux_loss_tracker'):
                 module.reset_global_aux_loss_tracker()
+            if getattr(module, 'qb_beta_accum', None) is not None:
+                module.qb_beta_accum.zero_()
+                module.qb_beta_count.zero_()
 
 
 def _update_router_expert_bias(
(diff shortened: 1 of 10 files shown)


andresnowak:
Hello, I have been also trying to check how to integrate the quantile balancing into Megatron LM for SwissAI, but really thank you for doing this implementation. I just wanted to ask some questions if I could:
- One is what is the idea behind the EMA option if I could ask, from what I understand the idea is that the QB will be calculated for each individual global batch no? (or well the necessary implementation is doing an average because it would be to costly to do it from the whole global batch). Like did you implement this method because you saw very big balance differences between global batches and the idea was to use an EMA to smooth this?
- If you see a difference from speed from not being able to use the fused_topk (I know there should be but just asking, this was the part that I was trying to see if it would be necessary to have the fused_topk).
- And just final question, what other things would this PR need because reading it I thought it is already complete (having the quantile computation per microbatch and then doing the stack to then be able to do the averaging across the global batch)

Sorry to bother you with the questions.

Mellonta:
Hi @andresnowak , glad it's useful. For you questions:
1. Your understanding is right: The EMA just smooths that across batches, but it's off by default (ema=0), I didn't see much difference on the small models I tried, so it's just there as a knob.
2.  Fused topk would help, but QB selects experts outside the score function, so the fused kernel can't be used as-is. This PR is mainly for studying the algorithm rather than efficiency, so I left it on the eager path.
3. Yep, it's complete.

Mellonta:
Address legitimate comments from Claude and rebase.

fanshiqing (approved):

deepakn94 (approved):

Phlip79 (approved):

yaoyu-33 (approved):

Mellonta:
Also fixes a pre-existing order-dependent hang in `test_shared_expert_forward_backward`: it seeded only CUDA but uses use_cpu_initialization, so inherited CPU-RNG state could drift in some ranks and cause a deadlock. now seeded via `_set_random_seed`.

PR #5371, opened 2026-06-16
Thread tensor-parallel group into the RADIO patch embedder

yashaswikarnati:
## What
Thread the encoder's tensor-parallel group from a provided `ProcessGroupCollection` into the RADIO patch embedder (and the separate video embedder) `ColumnParallelLinear`'s `tp_group`. The argument defaults to `None` when no `pg_collection` is supplied, preserving stock behavior for existing callers.

## Why
The RADIO ViT patch embedder built its `ColumnParallelLinear` without an explicit tensor-parallel group, so it fell back to the global TP group. Under a heterogeneous MIMO grid the encoder lives on its own grid whose TP group is not the global one, producing a wrong/empty TP group for the embedder.

Purely additive / default-None; no change to stock behavior.

## Validation
Validated in the 8-GPU 20L Nemotron VLM e2e (trains + checkpoint save/resume, lm loss 12.18->11.54 across resume).

## CODEOWNERS
`megatron/core/` -> @NVIDIA/core-adlr @NVIDIA/core-nemo

### DIFF
--- a/megatron/core/models/vision/radio.py
+++ b/megatron/core/models/vision/radio.py
@@ -173,6 +173,7 @@ class RADIOViTModel(VisionModule):
                 gather_output=True,
                 disable_grad_reduce=True,
                 init_method=lambda tensor: torch.nn.init.normal_(tensor, mean=0.0, std=1.0),
+                tp_group=pg_collection.tp if pg_collection is not None else None,
             )
             self.video_embedder = ColumnParallelLinear(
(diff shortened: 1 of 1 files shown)


deepakn94 (approved):

yashaswikarnati:
/ok to test [db47fe2](https://github.com/NVIDIA/Megatron-LM/pull/5371/commits/db47fe260db2bd73ea9991166fb1f9cf91fe5176)

Phlip79 (approved):

yaoyu-33 (approved):

PR #5373, opened 2026-06-16
Support the MIMO cross-grid path in training loop

yashaswikarnati:
## What
Three adjustments so the stock training loop supports the MIMO cross-grid path, all no-ops for non-MIMO runs:

- `get_forward_backward_func` selects `forward_backward_pipelining_without_interleaving` when passed a `MultiModuleProcessGroupCollection` (the cross-grid bridge); `train()` calls the selector unconditionally instead of branching on `p2p_communicator`.
- Guard the `losses_reduced[0]` access: MIMO encoder-grid ranks are their module's last PP stage but compute no language loss, so the list is empty there.
- Skip the ModelOpt distillation shape hook unless `modelopt_enabled`; it reads `parallel_state`, which is uninitialized on the MIMO path.

## Validation
8-GPU 20L Nemotron VLM e2e: trains + checkpoint save/resume (lm loss 12.18 → 11.54 across resume).

### DIFF
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -3354,8 +3354,6 @@ def train(
     eval_iterations = 0
     # Wrap forward_backward_func for Full iteration CUDA graph
     if p2p_communicator is not None:
-        # MIMO cross-grid bridge: pick the schedule by p2p_communicator presence, not
-        # parallel_state pp size (unset without mpu). without-interleaving supports both.
         from megatron.core.pipeline_parallel.schedules import (
             forward_backward_pipelining_without_interleaving,
         )
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -3353,14 +3353,9 @@ def train(
     eval_duration = 0.0
     eval_iterations = 0
     # Wrap forward_backward_func for Full iteration CUDA graph
-    if p2p_communicator is not None:
-        from megatron.core.pipeline_parallel.schedules import (
-            forward_backward_pipelining_without_interleaving,
-        )
-
-        forward_backward_func = forward_backward_pipelining_without_interleaving
-    else:
-        forward_backward_func = get_forward_backward_func()
+    forward_backward_func = get_forward_backward_func(
+        schedule_pg_collection=schedule_pg_collection
+    )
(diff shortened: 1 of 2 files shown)


megatron/training/training.py
```diff
@@ -2388,8 +2391,11 @@ def _save_state_dict(attr_name, label):
     if args.empty_unused_memory_level >= 2:
         torch.cuda.empty_cache()
 
-    if is_last_stage:
-        # Average loss across microbatches.
+    if is_last_stage and losses_reduced:
```
yashaswikarnati, line 2391: what exactly is last stage is here and how its computed? if its only LLM why we also need losses_reduced ?

megatron/training/training.py
```diff
@@ -2234,8 +2234,11 @@ def train_step(forward_step_func, data_iterator, model, optimizer, opt_param_sch
             model_chunk.force_all_reduce = save_wgrads_in_this_iteration
         optimizer.zero_grad()
 
-        if has_nvidia_modelopt:
-            # [ModelOpt]: Pipeline-parallel Distillation stacks student and teacher tensors
+        if has_nvidia_modelopt and p2p_communicator is None:
```
yashaswikarnati: why did we add  p2p_communicator is None ? when is use_model_opt triggered ?

megatron/training/training.py
```diff
@@ -3352,7 +3358,18 @@ def train(
     eval_duration = 0.0
     eval_iterations = 0
     # Wrap forward_backward_func for Full iteration CUDA graph
-    forward_backward_func = get_forward_backward_func()
+    if p2p_communicator is not None:
+        # MIMO cross-grid bridge: select the schedule by the presence of a
```
yashaswikarnati: verbose doc strings

Phlip79 (approved):

yaoyu-33 (approved):

deepakn94 (approved):

yashaswikarnati:
/ok to test [649065d](https://github.com/NVIDIA/Megatron-LM/pull/5373/commits/649065df094944dbee42708712df2e84f622e4a2)

PR #5385, opened 2026-06-16
route collectives through nccl2

tushar00jain:
Summary:
# Route Megatron-LM collectives through the in-tree nccl2 backend

## Summary

This makes Megatron-LM's distributed stack run over PyTorch's in-tree `nccl2`
backend. The world process group is initialized as an eagerly device-bound
`cpu:gloo,cuda:nccl2` group. All covering-partition subgroups -- the TP / DP /
CP / model-parallel families, their expert-parallel counterparts, the embedding
subgroups, and the gloo metadata groups -- are created with a single
`torch.distributed.split_group` collective over that world PG. The
pipeline-parallel groups are the sole exception: they use
`new_group(backend="nccl-lazy", ...)` for per-peer lazy point-to-point
communicators.

## Why this shape

`nccl2` is a faithful port of torchcomms: it supports comm-split
(`ncclCommSplit`) but not the stock eager-`new_group` no-color-split path. So
every subgroup that is a subset of the world PG is created via `split_group`,
which is collective and uses hashed group names -- self-contained, with no
dependence on `_world.group_count`, hence no members-only workarounds and no
deadlocks. The one exception is pipeline parallelism, which is pure
point-to-point (send/recv between stages): those groups use the `nccl-lazy`
backend so each peer pair gets its own lazily-created 2-rank communicator and
concurrent P2P to different stages can overlap.

### `megatron/core/parallel_state.py`

`create_partition_groups` builds every cuda and gloo covering partition with one
`split_group(split_ranks=...)` call over the world PG, via a
`_split_group_backend` helper that device-qualifies the filter: `cuda:nccl2` for
cuda families, and `cpu:gloo,cuda:nccl2` for gloo (the split filter must include
the parent's default cuda device, so the gloo group carries a redundant cuda
backend -- the same shape vLLM uses). Embedding / position-embedding groups are
built the same way from their partition lists. The pipeline-parallel device
group is created via `new_group(backend="nccl-lazy",
use_local_synchronization=True, device_id=...)` (the ucc path stays eager);
`create_group` gains an optional `device_id` used only there.

### `megatron/training/initialize.py`

The world PG is initialized as `cpu:gloo,cuda:<backend>` and eagerly bound to
the rank's cuda device via `device_id`; `split_group` requires a device-bound
parent that carries a cpu:gloo backend for the gloo splits to filter.

### `megatron/training/config/common_config.py`

`distributed_backend` defaults to `nccl2` (choices: `nccl2`, `nccl`, `gloo`).

### `megatron/core/process_groups_config.py`

The singleton `expt_dp_group` fallback routes through
`parallel_state.create_group` so it shares the common group-creation path.

### `tests/unit_tests/test_utilities.py`

Unit-test world init mirrors the `cpu:gloo,cuda:nccl` qualified backend.

## Test Plan

4xH100, CUDA 13, torchcomms conda env. pp2_dp2 bit-exact vs stock nccl over 20
steps:

```
CONFIGS=pp2_dp2 STEPS=20 bash run_bitexact_matrix_explicit.sh
```

Result: `[pg-probe] backend=cpu:gloo,cuda:nccl2 world=ProcessGroupNCCL2
tp=ProcessGroupNCCL2 pp=ProcessGroupNCCLLazy dp=ProcessGroupNCCL2`, VERDICT
BIT_EXACT_PASS (byte-identical loss/grad/weights, no hang). The rest of the
parallelism matrix (tp4/tp2_dp2/tp2_pp2/pp4) is pending.

This change was authored with the assistance of an AI coding assistant.

Signed-off-by: Tushar Jain <tushar00jain@users.noreply.github.com>

### DIFF
--- a/megatron/core/parallel_state.py
+++ b/megatron/core/parallel_state.py
@@ -217,6 +217,7 @@ def create_group(
     pg_options=None,
     use_local_synchronization=False,
     group_desc=None,
+    device_id=None,
 ):
     """Creates a ProcessGroup."""
     kwargs = {
@@ -227,6 +228,8 @@ def create_group(
         "use_local_synchronization": use_local_synchronization,
         "group_desc": group_desc,
     }
+    if device_id is not None:
+        kwargs["device_id"] = device_id
     if not is_torch_min_version("2.4.0"):
         kwargs.pop("group_desc")
         if timeout is None:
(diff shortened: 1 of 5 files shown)


guihong-nv:
Thanks for opening this PR.

Could you please update the PR description to follow the Megatron-LM [pull request template](https://github.com/NVIDIA/Megatron-LM/blob/main/.github/pull_request_template.md)? The current PR body is empty, so reviewers do not have the required summary, issue tracking, testing, documentation, and pre-check information.

A few specific items to address:

- Fill out the PR template, including the overview and pre-checks: https://github.com/NVIDIA/Megatron-LM/blob/main/.github/pull_request_template.md#L1-L25
- If this is a new feature, link the required feature request issue: https://github.com/NVIDIA/Megatron-LM/blob/main/.github/pull_request_template.md#L8-L15
- Please make sure the contribution follows the code submission policy: https://github.com/NVIDIA/Megatron-LM/blob/main/docs/developer/contribute.md#L31-L49
- Please also sign off your commit as required by the DCO policy: https://github.com/NVIDIA/Megatron-LM/blob/main/docs/developer/contribute.md#L50-L56

Once those are updated, we can continue review. If the PR remains without the required template/policy information, we may need to close it until it is resubmitted or updated accordingly.

Issue #5394, opened 2026-06-17
[BUG] ChainedOptimizer applies global grad-norm clipping to Muon (orthogonalizing) param groups, silently stalling training


yuchenwang3:
**Describe the bug**

`ChainedOptimizer.step()` computes a single global `grad_norm` across all chained sub-optimizers and applies `clip_grad_by_total_norm_fp32(..., total_norm=grad_norm)` to **every** sub-optimizer's parameters, including a Muon / orthogonalizing group (`dist_muon`). For an orthogonalizing optimizer this is (a) **meaningless** — Muon discards gradient magnitude by design (Newton–Schulz orthogonalization is scale-invariant) — and (b) **harmful**: when the global `grad_norm` is large, the clip coefficient `c = clip_grad/grad_norm` is tiny and scales the per-matrix gradients fed into Newton–Schulz below its internal `F.normalize(eps=1e-7)` floor, so orthogonalization **degenerates** (near-zero, non-orthogonal update). Affected layers stop updating, loss stalls, `grad_norm` grows, `c` shrinks further, more layers collapse — a silent positive-feedback stall.

**Steps/Code to reproduce bug**

We hit this fine-tuning a GatedDeltaNet model whose gradients are inherently large and per-layer-imbalanced (`grad_norm ≈ 5e7`, growing to ~2e11). Minimal mechanism repro (torch-only) showing the clip coefficient pushing a per-matrix gradient below the Newton–Schulz floor:

```python
import torch, torch.nn.functional as F
clip_grad, grad_norm = 1.0, 5e7
c = clip_grad / (grad_norm + 1e-6)                 # ≈ 2e-8

def newton_schulz(x, steps=6, eps=1e-7):
    a, b, cc = 3.4445, -4.7750, 2.0315
    X = F.normalize(x, p=2, dim=(-2, -1), eps=eps)
    for _ in range(steps):
        A = X @ X.transpose(-2, -1); X = a*X + (b*A + cc*(A@A)) @ X
    return X

G = torch.randn(2048, 2048); G = G / G.norm() * 0.1     # a late-layer matrix grad
print("no clip :", newton_schulz(G).norm().item())       # ~40  (orthogonal)
print("clipped :", newton_schulz(G * c).norm().item())   # collapses toward 0
```

Training-level, single-variable (fixed 32-example overfit batch, `optimizer=dist_muon`, **only `clip_grad` changed**):
```
[...]

PR #5395, opened 2026-06-17
fix(optimizer): skip grad-norm clipping for Muon optimizers

yuchenwang3:

**Background.** Encountered while running **ms-swift + Megatron-Core SFT of Qwen3.5-35B-A3B** (GatedDeltaNet hybrid, `optimizer=dist_muon`) on **16× B200 across 2 nodes (2×8)**. Full root-cause write-up in #5394.

## Scope & root cause (optimizer-agnostic)

The underlying failure is **not Muon-specific**. A benign-but-large global `grad_norm` (here ~5e7 from the GatedDeltaNet layers, growing to ~2e11) makes the clip coefficient `c = clip_grad / grad_norm ≈ 2e-8`, which scales per-matrix gradients below the optimizer's numerical floor and freezes updates. The same condition stalls **AdamW** too (its `eps` floor) — i.e. the clip, **not the optimizer**, is the root cause. This is now confirmed directly (see Testing): AdamW with `clip_grad=1.0` stalls at ~0.5, while AdamW with `clip_grad=0` overfits a fixed batch to **0.029** (vs Muon's 0.019). (We used `dist_muon` here following the model's training recipe; it is not required — it is simply naturally immune to this clip, since magnitude clipping is a no-op for an orthogonalizing update.)

This PR fixes the case that has a **clean, side-effect-free fix**: for an orthogonalizing optimizer, magnitude clipping is *semantically meaningless* — Newton–Schulz discards gradient magnitude — so skipping it is unambiguously correct. There is no equally clean "skip clip" for a magnitude-based optimizer (it legitimately uses clipping for spike protection); the optimizer-agnostic alternative is to clamp the clip coefficient so a benign-large `grad_norm` can't drive gradients under the eps floor (listed below).

## The harmful path, concretely

`ChainedOptimizer.step` clips every non-stub sub-optimizer's `main_params` with the global `grad_norm` whenever `optimizer.config.clip_grad > 0`. Applied to a Muon sub-optimizer, the tiny `c ≈ 2e-8` pushes its per-matrix gradients below Newton–Schulz's internal `F.normalize` floor, so the orthogonalization **silently degenerates** (near-zero, non-orthogonal update) and training stalls — with the forward/loss looking completely normal.

## Why not a config-level fix

Setting `clip_grad=0` only for the Muon sub-optimizer is not possible via `config`, because `ChainedOptimizer` requires all sub-optimizers to share the same `OptimizerConfig` (the DistOpt+LayerWise path even resets `result.config = config` to satisfy this). So this PR uses a per-optimizer attribute instead.

## Change

- `ChainedOptimizer.step`: skip the norm-clip for a sub-optimizer carrying `skip_grad_norm_clip = True`. The attribute defaults to `False` on `MegatronOptimizer`, so all existing optimizers are unaffected. `grad_norm` is still computed and returned for logging; only the clip is skipped, and only for the flagged (Muon) sub-optimizer — Adam sub-optimizers still clip normally. Separate grad-norm groups such as MTP are computed over that same non-exempt subset.
- `get_megatron_optimizer` / `_build_emerging_optimizer`: set `skip_grad_norm_clip = True` on the emerging (Muon) optimizer — both the standard-chain wrapper and the `LayerWiseDistributedOptimizer` (which owns the Muon-managed matrix params for `dist_muon`).

## Testing

The runtime effect — not magnitude-clipping the Muon param group — was validated on the actual **Qwen3.5-35B-A3B (GatedDeltaNet) model on 16× B200 across 2 nodes (2×8)**, `optimizer=dist_muon`:

- Muon group's gradient clipping **off** (equivalent to this PR; in our run via `clip_grad=0`): a fixed 32-example batch overfits to **~0.02**, real-data training learns.
- Global clip **on** (`clip_grad=1.0`, current behavior): training **stalls** (loss flat ~0.5, `grad_norm` 7.5e7 → 2.2e11).
- Dual single-variable control: keeping `clip_grad=1.0` but only lowering Newton–Schulz `eps` 1e-7→1e-30 also unblocks learning, confirming the `clip → NS-eps` interaction.
- **Optimizer-agnostic confirmation** (`optimizer=adam`, fixed 64-example overfit, 200 steps): `clip_grad=0` → loss `0.69 → 0.029` (vs Muon `clip_grad=0` → 0.019), while `clip_grad=1.0` stalls at ~0.5. `grad_norm` stayed 1e8–3e12 throughout with no NaN — Adam's per-parameter second-moment normalization digests the raw gradients fine once they are not crushed by the clip. So the clip is the sole cause across optimizers; Adam is viable without it, and Muon is not required for correctness (just lower/steadier).

This PR reproduces that validated behavior via the `skip_grad_norm_clip` attribute instead of a global `clip_grad=0`; the exact attribute wiring should additionally be exercised by CI.

## Alternatives

1. A config/CLI flag (e.g. `--no-clip-orthogonalizing-optimizers`) instead of an implicit attribute.
2. Detect orthogonalizing optimizers by type rather than an attribute.
3. A more general fix that also helps magnitude-based optimizers: clamp the clip coefficient with a floor so a benign-but-large `grad_norm` can't drive per-matrix gradients under the optimizer's numerical eps. (Broader blast radius — affects all training — hence not chosen here.)
4. Pair with the related robustness fix so the orthogonalizer doesn't silently degenerate on small inputs: NVIDIA-NeMo/Emerging-Optimizers#229 / PR NVIDIA-NeMo/Emerging-Optimizers#230.

Refs #5394


megatron/core/optimizer/__init__.py
```diff
@@ -963,6 +966,9 @@ def _get_megatron_emerging_optimizer(
             init_state_fn_list=list(init_fns),
             model_chunks=model_chunks,
         )
+        # LayerWise owns the Muon-managed (orthogonalizing) matrix params; their update
+        # is scale-invariant, so skip magnitude gradient clipping for it (see issue #5394).
+        setattr(layer_wise_optimizer, 'skip_grad_norm_clip', True)
```
ShauryaaSharma: **Critical — flag set on the container, not on the inner sub-optimizers; fix is a no-op when `results` is empty**

`LayerWiseDistributedOptimizer` is a subclass of `ChainedOptimizer` and does not
override `step()`. Its `self.chained_optimizers` are the per-layer sub-optimizers
built from `layer_wise_base_results` — and none of those objects receive
`skip_grad_norm_clip=True`.

**When `results` is non-empty** the caller gets
`ChainedOptimizer([layer_wise_optimizer] + results)`. The outer `step()` iterates
`layer_wise_optimizer` as a single element, reads the flag from it, and skips
clipping — this path works correctly.

**When `results` is empty** `layer_wise_optimizer` is returned directly to the
training loop. Calling `.step()` on it invokes the inherited `ChainedOptimizer.step()`
with `self.chained_optimizers` = the per-layer sub-optimizers. The guard at line 1677:

```python
if optimizer.config.clip_grad > 0.0 and not getattr(optimizer, "skip_grad_norm_clip", False):
```

evaluates to `True` for every inner sub-optimizer — none of them carry the flag.
Clipping still fires. **The fix is entirely bypassed for this path.**

Suggested fix — propagate the flag to each raw sub-optimizer before appending:

```python
# in the use_layer_wise branch, before layer_wise_base_results.append(...)
setattr(optimizer, 'skip_grad_norm_clip', True)
layer_wise_base_results.append((optimizer, init_state_fn))
```

Setting it on the assembled container after the fact is insufficient whenever that
container is the top-level optimizer returned to the training loop.

megatron/core/optimizer/__init__.py
```diff
@@ -884,6 +884,9 @@ def _get_megatron_emerging_optimizer(
             else:
                 optimizer = FP32Optimizer(optimizer, config, init_state_fn)
             setattr(optimizer, 'grad_stats_parallel_group', model_parallel_group)
+            # Orthogonalizing optimizers (Muon) have scale-invariant updates; magnitude
+            # gradient clipping is a no-op at best and harmful at worst (see issue #5394).
+            setattr(optimizer, 'skip_grad_norm_clip', True)
```
ShauryaaSharma: **High — `skip_grad_norm_clip` applied unconditionally to all emerging optimizers, not just orthogonalizing ones**

`_get_megatron_emerging_optimizer` is the entry point for every optimizer registered
in `_EMERGING_OPTIMIZERS`, which currently includes SOAP and Lion in addition to Muon
variants. The `setattr(..., 'skip_grad_norm_clip', True)` at this line fires for all
of them.

The exemption is only semantically valid for **orthogonalizing** optimizers: Muon's
Newton–Schulz step discards gradient magnitude by construction, making L2 norm
clipping a no-op on the update direction and actively harmful when the clip coefficient
is small. SOAP and Lion have no such property — they are magnitude-sensitive and should
remain subject to clipping. Silently disabling it risks training instability for any
non-Muon emerging optimizer with `clip_grad > 0.0`.

Suggested fix — introduce an explicit allowlist and gate on it:

```python
_ORTHOGONALIZING_OPTIMIZERS = frozenset({'muon', 'dist_muon'})
if opt_name in _ORTHOGONALIZING_OPTIMIZERS:
    setattr(optimizer, 'skip_grad_norm_clip', True)
```

This also future-proofs the function: any new optimizer added to `_EMERGING_OPTIMIZERS`
will clip by default and must explicitly opt out, rather than silently inheriting an
exemption it may not deserve.

megatron/core/optimizer/optimizer.py
ShauryaaSharma: **Nit/Medium — Muon grads inflate the shared `grad_norm` used for `grad_norm_skip_threshold`**

`grad_norm` is computed at the top of `ChainedOptimizer.step()` via
`self.get_grad_norm()` — a combined norm across **all** sub-optimizers, including
Muon. This same scalar is then used in the threshold comparison for every
sub-optimizer in the loop, including AdamW.

The issue report documents Muon grad norms of ~5×10⁷ growing to ~2×10¹¹. If a user
configures a finite `grad_norm_skip_threshold` (e.g. to guard against spike-and-NaN
cascades), the AdamW sub-optimizer's update will be suppressed based on a norm Muon's
gradients are dominating — even if AdamW's own gradients are well-behaved. The entire
step is dropped via `should_skip_update = True`.

This is consistent with the stated goal of the PR: if Muon grads are excluded from
clipping, they should arguably also be excluded from the threshold comparison for
non-Muon sub-optimizers. The straightforward fix is to compute a separate `grad_norm`
that excludes parameters whose sub-optimizer has `skip_grad_norm_clip=True`, and use
that narrower norm for the threshold check.

`grad_norm_skip_threshold` defaults to `float('inf')` so this won't affect most users
today, but it is a quiet footgun for production configs that set a finite value.

ShauryaaSharma, line 1963: **Low — `should_clip` predicate ignores `skip_grad_norm_clip`, triggering a wasted collective**

`should_clip` is evaluated by checking `optimizer.config.clip_grad > 0.0` across all
sub-optimizers, but does not consult `skip_grad_norm_clip`. When every sub-optimizer
that has `clip_grad > 0.0` also carries `skip_grad_norm_clip=True` (e.g. a Muon-only
chain), `should_clip` evaluates to `True` and `_compute_grad_norms_by_group()` is
called — triggering one or more distributed AllReduces to compute group norms that no
sub-optimizer will actually use for clipping.

Suggested fix — add the flag to the predicate:

```python
should_clip = any(
    not (hasattr(optimizer, 'is_stub_optimizer') and optimizer.is_stub_optimizer)
    and optimizer.config.clip_grad > 0.0
    and not getattr(optimizer, 'skip_grad_norm_clip', False)
    for optimizer in self.chained_optimizers
)

ShauryaaSharma (changes requested):
**Request Changes** — the fix is directionally correct but has two blocking issues.

**Bug 1 (Critical):** The `skip_grad_norm_clip` flag is set on the `layer_wise_optimizer`
container, but when `results` is empty the training loop calls `.step()` directly on it.
The inherited `ChainedOptimizer.step()` then iterates the inner per-layer sub-optimizers —
none of which carry the flag — so clipping still fires. The fix is entirely bypassed for
this code path.

**Bug 2 (High):** The flag is applied unconditionally to every optimizer routed through
`_get_megatron_emerging_optimizer`, including SOAP and Lion. These are not scale-invariant
and should remain subject to clipping.

Two medium/low findings are also included as inline comments (`grad_norm_skip_threshold`
pollution and the `should_clip` wasted AllReduce).

yuchenwang3:
Thanks for the thorough review, @ShauryaaSharma — all four findings are addressed, and I verified the result on an 8-GPU (sm90/H100-class) node.

**1 (Critical).** The flag is now set on the raw orthogonalizing sub before it's appended to the LayerWise base list, **and** re-propagated onto `layer_wise_optimizer.chained_optimizers` after construction. The second step is required: `LayerWiseDistributedOptimizer` re-wraps each base optimizer in `Float16OptimizerWithFloat16Params` under bf16, and the wrappers don't forward attribute access, so the inherited `ChainedOptimizer.step()` (the `results`-empty direct path) reads the flag off the wrapper. Verified: a Muon-only bf16 build returns a bare `LayerWiseDistributedOptimizer` whose every chained member carries the flag; a Muon+Adam build leaves the Adam sub unflagged.

**2 (High).** Gated on `isinstance(optimizer, OrthogonalizedOptimizer)`, computed on the raw optimizer. On `emerging_optimizers` 0.4.0a0 this hits exactly the Muon family (Muon, AdaptiveMuon, Scion, Spel, MOP, PolarGrad, MuonHyperball, SinkhornMuon); SOAP and Lion are not subclasses and keep clipping. Container flag set only when all base subs are orthogonalizing. Added a guarded `OrthogonalizedOptimizer` import with an empty-tuple sentinel so `isinstance` is safely `False` if the package is absent.

**3 (Medium).** Added `_get_grad_norm_skip_threshold()`, which excludes skip-flagged subs' grads from the norm used for the `grad_norm_skip_threshold` comparison (and exempts skip-flagged subs from the check). It returns `get_grad_norm()` unchanged when nothing is flagged, so no behavior change for non-Muon users. Scoped to the threshold check (the clip already short-circuits for skip-flagged subs). Verified: a Muon sub at grad_norm ≈ 6.2e6 + a well-behaved Adam sub → `get_grad_norm()`=6.2e6 vs `_get_grad_norm_skip_threshold()`=4.0e-4, so the Adam update is no longer wrongly skipped.

**4 (Low).** `should_clip` now also checks `skip_grad_norm_clip`, so a Muon-only chain no longer triggers `_compute_grad_norms_by_group()` (verified: 0 calls for a Muon-only chain, ≥1 for an Adam chain).

**One issue I caught and fixed while testing (flagging for transparency).** My first cut of `_get_grad_norm_skip_threshold()` called `get_grad_stats_parallel_group()` unconditionally, which asserts all chained subs share one grad-stats group. On the distributed-optimizer path (Muon `LayerWiseDistributedOptimizer` + Adam `DistributedOptimizer`) the groups differ, so it raised `AssertionError` and broke the `use_param_layout=True` layer-wise tests. The latest commit restores `get_grad_norm()`'s shared/non-shared handling — sharedness is judged over the non-skip subs only (sidestepping the container assertion), with a per-sub-norm fallback when non-shared and an early return for the Muon-only case. Those `[True]` tests are green again.

**Verification (8-GPU node, torchrun, vs base `ad5a93b`).** No new failures vs base: `test_layer_wise_optimizer.py` is back to 41/0 (the 15 `[True]` failures cleared), `optimizer/` and `test_lion_optimizer.py` unchanged. (There are 27 pre-existing `adaptive_muon`/`soap` failures from an `emerging_optimizers` 0.4.0a0 API skew — identical on base and branch, unrelated to this PR.) I also added `tests/unit_tests/optimizer/test_skip_grad_norm_clip.py` covering the four behaviors above plus the distributed-optimizer path (8/8 pass).

The clip `total_norm` for non-skip subs still uses the full `grad_norm` — I scoped finding 3 to the threshold as you suggested; happy to narrow the clip norm too in a follow-up if you'd prefer. Ready for another look.

ShauryaaSharma (approved):

yuchenwang3:
Thanks again for the review and approval, @ShauryaaSharma 🙏 This still shows `REVIEW_REQUIRED` and the copy-pr-bot CI hasn't been triggered yet, so it can't merge as-is. Anything needed from my side — happy to rebase onto latest `main`? If it's just the CI / CODEOWNER step, could you `/ok to test` or loop in someone from the optimizer CODEOWNERS to help land it? Cheers.

yuchenwang3:
Following up here 🙏 Thanks again for the review and approval, @ShauryaaSharma. The PR still shows `REVIEW_REQUIRED` and the copy-pr-bot CI has not been triggered, so it cannot merge as-is. Per CODEOWNERS, `megatron/core/optimizer/` needs approval from @NVIDIA/mcore-optimizer (or @NVIDIA/core-adlr). Could you help kick off CI and/or loop in a CODEOWNER? Happy to rebase if useful — the change is small and sits behind the existing Muon path.

Connor-XY:
Thanks for the detailed follow-up and validation. The remaining blockers are full CI and required optimizer CODEOWNER approval; a rebase is not currently necessary.

I’m marking this as medium complexity, starting functional CI on the current commit, and routing the remaining review to @NVIDIA/mcore-optimizer. No further action is needed from your side unless CI reports an issue.

Connor-XY:
/claude fix Apply only the Black formatting changes reported by the current lint job in tests/unit_tests/optimizer/test_skip_grad_norm_clip.py. Do not change behavior, assertions, test coverage, or production code.

janEbert:
Dear @yuchenwang3, thanks a lot for raising the issue and the PR contribution. I'd recommend a larger overhaul of the changes, mostly to avoid duplication, remove verbosity and better align with the rest of the codebase.

1. `_get_grad_norm_skip_threshold` duplicates some of the `get_grad_norm` logic. Maybe it makes more sense to refactor `get_grad_norm` and `_get_grad_norm_skip_threshold` to use a shared private helper, which takes a list of sub-optimizers to compute the grad norm for.
2. Please use `OrthogonalOptimizer = object` as the sentinel, as already established in `emerging_optimizers.py` (and remove the accompanying comment, or at least reduce its verbosity).
3. Please refactor the `skip_grad_norm_clip` attribute assignments, e.g., by introducing a helper to the `MegatronOptimizer` class, to set/read the attribute at/from a single place instead of duplicating it across the entire instance hierarchy.
4. Similarly, concerning the behavior/solution surrounding the handling of a `layer_wise_optimizer` wrapped by a `ChainedOptimizer`: it seems unclean and like it would easily introduce errors, since we ignore the individual optimizer settings in this code path.
5. I think we should already narrow the `grad_norm` used for actual clipping to not consider the skipped values in _this_ PR, unlike suggested in a comment to leave it for a follow-up. Otherwise, while we properly skip clipping Muon, we still clip Adam according to the undesired Muon grads, which seems like behavior this PR should also be trying to avoid.
6. In general, the comments are extremely verbose, sometimes repeat what the code is already stating and can often be meaningfully reduced in content while still being helpful for future reference.
7. Replace `setattr` and `getattr` use with standard assignment/queries, which also relates to point 3.

yuchenwang3:
Thanks for the thorough review @janEbert — all seven points addressed in 613425d (net −43 lines):

1. **Shared helper**: `get_grad_norm()` and the clip/skip-threshold norm now both go through a new `_get_grad_norm_over(optimizers)`, which mirrors the shared/non-shared grad-stats-group handling for any subset. `_get_grad_norm_skip_threshold` is gone.
2. **Sentinel**: `OrthogonalizedOptimizer` is now imported from `megatron.core.optimizer.emerging_optimizers`, reusing its established `= object` sentinel; the local fallback and its comment are removed.
3. **Single place for the flag**: declared once as a class attribute on `MegatronOptimizer` (default `False`); construction sets it only on the leaf Megatron wrappers.
4. **Chained/LayerWise handling**: the stored container flag and the raw-optimizer flag relay are gone. `ChainedOptimizer.skip_grad_norm_clip` is now a derived property (`all(sub.skip_grad_norm_clip)`), so a Muon-only LayerWise nested in an outer chain is exempt and a mixed one is conservatively clipped — no per-path special cases.
5. **Clip norm narrowed in this PR**: `step()` computes one `clippable_grad_norm` over the non-exempt subs and uses it for both the clip `total_norm` and the skip-threshold check, so Adam is no longer clipped against Muon's discarded magnitudes. The *reported* `grad_norm` metric intentionally stays the full norm for continuity — happy to narrow that too if you'd prefer.
6. **Comments** trimmed throughout (the long narrations are gone).
7. **setattr/getattr** for the flag replaced with plain attribute access; tests updated accordingly, plus a new test asserting the clip receives the narrowed norm.

With point 5 in place here, #5431 (which isolated the same effect at the clip level) is covered by this PR — I can close it as absorbed if you agree.

yuchenwang3:
@janEbert circling back — all seven points from your review were addressed in 613425d three weeks ago (net −43 lines, including the shared grad-norm helper you suggested). Could you take another look when you get a chance? If anything still reads off, I'll turn it around quickly.

yuchenwang3:
@janEbert before I rebase this across the newer separate-grad-norm-group changes, is the approach in 613425d still the direction you'd want? If so I'll refresh the branch and rerun the optimizer tests.

yuchenwang3:
Rebased the full change onto current main and folded the old CI merge-only formatting into the authored commits. The branch is linear again; formatter, syntax, and focused structural checks pass locally. @janEbert, could you re-review this head and run `/ok to test 281bc947005ca7c06f9e6c778636c3ab0279edb8` if it still looks good?

yuchenwang3:
@janEbert one more follow-up landed in 2232ac83a3fc31aaace3927bc196f89e7c687993. While rechecking the current optimizer paths I found that the separate MTP norm was still being aggregated across exempt Muon and clippable Adam sub-optimizers, so a large Muon MTP grad could still over-clip Adam. The grouped norm now uses the same clippable subset as the main norm, including the LayerWise path, and there is a regression for that mixed case. I also kept duck-typed optimizer compatibility. Formatting, Ruff, isort, syntax, and diff checks pass; could you review this SHA and run `/ok to test 2232ac83a3fc31aaace3927bc196f89e7c687993`?

yuchenwang3:
@janEbert small correction: I amended that follow-up as afc6d848f18828a99974478cb979cd417e9e520c so the production path keeps plain `skip_grad_norm_clip` access, matching your earlier point 7; the current-main test doubles now declare the interface explicitly. The grouped-norm fix itself is unchanged. Please use `/ok to test afc6d848f18828a99974478cb979cd417e9e520c`.

yuchenwang3:
The focused code-path checks are green on 8×B200 too. With the repository-pinned Emerging Optimizers v0.3.0 installed, the two-rank `skip_grad_norm_clip` suite completed with 9 passed and no skips, and the three chained/MTP regressions all passed. GitHub lint, installation, wheel, and docs jobs are green; the functional workflow is currently waiting at the repository's `test` environment gate.

yuchenwang3:
@janEbert the functional run has finished. The unit suite itself is green (`1892 passed, 106 skipped, 105 deselected`); the job only failed afterward because the coverage wrapper reported `No data was collected`. I tried rerunning the failed jobs, but GitHub requires repository admin permission. Could you rerun the failed jobs when you get a chance? No code assertion failed.

yuchenwang3:
@janEbert I rebased the six commits onto current main, including the newly merged #6754 optimizer changes. The patch replayed cleanly and preserves the same patch-id; the two changes operate at separate stages (#5395 narrows clipping before optimizer step, while #6754 syncs parameters afterward), so no API adjustment was needed.

Current head: `0f61a13574caf9c4dd40f26740867a8f7fd80f58`. Diff check, Black, isort, Pylint, Ruff, and Python compilation pass on the refreshed tree. Could you re-review and run `/ok to test 0f61a13574caf9c4dd40f26740867a8f7fd80f58` so the functional matrix reruns on the exact current-main head?

yuchenwang3:
@NVIDIA/mcore-oncall — I rewrote only the six commit identities/signatures to my personal GitHub noreply address; the source tree is byte-identical to the tested `0f61a135` head. The new head is `82ebe52f708bd4a0f948af8ff80357d3cd4256f2`, and DCO is green. This already has reviewer approval and the previous functional matrix reached the test stage. Could you route the current SHA for `/ok to test 82ebe52f708bd4a0f948af8ff80357d3cd4256f2` and the remaining CODEOWNER step?

yuchenwang3:
@janEbert Could you authorize CI for the current head, `1add0b2151507c5603a316b2e84a617e7ca16961`, when you have a chance? Your September 8 authorization covered `249a5aac1d3284c6119d3c8615f37c6c6c18219b`; the current head has only the basic checks and is still awaiting review. Please let me know if any author-side changes are needed before rerunning the functional tests.

PR #5396, opened 2026-06-17
perf(gated_delta_net): fold q/k L2-norm into the gated_delta_rule kernel

yuchenwang3:

## What

Move GatedDeltaNet q/k L2 normalization into the FLA `gated_delta_rule` kernel instead of materializing normalized q/k before the call.

The current GDN and GDN2 implementations share `_prepare_input_for_gated_delta_rule`, but GDN2's kernel path does not enable in-kernel normalization. The rebased change therefore makes deferral an explicit helper option:

- GDN defers normalization and passes `use_qk_l2norm_in_kernel=self.use_qk_l2norm`.
- GDN2 keeps its existing caller-side normalization and continues to pass `False` to its kernel.

This preserves GDN2 behavior added after the original PR rather than silently removing its normalization during the rebase.

## Why

Caller-side normalization leaves an additional pre-split normalized `query_key` activation live for backward alongside the q/k tensors the gated-delta kernel already needs. Letting FLA own normalization removes that extra caller-side activation and keeps the normalization state with the kernel-owned q/k, which saves activation memory in long-context GDN-heavy hybrids.

The numerical contract is unchanged:

- both caller-side FLA `l2norm` and the in-kernel path use `eps=1e-6`;
- normalization is per head over the last dimension, so GQA head replication does not change the result;
- `use_qk_l2norm=False` remains a no-op.

## Validation

The original change ran in Qwen3.5-35B-A3B 128K SFT on 16×B200. The current-main replay passes Black, isort, Ruff, Pylint, `py_compile`, and `git diff --check`.

The final `398a42b4` head was also exercised with an 8-rank torchrun on 8×B200:

- the GDN/GDN2 kernel flag-vs-explicit-normalization parity test passed on every rank;
- the BF16 end-to-end `GatedDeltaNet.forward` comparison passed on every rank, covering outputs, input gradients, all parameter gradients, and both routing flags;
- all 12 parameterized deterministic-mode GDN/GDN2 cases passed on every rank.


yuchenwang3:
@xuantengh pinging you as the most recent committer on `gated_delta_net.py`. This folds the explicit q/k `l2norm` into the FLA `gated_delta_rule` kernel via `use_qk_l2norm_in_kernel=True`, dropping a materialized intermediate at long context; it ran in real Qwen3.5-35B-A3B hybrid training at 128K. One-file change, open since Jun 17 — could you review or `/ok to test`?

xuantengh:
Thanks for the contribution! We had a refactor to the GDN module recently in #5843, would you please rebase that and apply this to the new file?

yuchenwang3:
Done, @xuantengh — merged latest `main` and re-applied the change onto the refactored layout from #5843:

- `common.py`: the explicit `l2norm()` in `_GDNBase._prepare_input_for_gated_delta_rule` is removed (replaced by a NOTE explaining the fold and the eps=1e-6 equivalence).
- `gdn.py`: the `gated_delta_rule` call in `GatedDeltaNet.forward` now passes `use_qk_l2norm_in_kernel=self.use_qk_l2norm`.

Net diff vs main is the same two-line semantic change as before (+7/−4 with the comment block). Ready for another look.

yuchenwang3:
@xuantengh one more follow-up before I refresh this again: current main added GDN2 sharing `_prepare_input_for_gated_delta_rule`. I've kept GDN2's caller-side normalization and only defer GDN to the kernel. Does that split match what you'd expect? If so I'll update the branch and request CI.

yuchenwang3:
Rebased onto current main and adjusted the shared input helper added since the original PR: GDN now defers q/k normalization to its kernel, while GDN2 keeps the existing caller-side normalization. Local formatting, syntax, and routing checks pass. @xuantengh, could you re-review this head when you get a chance?

xuantengh:
Sorry for the delayed reply. Would you please add a test to `tests/unit_tests/ssm/test_gated_delta_net.py` to validate toggling `use_qk_l2norm` can produce on-par results?

tests/unit_tests/ssm/test_gated_delta_net.py
```diff
@@ -54,6 +55,229 @@ def _unpack_sequence(x: torch.Tensor, cu_seqlens: torch.Tensor, dim=1) -> list[t
     return unpacked_x
 
 
+@pytest.mark.parametrize("use_gdn2", [False, True], ids=["gdn", "gdn2"])
+@pytest.mark.skipif(not HAVE_FLA, reason="FLA is not installed.")
+@pytest.mark.internal
+def test_torch_gdn_l2norm_flag_matches_explicit_normalization(use_gdn2):
```
xuantengh: I think this test is unnecessary if the module-level parity test is passed? Does this make sense?

yuchenwang3:
Done, @xuantengh. I added a small kernel-level parity check plus a BF16 end-to-end test that runs the real `GatedDeltaNet.forward` path. The latter compares the old caller-normalized route with kernel-owned normalization across outputs, input gradients, and every parameter gradient, and also asserts that both the helper and kernel receive the in-kernel flag and that raw q/k reach the kernel.

The new tests pass on all 8 ranks on 8×B200. While running the existing deterministic matrix I also caught and fixed a current FLA API mismatch in the newly activated fallback (`l2norm` has no `dim` argument); all 12 deterministic GDN/GDN2 cases now pass on every rank. The final head is `398a42b4` and DCO is green. Could you take another look and run `/ok to test 398a42b4af753c76ffeeb14198d9b41a1bb2de1a` when you get a chance?

xuantengh:
Hi @yuchenwang3 thanks for your contribution. Generally LGTM, but currently there is no way to specify `use_qk_norm_in_kernel` at the module spec level. So would you please also add this as an optional knob (default to `False`, aligned with upstream FLA interface signature) to `TransformerConfig` and `ModuleSpec`. And we can toggle this via config in test, rather than the `install_route_probe` hook.

yuchenwang3:
Done, @xuantengh — I rebased onto current `main` and added `TransformerConfig.use_qk_l2norm_in_kernel` with a `False` default, using the exact upstream FLA argument name. The value is forwarded through the GDN `ModuleSpec.params` and remains available from the config for other construction paths.

The BF16 parity test now builds the caller-normalized and kernel-normalized modules by toggling that config value through the real `ModuleSpec` path; the route-probe hook is gone. It still compares outputs, input gradients, and every parameter gradient. The module-spec test file is 47/47 locally, and formatting/static checks pass.

The updated head is `edf2ff50cbfdba177c5f27eb8d01a82d792a6bcd`. Could you take another look and run `/ok to test edf2ff50cbfdba177c5f27eb8d01a82d792a6bcd`?

yuchenwang3:
The current head `edf2ff50cbfdba177c5f27eb8d01a82d792a6bcd` has now completed the reviewer-requested focused validation on 8×B200 with `flash-linear-attention==0.5.1` (`HAVE_FLA=True`, `HAVE_FLA_GDN2=True`). Every rank finished with `15 passed, 0 failed, 0 skipped`: the GDN/GDN2 normalization parity cases, the BF16 end-to-end config/ModuleSpec route, and all 12 deterministic GDN/GDN2 cases. The logs show the Blackwell FLA/TileLang kernel path executing; there were no tracebacks, NaNs, or OOMs, and the source tree remained clean.

Could you take another look and run `/ok to test edf2ff50cbfdba177c5f27eb8d01a82d792a6bcd`?

megatron/core/transformer/transformer_config.py
```diff
@@ -402,6 +402,9 @@ class TransformerConfig(ModelParallelConfig):
     linear_num_value_heads: Optional[int] = 32
     """Number of value and gate heads for the gated delta net."""
 
+    use_qk_l2norm_in_kernel: bool = False
```
xuantengh: It's better to make this to `gdn_use_qk_l2norm_in_kernel` for a more informative name?

megatron/core/ssm/gated_delta_net/common.py
```diff
@@ -161,6 +164,11 @@ def __init__(
         assert A_init_range[0] >= 0 and A_init_range[1] >= A_init_range[0]
         self.A_init_range = A_init_range
         self.use_qk_l2norm = use_qk_l2norm
+        self.use_qk_l2norm_in_kernel = (
+            config.use_qk_l2norm_in_kernel
+            if use_qk_l2norm_in_kernel is None
+            else use_qk_l2norm_in_kernel
+        )
```
xuantengh: We may add an assertion here to raise an error if one has `use_qk_l2norm=False` but `use_qk_l2norm_in_kernel=True`?

tests/unit_tests/ssm/test_gated_delta_net.py
```diff
@@ -19,11 +20,167 @@
...
+
+
+@pytest.mark.skipif(not HAVE_FLA, reason="FLA is not installed.")
+@pytest.mark.internal
+def test_gdn_in_kernel_l2norm_matches_caller_end_to_end():
+    """GDN's BF16 in-kernel route must match the former caller-normalized route."""
+    Utils.initialize_model_parallel(
+        tensor_model_parallel_size=1, pipeline_model_parallel_size=1, context_parallel_size=1
+    )
+    try:
```
xuantengh: Why we wrap this test with `try`-`except`?

xuantengh (changes requested):

xuantengh:
/ok to test [edf2ff5](https://github.com/NVIDIA/Megatron-LM/commit/edf2ff50cbfdba177c5f27eb8d01a82d792a6bcd)

megatron/core/transformer/transformer_config.py
Phlip79: Yes, `use_qk_l2norm_in_kernel` is too generic.

megatron/core/models/gpt/experimental_attention_variant_module_specs.py
```diff
@@ -84,6 +84,7 @@ def get_gated_delta_net_module_spec(
     )
     attention = ModuleSpec(
         module=gdn_module,
+        params={"use_qk_l2norm_in_kernel": config.use_qk_l2norm_in_kernel},
```
Phlip79: This line is not needed. The GPT spec already passes config into GDN.

xuantengh:
@yuchenwang3 Would you please also check the failing CI. Is it related to the changes of this PR?

yuchenwang3:
@xuantengh I reproduced the reported failure on 8×B200. Across 3 repeated 8-rank runs, current main passed 24/24 and the PR tree passed 22/24. The two misses were both the existing non-GDN2 SBHD-vs-THD BF16 comparison: 6/131072 elements differed, with max abs error 0.00048828125 (one BF16 ULP; current tolerance 0.0003) and max relative error 0.03564453125. It was not deterministic, so I did not hide it by widening the tolerance.

The PR-specific normalization parity/end-to-end coverage and all deterministic GDN/GDN2 cases passed on every rank. Apex `fused_weight_gradient_mlp_cuda` imports successfully, and the HybridMoE constructor passed on all 8 ranks with 0 skips/failures. No traceback, NaN, or OOM was observed.

I also rewrote the four commit identities to my personal GitHub noreply address; the current head `f152abd9866da553a72d3b413fcdc7e69c7278f1` has a byte-identical source tree to the tested `3ce05c5f` head. DCO and the approval gate are green. Could you please re-review and `/ok to test f152abd9866da553a72d3b413fcdc7e69c7278f1`?

megatron/core/transformer/transformer_config.py
yuchenwang3: Renamed it to gdn_use_qk_l2norm_in_kernel in f152abd.

megatron/core/ssm/gated_delta_net/common.py
yuchenwang3: Added that assertion and a focused invalid-configuration test in f152abd.

tests/unit_tests/ssm/test_gated_delta_net.py
yuchenwang3: Replaced the try/finally block with a model-parallel fixture so teardown is still guaranteed without wrapping the test body.

megatron/core/models/gpt/experimental_attention_variant_module_specs.py
yuchenwang3: Removed this line in f152abd; the existing config path now supplies the GDN option.

xuantengh:
/ok to test [f152abd](https://github.com/NVIDIA/Megatron-LM/commit/f152abd9866da553a72d3b413fcdc7e69c7278f1)

xuantengh (approved):
LGTM.

zhongbozhu (approved):

zhongbozhu:
There seems to be a CI failure: `[CICD Megatron-LM / tests/unit_tests/ssm/test_gated_delta_net.py - latest (push)](https://github.com/NVIDIA/Megatron-LM/actions/runs/33468770732/job/99784635166?pr=5396)`

yuchenwang3:
Thanks @zhongbozhu — I checked the failed job. The failure was the existing `test_gpu_forward_thd_correctness[2-False-2-gdn]`; all PR-added normalization tests passed.

The old fixture reset CUDA RNG but not CPU RNG even though the module uses CPU initialization, so adding earlier tests changed the weights seen by this input-sensitive SBHD/THD comparison. This was independently fixed on `main` by #6961 with `torch.manual_seed(123)`. I rebased this PR onto current `main` to pick up that fix and dropped my duplicate patch. The new head is `3ba7c1477c`; it is mergeable, DCO is green, and Black/Ruff pass.

Could you please `/ok to test 3ba7c1477cec871da93c4a223040bfb820cb2e81`?

yuchenwang3:
@zhongbozhu @xuantengh follow-up on the new head: after rebasing onto #6961, the isolated case passed, but the full file exposed a second order dependency. The minimal reproducer was the first 23 collected nodes followed by `test_gpu_forward_thd_correctness[2-False-2-gdn]`; disabling Dynamo or resetting its process-global cache made it pass. This points to cumulative Dynamo cache state from the standalone compiled GDN tests, not a normalization regression.

I added fixture-scoped teardown cleanup after those standalone tests. It does not disable compilation or widen tolerances; it prevents their global Dynamo state from affecting the later TP/CP matrix.

Validated on 8×B200 at `3243bae5e57dda84ae328cac7001866b6d75a8f9`:
- historical failing prefix: every rank `22 passed, 2 skipped`; target passed on all 8 ranks
- full file, two consecutive runs: every rank `72 passed, 10 skipped, 0 failed`
- normalization coverage: 32/32 rank-test executions passed in each run
- no mismatch, NaN, or OOM

Could you please `/ok to test 3243bae5e57dda84ae328cac7001866b6d75a8f9` when you have a chance?

xuantengh:
/ok to test [72e0c40](https://github.com/NVIDIA/Megatron-LM/commit/72e0c40996bd77ee6d1a67236b030634f3d2c4d3)

Phlip79 (approved):

yuchenwang3:
Merged the latest main to resolve the conflicts, keeping the new backend coverage and the normalization regressions. Local Black/isort/Ruff and syntax checks pass; the previous 8-GPU results are not a fresh run of this merged head. @xuantengh could you authorize CI for b161f8fef17d1c4380c9bf171d2f9ef019239afc?

xuantengh:
Seems like there are new conflicts against main, @yuchenwang3 would you mind rebasing again?

yuchenwang3:
Synced with main again, including the backend rollback in #7326. The normalization regressions and Dynamo cleanup are preserved; the reverted backend code stays removed. Black/isort/Ruff, syntax checks, and a source-level test-preservation audit pass. This is not a fresh GPU run. @xuantengh could you authorize CI for a81333a06d9db277d970ea2eae2b5681310dab8e?

xuantengh:
/ok to test [a81333a](https://github.com/NVIDIA/Megatron-LM/pull/5396/commits/a81333a06d9db277d970ea2eae2b5681310dab8e)

guihong-nv (approved):

xuantengh:
/ok to test [192d5d3](https://github.com/NVIDIA/Megatron-LM/pull/5396/commits/192d5d36464927714223d95d6b603f48d1560a31)

yuchenwang3:
@xuantengh Thanks again for the reviews and enabling auto-merge. The CODEOWNERS check on `192d5d36464927714223d95d6b603f48d1560a31` still reports missing approval from `@NVIDIA/core-adlr`, and GitHub now also reports a conflict with main. Could you help route the remaining owner approval and coordinate the branch refresh? I have left the approved head unchanged so the existing review and CI state remain available. Happy to resolve the conflict and rerun the relevant checks if that is the preferred next step.

PR #5398, opened 2026-06-18
Fix fast-cache-load rank synchronization guard

sandyhouse:

# What does this PR do ?
Fix the `--dataloader-fast-cache-load` rank synchronization guard in `BlendedMegatronDatasetBuilder`.

The existing check used the dataset class object when testing for `GPTDatasetConfig`, which means the guard could fail to disable rank synchronization in some builder paths. This updates the check to inspect `self.config`, which is the actual dataset config instance.

## Motivation

`--dataloader-fast-cache-load` is intended to skip the rank-0-first synchronization path when loading dataset indices from an existing cache. However, the inner `_build_megatron_dataset_splits()` guard checked:

```python
isinstance(self.cls, GPTDatasetConfig)
```
self.cls is the dataset class, such as GPTDataset, not the config object. As a result, this condition is false even when self.config.fast_cache_load is enabled.

Some multi-prefix fast-cache cases may still work because higher-level builder paths already bypass synchronization, but the inner guard remains incorrect and can affect single-prefix paths, direct calls, or future refactors.
Changes

Use self.config instead of self.cls when checking for GPTDatasetConfig. Ensure synchronize_ranks is forced to False when --dataloader-fast-cache-load is enabled for GPT datasets.

### DIFF
--- a/megatron/core/datasets/blended_megatron_dataset_builder.py
+++ b/megatron/core/datasets/blended_megatron_dataset_builder.py
@@ -440,7 +440,7 @@ class BlendedMegatronDatasetBuilder(object):
             False
             if (
                 synchronize_ranks
-                and (isinstance(self.cls, GPTDatasetConfig) and self.config.fast_cache_load)
+                and (isinstance(self.config, GPTDatasetConfig) and self.config.fast_cache_load)
             )
             else synchronize_ranks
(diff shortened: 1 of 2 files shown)


asolergi-nv:
Hey! Thanks for catching this issue. Can you drop the tests? Thanks!

PR #5406, opened 2026-06-20
Fix merges_file kwarg name in HuggingFaceTokenizer

muyihao:

# What does this PR do ?

Fixes a keyword-argument typo in `HuggingFaceTokenizer`. When both `vocab_file` and `merges_file` are supplied, the tokenizer called `AutoTokenizer.from_pretrained` with the misspelled keyword `merge_files` instead of `merges_file`. Slow HF tokenizers such as `GPT2Tokenizer` require the correctly spelled keyword, so this code path silently failed for any user who passed both files. The fix passes `merges_file=` as the transformers API actually expects.

### Pre-checks


Phlip79 (approved):
HuggingFace uses [merges_file](https://github.com/huggingface/transformers/blob/v5.12.0/src/transformers/models/gpt2/tokenization_gpt2.py#L25-L27)

dimapihtar (approved):

deepakn94 (approved):

PR #5411, opened 2026-06-21
Add fully_shard_optimizer for mixed-precision FSDP

wujingyue:
## Summary
- Add `fully_shard_optimizer(optimizer)` and export it for the experimental `fully_shard` path.
- Adapt optimizer steps in place with pre/post hooks that temporarily cast FSDP-managed grads while the optimizer consumes them, then restore the original grad tensors.
- Reject optimizer step closures because the hook-only adapter cannot intercept grads materialized inside the closure.
- Add focused MFSDP v2 optimizer coverage for mixed param/grad precision:
  - raw `torch.optim.Adam` fails on the mismatch without adaptation
  - TE `FusedAdam` accepts mismatched grads and updates parameters
  - `fully_shard_optimizer` covers Adam mixed-precision stepping through the adapter

## Notes
- `DistributedOptimizer` is intentionally not used for the MFSDP v2 test. With `use_megatron_fsdp=True`, it delegates to the wrapped optimizer for this path; with `False`, it expects the older `_ParamAndGradBuffer` setup rather than v2 `FsdpParameterGroup`/`DBuffer` state.
- A follow-up can extend the adapter to sync `main_weight` to `model_weight` from a post-step hook. This PR keeps the current per-microbatch sync behavior.

## Testing
- `BASE_REF=main CHECK_ONLY=true SKIP_DOCS=false bash tools/autoformat.sh`
- `uv run --no-sync python -m torch.distributed.run --nproc-per-node 1 -m pytest -q tests/unit_tests/distributed/megatron_fsdp/test_optimizer.py --capture=fd --tb=short --disable-warnings -rN` (`2 passed`)
- `uv run --no-sync python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/megatron_fsdp/test_optimizer.py --capture=fd --tb=short --disable-warnings -rN` (`2 passed` on each rank)
- `uv run --no-sync python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/megatron_fsdp/test_experimental_fully_shard.py -k "adam or optimizer" --capture=fd --tb=short --disable-warnings -rN` (`2 passed, 8 deselected` on each rank)

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
@@ -14,18 +14,15 @@
 
 """Optimizer adapter for the minimal Megatron-FSDP path."""
 
-from dataclasses import dataclass
-from typing import Any, TypeVar
+from typing import Any, NamedTuple
 
 import torch
 from torch import nn
 
 from .parameter_group import contained_in_parameter_group
 
-_OptimizerT = TypeVar("_OptimizerT", bound=torch.optim.Optimizer)
 
-
-def fully_shard_optimizer(optimizer: _OptimizerT) -> None:
+def fully_shard_optimizer(optimizer: torch.optim.Optimizer) -> None:
     """Attach FSDP-aware step hooks to an optimizer instance.
 
     The adapted optimizer preserves its existing parameter groups and only adds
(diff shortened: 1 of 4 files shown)


shjwudp (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def fully_shard_optimizer(optimizer: torch.optim.Optimizer) -> None:
+    """Attach FSDP-aware step hooks to an optimizer instance.
+
+    The adapted optimizer preserves its existing parameter groups and only adds
+    temporary gradient casting around optimizer steps for FSDP sharded
+    parameters whose data dtype differs from their grad dtype.
+
+    Alternatives considered:
```
wujingyue, line 32: @shjwudp: Also mention that the cast from main_grad.dtype (typically 16-bit or lower) to main_weight.dtype (typically fp32) can be done during the last microbatch and back during the first microbatch. 

Question: If we do this, won't the high-precision sharded grads be alive for too long? For example, the sharded grad of the last layer (therefore the first to run in backprop) will be alive for the entire backprop?! Maybe for drop-in replacement we just set main_weight.dtype to param.dtype. This will get rid of the cast entirely. Wdyt?

shjwudp: Yes, alternatively we could handle this in the post-backward final callback to minimize the window as much as possible.

For the drop-in replacement scenario, the main weight is not needed at this point, so indeed no cast is required for now.

wujingyue: > we could handle this in the post-backward final callback to minimize the window as much as possible.

Oh, the final callback happens on a **root** FsdpModule so it can find **all** sub-FsdpModules and their ParameterGroups. Makes sense.

Phlip79 (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
wujingyue: Done -- incorporated to code comment.

deepakn94 (approved):

PR #5417, opened 2026-06-22
Add CUDA graph training iteration test

### LINKED ISSUE #5646: CUDA graph support for MFSDP v2
MFSDP v2 should support CUDA graphs across the relevant CUDA graph modes and integration surfaces.

Use-case matrix:

| | Standalone MFSDP v2 `fully_shard()` | Megatron Core training |
|---|---|---|
| Full-iteration CUDA graph | Not yet supported | To support after #5614 |
| Partial/per-layer CUDA graph | Not yet supported | To support after #5614 |

We should decide which combinations need first-class support. The Megatron Core cases are dependent on #5614, so standalone MFSDP v2 CUDA graph work should not block on MCore integration.

Partial CUDA graph support is important for MoE. Per-expert routing can create dynamic shapes, making full MoE layer capture difficult. Without more sophisticated techniques like paged stashing (#4247), CUDA graph capture for MoE layers usually requires oversized static buffers and can use a lot of memory. Partial capture lets users graph static subregions while leaving dynamic expert work outside capture.

Relevant starting point:

- `tests/unit_tests/distributed/megatron_fsdp/test_cuda_graph.py` already covers a standalone full training iteration with experimental MFSDP v2.

### LINKED ISSUE #5717: Full-iteration CUDA graph support for MFSDP v2
https://github.com/NVIDIA/Megatron-LM/pull/5417 already tested that for a toy model. However, we need to test this more thoroughly. 

See the parent issue for context.

wujingyue:
## Summary
- Add a CUDA graph capture/replay test for a full sharded training iteration.
- Warm up and capture on an explicit side stream, then replay on the default stream while snapshotting losses on GPU before one host readback.
- Keep the test on the new fully_shard path with SGD and fixed static tensors.

## Validation
- BASE_REF=pull-request/5387 CHECK_ONLY=true SKIP_DOCS=false bash tools/autoformat.sh
- python -m py_compile tests/unit_tests/distributed/megatron_fsdp/test_experimental_fully_shard.py
- python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/megatron_fsdp/test_experimental_fully_shard.py::test_cuda_graph_captures_full_training_iteration --log-cli-level=INFO

Note: autoformat check passed Black, isort, pylint, and ruff; mypy is not installed in this local environment and autoformat.sh treats that as non-fatal.

Related issue: #5646

### DIFF
--- a/tests/unit_tests/distributed/megatron_fsdp/test_experimental_fully_shard.py
+++ b/tests/unit_tests/distributed/megatron_fsdp/test_experimental_fully_shard.py
@@ -245,6 +245,65 @@ def test_next_forward_uses_optimizer_updated_weights(distributed_setup):
         torch.testing.assert_close(second_loss, first_loss)
 
 
+def test_cuda_graph_captures_full_training_iteration(distributed_setup):
+    """A full training iteration should be CUDA-graphable."""
+    world_size = distributed_setup.world_size
+    device = distributed_setup.device
+    if world_size < 2:
+        pytest.skip("This test requires at least 2 ranks.")
+
+    mesh = init_device_mesh(device.type, (world_size,))
+    torch.manual_seed(1234)
+    model = nn.Linear(4, 2, bias=False).to(device)
+
+    fully_shard(model, mesh=mesh, placements=_flat_placements())
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.25, foreach=False)
+
+    static_input = torch.eye(4, device=device)
+    static_target = torch.tensor(
+        [[1.0, -0.5], [-0.25, 0.75], [0.5, 0.25], [-0.75, -1.0]], device=device
+    )
+
+    def train_iteration() -> torch.Tensor:
+        optimizer.zero_grad(set_to_none=False)
+        output = model(static_input)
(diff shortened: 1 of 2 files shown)


tests/unit_tests/distributed/megatron_fsdp/test_experimental_fully_shard.py
```diff
@@ -245,6 +245,65 @@ def train_iteration() -> torch.Tensor:
...
+    # capture and does not introduce default-stream dependencies:
+    # https://docs.nvidia.com/dl-cuda-graph/troubleshooting/capture-failures.html
+    with torch.cuda.stream(capture_stream):
+        # The first warmup installs the reusable sharded gradient views; subsequent
+        # iterations zero them in place for CUDA graph replay.
+        for _ in range(3):
+            train_iteration()
+
+    graph = torch.cuda.CUDAGraph()
+    with torch.cuda.graph(graph, stream=capture_stream):
```
wujingyue: `stream=` may not be needed.

shjwudp (approved):
LGTM. I’ve been wanting this UT for a while—thanks for adding it!

Phlip79 (approved):

PR #5420, opened 2026-06-22
Track effective sample size

yobibyte:
Add ESS computed from IS weights to wandb.


train_rl.py
```diff
@@ -172,6 +180,14 @@ def loss_func(
     reporting_truncated_from_below = torch.cat(
         [masked_truncated_from_below.clone().detach().view(1), total_tokens.view(1)]
     )
+    reporting_is_weight = torch.cat([masked_is_weight.clone().detach().view(1), total_tokens.view(1)])
+    reporting_is_weight_sq = torch.cat([masked_is_weight_sq.clone().detach().view(1), total_tokens.view(1)])
+    reporting_is_ess_frac = torch.cat( [ masked_is_ess_frac.clone().detach().view(1), torch.ones_like(total_tokens, dtype=torch.float).view(1), ])
```
tdene, line 185: Will this not break with `num_microbatch > 1`?

If microbatch A has ESS fraction 1.0 because the IS weights are `[1,1,1,1]`, and microbatch B has ESS fraction 1.0 because the IS weights are `[2,2,2,2]`, this reports ESS fraction of 1.0 across the entire batch.

yobibyte: I am not sure what the problem is. Both microbatches [1,1,1,1] and [2,2,2,2] should have ESS of 1 because the weights are evenly distributed.

tdene: Sounds good, approved.

tdene (approved):

dimapihtar (approved):

PR #5400, opened 2026-06-18
fix(optimizer): route GDN in_proj to Adam under Muon

yuchenwang3:
- [x] I, the PR author, have personally reviewed every line of this PR.

## Summary

Routes the shared GatedDeltaNet/GatedDeltaNet2 `in_proj` weight to the scalar optimizer instead of Muon by setting the existing `param.use_muon = False` marker.

## Why

`in_proj` is a heterogeneous fused matrix containing q, k, v, convolution, gate, and beta projections. Orthogonalizing the whole matrix as one 2D block is not meaningful. `out_proj` remains on Muon.

This addresses the open routing question in #2885 without adding another optimizer-group mechanism. Current `main` already consumes `use_muon=False` in `_is_muon_excluded`, so the original generic `skip_orthogonalization` change is no longer needed; this PR now only tags the GDN parameter at its source.

## Change

- Set `self.in_proj.weight.use_muon = False` in `_GDNBase`, covering both GDN and GDN2.
- Extend the existing parameterized constructor test to assert that the fused input projection is excluded from Muon.

## Validation

- Repository autoformat checks passed: Black, isort, Pylint, and Ruff.
- `git diff --check` and Python syntax compilation passed.
- On 8×B200, the 8-rank parameterized GDN/GDN2 constructor test completed with 12 passed, including the `in_proj.weight.use_muon = False` assertion.

PR #5431, opened 2026-06-22
fix(optimizer): isolate GDN in_proj from global grad-norm clip

yuchenwang3:

# fix(optimizer): isolate GDN `in_proj` from the global grad-norm clip

## Problem

Global gradient clipping uses one norm for every parameter. In GatedDeltaNet hybrid training I observed `in_proj` gradients around `5e7`; that single tensor then drives the global clip coefficient close to zero and effectively starves the rest of the model. Disabling clipping avoids that failure mode but also removes the safety net from `in_proj` itself.

## Change

This reuses Megatron's existing separate grad-norm-group path, which is already used for MTP:

- add an immutable built-in `gated_delta_net` grad-norm group;
- add the default-off `gated_delta_net_separate_grad_norm` config flag (also exposed as `--gated-delta-net-separate-grad-norm`);
- when enabled, tag the shared `_GDNBase.in_proj`, covering both GDN and GDN2.

The tagged projection is excluded from the main norm and clipped against its own norm. With the flag off, existing behavior is unchanged. The group is built in rather than registered dynamically so every rank follows the same collective sequence even when only some pipeline stages own a GDN layer.

This is complementary to #5395 and #5400: those control Muon clipping and optimizer routing, while this also addresses the pure-Adam global-clip failure.

## Tests

- Added a clip-outcome regression showing that a large GDN gradient is clipped without shrinking a normal gradient.
- Added default-off/enabled tagging coverage on the current shared GDN/GDN2 implementation.
- Rebased on current `main`; Black, isort, Ruff, `py_compile`, `git diff --check`, and focused structural checks pass locally.
- On 8×B200, the focused launcher exited 0 on the current head. Every rank reported `3 passed, 0 failed, 10 skipped`: the clip outcome passed on 8/8 ranks and the applicable GDN/GDN2 tagging cases passed on 16/16 rank-case executions. The skipped layout variants are intentionally out of scope for layout-independent tagging.
- An 8-rank DistOpt Muon+Adam constructor smoke reaches a `ChainedOptimizer` on every rank with both `LayerWiseDistributedOptimizer`/`TensorParallelMuon` and `DistributedOptimizer`/`FusedAdam` children, exercising the mixed routing this change must coexist with.

An additional fixed-batch 8-rank A/B used identical initial weights, seed, and batch with real GDN/GDN2 forward/backward, DDP, Muon + Adam, and `clip_grad=0.005`. Across steps 2–6, separating the GDN norm raised the main clip coefficient by 8.9%–16.4%, while representative GDN and non-GDN parameters updated on every step; final loss was `0.007268` with the flag off and `0.007166` with it enabled. This confirms that the GDN norm no longer depresses the main clipping coefficient. The natural small-model workload did not reproduce complete non-GDN starvation, so no starvation-rescue claim is made. Save/resume was also exercised within the predeclared BF16 smoke tolerances; details are in the PR discussion.

The asymmetric pipeline case was exercised separately with PP2/DP4/TP1/CP1 on 8×B200. Pipeline stage 0 (ranks 0–3) owned 8,512 tagged GDN parameters while stage 1 (ranks 4–7) owned none. All ranks completed three BF16 forward/backward, DP synchronization, distributed Adam, and group-clipping steps without a collective mismatch or hang. Loss decreased from `0.0106501` to `0.0103764`; the main clip coefficient rose from `0.7006` to `0.7957`, while the independently computed GDN coefficient remained `1.0`. Parameters updated on both stages, with no NaN, Inf, or OOM. This is a focused mechanism harness, not a full-model convergence claim.

### DIFF
--- a/megatron/core/optimizer/optimizer.py
+++ b/megatron/core/optimizer/optimizer.py
@@ -96,8 +96,37 @@ def _multi_tensor_copy_this_to_that(
 
 param_group_identifier_keys = ('wd_mult', 'lr_mult', 'is_expert_parallel', 'is_decoupled_lr')
 MTP_GRAD_NORM_GROUP = 'mtp'
+GDN_GRAD_NORM_GROUP = 'gated_delta_net'
 GRAD_NORM_GROUP_ATTR = 'grad_norm_group'
-SEPARATE_GRAD_NORM_GROUPS = (MTP_GRAD_NORM_GROUP,)
+
+# Parameters tagged with one of these groups are excluded from the *global* gradient
+# norm and clipped independently against their own group norm. This isolates
+# pathological-gradient parameters -- e.g. a GatedDeltaNet ``in_proj`` whose gradient
+# is empirically several orders of magnitude larger than every other tensor -- so that
+# a single exploding tensor cannot dominate the global clip coefficient and starve
+# (drive toward zero the update of) every other parameter. The mechanism behind the
+# large in_proj gradient is under separate investigation; this isolation is
+# cause-agnostic and only relies on the observed magnitude.
+#
+# The registry must stay globally consistent: it is iterated to issue per-group
+# collectives, so every rank has to hold the same groups in the same order. Keep
+# group names as module-level constants registered at import time, or call
+# ``register_grad_norm_group`` from model-construction code that runs on all ranks.
+SEPARATE_GRAD_NORM_GROUPS = [MTP_GRAD_NORM_GROUP, GDN_GRAD_NORM_GROUP]
+
+
+def register_grad_norm_group(name: str) -> str:
+    """Register ``name`` as a separate gradient-norm group (idempotent).
+
+    Tag parameters with ``param.grad_norm_group = name`` to have every optimizer
+    variant exclude them from the main grad norm and clip them against their own
+    group norm. Must be called identically on all ranks (e.g. at model
+    construction) so the per-group collectives issued in
+    ``_compute_grad_norms_by_group`` stay aligned across ranks.
+    """
+    if name not in SEPARATE_GRAD_NORM_GROUPS:
+        SEPARATE_GRAD_NORM_GROUPS.append(name)
+    return name
 
 
 def _get_param_grad_norm_group(param: torch.nn.Parameter) -> Optional[str]:
@@ -109,8 +138,8 @@ def _validate_grad_norm_group(grad_norm_group: str) -> None:
     """Raise if the grad-norm group is not registered for separate clipping."""
     if grad_norm_group not in SEPARATE_GRAD_NORM_GROUPS:
         raise ValueError(
-            f"Unknown grad_norm_group '{grad_norm_group}'. Register it in "
-            "SEPARATE_GRAD_NORM_GROUPS before tagging parameters with it."
+            f"Unknown grad_norm_group '{grad_norm_group}'. Register it with "
+            "register_grad_norm_group() before tagging parameters with it."
         )
 
 
--- a/megatron/core/ssm/gated_delta_net.py
+++ b/megatron/core/ssm/gated_delta_net.py
@@ -165,6 +165,17 @@ class GatedDeltaNet(MegatronModule):
             name=(name + ".in_proj") if name is not None else None,
         )
 
+        if getattr(self.config, "gated_delta_net_separate_grad_norm", False):
+            # Isolate in_proj grads from the global grad-norm clip: in_proj's
+            # gradient is empirically orders of magnitude larger than other tensors,
+            # which would otherwise dominate the global clip coefficient and starve
+            # every other parameter. Clip it against its own norm via a dedicated
+            # grad-norm group. (Exact cause of the large gradient is under separate
+            # investigation; this isolation is cause-agnostic.) Mirrors the MTP idiom;
+            # literal kept in sync with optimizer.optimizer.GDN_GRAD_NORM_GROUP.
+            for param in self.in_proj.parameters():
+                param.grad_norm_group = 'gated_delta_net'
+
         # Conv1d for QKV
         self.conv_dim = self.qk_dim * 2 + self.v_dim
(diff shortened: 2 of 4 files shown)


yuchenwang3:
@fanshiqing pinging you as a recent `optimizer.py` maintainer. This isolates pathological-gradient params (GatedDeltaNet `in_proj`) from the global grad-norm clip so one layer's spikes stop scaling down every other param's update; mechanism plus unit tests are in. It was opened early specifically to confirm the design direction and hasn't had maintainer feedback in ~6 weeks — a quick read on whether the approach is right would unblock the remaining convergence-evidence work.

yuchenwang3:
@fanshiqing before I rework this on the current GDN layout, is isolating the GDN `in_proj` grad-norm group still a direction you'd consider? If not, I'm happy to close it rather than carry the old branch forward.

yuchenwang3:
@fanshiqing I rebased this onto the current GDN/GDN2 layout and removed the old mutable registry design. The branch now uses a built-in group, a default-off config flag, and includes an actual clip-outcome regression in addition to the tagging test. Static checks pass locally; the distributed CUDA cases still need runner approval. Could you take another look and, if the direction makes sense, run `/ok to test f6841a5db4427f99cce5588ce7d1979e3f36436f`?

yuchenwang3:
Ran the rebased `f6841a5d` head on 8×B200. The 8-rank clip-outcome test completed cleanly, and the GDN/GDN2 tagging test passed all applicable cases; the remaining parametrizations skip by design because tagging is independent of the parallel layout.

I also exercised the mixed DistOpt path with Muon + Adam. Every rank built a `ChainedOptimizer` containing both a `LayerWiseDistributedOptimizer` backed by `TensorParallelMuon` and a `DistributedOptimizer` backed by `FusedAdam`, with no hang or traceback. That covers the mixed optimizer routing this needs to coexist with. The longer fixed-batch convergence and save/resume checks are still separate follow-ups.

yuchenwang3:
@NVIDIA/mcore-oncall could you help route this to an optimizer/GDN reviewer and authorize CI? The current design uses a default-off built-in grad-norm group, with 8×B200 coverage for the clip outcome, GDN/GDN2 tagging, and mixed Muon + Adam distributed optimizer construction. This has not had maintainer feedback yet.

yuchenwang3:
@NVIDIA/mcore-oncall follow-up: the remaining production-path checks are complete on the current head.

On 8×B200, all 8 ranks passed a fixed-batch six-step run through real GDN/GDN2 forward/backward, DDP reduce-scatter, mixed `LayerWiseDistributedOptimizer` (Muon) + `DistributedOptimizer` (Adam), and `clip_grad=0.005`. Loss remained finite and decreased in both modes:

- default-off: `0.009775 -> 0.007268`
- separate GDN norm enabled: `0.009775 -> 0.007166`

With the flag enabled, both the main norm and separate GDN norm exceeded the clip threshold during the run, so this exercised actual clipping rather than only tagging. GDN/GDN2, representative Muon, and Adam parameters updated on every step.

I also compared continuous 6 steps with `3 steps -> distributed checkpoint save/load -> 3 steps`. The first resumed loss was bitwise identical and scheduler state matched exactly. Final BF16 parity was within the predeclared smoke tolerances, but not bitwise: max loss error `1.70e-4`, max representative parameter error `0.0048828125`, and max optimizer-state error `0.00547` (limits: loss `2e-4`, parameter/optimizer `0.015`). There was no NaN, Inf, OOM, or hang.

Could you route this for optimizer/GDN review and `/ok to test f6841a5db4427f99cce5588ce7d1979e3f36436f`?

yuchenwang3:
Revalidated the current rebased head on 8×B200. The focused launcher exited 0 with `3 passed, 0 failed, 10 skipped` per rank: clip outcome 8/8 and applicable GDN/GDN2 tagging 16/16.

I also covered the asymmetric PP case with a focused PP2/DP4 harness: ranks 0–3 owned 8,512 tagged GDN parameters, ranks 4–7 owned none, and all eight ranks completed three forward/backward, DP-sync, distributed-Adam, and group-clipping steps without collective mismatch, hang, NaN, Inf, or OOM. The PR description now records the scope and exact boundary of that evidence.

@NVIDIA/mcore-oncall could you route this for optimizer/GDN review and run `/ok to test 3b69c0e0c56134335f61952f14c0366ce2e7e4ea`?

yuchenwang3:
@NVIDIA/mcore-oncall, could you help route this separate-GDN clipping PR to an optimizer reviewer? The latest validation comment includes the asymmetric PP2/DP4 case, including ranks with no GDN parameters. The fixed-batch evidence supports clipping separation, not a claim of full-model starvation rescue. A design review and maintainer CI authorization would help move this forward. Thanks.

yuchenwang3:
@ShauryaaSharma Could you review this opt-in separate clipping group for GDN `in_proj`, or point me to the right optimizer reviewer? I rebased it at `ceee3a9574`; local lint checks pass. The earlier PP2/DP4 validation is recorded above; I haven't rerun it on this head.

yuchenwang3:
Rebased onto `c343bb6d337eae14e195835262854361680bcab8`; new head: `2f9217e3a4fdc0913af88f9f184bce8c0cdb6c26`.

Preserved the upstream GDN fusion settings and added the `_uses_decoupled_grad` delegation required by the updated clipping path to the regression fixture. The production change remains the default-off, separate GDN clipping group.

Black, isort, Ruff, Pylint, source compilation and whitespace checks pass. Six bounded CPU routing checks pass, including an old-fixture negative control. These use the actual optimizer methods and fixture with CPU norm/clip and single-rank adapters; they do not exercise GPU kernels or collectives. The isolated mypy run produces exactly the same 46 diagnostics as the unmodified base, with no additional diagnostics.

The PR description now explicitly distinguishes these fresh checks from the earlier B200 results. GPU/TE/FLA and distributed tests have not been rerun on this head; upstream CI is still needed.

PR #5453, opened 2026-06-23
Implement async scheduling for dynamic inference

lmcafee-nvidia:
## Summary

This PR adds an opt-in async scheduling mode for dynamic inference. The legacy path remains the default and keeps the existing request update behavior intact.

## Motivation

The current dynamic inference loop resolves request lifecycle changes immediately after sampling:

```text
sample(N) -> resolve(N) -> prepare(N+1) -> forward(N+1)
```

Async scheduling enables the decode-only path to run the next prepare/forward step speculatively before applying request lifecycle changes from the previous sample:

```text
sample(N) -> prepare(N+1) -> forward(N+1) -> resolve(N)
```

This lets the steady-state decode path move request lifecycle bookkeeping after the next forward pass when the active request set remains compatible with the speculative step. When requests finish, the controller compacts the cached forward output so only survivor rows are retained for the next sampling step.

## Architecture

- Adds `AsyncScheduleMode` to `InferenceConfig`, with `legacy` as the default and `serial` as the opt-in async scheduling mode.
- Adds the CLI flag `--inference-dynamic-batching-async-sched-mode {legacy,serial}` and wires it into inference config construction.
- Preserves the legacy `DynamicInferenceContext.update_requests` behavior as the compatibility path.
- Splits the new request bookkeeping surface into:
  - `resolve_requests`, which applies lifecycle changes after the speculative forward step.
  - `prepare_requests`, which advances currently active requests by one token and allocates new blocks as needed.
- Makes `TextGenerationController.async_generate_output_tokens_dynamic_batch` route between the legacy path and the async-scheduled decode-only path.
- Adds a forward primer so the async scheduling path has a valid previous forward result before entering the speculative steady-state loop.
- Keeps async scheduling guarded to decode-only cases and unsupported feature combinations so the legacy path remains the correctness path for broader workloads.


megatron/training/config/inference_config.py
```diff
@@ -136,6 +136,11 @@ class InferenceSetupConfig:
     """Which sampling kernels to use during inference. Falls back to "torch" with a warning if
     "flashinfer" is requested but the package is not installed."""
 
+    inference_dynamic_batching_request_resolution_mode: Literal["legacy", "defer"] = "legacy"
+    """Request lifecycle resolution mode for dynamic batching. "legacy" (default) preserves the
+    existing resolve-before-prepare path. "defer" speculatively prepares and forwards decode-only
+    greedy GPT steps before resolving finished requests."""
```
jaredcasper: Use of "GPT" here seems weird, is there a better way to describe this?

cuichenx (approved):

shanmugamr1992 (approved):

maanug-nv (approved):

santhnm2 (approved):

kvareddy (approved):

megatron/training/config/inference_config.py
lmcafee-nvidia: thanks, fixed!

jaredcasper (approved):

ko3n1g (approved):

PR #4244, opened 2026-04-09
Add the YARN support for hybrid_model

guihong-nv:
# Add YaRN positional embedding support to MambaModel

## Summary

- **`megatron/core/models/mamba/mamba_model.py`**: Extended `position_embedding_type` to accept `'yarn'`. Adds a `YarnRotaryEmbedding` in `__init__` and calls it in `forward` (discarding the returned mscale), mirroring the existing RoPE path.
- **`mamba_builders.py`**: Injects YaRN-specific config attributes (`yarn_rotary_scaling_factor`, `yarn_original_max_position_embeddings`, `yarn_beta_fast`, `yarn_beta_slow`, `yarn_mscale`, `yarn_mscale_all_dim`, `yarn_correction_range_round_to_int`) onto `TransformerConfig` before `MambaModel` is constructed, so `pretrain_mamba.py` can accept `--position-embedding-type yarn`.
- **`megatron/training/arguments.py`**: Added `'yarn'` to the allowed values for `--position-embedding-type`.
- **`tests/unit_tests/models/test_mamba_model.py`**: Added `TestMambaModelWithYarn` with three tests — constructor (checks `YarnRotaryEmbedding` is created), forward pass, and inference (static context, prefill + decode steps).

### Pre-checks

- [x] I have added relevant unit tests
- [x] I have added relevant functional tests
- [x] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [x] I have added relevant documentation
- [x] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

[...]

PR #5461, opened 2026-06-23
Add formal YaRN RoPE config fields and CLI plumbing for hybrid models

guihong-nv:

# What does this PR do ?

Declares the YaRN RoPE parameters as formal `TransformerConfig` fields so that `--position-embedding-type yarn` is actually usable for hybrid models.

`HybridModel` already instantiates `YarnRotaryEmbedding` from `getattr(self.config, "yarn_*")` (added in #4244), but those `yarn_*` attributes were never declared on `TransformerConfig` — they were only set dynamically in tests. As a result, launching with `--position-embedding-type yarn` raised `AttributeError`.

- **`megatron/core/transformer/transformer_config.py`** — declares the 7 YaRN fields on `TransformerConfig` with safe defaults (`yarn_rotary_scaling_factor`, `yarn_original_max_position_embeddings`, `yarn_beta_fast`, `yarn_beta_slow`, `yarn_mscale`, `yarn_mscale_all_dim`, `yarn_correction_range_round_to_int`). The default `yarn_rotary_scaling_factor=1.0` is a no-op, so existing configs are unaffected.
- **`tests/unit_tests/models/test_hybrid_model.py`** — drives the YaRN test config via the constructor (the formal-field path) and adds two tests: one asserting the fields exist as dataclass fields with the expected defaults, and one asserting the config values flow into the `YarnRotaryEmbedding` instance.

The matching `--yarn-*` CLI flags are produced **automatically**: `_add_network_size_args` builds a `TransformerConfig` argument group via `ArgumentGroupFactory(TransformerConfig)`, which derives one CLI flag (and the args→config plumbing) per dataclass field. No manual argument plumbing is needed — adding the flags by hand would duplicate the auto-generated ones and raise `argparse: conflicting option string`.

No behavior change for non-YaRN paths. Validated on GPU: the full arg parser builds with the `--yarn-*` flags auto-generated (no conflict), and `tests/unit_tests/models/test_hybrid_model.py` plus `tests/unit_tests/transformer/test_multi_latent_attention.py` pass (one pre-existing, environment-related context-parallel SIGABRT reproduces on unmodified `main` and is unrelated to this change).

## Issue tracking

Related to #4244 (which added the YaRN wiring in `HybridModel`).

### Pre-checks

### DIFF
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -226,6 +226,30 @@ class TransformerConfig(ModelParallelConfig):
     """True is rotate pairs of even and odd dimensions (RoFormer style), False is rotate pairs of
     first half and second half (LLaMa style). Default to False."""
 
+    yarn_rotary_scaling_factor: float = 1.0
+    """Scaling factor for YaRN RoPE, used when position_embedding_type == 'yarn'. Default 1.0
+    (no scaling)."""
+
+    yarn_original_max_position_embeddings: int = 4096
+    """Original (pre-extension) maximum sequence length the model was trained on, used by YaRN
+    RoPE when position_embedding_type == 'yarn'."""
+
+    yarn_beta_fast: float = 32.0
+    """Fast beta value for YaRN RoPE, used when position_embedding_type == 'yarn'."""
+
+    yarn_beta_slow: float = 1.0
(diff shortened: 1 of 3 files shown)


guihong-nv:
/ok to test [8ec9a40](https://github.com/NVIDIA/Megatron-LM/pull/5461/commits/8ec9a400f45f7bf286af688ec46cc6e5119fb1e4)

PR #5462, opened 2026-06-23
[Megatron-FSDP] MaxPoolAllocator for double-buffering hybrid architectures.

cspades:

# What does this PR do ?

image

> Iterating through all FSDP units, data buckets are categorized by data-type, sorted from small to large, and compared to the current MaxPool. If there are not enough buckets in the pool to support the unit, buckets are added to the pool (with size 0). If the largest buckets of the pool are not large enough to support the buckets in the unit (assigned to the pool from smallest to largest), the buckets in the pool are enlarged. After this process, we arrive at a minimal set of buckets that can symmetrically double-buffer every FSDP unit in the model.

- Adds hybrid architecture double buffering via FSDP unit max-pooling for Megatron-FSDP. (V1)
  - Opens up CG or NCCL UBR support for hybrid architectures, which will help support users for a while.
- Adds the `strict_assignment` state to attempt to assign the same bucket previously assigned to an FSDP unit before warning the user and assigning a different bucket to the unit.
  - If this warning appears during warmup or CUDA graph capture, likely some memory is being orphaned and you will hit numerical errors.
- Fixes an issue where parameters / buckets that are not members of an FSDP unit will pre-fetch subsequent buckets that aren't subsequently used, exhausting buffers in the double buffer allocator and causing an allocation error.
  - Only necessary for double buffer allocators, which require careful management of the 2 buffers in the pool.
- Deprecates `--grad-reduce-in-bf16` / `reduce_grad_in_fp32` for Megatron-FSDP, which has been incredibly confusing to use. Default arguments (`auto`) assume BF16 for both, so will not OOM any existing user's configs.
- Deprecate `fsdp_unit_id == -1`. It is never set to -1.
- Adds a call to `torch.autograd.graph.set_override_stale_capture_stream(True)` (only supported on new PyTorch versions since https://github.com/pytorch/pytorch/pull/180090) to prevent full-iteration CG errors like this:
```
[rank0]: RuntimeError: During CUDA graph capture, autograd node 'torch::autograd::AccumulateGrad' has a stale reference to the default stream (stream 0) from warmup. This will invalidate the capture because cudaStreamWaitEvent on the default stream pulls a non-capturing stream into the graph.

[rank0]: To fix, either:
[rank0]:   (a) Run warmup on the same stream that capture will use, or
[rank0]:   (b) Delete references to the loss / autograd graph (e.g. `del loss`) before capture, or
[rank0]:   (c) Call torch.autograd.graph.set_override_stale_capture_stream(True) to automatically redirect stale nodes to the capturing stream.
```
^ (a) is annoying to implement, (b) is dirty, and (c) is EZ-PZ and recommended.

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -4023,11 +4433,16 @@ def all_gather_params(
...
+            if fsdp_unit_id is not None and fsdp_unit_id >= 0:
+                no_fsdp_units = False
 
         # If prefetch is enabled, we will add prefetch buckets to ag_buckets.
-        if prefetch:
+        # If there are no FSDP units associated with params, we should not prefetch.
+        if prefetch and not no_fsdp_units:
 
             def next_bucket_id(ag_buckets):
                 """
```
cspades, line 4445: @shjwudp Please take a look at this code. It is a behavior change in how we pre-fetch.

Without this change, Nemotron will hit an error where a parameter owned by a module that is not an FSDP unit will pre-fetch buckets that will not be used for a long time. In this case, I believe the `LanguageModelEmbedding` will pre-fetch Layer 2 during the last MTP layer of the model.

This pre-fetch of an irrelevant FSDP unit will cause both buffers in the double buffer pool to be used, preventing our code from allocating the MTP layer because we do not have enough free buckets to support it.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -2122,24 +2517,40 @@ def _init_each_parameter_group_buffers(self, meta_device_init_fp8_params):
...
-            self.main_grad_alloc = FixedPoolAllocator(
+            self.main_grad_alloc = FIXED_POOL_ALLOC_TYPE(
                 name="fsdp_grads",
                 fsdp_param_groups=self.parameter_groups,
                 size=UB_BUFFER_NUM,
+                dtype_attr=(
+                    self.mp_policy.grad_comm_dtype
+                    if isinstance(self.mp_policy.grad_comm_dtype, torch.dtype)
+                    else "grad_dtype"
+                ),
```
cspades: Translate: If the gradient communication data-type is set, then that is what this allocator will allocate, and the `MaxPoolAllocator` needs to know the correct `dtype` to properly plan ahead for the bucket assignments. Otherwise, just check the main gradient data-type for the ParameterGroup.

Before this change, there was no data-type argument, the `FixedPoolAllocator` just used the `dtype` to find symmetric buckets, so this doesn't change any behavior besides the case where we use a custom gradient communication data-type.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -825,6 +882,307 @@ def free(self, bucket_id: int):
...
+                    self.idle_buffer.append((buf_group_id, dtype, bucket_offset))
+
+        # Fallback allocator used if the fixed pool allocator cannot fulfill a request.
+        self.fallback_to_persistent_buffer = fallback_to_persistent_buffer
+        self.backup_allocator = StorageResizeBasedBucketAllocator()
+
+    def _build_fixed_max_pool(self):
+        """
+        Compute the maximum double-buffer pool required to support all FSDP units.
+        """
```
cspades, line 973: The max pooling algorithm is here. The rest of the code is similar to `FixedPoolAllocator`.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -151,7 +143,7 @@ def __init__(
             self.fsdp_unit_modules = fsdp_unit_modules
         else:
             if self.ddp_config.data_parallel_sharding_strategy == "optim_grads_params":
-                self.fsdp_unit_modules = [TransformerLayer]
+                self.fsdp_unit_modules = [TransformerLayer, MoETransformerLayer, MambaLayer]
```
cspades, line 162: Before, we did not shard `MambaLayer` at all.

megatron/core/full_cuda_graph.py
```diff
@@ -207,6 +207,15 @@ def __call__(self, *args, **kwargs):
...
             logger.info(f'Capture CUDA graph for {training_str}!!!')
+            if hasattr(torch.autograd.graph, 'set_override_stale_capture_stream'):
+                torch.autograd.graph.set_override_stale_capture_stream(True)
+            else:
+                logger.warning(
+                    'torch.autograd.graph.set_override_stale_capture_stream is not '
+                    'available in this PyTorch version; CUDA graph capture may fail '
+                    'if autograd nodes hold stale references to non-capturing streams. '
+                    'Upgrade to a PyTorch build that includes pytorch/pytorch#180090.'
+                )
```
cspades, line 218: This should just be something that we should call if we have a new enough PyTorch version: https://github.com/pytorch/pytorch/pull/180090 (The PyTorch version has not been published yet.)

It harmlessly makes things a lot easier w.r.t. stragglers on the Autograd / accumulate stream. cc @nanz-nv

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
shjwudp: Can we add a precondition for using a double buffer allocator to avoid this modification broadcast to all use cases?

cspades: Absolutely, will add.

cspades:
```
        if prefetch and not (
            # When double buffering, if parameters are not members of FSDP units,
            # we should skip pre-fetch to efficiently supply buffers from the pool.
            # Non-unit module pre-fetch can run inside other FSDP unit modules and
            # un-shard irrelevant model components that pointlessly steal buffer
            # allocations from the expected FSDP unit allocation and violating
            # the maximum limit of 2 buffers allocated at any point in time.
            self.buffer.ddp_config.fsdp_double_buffer
            and no_fsdp_units
        ):
```

So basically, we'll still do a "naive" pre-fetch if we are not using double buffers.

That being said, I feel like sometimes this can increase memory overhead. I think it is a trade-off. If we "naively" pre-fetch the next bucket even though the current bucket is not an FSDP unit, then it means:

- If the next layer is relevant and is computed after the current layer, then we will have better overlap and performance.
  - `LanguageModelEmbedding` -> first Transformer layer.
- If the next layer is not relevant and is not computed after the current layer, then we will un-shard some extra bucket(s) and increase the memory overhead to support the current layer, the actual next compute layer, and the next un-used extra layer.
  - `LanguageModelEmbedding` tied to `MultiTokePrediction` so we pre-fetch maybe 1 Transformer, 1 MoETransformer, and some Layer 2 Transformer.

I think the above 2 points are both somewhat common, but one thing we can do is to suggest users use double buffer for weight-tied output layer, otherwise there will be higher memory overhead at the end of the model if we do not skip this pre-fetch.

If this can be customized by the user at a per-module level, the user can decide the pre-fetch graph instead of us.

wujingyue (commented):
> Deprecates --grad-reduce-in-bf16 / reduce_grad_in_fp32 for Megatron-FSDP, which has been incredibly confusing to use. Default arguments (auto) assume BF16 for both, so will not OOM any existing user's configs.
> Adds a call to torch.autograd.graph.set_override_stale_capture_stream(True) (only supported on new PyTorch versions since https://github.com/pytorch/pytorch/pull/180090) to prevent full-iteration CG errors like this:

Thanks for the PR and the figures! 

While I'm still reviewing the rest, can these two changes go to a separate PR(s)? https://google.github.io/eng-practices/review/developer/small-cls.html

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -504,6 +504,7 @@ def allocate(
         dtype: torch.dtype,
         device: torch.device,
         mem_alloc_context: Optional[Callable] = None,
+        strict_assignments: bool = True,
```
wujingyue, line 508: > Adds the strict_assignment state to attempt to assign the same bucket previously assigned to an FSDP unit before warning the user and assigning a different bucket to the unit.

What's the downside of strict_assignment=True? I wonder if it should be always on, or always on/off for certain allocators so we have fewer knobs to worry about.

cspades:
> While I'm still reviewing the rest, can these two changes go to a separate PR(s)? https://google.github.io/eng-practices/review/developer/small-cls.html

@wujingyue Considering this exact commit needs to be merged for the NeMo release code freeze in a few days, could we make an exception in this case? These three features are all needed for Nemotron benchmarks. I'm concerned that waiting on 3 PR's to be merged in a few work days is not feasible.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: It should always be on, it falls back to the original behavior if it is unsuccessful, but tries to allocate what it has allocated before, the scope of which is defined by this boolean being set to True. So it fixes CG issues while improving the rigor of the double buffer assignment strategy in general.

In V2, we should do this with context managers or something for `TracePoolAlloc`, but here it is far too messy to implement that when we don't have a use case for multiple model call patterns.

examples/megatron_fsdp/train_llama3_8b_fsdp_h100_fp8.sh
```diff
@@ -215,6 +219,8 @@ if [ "${NSYS_PROFILE}" = 1 ]; then
         --profile-step-start 8
         --profile-step-end 12
         --profile-ranks 0
+        --record-memory-history
+        --memory-snapshot-path "${NSYS_PROFILE_PATH}/torch_memprof_node${SLURM_NODEID}_rank${SLURM_PROCID}.pickle"
```
wujingyue: Remove?

examples/megatron_fsdp/train_llama3_8b_fsdp_h100_fp8.sh
```diff
@@ -202,6 +205,7 @@ EVAL_AND_LOGGING_ARGS=(
     --eval-interval 100
     --save-interval 1000
     --log-throughput
+    --logging-level 20
```
wujingyue, line 208: Remove?

megatron/core/distributed/fsdp/src/megatron_fsdp/distributed_data_parallel_config.py
```diff
@@ -187,3 +196,7 @@ def __post_init__(self):
                     "PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True is currently not supported "
                     "with nccl_ub due to compatibility issue with torch.cuda.MemPool API."
                 )
+
+        if self.megatron_fsdp_max_pool_double_buffer:
+            # MaxPoolAllocator is a type of double-buffer allocator.
+            self.fsdp_double_buffer = True
```
wujingyue, line 202: Instead of quietly overriding fsdp_double_buffer, we may want to `assert fsdp_double_buffer` to make sure the user understands the contract.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
wujingyue: Translation looks good. Even better if it's a code comment. The comment can go either here or to the allocator's constructor.

wujingyue:
> I'm concerned that waiting on 3 PR's to be merged in a few work days is not feasible.

In my experience, reviewing three [stacked PRs](https://github.com/NVIDIA/Megatron-LM/pull/5496) is usually faster than reviewing a single large PR. Stacked PRs can also be reviewed in parallel, though I may be missing something about how the review process works in Megatron-LM.

As a less ideal alternative, you could keep everything in a single PR but split it into three well-structured commits. GitHub's UI supports reviewing commits individually, which provides a similar incremental review experience.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -825,6 +882,307 @@ def free(self, bucket_id: int):
...
+        self._build_fixed_max_pool()
+
+        # --- Fixed Pool Buffering Check ---
+        # Ensure there is at least one group of FSDP units eligible for fixed pool buffering.
+        # If not, the allocator cannot provide its intended memory recycling benefits.
+        self.fsdp_double_buffer_units = list(self.fsdp_unit_buckets.keys())
+        assert (
+            len(self.fsdp_double_buffer_units) > 0
+        ), "Found no FSDP units to use max-sized buffering."
+        if torch.distributed.get_rank() == 0:
```
wujingyue: Consider log_single_rank to reduce indentation

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -825,6 +882,307 @@ def free(self, bucket_id: int):
...
+            if any(
+                pg.fsdp_unit_id == -1 or pg.fsdp_unit_id is None for pg in self.fsdp_param_groups
```
wujingyue: It seems weird to have both -1 and None to represent the same meaning. But I guess this is likely a pre-existing problem.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -825,6 +882,307 @@ def free(self, bucket_id: int):
...
+                    if bucket_size > max_offset_size:
+                        max_bucket_sizes[bucket_offset] = bucket_size
+                        # Track which bucket IDs define the maxima.
+                        max_bucket_ids[bucket_offset] = (fsdp_unit_id, bucket_id)
+                    # Assign bucket ID to this offset for this dtype,
+                    # to recycle the appropriate buffer.
+                    self.bucket_alloc_index[bucket_id] = (-1, bucket_offset)
+
+        # Log the max pool bucket sizes and bucket IDs responsible.
+        if torch.distributed.get_rank() == 0:
```
wujingyue: Still needed given log_single_rank below?

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -825,6 +882,307 @@ def free(self, bucket_id: int):
...
+    For every parameter group / bucket, the maximum storage required across all FSDP units
+    is pre-computed to recycle buffers across different FSDP units. To efficiently allocate
+    buckets, size maxima are stratified by dtype, and FSDP unit bucket assignments are
+    sorted such that the smallest bucket is assigned to the smallest buffer in the pool.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        fsdp_param_groups: List["ParameterGroup"],
```
wujingyue, line 901: Note to myself: these parameter groups may span across FSDP units.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
wujingyue: Do max pooling decisions depend on prefetching/overlapping? Conceptually, more aggressive prefetching needs more memory and therefore affects the max pooling algorithm?

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -825,6 +882,307 @@ def free(self, bucket_id: int):
...
+        for fsdp_unit_id, fsdp_unit_bucket_ids in self.fsdp_unit_buckets.items():
+            unit_dtype_bucket_sizes = {}
+            for bucket_id in fsdp_unit_bucket_ids:
+                # Get the parameter group dtype and size.
+                pg = self.fsdp_param_groups[bucket_id]
+                num_group_elements = sum(p.numel() for p in pg.params)
+                bucket_dtype = (
+                    self.dtype_attr
+                    if isinstance(self.dtype_attr, torch.dtype)
+                    else getattr(pg, self.dtype_attr)
```
wujingyue: This pattern appears many times. 

We may consider, instead of dtype_attr, taking a `self.dtype_fn: Callable[[ParameterGroup], torch.dtype]`. This avoids parameter polymorphism and leaves the if-then-else to the user of allocators.

Alternatively, we can make a method `Allocator.get_bucket_dtype(ParameterGroup)` wrapping this.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -825,6 +882,307 @@ def free(self, bucket_id: int):
...
+                # Get maximum dtype bucket sizes.
+                if dtype == "float8":
+                    # Map to actual dtype, which is uint8.
+                    dtype = torch.uint8
+                max_bucket_sizes = self.max_dtype_bucket_sizes.setdefault(dtype, [])
+                max_bucket_ids = dtype_max_bucket_id.setdefault(dtype, [])
+                # If more buckets are needed for this unit, extend the pool with 0's.
+                if len(bucket_sizes) > len(max_bucket_sizes):
+                    extend_len = len(bucket_sizes) - len(max_bucket_sizes)
+                    max_bucket_sizes.extend([0] * extend_len)
```
wujingyue, line 1000: Isn't max_bucket_sizes already sorted so we can **prepend** 0s without having to sort max_bucket_sizes again?

wujingyue (commented):
LGTM otherwise

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
shjwudp: The "next layer" model weights being used much later is an edge case — both `LanguageModelEmbedding` and `MultiTokenPrediction` belong to `GPTModel` rather than `TransformerLayer`. Although the execution gap between these two is large, we may still need to unshard them simultaneously.

At the moment, prefetch can only handle the more general case and does not offer fine-grained control over this specific scenario.

shjwudp (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: Max-pooling is a hard-requirement on memory efficiency unless we have a more sophisticated FSDP scheduling algorithm to pre-compute all possible paths for this set of asymmetric model layers.

Once you consider the max pool the representative size of the unit in this model, pre-fetch sizing can be controlled using `suggested_communication_unit_size` which controls how many max pool buckets we pre-fetch. It defaults to 500M or 1B `numel`.

Finally, if you have performance requirements where the size of a max pool bucket lands you in an awkward position with the required communication size, and you don't want to increase the communication size, then the last thing you can do is to use fine-grained AG or fine-grained RS to allow for more checkpoints where we permit an AG or RS to be launched. This will allow multiple episodes of AG or RS to be called within a single FSDP unit.

examples/megatron_fsdp/train_llama3_8b_fsdp_h100_fp8.sh
cspades: Intended, I use this all the time that I think it should be documented in our example script, but I should comment this out since it does damage performance a little bit.

examples/megatron_fsdp/train_llama3_8b_fsdp_h100_fp8.sh
cspades: This should have always been here. Without it, FSDP unit assignments don't print out when using Megatron-FSDP, and those logs are extremely useful / important. Moreover, zero impact on performance.

megatron/core/distributed/fsdp/src/megatron_fsdp/distributed_data_parallel_config.py
cspades: I prefer if the user does not need to turn on 2 arguments just to activate the max pooling double buffer. I've been told that MLM CLI should be simplified, not made more complex.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: I completely missed this, I purposefully changed it to log_single_rank but forgot to remove the if condition, thanks!

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: I'm deprecating `-1`. I did a search on `fsdp_unit_id` as well as `-1` in the Megatron-FSDP codebase and found nothing. It must have been removed in one of the past PR's refactoring `_get_parameter_groups` or something. cc @shjwudp who has been refactoring this part.

cspades: It's not in git history either. Must have been an artifact of nvFSDP or even earlier. Most likely very safe to deprecate.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: Well, we have already assigned the previous bucket ID's and I'm using the enumerated index of this list as a bucket offset. If I prepend, it will shift all the buckets to the right by one relative to their bucket offset, and break this algorithm.
```
sorted(enumerate(max_bucket_sizes), key=lambda x: x[1])
```
We can avoid this by reversing the `zip`, adding the new buckets to the end of the pool but getting the largest N buckets from the top of the pool and assigning them to the largest N buckets of the unit (so also `bucket_sizes.sort() -> bucket_sizes.sort(reverse=True)`. I think that should preserve a reversed sorting order.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: Sure, done.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: Discussion continued in Slack, we'll think about it for V2.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
cspades: To add, they are all parameter groups across the entire model. Including non-unit module parameters `fsdp_unit_id=None`.

yashaswikarnati (approved):

deepakn94 (approved):

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -117,17 +117,8 @@ def __init__(
...
-                torch.float32
-                if ddp_config.grad_reduce_in_fp32
-                else ddp_config.megatron_fsdp_main_grads_dtype
-            ),
-            grad_comm_dtype=(
-                torch.float32
-                if ddp_config.grad_reduce_in_fp32
-                else ddp_config.megatron_fsdp_grad_comm_dtype
-            ),
+            main_grads_dtype=ddp_config.megatron_fsdp_main_grads_dtype,
```
ericharper, line 136: does this break bridge? https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/d12548a3cf7a72e0b2f38cd67da5598624abb3fd/src/megatron/bridge/recipes/qwen_vl/qwen35_vl.py#L131-L141

ericharper: @gautham-kollu , can you update bridge?

FYI @yaoyu-33 and @cuichenx

ericharper (approved):
Approve but bridge needs to be updated.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
cspades: The migration action items are:
```
--megatron-fsdp-main-grads-dtype fp32 / ddp.megatron_fsdp_main_grads_dtype=torch.float32
--megatron-fsdp-grad-comm-dtype fp32 / ddp.megatron_fsdp_grad_comm_dtype=torch.float32
```
for any recipe that uses `grad_reduce_in_fp32=True` (i.e. does not use `--grad-reduce-in-bf16`).

For completeness, if `--grad-reduce-in-bf16 / grad_reduce_in_fp32=False`, then the default `megatron_fsdp_grad_comm_dtype` and `megatron_fsdp_main_grads_dtype` are both BF16 so that's also aligned with turning that argument on and does not need any changes. (This is the logical spaghetti I was talking about, two levels of arguments.)

cc @gautham-kollu if you can hit this in your next benchmark update. 🙏🏻 IMO low-ish priority because this will not OOM anyone's script.

PR #4769, opened 2026-05-13
Fix mxfp8 param gather numerical issue when DP overlap is off

WanZzzzzz:
# What does this PR do ?

main: https://github.com/NVIDIA/Megatron-LM/pull/4800

 With MXFP8 grad-buffer reuse and non-overlap param gather, each DistOpt stages its own updated main-param shards into its param buffers during step. However, param sync is a DDP model-chunk operation: model_chunk.start_param_sync() gathers both dense and expert bucket groups, copies gathered values into model weights, and zeros the shared MXFP8 param/grad buffers. For MoE, dense and expert DistOpts may share the same model chunk, so defer param sync until all chained optimizers have staged their params, then sync each model chunk once. Otherwise the expert DistOpt may copy the zeroes into expert model weights.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

## Issue tracking

For PRs from open-source community contributors:

- **New features**: a linked issue is **required**. Please open a [feature request](https://github.com/NVIDIA/Megatron-LM/issues/new?template=feature_request.md) and reference it here before submitting the PR.
- **Small updates (bug fixes, minor improvements)**: a linked issue is **recommended** and will accelerate the PR review process.

Linked issue: 

### Pre-checks

- [x] I have added relevant unit tests
- [x] I have added relevant functional tests
- [x] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [x] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

[...]

PR #5470, opened 2026-06-24
[dev] Muon fp8 param gather for decoupled layout

Wohox:

# What does this PR do ?

Adds **FP8 parameter gather to the Muon optimizer** on the decoupled compact LayerWise DDP layout
(`--fp8-param-gather`), so Muon persists and syncs its matrix
parameters as FP8 instead of BF16 — bringing its memory footprint down to Adam's level while staying
bit-for-bit equal to the fp8-param-gather-OFF run.

Builds on #5388 (the decoupled compact LayerWise layout). Previously the non-DistOpt Muon path could
only hold and all-gather its parameters in BF16. With this change Muon's matrix params are persisted
as FP8 (mxfp8 on Blackwell, blockwise on Hopper), removing the standing BF16 parameter copy; combined
with the no-padding layout from #5388 this closes the Muon-vs-Adam memory gap.

### What's new

**1. An FP8-aware whole-parameter all-gather for the non-DistOpt Muon path.** Muon's matrix buffers
own whole params via ping-pong assignment and sync through `allgather_params` (not DistOpt's
byte-shard reduce-scatter/all-gather). FP8 tensors can't be all-gathered directly — mxfp8's row/col
block scales can't be derived from one another — so the transport rides BF16 and requantizes on
landing: stage each owned param from its **fp32 master → bf16** (`_stage_param_to_bf16`, a
high-precision source, not a lossy fp8 dequant) → **uneven all-gather-v** of the bf16 whole-params →
**requantize** the gathered bf16 into *every* rank's fp8 `param.data`
(`copy_back_gathered_bf16_into_fp8_param`) so all ranks hold `Q(bf16(master))` (== OFF / Adam) →
`post_all_gather_processing` rebuilds the fp8 columnwise/transpose. Wired into **both** the overlap
path (`start_param_sync`/`finish_param_sync`) and the non-overlap path (`allgather_params`); plain
bf16 (fp8pg OFF) collapses back to the original code.

**2. `use_distributed_optimizer` becomes a per-buffer property.** On the decoupled layout,
LayerWise-managed (Muon 2D-matrix) buffers locally disable DistOpt — compact no-padding layout,
all-reduce gradients, whole-param ping-pong ownership — while sibling buffers (embeddings, biases,
layernorm) keep the standard byte-level `DistributedOptimizer` path.

**3. Rank-independent ping-pong ownership.** `allgather_params` requires every DP rank to agree on
each param's single owner. `numel` alone isn't a total order (a stable sort tie-breaks equal-numel
params by insertion order → different owners per rank), so ownership is now keyed by a canonical
identity `(chunk_idx, buffer_idx, global_start_index)` from model construction, with an assert that
every Muon param has one.

**4. iter-0 master parity.** The fp32 master is seeded from each param's high-precision
pre-quantization init (`get_high_precision_init_val`), not the lossy fp8 dequant, and the
master→model copy routes through bf16 (`Q(bf16(master))`) — so fp8pg ON and OFF hold an identical
master at iter 0 (matching DistOpt). Also covers MoE expert weights at `expt_dp==1`, which aren't
gathered.

**5. Validation guards** (`arguments.py`): FP8 param gather is supported only on the decoupled layout
and is rejected on the padded LayerWise layout. On decouple it requires `fp8_recipe ∈ {mxfp8,
blockwise}` (fp4 out of scope); mxfp8 additionally requires `--reuse-grad-buf-for-mxfp8-param-ag`.

**6. Single all-reduce buffer (a correctness fix this combination exposed).** With fp8pg ON,
LayerWise FP8 grads were keyed by `torch.uint8` in `group_params_for_buffers`, splitting Muon's
gradients into two all-reduce buffers (uint8 + bf16) where OFF has one. Both reduce in fp32, but
NCCL's fp32 accumulation order is buffer-size sensitive, so the two-buffer order diverged ~1 ULP from
OFF on the DP-reduced dense FP8 2D weights — amplified by Muon's Newton–Schulz into a visible loss
divergence. Fix: a new `merge_layerwise_fp8_grads` flag (decouple-only) keys FP8 Muon grads by their
bf16 logical dtype so FP8 + bf16 grads share one fp32 all-reduce buffer, matching OFF. Two
all-reduces collapse to one (no extra communication); NVFP4's packed param buffer is untouched.

### Correctness — bitwise ON == OFF

Deterministic, `clip_grad=0`. Compared per-iter `lm loss`, forward output, per-param reduced
`main_grad`, fp32 master, and bf16 model param.

| setup | result |
|---|---|
| H100, DP8, blockwise, deepseek_v3_proxy, 20 it | **bitwise identical, all 20 iters** |
| GB200, DP4, mxfp8, deepseek_v3_proxy, 20 it | **bitwise identical, all 20 iters** |

(Before the single-buffer fix, loss diverged at it12 on H100 / it3 on GB200; per-param pre-reduce
grads were already identical, isolating the divergence to the DP all-reduce.) With `clip_grad=0`,
Adam (DistOpt byte-shard, not split) was already bitwise. `clip_grad>0` adds a separate
layout-dependent global grad-norm fp32 sum term, so the `clip_grad=1.0` benchmarking recipes are not
bitwise — bit-exactness is established at `clip_grad=0`.

### Memory & throughput

**GB200 DP64** (mxfp8, deepseek_v4_pro_proxy 8L, TP1/PP1/EP64, SL4K):

| arm | peak alloc |
|---|---|
| Muon decouple **+ fp8pg** | **105.8 GB** |
| Muon decouple, no fp8pg | 116.6 GB |
| Muon **padded** (legacy), no fp8pg | 142.9 GB |
| Adam (DistOpt) + fp8pg | 106.1 GB |
| Adam (DistOpt), no fp8pg | 116.9 GB |

FP8 param persistence −10.8 GB (identical on Muon and Adam); compact-layout padding removal −26.2 GB;
decouple+fp8pg vs padded Muon −37 GB (~26%), and ~11 GB below Adam-no-fp8pg. **H100 dev proxy** (DP8,
blockwise) reproduces the mechanism: fp8pg −1.1 GB (35,460 vs 36,594 MB); padding `0.0%` vs `0.6%`.
**Refactor integrity:** Muon padded / fp8pg OFF vs the pre-refactor DSv4+padded-Muon baseline is
142,858 vs 142,857 MB (1 MB) — the padded path is left equivalent.

**Throughput** (GB200 wall-clock, comparable within branch): Adam < Muon decouple < Muon padded. Muon
is ~5–6% slower than Adam (Newton–Schulz + uneven all-gather-v); fp8 param-gather itself adds ~1%;
the compact layout is ~4.8% faster than padded.

## Issue tracking

Linked issue: Builds on #5388

### Pre-checks

Unit test: `tests/unit_tests/test_muon_decouple_fp8_param_gather.py` — bitwise ON-vs-OFF check on the
decoupled layout (`muon` + `use_distributed_optimizer` → `LayerWiseDistributedOptimizer`), asserting
per-step loss / forward output / `main_grad` / fp32 master / bf16 param are bitwise-identical under
`deterministic_mode`. Parametrized over `overlap_grad_reduce` + `overlap_param_gather` both ON and
OFF; blockwise runs on Hopper, mxfp8 skips off Blackwell. Verified on H100: 2 passed (blockwise ×
overlap{off,on}), 2 skipped (mxfp8).

### Code review

Feel free to message or comment @NVIDIA/mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged\!

#### Step 2: Final Review

#### Step 3: Approved

### Merge

---

Main-branch counterpart: https://github.com/NVIDIA/Megatron-LM/pull/5479


Wohox:
/ok to test [f75afdb](https://github.com/NVIDIA/Megatron-LM/pull/5470/commits/f75afdba72552476beeafc180737a1d078c727d8)

megatron/core/optimizer/optimizer.py
```diff
@@ -924,12 +947,43 @@ def _copy_model_grads_to_main_grads(self):
...
     def _copy_main_params_to_model_params(self):
+        # Non-DistOpt LayerWise fp8: route master->model through bf16 (Q(bf16(master))) to match the
+        # fp8-param-gather-OFF baseline (a direct fp32->fp8 copy would write Q(fp32 master)). This
+        # also covers MoE expert weights at expt_dp==1, which are not gathered.
+        if self._layer_wise_non_distopt_child:
+            other_model_data, other_main_data = [], []
+            for model_group, main_group in zip(self.float16_groups, self.fp32_from_float16_groups):
+                for model_param, main_param in zip(model_group, main_group):
+                    if is_float8tensor(model_param):
+                        copy_back_gathered_bf16_into_fp8_param(
```
FDecaYed: for the optimizer itself, this copy is correct. But in case where later a bf16 copy is created and gathered, this copy will be overwritten and is not necessary, right?

Wohox: You're right — for the params that go through the all-gather, this copy is redundant. The fp8 `param.data` written here is never read by the gather (`_stage_param_to_bf16` stages from `main_param`, the fp32 master, not from `param.data`), and `_allgather_helper_fp8` then requantizes the gathered bf16 back into every rank's `param.data` anyway — writing the same `Q(bf16(master))` a second time and overwriting this one.

The reason this branch stays is the non-gathered case: MoE expert weights at `expt_dp == 1` (where `expt_dp_params_list` is `None`, so the all-gather skips them). For those, this is the only place they get the correct `Q(bf16(master))` written back, so the copy is necessary there.

So the two options are:
1. Keep it as-is — correctness is fine; the only cost is one extra requantize per step on the gathered params.
2. Narrow this copy to non-gathered params only (skip params owned by a buffer that will be all-gathered), avoiding the redundant work at the cost of a bit more branching here.

I'm inclined to keep (1) for simplicity since the extra cost is a single in-place requantize on the owned shard, but happy to do (2) if you'd prefer to avoid the redundant copy.

Wohox: Fixed in f45783085a27b0b623bbe9acd61125a12fa4da61, only the non-dp sharded params go through the cast here.

megatron/core/optimizer/optimizer.py
```diff
@@ -823,8 +834,20 @@ def __init__(
...
+                            # fp8_param_gather ON/OFF hold an identical master at iter 0.
+                            if hasattr(param, 'get_high_precision_init_val'):
+                                main_param = (
+                                    param.get_high_precision_init_val()
+                                    .detach()
+                                    .clone()
+                                    .to(param.device)
+                                    .float()
+                                )
+                                param.clear_high_precision_init_val()
```
kunlunl, line 847: Does this `clear high_precision_init_val` for all parameters, including the non-owned parameters that were removed from the local optimizer groups by `shard_params()`?

megatron/training/arguments.py
```diff
@@ -1894,19 +1894,39 @@ def validate_args(args, defaults={}):
...
-            "Layer-wise (Muon) distributed optimizer does not support FP8/FP4 parameter gather "
-            "(fp8_param_gather / fp4_param_gather). Use fp8_param_gather=False (e.g. blockwise/"
-            "MXFP8 compute with parameters persisted in bf16)."
-        )
         if not args.use_layer_wise_param_layout:
+            # Decoupled compact LayerWise: fp8 parameter gather is supported via the FP8-aware
+            # whole-param all-gather. Only mxfp8/blockwise (fp4 out of scope); mxfp8 needs
+            # reuse_grad_buf.
+            if args.fp8_param_gather:
+                assert not getattr(args, 'fp4_param_gather', False), (
```
kunlunl: Should `fp4_param_gather` be rejected unconditionally in the compact-layout branch? Currently this assertion only runs when `fp8_param_gather` is also enabled, so using `--fp4-param-gather` alone passes validation even though FP4 is documented as out of scope here.

megatron/core/distributed/param_and_grad_buffer.py
kunlunl, line 456: Should the `force_sync` case also perform the LayerWise copy-back and cleanup before returning?

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1577,6 +1648,10 @@ def partition_buckets(
          has completed. This is because we need to wait for the non-fp8 params from the beginning
          layers to obtain their gradients.
        - Combining the non-fp8 bucket with the last fp8 bucket can help avoid this issue.
+       - A bucket group runs one collective type, so only buckets agreeing on the effective
```
kunlunl, line 1764: Could splitting the LayerWise all-reduce buckets from the DistOpt reduce-scatter buckets reintroduce the back-to-back NCCL launches that the original FP8/non-FP8 bucket merging was intended to avoid, especially with CUDA_DEVICE_MAX_CONNECTIONS=1?

tests/unit_tests/test_muon_decouple_fp8_param_gather.py
```diff
@@ -0,0 +1,287 @@
...
+    @pytest.mark.skipif(not is_te_min_version("2.3.0.dev0"), reason="TE 2.3.0.dev0 is required")
+    def test_on_vs_off_bitwise_identical(self, fp8_recipe, overlap):
+        """fp8_param_gather ON must match OFF bitwise on the decoupled layout, for
+        overlap grad-reduce + param-gather both ON and OFF."""
+        arch = get_device_arch_version()
+        if fp8_recipe == "blockwise" and arch != 9:
+            pytest.skip("blockwise FP8 is Hopper-only")
+        if fp8_recipe == "mxfp8" and arch < 10:
+            pytest.skip("mxfp8 requires Blackwell architecture or newer")
+        n = 5
```
kunlunl: Maybe add this number to 30 or bigger?
We often observed mismatch after many iterations.

tests/unit_tests/test_muon_decouple_fp8_param_gather.py
```diff
@@ -0,0 +1,287 @@
...
+        args.num_attention_heads = 4
+        args.max_position_embeddings = self.seq_length
+        args.seq_length = self.seq_length
+        args.micro_batch_size = self.micro_batch_size
+        args.create_attention_mask_in_dataloader = True
+        args.tensor_model_parallel_size = 1
+        args.pipeline_model_parallel_size = 1
+        args.context_parallel_size = 1
+        args.train_iters = 10
+        args.lr = 3e-5
```
kunlunl: Maybe increase `lr` so the mismatch (if exists) will be more obvious.

tests/unit_tests/test_muon_decouple_fp8_param_gather.py
```diff
@@ -0,0 +1,287 @@
...
+class TestMuonDecoupleFP8ParamGather:
+
+    def setup_method(self, method):
+        self.seq_length = 128
+        self.micro_batch_size = 1
+        os.environ['CUDA_DEVICE_MAX_CONNECTIONS'] = '1'
+        Utils.initialize_model_parallel(
+            tensor_model_parallel_size=1,
+            pipeline_model_parallel_size=1,
+            expert_model_parallel_size=1,
```
kunlunl, line 123: Could we add a MoE variant of this test, ideally covering both overlap on/off and expt_dp == 1 / expt_dp > 1? PRs #4769/#4800 fixed a related MXFP8 grad-buffer-reuse bug where dense and expert DistOpts shared a model chunk: one optimizer’s param sync gathered and zeroed both reused buffers before the other optimizer had staged its updated parameters, which could copy zeroes into the model weights. Since this PR changes the per-buffer DistOpt ownership, param-sync grouping, and LayerWise buffer cleanup paths, a Muon + MoE regression test would help ensure that the same buffer-lifecycle issue is not reintroduced.

megatron/core/optimizer/optimizer.py
Wohox: Good catch! Currently other DP rank's high-precision param are leaked, but distributed optimizer has the same issue, will fix layerwise here.

kunlunl: `dist-opt` is slightly different. `dist-opt` will call the partial cast method of the TE. After that method is called for the first time, it will call `clear_high_precision_xxx` to clean up the CPU tensor. In your case, that function of the TE was not called.

megatron/core/distributed/param_and_grad_buffer.py
Wohox: I think so, but this is hard to avoid, since AR and RS cannot be merged, same for AG and AG-v.

megatron/training/arguments.py
Wohox: Thanks for catching this up, added assertion for fp4-param-gather as well.

tests/unit_tests/test_muon_decouple_fp8_param_gather.py
Wohox: Done.

megatron/core/distributed/param_and_grad_buffer.py
Wohox: Fixed.

kunlunl (approved):

kunlunl:
/ok to test [43cfb67](https://github.com/NVIDIA/Megatron-LM/pull/5470/commits/43cfb67dec2c5b582415f5d266af8316dec18981)

Wohox:
/ok to test [4b3190b](https://github.com/NVIDIA/Megatron-LM/pull/5470/commits/4b3190b7803bcea307dc0e9534ffc5c171675ea0)

FDecaYed (approved):
lgtm

PR #5471, opened 2026-06-24
[dev]: Fix mHC boundaries in EP overlap schedule

jingqiny-99:

# What does this PR do ?
  ## What changed

  - Fix mHC boundary and recompute handling with EP overlap.
  - Keep mHC post-processing in the combine node by default.
  - Add `--mhc-post-on-compute-stream` to optionally schedule mHC post as a standalone compute-stream node.
  - Preserve PP, MTP, and FSDP compatibility.

mHC post and mHC recompute have no dedicated stream; they reuse the two EP-overlap streams: comp (get_comp_stream()) and comm (EP all-to-all / PP p2p, get_comm_stream()). All per-layer 1F1B scheduling is in TransformerLayerSchedulePlan.run (model_chunk_schedule_plan.py:264-342); combined_1f1b.py is unchanged vs dev.

mHC post — two modes, gated by mhc_post_on_compute_stream:

Default (False): post is fused inside the moe_combine node on the comm stream (the mhc_post node is a NoopScheduleNode). Forward runs in combine_fwd, backward in combine_bwd. (fine_grained_callables.py:703,725)

Compute-stream (True): post is a standalone node on the comp stream — post.F after combine_fwd (run:331), post.B before combine_bwd (run:297).

mHC recompute — only when recompute_granularity="selective" and "mhc" in recompute_modules. It is a ScheduleNode on the comp stream, created only on the last layer of each recompute group (model_chunk_schedule_plan.py:186-199). In backward it runs first for that layer, immediately before the combine/post backward (run:295-296), replaying the whole group's forward. Since backward walks layers in descending order, the group's last layer runs first, so the replay precedes the group's per-layer backward. NVTX: mhc/recompute/group_{i}/B.
image

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


megatron/core/models/gpt/fine_grained_callables.py
```diff
@@ -666,30 +738,90 @@ def submodule_combine_forward(node: ScheduleNode, output: torch.Tensor):
...
+        )
+
+        mlp_norm_manager = getattr(node.layer_state, 'mlp_norm_manager', None)
+        if mlp_norm_manager is not None:
+            hidden_states = mlp_norm_manager.group_offload(
+                hidden_states, forced_released_tensors=[residual]
+            )
+            node.layer_state.mlp_norm_manager = None
+
+        if not node.is_mtp and node.is_last_layer:
```
Victarry: **[CRITICAL Correctness]** Decoder boundary finalization is currently reachable only from the MoE `combine` / `mhc_post` callables. For a supported mixed pattern such as `moe_layer_freq=[1, 0]`, the final dense layer has no real combine or mHC-post node, so its `[s, b, n*h]` output can reach GPT postprocessing without `learned_output_contract` or the final layer norm.

Please factor this into one shared helper, for example `finalize_decoder_layer_output(node, hidden_states)`, and attach it to whichever operation is terminal for the layer (`mhc_post`, `combine`, or the dense `mlp`) instead of embedding the boundary logic only in the MoE closures. This keeps the math independent of placement and covers dense/MoE and merged/separate-post paths consistently. A regression test with mHC + EP overlap and a final dense layer (for example `moe_layer_freq=[1, 0]`) would verify the boundary and output shape.

megatron/core/models/gpt/fine_grained_callables.py
```diff
@@ -666,30 +738,90 @@ def submodule_combine_forward(node: ScheduleNode, output: torch.Tensor):
...
+            output, mhc_multistream = node.chunk_state.model.decoder.postprocess_for_layer_schedule(
+                hidden_states, return_mhc_multistream=True
+            )
+            output = make_viewless_tensor(
+                inp=output, requires_grad=output.requires_grad, keep_graph=True
+            )
+            node.chunk_state.mhc_multistream = mhc_multistream
```
Victarry: **[CRITICAL Correctness]** `mhc_multistream` currently bypasses the autograd boundary established by the fine-grained schedule.

Each `ScheduleNode` detaches its formal inputs before executing its local forward graph. During backward, the scheduler runs one node at a time and explicitly passes the resulting input gradients to the preceding node. This is what lets the scheduler control backward ordering, CUDA-stream dependencies, mHC recomputation, and FSDP parameter lifetime instead of allowing one `autograd.backward()` call to traverse the whole layer.

Here the decoder boundary has two differentiable outputs from the same graph: the contracted `[s, b, h]` tensor returned as the formal node output, and the pre-contraction `[s, b, n*h]` `mhc_multistream` needed by MTP. Only the contracted tensor crosses the normal `ScheduleNode` detach boundary. `mhc_multistream` is stored in `chunk_state` without `node.detach()`, and the MTP attention callable later reads that original tensor directly. Therefore MTP backward can leave its local schedule node and traverse into the decoder mHC graph *before* the scheduler reaches that decoder node's `mhc_post.backward()`. The later scheduled backward then re-enters the same graph from the contracted branch, which can produce a second-backward error after saved tensors have been freed, or at minimum bypass the intended point where the contracted and MTP gradient branches should be merged.

The existing handling of `residual`, `mlp_h_res`, and `mlp_hc_h_post` shows the required pattern: detach the side output at its producer and let `TransformerLayerNode.backward_impl` reconnect its accumulated detached gradient when the producer node is scheduled for backward. Please apply the same bridge here, for example:

```python
node.chunk_state.mhc_multistream = (
    node.detach(mhc_multistream) if mhc_multistream is not None else None
)
```

Please also add an **mHC + EP overlap numerical test with MTP** (`mtp_num_layers=1`, default `mtp_detach_heads=False`) that compares eager and scheduled forward outputs and parameter gradients for both merged and compute-stream mHC-post placements. Covering two in-flight schedule plans would additionally verify that the MTP side gradient stays attached to the correct microbatch.

megatron/core/transformer/transformer_config.py
```diff
@@ -2132,13 +2134,26 @@ def __post_init__(self):
...
+            and self.recompute_granularity == "selective"
+            and "mhc" in self.recompute_modules
+            and (
+                self.cuda_graph_impl != "none" or self.enable_cuda_graph or self.external_cuda_graph
+            )
+        ):
+            raise ValueError(
+                "mHC recompute with overlap_moe_expert_parallel_comm requires CUDA graphs "
+                "to be disabled because explicit group replay is eager-only."
+            )
```
dingqingy-nv, line 2228: mhc recompute + EP overlap + full iter cg will be an important use case for DSV4 Pro. Could you share the plan? Is it outside the scope of this PR and will be a follow-up? thanks

megatron/core/models/gpt/fine_grained_callables.py
dingqingy-nv: Under hyper-connections, `MultiTokenPredictionLayer` sets `eh_proj=None `and uses separate `e_proj`/`h_proj`. Both branches here therefore add None to the delayed-wgrad callable list. Please select `[e_proj, h_proj]` when mHC is enabled, retain `[eh_proj]` otherwise, and add a test that executes the delayed-wgrad backward path.

e.g. something like
```python
mtp_projs = (
    [layer.e_proj, layer.h_proj]
    if layer.config.enable_hyper_connections
    else [layer.eh_proj]
)
if isinstance(backward_dw["attn"], list):
    backward_dw["attn"].extend(mtp_projs)
else:
    backward_dw["attn"] = [backward_dw["attn"], *mtp_projs]
```

dingqingy-nv:
A question on **DSv4 CSA delayed-wgrad compatibility:** the EP-overlap scheduler reaches `DSv4HybridSelfAttention.backward_dw()`, but that method visits only the KV, Q, and output projections—not `self.core_attention`. For CSA, this skips the compressor/indexer projections, potentially leaving their wgrads uncomputed and delayed state retained. This seems pre-existing and may be outside #5471. Is DSv4 CSA + delayed wgrad intended to be supported here? If not, can we document/guard it and handle the recursive CSA `backward_dw()` traversal in a follow-up PR?

tests/unit_tests/a2a_overlap/test_mhc_schedule.py
```diff
@@ -0,0 +1,501 @@
...
+    def test_mtp_schedule_matches_eager(self):
+        # With MTP (mtp_num_layers=1, default mtp_detach_heads=False) the decoder boundary
+        # produces the pre-contraction mHC multi-stream consumed by the MTP depths. It must be
+        # detached at its producer so MTP backward does not traverse the decoder mHC graph out
+        # of schedule order; this node's backward_impl reconnects the accumulated gradient.
+        mtp = {"mtp_num_layers": 1}
+        reference_config = _make_mhc_numerical_config(
+            overlap=False, recompute=False, extra_config=mtp
+        )
+        overlap_config = _make_mhc_numerical_config(recompute=False, extra_config=mtp)
```
dingqingy-nv: Could our test be parameterized with `recompute=False/True` instead of forcing `recompute=False`? the previous recompute test doesn't have MTP. It'd be nice to exercise a combination of "mHC + MTP + EP overlap + recompute"

dingqingy-nv:
BTW, why is `--mhc-post-on-compute-stream` removed in  66e6d5c3ab25c22b8a07a4232c3d585f1fc87c14?

dingqingy-nv (changes requested):
Thanks for great effort! I want to flag 2 things during my e2e dsv4 tests and discuss if it belongs to scope of this PR or we can create follow-up PRs later:

### full-iter cg w/ mhc recompute and ep overlap
- full iter cg and ep overlap are both great for perf, but adds more memory, which is extremely challenging for Full Pro model. So it is critical that memory saving techniques are compatible.

### delayed-wgrad compatibility
- Delayed wgrad is typically used together with EP overlap, and this path seems less tested in this PR. I commented concern for mHC+MTP and CSA delayed-wgrad traversal.

jingqiny-99:
> BTW, why is `--mhc-post-on-compute-stream` removed in [66e6d5c](https://github.com/NVIDIA/Megatron-LM/commit/66e6d5c3ab25c22b8a07a4232c3d585f1fc87c14)?

We have tested this flag and found that it provides no perf benefits.

megatron/core/models/gpt/fine_grained_callables.py
jingqiny-99: Fixed as suggested, added test_mtp_schedule_with_delayed_wgrad_matches_eager

tests/unit_tests/a2a_overlap/test_mhc_schedule.py
jingqiny-99: Sure, both MTP tests (test_mtp_schedule_matches_eager, test_mtp_two_inflight_plans_match_eager) are now parameterized over recompute=[False, True]

megatron/core/transformer/transformer_config.py
jingqiny-99: I am working on a follow-up PR (5841) to add support for partial CG and full iter CG with mHC recompute. The implementation is complete, and it is currently being tested.

dingqingy-nv: 5841 says partial CG but not full cg?

tests/unit_tests/a2a_overlap/test_mhc_schedule.py
```diff
@@ -0,0 +1,642 @@
...
+            reset_model(overlap_model, initial_parameters)
+            overlap_output, overlap_gradients = _run_schedule_and_capture(overlap_model, data)
+
+        torch.testing.assert_close(overlap_output, reference_output, rtol=5e-3, atol=5e-3)
+        _assert_close_grads(overlap_gradients, reference_gradients)
+
+    @pytest.mark.skipif(
+        not is_te_min_version("2.3.0"), reason="delay_wgrad_compute requires TE >= 2.3.0"
+    )
+    def test_mtp_schedule_with_delayed_wgrad_matches_eager(self):
```
dingqingy-nv, line 475: This test validates delayed wgrad for the generic GPT/MLA path, but does not instantiate DSv4 hybrid attention/CSA. In that path, `DSv4HybridSelfAttention.backward_dw()` does not traverse `core_attention`, so CSA compressor/indexer wgrads may remain unflushed. 

Let's add a test for that, and I’ll discuss this separately in more detail in a comment.

dingqingy-nv (changes requested):

dingqingy-nv:
### Delayed wgrad does not flush all DSv4 CSA projection weights.
`DSv4HybridSelfAttention.backward_dw()` currently flushes the Q, KV, and output projections, but it never traverses `self.core_attention`. For CSA layers, CompressedSparseAttention owns the outer compressor and optional indexer, which contain `linear_wkv`, `linear_wgate`, `linear_wq_b`, and `linear_weights_proj`. With `delay_wgrad_compute=True`, their weight gradients remain deferred and are never flushed.

Could we add `backward_dw()` traversal to `Compressor`, `CSAIndexer`, and `CompressedSparseAttention`, then call `self.core_attention.backward_dw()`?

e.g.
```python
class Compressor:
    def backward_dw(self):
        self.linear_wgate.backward_dw()
        self.linear_wkv.backward_dw()


class CSAIndexer:
    def backward_dw(self):
        self.linear_weights_proj.backward_dw()
        self.compressor.backward_dw()
        self.linear_wq_b.backward_dw()


class CompressedSparseAttention:
    def backward_dw(self):
        if self.indexer is not None:
            self.indexer.backward_dw()
        if self.compressor is not None:
            self.compressor.backward_dw()
```

Then in `deepseek_v4_hybrid_attention.py` flush `self.core_attention`:

```python
def backward_dw(self):
    self._backward_kv_proj()
    self._backward_q_proj()
    self.core_attention.backward_dw()
    self._backward_output_proj()
```

megatron/core/transformer/transformer_config.py
jingqiny-99: I'm working to support full cg in 5841, but it's not implemented yet. I think it can be supported in next week if everything works fine.

tests/unit_tests/a2a_overlap/test_mhc_schedule.py
jingqiny-99: Ok, I'll add the test as you have commented soon.

megatron/core/transformer/transformer_config.py
dingqingy-nv: Great, thanks!

dingqingy-nv (approved):

Victarry (approved):

PR #5391, opened 2026-06-17
[main] Add experimental decoupled compact LayerWise DDP layout for Muon

Wohox:
- [x] I, the PR author, have personally reviewed every line of this PR.

# What does this PR do ?

> **`main` counterpart of #5388** (which targets `dev`). Same change, re-implemented on top of `main`. See #5388 for the `dev`-branch version.

Add an experimental **compact, decoupled LayerWise DDP buffer layout** for the Muon (layer-wise) distributed optimizer that removes the persistent `dp_size * max(shard_load)` padding from the long-lived param/grad buffers.

There is **no new flag**: this reuses the existing `--no-use-layer-wise-param-layout`. With a Muon layer-wise distributed optimizer under `--use-distributed-optimizer`, disabling the shard-aligned padded LayerWise layout (`use_layer_wise_param_layout=False`) selects the compact decoupled path. `use_layer_wise_param_layout` is mirrored onto both `DistributedDataParallelConfig` and `OptimizerConfig` (auto-populated by field name, default `True` = padded layout), so the configs read it directly — no derived switch.

In this mode `use_distributed_optimizer` becomes a **per-buffer** property:

- **LayerWise-managed (Muon 2D matrix) buffers** use a compact no-padding DDP layout and locally disable `DistributedOptimizer` semantics: all-reduce gradients, legacy whole-param ping-pong ownership, and `allgather_params` param sync.
- **Sibling non-LayerWise buffers** (embeddings, biases, layernorm) keep the standard byte-level `DistributedOptimizer` layout.

The effective flag is computed per `_ParamAndGradBuffer` / `_ParamAndGradBucketGroup`. `partition_buckets` splits a force-single bucket group (disable_bucketing / non-first VPP chunks) by the effective per-bucket `use_distributed_optimizer`, so Muon (all-reduce) and sibling (reduce-scatter) buckets never share a group; when all buckets agree this collapses to a single group, identical to the prior behavior.

`get_model` and `wrap_model_chunks_with_ddp` share a single `if use_layer_wise_distributed_optimizer:` branch — both the padded and compact cases force `ddp_config.use_distributed_optimizer=True`, tag params for buffer routing, and compute the LayerWise `full_param_layout`. The padded-vs-compact decision lives entirely in `compute_full_param_layout` / `_ParamAndGradBuffer`, which read `ddp_config.use_layer_wise_param_layout`.

**Compatibility**

- The **default padded LayerWise layout is unchanged** (`use_layer_wise_param_layout=True`).
[...]

PR #5479, opened 2026-06-24
[main] Muon fp8 param gather for decoupled layout

Wohox:

# What does this PR do ?

Adds the **compact decoupled LayerWise DDP layout** for the Muon optimizer, plus **FP8 parameter
gather on that layout** (`--fp8-param-gather`), so Muon drops the persistent
`dp_size * max(shard_load)` padding *and* persists its matrix parameters as FP8 instead of BF16 —
bringing its memory footprint below Adam's while staying bit-for-bit equal to the
fp8-param-gather-OFF run.

Main-branch counterpart of the `dev` stack #5388 + #5470 (both merged).
Compact-layout checkpoint support is intentionally out of scope for this PR.

Self-contained: it supersedes #5391, which carried the layout half on `main` and is now
closed into this one. See *Parity with the dev stack* below.

> **Rebased onto #6754** ([Main] Fix Numerical issue for GTP + Muon + MXFP8 param gather + use
> layerwise optimizer layout), which landed while this PR was open. See *Rebase onto #6754* at the
> end for exactly what that changed here.

### Behaviour change: `--layer-wise-param-layout`

The boolean `use_layer_wise_param_layout` becomes a three-valued
`layer_wise_param_layout` on both `DistributedDataParallelConfig` and `OptimizerConfig`, with the
CLI knob `--layer-wise-param-layout {padded,decoupled,legacy}`:

| value | meaning |
|---|---|
| `decoupled` (**new default**) | LayerWise (Muon 2D) buffers get a compact no-padding DDP layout and locally disable `use_distributed_optimizer` (all-reduce grads + whole-param ping-pong + `allgather_params`), while sibling buffers keep the byte-level `DistributedOptimizer`. |
| `padded` | the shard-aligned padded LayerWise layout, synced through the DDP parameter buffer. Previously the default (`use_layer_wise_param_layout=True`). |
| `legacy` | DDP is not told about LayerWise at all: every parameter lands in one non-DistOpt buffer and the optimizer owns the sync end to end. Previously `--no-use-layer-wise-param-layout`. |

A boolean could not express this once #6754 gave the `legacy` path real MXFP8 param-gather support:
`padded`, `decoupled` and `legacy` are now three distinct, individually supported configurations.
`--no-use-layer-wise-param-layout` is gone; pass `--layer-wise-param-layout legacy` instead. The
three layouts use different bf16 reduction orderings, so they are not bit-for-bit comparable with
each other.

### What's new

**1. The compact decoupled DDP layout.** On `layer_wise_param_layout='decoupled'`,
LayerWise-managed (Muon 2D-matrix) buffers get `_compute_default_per_buffer_param_layout` — the
no-padding layout — instead of the shard-aligned `dp_size * max(shard_load)` one, and locally
disable DistributedOptimizer semantics (all-reduce gradients, whole-param ping-pong ownership),
while sibling buffers (embeddings, biases, layernorm) keep the standard byte-level
`DistributedOptimizer` path.

The decision has a single source of truth: it is baked into each buffer's own `ddp_config` via
`dataclasses.replace(...)`, so bucket groups inherit it through `buffer.ddp_config` rather than
through a separately threaded attribute. A bucket *group* runs a single collective type
(reduce-scatter for DistOpt buffers, all-reduce otherwise); #6754's partition-by-optimizer-owner in
`partition_buckets` already separates the two before any merging strategy runs, so this PR only
threads the authoritative per-buffer `ddp_config` through the fp8-merge branch and asserts the
invariant. When every bucket agrees — every non-decoupled configuration — this collapses to exactly
the previous grouping.

**2. An FP8-aware whole-parameter all-gather for the decoupled path.** Muon's matrix buffers
own whole params via ping-pong assignment and sync through `allgather_params` (not DistOpt's
byte-shard reduce-scatter/all-gather, and not the fixed-shard bucket-group gather #6754 uses on the
`padded` / `legacy` layouts — a compact-layout param's ping-pong owner is unrelated to its offset
in the DDP buffer). FP8 tensors can't be all-gathered directly — mxfp8's row/col block scales can't
be derived from one another — so the transport rides BF16 and requantizes on landing: stage each
owned param from its **fp32 master → bf16** (a high-precision source, not a lossy fp8 dequant) →
**uneven all-gather-v** of the bf16 whole-params → **requantize** the gathered bf16 into *every*
rank's fp8 `param.data` (`copy_back_gathered_bf16_into_fp8_params`) so all ranks hold
`Q(bf16(master))` (== OFF / Adam) → `post_all_gather_processing` rebuilds the fp8
columnwise/transpose.

Both copy-back sites — the overlap path (`start_param_sync` / `finish_param_sync`) and the
non-overlap path (`allgather_params`) — share one helper,
`_layerwise_copy_back_gathered_params(bucket, local_rank, fp8_staged=...)`. The staging decision is
persisted on the bucket as `bucket.layerwise_fp8_staged` at `start_param_sync` time and read back by
the (possibly deferred) copy-back, so transport dtype and copy-back can never disagree. Plain bf16
(fp8pg OFF) collapses back to the original code.

Params are dispatched by **transport dtype**: fp8 and bf16 ride the bf16-staged helper, while native
fp32 params (weights marked `keep_in_fp32`, e.g. the DeepSeek-V4 CSA `ape`) are gathered in fp32 —
routing them through the bf16-staged path would silently downcast them, and mixing fp32 with bf16 in
one flatten is invalid. For a pure-bf16 model the native group is empty and this collapses to the
original single-helper dispatch.

**3. Rank-independent ping-pong ownership.** `allgather_params` requires every DP rank to agree on
each param's single owner. `numel` alone isn't a total order (a stable sort tie-breaks equal-numel
params by insertion order → different owners per rank), so ownership is now keyed by a canonical
identity `(chunk_idx, buffer_idx, global_start_index)` from model construction, with an assert that
every Muon param has one.

**4. No double-write of gathered fp8 params.** The master→model copy routes through bf16
(`Q(bf16(master))`), gated by `_layer_wise_non_distopt_child` — a tag
`LayerWiseDistributedOptimizer` sets on the `Float16OptimizerWithFloat16Params` children it wraps.
Gathered fp8 params are tagged `_layer_wise_fp8_gathered` and skipped there, because the
all-gather's requantize already writes `Q(bf16(master))` into their `param.data`; non-gathered fp8
params (e.g. MoE experts at `expt_dp == 1`, which the all-gather skips) are untagged and still get
their copy here.

Routing those Float16 children away from the DDP param buffer in `step_with_ready_grads` is handled
by the existing `hasattr(self, "_copy_main_params_to_param_buffer")` probe — only
`DistributedOptimizer` owns that byte-shard buffer — so this PR deliberately does **not** define a
raising stub of that method on `Float16OptimizerWithFloat16Params`, which would make the probe always
succeed.

(iter-0 fp32-master parity — seeding the master from `get_high_precision_init_val` rather than the
lossy fp8 dequant — was part of this PR before the rebase; #6754 landed the identical fix, so this
PR now only factors it into the shared `pop_high_precision_init_val()` helper that
`distrib_optimizer.py` already needed.)

**5. Memory: drop the non-owned high-precision init copies.** `shard_params()` removes non-owned
params from the local optimizer groups, so the Float16 wrapping only clears TE's high-precision init
copy (a full-size CPU tensor per fp8 param) for locally *owned* params. Without an explicit sweep
every DP rank would retain ~`(dp-1)/dp` of the LayerWise matrix params' bf16 CPU copies for the whole
run. The sweep is scoped to LayerWise-managed params — sibling DistOpt params must keep their init
value until their own optimizer's master creation consumes it.

**6. `force_sync` finalizes a pending LayerWise gather.** A pending LayerWise gather normally lands
in `finish_param_sync` (the forward pre-hook), which `force_sync` bypasses. `start_param_sync` now
finalizes it explicitly — otherwise every rank keeps stale `param.data` and the gather payload
pollutes `grad_data` (which the LayerWise path reuses as the all-gather receive buffer).

**7. Validation guards** (`arguments.py`), scoped to `layer_wise_param_layout='decoupled'`: it
requires `fp8_recipe ∈ {mxfp8, blockwise}`; `fp4_param_gather` is rejected unconditionally (the
LayerWise gather routes buckets by `is_layerwise_fp8_param`, so an NVFP4 param would silently take
the raw flatten path). mxfp8 additionally requires `--reuse-grad-buf-for-mxfp8-param-ag`, and the
layout requires `num_distributed_optimizer_instances == 1`, since the non-DistOpt Muon buffers
all-reduce within a single instance and partial DistOpt would under-reduce Muon gradients across the
DP domain. The `padded` and `legacy` layouts are untouched — #6754 owns their fp8 param-gather
support and its own validation.

**8. Single all-reduce buffer (a correctness fix this combination exposed).** With fp8pg ON,
LayerWise FP8 grads were keyed by `torch.uint8` in `group_params_for_buffers`, splitting Muon's
gradients into two all-reduce buffers (uint8 + bf16) where OFF has one. Both reduce in fp32, but
NCCL's fp32 accumulation order is buffer-size sensitive, so the two-buffer order diverged ~1 ULP from
OFF on the DP-reduced dense FP8 2D weights — amplified by Muon's Newton–Schulz into a visible loss
divergence. Fix: a new `merge_layerwise_fp8_grads` flag (decoupled-only) keys FP8 Muon grads by their
bf16 logical dtype so FP8 + bf16 grads share one fp32 all-reduce buffer, matching OFF. Two
all-reduces collapse to one (no extra communication); NVFP4's packed param buffer is untouched.

**9. Mixed fp8/bf16 buckets (a direct consequence of 8).** Once fp8 Muon grads key to their bf16
logical dtype, one buffer — and therefore one bucket — can hold both fp8 and bf16 params (e.g. an
fp8 matrix next to a non-quantized MoE router / DSA indexer / mHC weight). The bf16-staged transport
handles both dtypes, so the staging decision is `any(is_layerwise_fp8_param(p) for p in
bucket.params_list)` rather than a peek at `params_list[0]`: a bf16-first mixed bucket would
otherwise be routed into the raw `_flatten_dense_tensors()` path and crash on the MXFP8 `.view(-1)`.

### Correctness — bitwise ON == OFF

Deterministic, `clip_grad=0`. Compared per-iter `lm loss`, forward output, per-param reduced
`main_grad`, fp32 master, and bf16 model param.

| setup | result |
|---|---|
| H100, DP8, blockwise, deepseek_v3_proxy, 20 it | **bitwise identical, all 20 iters** |
| GB200, DP4, mxfp8, deepseek_v3_proxy, 20 it | **bitwise identical, all 20 iters** |

(Before the single-buffer fix, loss diverged at it12 on H100 / it3 on GB200; per-param pre-reduce
grads were already identical, isolating the divergence to the DP all-reduce.) With `clip_grad=0`,
Adam (DistOpt byte-shard, not split) was already bitwise. `clip_grad>0` adds a separate
layout-dependent global grad-norm fp32 sum term, so the `clip_grad=1.0` benchmarking recipes are not
bitwise — bit-exactness is established at `clip_grad=0`.

### Memory & throughput

**GB200 DP64** (mxfp8, deepseek_v4_pro_proxy 8L, TP1/PP1/EP64, SL4K):

| arm | peak alloc |
|---|---|
| Muon decoupled **+ fp8pg** | **105.8 GB** |
| Muon decoupled, no fp8pg | 116.6 GB |
| Muon **padded** (legacy default), no fp8pg | 142.9 GB |
| Adam (DistOpt) + fp8pg | 106.1 GB |
| Adam (DistOpt), no fp8pg | 116.9 GB |

FP8 param persistence −10.8 GB (identical on Muon and Adam); compact-layout padding removal −26.2 GB;
decoupled+fp8pg vs padded Muon −37 GB (~26%), and ~11 GB below Adam-no-fp8pg. **H100 dev proxy** (DP8,
blockwise) reproduces the mechanism: fp8pg −1.1 GB (35,460 vs 36,594 MB); padding `0.0%` vs `0.6%`.
**Refactor integrity:** Muon padded / fp8pg OFF vs the pre-refactor DSv4+padded-Muon baseline is
142,858 vs 142,857 MB (1 MB) — the padded path is left equivalent.

**Throughput** (GB200 wall-clock, comparable within branch): Adam < Muon decoupled < Muon padded. Muon
is ~5–6% slower than Adam (Newton–Schulz + uneven all-gather-v); fp8 param-gather itself adds ~1%;
the compact layout is ~4.8% faster than padded.

### Parity with the dev stack

The FP8/layout source and test commits from the `dev` side are carried here:

| dev | content |
|---|---|
| #5388 | the decoupled compact LayerWise DDP layout (main counterpart #5391, closed into this PR) |
| #5470 | FP8 parameter gather, plus its post-review work: compact-layout default flip + single-source-of-truth per-buffer distopt, gathered-fp8 copy-back skip, `force_sync` finalize / non-owned init cleanup / fp4 guard, native-fp32 gather dispatch, added MoE + `force_sync` coverage |

The FP8/layout feature symbols were checked symbol-by-symbol against the dev stack.

Main-only adaptations, all intentional:

- the unit test stages masters via `ChainedOptimizer.prepare_model_params_for_param_sync()` (main's
  purpose-built API, added after `dev` forked) instead of iterating `chained_optimizers`;
- the shared-buffer guard keeps main's `is_grouped_mxfp8tensor()` widening from #5487, and the
  LayerWise all-gather reads main's cached `self.dp_cp` / `self.expt_dp` groups from #6048;
- the tri-state `layer_wise_param_layout`, which `dev` does not need because #6754 is main-only.

**Golden values regenerated.** The `dist_dist_muon` ckpt-resume goldens are refreshed on this
branch, using CI's own harness (`tests/functional_tests/shell_test_utils/run_ci_test.sh`) on the
common_pile CI dataset and matching hardware, each running the full ckpt-resume flow (both phases
to iteration 100):

| test case | platform | shape |
|---|---|---|
| `..._dist_dist_muon` | `dev_dgx_h100` | 8×H100, EP8 |
| `..._dist_dist_muon` | `dev_dgx_gb200` | 2× GB200 nodes, EP8 |
| `..._dist_dist_muon_1node` | `dev_dgx_gb200` | 1 GB200 node, EP4 |

Every run reported *Exact comparison: FAILED / APPROXIMATE test: PASSED* before the refresh — a
reordering, not a regression; peak `|rel Δ|` on lm loss is ~1.6e-4. (dev's refreshed goldens are
not transferable: `main` and `dev` have independently drifted baselines for these cases, ~1e-3
across 94–97 of 100 iters, the same order as the change itself.) They were re-verified in compare
mode at the final tip: all three pass with zero exact-comparison failures.

Those three cases previously passed `--no-use-layer-wise-param-layout` (the `legacy` layout) and now
run the new default `decoupled` layout, which is what the refreshed goldens capture. The two
`*_param_layout` cases now pass `--layer-wise-param-layout padded` explicitly, since they exist to
cover the padded layout; their goldens are unchanged. `legacy` keeps the unit-test coverage #6754
added for it.

### Rebase onto #6754

#6754 merged into `main` after this PR was opened and overlaps it heavily. What changed here:

- **`layer_wise_param_layout` became tri-state** (see *Behaviour change* above). #6754 gave
  `use_layer_wise_param_layout=False` — the legacy non-DistOpt single-buffer path — real MXFP8
  param-gather support and a padded-vs-legacy parity test, so this PR no longer reuses that value
  for the new compact decoupled layout. #6754's tests were updated to the new spelling
  (`True` → `'padded'`, `False` → `'legacy'`); their behaviour is unchanged.
- **Dropped this PR's "the padded layout does not support FP8/FP4 parameter gather" guard.** #6754
  added that support.
- **Dropped this PR's copy of the `get_high_precision_init_val` FP32-master seeding** in
  `Float16OptimizerWithFloat16Params`; #6754 landed the same fix. Kept the shared
  `pop_high_precision_init_val()` helper so `optimizer.py` and `distrib_optimizer.py` use one
  implementation.
- **Dropped this PR's `partition_buckets` owner-splitting** in the `force_single_bucket_group`
  branch and the `_bucket_distopt` / `_merged_use_distributed_optimizer` helpers: #6754's
  partition-by-optimizer-owner already separates LayerWise from DistOpt buffers before any merging
  strategy runs. Kept the per-buffer `ddp_config` threading in the fp8-merge branch (which also
  fixes a latent loop-variable `buffer.ddp_config` → `fp8_buffer.ddp_config` bug there) and reduced
  the rest to one invariant assert.
- **`layerwise_param_sync_via_bucket_group`** (#6754) is forced to `overlap_param_gather` on the
  decoupled layout. It otherwise evaluates `True` — the model-level config still has
  `use_distributed_optimizer=True` for the sibling Adam buffers — and would route the decoupled
  path into `start_param_sync_for_bucket_group_subset()`. A compact-layout param's ping-pong owner
  is unrelated to its DDP-buffer offset, so the fixed-shard bucket-group gather does not apply.
- **Golden values** keep this PR's regenerated `dist_dist_muon` values (the default is still the
  decoupled layout) with main's #6539 convention applied — leading NaN iteration-time entry
  dropped, `start_step: 2`.

The measured memory / throughput / bitwise numbers above were taken before the rebase; none of the
resolutions above change the decoupled code path's behaviour.

## Issue tracking

Linked issue: Supersedes #5391 (the main-branch prerequisite, closed into this PR)

### Pre-checks

Unit test: `tests/unit_tests/test_muon_decouple_fp8_param_gather.py` — bitwise ON-vs-OFF checks on the
decoupled layout (`muon` + `use_distributed_optimizer` → `LayerWiseDistributedOptimizer`), asserting
per-step loss / forward output / `main_grad` / fp32 master / bf16 param are bitwise-identical under
`deterministic_mode`. Three cases, parametrized over `fp8_recipe ∈ {blockwise, mxfp8}`:
`test_on_vs_off_bitwise_identical` (× overlap {OFF, ON}), `test_moe_on_vs_off_bitwise_identical`
(× overlap × EP), and `test_force_sync_finalizes_pending_layerwise_gather`. blockwise runs on Hopper
only and mxfp8 on Blackwell and newer, so each platform runs 7 and skips 7. Both halves verified
against the CI `Dockerfile.ci.dev` image:

| suite | platform | result |
|---|---|---|
| `test_muon_decouple_fp8_param_gather.py` | H100 ×8 | `7 passed, 7 skipped` (blockwise; mxfp8 skipped) |
| `test_muon_decouple_fp8_param_gather.py` | GB200 ×4 | `7 passed, 7 skipped` (mxfp8; blockwise skipped) |

### Code review

Feel free to message or comment @NVIDIA/mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged\!

#### Step 2: Final Review

#### Step 3: Approved

### Merge

---

Dev-branch counterpart: https://github.com/NVIDIA/Megatron-LM/pull/5470 (merged)


kunlunl (approved):

megatron/training/arguments.py
```diff
@@ -3077,6 +3102,16 @@ def _add_distributed_args(parser):
                        help='If set, initialize with fake distributed process group and all distributed communication operations will be skipped. \
                        This is quite useful for profiling memory usage of distributed training with just one GPU. \
                        Setting WORLD_SIZE and RANK to the specific values for target distribtued scale.')
+    group.add_argument(
```
zhongbozhu, line 3345: do we need to add this? isn't the config struct automatically populated into the args?

megatron/training/training.py
```diff
@@ -1634,13 +1632,18 @@ def wrap_model_chunks_with_ddp(
     # Compute per-chunk layouts (DDP only).
     per_chunk_layouts = [None] * n
     if DP is DDP:
-        if use_layer_wise_distributed_optimizer and use_layer_wise_param_layout:
+        if use_layer_wise_distributed_optimizer:
```
zhongbozhu: why do we need to change this?

megatron/core/fp8_utils.py
```diff
@@ -308,6 +308,31 @@ def dequantize_fp8_tensor(fp8_tensor: torch.Tensor) -> torch.Tensor:
         return fp8_tensor.from_float8()
 
 
+def copy_back_gathered_bf16_into_fp8_param(model_p: torch.Tensor, src_bf16: torch.Tensor) -> None:
+    """Requantize a gathered bf16 whole-param into an fp8 (Float8/MXFP8) model param in place.
+
+    mxfp8 columnwise can't be derived from rowwise, so force columnwise before copy_ (TE rebuilds
+    both directions from the bf16); blockwise/Float8 columnwise is a lossless transpose.
+    """
+    if is_mxfp8tensor(model_p):
```
zhongbozhu: this won't account for single weight, so we should either support it or ban it properly in the transformer config level

megatron/core/fp8_utils.py
```diff
@@ -308,6 +308,31 @@ def dequantize_fp8_tensor(fp8_tensor: torch.Tensor) -> torch.Tensor:
...
+
+def _stage_param_to_bf16(p: torch.Tensor) -> torch.Tensor:
+    """Stage a param to a detached bf16 whole-param for fp8 param-gather transport.
+
+    Prefer the fp32 master (high-precision source); else dequantize an fp8 param; else copy bf16.
+    """
+    main_param = getattr(p, "main_param", None)
+    if main_param is not None:
+        return main_param.detach().to(torch.bfloat16)
+    if is_float8tensor(p):
```
zhongbozhu: `is_float8tensor` is actually very vague, it could mean many fp8 recipes, so I suggest being a bit more explicit here ( if it's fp8 blockwise recipe, then fp8 blockwise, or mxfp8 tensor, etc.)

megatron/core/fp8_utils.py
```diff
@@ -308,6 +308,31 @@ def dequantize_fp8_tensor(fp8_tensor: torch.Tensor) -> torch.Tensor:
...
+        return dequantize_fp8_tensor(p).detach().to(torch.bfloat16)
```
zhongbozhu: this seems like a numerical lossy operation, when will `_stage_param_to_bf16` be used?

megatron/core/optimizer/optimizer_config.py
```diff
@@ -327,6 +327,13 @@ class OptimizerConfig:
     arguments layer sets this flag and resets ``use_distributed_optimizer`` to False so
     that the standard distributed-optimizer path is not triggered."""
 
+    use_layer_wise_param_layout: bool = False
```
zhongbozhu: can you elaborate what does `layout` mean here?

megatron/core/optimizer/optimizer.py
```diff
@@ -1110,12 +1131,46 @@ def _copy_model_grads_to_main_grads(self):
                 model_param.grad = model_param.main_grad
 
     def _copy_main_params_to_model_params(self):
```
zhongbozhu, line 1139: this name is too similar with `_copy_main_params_to_param_buffer` and why do we need this in the first place?

megatron/training/arguments.py
Wohox: I checked this more carefully: this flag is not auto-generated today. Neither `OptimizerConfig` nor `DistributedDataParallelConfig` is registered with `ArgumentGroupFactory`; `get_megatron_optimizer_config()` and `get_megatron_ddp_config()` only copy attributes that already exist on the argparse namespace.

Since this PR removes the old `--no-use-layer-wise-param-layout` flag and introduces the opt-in `--use-layer-wise-param-layout` form, this explicit parser entry is still required. I’ll keep it.

megatron/training/training.py
Wohox: This change makes `ddp_config.use_layer_wise_param_layout` the single source of truth for selecting the layout.

When the LayerWise optimizer is enabled, `LayerWiseDistributedOptimizer.compute_full_param_layout()` needs to run for both values: `True` builds the shard-aligned padded layout for LayerWise-managed parameters, while `False` builds the compact no-padding layout for those parameters and retains the standard DistOpt layout for sibling parameters such as embeddings, biases, and layer norms.

Therefore, skipping layout computation when the flag is `False` would also lose the required per-buffer routing and layout information.

megatron/core/fp8_utils.py
Wohox: Good catch. This copy-back path currently handles plain Float8/MXFP8 parameters, but it does not handle the Transformer Engine single-grouped-weight `GroupedTensor` representation correctly.

I’ll explicitly reject `moe_single_grouped_weight` with decoupled LayerWise FP8 parameter gather in config validation, unless we add a GroupedTensor-aware copy-back path and corresponding test coverage.

megatron/core/fp8_utils.py
Wohox: For this Muon path, every locally owned source parameter must have an FP32 `main_param`, so the intended operation is FP32 master → BF16. This is intentionally lossy relative to FP32, but it matches the FP8-param-gather-OFF baseline, which produces `Q(bf16(master))`.

The FP8-dequant fallback is not expected to execute in this path. I’ll narrow the helper or assert the `main_param` invariant so that it cannot silently take the lossier fallback.

Wohox: `_stage_param_to_bf16` is used by the two alternative LayerWise parameter-sync implementations: `LayerWiseDistributedOptimizer.allgather_params()` for the non-overlap path and `_ParamAndGradBucketGroup.start_param_sync()` for the overlap path. Only one of these paths is used for a given sync configuration.

megatron/core/optimizer/optimizer.py
Wohox: `_copy_main_params_to_model_params()` is the existing `Float16OptimizerWithFloat16Params` hook; the method name predates this change and pairs with `_copy_model_params_to_main_params()`.

`_copy_main_params_to_param_buffer()` has a different role: it is the DistributedOptimizer/MXFP8 staging helper that writes FP32 master shards into the DDP parameter buffer before parameter all-gather.

A compact non-DistOpt LayerWise child does not own that byte-sharded parameter buffer, so after the optimizer step it must use `_copy_main_params_to_model_params()` instead. The new branch also ensures FP8 parameters are written as `Q(bf16(master))` and skips parameters that will already be written by the LayerWise gather.

megatron/core/optimizer/optimizer_config.py
Wohox: Here, `layout` means the mapping of each parameter into the contiguous DDP buffer: its start/end indices, bucket assignment, and the padding and shard boundaries used by that buffer.

For LayerWise-managed parameters, the padded layout places every whole parameter inside one DP shard, aligns parameter starts, and pads the shards to the same size. The compact layout instead uses the normal no-padding DDP ordering and keeps whole-parameter ownership in the LayerWise optimizer’s ping-pong assignment.

The layout always determines the grad-buffer placement. It also determines the param-buffer placement when that buffer uses DistributedOptimizer; compact non-DistOpt LayerWise buffers do not allocate a DDP param buffer.

megatron/core/fp8_utils.py
Wohox: `is_float8tensor` is intentionally a broad abstraction over Transformer Engine quantized tensor implementations, but I agree that it is too broad as the capability check in this path.

The LayerWise parameter-gather path currently supports only MXFP8 and blockwise FP8, so I’ll make that restriction explicit in the predicate/config validation rather than relying on the generic classification.

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -255,6 +255,26 @@ def _build_model_gbuf_range(cls, param_and_grad_buffer: _ParamAndGradBuffer, buc
 
         return data
 
+    @classmethod
+    def _filter_gbuf_range_map(
```
zhongbozhu: why do we need this?

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
wujingyue, line 1: Sorry, I don't follow the changes in this file. 

Could you move these changes to a **separate** PR and code-comment the **why** better? 

https://google.github.io/eng-practices/review/developer/small-cls.html

megatron/core/optimizer/distrib_optimizer.py
Wohox: removed

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
Wohox: There are multiple places that call the `high_precision_init_value` method, this is to reduce the duplicated code.

wujingyue: I'm still confused. The change in this file made the code strictly harder to read -- what does this dedup buy us here?

PR #5491, opened 2026-06-25
Refactor RL rollout pipeline

lauradang:

# What does this PR do?

Refactors the `GroupedRolloutGenerator` rollout pipeline introduced in #5306 so each pipeline stage (submission gating, inference, assembly, consumption ordering) is explicit and independently legible. Behavior is unchanged for the five supported granularity modes (`B/B`, `G/B`, `G/G`, `R/B`, `R/G`); this is a code-organization change validated against the same five-mode matrix.

Refactor highlights:

- **Inverts control inside agents.** Agents now expose three small primitives instead of a single monolithic `group_rollout()`:
  - `prepare_group_rollout(request) -> GroupRolloutParams` — builds the per-group `InferenceRequest` and a `build_rollout` callback once per group.
  - `_agenerate(request, inference_request) -> InferenceResponse` — the inference call, overridable per agent.
  - `_rollout_from_response(...)` — post-processes a single response into a `Rollout`.

  The pipeline scheduler now lives in `Agent`/`GroupedRolloutGenerator` in `api.py` and drives those primitives stage-by-stage.

- **Adds explicit pipeline-stage types.** New `_GranularityConfig`, `_SubmissionGate`, `_InferWorkItem`, and `_InferredItem` dataclasses make each stage's inputs/outputs and the submission-gate release semantics legible at the call site.

- **Hardens granularity types.** `SubmissionGranularity` (`R|G|B`), `ConsumptionGranularity` (`G|B`), and `ReleaseState` live in `rollout_granularity.py` with a `RELEASE_STATE_BY_SUBMISSION` mapping. `consumption_granularity = "R"` is now rejected at the type level (was a runtime rejection in #5306).

- **Hardens request validation.**
  - `filter_groups_with_same_reward = True` is rejected with a message noting that dropped groups are not regenerated, so non-streaming callers would receive fewer groups than requested and batch-order consumers would stall on incomplete batches.
  - The `B-submit / G-consume` combination continues to be rejected.

- **Preserves rollout group count for streaming submissions.** Streaming submissions no longer collapse `num_groups` to 1; the pipeline handles per-rollout fan-out internally and callers receive a consistent group count regardless of submission granularity.

- **Threads the new agent surface through `RewardOnlyAgent`, `WeightedMultiTaskAgent`, and `FastAPIEnvServer`.** `FastAPIEnvServer._agenerate` raises `NotImplementedError` because it forwards rollouts to a remote env server rather than running local inference, and the rest of the FastAPI path now uses the pipeline-driven flow.

- **Adds unit coverage** for the new request validation (filter-groups rejection), streaming group-count preservation, `_agenerate` rejection on the FastAPI server, and a new `RemoteAgent` instantiation test.

## Issue tracking

Linked issue: N/A

### Pre-checks

- [x] I have added relevant functional/manual validation
- [x] I have added relevant documentation through CLI help text (no new CLI flags — granularity surface unchanged from #5306)

### Testing

Static/unit validation:

- `uv run --no-sync pytest tests/unit_tests/rl/test_grouped_rollouts.py tests/unit_tests/rl/test_remote_agent.py -q`
- `uv run --no-sync ruff check megatron/rl/agent/api.py megatron/rl/agent/reward_only_agent.py megatron/rl/agent/weighted_multi_task.py megatron/rl/server/agent/fastapi_env_server.py megatron/rl/rollout_granularity.py megatron/rl/rl_utils.py tests/unit_tests/rl/test_grouped_rollouts.py tests/unit_tests/rl/test_remote_agent.py`

Manual hsg2 validation (hsg2, 2026-06-26):

All five runs used the same Qwen2.5-3B / GSM8K setup with TP=2, PP=1, `grpo_group_size=2`, `grpo_prompts_per_step=2`, `--rl-partial-rollouts`, `--rl-generation-lag 4`, detailed timing enabled (`--timing-log-level 1`, `--timing-log-option minmax`, `--log-timers-to-tensorboard`), and `--train-samples 48828125`. Runs were submitted against megatron-rl commit `f7a284759aaa4700446240997c1aeaefc152382c` on the `nemotron_sw_post` account, partition `batch`, 2 nodes × 4 GPUs each, 4h SLURM wall time.

Run matrix:

| Mode | SLURM | W&B | Final SLURM state | Elapsed | Iterations |
|---|---:|---|---|---:|---:|
| B-submit / B-consume | `3600617` | https://wandb.ai/adlr/megatron-rl/runs/fevbefei | `COMPLETED` (early exit at iter 110) | 1:15:06 | 110 |
| G-submit / B-consume | `3600621` | https://wandb.ai/adlr/megatron-rl/runs/5i7vozah | `COMPLETED` (near 4h wall) | 3:58:34 | 702 |
| G-submit / G-consume | `3600630` | https://wandb.ai/adlr/megatron-rl/runs/aj1hfz0s | `TIMEOUT` at 4h | 4:00:12 | 419 |
| R-submit / B-consume | `3600638` | https://wandb.ai/adlr/megatron-rl/runs/f23te6mr | `COMPLETED` (clean exit, `DerivedExitCode 0:0`) | 1:08:28 | 218 |
| R-submit / G-consume | `3600642` | https://wandb.ai/adlr/megatron-rl/runs/t5ubgpkx | `COMPLETED` | 2:31:52 | 270 |

Performance summary, parsed from W&B `iteration-time` after warmup:

| Mode | Median step | Mean step | p90 step | p99 step | Max step |
|---|---:|---:|---:|---:|---:|
| B/B | 7.84s | 12.88s | 14.64s | 97.16s | 115.24s |
| G/B | 7.02s | 7.56s | 8.65s | 14.12s | 99.58s |
| G/G | 7.71s | 7.97s | 9.26s | 10.83s | 13.79s |
| R/B | 7.33s | 7.52s | 8.03s | 11.40s | 16.65s |
| R/G | 7.16s | 7.22s | 7.50s | 8.49s | 11.20s |

Staleness summary, from W&B scalar history (`*_policy_staleness`, `*_policy_last_token_staleness`):

| Mode | Latest mean/min/max policy staleness | Run mean policy staleness | Run median policy staleness | Max observed policy staleness | Max observed last-token mean staleness | Readout |
|---|---:|---:|---:|---:|---:|---|
| B/B | 4 / 4 / 4 | 3.97 | 4.00 | 4 | 4.00 | bounded exactly at configured lag |
| G/B | 8 / 8 / 8 | 12.31 | 10.00 | 62 | 61.00 | bounded near lag with transient tail spikes |
| G/G | 4 / 3 / 5 | 7.35 | 7.00 | 24 | 10.00 | usually low, with transient stale spikes |
| R/B | 159 / 159 / 159 | 75.53 | 71.00 | 159 | 159.00 | fast median path; builds/drains stale backlog |
| R/G | 201.5 / 200 / 203 | 93.77 | 78.00 | 208 | 202.00 | high transient staleness during backlog drain |

Validation notes:

- `R-submit / G-consume` again had the tightest step-time distribution (median 7.16s, p99 8.49s) — best wall-clock progress per second of cluster time, mirroring the #5306 AWS1 finding.
- `G/G` was the stable middle option: tighter step-time distribution than R-submit modes but slower median, and the only run to consume the full 4h wall (TIMEOUT at iter 419).
- `B/B` and `R/B` exited cleanly at ~1h (iter 110 / 218) without consuming the full wall. The post-mortem on B/B shows the worker was inside `save_checkpoint`/`barrier` when SIGTERM landed; both `B/B` and `R/B` reported `DerivedExitCode 0:0` or `0:15` and SLURM `COMPLETED`. The remaining three runs progressed long enough to dominate the staleness/perf comparison.
- Qualitative staleness shape matches #5306's AWS1 run: `B/B` exactly lag-bounded; `G/B` bounded near lag with infrequent large spikes; `G/G` usually low with rare transient spikes; R-submit modes lead step-time but show large transient staleness during backlog drain.
- No tracebacks, OOMs, or NCCL errors during the active training phase across the five logs.


tests/unit_tests/rl/test_remote_agent.py
tdene, line 1: I don't see the point of this test.

megatron/rl/rl_utils.py
```diff
@@ -572,11 +572,8 @@ def get_rollout_generator(args, inference_interface, n_prompts, samples_per_grou
     if not (streaming := args.rl_partial_rollouts) or _ROLLOUT_GENERATOR is None:
         parallel_generation_tasks = get_rl_parallel_generation_tasks(args)
         agent = get_agent(args, parallel_generation_tasks=parallel_generation_tasks)
-        num_groups = n_prompts
-        if streaming and args.rl_submission_granularity != "B":
-            num_groups = 1
         request = GroupedRolloutRequest(
-            num_groups=num_groups,
```
tdene, line 579: Can you add a test to cover this multi-env failure that you fixed in this PR? There should be a unit-test to catch any such issues in the future.

megatron/rl/agent/weighted_multi_task.py
```diff
@@ -153,20 +154,24 @@ def _distribute_counts(self, total_count: int, distribute_remainder: bool = True
...
-    async def group_rollout(
+    async def prepare_group_rollout(
         self,
         request: GroupedRolloutRequest,
-        submission_gate: asyncio.Semaphore | None = None,
-    ) -> list[Rollout]:
+    ) -> GroupRolloutParams:
         raise NotImplementedError(
-            "WeightedMultiTask is a collection of tasks and therefore doesn't implement this method directly. Use get_grouped_rollouts instead to generate grouped rollouts."
+            "WeightedMultiTask overrides get_grouped_rollouts; prepare_group_rollout is not used."
```
tdene: I like the original error message more.

megatron/rl/agent/reward_only_agent.py
```diff
@@ -131,31 +139,25 @@ async def rollout(self, request: RolloutRequest) -> Rollout:
...
-                response = await request.inference_interface.agenerate(inference_request)
-            else:
-                async with submission_gate:
-                    response = await request.inference_interface.agenerate(inference_request)
-            return await self.rollout_from_response(request, response, golden)
+        async def build_rollout(response: InferenceResponse) -> Rollout:
+            return await self._rollout_from_response(request, response, golden)
 
-        return await asyncio.gather(*[generate_one() for _ in range(request.rollouts_per_group)])
+        return GroupRolloutParams(inference_request=inference_request, build_rollout=build_rollout)
```
tdene: If left like this, this should be a `functools.partial` call instead of a new method definition.

Personally, I would just return `(request, response, golden)` instead of a `Callable`: it's lighter-weight. But if a `Callable` is the returnable, it should use `functools.partial`.

megatron/rl/agent/reward_only_agent.py
```diff
@@ -123,6 +124,13 @@ async def rollout_from_response(
 
         return rollout
 
+    async def _agenerate(
```
tdene: I would be more comfortable with this being called `get_rollout`. The fact that doing so involves `generate` and `a` is an implementation choice. Ultimately the rollout can be gotten from non-generation sources, and it is likely we will move in that direction by September.

megatron/rl/agent/reward_only_agent.py
```diff
@@ -123,6 +124,13 @@ async def rollout_from_response(
...
+        self,
+        request: RolloutRequest | GroupedRolloutRequest | EvaluationRequest,
+        inference_request: InferenceRequest,
+    ) -> InferenceResponse:
+        return await request.inference_interface.agenerate(inference_request)
+
     async def rollout(self, request: RolloutRequest) -> Rollout:
 
         prompt, golden = await self.get_prompt(validation=request.validation)
```
tdene: This method is strange to me now. The only thing that still uses it are `get_reward_rollouts` calls inside abstract child classes.

We should not break the abstract child classes. It's good organization.

But, am I correct that we can eliminate this method altogether from `reward_only_agent` at this point? It's still defined in the abstract child class as `...`, and that seems good enough to me?

Maybe I'm wrong in how I'm looking at the code organization here, let me know.

megatron/rl/agent/api.py
```diff
@@ -187,6 +215,217 @@ async def get_reward_rollouts(self, request: RolloutRequest) -> list[TokenRollou
...
+    def from_request(cls, request: GroupedRolloutRequest) -> "_GranularityConfig":
+        cls._validate(request)
+        return cls(
+            submission=request.submission_granularity,
+            consumption=request.consumption_granularity,
+            num_groups_per_batch=request.num_groups,
+        )
+
+    @property
+    def consume_in_batch_order(self) -> bool:
```
tdene: It would be nice to keep this method name more high-level and descriptive.

I'm thinking. The purpose of this toggle is to determine whether or not the dataset is re-ordered (see the diagrams from the slides we're relying on).

Can we call this something like `prevent_dataset_reorder` or `freeze_dataset_order`?

megatron/rl/agent/api.py
```diff
@@ -187,6 +215,217 @@ async def get_reward_rollouts(self, request: RolloutRequest) -> list[TokenRollou
...
+    index_in_batch: int
+    params: GroupRolloutParams
+
+
+@dataclass(frozen=True)
+class _InferredItem:
+    """One rollout post-inference, flowing from infer to assemble."""
+
+    item: _InferWorkItem
+    response: InferenceResponse
```
tdene: I think that this (and other similar frozen dataclasses in this codepath) is one of the few times when it is better to use `NamedTuple`.

These are light-weight objects that are frozen and passed through queues. That sounds like the exact use-case that `NamedTuple` were kept around in Python for.

megatron/rl/agent/api.py
```diff
@@ -199,134 +438,42 @@ def __init__(self, *, parallel_generation_tasks: int | None = None, **kwargs):
...
+        """Return the params for one group's rollouts.
+
+        Called once per group by _RolloutPipeline.stage_prepare. The returned
+        build_rollout closure is invoked once per inference response in
+        _RolloutPipeline.stage_assemble.
+        """
         ...
 
-    async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
+    async def get_grouped_rollouts(
```
tdene, line 522: Beautiful code!!

megatron/rl/agent/api.py
```diff
@@ -187,6 +215,217 @@ async def get_reward_rollouts(self, request: RolloutRequest) -> list[TokenRollou
...
+        in_flight: set[asyncio.Task[None]] = set()
+        try:
+            while True:
+                try:
+                    item = await self.infer_queue.get()
+                except asyncio_QueueShutDown:
+                    break
+                task = asyncio.create_task(self._infer_one(item))
+                in_flight.add(task)
+                task.add_done_callback(in_flight.discard)
```
tdene: I still really dislike the fact that we are building tasks on-the-fly, specifically because this is a case where we know the total number of tasks ahead of time.

We can just spawn `parallel_generation_tasks` number of tasks during init, and have them all run `.get()` on the queue.

I would prefer to have permanent, persistent, tasks rather than spinning them up and down and tagging them with callbacks.

megatron/rl/agent/api.py
```diff
@@ -187,6 +215,217 @@ async def get_reward_rollouts(self, request: RolloutRequest) -> list[TokenRollou
...
+                )
+                if keep:
+                    first = completed[0]
+                    await self.output_queue.put(
+                        RolloutGroup(
+                            rollouts=rollouts,
+                            batch_id=first.item.batch_id,
+                            index_in_batch=first.item.index_in_batch,
+                        )
+                    )
```
tdene, line 456: Please either comment out this code-block, or write a note saying that it is currently broken non-functional code.

It's good to keep it around so a later PR can fix it, but it's confusing that it seems to be live and is only being disabled by `_validate`.

megatron/rl/agent/api.py
```diff
@@ -187,6 +216,291 @@ async def get_reward_rollouts(self, request: RolloutRequest) -> list[TokenRollou
...
+        parallel_generation_tasks: int,
+    ) -> None:
+        self.agent = agent
+        self.request = request
+        self.gran_policy = _GranularityConfig.from_request(request)
+        self.gate = _SubmissionGate(
+            capacity=parallel_generation_tasks,
+            submission=self.gran_policy.submission,
+        )
+        self.output_queue_maxsize = buffer_size if request.streaming else 0
```
tdene: Just noticed this maxsize. I don't understand why it's needed. In fact, it looks very dangerous.

megatron/rl/rl_utils.py
```diff
@@ -1100,6 +1099,89 @@ def prep_wandb_metrics(
...
+        pipeline.prepared_count = 0
+        pipeline.inferred_count = 0
+        pipeline.assembled_count = 0
+        pipeline.yielded_count = 0
+        gate.prepare_blocked_seconds = 0.0
+        gate.acquire_calls = 0
+        gate.release_calls = 0
+
+    # WeightedMultiTask work distribution (agent_slots / agent_pgts).
+    dist = getattr(_ROLLOUT_AGENT, "latest_distribution", None)
```
tdene, line 1172: Is this debug code that should be removed?

megatron/rl/agent/reward_only_agent.py
lauradang: Confirmed `rollout()` is never reached from `train_rl.py` — training goes through `get_grouped_rollouts` and evaluation through `run_evaluation`. Its only live path is the standalone env server: `/rollouts/` → `get_reward_rollouts` → `rollout()`.

Deleting it outright still breaks though: `rollout` is `@abstractmethod` on `RolloutGenerator`, so without the `RewardOnlyAgent` override every env subclass fails at startup (`TypeError: Can't instantiate abstract class MathAgent without an implementation for abstract method 'rollout'`).

Proposal: invert the interface — make `get_reward_rollouts` the abstract method and drop `rollout` entirely (the default `get_reward_rollouts` was its only caller):

```python
class RolloutGenerator(Agent, ABC):
    @abstractmethod
    async def get_reward_rollouts(self, request: RolloutRequest) -> list[Rollout]: ...
```

`RewardOnlyAgent` implements it as N × (`prepare_group_rollout` → `get_rollout_response` → `build_rollout`) gathered — same per-rollout composition as today, so `/rollouts/` behavior is unchanged. `WeightedMultiTask` and `FastAPIEnvServer` already override `get_reward_rollouts`, so their `rollout` stubs just get deleted, and envs stay instantiable via the inherited implementation.

megatron/rl/agent/reward_only_agent.py
lauradang: Renamed in 982b81166, with one adjustment: `get_rollout_response` instead of `get_rollout`. This method returns the raw `InferenceResponse`, and `rollout()` already exists on the same class returning a built `Rollout`, so `get_rollout` would collide with it. `get_rollout_response` keeps the generation-agnostic framing — a non-generation source can still be the thing that returns the response backing a rollout. If you'd still rather have `get_rollout`, easy to change.

megatron/rl/agent/api.py
lauradang: This isn't new in this PR, it's carried over from the previous implementation (`asyncio_Queue(maxsize=self.buffer_size if request.streaming else 0)`), which documented it as "backpressure for balanced generation in a multi-task setting."

megatron/rl/rl_utils.py
lauradang: Not leftover as this logs how rollout work is split across environments each iteration: groups and generation slots assigned per env. In multi-env runs, a bad weight config or a starved env (zero groups) is otherwise invisible until the run stalls which is exactly the failure mode this PR fixed.

megatron/rl/agent/api.py
tdene: It's not new, but it's part of the lag strategy code. It was useful before the previous lag-shaping PR, but at this point it's become a dangerous artifact that will break the lag-shaping logic.

megatron/rl/agent/reward_only_agent.py
tdene: I'm very happy with this proposal; it's the right way to do things.

megatron/rl/rl_utils.py
tdene: Sounds good!

megatron/rl/agent/api.py
lauradang: Removed in c6e158834: the output queue is now unconditionally unbounded and the buffer_size plumbing (and the output_queue_maxsize metric) is gone. The submission gate is the sole flow control — with the bound in place, once 10 assembled groups sat unconsumed, stage_assemble blocked on put(), stopped draining assemble_queue, gate slots stopped releasing, and effective run-ahead clamped to ~10 groups regardless of --rl-generation-lag.

tdene (approved):

PR #5492, opened 2026-06-25
Add statistics logging for params and activations

philipcmonk:

# What does this PR do ?

This introduces per-parameter and per-activation-tensor statistics, logged to a jsonl file for later processing.

We log statistics for parameters, wgrads, activations, and dgrads.  These form two natural groups: parameters and wgrads have the size of the model, are calculated after the forward and backward passes have completed, and are generally fast enough to run on every step.  We store statistics for each of these at the granularity of a parameter -- for example, `decoder.layers.3.mlp.linear_fc2.weight`.

Activations and dgrads are summarized over the size of the activations times the number of times we store statistics in the model and are calculated by hooks during the forward and backward passes (and all-reduced after).  We store statistics at the beginning and end of most linear layers -- for example `decoder.layers.3.mlp.linear_fc2/input0`.  This incurs a nontrivial cost, I measured up to 30% of the step time.  Consider setting `--activation-log-interval` to 100 steps to reduce the impact.

For all of these, we're most interested in the mean, variance/stddev, L2 norm, RMS norm, and kurtosis.  We want to be able to compute these statistics for arbitrary combinations -- for example, across all the parameters in a logical layer.  These statistics don't necessarily aggregate well directly, but they're all derivable from the first four raw moments (sum, sum of squares, sum of cubes, and sum of fourth powers), so we compute and log those.

A performance consideration is that computing these statistics across many tensors can be slow without a multi_tensor-style kernel that fuses the four statistics.  I've opened [a PR](https://github.com/NVIDIA/TransformerEngine/pull/3144) in TransformerEngine with an implementation of this, and a raw torch backup is used if that kernel can't be found.

For reference, these are the step times I observed on a 1B dense model with and without the TE kernel.
```
                        step time change w/TE kernel (ms)
baseline                418 -> 419
log params+wgrads       450 -> 428
log all 4               880 -> 500
log all 4 w/interval=10 450 -> 430 (except every 10 steps, when it's roughly 880 -> 500)
```

I have a simple streamlit dashboard to visualize the logged data, but I have not included it in this PR.  It could perhaps be added to the tools/ directory, but this PR is already pretty large.

A stylistic note is that this adds core/parameter_names.py to encapsulate the messy details of getting globally canonical names of parameters.  This includes both name rewriting and `all_gather_index()` to get a globally consistent mapping of names to an index.  This problem is solved ad-hoc in several other places, so hopefully over time they can be migrated to use this.

### Pre-checks


megatron/core/optimizer/layer_wise_optimizer.py
```diff
@@ -744,6 +745,21 @@ def _get_grad_norm_for_group(self, grad_norm_group: str):
...
+        names = []
+        grads = []
+        for optimizer in self.chained_optimizers:
+            for name, param in optimizer.get_named_parameters_for_grad_norm(registry):
+                grad = optimizer._get_grad_for_grad_norm(param)
+                if not optimizer._include_param_in_grad_norm(param, grad):
+                    continue
+                names.append(name)
+                grads.append(grad.detach())
+
```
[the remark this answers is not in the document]
philipcmonk: I struggled with this one a bit and am not sure I got it right in the end.  `None` as the world group is used based on the similar logic in [get_grad_norm](https://github.com/philipcmonk/Megatron-LM/blob/34f71bc9cdb3f1ca741134091af6c84913d4589d/megatron/core/optimizer/layer_wise_optimizer.py#L709) and [count_zeros](https://github.com/philipcmonk/Megatron-LM/blob/34f71bc9cdb3f1ca741134091af6c84913d4589d/megatron/core/optimizer/layer_wise_optimizer.py#L768).

However, I notice in their [helper](https://github.com/philipcmonk/Megatron-LM/blob/34f71bc9cdb3f1ca741134091af6c84913d4589d/megatron/core/optimizer/clip_grads.py#L132-L138) [functions](https://github.com/philipcmonk/Megatron-LM/blob/34f71bc9cdb3f1ca741134091af6c84913d4589d/megatron/core/optimizer/clip_grads.py#L264-L272) that if they're dtensors, they also reduce across that data parallel group.  That seems like a double-count to me, since reducing across `None` should already include those devices, I believe.

megatron/core/per_parameter_stats.py
```diff
@@ -0,0 +1,358 @@
...
+    if sequential_match is not None:
+        prefix, local_expert_index, suffix = sequential_match.groups()
+        return f"{prefix}{int(local_expert_index) + expert_offset}{suffix}"
+
+    return local_name
+
+
+def _get_local_expert_offset(num_experts: int) -> int:
+    expert_group = parallel_state.get_expert_model_parallel_group(check_initialized=False)
+    if expert_group is None:
```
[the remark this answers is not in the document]
philipcmonk: I see this now in AGENTS.md.  Addressed in 27c35163

megatron/training/config/training_config.py
```diff
@@ -251,6 +251,30 @@ class LoggerConfig:
...
+    """If set, calculate and log count and raw sums of powers 1-4 for backward data gradients
+    keyed by module site to JSONL statistics files."""
+
+    activation_log_interval: int | None = None
+    """Interval for activation and dgrad raw-moment statistics. If unset, uses
+    tensorboard_log_interval."""
+
+    statistics_log_dir: str | None = None
+    """Directory for high-cardinality JSONL statistics. If unset, statistics use
+    tensorboard_dir when available, then save as a fallback."""
```
deepakn94: I don't follow "save as a fallback"

megatron/training/config/training_config.py
```diff
@@ -251,6 +251,30 @@ class LoggerConfig:
     log_params_norm: bool = False
     """If set, calculate and log parameters norm."""
 
+    log_param_raw_moments_by_param: bool = False
+    """If set, calculate and log count and raw sums of powers 1-4 for each parameter separately
```
deepakn94: Maybe all of these should be integer intervals as opposed to booleans?

megatron/training/config/training_config.py
philipcmonk: Will reword to "then the --save directory as a fallback."

megatron/training/config/training_config.py
philipcmonk: Intervals of steps, so 0 means don't log, 10 means log every 10 steps, etc?  That makes sense

megatron/training/config/training_config.py
```diff
@@ -250,6 +250,28 @@ class LoggerConfig:
     log_params_norm: bool = False
     """If set, calculate and log parameters norm."""
 
+    log_param_raw_moments_by_param: int = 0
```
deepakn94: Maybe all of these need "_interval" suffixes at the end?

philipcmonk: Fair.  It's becoming quite a mouthful, maybe we can reduce to something like this:

```
--log-param-stats-interval
--log-wgrad-stats-interval
--log-activation-stats-interval
--log-dgrad-stats-interval
```

This also opens the door to other stats beside raw moments that we might want to add later using the same machinery.

philipcmonk:
I pushed some commits cleaning up some verbose code, as well as renaming the cli args (suggested by @deepakn94) and adding two more logs (suggested by @mkhona-nvidia):
```
--log-residual-stats-interval
--log-residual-grad-stats-interval
```

These log only the residual stream.  This makes them much cheaper and compatible with cudagraphs (except full-iteration), unlike `--log-activation-stats-interval` and `--log-dgrad-stats-interval`, which log the input and output of modules.

I also added behavior for act/dgrad logging to temporarily suspend cudagraphs for the steps when they're run.  In core/, this is a generic utility to suspend cudagraphs, and in training/ we invoke it on the relevant steps.

guihong-nv:
@philipcmonk May you fix the linting issues? You can run `bash tools/autoformat.sh`

philipcmonk:
This looks like a new lint check added in #5989, which is triggering because of the changes in #5916 that i pulled into this branch because they're in the same area of the code.  I'm working on fixing that.

Probably best to wait for that to hit main in a few hours, then I'll just update this branch from main.

megatron/core/models/gpt/fine_grained_callables.py
```diff
@@ -75,6 +76,7 @@ def submodule_pre_dispatch_forward(node: ScheduleNode, hidden_states: torch.Tens
             isinstance(layer, GraphableMegatronModule)
             and hasattr(layer, 'cuda_graphs')
             and layer.cuda_graphs
+            and not is_cuda_graph_replay_suspended()
```
zhongbozhu, line 79: why do we need to change this?

megatron/core/tensor_parallel/layers.py
```diff
@@ -89,11 +89,17 @@
...
+def param_is_not_tensor_parallel_duplicate(param, tp_group=None, expert_tp_group=None):
+    """Return whether a parameter contributes to a unique model-parallel shard.
+
+    Parameters reduced over expert data parallel groups use the expert tensor-parallel
+    group for duplicate filtering. Other parameters use the regular tensor-parallel group.
+    """
     if hasattr(param, "tensor_model_parallel") and param.tensor_model_parallel:
         return True
+    # allreduce=False marks parameters reduced over expert DP, so filter their duplicates over ETP.
+    if not getattr(param, "allreduce", True) and expert_tp_group is not None:
```
zhongbozhu, line 104: Aren't these code belong to this PR https://github.com/NVIDIA/Megatron-LM/pull/5916 ? Need to rebase after that one lands.

philipcmonk: Yes, that was in the merge queue yesterday so I thought it was going to be in soon, but the queue has stalled.  Will fix once that goes in.

megatron/core/models/gpt/fine_grained_callables.py
philipcmonk: `--log-activation-stats-interval` and `--log-dgrad-stats-interval` are not compatible with cudagraphs, so I added a way to suspend cudagraphs for the specific steps when those are active.  Since those are kind of slow, I would expect them to only be logged every 100 or 1k steps.

philipcmonk:
@guihong-nv @zhongbozhu This PR has been rebased from main, so the linting issues and extra changes are fixed.  The failing GB200 test is an unrelated infra issue the CI team is working on.

philipcmonk:
Closing in favor of #6552

PR #5497, opened 2026-06-25
Add cos/sin width guard to fused MLA RoPE kernels

ShauryaaSharma:

# What does this PR do ?

Add a width guard to the fused MLA RoPE kernels so an undersized cos/sin cache fails with a clear error instead of silently reading out of bounds.

The fused kernels (`ApplyMLARotaryEmbQ` / `ApplyMLARotaryEmbKV`) read `emb_dim` cos/sin values per token and assume the cache is `emb_dim` wide. When the rotary cache is built with `rotary_percent < 1`, the base `RotaryEmbedding` shrinks it to `int(emb_dim * rotary_percent)` , narrower than `emb_dim` , and the kernel reads past the buffer, picks up uninitialized GPU memory, and produces NaN gradients. This is reachable on `main`: `MultiLatentAttention` forwards `rotary_percent` into the base `RotaryEmbedding` for `rope_type="rope"` (`multi_latent_attention.py:184`), and the fused branch (`:870`) calls the kernel regardless of rope type with no width check today (`fused_mla_yarn_rope_apply.py:251-255`). The guard asserts `cos.shape[-1] == emb_dim` (and `sin`) in both forward paths, turning a silent ~1e14 out-of-bounds read into an immediate, readable failure.

## Issue tracking

Linked issue: Related to #5317

This guard makes the failure mode in #5317 loud instead of silent; the root-cause fix for that issue is `rotary_percent = 1.0` in the Megatron-Bridge DSv4 config mapping. Independent of #5412, which fixes a separate in-place aliasing bug.

### Pre-checks


hxbai (approved):

ShauryaaSharma:
Hey team, its been 2 weeks since the PR is open, Kindly approve the rest of the workflow so that if any errors occur and code change is required, i can proceed with the same. Also, the code has been reviews and approved by: @hxbai 
Thank you

guihong-nv:
/ok to test [3ede266](https://github.com/NVIDIA/Megatron-LM/pull/5497/commits/3ede2666a5b2ef7b56f7c23a2acec3c67d9df7b5)

megatron/core/fusions/fused_mla_yarn_rope_apply.py
```diff
@@ -253,6 +253,11 @@ def forward(
         assert sin.is_contiguous()
         assert headdim == qk_head_dim + emb_dim
         assert emb_dim % 4 == 0
+        assert cos.shape[-1] == emb_dim and sin.shape[-1] == emb_dim, (
```
guihong-nv: Because this is an `assert`, CPython removes the entire check under `python -O` / `PYTHONOPTIMIZE=1`. In that mode an undersized cache still reaches Triton and can read past the buffer, so the safety guarantee this PR is adding disappears. Please make both the Q and KV checks unconditional (for example, `if ...: raise ValueError(...)`) and update the tests to expect that exception; ideally add an optimized-mode regression so this cannot slip back.

guihong-nv (changes requested):
One blocking safety issue: the new width guard must remain active in optimized Python; details inline.

megatron/core/fusions/fused_mla_yarn_rope_apply.py
ShauryaaSharma: Fixed it just now. In `430e64fb7`: both the Q and KV checks are now a plain `if`/`raise ValueError` instead of `assert`, so they run regardless of optimization mode.

```python
if cos.shape[-1] != emb_dim or sin.shape[-1] != emb_dim:
    raise ValueError(
        f"cos/sin last dim must equal emb_dim={emb_dim} "
        f"(got cos={cos.shape[-1]}, sin={sin.shape[-1]}); a narrower rotary cache "
        f"(e.g. rotary_percent < 1) makes the fused MLA kernel read past the buffer."
    )
```

Also updated the two unit tests to expect `ValueError` instead of `AssertionError` so they still pass.

Before pushing I double-checked the behavior directly (mismatched width raises, matching width passes through fine, and the asymmetric case where only one of cos/sin is narrow still raises) under both normal and `-O` execution, confirmed the old `assert` version really did let a narrow cache through silently under `-O`, and the new version catches it in both modes.

ShauryaaSharma:
Fixed this now. In `430e64fb7`: both the Q and KV checks are now a plain `if`/`raise ValueError` instead of `assert`, so they run regardless of optimization mode.

```python
if cos.shape[-1] != emb_dim or sin.shape[-1] != emb_dim:
    raise ValueError(
        f"cos/sin last dim must equal emb_dim={emb_dim} "
        f"(got cos={cos.shape[-1]}, sin={sin.shape[-1]}); a narrower rotary cache "
        f"(e.g. rotary_percent < 1) makes the fused MLA kernel read past the buffer."
    )
```

Also updated the two unit tests to expect `ValueError` instead of `AssertionError` so they still pass.

Before pushing I double-checked the behavior directly (mismatched width raises, matching width passes through fine, and the asymmetric case where only one of cos/sin is narrow still raises) under both normal and `-O` execution — confirmed the old `assert` version really did let a narrow cache through silently under `-O`, and the new version catches it in both modes.

PR #5516, opened 2026-06-26
[training migration] Finish ModelBuilder integration

maanug-nv:

# What does this PR do ?
Replaces `model_builder` function pointer with ModelBuilder and calls the correct methods in place of `get_model`. Fixes necessary unit tests.

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

### DIFF
--- a/examples/multimodal/train.py
+++ b/examples/multimodal/train.py
@@ -385,9 +385,9 @@ if __name__ == "__main__":
 
     pretrain(
         train_valid_test_dataloaders_provider,
-        model_provider,
         ModelType.encoder_or_decoder,
         forward_step,
+        model_provider,
         args_defaults={'tokenizer_type': 'GPT2BPETokenizer'},
         extra_args_provider=add_multimodal_extra_args,
         process_non_loss_data_func=write_online_eval_to_tensorboard,
--- a/examples/multimodal/train.py
+++ b/examples/multimodal/train.py
@@ -5,6 +5,7 @@ import os
 import sys
 from functools import partial
 
+from megatron.training.arguments import parse_and_validate_args
 import torch
 import yaml
 
@@ -383,13 +384,16 @@ if __name__ == "__main__":
 
     train_valid_test_dataloaders_provider.is_distributed = True
 
+    args = parse_and_validate_args(
+        extra_args_provider=add_multimodal_extra_args,
+        args_defaults={'tokenizer_type': 'GPT2BPETokenizer'},
+    )
+    full_config = pretrain_cfg_container_from_args(args)
     pretrain(
         train_valid_test_dataloaders_provider,
         ModelType.encoder_or_decoder,
         forward_step,
         model_provider,
-        args_defaults={'tokenizer_type': 'GPT2BPETokenizer'},
-        extra_args_provider=add_multimodal_extra_args,
         process_non_loss_data_func=write_online_eval_to_tensorboard,
         get_embedding_ranks=llava_embedding_ranks,
         get_position_embedding_ranks=llava_position_embedding_ranks,
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -1004,9 +1004,9 @@ def preprocess_common_state_dict(common_state_dict):
 def pretrain(
     cfg_container: PretrainConfigContainer,
     train_valid_test_dataset_provider,
-    model_provider,
     model_type,
     forward_step_func,
+    model_provider=None,
     process_non_loss_data_func=None,
     get_embedding_ranks=None,
     get_position_embedding_ranks=None,
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -1223,7 +1223,11 @@ def pretrain(
     # Model, optimizer, and learning rate.
     timers('model-and-optimizer-setup', log_level=0).start(barrier=True)
     model, optimizer, opt_param_scheduler = setup_model_and_optimizer(
-        model_provider, model_type, checkpointing_context=checkpointing_context
+        model_provider,
+        model_type,
+        checkpointing_context=checkpointing_context,
+        cfg_container=cfg_container,
+        pg_collection=pg_collection,
     )
 
     timers('model-and-optimizer-setup').stop()
@@ -1995,7 +1999,9 @@ def setup_model_and_optimizer(
     model_provider_func,
     model_type,
     checkpointing_context=None,
-    pg_collection=None,
+    *,
+    cfg_container: PretrainConfigContainer,
+    pg_collection: ProcessGroupCollection,
 ):
     """Setup model and optimizer."""
     args = get_args()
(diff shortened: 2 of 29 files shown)


megatron/training/training.py
```diff
@@ -2006,9 +2014,27 @@ def setup_model_and_optimizer(
...
-    )
+
+    def _build_model_wrapper(wrap_with_ddp: bool):
+        from megatron.training.utils import start_memory_history_recording
+
+        start_memory_history_recording(cfg_container.profiling)
+
+        cfg = cfg_container
+        model_config = cfg.model
+        builder_cls = model_config.get_builder_cls()
```
yaoyu-33: `model_provider_func` is now unused, but the callers migrated in this PR (BERT, MIMO, T5, and VLM) still call `pretrain_cfg_container_from_args(args)` without a model config and pass their provider here. That helper deliberately leaves `cfg.model` as `None`, so each of those entry points now fails at first model setup with `AttributeError: NoneType has no attribute get_builder_cls`. Please either retain the provider fallback for configs without a builder or migrate those models to concrete ModelConfigs/builders before ignoring the provider.

examples/multimodal/train.py
```diff
@@ -385,9 +385,9 @@ def write_online_eval_to_tensorboard(data, iteration, writer, walltime=None):
 
     pretrain(
         train_valid_test_dataloaders_provider,
-        model_provider,
         ModelType.encoder_or_decoder,
         forward_step,
+        model_provider,
```
yaoyu-33, line 396: This entry point still cannot call the current `pretrain` API: the required first argument is a `PretrainConfigContainer`, and `args_defaults`/`extra_args_provider` are no longer accepted keyword arguments. As written, multimodal training raises `TypeError` before initialization. Please parse/validate the multimodal args, construct an appropriate model/config container, and then call `pretrain` with that container first.

megatron/training/training.py
```diff
@@ -2081,7 +2107,7 @@ def setup_model_and_optimizer(
         args.ffn_hidden_size = moe_ffn_hidden_size * args.moe_upcycling_granularity
 
         # get dense model
-        dense_model_for_upcycling = get_model(model_provider_func, model_type)
+        dense_model_for_upcycling = _build_model_wrapper(wrap_with_ddp=True)
```
yaoyu-33, line 2114: The temporary mutations above no longer affect model construction: `_build_model_wrapper` always reuses `cfg_container.model`, whose transformer config was created earlier with the MoE `num_experts` and FFN size. This therefore builds a second MoE model instead of the dense source model that `load_and_upcycle_model` expects. Please derive/clone a dense ModelConfig after applying these overrides (and build from that), rather than reusing the original MoE config.

megatron/training/training.py
```diff
@@ -2006,9 +2014,27 @@ def setup_model_and_optimizer(
...
+            pg_collection=pg_collection,
+            ddp_config=cfg.ddp,
+            overlap_param_gather_with_optimizer_step=cfg.optimizer.overlap_param_gather_with_optimizer_step,
+            use_megatron_fsdp=cfg.dist.use_megatron_fsdp,
+            use_torch_fsdp2=cfg.dist.use_torch_fsdp2,
+            wrap_with_ddp=wrap_with_ddp,
+            data_parallel_random_init=cfg.rng.data_parallel_random_init,
+        )
+
+    model = _build_model_wrapper(wrap_with_ddp)
```
yaoyu-33, line 2041: The old `get_model` path detected ModelOpt state in `args.load` (or `export_kd_teacher_load`) and set `args.modelopt_enabled` before selecting the provider/spec. This direct builder path skips that detection, and `gpt_config_from_args`/`hybrid_config_from_args` never set `restore_modelopt_state`, so loading a ModelOpt checkpoint now constructs the ordinary model/spec and also takes the ordinary loss path. Please preserve the detection and propagate it into the ModelConfig before resolving the builder.

megatron/training/training.py
```diff
@@ -1094,6 +1094,8 @@ def pretrain(
         seed_ep_group=getattr(init_pg_collection, "ep", None),
         seed_etp_group=getattr(init_pg_collection, "expt_tp", None),
     )
+    # TODO (@maanug): temporary until initialize.py is refactored to build pgcollection as bridge does
+    pg_collection = ProcessGroupCollection.use_mpu_process_groups()
```
yaoyu-33, line 1098: `get_model` used to honor `--create-all-gather-group` here by creating the separate DP/EP all-gather communicators and assigning `dp_cp_ag`/`expt_dp_ag` on this collection. `use_mpu_process_groups()` initializes both fields to `None`, and the new builder path never creates them, so Megatron FSDP silently loses the requested AG/RS overlap communicators. Please preserve that setup when materializing this collection.

megatron/training/utils/utils.py
```diff
@@ -0,0 +1,55 @@
...
+    Without this call, dumped snapshots contain only the current live
+    allocations — no timeline, no call sites.
+
+    Must be invoked before model construction so every tensor allocation is
+    captured. Guarded by ``profile_ranks`` so only ranks that will dump a
+    snapshot pay the recording overhead.
+    """
+    if profiling is None or not profiling.record_memory_history:
+        return
+    if safe_get_rank() not in profiling.profile_ranks:
```
yaoyu-33: An empty `profile_ranks` is the default and means “all ranks” in the existing profiling checks (`len(args.profile_ranks) == 0 or rank in ...`). This condition instead returns on every rank, so `--record-memory-history` with default rank settings never enables history and the later snapshot contains no allocation timeline. Please preserve the empty-list semantics (and align the selected rank with the existing last-rank snapshot dump).

megatron/training/training.py
```diff
@@ -2006,9 +2014,27 @@ def setup_model_and_optimizer(
...
+        builder = builder_cls(model_config)
+        return builder.build_distributed_models(
+            pg_collection=pg_collection,
+            ddp_config=cfg.ddp,
+            overlap_param_gather_with_optimizer_step=cfg.optimizer.overlap_param_gather_with_optimizer_step,
+            use_megatron_fsdp=cfg.dist.use_megatron_fsdp,
+            use_torch_fsdp2=cfg.dist.use_torch_fsdp2,
+            wrap_with_ddp=wrap_with_ddp,
+            data_parallel_random_init=cfg.rng.data_parallel_random_init,
+        )
```
yaoyu-33: The builder path also drops the `get_model` handling for `--freeze-all-layers`. That option was added for logits-saving workflows and must call `requires_grad_(False)` (plus set router `frozen_expert_bias`) before DDP/optimizer construction. With this return path every parameter remains trainable, so a supposedly frozen logits-generation run can update the model. Please carry the freeze operation into a pre-wrap hook or otherwise apply it before wrapping.

megatron/training/training.py
```diff
@@ -1004,9 +1004,9 @@ def reorder_inner_param_groups(optimizer_state_dict):
 def pretrain(
     cfg_container: PretrainConfigContainer,
     train_valid_test_dataset_provider,
-    model_provider,
     model_type,
     forward_step_func,
+    model_provider=None,
```
yaoyu-33, line 1009: There are still callers using the old positional order. In particular, `examples/post_training/modelopt/finetune.py` and `megatron/elastification/pretrain_hybrid_flex.py` pass `model_provider` as argument 3, so after this signature change it is interpreted as `model_type` (and the enum is interpreted as `forward_step_func`). Please migrate all repository `pretrain(...)` callers in the same change; these paths are not covered by the current green CI selection.

megatron/training/training.py
```diff
@@ -1993,7 +1999,9 @@ def setup_model_and_optimizer(
     model_provider_func,
     model_type,
     checkpointing_context=None,
-    pg_collection=None,
+    *,
+    cfg_container: PretrainConfigContainer,
```
yaoyu-33: These new required keyword-only arguments also leave direct callers unmigrated. `tasks/finetune_utils.py` and `tools/bert_embedding/embed.py` still call this function with only provider/type, as do `test_fp4_param.py`, `test_cuda_graphs.py`, and several MTP tests. They now raise `TypeError` before model construction. Please update all direct callers or retain a compatible fallback/default while the migration is incomplete.

megatron/training/training.py
maanug-nv: should work now if the get_model fallback is used

megatron/training/training.py
maanug-nv: should work now if the get_model fallback is used

megatron/training/training.py
maanug-nv: should be preserved if the get_model fallback is used

megatron/training/training.py
maanug-nv: should be preserved if the get_model fallback is used

megatron/training/training.py
maanug-nv: updated in 073053a72f612618d38f13f34837c96c7ba2d129

megatron/training/training.py
maanug-nv: updated in 073053a72f612618d38f13f34837c96c7ba2d129

megatron/training/training.py
maanug-nv: resolved in 073053a72f612618d38f13f34837c96c7ba2d129

examples/multimodal/train.py
maanug-nv: resolved in af7f1b60f96c0653dc442c76ae04313929788622

megatron/training/utils/utils.py
maanug-nv: resolved in 0b7c42ce85f7438f56d14081bc6ba7a837b99f86

yaoyu-33 (approved):

deepakn94 (approved):

tdene (approved):

PR #5530, opened 2026-06-28
fix(clip_grads): handle empty grads_for_norm in inf-norm and p-norm paths

### LINKED ISSUE #5529: `get_grad_norm_fp32` crashes on empty gradient list with `inf` norm or custom p-norm
**Describe the bug**

`get_grad_norm_fp32` in `megatron/core/optimizer/clip_grads.py` has two
separate crash paths when `grads_for_norm` is an empty list. This is a
realistic condition whenever all parameters on a given rank are frozen,
shared, or filtered out by `_filter_grads_for_norm`.

@NVIDIA/mcore-oncall

**Bug 1 : `inf` norm: `ValueError` from empty generator**

Line 95:
```python
total_norm = max(grad.abs().max() for grad in grads_for_norm)
```
When `grads_for_norm` is empty this evaluates `max()` over an empty
generator, which raises:
```
ValueError: max() arg is an empty sequence
```

**Bug 2 : custom p-norm (`norm_type != 2.0 and != inf`): `TypeError` in `all_reduce`**

Lines 127–129 leave `total_norm` as a Python `float 0.0` when the loop
body never executes:
```python
for grad in grads_for_norm:           # never runs
    grad_norm = torch.norm(grad, norm_type)
    total_norm += grad_norm**norm_type  # total_norm stays float 0.0
```
Then line 133:
```python
torch.distributed.all_reduce(total_norm, ...)   # TypeError: expects Tensor
```

Note: the L2 path (lines 113–122) already handles the empty case
correctly with an explicit `if grads_for_norm:` guard and a
`torch.zeros(1, …, device='cuda')` fallback. The other two paths need
the same treatment.

Minimal non-distributed reproducer (no GPU required for the crash itself):

```python
from torch import inf
from megatron.core.optimizer.clip_grads import get_grad_norm_fp32

# Bug 1
get_grad_norm_fp32([], norm_type=inf)
# ValueError: max() arg is an empty sequence

# Bug 2
get_grad_norm_fp32([], norm_type=1.0)
# TypeError: argument 'tensor' must be Tensor, not float
```

In a real training run the empty list arises when all parameters on a
rank are filtered by `param_is_not_shared` / `param_is_not_tensor_parallel_duplicate`
inside `_filter_grads_for_norm`.

**Expected behavior**

`get_grad_norm_fp32([])` should return `0.0` for any `norm_type`,
consistent with the L2 path's existing behaviour and with
`torch.nn.utils.clip_grad_norm_`'s behaviour on an empty parameter list.

**Additional context**

- Affected file: `megatron/core/optimizer/clip_grads.py`
- Bug 1: line 95 (`inf` branch, `max()` over empty generator)
- Bug 2: lines 127–136 (generic p-norm branch, `float` passed to `all_reduce`)
- The L2 path (lines 113–122) already has the correct pattern; the fix
  is to mirror it to the other two branches.
- `test_clip_grads.py` contains only one trivial config test and has no
  coverage for any of the three norm-type paths.

Mattral:
## Summary

`get_grad_norm_fp32` crashes when called with an empty gradient list,
which occurs in practice when all parameters on a rank are filtered out
(frozen layers, shared params, TP duplicates). Two out of three norm-type
branches are affected; the L2 branch already has a correct guard.

## Changes

**`megatron/core/optimizer/clip_grads.py`**

| Branch | Bug | Fix |
|---|---|---|
| `inf` norm (line 95) | `max()` over empty generator → `ValueError` | Add `default=torch.tensor(0.0)` to `max()` |
| generic p-norm (line 127) | `total_norm` stays `float 0.0`; `all_reduce` receives a non-Tensor → `TypeError` | Initialise `total_norm = torch.zeros(1, dtype=torch.float, device='cuda')` before loop |

No change to the L2 path or any non-empty-list behaviour.

## Testing

- Added unit tests in `tests/unit_tests/optimizer/test_clip_grads.py`
  covering empty-list calls for all three `norm_type` paths (inf, 2.0,
  custom p). Tests run without GPU using the local fallback
  implementations.
- Existing CI passes unchanged.

## Checklist
- [x] Commits signed off (`git commit -s`)
- [x] Rebased on `main`
- [x] No unrelated formatting changes
- [x] One logical change per commit

Fixes #5529

### DIFF
--- a/megatron/core/optimizer/clip_grads.py
+++ b/megatron/core/optimizer/clip_grads.py
@@ -92,7 +92,7 @@ def get_grad_norm_fp32(
 
     # Calculate norm.
     if norm_type == inf:
-        total_norm = max(grad.abs().max() for grad in grads_for_norm)
+        total_norm = max((grad.abs().max() for grad in grads_for_norm), default=torch.tensor(0.0))
         total_norm_cuda = torch.tensor([float(total_norm)], dtype=torch.float, device='cuda')
         # Take max across all data-parallel GPUs if using FSDP and then all model-parallel GPUs.
         if data_parallel_group:
@@ -105,24 +105,20 @@ def get_grad_norm_fp32(
         total_norm = total_norm_cuda[0].item()
 
     else:
-        if norm_type == 2.0:
+        total_norm = torch.zeros(1, dtype=torch.float, device='cuda')
+        if not grads_for_norm:
+            pass
+        elif norm_type == 2.0:
             dummy_overflow_buf = torch.zeros(1, dtype=torch.int, device='cuda')
             # Use apex's multi-tensor applier for efficiency reasons.
             # Multi-tensor applier takes a function and a list of list
             # and performs the operation on that list all in one kernel.
-            if grads_for_norm:
-                grad_norm, _ = multi_tensor_applier(
-                    l2_norm_impl,
-                    dummy_overflow_buf,
-                    [grads_for_norm],
-                    False,  # no per-parameter norm
-                )
-            else:
-                grad_norm = torch.zeros(1, dtype=torch.float, device='cuda')
+            grad_norm, _ = multi_tensor_applier(
+                l2_norm_impl, dummy_overflow_buf, [grads_for_norm], False  # no per-parameter norm
+            )
             # Since we will be summing across data parallel groups,
             # we need the pow(norm-type).
             total_norm = grad_norm**norm_type
-
         else:
             for grad in grads_for_norm:
                 grad_norm = torch.norm(grad, norm_type)


Mattral (commented):
looking forward to have this reviewed

janEbert (commented):
We can simplify a bit in the `norm_type != inf` code path, since both `norm_type == 2.0` and `norm_type != 2.0` need the same zero-tensor if `grads_for_norm` is an empty list. `norm_type == 2.0` currently solves it with the `grad_norm`, which we could remove if we initialized the `total_norm` zero-tensor earlier. The entire `norm_type == 2.0` path could then be simplified around that.

E.g.:

```python
total_norm = torch.zeros(1, dtype=torch.float, device='cuda')
if not grads_for_norm:
    pass
elif norm_type == 2.0:
    dummy_overflow_buf = torch.zeros(1, dtype=torch.int, device='cuda')
    # Use apex's multi-tensor applier for efficiency reasons.
    # Multi-tensor applier takes a function and a list of list
    # and performs the operation on that list all in one kernel.
    grad_norm, _ = multi_tensor_applier(
        l2_norm_impl,
        dummy_overflow_buf,
        [grads_for_norm],
        False,  # no per-parameter norm
    )
    # Since we will be summing across data parallel groups,
    # we need the pow(norm-type).
    total_norm = grad_norm**norm_type

else:
    for grad in grads_for_norm:
        grad_norm = torch.norm(grad, norm_type)
        total_norm += grad_norm**norm_type
```

Mattral:
Thanks for the suggestion 

that's a cleaner structure. I'll hoist `total_norm = torch.zeros(1, dtype=torch.float, device='cuda')` before the `norm_type` branches so both `norm_type == 2.0` and the generic p-norm path share the same zero-init, and drop the redundant `else: grad_norm = torch.zeros(...)` guard from the L2 path. Will push an updated commit shortly.

janEbert (approved):
Thank you for the contribution and quick adjustment!

Mattral (commented):
@janEbert 
Thanks for the review! Updated the commit per your suggestion. Hoisted the zero-tensor init before both branches and dropped the redundant L2 guard. Also fixed the DCO sign-off. LMK if anything else needs changing.

janEbert:
You'd quickly need to run `bash tools/autoformat.sh`.

Mattral:
@janEbert 
Done, ran `bash tools/autoformat.sh` and amended the commit. Thank you for your patience and following up

Mattral (commented):
The autoformatter fix has been pushed. CI is now waiting for approvals from `@NVIDIA/core-adlr` and `@NVIDIA/core-nemo`.

jiemingz (approved):

deepakn94 (approved):

Phlip79 (approved):

janEbert:
Previous fail due to deterministic throughput being 1.29 slower than non-deterministic baseline. I doubt it's related to this PR, probably just flaky CI.

PR #5536, opened 2026-06-29
ci: add install check for megatron.inference imports

svcnemo-autobot:
Added a "Check imports for megatron.inference" step to the pip-test-pytorch job in .github/workflows/install-test.yml, mirroring the existing megatron.training check.


ko3n1g:
@svcnemo-autobot can we add this check _without_ updating the pyproject.toml?

ko3n1g:
/ok to test [59bc829](https://github.com/NVIDIA/Megatron-LM/pull/5536/commits/59bc829a5087b2b45c0c2f60c9b5611bbecaa04f)

ko3n1g:
@svcnemo-autobot `  /opt/venv/bin/python -c "import megatron.inference"
  shell: bash -x -e -u -o pipefail {0}` this is not what i meant. we want to use the existing CI framework to check all imports of `megatron.inference`, but not add megatron.inference to the pyproject.toml

ko3n1g:
@svcnemo-autobot how about we pip install megatron.training in .github/workflows/install-test.yml?

ko3n1g:
@svcnemo-autobot but we also need to add megatron.inference to the pythonpath i think

ko3n1g:
@svcnemo-autobot remove the megatron.training install again.. that was a mistake of mine

PR #5541, opened 2026-06-29
[codex] Preserve real and padded THD sequence lengths

xiaoyao0115:
# What does this PR do?

This PR fixes THD tail-padding metadata in two cases:

1. **Dummy sequence padding:** preserves existing gaps between valid and padded sequence boundaries when appending the dummy sequence.
2. **Non-dummy padding:** keeps `cu_seqlens` unchanged and extends `cu_seqlens_padded[-1]` to the actual padded tensor size.

For non-dummy padding, tensors are padded before context-parallel slicing to ensure the correct token distribution.

The default dummy-sequence behavior remains unchanged.

## Why?

Using physical padded boundaries as valid-token boundaries makes padding appear to be real data and prevents Transformer Engine from inferring gaps between sequences correctly.

## Impact

THD attention still receives physical storage boundaries, while routing and loss paths retain the real token boundaries. The default dummy-sequence behavior remains unchanged. The `extend_last` policy is opt-in.

The sequence-packing test now validates deterministic metadata locally on every rank and checks padding presence across the CP group. This avoids incorrectly comparing independent DP replicas or requiring every CP slice to contain padding.

## Validation

- `tests/unit_tests/test_sequence_packing.py`: 29 passed
- `tests/unit_tests/transformer/test_thd_cuda_graph.py`: 28 passed, 2 skipped pending the required TE change
- `tests/unit_tests/transformer/moe/test_token_dispatcher.py`: 19 passed, 92 skipped by configuration

### DIFF
--- a/megatron/core/model_parallel_config.py
+++ b/megatron/core/model_parallel_config.py
@@ -118,13 +118,21 @@ class ModelParallelConfig:
     tensors are padded to a multiple of N.
     """
 
-    pad_packed_seq_by_appending_dummy_seq: bool = True
-    """Represent a THD packed-sequence padding tail by appending a dummy sequence.
+    thd_tail_padding_policy: Optional[Literal["append_dummy_seq", "extend_last"]] = None
+    """Policy for representing the THD packed-sequence padding tail.
 
-    When disabled, token-like tensors are still padded according to
-    pad_packed_seq_alignment, but cu_seqlens sequence boundaries are not extended
-    for the padding tail. When thd_max_packed_sequences is set, static-input
-    padding may still pad cu_seqlens tensors to that value + 1 entries.
+    - append_dummy_seq: cover the post-pack padding tail with an ordinary
+      dummy sequence appended to the cu_seqlens metadata. Existing
+      valid/physical gaps between real sequences are preserved. This is the
+      default behavior.
+    - extend_last: keep valid cu_seqlens boundaries unchanged and extend the
+      final padded boundary so the tail is physical padding of the last
+      sequence. With context parallelism the extension is applied to the
+      global metadata before CP slicing.
+    - None (default): treated as append_dummy_seq.
+
+    When thd_max_packed_sequences is set, cu_seqlens tensors are padded to
+    that value + 1 entries in both eager and CUDA Graph modes.
     """
 
     expert_model_parallel_size: int = 1
@@ -513,6 +521,12 @@ class ModelParallelConfig:
                     f"min_dynamic_context_parallel_size must be >= 1, "
                     f"got {self.min_dynamic_context_parallel_size}"
                 )
+
+        if self.thd_tail_padding_policy not in (None, "append_dummy_seq", "extend_last"):
+            raise ValueError(
+                "thd_tail_padding_policy must be 'append_dummy_seq', 'extend_last', or None, "
+                f"got {self.thd_tail_padding_policy!r}."
+            )
 
         if self.pad_packed_seq_alignment is not None:
             self.pad_packed_seq_alignment = _parse_pad_packed_seq_alignment(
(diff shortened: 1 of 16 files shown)


HaochenYuan:
Suggestion: replace pad_packed_seq_by_appending_dummy_seq: bool with an explicit tail-policy enum:
```
--thd-tail-padding-policy {
    extend_last,
    zero_valid_dummy,
    local_hidden_only
}
# - extend_last (default): extend only cu_seqlens_padded[-1]; keep cu_seqlens unchanged.  
# - zero_valid_dummy: append the original real endpoint to cu_seqlens and the target endpoint to cu_seqlens_padded. This preserves the original sequence partition but consumes one sequence slot.  
# - local_hidden_only (experimental): pad local hidden after CP slicing without changing metadata. Restrict this to P2P and tail-safe implementations such as the DSV4 custom RoPE; reject A2A and generic TE fused THD RoPE.
```  
This avoids silently changing the meaning of the existing boolean option and makes the trade-offs explicit

megatron/core/packed_seq_params.py
```diff
@@ -439,16 +482,77 @@ def pad_sequence_for_thd(
...
+            cu_seqlens_kv = _append_dummy_seq(cu_seqlens_kv, global_target_len)
         cu_seqlens_q_padded = _append_dummy_seq(cu_seqlens_q_padded, global_target_len)
         cu_seqlens_kv_padded = _append_dummy_seq(cu_seqlens_kv_padded, global_target_len)
+    elif has_non_dummy_padding_tail:
+        # Keep compact valid-token coordinates unchanged. Only physical
+        # metadata grows, so the tail belongs to the final real sequence.
+        non_dummy_cp_size, _ = _resolve_thd_cp_geometry(
+            packed_seq_params, cp_group=cp_group, cp_size=cp_size, cp_rank=cp_rank
+        )
+        assert non_dummy_cp_size == 1, (
```
Victarry, line 621: **[IMPORTANT Correctness]** Handle the non-scheduler packed-SFT CP path before rejecting slice-first calls.

This assertion is reachable from `pretrain_gpt.py`: when `sequence_packing_scheduler` is `None`, `get_thd_batch_on_this_cp_rank()` partitions the THD tensors first, and `pad_sequence_for_thd(..., pad_by_appending_dummy_seq=False)` is called afterwards. With CP > 1 and any padding tail, this now always triggers the assertion. `--sft` does not automatically enable a packing scheduler, so this is a supported path.

Please apply the global non-dummy padding and metadata extension before `get_thd_batch_on_this_cp_rank()` as well, ideally through a helper shared with `data_schedule.py`, and add a CP regression test for the non-scheduler SFT path.

megatron/core/model_parallel_config.py
```diff
@@ -122,9 +122,10 @@ class ModelParallelConfig:
     """Represent a THD packed-sequence padding tail by appending a dummy sequence.
 
     When disabled, token-like tensors are still padded according to
-    pad_packed_seq_alignment, but cu_seqlens sequence boundaries are not extended
-    for the padding tail. CUDA Graph static-input padding may still pad the
-    cu_seqlens tensors to thd_max_packed_sequences + 1 entries.
+    pad_packed_seq_alignment. Valid cu_seqlens boundaries remain unchanged, while
```
Victarry: **[IMPORTANT Compatibility]** I agree with [@HaochenYuan's suggestion in the PR discussion](https://github.com/NVIDIA/Megatron-LM/pull/5541#issuecomment-4852805681) to replace this boolean with an explicit THD tail-padding policy.

This PR changes the existing `False` semantics from padding token-like tensors without extending sequence boundaries to extending `cu_seqlens_padded[-1]`. Existing callers using `False` would therefore silently receive different THD metadata and backend behavior. The proposed `extend_last`, `zero_valid_dummy`, and `local_hidden_only` policies make these materially different contracts explicit instead of overloading one boolean.

Please introduce a policy enum (for example, `--thd-tail-padding-policy`) and either preserve the legacy boolean behavior during migration or add an explicit compatibility/deprecation path.

HaochenYuan: After discussion with @xiaoyao0115, we think it's ok to replace `local_hidden_only` with `extend_last `, it's compatible with existing callers, but I agree with introducing a policy num for `extend_last` & `zero_valid_dummy` : ).

megatron/core/packed_seq_params.py
xiaoyao0115: i'll fix it

HaochenYuan (approved):

yaox12 (approved):

PR #5542, opened 2026-06-29
[codex] Exclude padding tokens from MoE routing

xiaoyao0115:
# What does this PR do?

This splits the MoE padding-routing fix out of #5535.

- Zeros router probabilities for padding tokens.
- Clears padding-token entries from the routing map before capacity enforcement and token dispatch.
- Covers both fused and non-fused router paths.

## Why?

Masking padding only in auxiliary-loss calculations still lets padding consume expert capacity, enter permutations, and change dispatcher and grouped-GEMM shapes. Padding must be removed before token dropping and dispatch.

## Impact

Padding tokens no longer affect auxiliary losses, expert capacity, permutation, dispatch counts, or grouped GEMM shapes.

## Validation

- `tests/unit_tests/transformer/moe/test_routers.py::TestTop2Router::test_router_with_padding_mask`: 2 passed


HaochenYuan:
The new router masking works as intended for dropless HybridEP, but it cannot currently be applied unconditionally because it breaks or bypasses assumptions in the other dropless dispatchers.

## Runtime validation

### HybridEP works as intended

With 4 ranks, EP=4, 64 physical tokens/rank, 48 valid tokens/rank, and top-k=2:

- The parent commit dispatched 512 routes globally, including 128 padding routes.
- This PR dispatched exactly `4 * 48 * 2 = 384` routes.
- Zero padding rows reached the expert input, and dispatch/combine reconstructed all valid tokens exactly.

This confirms that dropless HybridEP consumes the sparse `routing_map` correctly.

### Dropless AlltoAll is broken

With 4 ranks, EP=4, 16 physical tokens/rank, 8 valid tokens/rank, and top-k=2:

- The parent commit had `route_count=32`, `num_out_tokens=32`, and split sum 32; forward succeeded.
- This PR had `route_count=16`, but `num_out_tokens` remained 32 while the split sum was 16.
- Every rank failed in `all_to_all_single` with:

  ```text
  RuntimeError: Split sizes doesn't match total dim 0 size
  ```

After the new mask, `routing_map` has only `num_valid_tokens * topk` set bits, while the dropless AlltoAll fast path still assumes every physical row has `topk` routes and sets `num_out_tokens = routing_map.size(0) * topk`. The communication metadata is computed from the smaller actual route count, so the permutation buffer and A2A splits disagree.

### Dropless DeepEP still dispatches padding

With 2 ranks, two valid and two padding tokens per rank, and top-k=2, there should be eight valid routes globally. The actual DeepEP run produced 16 dispatch/GEMM rows: the extra eight routes were all padding. Padding source-token sentinels were present in the tensor passed to expert computation.

`_DeepepManager.setup_metadata()` reconstructs fixed-width indices with `topk(probs)` and masks zero-probability indices to `-1` only when `moe_expert_capacity_factor` is set. In the dropless case, each all-zero padding row therefore receives arbitrary valid expert IDs. DeepEPv2 inherits the same metadata method.

## Two possible ways forward

### Option 1: Make the change generic

Please update the affected dispatchers to support sparse routing maps:

1. For dropless AlltoAll, when padding makes routing sparse, size the permutation output consistently with the actual route count and use matching A2A split metadata. The existing static fast path can remain for unmasked batches.
2. For DeepEP/DeepEPv2, preserve the supplied `routing_map` when converting probabilities to indices. For example, gather the selected bits from `routing_map` and set indices whose gathered bit is false to `-1`, instead of conditioning this only on `moe_expert_capacity_factor`.
3. Add end-to-end MoE tests for padding with dropless AlltoAll, DeepEP, DeepEPv2, and HybridEP. The current router-only test does not exercise these dispatcher invariants.

### Option 2: Scope the fix to HybridEP

If this change is intended only for the HybridEP use case, guard only the new dispatch masking block:

```python
use_hybridep = (
    self.config.moe_token_dispatcher_type == "flex"
    and self.config.moe_flex_dispatcher_backend == "hybridep"
)

if padding_mask is not None and use_hybridep:
    valid_tokens = (~padding_mask).unsqueeze(-1)
    probs = probs * valid_tokens
    routing_map = routing_map & valid_tokens
```

The `padding_mask` itself should still be passed through for auxiliary loss, z-loss, and expert-bias handling. With this option, please narrow the PR title, description, and validation claims to dropless HybridEP.

## Capacity-mode caveat

The current implementation does not make capacity-based routing padding-independent:

- `moe_expert_capacity_factor` still computes capacity from the physical row count rather than the valid-token count, so padding changes the drop threshold.
- In drop-and-pad mode, `final_map = capacity_mask` fills the physical-token-derived capacity with zero-probability routes. On both AlltoAll and HybridEP, runtime tests confirmed that these fillers consume communication and expert-GEMM rows; depending on zero-score tie ordering, the fillers can be the original padding rows.
- DeepEP/DeepEPv2 explicitly reject true drop-and-pad (`moe_pad_expert_input_to_capacity=True`); setting only `moe_expert_capacity_factor` is drop-only.
- HybridEP's `moe_expert_rank_capacity_factor` budget is also based on physical rows. With identical 128 valid tokens, adding 64 padding tokens doubled the measured static budget and expert buffer from 256 to 512 rows/rank and changed whether overflow/drop occurred.

If the intended scope is only dropless HybridEP, these modes can be excluded explicitly. Otherwise, the capacity calculation and drop-and-pad behavior also need to be made padding-aware.

HaochenYuan (approved):
LGTM

yaox12 (approved):

PR #5543, opened 2026-06-29
[codex] Restore per-microbatch MTP loss logging

xiaoyao0115:
# What does this PR do?

This splits the MTP logging restoration out of #5535.

- Normalizes each local MTP loss contribution by its microbatch token count before accumulation.
- Reduces accumulated normalized losses while averaging across the configured average group.
- Scales logging by the scheduled microbatch count.
- Removes the global token-weighted `sum(loss) / sum(tokens)` tracker state.

## Why?

MTP logging historically accumulated normalized per-microbatch losses. Globally weighting all microbatches by token count changes the metric when packed microbatches have different token counts.

## Impact

Logged MTP loss again represents the average of normalized microbatch contributions. Existing runs may see a metric-definition discontinuity at the upgrade boundary when microbatch token counts differ.

## Validation

- `tests/unit_tests/transformer/test_multi_token_prediction.py::TestMTPLossLoggingHelper`: 6 passed


Victarry (approved):

liuzhenhai93:
make per-microbatch/per-token mtp loss logging as an option may be better

PR #5551, opened 2026-06-29
Eliminate the ShardedObject all_gather_object on the FullyParallel load path

asolergi-nv:

# What does this PR do ?

Remove the world `all_gather_object` that exchanges `ShardedObject`s on the FullyParallel **load** path: every rank loads its objects directly (content-addressable by `unique_key`) fused into a single `base_strategy.load`.
 
## Issue tracking

Linked issue: 

### Pre-checks

---

# FullyParallel load: load ShardedObjects per-rank, drop the object `all_gather_object`

## Problem

On the fully-parallel checkpoint **load** path
(`FullyParallelLoadStrategyWrapper.load`), `ShardedObject`s (RNG state,
`_extra_state`, other small artifacts) were loaded only by their main replica
and then exchanged across **all** ranks with a WORLD-wide
`exchange_loaded_objects_gather_object` (an `all_gather_object`). This collective
scales with world size and serializes/transfers every object to every rank, even
though objects are tiny and each rank could simply read what it needs directly.
It is a recurring, latency-sensitive cost at the start of every load (and of
every resume), with two separate base-strategy `.load()` calls (objects, then
tensors) each incurring their own metadata read and load plan.

## Solution

A `ShardedObject` is addressed in the checkpoint by its `unique_key`
(`key + global_offset + global_shape`), which **excludes `replica_id`**. Each
distinct object position is therefore written to disk exactly once and can be
read locally by any rank with `no_dist=True`. So instead of exchanging objects:

- Every rank loads **all** of its objects directly from storage by merging the
  main (`to_load_objects`) and non-main (`unloaded_objects`) replica maps
  (`all_objects_to_load`). The WORLD `all_gather_object` for objects — and its
  trailing `torch.cuda.synchronize()` and missing-shard check — are removed.
- Object loads are **fused with the tensor-shard load** into a single
  `base_strategy.load({**to_load_shards, **all_objects_to_load})` call (one
  metadata read + one load plan instead of two). The results are split back into
  tensors and objects by membership in the original maps (tensor and object
  shard ids never collide).

`ShardedTensor`s are unchanged: they keep the fully-parallel distribution (each
rank reads only its assigned shards and receives the rest via
`exchange_by_distribution` in step 4).

Net effect: **one fewer world collective per load** and one fewer base-strategy
load, traded for cheap extra local reads of small object artifacts. No change to
what ends up in the loaded state dict.

## Files changed

- `megatron/core/dist_checkpointing/strategies/fully_parallel.py`
  - Removed the `exchange_loaded_objects_gather_object` import and its use.
  - Reworked `FullyParallelLoadStrategyWrapper.load` to load all objects
    per-rank and fuse the tensor+object base-strategy load.

## Testing notes

- Backward compatible: loads checkpoints written by the existing save path
  (object on-disk layout is unchanged; this only changes how objects are read
  back). Verified that a checkpoint saved before this change loads correctly.
- Covered by the existing fully-parallel load tests
  (`tests/unit_tests/dist_checkpointing/test_fully_parallel.py`); object content
  (RNG/`_extra_state`) matches the pre-change loader on every rank.

### DIFF
--- a/megatron/core/dist_checkpointing/strategies/fully_parallel.py
+++ b/megatron/core/dist_checkpointing/strategies/fully_parallel.py
@@ -178,6 +178,7 @@ class FullyParallelLoadStrategyWrapper:
         parallelization_group: Optional[torch.distributed.ProcessGroup] = None,
         do_cache_distribution: bool = False,
         exchange_algo: str = 'broadcast',
+        per_rank_object_load: bool = False,
     ):
         self.base_strategy = strategy
         if parallelization_group is None:
@@ -187,6 +188,11 @@ class FullyParallelLoadStrategyWrapper:
         self.parallelization_group = parallelization_group
         self.do_cache_distribution = do_cache_distribution
         self.exchange_algo = exchange_algo
+        # When True, every rank loads *all* of its ShardedObjects directly from
+        # storage instead of loading only its main replicas and exchanging them
+        # with a WORLD-wide `all_gather_object`. Opt-in; defaults to the legacy
+        # gather-based exchange. See `load` for correctness rationale.
+        self.per_rank_object_load = per_rank_object_load
 
         self.cached_distribution: Optional[ShardDistribution] = None
         self.cached_global_metadata: Optional[Metadata] = None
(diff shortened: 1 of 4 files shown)


deepakn94 (approved):

Phlip79 (approved):

yaoyu-33 (approved):

asolergi-nv:
/ok to test [519338b](https://github.com/NVIDIA/Megatron-LM/pull/5551/commits/519338b196cd93505d1c5af4f90c521929d2b622)

dimapihtar (changes requested):
Is it possible to add some unit tests for newly added logic? Thanks.

asolergi-nv:
> Is it possible to add some unit tests for newly added logic? Thanks.

Before merging this PR, let's wait for https://github.com/NVIDIA/Megatron-LM/pull/6347 & include this feature in the new functional test which is loading a checkpoint. RNG states are ShardedObjects, so it should arise any issue. WDYT?

dimapihtar (approved):

deepakn94:
/ok to test [284ced6](https://github.com/NVIDIA/Megatron-LM/pull/5551/commits/284ced611717bd3b5b5b0620c37c785fbd9083e3)

PR #5552, opened 2026-06-29
Add --ckpt-drop-redundant-extra-state to skip persisting redundant TE _extra_state

asolergi-nv:

# What does this PR do ?

Add an opt-in `--drop-redundant-extra-state` flag that stops persisting TE `_extra_state` artifacts that carry no irreplaceable state (everything except delayed-scaling FP8).
 
## Issue tracking

Linked issue: 

### Pre-checks

---

# Don't persist redundant TransformerEngine `_extra_state` in distributed checkpoints

## Problem

Every TransformerEngine module (`Linear`, `LayerNormLinear`, `DotProductAttention`,
grouped/MoE experts, fused ops, …) contributes an `_extra_state` entry to the
checkpoint via `TransformerEngineBaseModule.get_extra_state`. In a large model
these dominate the checkpoint's object count — on a 55B hybrid-MoE run (512
experts) they are on the order of ~50k `ShardedObject`s, each saved (and its
metadata gathered/validated) on every checkpoint.

But that payload is only meaningful in one case. `get_extra_state` writes real
state **only under `if recipe.delayed():`** — the FP8 delayed-scaling amax
history and scale factors, which evolve during training and cannot be
reconstructed from config. In every other configuration the `_extra_state` is
either:

* **No FP8** → `None` / an empty `uint8` tensor, or
* **Block scaling (NVFP4, MXFP8) / current scaling (Float8CurrentScaling)** →
  a recipe + scalar blob that the freshly-built model reproduces from the run
  config at build time.

So unless you train with delayed-scaling FP8, these objects are pure overhead:
they bloat the checkpoint, inflate `.metadata`, and add tens of thousands of
items to the save-side metadata gathers and sharding validation — for data that
is either empty or trivially reconstructable. (Checkpoints are typically stored
in BF16 even when compute uses FP8/FP4, so this is the common case.)

## Solution

Add an opt-in `--drop-redundant-extra-state` flag (**off by default**, so the
prior behavior — persisting all `_extra_state` — is preserved) that stops
persisting the redundant `_extra_state` while keeping the genuinely useful
delayed-scaling ones, with no change to the load contract.

When enabled, in `generate_state_dict` (the shared chokepoint for both the save
state dict and the load *request*), for `torch_dist` checkpoints, every
non-persistent `_extra_state` `ShardedObject` is rewritten into a
`LocalNonpersistentObject`:

* **SAVE**: `save_preprocess` drops `LocalNonpersistentObject`s, so they are
  never written — fewer objects, smaller `.metadata`, lighter metadata gather
  and sharding validation.
* **LOAD**: `load_preprocess` unwraps them back into the loaded state dict with
  the local (freshly-built) value, so the module's `_extra_state` key is still
  present (strict `load_state_dict` stays satisfied) and `set_extra_state`
  no-ops on the empty/recipe payload.

Persistence is decided by `_fp8_extra_state_is_persistent`, which inspects the
**raw serialized bytes** of the `uint8` payload for the delayed-scaling key
markers (`amax_history_fwd` / `scale_fwd` / `scale_bwd`) — no unpickling, no TE
import. Anything unrecognized defaults to *persistent*, so real state is never
dropped silently. Delayed-scaling `_extra_state` remains a `ShardedObject` and
is saved/loaded exactly as before.

## Backward compatibility

* **Old checkpoints load unchanged.** Because the decision runs in
  `generate_state_dict`, the dropped keys are never *requested* at load time, so
  an older checkpoint that still stores those `_extra_state`s simply does not
  read them — they are neither "missing" nor "unexpected".
* **New checkpoints load on old code** the same way an empty `_extra_state`
  always has (`set_extra_state` no-ops).
* **Delayed-scaling FP8 runs are unaffected** — their `_extra_state` is detected
  as persistent and kept.

## Usage

The optimization is **opt-in**; by default all `_extra_state` are persisted, so
existing behavior is unchanged. Enable it to drop the redundant ones:

```
--drop-redundant-extra-state
```

## Files changed

* `megatron/training/checkpointing.py` — `_fp8_extra_state_is_persistent`,
  `_localize_redundant_extra_states`, and the gated call in `generate_state_dict`.
* `megatron/training/arguments.py` — `--drop-redundant-extra-state` flag.
* `tests/unit_tests/dist_checkpointing/test_drop_redundant_extra_state.py` — new
  unit tests (below).

## Testing notes

* **Unit tests** (`test_drop_redundant_extra_state.py`, no GPU/FP8 required, so
  they run identically on FP8-capable CI and elsewhere): exercise
  `_fp8_extra_state_is_persistent` on TE-format payloads — empty / `None` /
  recipe-only (block & current scaling) classify as non-persistent; the
  delayed-scaling markers (`amax_history_fwd` / `scale_fwd` / `scale_bwd`)
  classify as persistent — and `_localize_redundant_extra_states`, which must
  wrap only the redundant `_extra_state` objects as `LocalNonpersistentObject`
  while leaving delayed-scaling objects, non-`_extra_state` objects, and tensors
  untouched.
* No-FP8 / BF16 run: confirm a checkpoint saved with `--drop-redundant-extra-state`
  contains no `*_extra_state` data objects and the `.metadata` object count drops
  by the expected ~50k; train→save→load→resume is bit-exact vs the default
  (persist-all) baseline.
* Delayed-scaling FP8 run (H100 / GB200): confirm `_extra_state` is still
  persisted with `--drop-redundant-extra-state` and that amax history/scale
  survive a save/load round-trip.
* Cross-version: load a pre-change checkpoint with the new code (and vice versa)
  and verify resume works.

### DIFF
--- a/megatron/training/arguments.py
+++ b/megatron/training/arguments.py
@@ -2893,6 +2893,17 @@ def _add_checkpointing_args(parser):
     group.add_argument('--ckpt-fully-parallel-save', action='store_true',
                        dest='ckpt_fully_parallel_save_deprecated',
                        help='Deprecated: see --no-ckpt-fully-parallel-save.')
+    group.add_argument('--drop-redundant-extra-state', action='store_true',
+                       default=False,
+                       help='Drop TE `_extra_state` artifacts that carry no '
+                       'irreplaceable state (no FP8, or block/current FP8 '
+                       'scaling) from the distributed checkpoint, keeping them '
+                       'local instead of writing them. Only delayed-scaling '
+                       '`_extra_state` (amax history + scale) is ever needed and '
+                       'it is always persisted regardless of this flag. By '
+                       'default (flag off) all `_extra_state` are persisted, '
+                       'preserving the previous behavior. Older checkpoints load '
+                       'unchanged either way.')
     return parser
 
 
--- a/megatron/training/arguments.py
+++ b/megatron/training/arguments.py
@@ -2893,6 +2893,17 @@ def _add_checkpointing_args(parser):
     group.add_argument('--ckpt-fully-parallel-save', action='store_true',
                        dest='ckpt_fully_parallel_save_deprecated',
                        help='Deprecated: see --no-ckpt-fully-parallel-save.')
+    group.add_argument('--ckpt-drop-redundant-extra-state', action='store_true',
+                       default=False,
+                       help='Drop TE `_extra_state` artifacts that carry no '
+                       'irreplaceable state (no FP8, or block/current FP8 '
+                       'scaling) from the distributed checkpoint, keeping them '
+                       'local instead of writing them. Only delayed-scaling '
+                       '`_extra_state` (amax history + scale) is ever needed and '
+                       'it is always persisted regardless of this flag. By '
+                       'default (flag off) all `_extra_state` are persisted, '
+                       'preserving the previous behavior. Older checkpoints load '
+                       'unchanged either way.')
     return parser
 
 
--- a/megatron/training/checkpointing.py
+++ b/megatron/training/checkpointing.py
@@ -25,7 +25,8 @@ import torch
 from torch.distributed.checkpoint import FileSystemReader, default_planner
 
 from megatron.core import dist_checkpointing, mpu, tensor_parallel
-from megatron.core.dist_checkpointing.mapping import ShardedObject
+from megatron.core.dist_checkpointing.dict_utils import dict_list_map_inplace
+from megatron.core.dist_checkpointing.mapping import LocalNonpersistentObject, ShardedObject
 from megatron.core.dist_checkpointing.strategies.async_utils import _disable_gc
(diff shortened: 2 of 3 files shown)


megatron/training/arguments.py
```diff
@@ -2893,6 +2893,17 @@ def _add_checkpointing_args(parser):
     group.add_argument('--ckpt-fully-parallel-save', action='store_true',
                        dest='ckpt_fully_parallel_save_deprecated',
                        help='Deprecated: see --no-ckpt-fully-parallel-save.')
+    group.add_argument('--drop-redundant-extra-state', action='store_true',
```
deepakn94: Prefix with "ckpt"?

asolergi-nv:
/ok to test [e933092](https://github.com/NVIDIA/Megatron-LM/pull/5552/commits/e933092c2bcb3ea3732b827613993b8e96aff688)

deepakn94 (approved):

yaoyu-33 (approved):

PR #5553, opened 2026-06-29
Cache the FullyParallel save distribution on disk (--ckpt-pg-tensors-cache-path)

asolergi-nv:

# What does this PR do ?

Cache the FullyParallel save distribution on disk (`--ckpt-pg-tensors-cache-path` / `--ckpt-pg-tensors-cache-create`) so the planning `all_gather_object` is skipped across jobs.
 
## Issue tracking

Linked issue: 

### Pre-checks

---

# Cache the fully-parallel save/load shard distribution on disk

## Problem

When saving or loading a distributed checkpoint with fully-parallel
save/load enabled, `determine_main_replica_uniform_distribution` runs a
world-wide `all_gather_object` over the parallelization group to exchange every
rank's shard *metadata* and decide which rank is the main saver/loader of each
shard. At large scale this gather is the dominant cost of the planning step and
runs on **every job**.

Its result, however, is a pure deterministic function of the sharded-state-dict
structure and the parallel layout. For a fixed model config and world size it is
**identical across jobs and across resumes** — so paying for the collective on
every job is wasteful.

## Solution

Allow the per-parallelization-group distribution to be cached on disk and reused,
skipping the collective entirely on subsequent jobs. Two new (optional) knobs:

- `--ckpt-pg-tensors-cache-path <DIR>`: when set (and *create* is off), the
  distribution for each group is **read from disk** and the `all_gather_object`
  is skipped — both for save and load.
- `--ckpt-pg-tensors-cache-create`: run once with a matching config/world size.
  The collective runs as usual, but **both** the save (`ignore_groups=False`)
  and load (`ignore_groups=True`) distributions are derived from that single
  gather and written to disk (one file per group). Later jobs set only
  `--ckpt-pg-tensors-cache-path` to consume them.

Default (neither set) preserves the original collective-based behaviour exactly.

### Implementation

`megatron/core/dist_checkpointing/exchange_utils.py` — `determine_main_replica_uniform_distribution`
is refactored into composable pieces so the gather and the post-gather logic can
be cached/replayed independently:

- `_gather_shards_metadata()` — the single `all_gather_object` (data stripped via
  `without_data()`, so the payload is tiny).
- `_build_shard_distribution(all_shards, ignore_groups)` — the pure,
  rank-independent algorithm; rank-independence is what makes the result safe to
  persist and reload.
- `_pg_dist_cache_file_path()` — one file per group, keyed by `min(global ranks)`
  of the group (a collision-free id derivable with no collective).
- `_load_pg_dist_cache()` / `_create_pg_dist_cache()` — read / atomically write
  (`temp + os.replace`) the `{ignore_groups: ShardDistribution}` map; only the
  group's writer rank writes.
- A process-global memo (`_PG_DIST_CACHE`) makes each group create-or-read at most
  once per job (later saves / the load wrapper's reuse hit memory). The memo is
  checked **before** the collective so all ranks of a group hit/miss in lockstep.

`fully_parallel.py` — `FullyParallelSaveStrategyWrapper` and
`FullyParallelLoadStrategyWrapper` gain `pg_cache_path` / `pg_cache_create` and
pass them through. In **read** mode the save wrapper also skips the first-save
`validate_sharding_integrity` check, because that triggers its own world collective
(`determine_global_metadata`) which would defeat skipping the distribution gather.

`checkpointing.py` — wires the two args into both wrappers.

`config/training_config.py` — defines the two `CheckpointConfig` fields.

## Files changed

- `megatron/core/dist_checkpointing/exchange_utils.py`
- `megatron/core/dist_checkpointing/strategies/fully_parallel.py`
- `megatron/training/checkpointing.py`
- `megatron/training/config/training_config.py`

## Usage

First job (create the cache once, on a matching config/world size):
```
--ckpt-fully-parallel-save --ckpt-fully-parallel-load \
--ckpt-pg-tensors-cache-path $CONTAINER/pg_dist_cache --ckpt-pg-tensors-cache-create
```
Subsequent jobs (reuse, collective skipped):
```
--ckpt-fully-parallel-save --ckpt-fully-parallel-load \
--ckpt-pg-tensors-cache-path $CONTAINER/pg_dist_cache
```

## Correctness constraints

The cache is consumed with **no existence/validity checks** by design (lowest
latency). It is only valid when the consuming run has the **same model config,
parallel layout, and world size** as the run that created it. A mismatch yields a
wrong shard map. Treat the cache as config-specific and regenerate it whenever the
structure changes.

## Testing notes

- **Unit test** (`tests/unit_tests/dist_checkpointing/test_pg_distribution_cache.py`,
  `TestPgDistributionCache`): (1) CREATE writes the cache file and, after clearing
  the in-process memo, READ loads an **identical** distribution from disk
  (`main_rank_for_shard` / `all_ranks_for_shard` match); (2) the READ path is proven
  to **skip the collective** by monkeypatching `_gather_shards_metadata` to raise and
  asserting it is never called. World-size-agnostic (validated locally at ws=2;
  holds at ws=8 in CI).
- Default path (no flags) is byte-for-byte the original behaviour — the refactor
  splits the function but the create/default code path produces the same
  `ShardDistribution`.
- End-to-end: create a cache with `--ckpt-pg-tensors-cache-create`, then run a
  second job with only `--ckpt-pg-tensors-cache-path` and confirm saves/loads
  succeed and the distribution all_gather no longer appears.

### DIFF
--- a/megatron/core/dist_checkpointing/exchange_utils.py
+++ b/megatron/core/dist_checkpointing/exchange_utils.py
@@ -3,6 +3,8 @@
 """Utilities for exchanging data between ranks."""
 
 import logging
+import os
+import pickle
 from collections import defaultdict
 from functools import reduce
 from itertools import zip_longest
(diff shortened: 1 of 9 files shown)


dimapihtar (approved):

asolergi-nv:
/ok to test [27f48cd](https://github.com/NVIDIA/Megatron-LM/pull/5553/commits/27f48cd8271a44509ffe90326e2f95c857e1fdea)

asolergi-nv:
/ok to test [bd60d14](https://github.com/NVIDIA/Megatron-LM/pull/5553/commits/bd60d144c56b8939140a0a1434fd7f4ffaa6cbb1)

Phlip79 (approved):

maanug-nv (approved):

deepakn94 (approved):

deepakn94:
/ok to test [620f02a](https://github.com/NVIDIA/Megatron-LM/pull/5553/commits/620f02a1058aab667cc9d93d5a0214cce38c20c2)

asolergi-nv:
/ok to test [47345a9](https://github.com/NVIDIA/Megatron-LM/pull/5553/commits/47345a9069b571f19035529d243dde6862eefee3)

asolergi-nv:
/ok to test [4948693](https://github.com/NVIDIA/Megatron-LM/pull/5553/commits/494869337c1a116429ea328da8c865de5361a5dd)

PR #5567, opened 2026-06-30
Fix `isort` target Python version

janEbert:
Now `isort` target Python version matches `requires-python` again, meaning that packages like `tomllib` will be correctly included in the stdlib import section.


ko3n1g (approved):

janEbert:
@ko3n1g this is what I mentioned in our last meeting about the divergence between `pyproject.toml` Python version specifications.

PR #5568, opened 2026-06-30
Add RoPE config

janEbert:
- [X] I, the PR author, have personally ~reviewed~ written every line of this PR.

# What does this PR do?

Add a config container for RoPE and related functionality. It's also integrated into the argument handling logic instead of the previous workaround.

Since there are some divergences and resulting duplications in the codebase, we add some duplicate handling (regarding the ), which is unclean, but not a result of these changes. This refers to the attributes, which `MLATransformerConfig` already had in its dataclass, but which `TransformerConfig` started to now also set, but only at runtime, in the latter case with a `yarn_` prefix.

Also, there are other duplications between `RoPEConfig` and `{Hybrid,GPT}ModelConfig` attributes. The aim is to deprecate the `rotary_base` and similar duplicates in `{Hybrid,GPT}ModelConfig`, so that they're only fetched from `RoPEConfig`. However, the actual deprecation is left for a future PR. For now, we only make sure that the settings are mirrored even before applying the `RoPEConfig` to the `{Hybrid,GPT}ModelConfig`.


megatron/training/models/gpt.py
```diff
@@ -164,6 +165,7 @@ class GPTModelConfig(ModelConfig):
     rope_scaling_factor: float = 8.0
     scatter_embedding_sequence_parallel: bool = True
     seq_len_interpolation_factor: float | None = None
+    rope_config: RoPEConfig
```
maanug-nv: should it be optional? if we just want default rope, we can use the existing settings above? non-optional forces adoption, which will make deprecating the above settings faster i guess.

maanug-nv: sorry ignore my last comment. i think the right thing to do here is to use a default_factory because RoPEConfig has defaults for all attributes. so if a user touches nothing, defaults in RoPEConfig match above duplicate settings

tests/unit_tests/training/models/test_hybrid_builder.py
maanug-nv, line 1: changes in this file are unnecessary if rope_config attribute has a default

megatron/training/argument_utils.py
```diff
@@ -427,6 +433,44 @@ def _default_config_from_args(cls: type, args: Namespace, return_instance: bool
...
+        return value
+
+    kwargs = dict(
+        rotary_base=args.rotary_base,
+        rotary_percent=args.rotary_percent,
+        rotary_seq_len_interpolation_factor=args.rotary_seq_len_interpolation_factor,
+        use_rotary_scaling=args.use_rope_scaling,
+        rotary_scaling_factor=args.rope_scaling_factor,
+        yarn_rotary_scaling_factor=args.rotary_scaling_factor,
+        yarn_original_max_position_embeddings=default_if_none("yarn_original_max_position_embeddings", args.yarn_original_max_position_embeddings),
```
maanug-nv, line 460: why is the dataclass default for these different than the args default?

megatron/training/argument_utils.py
```diff
@@ -427,6 +433,44 @@ def _default_config_from_args(cls: type, args: Namespace, return_instance: bool
         return kwargs
 
 
+def rope_config_from_args(args: Namespace) -> RoPEConfig:
+    """Create a `RoPEConfig` from the appropriate values in the `args` `Namespace`."""
+    rope_config_sig = inspect.signature(RoPEConfig)
+    rope_config_params = rope_config_sig.parameters
```
maanug-nv, line 439: preference of using `dataclass.fields()` instead of inspect. Justification from codex:

> for dataclasses, `dataclasses.fields()` is usually better than `inspect.signature()`.
> 
>   `fields()` is dataclass-aware:
> 
```
  from dataclasses import MISSING, fields

  def dataclass_defaults(cls):
      result = {}
      for f in fields(cls):
          if f.default is not MISSING:
              result[f.name] = f.default
          elif f.default_factory is not MISSING:
              result[f.name] = f.default_factory()
      return result
```
> 
>   Why it’s better:
> 
>   - It directly represents dataclass fields, not just constructor parameters.
>   - It handles `field(default=...)`.
>   - It can detect `field(default_factory=...)`.
>   - It gives you field metadata, init flags, kw-only state, etc.
>   - It avoids being confused by a custom `__init__`.

megatron/training/argument_utils.py
```diff
@@ -372,42 +372,48 @@ def core_transformer_config_from_args(args, config_class=None):
...
+        config.rotary_percent = rope_config.rotary_percent
+        config.seq_len_interpolation_factor = rope_config.rotary_seq_len_interpolation_factor
+        if isinstance(config, GPTModelConfig):
+            config.rope_scaling = rope_config.use_rotary_scaling
+            config.rope_scaling_factor = rope_config.rotary_scaling_factor
+    else:
+        transformer_config = config
+
+    # `TransformerConfig` does not contain these fields, but they're set at runtime.
+    transformer_config.yarn_rotary_scaling_factor = rope_config.yarn_rotary_scaling_factor
```
maanug-nv, line 398: since they aren't defined on transformer config, this will break i think? needs to use setattr right?

megatron/training/models/gpt.py
janEbert: Made it initialize to the default `RoPEConfig` and removed the respective argument from the tests in which the default `RoPEConfig` was used.

megatron/training/argument_utils.py
janEbert: These arguments, when unspecified (i.e., `None`), either break or map to the defaults in the dataclass. These seemed super hacky in general and required a bunch of workaround code and snooping around the codebase to even grasp their behavior, since they're undefined attributes that are only added to the `TransformerConfig` at runtime.

megatron/training/argument_utils.py
janEbert: I was considering this, but the exact point of using `inspect` was that it correctly maps to `__init__` and not some other behavior.

[`dataclasses.fields` does not return `InitVar`s](https://docs.python.org/3/library/dataclasses.html#dataclasses.fields), so its use would be incorrect here. I'm open to discussion, though, if I'm missing something.

megatron/training/argument_utils.py
janEbert: I don't think what that would change anything. `TransformerConfig` isn't declared `frozen=True`, so it's fine to add new fields.

tests/unit_tests/training/models/test_hybrid_builder.py
janEbert: I did leave one usage in `test_custom_initialization`, same for `TestGPTModelConfigInitialization`.

PR #5590, opened 2026-06-30
Suppress noisy external-library log lines on non-rank-0 processes

deepakn94:
## Summary
- Add `DeprecationWarning` to the warning filter list on non-rank-0 processes in `pretrain_gpt.py` and `pretrain_hybrid.py`. Override `warnings.showwarning` as a fallback for libraries like CUTLASS DSL that use `catch_warnings()` with `simplefilter("always")` to bypass filters.
- Suppress noisy Python loggers (`GroupedGemmQuantSm100`, `GroupedGemmDsreluSm100`, `GroupedGemmSreluSm100`, `GroupedGemmWgradSm100`, `absl`) on non-rank-0 in `setup_logging()`.
- Pass `device_id` to `torch.distributed.init_process_group()` to eliminate the NCCL "Guessing device ID" C++ warning on every rank.

Together these reduce log output from ~160K lines to ~3K lines on a 128-GPU job.

## Test plan
- [x] Verified on oci-hsg with 32-node (128 GPU) job: log went from 159K → 21K → 3K lines
- [x] Remaining non-rank-0 lines are only NCCL version strings (~128) and legitimate rank-127 Megatron output (timers, iteration logs)

### DIFF
--- a/megatron/training/initialize.py
+++ b/megatron/training/initialize.py
@@ -336,6 +336,7 @@ def _initialize_distributed(get_embedding_ranks, get_position_embedding_ranks, s
             'world_size': args.world_size,
             'rank': args.rank,
             'timeout': timedelta(minutes=args.distributed_timeout_minutes),
+            'device_id': device_id,
         }
         if args.fake_process_group:
             assert is_torch_min_version(
@@ -562,3 +563,13 @@ def setup_logging() -> None:
         if is_rank0():
             logger.info(f'Setting logging level to {logging_level}')
         logging.getLogger().setLevel(logging_level)
+
+    if not is_rank0():
+        for noisy_logger_name in [
+            'GroupedGemmQuantSm100',
+            'GroupedGemmDsreluSm100',
+            'GroupedGemmSreluSm100',
+            'GroupedGemmWgradSm100',
+            'absl',
+        ]:
+            logging.getLogger(noisy_logger_name).setLevel(logging.ERROR)
--- a/megatron/training/initialize.py
+++ b/megatron/training/initialize.py
@@ -580,3 +580,13 @@ def setup_logging() -> None:
         if is_rank0():
             logger.info(f'Setting logging level to {logging_level}')
         logging.getLogger().setLevel(logging_level)
+
+    if not is_rank0():
+        for noisy_logger_name in [
+            'GroupedGemmQuantSm100',
+            'GroupedGemmDsreluSm100',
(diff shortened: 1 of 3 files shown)


jaredcasper (approved):

megatron/training/initialize.py
```diff
@@ -562,3 +522,13 @@ def setup_logging() -> None:
         if is_rank0():
             logger.info(f'Setting logging level to {logging_level}')
         logging.getLogger().setLevel(logging_level)
+
+    if not is_rank0():
```
maanug-nv, line 584: can we just use this instead: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/utils/log_utils.py#L66-L70

was going to remove the version in this file (initialize.py) in a different PR that hasn't been merged yet. but it seems that it would do the same thing?

deepakn94: Where should I call the version in `log_utils`?

maanug-nv: does it not work if called in the same place this one is called? ie log_utils.setup_logging should just replace initialize.setup_logging?

megatron/training/initialize.py
```diff
@@ -336,6 +336,7 @@ def _initialize_distributed(get_embedding_ranks, get_position_embedding_ranks, s
             'world_size': args.world_size,
             'rank': args.rank,
             'timeout': timedelta(minutes=args.distributed_timeout_minutes),
+            'device_id': device_id,
```
deepakn94: Remove this line for now, doesn't seem it's safe.

deepakn94: Removed, and rebased onto latest `main`. That line was unrelated to log suppression and should not have been in this PR.

deepakn94: Correcting my previous reply: it was wrong to say the line was unrelated. Passing `device_id` to `init_process_group` eagerly binds the device, which suppresses a PyTorch warning, so it was in scope for this PR.

It is removed because it appears to cause hangs, not because it was off-topic. Worth revisiting separately if the hang is understood.

maanug-nv (approved):

deepakn94:
/ok to test [7027e38](https://github.com/NVIDIA/Megatron-LM/pull/5590/commits/7027e384385ade40f90a6afb92b887add56ce770)

Phlip79 (approved):

PR #5600, opened 2026-07-01
Add per-layer transformer resource profiling

### LINKED ISSUE #5595: Add per-layer measured logging for activation memory and forward/backward time
This item is listed in the Q2 2026 roadmap (#4997) under Ease of Use ("Per-layer logging and memory estimation").

When planning a model configuration against a fixed memory budget, or debugging which layers dominate step latency, Megatron currently offers no on-line, per-layer view of measured resource usage. The existing tools each cover a different part of the space but leave a gap:

report_memory (megatron/training/utils/common_utils.py) reports process-global CUDA memory (memory_allocated / max_memory_allocated / reserved), not per-layer attribution.
timers.py (megatron/core/timers.py) provides operation-level timing (e.g. optimizer-copy-to-main-grad, params-all-gather), but nothing at transformer-layer granularity.
theoretical_memory_usage.py provides detailed analytical per-layer estimates, but they are never reconciled against measured values (the activation path relies on a hardcoded overhead_factor = 1.05).
nvtx instrumentation exists, but requires an external profiler (nsys) and offline analysis, and the existing ranges cover specific ops (e.g. gated_delta_net, fused swiglu) rather than systematically per transformer layer.

None of these give an in-framework, per-layer breakdown of measured memory and time during a normal training run.

A config-gated, measured per-layer profiling utility, scoped to keep the first contribution reviewable:

1. Per-layer logging of allocated activation memory and forward/backward wall-clock time, collected via module hooks on TransformerLayer (megatron/core/transformer/transformer_layer.py), iterating over layers in TransformerBlock.
2. Reuse the existing CUDA memory-query pattern from report_memory and the timing conventions in timers.py rather than introducing a parallel path.
3. Gated behind a config flag, disabled by default, with negligible overhead when off; structured per-layer summary emitted at a configurable interval (consistent with log_memory_interval).

First PR scoped to single-device measured logging. A natural follow-up would be a theoretical-vs-measured comparison view that reconciles these numbers against theoretical_memory_usage.py — which could, in turn, help calibrate the hardcoded overhead factor.

A few questions before I start:
1. Preferred location — under megatron/core/ profiling utilities, or alongside the existing logging code?
2. Any prior design discussion or owner I should align with?
3. is there interest in a measured counterpart to theoretical_memory_usage.py, and would reconciling the two (to calibrate the overhead factor) be in scope?

1. External profiling via existing nvtx ranges + nsys: works, but requires external tooling and offline analysis, isn't integrated with training logs or the config system, and current ranges aren't systematically per-layer.
2. Existing report_memory: process-global only, no per-layer attribution.
3. Existing theoretical_memory_usage.py: analytical per-layer estimates, but purely theoretical and never validated against measurements.

An in-framework, config-gated, per-layer measured logger fills the gap none of these cover: on-line, per-layer, no external tooling required.

**Additional context**
A few questions:
1. Is there a preferred location for this — under megatron/core/ profiling utilities, or alongside the existing logging in megatron/training/?
2. Is a measured counterpart to theoretical_memory_usage.py (and reconciling the two) of interest, or should the first PR stay strictly to per-layer logging?
3. Any prior design discussion or owner I should align with?
Validation hardware available: single cloud GPU (A100 / 4090). Happy to open a draft PR once scope is confirmed.

michaelsam94:
## Summary
- add a config-gated per-layer transformer profiler that uses module hooks to collect activation memory deltas plus forward/backward wall-clock time
- wire profiler construction into TransformerBlock and flush summaries from the training log at the configured interval
- add focused unit tests for profiler accumulation, reset, logging, and missing-model handling

Fixes #5595

## Testing
- python -m py_compile megatron/core/transformer/per_layer_profiling.py megatron/core/transformer/transformer_block.py megatron/core/transformer/transformer_config.py megatron/training/argument_utils.py megatron/training/training.py tests/unit_tests/transformer/test_per_layer_profiling.py
- git diff --check

Note: could not run pytest/isort locally because this environment does not have torch, pytest, isort, or uv available on PATH.

### DIFF
--- a/megatron/core/transformer/per_layer_profiling.py
+++ b/megatron/core/transformer/per_layer_profiling.py
@@ -1,0 +1,167 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+"""Per-layer measured resource profiling for transformer layers."""
+
+from __future__ import annotations
+
+import time
+from dataclasses import dataclass
+from typing import Iterable
+
+import torch
+
+from megatron.core.utils import get_pg_rank
+
+
+_BYTES_PER_MB = 1024.0 * 1024.0
+
+
+@dataclass
+class PerLayerProfileStats:
+    """Accumulated measured resource usage for one transformer layer."""
+
+    layer_number: int
+    forward_calls: int = 0
+    backward_calls: int = 0
+    activation_memory_bytes: int = 0
+    forward_time_ms: float = 0.0
+    backward_time_ms: float = 0.0
+
+    def reset(self) -> None:
+        """Reset counters after a log interval is emitted."""
+        self.forward_calls = 0
+        self.backward_calls = 0
+        self.activation_memory_bytes = 0
+        self.forward_time_ms = 0.0
+        self.backward_time_ms = 0.0
+
+
+class PerLayerProfiler:
+    """Collect rank-local memory and wall-clock timing with module hooks."""
+
+    def __init__(self, layers: Iterable[torch.nn.Module], layer_offset: int = 0) -> None:
+        self._handles: list[torch.utils.hooks.RemovableHandle] = []
+        self._stats_by_module: dict[torch.nn.Module, PerLayerProfileStats] = {}
+        self._forward_start: dict[torch.nn.Module, tuple[float, int]] = {}
+        self._backward_start: dict[torch.nn.Module, float] = {}
+
+        for local_idx, layer in enumerate(layers):
+            layer_number = getattr(layer, "layer_number", layer_offset + local_idx + 1)
+            self._stats_by_module[layer] = PerLayerProfileStats(layer_number=layer_number)
+            self._handles.extend(
+                [
+                    layer.register_forward_pre_hook(self._forward_pre_hook),
+                    layer.register_forward_hook(self._forward_hook),
+                    layer.register_full_backward_pre_hook(self._backward_pre_hook),
+                    layer.register_full_backward_hook(self._backward_hook),
+                ]
+            )
+
+    def remove(self) -> None:
+        """Remove all registered hooks."""
+        for handle in self._handles:
+            handle.remove()
+        self._handles.clear()
+
+    def summary(self, reset: bool = False) -> list[PerLayerProfileStats]:
+        """Return accumulated stats sorted by layer number."""
+        stats = sorted(self._stats_by_module.values(), key=lambda item: item.layer_number)
+        snapshot = [
+            PerLayerProfileStats(
+                layer_number=item.layer_number,
+                forward_calls=item.forward_calls,
+                backward_calls=item.backward_calls,
+                activation_memory_bytes=item.activation_memory_bytes,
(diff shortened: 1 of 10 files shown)


maxyzhu:
Hi @michaelsam94 — @Connor-XY suggested we coordinate here. I've been working on the same feature from #5595 and have a GPU-validated implementation; I've laid out the approach and some validation in #5595. Happy to compare notes and merge the best of both. Looking forward to working together.

maxyzhu:
Following up here per @Connor-XY's coordination request. I've opened a GPU-validated draft for the same per-layer scope at #5662 (forward time + memory, with backward and per-layer peak scoped as documented follow-ups). It's a separate PR only so the validation and tests travel with it.

The four points raised (hook timing, CUDA sync, memory attribution, core/training boundary) are addressed in the #5662 description, with GPU logs across single-GPU / recompute / TP2 / PP2 attached. Some other possibilities with other APIs are tested also.

michaelsam94:
Updated the PR to address the modification request and align the implementation with the GPU-validated direction discussed in #5662.

What changed:
- Reworked per-layer profiling so hooks are installed lazily only on logged steps, instead of being permanently attached during model construction.
- Removed per-layer backward hooks and per-hook CUDA synchronization. Forward timing now uses deferred CUDA events and is resolved only at log time.
- Changed memory reporting to use CUDA allocator snapshots for allocated delta, reserved delta, peak-after, peak-rise, and global allocated/reserved peaks.
- Renamed the user-facing flag to `--log-per-layer-profiling` and require it to be paired with `--log-memory-interval`.
- Kept distributed/rank decisions in training code; the core profiler now returns structured summaries and prints only when asked by the caller.
- Added CPU unit coverage for aggregation/summarization/log-rank behavior.
- Added two functional test configs intended to validate the GPU paths for TP1/PP1 and TP1/PP2.

Locally verified before pushing:
- `git diff --cached --check`
- `python3 -m py_compile` on the changed Python files
- Added-line length scan for the changed Python files

Still needs verification in the Megatron CI/container GPU environment:
- Unit test: `tests/unit_tests/transformer/test_per_layer_profiling.py`
- Functional GPU configs:
  - `tests/functional_tests/test_cases/gpt/gpt3_per_layer_profiling_tp1_pp1_1node/model_config.yaml`
  - `tests/functional_tests/test_cases/gpt/gpt3_per_layer_profiling_tp1_pp2_1node/model_config.yaml`
- Confirm log output and memory/timing semantics on real CUDA runs, especially recompute and pipeline-parallel behavior.

I could not run those GPU/container tests on the local machine because this environment does not have `uv`, `pytest`, `torch`, Docker, or `nvidia-smi` available.

maxyzhu:
Hi @Connor-XY, #5662 PR is updated . Now supports fwd/bwd timing and per-layer memory stats (peak, retained, largest,comm...) in single model w/ zero-sync in hot path. Roadmap is also updated to cover different parallel scenarios and MoE specific analysis soon. Welcome to review and leave comments!

Issue #15, opened 2019-10-17
Error when runinng script pretrain_gpt2_distributed.sh


Lepiloff:
When I run 
`OMP_NUM_THREADS=10  bash scripts/pretrain_gpt2_distributed.sh`

I got an error

```
> initializing model parallel cuda seeds on global rank 0, model parallel rank 0, and data parallel rank 0 with model parallel seed: 3952 and data parallel seed: 1234
configuring data
Traceback (most recent call last):
  File "pretrain_gpt2.py", line 625, in <module>
    main()
  File "pretrain_gpt2.py", line 569, in main
    args.eod_token = get_train_val_test_data(args)
  File "pretrain_gpt2.py", line 515, in get_train_val_test_data
    args)
  File "/home/ubuntu/Megatron-LM/configure_data.py", line 34, in apply
    return make_loaders(args)
  File "/home/ubuntu/Megatron-LM/configure_data.py", line 170, in make_loaders
    train, tokenizer = data_utils.make_dataset(**data_set_args)
  File "/home/ubuntu/Megatron-LM/data_utils/__init__.py", line 114, in make_dataset
    ds = [GPT2Dataset(d, max_seq_len=seq_length) if d is not None else None for d in ds]
  File "/home/ubuntu/Megatron-LM/data_utils/__init__.py", line 114, in <listcomp>
    ds = [GPT2Dataset(d, max_seq_len=seq_length) if d is not None else None for d in ds]
  File "/home/ubuntu/Megatron-LM/data_utils/datasets.py", line 477, in __init__
    self.init_weighting()
  File "/home/ubuntu/Megatron-LM/data_utils/datasets.py", line 487, in init_weighting
    self.weighting = list(accumulate(lens))
TypeError: iteration over a 0-d array
Traceback (most recent call last):
  File "pretrain_gpt2.py", line 625, in <module>
    main()
  File "pretrain_gpt2.py", line 569, in main
    args.eod_token = get_train_val_test_data(args)
  File "pretrain_gpt2.py", line 515, in get_train_val_test_data
    args)
  File "/home/ubuntu/Megatron-LM/configure_data.py", line 34, in apply
    return make_loaders(args)
  File "/home/ubuntu/Megatron-LM/configure_data.py", line 170, in make_loaders
    train, tokenizer = data_utils.make_dataset(**data_set_args)
  File "/home/ubuntu/Megatron-LM/data_utils/__init__.py", line 114, in make_dataset
[...]

Issue #14, opened 2019-10-12
1:train loss decrease too faster.2:learning rates did not change after warmup iter,always kept on 1.5e-4. Is it normal phenomenon? 


kangzhonghua:
Thank your very much! I was encounter some question.
     1:train loss decrease too faster.
     2:learning rates did not change after warmup iter,always kept on 1.5e-4. 
Is it normal phenomenon?

total 10GB chinese corpus ,about 3000000 sample.
python3 -m torch.distributed.launch \
    --nnodes 1 \
    --nproc_per_node 2 \
    pretrain_gpt2.py \
        --num-layers 24 \
        --hidden-size 1024 \
        --num-attention-heads 16 \
        --max-position-embeddings 1024 \
        --seq-length 1024 \
        --batch-size 8 \
        --train-iters 1000000 \
        --save-interval 1000 \
        --save checkpoints/gpt2_345m_hm10g \
        --load checkpoints/gpt2_345m_hm10g \
        --tensorboard-dir logs/gpt2_345m_hm10g \
        --resume-dataloader \
        --train-data corpus_data \
        --lazy-loader \
        --tokenizer-type SentencePieceTokenizer \
        --tokenizer-path data/spm/corpus_bpe_32k.model \
        --cache-dir cache \
        --split 949,50,1 \
        --distributed-backend nccl \
        --lr 0.00015 \
        --lr-decay-style cosine \
        --weight-decay 1e-2 \
        --clip-grad 1.0 \
        --warmup .01 \
        --checkpoint-activations \
        --fp16 \

![a_t](https://user-images.githubusercontent.com/12265127/66694293-9d768680-ece4-11e9-9ab8-cca43588c41e.png)
![b_g_t](https://user-images.githubusercontent.com/12265127/66694294-9d768680-ece4-11e9-8d27-ff3ed20f5aeb.png)
![b_t](https://user-images.githubusercontent.com/12265127/66694295-9ea7b380-ece4-11e9-89d1-7fc1fdce9db1.png)
![d_l_t](https://user-images.githubusercontent.com/12265127/66694296-9ea7b380-ece4-11e9-8a1e-38dd269886e8.png)
![f_t](https://user-images.githubusercontent.com/12265127/66694299-9f404a00-ece4-11e9-80c1-2d04108bf28f.png)
![i_t](https://user-images.githubusercontent.com/12265127/66694300-9f404a00-ece4-11e9-8dfd-950a771741f9.png)
[...]

Issue #13, opened 2019-10-10
No module named 'apex'


GrahamboJangles:
I get this when running `generate_text.sh`
```
Traceback (most recent call last):
  File "generate_samples.py", line 28, in <module>
    from utils import Timers
  File "/content/Megatron-LM/utils.py", line 25, in <module>
    from fp16 import FP16_Optimizer
  File "/content/Megatron-LM/fp16/__init__.py", line 15, in <module>
    from .fp16util import (
  File "/content/Megatron-LM/fp16/fp16util.py", line 21, in <module>
    import mpu
  File "/content/Megatron-LM/mpu/__init__.py", line 35, in <module>
    from .layers import ColumnParallelLinear
  File "/content/Megatron-LM/mpu/layers.py", line 28, in <module>
    from apex.normalization.fused_layer_norm import FusedLayerNorm as LayerNorm
ModuleNotFoundError: No module named 'apex'
```

raulpuric:
It's included in our nvidia docker [containers](https://github.com/NVIDIA/Megatron-LM/blob/master/docker/Dockerfile)

but you can also get it from [here](https://github.com/NVIDIA/apex)

GrahamboJangles:
@raulpuric 

Okay, I have installed apex and all of the other dependencies but now I get this:
```
Generate Samples
WARNING: No training data specified
using world size: 1 and model-parallel size: 1 
 > using dynamic loss scaling
> initializing model parallel with size 1
> initializing model parallel cuda seeds on global rank 0, model parallel rank 0, and data parallel rank 0 with model parallel seed: 3952 and data parallel seed: 1234
100% 1042301/1042301 [00:00<00:00, 5937749.24B/s]
100% 456318/456318 [00:00<00:00, 3423447.30B/s]
prepare tokenizer done
building GPT2 model ...
 > number of parameters on model parallel rank 0: 124475904
WARNING: could not find the metadata file checkpoints/gpt2_345m/latest_checkpointed_iteration.txt 
    will not load any checkpoints and will start from random
Traceback (most recent call last):
  File "generate_samples.py", line 496, in <module>
    main()
  File "generate_samples.py", line 492, in main
    write_and_generate_samples_unconditional(model, tokenizer, args)
  File "generate_samples.py", line 319, in write_and_generate_samples_unconditional
    for datum in generate_samples_unconditional(model, tokenizer, args):
  File "generate_samples.py", line 295, in generate_samples_unconditional
[...]

Issue #12, opened 2019-10-07
Collecting Wikipedia Training Data issues


great-attraction:
Hi!
I'm dealing with the work of your library and I have a misunderstanding of the point Collecting Wikipedia Training Data.
I run command
[python WikiExtractor.py --json enwiki-latest-pages-articles.xml.bz2](url)
and expected get one json file as output ( do I understand correctly that I need an available .json file to start learning the model ?), but got a  folder with many inner folders like AA, AB, AC ..

Could you explain what I'm doing wrong and how I can start training a model?

jon-barker:
You need to convert the extracted .json files into a single .json file to pass to `preprocess_data.py` as described here: https://github.com/NVIDIA/Megatron-LM/tree/main#data-preprocessing

This will process the data into mmap files that can then be used for training.

PR #5606, opened 2026-07-01
  Add VLM support to DynamicInferenceEngine

RPrenger:
Extend the dynamic-batching inference path to accept per-request image
  inputs. The engine now offers VLM kwargs on add_request (imgs,
  imgs_sizes, num_tiles, num_img_embeddings_per_tile) that:

    * expand <image> placeholders into pad tokens (static-resolution and dynamic-resolution cases both supported);
    * run the vision encoder on the first PP stage to produce per-request image embeddings;
    * attach the embeddings and image-token mask to the context so the decoder forward can splice them back in.

  DynamicInferenceContext gains _request_to_image_embeddings/
  _request_to_image_token_mask/_request_to_image_token_count dicts and
  four new accessors (add_vlm_request_data, remove_vlm_request_data,
  current_image_token_mask, current_image_embeddings); the per-request
  data is cleared at reset_metadata and on request completion in
  post_process_requests.

  VLMInferenceWrapper gains expand_image_tokens, _forward_vision_encoder
  and _forward_dynamic to back the engine path. All preexisting wrapper
  methods are unchanged, and the new behavior is dormant unless a caller
  passes imgs/imgs_sizes/num_tiles to add_request, so text-only inference
  is unaffected.

  TextGenerationController._dynamic_step_forward_logits picks up the
  per-request image data from the context and forwards it through the
  wrapper's inference_input dict. Out-of-vocabulary placeholders
  introduced by VLM expansion would otherwise leak into expert-parallel
  dummy forwards (whose logits are discarded but whose token IDs still
  have to be in vocab); _sanitize_dummy_input_ids replaces them with a
  stable valid token on the dummy path only.

  DynamicVLMInferenceRequest combines DynamicInferenceRequest and
  VLMInferenceRequest, carrying the precomputed image_embeddings and
  image_token_mask alongside the usual request state.

# What does this PR do?

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

### DIFF
--- a/megatron/core/inference/contexts/dynamic_context.py
+++ b/megatron/core/inference/contexts/dynamic_context.py
@@ -284,6 +284,11 @@ class DynamicInferenceContext(BaseInferenceContext):
         self.async_sched_step_count = 0
         self.async_sched_compaction_step_count = 0
 
+        # Per-request VLM data (empty when not using multimodal).
+        self._request_to_image_embeddings: Dict[int, Optional[Tensor]] = {}
+        self._request_to_image_token_mask: Dict[int, Optional[Tensor]] = {}
+        self._request_to_image_token_count: Dict[int, int] = {}
+
         self.cache_mla_latent = (
             isinstance(model_config, MLATransformerConfig) and model_config.cache_mla_latents
         )
@@ -2534,6 +2539,11 @@ class DynamicInferenceContext(BaseInferenceContext):
             token_count=0, prefill_req_count=0, decode_req_count=0
         )
 
+        # Reset VLM data.
+        self._request_to_image_embeddings.clear()
+        self._request_to_image_token_mask.clear()
+        self._request_to_image_token_count.clear()
+
     def reset(self) -> None:
         """Reset entire context.
 
@@ -4137,3 +4147,113 @@ class DynamicInferenceContext(BaseInferenceContext):
             'total_request_count': int(total_request_count),
             'max_requests': int(self.max_requests),
         }
+
+    # ----- VLM per-request data management -----
+
+    def add_vlm_request_data(
+        self,
+        request_id: int,
+        image_embeddings: Optional[Tensor],
+        image_token_mask: Optional[Tensor],
+    ) -> None:
+        """Attach per-request image data to the context.
+
+        Called at add_request time when a multimodal request has images.
+        No-op overhead for text-only requests that never call this.
+
+        Args:
+            request_id: The request identifier.
+            image_embeddings: Tensor of shape [seq_img, 1, hidden] or None.
+            image_token_mask: 1-D tensor with -1 for text positions and
+                non-negative indices for image embedding positions; or None.
+        """
+        self._request_to_image_embeddings[request_id] = image_embeddings
+        self._request_to_image_token_mask[request_id] = image_token_mask
+        if image_embeddings is not None:
+            self._request_to_image_token_count[request_id] = int(
+                image_embeddings.shape[0] * image_embeddings.shape[1]
+            )
+        else:
+            self._request_to_image_token_count[request_id] = 0
+
+    def remove_vlm_request_data(self, request_id: int) -> None:
+        """Remove image data for a finished request."""
+        self._request_to_image_embeddings.pop(request_id, None)
+        self._request_to_image_token_mask.pop(request_id, None)
+        self._request_to_image_token_count.pop(request_id, None)
+
+    def current_image_token_mask(self) -> Optional[Tensor]:
+        """Flattened image-token mask aligned with current_input_ids.
+
+        Returns a [1, padded_active_token_count] tensor with -1 for non-image
+        positions and non-negative indices into the concatenated image
+        embeddings, or None when there are no active tokens. During
+        decode-only steps (no image tokens consumed) returns all -1.
+        """
+        if self.padded_active_token_count is None or self.padded_active_token_count == 0:
+            return None
+
+        if self.is_decode_only():
+            return torch.full(
+                (self.padded_active_token_count,),
+                -1,
+                dtype=torch.long,
+                device=torch.cuda.current_device(),
+            )
+
+        segments: List[Tensor] = []
+        cumulative_offset = 0
+        for row_idx in range(self.paused_request_count, self.total_request_count):
+            request_id = int(self.request_ids[row_idx].item())
+            query_len = int(self.request_query_lengths[row_idx].item())
+
+            per_request_mask = self._request_to_image_token_mask.get(request_id, None)
+            if per_request_mask is None:
+                seg = torch.full(
+                    (query_len,), -1, dtype=torch.long, device=torch.cuda.current_device()
+                )
+            else:
+                seg = per_request_mask[:query_len].clone()
+                positive = seg >= 0
+                if positive.any():
+                    seg[positive] += cumulative_offset
+
+            segments.append(seg)
+            cumulative_offset += int(self._request_to_image_token_count.get(request_id, 0))
+
+        if len(segments) == 0:
+            return None
+
+        mask = torch.cat(segments, dim=0)
+        if mask.numel() < int(self.padded_active_token_count):
+            pad = torch.full(
+                (int(self.padded_active_token_count) - mask.numel(),),
+                -1,
+                dtype=torch.long,
+                device=mask.device,
+            )
+            mask = torch.cat([mask, pad], dim=0)
+        return mask.unsqueeze(0)
+
+    def current_image_embeddings(self) -> Optional[Tensor]:
+        """Concatenate image embeddings for active requests.
+
+        Each per-request embedding has shape [seq_img_i, 1, hidden] where
+        seq_img_i may vary across requests under dynamic resolution. Each is
+        flattened to [seq_img_i, hidden], concatenated along dim 0, and
+        unsqueezed to [sum(seq_img_i), 1, hidden]. The downstream consumer
+        does ``permute(1, 0, 2).reshape(-1, hidden)`` to yield the flat
+        [total_image_tokens, hidden] array indexed by
+        :meth:`current_image_token_mask`.
+
+        Returns tensor of shape [total_image_tokens, 1, hidden] or None.
+        """
+        parts: List[Tensor] = []
+        for row_idx in range(self.paused_request_count, self.total_request_count):
+            request_id = int(self.request_ids[row_idx].item())
+            emb = self._request_to_image_embeddings.get(request_id, None)
+            if emb is not None:
+                parts.append(emb.reshape(-1, emb.shape[-1]))
+        if not parts:
+            return None
+        return torch.cat(parts, dim=0).unsqueeze(1)
--- a/megatron/core/inference/contexts/dynamic_context.py
+++ b/megatron/core/inference/contexts/dynamic_context.py
@@ -4201,12 +4201,13 @@ class DynamicInferenceContext(BaseInferenceContext):
                 device=torch.cuda.current_device(),
             )
 
+        active_slice = slice(self.paused_request_count, self.total_request_count)
+        active_request_ids = self.request_ids[active_slice].tolist()
+        active_query_lengths = self.request_query_lengths[active_slice].tolist()
+
         segments: List[Tensor] = []
         cumulative_offset = 0
-        for row_idx in range(self.paused_request_count, self.total_request_count):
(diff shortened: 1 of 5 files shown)


megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -4137,3 +3990,113 @@ def get_kvcache_utilization_stats(self) -> dict:
...
+                -1,
+                dtype=torch.long,
+                device=torch.cuda.current_device(),
+            )
+
+        segments: List[Tensor] = []
+        cumulative_offset = 0
+        for row_idx in range(self.paused_request_count, self.total_request_count):
+            request_id = int(self.request_ids[row_idx].item())
+            query_len = int(self.request_query_lengths[row_idx].item())
```
santhnm2: Instead of doing individual `.item` calls for each `row_idx`, can we do batched transfers of `request_ids[self.paused_request_count : self.total_request_count]` and `self.request_query_lengths[self.paused_request_count : self.total_request_count]` before the loop?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -4137,3 +3990,113 @@ def get_kvcache_utilization_stats(self) -> dict:
...
+        unsqueezed to [sum(seq_img_i), 1, hidden]. The downstream consumer
+        does ``permute(1, 0, 2).reshape(-1, hidden)`` to yield the flat
+        [total_image_tokens, hidden] array indexed by
+        :meth:`current_image_token_mask`.
+
+        Returns tensor of shape [total_image_tokens, 1, hidden] or None.
+        """
+        parts: List[Tensor] = []
+        for row_idx in range(self.paused_request_count, self.total_request_count):
+            request_id = int(self.request_ids[row_idx].item())
```
santhnm2: Same comment about batched transfer on `request_ids`

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1174,16 +1154,84 @@ def add_request(
...
 
-        # Initialize request.
-        request = DynamicInferenceRequest(
-            request_id=request_id,
-            prompt=prompt_str,
-            prompt_tokens=tokens,
-            sampling_params=sampling_params,
-            block_size_tokens=self.context.block_size_tokens,
-            enable_prefix_caching=self.context.enable_prefix_caching,
+        # --- VLM: expand image tokens and run vision encoder ---
```
santhnm2: Can we move the image-related code to an internal helper function? 
```
if has_images:
    request = self._build_vlm_request(imgs, num_tiles, imgs_sizes)
else:
    request = DynamicInferenceRequest(...)
```

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -705,25 +616,66 @@ def _dynamic_step_context_init(
...
+        Expert-parallel dummy forwards are synthetic; they only keep ranks with
+        no local request participating in the same collectives as active ranks.
+        Their logits are discarded, so any out-of-vocabulary placeholder token
+        can be replaced with a stable valid token without affecting user-visible
+        output. This keeps the dummy path independent of how dynamic VLM
+        resolution expands image placeholders for real requests.
+        """
+        invalid_mask = (input_ids < 0) | (input_ids >= self.vocab_size)
+        if not invalid_mask.any():
+            return input_ids
```
santhnm2, line 721: Can we just directly do `input_ids[invalid_mask] = 0`? This should be a no-op if `invalid_mask.any()` is False right?

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -705,25 +616,66 @@ def _dynamic_step_context_init(
...
+
+        input_ids = input_ids.clone()
```
santhnm2, line 723: Why do we need the `.clone()`?

megatron/core/inference/contexts/dynamic_context.py
RPrenger: Yes, I pushed a change.  Though I don't think these request_ids are on the CPU, so `.item()` isn't causing a GPU sync or anything right now.  But probably good to avoid it.

megatron/core/inference/contexts/dynamic_context.py
RPrenger: Same as the last comment.  Fixed now.

megatron/core/inference/engines/dynamic_engine.py
RPrenger: Yep.  Changed it.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
RPrenger: So `input_ids` is a view into `context.gpu_view.token_to_input_ids` and I think it's cached for reuse across steps in `context._input_position_views`.  So we need the `.clone()` I think.  And then the `.any()` is so we can skip both `.clone()` and the assign if there are no invalid tokens.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
RPrenger: See above.  I think we need the clone because it gets cached and reused.

santhnm2: I see, could you add a comment explaining this?

santhnm2 (approved):

RPrenger:
To keep this PR reviewable, it intentionally ships engine + wire plumbing for VLM dynamic batching without the LLaVAModel companion changes

(forward_lm_only, _dynamic_resolution / _patch_dim / _class_token_len), which will land
in a follow-up PR. Text-only inference is unaffected.

Latest commit addresses each comment:

- #12 — _forward_dynamic now raises NotImplementedError with an explicit "requires
companion LLaVAModel changes" message when forward_lm_only isn't present, instead of an
AttributeError at call time.
- #13 — expand_image_tokens and _forward_vision_encoder now raise NotImplementedError
when a caller passes imgs_sizes (dynamic resolution) but the model lacks
_dynamic_resolution, instead of silently falling through to the static path. Also removed
the unused drop_class_token / class_token_len locals inside the dynamic branch.
- #14 — _build_vlm_request asserts pipeline-parallel world size 1 (PP>1 needs the
non-first-stage embedding recv path that the wrapper's _recv_only_vision_embeds TODO
refers to; not upstream yet). Added the shape assertion you asked for at the splice site
so token-count / embedding-count disagreement fails loudly rather than indexing out of
bounds.
- #15 — Both current_image_token_mask and current_image_embeddings now short-circuit and
return None when the per-request VLM dicts are empty. Text-only workloads skip the
.tolist() / torch.cat on every step.

kvareddy:
@lmcafee-nvidia can you please take a look at this MR?

kvareddy:
@RPrenger  can we add unit test cases to Megatron Inference for VLM support?

RPrenger:
> @RPrenger can we add unit test cases to Megatron Inference for VLM support?

Happy to add tests, but I'd like to propose deferring them to the follow-up PR that lands the `LLaVAModel` companion changes (`forward_lm_only`, `_dynamic_resolution` etc.). Rationale:

This PR ships engine + wire plumbing only. Every new code path either (a) fires only when the `LLaVAModel` companion is present or (b) is a `NotImplementedError` guard for the case where it isn't. Tests I could write against upstream today would look like:

- Pure `DynamicInferenceContext` mask/embedding accessor tests with hand-built tensors (verifies the perf early-return + offset math).
- `_sanitize_dummy_input_ids` tensor tests (out-of-vocab → 0 replacement).
- `NotImplementedError` guards fire when `forward_lm_only` / `_dynamic_resolution` are absent.
- `expand_image_tokens` static path with a `MagicMock` module.

Those exercise the plumbing but don't validate any actual VLM inference — the mocks stand in for the very code that isn't upstream yet. A follow-up PR with the `LLaVAModel` companion + a real end-to-end VLM inference test would be a stronger safeguard against regressions than mock-heavy plumbing tests here.

If you'd prefer we get some coverage in this PR anyway, happy to add the guard-and-context tests above — just wanted to check on scope first.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1189,19 +1215,111 @@ def add_request(
...
+            request_id, image_embeddings=image_embeddings, image_token_mask=mask_tensor
+        )
+
+        return DynamicVLMInferenceRequest(
             request_id=request_id,
             prompt=prompt_str,
             prompt_tokens=tokens,
             sampling_params=sampling_params,
             block_size_tokens=self.context.block_size_tokens,
             enable_prefix_caching=self.context.enable_prefix_caching,
```
lmcafee-nvidia, line 1314: I could be wrong, but it seems placeholders (-1 values) are initially used for the image tokens until the first forward pass runs. this means that the placeholders are what's added to the request object, and added to the prefix cache if we are using prefix caching. since the cache sees the -1's and not the actual image, then multiple requests can point to the same cache blocks. for example, “describe <image>” with first a cat, and later a dog, the dog could reuse the cat’s saved state for the dog request. if this is the case, then we should disable prefix caching for image data; or if possible maybe we could cache only the text tokens before the first image.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -4137,3 +4147,127 @@ def get_kvcache_utilization_stats(self) -> dict:
...
+        segments: List[Tensor] = []
+        cumulative_offset = 0
+        for request_id, query_len in zip(active_request_ids, active_query_lengths):
+            per_request_mask = self._request_to_image_token_mask.get(request_id, None)
+            if per_request_mask is None:
+                seg = torch.full(
+                    (query_len,), -1, dtype=torch.long, device=torch.cuda.current_device()
+                )
+            else:
+                seg = per_request_mask[:query_len].clone()
```
lmcafee-nvidia, line 4222: this always reads image positions from the start of the prompt, even when processing a later chunk. for example, if the first chunk is `[image, image, "what"]` and the second is `["is", "this", "?"]`, the second chunk can reuse the first chunk’s image positions and replace `"is"` and `"this"` with image data. I think the image positions need to be aligned with the chunk currently being processed, and probably add a test for this case.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -4137,3 +4147,127 @@ def get_kvcache_utilization_stats(self) -> dict:
...
+
+        active_request_ids = self.request_ids[
+            self.paused_request_count : self.total_request_count
+        ].tolist()
+
+        parts: List[Tensor] = []
+        for request_id in active_request_ids:
+            emb = self._request_to_image_embeddings.get(request_id, None)
+            if emb is not None:
+                parts.append(emb.reshape(-1, emb.shape[-1]))
```
lmcafee-nvidia, line 4270: I believe this combines a fixed-size, multi-tile image in the wrong order. for example, if tile A has patches `A1, A2` and tile B has `B1, B2`, the expected order is `A1, A2, B1, B2`, but this produces `A1, B1, A2, B2`. the patches are then attached to the wrong image positions. we should probably put all patches from one tile together before flattening the data, and add a test for this.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -2534,6 +2539,11 @@ def reset_metadata(self) -> None:
             token_count=0, prefill_req_count=0, decode_req_count=0
         )
 
+        # Reset VLM data.
+        self._request_to_image_embeddings.clear()
```
lmcafee-nvidia, line 2543: in `RECOMPUTE` pause/resume mode, this reset erases the saved image positions and image data, but resume restores only the ordinary request information. for example, if a half-finished photo request is paused to free memory and then resumed, its reserved image positions remain but the actual photo data is gone. the request can then fail or continue without using the photo. we should preserve or restore the image data for every resumed request, or reject pause/resume while image requests are present. also a good idea to add a test for this.

megatron/core/inference/engines/dynamic_engine.py
lmcafee-nvidia, line 1: please add tests for the entire PR.

lmcafee-nvidia, line 1: also, we should add support for the client/server. I don't think the client currently has any way of sending image data.

PR #5636, opened 2026-07-03
[Dev] Fix FSDP backward hooks for TE-fused experts

lhb8125:
## Summary

- forward the original `GroupedLinear` post-forward hooks to the TE op-fuser output
- preserve Megatron-FSDP pre-backward parameter all-gathers when fused expert execution bypasses the original submodule `forward()` calls
- cover both regular and `with_kwargs=True` post-forward hooks in the grouped-MLP unit tests

## Root cause

The TE op-fuser path already forwards the original `GroupedLinear` pre-forward hooks, but it bypasses their post-forward hooks. Megatron-FSDP uses those post-forward hooks to attach pre-backward all-gathers. With `optim_grads_params`, expert parameters released after forward could therefore remain unavailable when deferred grouped-wgrad ran, resulting in an illegal CUDA memory access.

The non-op-fuser path continues to invoke `GroupedLinear.forward()` normally and is unaffected.

## Test plan

- `CHECK_ONLY=true BASE_REF=dev bash tools/autoformat.sh` (black, isort, pylint, and ruff passed)
- `python tools/check_copyright.py megatron/core/transformer/moe/experts.py tests/unit_tests/transformer/moe/test_grouped_mlp.py`
- Qwen3-235B, 94 layers, TP1/PP1/EP64, 128 GB200 GPUs, MBS=1, MXFP8, Megatron-FSDP `optim_grads_params`, paged stash, TE op fuser, and full-iteration CUDA graph: job `20260702-082849-85c1` completed 9 stable iterations with finite loss/grad norm, median 861.45 TFLOP/s/GPU, and 92,006 MB peak allocated memory

The CI-faithful target unit-test launch did not reach pytest because the Lyris login node could not import `mcore-ci-dev:latest` without registry/Enroot credentials; upstream CI should execute the added test.

### DIFF
--- a/megatron/core/transformer/moe/experts.py
+++ b/megatron/core/transformer/moe/experts.py
@@ -577,6 +577,7 @@ class TEGroupedMLP(MegatronModule):
 
         # Emulate submodule pre-forward hooks
         ops.register_forward_pre_hook(self._make_fused_impl_pre_forward_hook())
+        ops.register_forward_hook(self._make_fused_impl_post_forward_hook())
 
         return ops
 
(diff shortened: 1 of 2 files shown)


shjwudp (approved):

yaox12 (approved):

lhb8125:
/ok to test [29d72a8](https://github.com/NVIDIA/Megatron-LM/pull/5636/commits/29d72a8af550feca49979045f2894db2c87c1531)

PR #5642, opened 2026-07-03
Set is_first_microbatch when quant_recipe is configured

### LINKED ISSUE #5641: [BUG] TE FP8 cache not refreshed under per-module quant_recipe
When quantization is configured **only** through `--te-precision-config-file` (which sets `config.quant_recipe`), the per-module `is_first_microbatch` flag is never set to `True`, so Transformer Engine modules never refresh their FP8 weight (transpose) cache on the first micro-batch of each step. 

`MegatronModule.set_is_first_microbatch()` only enters its body when `config.fp8`, `config.fp4`, or `use_kitchen` is truthy. 

The `--te-precision-config-file` path (`megatron/training/argument_utils.py:362`) sets only `config.quant_recipe`, leaving `fp8`/`fp4` as `None` and `use_kitchen` as `False`, so the guard is skipped even though FP8 is active per-module.

1. Configure FP8 quantization via a recipe file only (no `--fp8-format`, no kitchen):

   ```bash
   python pretrain_gpt.py ... --te-precision-config-file recipe.yaml
   ```

   where `recipe.yaml` enables FP8 on some modules.
2. Inspect any TE linear layer's `is_first_microbatch` attribute during training — it stays `False` on every step.

**Expected behavior**

With a `quant_recipe` active, `set_is_first_microbatch()` should set `is_first_microbatch = True` on the first micro-batch of each step, exactly as it does for the global `--fp8` path, so TE refreshes its FP8 parameter cache correctly.

**Additional context**

Root cause is the guard condition in `set_is_first_microbatch()` not covering `config.quant_recipe`. Same class of bug as expert-only FP8.

Fix: add `getattr(self.config, 'quant_recipe', None) is not None` to the guard. The flag is a no-op for bf16 modules (TE ignores it outside the FP8 path), so it is safe for mixed bf16/FP8 recipes.

@NVIDIA/mcore-oncall

PR: https://github.com/NVIDIA/Megatron-LM/pull/5642

yezhengmao1:

# What does this PR do?

Enable the `is_first_microbatch` flag when quantization is driven by a
`quant_recipe` (e.g. via `--te-precision-config-file`), not only for the
`fp8` / `fp4` / kitchen paths.

## Issue tracking

Fixes https://github.com/NVIDIA/Megatron-LM/issues/5641

## Problem

`set_is_first_microbatch()` only propagates the flag when `config.fp8`,
`config.fp4`, or `use_kitchen` is set. When quantization is configured
solely through `--te-precision-config-file` (which sets `config.quant_recipe`
while leaving `fp8`/`fp4` as `None` and `use_kitchen` as `False`), the guard
never enters, so per-module `is_first_microbatch` is never set to `True`.

As a result TE modules never refresh their fp8 weight (transpose) cache on
the first micro-batch of each step under per-module quantization — the same
root cause as expert-only fp8.

## Fix

Add `quant_recipe is not None` to the guard. The flag is a no-op for modules
that run in bf16 (TE ignores it outside the fp8 path), so this is safe for
mixed bf16/fp8 recipes and mirrors the existing global `fp8` behavior.

### Pre-checks


Connor-XY:
Thanks—the fix direction looks correct and is appropriately scoped. Before approval, please add a focused regression test covering `fp8=None`, `fp4=None`, `use_kitchen=False`, and a non-`None` `quant_recipe`, asserting that a child module exposing `is_first_microbatch` is updated. Please also retain coverage showing that the flag is not touched when no quantization mode is configured. Once that is added, we can start CI.

yezhengmao1:
Thanks for the review! I've pushed a focused regression test (tests/unit_tests/transformer/test_module.py, class TestSetIsFirstMicrobatch) covering both cases you asked for. @Connor-XY

jaredcasper (approved):

Connor-XY (approved):

Phlip79 (approved):

PR #5661, opened 2026-07-06
[Megatron Lite] Fix qwen3_moe BSHD forward under context parallelism (#5617)

### LINKED ISSUE #5617: [Megatron Lite] CP+PP eval-only BSHD forward passes unsliced labels to cross entropy
**Describe the bug**

In the experimental Megatron-Lite runtime, a Qwen3-MoE eval-only forward with context parallelism and pipeline parallelism enabled together (`CP=2`, `PP=2`) crashes in vocab-parallel cross entropy because the final pipeline stage has CP-local logits for 64 tokens but still receives full-sequence labels for 128 tokens.

The failure is optimizer-backend independent: it reproduces with both Megatron distributed optimizer (`dist_opt`) and PyTorch FSDP2. 

Scope: this report is for the `use_thd=false` BSHD path. The THD protocol has separate packing/CP-splitting logic and is not claimed to be affected.

Environment used:

- 1 node / 8 GPUs
- Python 3.12
- Qwen3-30B-A3B-Instruct-2507 config/weights
- 2 model layers retained for the smoke test
- `TP=1, ETP=1, EP=1, CP=2, PP=2`
- sequence length 128, 2 microbatches

From a checkout containing `experimental/lite`:

```bash
export PYTHONPATH="$PWD/experimental/lite:/path/to/Megatron-LM"
export MEGATRON_LITE_DETERMINISTIC=1
export CUBLAS_WORKSPACE_CONFIG=:4096:8

torchrun --nnodes 1 --nproc_per_node 8 --master_port 29771 \
  experimental/lite/examples/bench/correctness.py run \
  --backend mlite \
  --hf-path /path/to/Qwen3-30B-A3B-Instruct-2507 \
  --model-name qwen3_moe \
  --impl lite \
  --tp 1 --etp 1 --ep 1 --pp 2 --cp 2 \
  --steps 1 --num-microbatches 2 --seq-len 128 \
  --seed 42 --truncate-layers 2 --same-data-across-dp \
  --override-optimizer-json '{"use_precision_aware_optimizer":false}' \
  --skip-weight-hash \
  --impl-cfg-json '{"optimizer":"fsdp2"}' \
  --output-json /tmp/fsdp2-cp2-pp2.json
```

The correctness runner performs eval-only forward before the training step, so the crash occurs immediately. Replacing the last argument with `{"optimizer":"dist_opt"}` produces the same failure.

Representative stack from FSDP2 (ranks 4-7, the last PP stage):

```text
correctness.py:338 in run_backend
  eval_logits = _hash_tensor(_forward_logits(rt, handle, eval_batch))
runtime/backends/mlite/runtime.py:445 in forward_backward
  outputs = forward_backward_pipelining(... forward_only=True)
primitive/parallel/pipeline.py:555 in _forward_only_pipeline_schedule
  out = _run_pipeline_chunk_forward(...)
model/qwen3_moe/lite/protocol.py:137 in _forward_step_bshd
  return model(input_ids=batch.input_ids.reshape(1, -1), labels=labels, ...)
model/qwen3_moe/lite/model.py:518 in forward
  token_loss = vocab_parallel_cross_entropy(logits, labels_sb, self.ps.tp_group)
primitive/ops/cross_entropy.py:53 in forward
  predicted_logits_1d = logits_2d[arange_1d, masked_target_1d]
IndexError: shape mismatch: indexing tensors could not be broadcast together with shapes [64], [128]
```

The apparent data flow is:

1. The bench supplies a raw model-agnostic `PackedBatch` with 128 input IDs and 128 labels.
2. With `PP>1`, `_infer_pipeline_tensor_shape()` sizes PP activations using the CP-local sequence length (`128 / 2 = 64`).
3. With `use_thd=false`, Qwen3-MoE selects `_forward_step_bshd()`, which reshapes and forwards the full `batch.labels` without applying the CP layout.
4. The final stage produces logits for 64 local tokens, while cross entropy receives 128 targets.

As controls, FSDP2 CP-only (`CP=2, PP=1`) and PP-only (`CP=1, PP=2`) smoke runs complete; the failure appears when CP and PP are enabled together in this eval-only path.

**Expected behavior**

Eval-only forward with `CP=2, PP=2` should apply the same CP token layout to labels/loss masks as to the final-stage activations, and should complete for both `dist_opt` and FSDP2.

At minimum, unsupported BSHD+CP combinations should fail early with a clear validation error rather than reaching cross entropy with inconsistent shapes.

**Additional context**

The relevant shared BSHD forward currently passes labels unchanged:

```python
def _forward_step_bshd(model: nn.Module, batch: PackedBatch) -> dict:
    labels = batch.labels.reshape(1, -1) if batch.labels is not None else None
    return model(input_ids=batch.input_ids.reshape(1, -1), labels=labels, packed_seq_params=None)
```

The THD forward instead calls `pack_thd_forward_kwargs()`, whose packing path CP-splits tokens, labels, masks, and position IDs. A likely fix is to give the BSHD path equivalent CP-aware label/loss-mask handling, or explicitly reject BSHD+CP until that layout is supported.

conver334:
## What

Fixes #5617.

In the experimental Megatron-Lite runtime, a Qwen3-MoE eval-only BSHD (`use_thd=false`) forward with context parallelism and pipeline parallelism enabled together (`CP>1`, `PP>1`) crashes in vocab-parallel cross entropy: the final pipeline stage produces CP-local logits but still receives full-sequence labels.

## Root cause

`_forward_step_bshd()` forwarded `batch.input_ids` / `batch.labels` unchanged, but:

- `GQAttention`'s non-THD branch (and its RoPE frequency slicing) assumes its input is **already CP-zigzag-sharded** (see the "q is CP-zigzag pre-sliced" comment in `primitive/modules/gqa.py`).
- The pipeline runtime's `_infer_pipeline_tensor_shape()` sizes PP activations from the **CP-local** sequence length.

So with `CP>1` the last PP stage's CP-local logits disagreed in shape with the still-full-sequence labels, crashing `vocab_parallel_cross_entropy` (and, with `PP=1`, silently misaligning the loss). Only the THD path did CP splitting (via `pack_thd_forward_kwargs`); the BSHD input side was missing it.

## Fix

`_forward_step_bshd()` now zigzag-splits `input_ids` / `labels` / `loss_mask` with `zigzag_slice_for_cp()` — the same primitive other models (`mla.py`, `dsa.py`, `gated_delta_net.py`) already use for CP — so the input layout matches what attention/RoPE expect. `unpack_forward_output()` already reconstructs zigzag-CP outputs back to full-sequence order via `unpack_thd_forward_output()`, so only the input side needed the change.

## Testing

- Existing `tests/unit/model/test_qwen_config_unit.py` passes (CPU-only, no GPU required).
- Numerically validated separately: BSHD `CP=2, PP=2` eval-only forward now completes and matches the CP-off baseline for both `dist_opt` and FSDP2.

## Scope

BSHD-only (`use_thd=false`) path, as reported. The THD path already handled CP and is unchanged.

### DIFF
--- a/experimental/lite/megatron/lite/model/qwen3_moe/lite/protocol.py
+++ b/experimental/lite/megatron/lite/model/qwen3_moe/lite/protocol.py
@@ -44,7 +44,8 @@ from megatron.lite.primitive.modules.lora import (
     normalize_lora_config,
     trainable_param_stats,
 )
-from megatron.lite.primitive.parallel import ParallelState, init_parallel
+from megatron.lite.primitive.parallel import ParallelState, init_parallel, zigzag_slice_for_cp
+from megatron.lite.primitive.parallel.thd import parallel_state_from_model
 from megatron.lite.primitive.recompute import apply_recompute, parse_recompute_spec
 from megatron.lite.runtime.contracts import OptimizerConfig, ParallelConfig
 from megatron.lite.runtime.contracts.data import PackedBatch
@@ -133,8 +134,21 @@ def _forward_step(model: nn.Module, batch: PackedBatch) -> dict:
 
 
 def _forward_step_bshd(model: nn.Module, batch: PackedBatch) -> dict:
+    ps = parallel_state_from_model(model) or ParallelState()
+    input_ids = batch.input_ids.reshape(1, -1)
     labels = batch.labels.reshape(1, -1) if batch.labels is not None else None
-    return model(input_ids=batch.input_ids.reshape(1, -1), labels=labels, packed_seq_params=None)
+    loss_mask = batch.loss_mask.reshape(1, -1) if batch.loss_mask is not None else None
+    if ps.cp_size > 1:
+        # GQAttention's non-THD branch expects CP-zigzag-sharded inputs, and
+        # unpack_forward_output() reconstructs zigzag-CP outputs; match both.
+        input_ids = zigzag_slice_for_cp(input_ids, ps.cp_rank, ps.cp_size, seq_dim=1)
+        if labels is not None:
+            labels = zigzag_slice_for_cp(labels, ps.cp_rank, ps.cp_size, seq_dim=1)
+        if loss_mask is not None:
+            loss_mask = zigzag_slice_for_cp(loss_mask, ps.cp_rank, ps.cp_size, seq_dim=1)
+    return model(
+        input_ids=input_ids, labels=labels, loss_mask=loss_mask, packed_seq_params=None
+    )
 
 
 def unpack_forward_output(model: nn.Module, batch: PackedBatch, output) -> Any:


experimental/lite/megatron/lite/model/qwen3_moe/lite/protocol.py
```diff
@@ -133,8 +134,21 @@ def _forward_step(model: nn.Module, batch: PackedBatch) -> dict:
...
 def _forward_step_bshd(model: nn.Module, batch: PackedBatch) -> dict:
+    ps = parallel_state_from_model(model) or ParallelState()
+    input_ids = batch.input_ids.reshape(1, -1)
     labels = batch.labels.reshape(1, -1) if batch.labels is not None else None
-    return model(input_ids=batch.input_ids.reshape(1, -1), labels=labels, packed_seq_params=None)
+    loss_mask = batch.loss_mask.reshape(1, -1) if batch.loss_mask is not None else None
+    if ps.cp_size > 1:
+        # GQAttention's non-THD branch expects CP-zigzag-sharded inputs, and
+        # unpack_forward_output() reconstructs zigzag-CP outputs; match both.
+        input_ids = zigzag_slice_for_cp(input_ids, ps.cp_rank, ps.cp_size, seq_dim=1)
```
guihong-nv, line 144: The BSHD shard length here can diverge from both pipeline buffer sizing and output reconstruction. This slices the unpadded flattened sequence and only requires divisibility by `2 * CP`, but `_infer_pipeline_tensor_shape()` and `unpack_thd_forward_output()` derive a layout padded per sequence to `TP * 2 * CP`. For example, with `S=12`, `TP=2`, `CP=2`, and `PP=2`, the model sends a three-token sequence-parallel activation while the receiver allocates four tokens; with `PP=1`, unpacking interprets a six-token CP shard using an eight-token layout. Please either pad the BSHD tensors using the same metadata before slicing, or make pipeline sizing and unpacking BSHD-aware and validate the supported alignment. A `TP>1` or unaligned-length regression test should cover this.

experimental/lite/megatron/lite/model/qwen3_moe/lite/protocol.py
```diff
@@ -44,7 +44,8 @@
     normalize_lora_config,
     trainable_param_stats,
 )
-from megatron.lite.primitive.parallel import ParallelState, init_parallel
+from megatron.lite.primitive.parallel import ParallelState, init_parallel, zigzag_slice_for_cp
```
guihong-nv, line 47: This file fails the repository-mandated `isort --check-only` check. Running isort inserts a blank line after `import torch.nn as nn` before the local `megatron` imports. Please run isort on the touched Python file.

guihong-nv (commented):
Please address the BSHD alignment mismatch before merging. The current validation covers the reported `TP=1`, `S=128` case but not `TP>1` or unaligned sequence lengths. I also noted the import-format check failure inline.

PR #5679, opened 2026-07-06
[Core] Add dynamic context-parallel THD scheduling

ilml:
Training loss:
image

Throughput:
image

## Summary

- Add a main-native sequence-packing scheduler for fixed context parallelism (`dp_balanced`) and dynamic context parallelism (`default_dynamic_cp`).
- Add dynamic CP process groups, per-microbatch CP sizing, and pipeline-aware scheduling for packed THD batches.
- Propagate packed-sequence metadata through attention, RoPE, Transformer Engine, and tensor-parallel broadcasts while restoring the configured process groups after each microbatch.
- Add tokenizer chat-template keyword forwarding and an SFT mock-data path for scheduler testing.

## User-facing flags

| Flag | Behavior and constraints |
| --- | --- |
| `--dynamic-context-parallel` | Enables per-microbatch CP resizing for THD batches. Requires `--calculate-per-token-loss`, `--max-seqlen-per-dp-cp-rank`, `--dataloader-type single`, and `--cuda-graph-impl none`. Megatron FSDP, MoE, MLA, and hybrid/Mamba entrypoints are not supported. |
| `--min-dynamic-context-parallel-size N` | Sets the smallest allowed CP size. The DP x CP domain must be a power of two unless `N` selects the full domain. |
| `--sequence-packing-scheduler default_dynamic_cp` | Uses dynamic CP scheduling. Packed capacity is `DP x CP x --max-seqlen-per-dp-cp-rank`. It is selected automatically by `--dynamic-context-parallel`. |
| `--sequence-packing-scheduler dp_balanced` | Balances packed work while keeping the configured CP size fixed. MoE requires all-to-all token dispatch. |
| `--hybrid-context-parallel` | Deprecated compatibility alias for dynamic CP. New recipes should use `--dynamic-context-parallel`. |
| `--max-seqlen-per-dp-cp-rank N` | Sets the packed-token capacity assigned to each scheduler rank. It is required whenever a sequence-packing scheduler is active. |
| `--sft-mock-dataset-config-json VALUE` | Supplies inline JSON or a JSON file for synthetic SFT sequence lengths when using `--sft --mock-data`. |

Dynamic CP requires Transformer Engine 2.9 or newer.

## Stack and scope

- PR #5681 is stacked on this PR and adds variable-length datasets and the Nemotron Math v3 loader.
- PR #5811 is the runnable baseline-versus-DCP benchmark harness stacked on #5681.
- The implementation incorporates the training path demonstrated in [xiaoyao0115's branch](https://github.com/xiaoyao0115/Megatron-LM/tree/tailaim/varlen-math-v3-pr5541).
- PR #5541 remains useful historical context for the earlier no-CUDA-graph subset. CUDA graph support from the coworker's full development branch is intentionally outside this PR.

## Validation

- `uv run isort --check-only` on changed Python files
- `uv run ruff check` on changed Python files
- Python byte compilation and `git diff --check`
- Signed commit verified locally: `533deb7b7e68a7ed5daa988fcfdbe0af73172cc1`

The available environment does not have a usable PyTorch test installation and has only two GPUs, so the distributed numerical tests were not run locally. CI was requested on the signed commit.

## Summary by CodeRabbit

* **New Features**
  * Added sequence-packing schedulers to improve variable-length workload balancing and GPU utilization.
  * Added dynamic context parallelism with configurable scheduling and minimum group sizes.
  * Added support for passing custom chat-template options, including tool definitions, during SFT tokenization.
  * Added configurable mock SFT datasets for testing and development.

* **Bug Fixes**
  * Improved padding-mask handling across sequence and tensor parallel execution.
  * Preserved accurate token and sequence-length statistics for packed batches.

* **Documentation**
  * Added guidance for sequence-packing schedulers and dynamic context parallelism.


megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +351,580 @@ def __next__(self) -> Any:
...
+                else:
+                    i -= 1
+
+        num_micro_batches = int(len(packed_id_groups) / self.dp_size)
+        for i in range(num_micro_batches):
+            sample_id_groups.append([])
+            for j in range(self.cp_size * self.dp_size):
+                seq_id = int(i * self.dp_size + j / self.cp_size)
+                sample_id_groups[i].append(packed_id_groups[seq_id])
+        return sample_id_groups
```
guihong-nv, line 209: The scheduler demands an exact multiple of DP × microbatch_group_size_per_vp_stage, but the interleaved schedule permits a final remainder ≥ PP. The documented PP=2, group-size=3, M=5 case with five capacity-sized singleton sequences tries to create a sixth group and raises Not enough samples to move. Apply the pipeline schedule’s remainder rule instead.

megatron/core/datasets/data_schedule_utils.py
```diff
@@ -0,0 +1,473 @@
...
+        ``cu_seqlens`` tensor and the tokens of multiple sub-samples
+        concatenated together. We slice them apart and synthesize
+        ``original_seq_len`` / ``padded_seq_len`` from the cu_seqlens deltas.
+
+      * **Already unpacked**: each sample is a single sub-sample that already
+        carries ``padded_seq_len`` (and usually ``original_seq_len``). We just
+        normalize an optional leading batch dimension and return it as-is.
+    """
+    # Short-circuit for datasets that already emit one sub-sample per index.
+    if batch and "padded_seq_len" in batch[0]:
```
guihong-nv, line 47: For already-unpacked inputs, _unpack_batch returns iterator-owned dictionaries. Pipeline filtering then deletes keys in place, while RerunDataIterator replays the same saved references. A PP rerun consequently fails the required-key assertion. Copy samples before normalization/filtering.

megatron/core/datasets/data_schedule.py
```diff
@@ -299,3 +351,580 @@ def __next__(self) -> Any:
...
+        Returns:
+            new_data_iterator: The new data iterator (or list for VPP).
+            num_micro_batches: Number of micro batches after scheduling.
+            seqlen_sum_this_global_batch: Total tokens for FLOPs calculation.
+            seqlen_squared_sum_this_global_batch: Sum of squared seqlens for FLOPs.
+        """
+        raise NotImplementedError
+
+
+class DpBalancedScheduler(BasePackingScheduler):
```
parthmannan, line 132: In our offline discussion, we spoke to de-duplicating some of the code changes here that already exist. Most of the logic for `DpBalancedScheduler` exists in `HybridCPDataLoaderWrapper`.

Can we re-use the functions from `HybridCPDataLoaderWrapper` instead of introducing in utils file OR remove all or most of `HybridCPDataLoaderWrapper` in favor of moving to DpBalancedScheduler for default Hybrid CP implementation?

parthmannan (changes requested):

megatron/core/datasets/data_schedule.py
ilml: Addressed by taking the second option in the updated commit `533deb7b7`. `HybridCPDataLoaderWrapper` has been removed entirely. Its unpack, global-length gather, and reroute operations now live once as scheduler-neutral helpers in `data_schedule_utils.py`; `DpBalancedScheduler` and `DefaultDynamicCPScheduler` share the same `run()` pipeline through `wrap_data_iterator()`.

The deprecated `--hybrid-context-parallel` option is normalized to dynamic CP with `sequence_packing_scheduler=default_dynamic_cp`, so it no longer selects the wrapper or the legacy forward/backward path. I also added a focused regression test for that compatibility mapping. The old `hybrid_cp_schedule.py` module remains importable during the deprecation window, but it has no production caller after this change.

megatron/core/datasets/data_schedule.py
ilml: Addressed in `533deb7b7`. `DpBalancedScheduler` now aligns to DP first and then applies the pipeline schedule remainder rule: a final VPP group is preserved when its remainder is at least the PP size, and it is expanded only when the remainder is smaller than PP. `wrap_data_iterator()` passes the actual PP size into the scheduler. The regression `test_dp_balanced_scheduler_allows_schedulable_vpp_final_remainder` covers the exact PP=2, group-size=3, M=5 case with five capacity-sized singleton sequences and verifies that it remains five groups.

megatron/core/datasets/data_schedule_utils.py
ilml: Addressed in `533deb7b7`. `_unpack_batch()` now copies every sample dictionary before normalizing dimensions or applying PP-stage field filtering, so deletes and assignments only affect scheduler-owned copies and the `RerunDataIterator` references remain intact. `test_unpack_batch_does_not_mutate_replay_owned_sample` covers dictionary identity, source tensor shape, and synthesized metadata ownership.

yaoyu-33 (approved):

megatron/core/models/gpt/gpt_model.py
Phlip79, line 1: Can you make the equivalent change to `hybrid_model.py`?

PR #5680, opened 2026-07-06
build: Update Transformer Engine to 2.17

balasaajay:
## Summary

- update Transformer Engine to 2.17.0
- pin the `release_v2.17` branch commit `2e559f062497bef768dfbe9d7e45548fadeca80a`
- refresh the corresponding `uv.lock` entries

## Validation

- `uv lock --check --offline --no-build` (inside `nvcr.io/nvidia/pytorch:26.04-py3`)
- `git diff --check`


YangFei1990 (changes requested):
Can we also update the NCCL to the third party NCCL inside of TE? NCCL EP will require that NCCL to run.

balasaajay:
> Can we also update the NCCL to the third party NCCL inside of TE? NCCL EP will require that NCCL to run.

@YangFei1990  we can install custom NCCL in the docker image if needed for any feature or to fix an issue. currently, NCCL from the base pytorch image is used.

YangFei1990 (approved):

ko3n1g (approved):

Phlip79 (approved):

PR #5684, opened 2026-07-07
fix: save chained optimizer state in torch checkpoints

FDecaYed:
## Summary

- Move the per-rank torch checkpoint save/load helpers from `LayerWiseDistributedOptimizer` to `ChainedOptimizer`.
- Save and restore the complete chained state so both LayerWise and DistributedOptimizer state are preserved.
- Add a round-trip regression test for a two-optimizer chain.

## Root cause

The non-Muon parameter routing change can construct a top-level `ChainedOptimizer` for LayerWise torch checkpoints. The checkpoint path called helpers implemented only on `LayerWiseDistributedOptimizer`, causing save and resume to fail with `AttributeError`.

## Validation

- `ruff check` on changed Python files
- `black --check` on changed Python files
- `isort --check-only` on changed Python files
- `compileall`
- `git diff --check`

The distributed GPU unit test was not run locally because this host has no PyTorch or supported container runtime.


tests/unit_tests/test_optimizer.py
```diff
@@ -326,6 +326,46 @@ def to_cuda(d):
...
+            return self._state_dict
+
+        def load_state_dict(self, state_dict):
+            self._state_dict = state_dict
+
+    state_dicts = [
+        {'optimizer': {'state': {'layer_wise': 'state'}, 'param_groups': []}},
+        {'optimizer': {'state': {'distributed': 'state'}, 'param_groups': []}},
+    ]
+    checkpoint_path = tmp_path / f'optimizer_{os.getpid()}.pt'
```
skyw, line 351: Does megatron require using tempfile in tests?

skyw (approved):

Wohox (approved):
LGTM!

Autumn1998 (approved):

PR #5707, opened 2026-07-08
Fix layer-wise (Muon) distributed optimizer + fp8-param

fanshiqing:

# What does this PR do?

Support Muon (LayerWiseDistributedOptimizer) with mxfp8 fp8-param-gather
fp8-param-gather relies on the DistributedOptimizer param-buffer machinery, which
LayerWiseDistributedOptimizer (Muon) does not have, so the optimizer's weight
update never reached the forward and loss stayed flat. Wire the missing pieces so
Muon trains with --fp8-param-gather + --reuse-grad-buf-for-mxfp8-param-ag at both
DP=1 and DP>1:

- optimizer.py: the reuse_grad_buf (master -> fp8 param buffer) copy only applies
  to DistributedOptimizer; gate it on _copy_main_params_to_param_buffer so
  non-DistOpt optimizers (Muon's Float16 base opts) fall through to
  master -> model params. The forward then reads the update.

- param_and_grad_buffer.py: in the LayerWise variable-size param all-gather, stage
  the high-precision fp32 master (param.main_param, cast to bf16) for MXFP8 params
  instead of flattening the MXFP8 wrapper (view(-1) is unsupported); the receive
  side re-quantizes via param.data.copy_. Also skip _post_param_sync for buckets
  with no param_data buffer (the LayerWise path already updated params in place).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -168,7 +168,7 @@ build = [
 ]
 linting = [
     "ruff~=0.9.0",
-    "black==24.4.2",
+    "black==26.3.0",
     "isort==5.13.2",
     "flake8==7.1.0",
     "pylint==3.2.6",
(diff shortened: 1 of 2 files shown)


fanshiqing:
Close this MR as this fix (https://github.com/fanshiqing/Megatron-LM/commit/1b0121d052400d8d1d55f6e0ff07e4ba1f0e6ebf) has already been merged with https://github.com/NVIDIA/Megatron-LM/pull/4967/ together.

Issue #2769, opened 2025-12-28
Correctness Issue with Torch FSDP2


haok1402:
Hi, I'm using `core_v0.15.0` release with Megatron-LM and running a mid-training from the Llama3-8B checkpoint. 

Exact same configuration, using the Megatron's `use-distributed-optimizer`, the loss starts off at 2.404998E+00, which is a correct range for a pretrained model.

```
 [2025-12-28 00:20:24] iteration        1/     100 | consumed samples:         1024 | elapsed time per iteration (ms): 27003.9 | throughput per GPU (TFLOP/s/GPU): 452.7 | learning rate: 9.900000E-05 | global batch size:  1024 | lm loss: 2.404998E+00 | loss scale: 1.0 | grad norm: 3.048 | number of skipped iterations:   0 | number of nan iterations:   0 |Number of parameters in transformer block in billions:  6.98 
 [2025-12-28 00:20:49] iteration        2/     100 | consumed samples:         2048 | elapsed time per iteration (ms): 24621.6 | throughput per GPU (TFLOP/s/GPU): 496.6 | learning rate: 9.800000E-05 | global batch size:  1024 | lm loss: 4.466884E+00 | loss scale: 1.0 | grad norm: 44.051 | number of skipped iterations:   0 | number of nan iterations:   0 |
```

However, when I try to use torch FSDP2 via `use-torch-fsdp2`, the loss starts off with at 1.386388E+01, which isn't a correct range...
 
```
 [2025-12-28 00:23:52] iteration        1/     100 | consumed samples:         1024 | elapsed time per iteration (ms): 37835.9 | throughput per GPU (TFLOP/s/GPU): 323.1 | learning rate: 9.900000E-05 | global batch size:  1024 | lm loss: 1.386388E+01 | loss scale: 1.0 | grad norm: 346.039 | number of skipped iterations:   0 | number of nan iterations:   0 |
 [2025-12-28 00:24:26] iteration        2/     100 | consumed samples:         2048 | elapsed time per iteration (ms): 34441.6 | throughput per GPU (TFLOP/s/GPU): 355.0 | learning rate: 9.800000E-05 | global batch size:  1024 | lm loss: 3.159755E+01 | loss scale: 1.0 | grad norm: 544.112 | number of skipped iterations:   0 | number of nan iterations:   0 |
```

The full command line arguments are attached below for reference.

```bash
#!/bin/bash

declare -A MODEL_CONFIG
[...]

PR #5714, opened 2026-07-08
Fix FSDP2 SwiGLU checkpointing.

cspades:

# What does this PR do?

- Fixes a bug (https://github.com/NVIDIA/Megatron-LM/issues/2769) with `torch_dist` checkpointing for FSDP2, where the MLP / MoE SwiGLU weights need to be TP-DP sharded. To remind everyone, the checkpoint files have `[W; V]` weights, and then we:
  - Swizzle / interleave the TP shards, i.e. `[W_tp0, V_tp0, W_tp1, V_tp1, ... ]`
  - For each TP shard pair of W/V, FSDP2 then shards `dim=0`, so you end up with `[ {W_tp0; V_tp0}_dp{1...DP}, {W_tp1; V_tp1}_dp{1...DP}, ... ]`, which is really hard to read but basically the DP sharding is done over the pair for each TP rank, and this is mapped from the simple `[W; V]` checkpointed format.

After making this fix, we have loss parity after loading a ND-parallel `torch_dist` checkpoint (non-trivial):
```
# ND Post-Load
 [2026-07-07 14:43:32.368419] iteration       51/     100 | consumed samples:          816 | elapsed time per iteration (ms): 42941.8 | throughput per GPU (TFLOP/s/GPU): 4.4 | learning rate: 9.900000E-05 | global batch size:    16 | lm loss: 3.642244E-02 | loss scale: 1.0 | grad norm: 0.913 | number ofskipped iterations:   0 | number of nan iterations:   0 |Number of parameters in transformer block in billions:  6.98

# Torch FSDP2 Post-Load
 [2026-07-07 14:44:43.435452] iteration       51/     100 | consumed samples:          816 | elapsed time per iteration (ms): 12756.0 | throughput per GPU (TFLOP/s/GPU): 15.0 | learning rate: 9.900000E-05 | global batch size:    16 | lm loss: 3.642244E-02 | loss scale: 1.0 | grad norm: 0.913 | number of skipped iterations:   0 | number of nan iterations:   0 |Number of parameters in transformer block in billions:  6.98
```
Debugged this with my usual activation bisect approach after confirming that `FSDPParam._sharded_param_data` was indeed updated by the distributed checkpoint load. It was just the middle sections of the SwiGLU weights that were in the wrong order.

Attached my minimal script in the Testing section.

# Testing

- Unit test on the `ShardedTensorFactory` built/acting on the `ShardedTensor`-converted `DTensor` parameters from FSDP2.

To reproduce the bug...
```
#!/bin/bash

declare -A MODEL_CONFIG
MODEL_CONFIG[tokenizer-type]=HuggingFaceTokenizer
MODEL_CONFIG[tokenizer-model]=meta-llama/Meta-Llama-3-8B
MODEL_CONFIG[untie-embeddings-and-output-weights]=true
MODEL_CONFIG[position-embedding-type]=rope
MODEL_CONFIG[rotary-base]=500000
MODEL_CONFIG[max-position-embeddings]=8192
MODEL_CONFIG[num-layers]=32
MODEL_CONFIG[hidden-size]=4096
MODEL_CONFIG[ffn-hidden-size]=14336
MODEL_CONFIG[hidden-dropout]=0.0
MODEL_CONFIG[disable-bias-linear]=true
MODEL_CONFIG[swiglu]=true
MODEL_CONFIG[group-query-attention]=true
MODEL_CONFIG[num-attention-heads]=32
MODEL_CONFIG[num-query-groups]=8
MODEL_CONFIG[kv-channels]=128
MODEL_CONFIG[attention-dropout]=0.0
MODEL_CONFIG[init-method-std]=0.02
MODEL_CONFIG[normalization]=RMSNorm
MODEL_CONFIG[norm-epsilon]=1e-5

declare -A TRAIN_CONFIG
TRAIN_CONFIG[lr]=1e-4
TRAIN_CONFIG[optimizer]=adam
TRAIN_CONFIG[adam-beta1]=0.9
TRAIN_CONFIG[adam-beta2]=0.95
TRAIN_CONFIG[weight-decay]=0.1
TRAIN_CONFIG[clip-grad]=1.0
TRAIN_CONFIG[train-iters]=200
TRAIN_CONFIG[micro-batch-size]=1
TRAIN_CONFIG[global-batch-size]=16
TRAIN_CONFIG[seq-length]=2048
TRAIN_CONFIG[bf16]=true
TRAIN_CONFIG[mock-data]=true
TRAIN_CONFIG[vocab-size]=128256

# TRAIN_CONFIG[use-distributed-optimizer]=true
TRAIN_CONFIG[use-torch-fsdp2]=true
TRAIN_CONFIG[no-gradient-accumulation-fusion]=true
# FSDP4 TP2 works!
TRAIN_CONFIG[tensor-model-parallel-size]=2

TRAIN_CONFIG[log-interval]=1
TRAIN_CONFIG[no-one-logger]=true
TRAIN_CONFIG[log-throughput]=true
TRAIN_CONFIG[load]=checkpoints/debug/megatron/  # torch_dist checkpoint -> FSDP2 model
TRAIN_CONFIG[save]=checkpoints/debug/megatron/
TRAIN_CONFIG[no-load-optim]=true
TRAIN_CONFIG[save-interval]=100
TRAIN_CONFIG[eval-interval]=250
TRAIN_CONFIG[eval-iters]=1

MAIN_ARGS=()
for key in ${!MODEL_CONFIG[@]}; do val=${MODEL_CONFIG[$key]}; [[ $val == true ]] && MAIN_ARGS+=(--$key) || MAIN_ARGS+=(--$key $val); done
for key in ${!TRAIN_CONFIG[@]}; do val=${TRAIN_CONFIG[$key]}; [[ $val == true ]] && MAIN_ARGS+=(--$key) || MAIN_ARGS+=(--$key $val); done

TRUN_ARGS=()
TRUN_ARGS+=(--nnodes=1 --nproc-per-node=8)
TRUN_ARGS+=(--rdzv-backend=c10d --rdzv-endpoint=localhost:15213)

SCRIPT=pretrain_gpt.py
STDOUT=checkpoints/debug/midtrain.log

mkdir -p $(dirname $STDOUT)
export WANDB_RUN_ID=$(echo ${TRAIN_CONFIG[WANDB_EXP_NAME]} | md5sum | cut -c1-5)
uv run python -m torch.distributed.run ${TRUN_ARGS[@]} $SCRIPT ${MAIN_ARGS[@]}
```

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

### DIFF
--- a/megatron/core/distributed/torch_fully_sharded_data_parallel.py
+++ b/megatron/core/distributed/torch_fully_sharded_data_parallel.py
@@ -96,6 +96,8 @@ class TorchFullyShardedDataParallel(_BaseDataParallel):
                     # micro-batch id, thus removing unnecessary memory stores
                     attrs['_fp8_attrs']['transpose_invalid'] = False
                     del attrs['_fp8_attrs']['transpose']
+                # Mark this parameter as an FSDP2 parameter.
+                attrs["is_torch_fsdp2_param"] = True
                 custom_attrs[name] = {k: v for k, v in attrs.items()}
             return custom_attrs
 
--- a/megatron/core/transformer/mlp.py
+++ b/megatron/core/transformer/mlp.py
@@ -435,13 +435,14 @@ def apply_swiglu_sharded_factory(
     )
     axis_frag = original_sh_ten.axis_fragmentations[swiglu_shard_axis + prepend_axis_num]
 
-    # Detect TP-DP strided sharding (FSDP), in which case we need to map [W; V] to
-    # [ {W_tp0; V_tp0}_dp{1...DP}, {W_tp1; V_tp1}_dp{1...DP}, ... ]
-    # TP sharding produces TP-sharded pairs of W/V, followed by DP sharding!
-    assert tp_group is not None and dp_group is not None
-    tp_size = get_pg_size(tp_group)
-    dp_size = get_pg_size(dp_group)
-    is_dp_sharded = getattr(original_sh_ten, "is_torch_fsdp2_param", False) and dp_size > 1
+    # Only FSDP2 supports torch_dist ShardedTensor. (Add other DP sharding algos here if needed.)
+    is_torch_fsdp2_param = getattr(original_sh_ten, "is_torch_fsdp2_param", False)
+    if is_torch_fsdp2_param:
+        assert dp_group is not None
+        dp_size = get_pg_size(dp_group)
+        is_dp_sharded = dp_size > 1
+    else:
+        is_dp_sharded = False
 
     @torch.no_grad()
     def sh_ten_build_fn(
@@ -486,10 +487,17 @@ def apply_swiglu_sharded_factory(
         ]
 
     @torch.no_grad()
-    def fsdp2_sh_ten_build_fn(
+    def dp_sh_ten_build_fn(
         key: str, t: torch.Tensor, replica_id: ReplicaId, flattened_range: Optional[slice]
     ):
-        # FSDP2 shards the TP-local [W; V] SwiGLU FC1 tensor over DP along dim 0.
+        assert not singleton_local_shards, (
+            "FSDP does not support singleton ShardedTensor for SwiGLU fused FC1. "
+            "Set singleton_local_shards=False, which is the default in MCore."
+        )
+        # FSDP shards the TP-local [W; V] SwiGLU FC1 tensor over DP along dim 0.
+        # TP sharding produces TP-sharded pairs of W/V, followed by DP sharding!
+        assert tp_group is not None and dp_group is not None
+        tp_size = get_pg_size(tp_group)
         global_axis = swiglu_shard_axis + prepend_axis_num
         tp_rank = get_pg_rank(tp_group)
         dp_rank = get_pg_rank(dp_group)
@@ -500,8 +508,8 @@ def apply_swiglu_sharded_factory(
         half_axis_size = tp_local_axis_size // 2
         # Check that the TP-local W or V is cleanly divisible by DP.
         assert half_axis_size % local_axis_size == 0, (
-            "SwiGLU FSDP2 torch_dist checkpoint loading requires each DP shard of "
-            "linear_fc1 to fall wholly inside either the W or V half."
+            "SwiGLU FC1 FSDP ShardedTensor requires each DP shard of "
+            "linear_fc1 to be completely inside either the W or V half."
         )
         # Number of DP shards per W or V TP-shard, and make sure
         # that DP sharding spans both W and V.
@@ -512,13 +520,13 @@ def apply_swiglu_sharded_factory(
         swiglu_half_idx, half_dp_shard_idx = divmod(dp_rank, shards_per_half)
         # If W, then 0. If V, then 1.
         assert swiglu_half_idx in (0, 1)
-        # Map [W; V] to this rank's shard [W_tpx_dpy; V_tpx_dpy].
+        # Map [ W; V ] to this rank's shard [ {W_tpx; V_tpx}_dpy ].
         shard_rank_offset = (
-            # W or V
+            # W or V half of the [W; V] global data.
             swiglu_half_idx * tp_size * shards_per_half
-            # TP Shard
+            # TP Shard Index
             + tp_rank * shards_per_half
-            # TP-DP Shard
+            # TP-DP Shard Index
             + half_dp_shard_idx
         )
 
@@ -534,7 +542,7 @@ def apply_swiglu_sharded_factory(
         ]
 
     # Construct a ShardedTensorFactory.
-    sh_ten_factory_build_function = fsdp2_sh_ten_build_fn if is_dp_sharded else sh_ten_build_fn
+    sh_ten_factory_build_function = dp_sh_ten_build_fn if is_dp_sharded else sh_ten_build_fn
     return ShardedTensorFactory(
         original_sh_ten.key,
         original_sh_ten.data,
--- a/megatron/core/transformer/mlp.py
+++ b/megatron/core/transformer/mlp.py
@@ -26,9 +26,15 @@ from megatron.core.fusions.fused_bias_swiglu import bias_swiglu_impl, weighted_b
 from megatron.core.process_groups_config import ProcessGroupCollection
 from megatron.core.transformer.module import MegatronModule
 from megatron.core.transformer.transformer_config import TransformerConfig
-from megatron.core.transformer.utils import cat_with_oom_fallback, sharded_state_dict_default
+from megatron.core.transformer.utils import (
+    cat_with_oom_fallback,
+    ensure_metadata_has_dp_cp_group,
+    sharded_state_dict_default,
+)
 from megatron.core.typed_torch import apply_module, not_none
 from megatron.core.utils import (
+    get_pg_rank,
+    get_pg_size,
     get_tensor_model_parallel_group_if_none,
     nvtx_range_pop,
     nvtx_range_push,
@@ -355,6 +361,7 @@ class MLP(MegatronModule):
         self, prefix: str = "", sharded_offsets: tuple = (), metadata: Optional[dict] = None
     ) -> ShardedStateDict:
         """Return the sharded state dictionary of the module."""
+        metadata = ensure_metadata_has_dp_cp_group(metadata)
         sharded_state_dict = {}
         singleton_local_shards = (metadata or {}).get('singleton_local_shards', False)
         for name, module in self._modules.items():
@@ -365,7 +372,11 @@ class MLP(MegatronModule):
                 for k, v in sub_sd.items():
                     if k in (f"{prefix}{name}.weight", f"{prefix}{name}.bias"):
                         sub_sd[k] = apply_swiglu_sharded_factory(
-                            v, sharded_offsets, singleton_local_shards
+                            v,
+                            sharded_offsets,
+                            singleton_local_shards,
+                            tp_group=self.tp_group,
+                            dp_group=metadata['dp_cp_group'],
                         )
             sharded_state_dict.update(sub_sd)
         return sharded_state_dict
@@ -404,7 +415,11 @@ class MLP(MegatronModule):
 
 # pylint: disable=missing-function-docstring
 def apply_swiglu_sharded_factory(
-    original_sh_ten, sharded_offsets, singleton_local_shards: bool = False
+    original_sh_ten,
+    sharded_offsets,
+    singleton_local_shards: bool = False,
+    tp_group: torch.distributed.ProcessGroup | None = None,
+    dp_group: torch.distributed.ProcessGroup | None = None,
 ):
     # We must split the tensor into 2 parts, each sharded separately.
     # This requires a ShardedTensorFactory which `chunk`s during saving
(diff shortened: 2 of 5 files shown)


megatron/core/transformer/moe/experts.py
```diff
@@ -848,7 +848,11 @@ def sharded_state_dict(
                     for k in (f'{name}.weight{i}', f'{name}.bias{i}'):
                         if k in sub_sd:
                             sub_sd[k] = apply_swiglu_sharded_factory(
-                                sub_sd[k], new_sharded_offsets, singleton_local_shards
+                                sub_sd[k],
+                                new_sharded_offsets,
+                                singleton_local_shards,
+                                tp_group=self.tp_group,
+                                dp_group=metadata['dp_cp_group'],
```
cspades, line 855: When using FSDP2, we may not be simply TP-sharding anymore (with EP offset), now that we have both TP and DP sharding we need the groups to figure out what goes where...

If FSDP2 isn't used, then neither the TP nor the DP group are needed. TP is inferred from simple 1-D sharding calculations.

megatron/core/distributed/torch_fully_sharded_data_parallel.py
```diff
@@ -96,6 +96,8 @@ def save_custom_attrs(module):
                     # micro-batch id, thus removing unnecessary memory stores
                     attrs['_fp8_attrs']['transpose_invalid'] = False
                     del attrs['_fp8_attrs']['transpose']
+                # Mark this parameter as an FSDP2 parameter.
+                attrs["is_torch_fsdp2_param"] = True
```
cspades, line 100: Too many FSDP2 checks dangerously just check for DTensor, when in the future DTensors may appear everywhere. Time to start housekeeping this logic.

megatron/core/transformer/mlp.py
```diff
@@ -462,10 +486,67 @@ def sh_ten_build_fn(
             ),
         ]
 
+    @torch.no_grad()
+    def dp_sh_ten_build_fn(
+        key: str, t: torch.Tensor, replica_id: ReplicaId, flattened_range: Optional[slice]
+    ):
```
cspades, line 492: Maps the global `[W;V]` to the TP W/V-swizzled TP-DP shards, e.g. DP4 and TP2 looks like:
```
W_tp0_dp0 W_tp0_dp1 V_tp0_dp2 V_tp0_dp3 W_tp1_dp0 W_tp1_dp1 V_tp1_dp2 V_tp1_dp3
```

yashaswikarnati (approved):

tests/unit_tests/distributed/test_torch_fully_sharded_parallel.py
```diff
@@ -106,3 +128,117 @@ def _is_fsdp_wrapped_module(instance):
...
+        global_offset=(6, 0),   # W/V TP-swizzled, DP-sharded!
+        axis_fragmentations=(8, 1),
+        replica_id=(0, 0, 0),
+        prepend_axis_num=0
+    )
+
+    This is a mapping from the checkpoint [W;V] rank offsets:
+
+        W_tp0_dp0 W_tp0_dp1 W_tp1_dp0 W_tp1_dp1 V_tp0_dp2 V_tp0_dp3 V_tp1_dp2 V_tp1_dp3
+        0         1         2         3         4         5         6         7
```
fanshiqing, line 209: > For each TP shard pair of W/V, FSDP2 then shards dim=0, so you end up with [ {W_tp0; V_tp0}_dp{1...DP}, {W_tp1; V_tp1}_dp{1...DP}, ... ]

TP should have higher priority than DP /FSDP, so here I think the sharding scheme should be:
```
W_tp0_dp0, W_tp1_dp0,  W_tp0_dp1, W_tp1_dp1, W_tp0_dp2, W_tp1_dp2, W_tp0_dp2, W_tp1_dp2,
```
Is that right?

megatron/core/transformer/mlp.py
fanshiqing: from this order why dp should have higher priority than tp?

megatron/core/transformer/mlp.py
```diff
@@ -404,7 +415,11 @@ def as_mlp_submodule(
 
 # pylint: disable=missing-function-docstring
 def apply_swiglu_sharded_factory(
-    original_sh_ten, sharded_offsets, singleton_local_shards: bool = False
+    original_sh_ten,
+    sharded_offsets,
+    singleton_local_shards: bool = False,
+    tp_group: torch.distributed.ProcessGroup | None = None,
+    dp_group: torch.distributed.ProcessGroup | None = None,
```
fanshiqing, line 422: TBH, Coupling FSDP deeply with generalized func like `apply_swiglu_sharded_factory` make it hard to understand and review.
(1) except from `apply_swiglu_sharded_factory`, if there are some other similar fix logics from FSDP in other functions? logically there should?
(2) can we figure our some more elegant / generalized fix without hacking into these shared functions? e.g., maintain there FSDP specific logic withing some FSDP utils files?

megatron/core/transformer/mlp.py
```diff
@@ -418,15 +433,24 @@ def apply_swiglu_sharded_factory(
...
+        is_dp_sharded = dp_size > 1
+    else:
+        is_dp_sharded = False
+
     @torch.no_grad()
     def sh_ten_build_fn(
         key: str, t: torch.Tensor, replica_id: ReplicaId, flattened_range: Optional[slice]
     ):
+        rank_offset = (
+            original_sh_ten.global_offset[swiglu_shard_axis + prepend_axis_num] // local_axis_size
```
fanshiqing, line 452: sorry but what's the necessity of moving `rank_offset` computation from L412 to here?

megatron/core/transformer/mlp.py
```diff
@@ -462,10 +486,67 @@ def sh_ten_build_fn(
...
+        # that DP sharding spans both W and V.
+        shards_per_half = half_axis_size // local_axis_size
+        assert dp_size == 2 * shards_per_half
+
+        # Compute if DP rank maps to W or V in [W; V].
+        swiglu_half_idx, half_dp_shard_idx = divmod(dp_rank, shards_per_half)
+        # If W, then 0. If V, then 1.
+        assert swiglu_half_idx in (0, 1)
+        # Map [ W; V ] to this rank's shard [ {W_tpx; V_tpx}_dpy ].
+        shard_rank_offset = (
```
fanshiqing, line 524: I suggest we should have some well-documentation will highlight diagrams for better explain why FDSP should require such a specific fix? There are footstone for FSDP's correctness and anyone who want to use/develop FSDP should understand what's happening under the hood.

`docs/user-guide/features/megatron_fsdp.md` could be one choice.

fanshiqing:
And I am confused if this is true bug then why previous related UTs just not catch it? 
Is it covered by the UTs for now?

megatron/core/transformer/mlp.py
cspades: This is a 1-D sharding formula that only works when this `ShardedTensor` is sharding on 1 dimension of devices, in which case you can derive the rank of this shard by looking at the global offset divided by the local shape.

So for example, if you are global offset 6 and your local size is 2, then obviously the (TP, but it could be _whatever_ parallelism that is actually sharding this) rank is 3.

Thus, this heuristic formula has no relevance beyond the 1-D sharding case. For 2-D sharding, we need to consider multiple rank calculations that cannot be derived from this formula.

Furthermore, leaving behind a higher-scoped formula that is not used for the (TP, DP) sharding case will confuse future developers wondering why this formula is relevant at all, so I decided to move it into the 1-D sharding builder function, where it naturally is relevant.

megatron/core/transformer/mlp.py
cspades: The reason it's coupled tightly is because this is used for checkpoint loading. Trying to re-shard the weights across DP after loading the checkpoint is not trivial or clean to do. (If you are curious, Megatron-FSDP does not save the W/V pair as a fused weight. It saves them separately, so we can very easily perform these transformations post-load, online, and in a separate part of the codebase. We cannot do it nicely for FSDP2 or ND as they both use `torch_dist`.)

This is why I created a new builder function `dp_sh_ten_build_fn` so it will not affect the non-FSDP case whatsoever.

1. Currently, I don't see any other FSDP-specific logic to fix. After swizzling this properly, we have activation and weight parity for an entire model.
2. Do you have any ideas for a more elegant fix that supports `torch_dist`?

megatron/core/transformer/mlp.py
cspades: Considering Torch FSDP2 is not well-tested (currently just 3 unit tests and maybe a few functional tests that don't really test anything) or documented, and possibly on a future deprecation path if Megatron-FSDP V2 becomes hot-swappable with FSDP2, I would prefer if we do the bare minimum maintenance of FSDP2 if possible.

I don't enjoy saying the above, but there are too many FSDP-like implementations and as a MCore engineer it will be impossible for me to rigorously maintain them all. 😩

tests/unit_tests/distributed/test_torch_fully_sharded_parallel.py
cspades: Also realize now that I probably should add the mapping to the result:
```
W_tp0_dp0 W_tp0_dp1 V_tp0_dp2 V_tp0_dp3 W_tp1_dp0 W_tp1_dp1 V_tp1_dp2 V_tp1_dp3
0         1         4         5         2         3         6         7
```

Your intuition is correct, but does not consider the W/V-swizzling. There are 3 steps going from `[W; V]` to `[ {W_tpx; V_tpx}_dpy ]` that mathematically are simplified into a single mapping in this PR:

- Shard W and V with TP.
  - This produces: `[ W_{tpx}; V_{tpx} ]`.
  - The W and V TP shards are still adjacent to each other.
- Swizzle W and V.
  - This changes the above into: `[ W_{tp0} V_{tp0} W_{tp1} V_{tp1} ... ]`
  - Each TP rank owns a TP-sharded pair of W and V for the fused SwiGLU FC1 weight.
  - Note that this is the whole purpose of the `apply_swiglu_sharded_factory` function.
- Shard the W/V-swizzled TP shards by DP. (For simplicity, let's use DP2 and TP2 so I don't have to write out too many terms. The numbering matters most here.)
  - This changes the above into: `[ W_{tp0}_{dp0} V_{tp0}_{dp1} W_{tp1}_{dp0} V_{tp1}_{dp1} ]`
  - DP shards the fused and swizzled W-V shard local to this TP rank. So unlike normal TP-DP sharding, the DP ranks do not span W or V individually, but span them together.
  - So indeed TP takes precedence over DP, which should always be the final sharding dimension, but because there is W/V swizzling in between them, DP ranks are completely partitioned between W and V. So the first half of the DP ranks will only shard W, and the second half of the DP ranks will only shard V. (We will never have `W_{tpx}_{dp1}` or `V_{tpx}_{dp0}`, for example.) This is also the job of this variable: `swiglu_half_idx`.

It's complex, but I've basically hidden it into its own section of code (`dp_sh_ten_build_fn`) that won't affect ND or GTP at all. Putting all this into the training loop or checkpointing code would be much worse to deal with in my opinion, and be far less clean than this builder function.

This unblocks some customers who really want to use FSDP2 with Megatron-Core: https://github.com/NVIDIA/Megatron-LM/issues/2769

This is what happens if you try to use the original factory on FSDP, loading an ND-parallel-trained checkpoint into FSDP2:

```
# ND Parallel
 [2025-12-28 00:20:24] iteration        1/     100 | consumed samples:         1024 | elapsed time per iteration (ms): 27003.9 | throughput per GPU (TFLOP/s/GPU): 452.7 | learning rate: 9.900000E-05 | global batch size:  1024 | lm loss: 2.404998E+00 | loss scale: 1.0 | grad norm: 3.048 | number of skipped iterations:   0 | number of nan iterations:   0 |Number of parameters in transformer block in billions:  6.98 
 [2025-12-28 00:20:49] iteration        2/     100 | consumed samples:         2048 | elapsed time per iteration (ms): 24621.6 | throughput per GPU (TFLOP/s/GPU): 496.6 | learning rate: 9.800000E-05 | global batch size:  1024 | lm loss: 4.466884E+00 | loss scale: 1.0 | grad norm: 44.051 | number of skipped iterations:   0 | number of nan iterations:   0 |
 
# Torch FSDP2
 [2025-12-28 00:23:52] iteration        1/     100 | consumed samples:         1024 | elapsed time per iteration (ms): 37835.9 | throughput per GPU (TFLOP/s/GPU): 323.1 | learning rate: 9.900000E-05 | global batch size:  1024 | lm loss: 1.386388E+01 | loss scale: 1.0 | grad norm: 346.039 | number of skipped iterations:   0 | number of nan iterations:   0 |
 [2025-12-28 00:24:26] iteration        2/     100 | consumed samples:         2048 | elapsed time per iteration (ms): 34441.6 | throughput per GPU (TFLOP/s/GPU): 355.0 | learning rate: 9.800000E-05 | global batch size:  1024 | lm loss: 3.159755E+01 | loss scale: 1.0 | grad norm: 544.112 | number of skipped iterations:   0 | number of nan iterations:   0 |
```
I've reproduced it with the script above, and bitwise parity checked the weights and debugged the checkpoint loading code to detect this. I believe allowing our users to take ND-parallel checkpoints and use them in other parallelisms as well is in the best interests of making Megatron flexible and Megatron models portable.

megatron/core/transformer/mlp.py
cspades: Another option would be to create a new factory builder like `apply_swiglu_fsdp_sharded_factory`, but it would require adding `FSDP2 + Interleaved SwiGLU FC1`-specific logic wherever this function is used, when a separate builder function DRY's them all here. It makes it easier for people to reuse the same logic (I wouldn't want someone else to rewrite this without my supervision...) while being pretty readable since it's just a `builder_fn` swap.

cspades:
> And I am confused if this is true bug then why previous related UTs just not catch it? Is it covered by the UTs for now?

@fanshiqing Mentioned this in DM's and before, but Torch FSDP2 has barely any unit tests, the unit test in this PR is the first of this kind, unfortunately. I'd prefer to keep the maintenance light until we have a single source-of-truth implementation of FSDP and our customers are migrated.

fanshiqing (approved):

ericharper (approved):

deepakn94 (approved):

PR #2, opened 2019-04-24
rename confusing arg

yaroslavvb:
WORLD_SIZE usually refers to total number of gpus while --nproc_per_node should be set to number of gpus per node

PR #5715, opened 2026-07-08
IMA fix by making the copy of book keeping buffer to GPU blocking

shanmugamr1992:
1. **The original issue.** Each inference step, Megatron copies a pinned CPU
   bookkeeping buffer (`_cpu_bookkeeping_buf`) to the GPU with an **async**
   (`non_blocking=True`) copy, then immediately reuses/overwrites that same CPU
   buffer on the next step. The host doesn't wait for the copy to finish, so it
   clobbers bytes that the in-flight GPU copy is still reading. The GPU ends up
   with corrupted token/KV-block indices and dereferences out-of-bounds memory
   → `illegal memory access`. CUDA-graph warmup runs steps back-to-back with
   almost no host work in between, so the race fires every time there.
   
   Fix: 
  to make the copy blocking would make sure this issue doesn't happen
  

Window | blocking (fix #1) | shadow (fix #2) | Δ shadow vs blocking
-- | -- | -- | --
all 6 steps | 115.1 | 114.7 | +0.3% slower
steady (steps 2–6) | 117.9 | 114.2 | +3.2% slower
steady (steps 3–6) | 121.4 | 119.8 | +1.3% slower
No measurable throughput benefit from the shadow buffer.The two are within run-to-run noise, and the small deltas actually favor which shadow physically can't lose to on real overlap, so it's just noise. 

timer aggregates ~370 decode iterations/seq; the per-iteration copy is a few µs and any blocking host-stall (~10–30 µs) is a fraction of a percent of the ~32 ms/iter forward pass — diluted below noise.

With larger models this is even less a concern,


kvareddy:
@tdene can you signoff of these changes?

tdene (approved):

kvareddy:
@lmcafee-nvidia can you please take a look at these changes?

santhnm2 (approved):
Can you update the PR title and description?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -2457,7 +2457,17 @@ def transfer_bookkeeping_to_gpu(self) -> None:
         # Copying the whole (max_tokens + max_requests)-sized buffer including
         # unused slots is cheap (~71 KB total, ~3-5 us on PCIe Gen4) and saves
         # 8 redundant launch overheads vs. the prior per-field copies.
-        self.gpu_view._buf.copy_(self._cpu_bookkeeping_buf, non_blocking=True)
+        # This copy MUST be blocking. `_cpu_bookkeeping_buf` is a pinned host
```
lmcafee-nvidia, line 2514: the method docstring above still says that all copies are done with `non_blocking=True`. the comment on this line is good, but let's update the method docstring to mention briefly why this copy is done with `False`.

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia, line 1: can you post your timing results that you shared on slack to the PR description, for reference? thanks

shanmugamr1992: done

cuichenx (approved):

lmcafee-nvidia (approved):

shanmugamr1992:
/ok to test [75dd58f](https://github.com/NVIDIA/Megatron-LM/pull/5715/commits/75dd58f5c9ec6134d7a21545f4cc8990d3762d42)

shanmugamr1992:
/ok to test [8cee542](https://github.com/NVIDIA/Megatron-LM/pull/5715/commits/8cee54292517cc633509c95418386b02244e35bd)

PR #5719, opened 2026-07-09
Overlap FSDP communication with compute

### LINKED ISSUE #5647: MFSDP v2 overlap, prefetching, and double-buffering schedule support
MFSDP v2 should support the scheduling mechanisms needed for communication/computation overlap, prefetching, and double buffering.

Design doc: http://nv/mfsdp-schedule-design

This issue tracks the schedule-level work across:
- all-gather / compute overlap
- reduce-scatter / compute overlap
- forward and backward prefetch behavior
- double-buffered staging and allocator behavior

In-progress PRs:
- #5416
- #5513

wujingyue:
## Summary

This implements the explicit prefetching approach used in MFSDP v1 for MFSDP v2 (`mfsdp_v2`). The current implementation assumes that static module orders match the execution order closely enough for one-step prefetch.

Each FsdpModule's pre-forward hook prefetches the next FsdpModule's parameter all-gather from `forward_order`. During backward, pre-hooks follow a separately computed `backward_order` instead of assuming backward is strictly the reverse of `forward_order`, which matters for nested module trees.

An example trace from the added unit test:

image

### Forward prefetch

Each FsdpModule tracks materialization with `_unshard_event`. `_unshard_parameter_groups()` is idempotent: if `_unshard_event` is already set, the FsdpModule was already unsharded or prefetched; otherwise it enqueues the all-gather on the shared all-gather stream and records `_unshard_event` after materialization.

`pre_forward` waits on `_unshard_event` instead of waiting on the whole all-gather stream, so compute depends only on the current FsdpModule's materialization and not on later release work queued on that stream. It then issues the next FsdpModule's all-gather from `forward_order`. `_reshard_parameter_groups()` queues release on the same all-gather stream where unsharded storage was allocated, then clears `_unshard_event` so the next user enqueues a fresh all-gather.

### Backward overlap

The shared FSDP context maintains separate static `forward_order` and `backward_order` sequences. Each backward pre-hook:

1. calls the same idempotent unshard path and waits only for that FsdpModule's `_unshard_event`;
2. prefetches the next FsdpModule needed by backward on the all-gather stream;
3. allows that all-gather to overlap the current FsdpModule's gradient computation.

Gradient reduction is split into allocation, packing, and communication stages. Once a module's gradients are packed, its reduce-scatter is launched immediately on a separate reduce-scatter stream, allowing it to overlap subsequent backward GEMMs and all-gathers. Existing symmetric-memory scaling, mixed gradient dtypes, first-backward assignment, and microbatch accumulation semantics are preserved.

An autograd-end callback registered by the root makes the caller's current stream wait for the reduce-scatter stream before `backward()` returns. This cannot rely on the root module's own post-backward hook: that hook tracks gradients for parameters owned by the root and may run before descendants finish, or may not run at all when the root owns no trainable parameters.

## Verification

- `test_fully_shard.py` passes with two CUDA ranks after the `_unshard_event` change: **14 passed per rank**.
- The nested-order unit test verifies the static forward and backward orders for a nested module tree, including the backward DFS traversal.
- The profiler test covers one forward and backward pass. For four child FsdpModules, it requires all **6 expected all-gather/GEMM overlaps** and all **3 expected reduce-scatter/GEMM overlaps**, rather than accepting any single overlap.
- The profiler test also verifies that all-gathers use one stream, reduce-scatters use another, and both streams are distinct from compute.
- A local Nsight Systems capture confirms backward all-gathers and reduce-scatters overlap GEMM execution on separate streams.
- Diff and byte-compilation checks pass for the latest edits.

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
@@ -58,6 +58,10 @@ class FsdpContext:
             self.allgather_stream = torch.cuda.Stream()
             self.reduce_scatter_stream = torch.cuda.Stream()
 
+    def current_stream(self) -> torch.cuda.Stream:
+        """Current stream on this context's device."""
+        return torch.cuda.current_stream(self.allgather_stream.device)
+
     def register_post_backward_final_callback(self) -> None:
         """Register this root context's final callback for the current backward.
 
@@ -68,8 +72,7 @@ class FsdpContext:
         """
 
         def post_backward_final_callback() -> None:
-            current_stream = torch.cuda.current_stream(self.reduce_scatter_stream.device)
-            current_stream.wait_stream(self.reduce_scatter_stream)
+            self.current_stream().wait_stream(self.reduce_scatter_stream)
 
         torch.autograd.Variable._execution_engine.queue_callback(post_backward_final_callback)
 
@@ -226,7 +229,7 @@ class FsdpModule:
         self._ready_grad_parameters.clear()
         context = self.context
         allgather_stream = context.allgather_stream
-        current_stream = torch.cuda.current_stream(allgather_stream.device)
+        current_stream = context.current_stream()
 
         if self.is_root():
             allgather_stream.wait_stream(current_stream)
@@ -252,8 +255,7 @@ class FsdpModule:
         if self._unshard_event is not None:
             return
 
-        context = self.context
-        allgather_stream = context.allgather_stream
+        allgather_stream = self.context.allgather_stream
         with torch.cuda.stream(allgather_stream):
             for group in self._parameter_groups:
                 if sync_model_weight:
@@ -278,7 +280,7 @@ class FsdpModule:
             group.reshard_parameters()
 
         allgather_stream = self.context.allgather_stream
-        allgather_stream.wait_stream(torch.cuda.current_stream(allgather_stream.device))
+        allgather_stream.wait_stream(self.context.current_stream())
         # Release on the all-gather stream where unsharded storage was allocated,
         # so no record_stream() call is required for the storage.
         with torch.cuda.stream(allgather_stream):
@@ -290,8 +292,7 @@ class FsdpModule:
         """Prepare full parameters and prefetch the next FsdpModule in backward order."""
         torch.cuda.nvtx.range_push(self._nvtx_label("backward"))
         context = self.context
-        allgather_stream = context.allgather_stream
-        current_stream = torch.cuda.current_stream(allgather_stream.device)
+        current_stream = context.current_stream()
         if self.is_root():
             context.register_post_backward_final_callback()
             # Fork the reduce-scatter stream from the current stream once, at the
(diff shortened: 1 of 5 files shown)


megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -27,22 +27,40 @@
...
+            device: Device on which this context schedules communication.
             root_module: Outermost module that owns this context.
         """
         self.root_module = root_module
         self.is_last_microbatch = True
+        self.forward_order = []
+        with torch.cuda.device(device):
+            self.allgather_stream = torch.cuda.Stream()
+
+    def next_of(self, module: "FsdpModule") -> "FsdpModule | None":
```
wujingyue: Consider a linked hashmap.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -114,7 +136,9 @@ def _lazy_init_context(self) -> None:
         if self._context is not None:
             return
 
-        context = FsdpContext(root_module=self)
+        context = FsdpContext(device=self._parameter_groups[0].main_weight.device, root_module=self)
+        # named_modules() yields units in registration order, which is the static
+        # forward execution order used to prefetch the next unit's all-gather.
```
wujingyue: This is a fragile assumption. Fine for the prototype.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -201,64 +220,121 @@ def grad_hook(_parameter: nn.Parameter) -> None:
...
+            next_module._issue_unshard(sync_model_weight=True)
 
+    def _issue_unshard(self, *, sync_model_weight: bool) -> None:
+        """Issue this unit's all-gather on the comm stream (no compute-side wait)."""
+        context = self.context
+        allgather_stream = context.allgather_stream
         with torch.cuda.stream(allgather_stream):
             for group in self._parameter_groups:
                 if sync_model_weight:
-                    # TODO: After NVIDIA/Megatron-LM#5411 lands, move this sync to the
```
wujingyue, line 229: This comment should be kept. Probably due to a bad merge.

shjwudp (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
wujingyue: Use a sliding window instead of relying on the each-module-appears-once assumption

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -201,64 +220,121 @@ def grad_hook(_parameter: nn.Parameter) -> None:
...
+        """Free this unit's unsharded storage on the comm stream once its own
+        compute (``F_i``/``B_i``) has consumed it.
+
+        With forward prefetch driving overlap, storage no longer needs to be
+        delay-queued to expose a later unit's all-gather; each unit's buffer can
+        be released right after its compute. The release still runs on the
+        all-gather stream (where the buffer was allocated), gated on a consumer
+        event recorded on the compute stream so the free never races the kernels
+        that read the buffer.
+        """
```
shjwudp, line 278: Should we collect the states of unsharded models and release module parameter from all models except the current and next ones?

wujingyue: When would that be useful? Some FsdpModule's post-forward or post-backward failed to be triggered? That sounds like a bug that we shouldn't hide.

jiemingz (approved):

Phlip79 (approved):

PR #5735, opened 2026-07-09
NCCL EP zero copy

YangFei1990:

# What does this PR do?
Adding support for zero copy feature for NCCL EP, to skip D2D copies from user buffer into NCCL's staging buffer. This PR works together with the TE PR https://github.com/NVIDIA/TransformerEngine/pull/3187

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


YangFei1990:
/ok to test [5e50386](https://github.com/NVIDIA/Megatron-LM/pull/5735/commits/5e503866e03178c91d24d6163414cd4d6fbbbb43)

megatron/core/pipeline_parallel/utils.py
```diff
@@ -228,7 +235,13 @@ def _forward(self, *inputs):
...
                 if input is not None:
                     input.record_stream(self.stream)
-                    input.untyped_storage().resize_(0)
+                    # Skip symmetric-memory (zero-copy EP) buffers
+                    if not (
+                        self.ncclep_zero_copy
+                        and is_symm_backed is not None
+                        and is_symm_backed(input)
+                    ):
+                        input.untyped_storage().resize_(0)
```
Autumn1998, line 276: Could you please explain why symm doesn’t need to manually free the storage here? It seems like there would be extra memory usage.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1404,6 +1406,32 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
...
+    def forward(ctx, recv_tokens):  # type: ignore[override]
+        return recv_tokens
+
+    @staticmethod
+    def backward(ctx, grad):  # type: ignore[override]
+        buf = _NCCLEPManager._zc_bwd_token_buf
+        assert buf is not None, "zero-copy staging buffer not allocated before dispatch-backward"
+        assert buf.shape == grad.shape, (
+            f"dispatch-bwd grad {tuple(grad.shape)} != staging buffer {tuple(buf.shape)}"
+        )
```
Autumn1998: Is there an extra copy happening here? 🤔

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1573,10 +1649,20 @@ def dispatch(
         #   tokens_per_expert: [num_local_experts]
         #   dispatched_probs: [recv_capacity_per_rank]
         recv_tokens, tokens_per_expert, dispatched_probs = nccl_ep_dispatch(
-            self._buffer, hidden_states, topk_idx, topk_weights
+            self._buffer,
+            hidden_states,
+            topk_idx,
+            topk_weights,
+            recv_tokens=_NCCLEPManager._zc_fwd_token_buf if self._zc_quant else None,
```
Autumn1998, line 1684: I noticed that the EP buffer itself already uses the symm pool in TE. However, it seems that for BF16 it uses the symm pool from TE, while for FP8 it doesn’t. Why are these two treated differently? If we always use the symm memory pool, what impact would that have?

megatron/core/pipeline_parallel/utils.py
YangFei1990: Yes it is extra memory usage, but is not "doesn't need", but "can't". Basically symm tensor can not be resized

megatron/core/transformer/moe/token_dispatcher.py
YangFei1990: Yes but unfortunately this is something can not be avoided. This autograd function is only invoked when 1F1B overlap is enabled. This is because the [input](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/pipeline_parallel/utils.py#L210) of mlp is a leaf tensor, in backward, Pytorch will trigger the `AccumulateGrad` which will do a [clone](https://github.com/pytorch/pytorch/blob/main/torch/csrc/autograd/functions/accumulate_grad.h#L225), this clone not only a copy, but also will cast the symm tensor back to normal pytorch tensor.
So walkaround here is to pass normal torch tensor to groupgemm, and copy it back to symm here.

megatron/core/transformer/moe/token_dispatcher.py
YangFei1990: The main reason is symm pool today does not support cuda graph. So we pre-allocate persistent symm buffers. The only different treatment is the dispatch recv buffer, the pattern is:
- FP8: dispatch recv buffer can be reused -> mcore allocate persistent buffer -> feed into TE dispatch/combine/groupgemm
- bf16: dispatch recv buffer can not be reused (because it needs to be saved as activation for bwd) -> TE allocate the buffer on the fly use symm pool -> use Pytorch to manage the lifecycle of the tensor

In the future when the symm pool is ready for cuda graph, we can move all to symm pool

YangFei1990:
/ok to test [e64e344](https://github.com/NVIDIA/Megatron-LM/pull/5735/commits/e64e344ae0321410c1a5b3eecbe12577e61c515e)

YangFei1990:
/ok to test [892a612](https://github.com/NVIDIA/Megatron-LM/pull/5735/commits/892a612c5b0517867bdadf8f35cfe240500c8b9c)

Autumn1998 (approved):

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1404,6 +1406,34 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
...
+
+    @staticmethod
+    def backward(ctx, grad):  # type: ignore[override]
+        """Stage the incoming gradient into the symm dispatch-backward buffer."""
+        buf = _NCCLEPManager._zc_bwd_token_buf
+        assert buf is not None, "zero-copy staging buffer not allocated before dispatch-backward"
+        assert (
+            buf.shape == grad.shape
+        ), f"dispatch-bwd grad {tuple(grad.shape)} != staging buffer {tuple(buf.shape)}"
+        buf.copy_(grad)
```
phu0ngng: Do we need this copy here? 
Can't we make FFN2 BWD DGRAD GroupedGEMM write to the buf directly?

megatron/core/transformer/moe/token_dispatcher.py
phu0ngng: I see, and we can't simply do a non-zero-copy for this Dispatch bwd here since we bootstrap with zero-copy; thus, there is no internal staging buffer allocated.

phu0ngng (approved):

YangFei1990:
/ok to test [b9470ed](https://github.com/NVIDIA/Megatron-LM/pull/5735/commits/b9470ed0ad472a2a539226c283d388cf0a9a0024)

guihong-nv (approved):

fanshiqing (approved):

megatron/core/transformer/moe/experts.py
```diff
@@ -61,9 +61,18 @@
     import transformer_engine as te
 
     from megatron.core.extensions.transformer_engine import Fp8Padding, Fp8Unpadding
+
+    try:
+        from transformer_engine.pytorch.ops.basic.grouped_linear import (
+            GRAD_INPUT_BUFFER_KEY,
+            OUTPUT_BUFFER_KEY,
+        )
```
Phlip79, line 69: Why is this in a `try` if there's already a check for `HAVE_TE`?

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -698,6 +707,9 @@ def nccl_ep_dispatch(buffer, tokens, topk_idx, topk_weights):
             topk_idx (torch.Tensor): ``int64`` ``[num_local_tokens, top_k]`` global expert
                 ids per token.
             topk_weights (torch.Tensor): ``float32`` ``[num_local_tokens, top_k]`` weights.
+            recv_tokens, recv_topk_weights (torch.Tensor, optional): caller-owned symm dispatch
```
Phlip79, line 714: These are two different arguments

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -546,8 +546,20 @@ def routed_experts_compute(self, hidden_states: torch.Tensor, probs: torch.Tenso
...
+                get_zc_buffers() if get_zc_buffers is not None else (None, None)
+            )
+            expert_kwargs = {}
+            if output_buffer is not None:
+                expert_kwargs["output_buffer"] = output_buffer
+            if grad_input_buffer is not None:
+                expert_kwargs["grad_input_buffer"] = grad_input_buffer
             expert_output, mlp_bias = apply_module(self.experts)(
-                dispatched_input, tokens_per_expert, permuted_probs
+                dispatched_input, tokens_per_expert, permuted_probs, **expert_kwargs
```
Phlip79, line 559: Why not just pass `output_buffer` and `grad_input_buffer` directly?

megatron/core/transformer/moe/experts.py
YangFei1990: These are newer version TE APIs just added in https://github.com/NVIDIA/TransformerEngine/pull/3161

Phlip79 (approved):
Please resolve my comments before merge

megatron/core/transformer/moe/moe_layer.py
YangFei1990: This is because not all group gemm backends accept these 2 kwargs. Dispatcher side has gate to make sure the right group gemm backend is used when these buffers are not None.

megatron/core/transformer/moe/fused_a2a.py
YangFei1990: The same description applies to both kwargs.

megatron/core/transformer/moe/moe_layer.py
Phlip79: Can you add that as a comment? Otherwise it's confusing.

YangFei1990: The above comment 549-551 has discussed about this, do you think if they are clear?

Phlip79: I missed that. It's probably fine.

YangFei1990:
/ok to test [21e747f](https://github.com/NVIDIA/Megatron-LM/pull/5735/commits/21e747f17bde0c93e88f10048bafb51bf9b72d03)

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -546,8 +546,20 @@ def routed_experts_compute(self, hidden_states: torch.Tensor, probs: torch.Tenso
                 dispatched_input, tokens_per_expert, permuted_probs, routing_map=routing_map
             )
         else:
+            # NCCL-EP zero-copy: experts write fc2 output and fc1 dgrad straight into the combine /
+            # dispatch symm buffers. Passed only when set (non-TEGroupedMLP experts don't accept
+            # these kwargs).
+            get_zc_buffers = getattr(self.token_dispatcher, "get_expert_zero_copy_buffers", None)
```
jiemingz: rather than an attribute check can 'get_expert_zero_copy_buffers' be added to the token dispatcher base class?

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1447,6 +1449,34 @@ def get_restored_hidden_states_by_experts(self, hidden_states: torch.Tensor) ->
         return hidden_states
 
 
+class _StageDispatchBwdGrad(torch.autograd.Function):
```
jiemingz: would somewhere in `core/pipeline_parallel/` be a better location? I think isolating 1f1b is more important here.

megatron/core/transformer/moe/experts.py
```diff
@@ -665,10 +676,22 @@ def _fused_forward(
             fine_grained_activation_offloading, permuted_local_hidden_states, offload_name
         )
         with fused_group_mlp_manager as permuted_local_hidden_states:
+            # NCCL-EP zero-copy: the fused-MLP input aliases the persistent symm buffer (also the
+            # fc2 output combine reads), whose storage is non-resizable — don't force-release it.
```
jiemingz: can you make this comment clearer that that specifically refers to the `output_buffer` being not None?

megatron/core/transformer/moe/token_dispatcher.py
YangFei1990: This is only relevant to NCCL EP zero-cp and only used in this `token_dispatcher.py`, I wonder if we should put in a different location, might create extra hops when agent tracing the code.

jiemingz: Its functionality is specific to both 1f1b and zero copy, and the precedent has been to keep 1f1b more encapsulated.

this just needs to wrap the dispatch output in order to insert an autograd node after the dispatch in the autograd. could we just apply it at the 1f1b node itself?
https://github.com/NVIDIA/Megatron-LM/blob/076f61fc7be6dd0d8a7a95f99cc66dd775ac5c42/megatron/core/models/gpt/fine_grained_callables.py#L156

YangFei1990: Moved the autograd function into the pipeline/utils and called it in the above method

jiemingz (approved):

YangFei1990:
/ok to test [80f8b55](https://github.com/NVIDIA/Megatron-LM/pull/5735/commits/80f8b55594df88312401aa5a8dea8ac085e45659)

PR #4686, opened 2026-05-07
feat(gpt): add output postprocess hook

Glitchfix:
# What does this PR do ?

Adds an objective-neutral GPT output postprocess hook so downstream RL callers can consume decoder hidden states and output-layer helpers without monkey-patching `GPTModel.forward` or the schedule-plan postprocess path.

## Issue tracking

Linked issue: Related to #4590

## Summary

- Adds keyword-only `output_processor` and `output_processor_context` arguments to `GPTModel.forward`.
- Runs the custom processor after decoder/MTP hidden-state preparation and before the default logits/loss path.
- Threads the same processor/context through `build_schedule_plan` and the 1F1B `PostProcessNode`.
- Adds unit coverage for a selected-token logprob processor that uses the output layer and matches the default logits path, plus schedule-plan postprocess threading.

This is intentionally a narrow first slice for #4590. It does not add RL-specific arguments to `GPTModel.forward`, and it leaves MTP-specific custom postprocess behavior as a follow-up unless maintainers prefer a broader API.

### Pre-checks

- [x] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [x] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [x] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

## Testing

- `pre-commit run --files megatron/core/models/common/model_chunk_schedule_plan.py megatron/core/models/gpt/fine_grained_callables.py megatron/core/models/gpt/gpt_model.py tests/unit_tests/models/test_gpt_model.py`
  - Black, pylint, and isort passed.
- `python -m py_compile megatron/core/models/gpt/gpt_model.py megatron/core/models/common/model_chunk_schedule_plan.py megatron/core/models/gpt/fine_grained_callables.py tests/unit_tests/models/test_gpt_model.py`
- `PATH=/tmp/megatron-autoformat-stable/bin:$PATH CHECK_ONLY=true SKIP_DOCS=true ./tools/autoformat.sh`
  - Black, isort, pylint, and ruff passed.
  - Mypy is advisory in `tools/autoformat.sh` (`|| true`); running it in this local environment reports existing project typing issues in the touched files' surrounding code.
- Standalone CUDA smoke test using the local GPT layer spec:
  - direct `GPTModel.forward(..., output_processor=...)` selected-token logprobs match logprobs computed from the default logits path;
[...]

Issue #4590, opened 2026-05-02
Reduce downstream Megatron patching for RL use cases


sbhavani:
Some Megatron Core features are difficult to use from external RL training loops without copying or monkey-patching GPTModel.forward, GPT postprocess, MTP postprocess, or 1F1B schedule plan.

This usually happens when the training loop owns data or semantics that Megatron Core should not model directly: selected-token labels, loss masks, packed sequence metadata, old/reference logprobs, KL/entropy terms, or custom fused logprob/loss computation.

**Current downstream symptoms**

- veRL patches/copies GPT/MTP postprocess logic:
    - verl/models/mcore/mtp_patch.py
    - verl/models/mcore/model_forward_fused.py
    - verl/models/mcore/model_forward_1f1b_overlap.py

This indicates that external training loops need a stable, objective-neutral extension point at the GPT postprocess boundary.

**Proposed direction**

Add a small optional GPT output/postprocess hook, keyword-only. The hook should run after decoder hidden states are available and before the default output-layer logits/loss path. This should avoid adding PPO/GRPO/RL-specific arguments to GPTModel.forward.

Schedule-plan support

Thread the same optional processor/context through build_schedule_plan and the 1f1b schedule-plan PostProcessNode. 

MTP follow-up

Handle MTP separately if needed. First investigate whether MTP can expose a narrow callable for custom loss/logprob computation while Megatron Core continues to own MTP shifting, packed-sequence handling, scaling, and logging behavior.

sbhavani:
Adding another concrete downstream patching example: R3 / MoE router replay.

R3 is supported in Megatron Core, but external RL frameworks still need to patch Megatron internals to use it e2e. In veRL this currently requires:

Downstream frameworks currently have to monkey-patch Megatron router/config/dispatcher internals and reimplement replay metadata layout/state handling across PP, VPP, SP, packed/THD, forward, recompute, and backward paths.

It would help downstream RL frameworks if Megatron Core exposed a stable public replay API that accepts/returns canonical router replay metadata and owns the PP/VPP/SP/packed/THD layout transforms internally, so frameworks can pass rollout-recorded top-k indices without monkey-patching router or dispatcher internals.

veRL references:
- https://github.com/verl-project/verl/blob/main/verl/utils/megatron/router_replay_patch.py
- https://github.com/verl-project/verl/blob/main/verl/workers/engine/megatron/transformer_impl.py

chengcuiping:
[...]

PR #5739, opened 2026-07-10
[core] Clarify GPT output_processor callback contract

chengcuiping:
## What does this PR do?

This PR clarifies the return contract of `GPTModel`'s optional `output_processor` callback.

Specifically, it:

- generalizes the callback return annotation from `Tensor` to `Any`;
- documents that, for direct `GPTModel.forward()` calls, the processor result becomes the forward result;
- adds direct-forward unit coverage for a structured, non-Tensor callback result.

## Motivation

The runtime hook already returns the processor result directly, but its existing `Callable[..., Tensor]` annotation is narrower than the actual behavior.

Downstream validation exposed this mismatch using a structured model output. This PR keeps the contract objective-neutral and does not introduce framework- or RL-specific output types.

## Scope

This is a typing, documentation, and unit-test change. It does not change:

- `_postprocess()` runtime behavior;
- the default logits or loss path;
- MTP behavior;
- 1F1B scheduling;
- schedule-plan output conversion;
- `float16_to_fp32()` behavior.

## Testing

Focused callback and schedule-plan tests:

```text
CUDA_VISIBLE_DEVICES=0 pytest -q tests/unit_tests/models/test_gpt_model.py \
  -k "output_processor or schedule_plan"

2 passed, 14 deselected
```

Full test file:

```text
CUDA_VISIBLE_DEVICES=0 pytest -q tests/unit_tests/models/test_gpt_model.py

13 passed, 1 skipped, 2 failed
```

This matches the clean `main` baseline exactly. The same two custom-process-group tests fail on `main`: both construct an 8-rank `HyperCommGrid` under a local `world_size=1` run, so no PR-introduced regression was found.

Additional checks completed successfully:

- `python3 -m py_compile`
- `git diff --check`
- `isort --check-only`
- `ruff check --no-fix`

The repository-pinned pre-commit and Black toolchain was not available locally, so the official project CI remains required for those checks.

Related to #4590.

Follow-up to #4686.

### DIFF
--- a/megatron/core/models/gpt/gpt_model.py
+++ b/megatron/core/models/gpt/gpt_model.py
@@ -531,9 +531,9 @@ class GPTModel(LanguageModule):
         inference_params: Optional[BaseInferenceContext] = None,
         loss_mask: Optional[Tensor] = None,
         padding_mask: Optional[Tensor] = None,
-        output_processor: Optional[Callable[..., Tensor]] = None,
+        output_processor: Optional[Callable[..., Any]] = None,
         output_processor_context: Optional[Any] = None,
-    ) -> Tensor:
+    ) -> Any:
         """Forward function of the GPT Model This function passes the input tensors
         through the embedding layer, and then the decoder and finally into the post
         processing layer (optional).
@@ -810,7 +810,7 @@ class GPTModel(LanguageModule):
         loss_mask: Optional[Tensor] = None,
         padding_mask: Optional[Tensor] = None,
         *,
-        output_processor: Optional[Callable[..., Tensor]] = None,
+        output_processor: Optional[Callable[..., Any]] = None,
         output_processor_context: Optional[Any] = None,
     ):
         """Builds a computation schedule plan for the model.
--- a/megatron/core/models/common/model_chunk_schedule_plan.py
+++ b/megatron/core/models/common/model_chunk_schedule_plan.py
@@ -351,7 +351,7 @@ class TransformerModelChunkSchedulePlan(AbstractSchedulePlan):
         loss_mask: Optional[Tensor] = None,
         padding_mask=None,
         *,
-        output_processor: Optional[Callable[..., Tensor]] = None,
+        output_processor: Optional[Callable[..., Any]] = None,
         output_processor_context: Optional[Any] = None,
     ):
         """Initialize the schedule plan of all Transformer layers' sub-modules.
--- a/tests/unit_tests/models/test_gpt_model.py
+++ b/tests/unit_tests/models/test_gpt_model.py
@@ -144,6 +144,7 @@ class TestGPTModel:
         ).cuda()
 
         context = {"selected_token_positions": torch.tensor([0, 2], device="cuda")}
+        created = {}
         seen = {}
 
         def output_processor(**kwargs):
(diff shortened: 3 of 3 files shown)


cspades:
/ok to test [a42ee1e](https://github.com/NVIDIA/Megatron-LM/pull/5739/commits/a42ee1efe58385775e594ed87386bbc120a5524d)

chengcuiping:
Thanks @cspades for approving the previous `a42ee1e` HEAD for testing.

I rebased this PR onto current `main` at `a54bd8cc`. The new HEAD is `e9da270`. I also consolidated the new contract coverage into the existing `output_processor` test, reducing the patch while retaining coverage for structured callback results, identity-preserving forwarding, keyword-only `context=`, and tied/untied output-weight semantics.

Local validation on one GPU completed with the focused tests passing (`3 passed`). The full test file reported `14 passed, 1 skipped, 2 failed`; a clean `main` run produced the same two custom-process-group failures because they construct an 8-rank grid under `world_size=1`, so no PR regression was found. Compile, diff, isort, and Ruff checks also passed.

A concrete downstream use of the structured callback result is [[verl-project/verl#6933](https://github.com/verl-project/verl/pull/6933)](https://github.com/verl-project/verl/pull/6933).

Could you please run `/ok to test e9da270` for the new HEAD and review when convenient? Thank you.

megatron/core/models/gpt/gpt_model.py
```diff
@@ -546,10 +546,16 @@ def forward(
...
-            output_processor_context (Any, optional): User-defined context object forwarded to
-                `output_processor`.
+            output_processor (Callable, optional): Custom postprocess callback invoked after
+                decoder hidden states are available and before the default logits/loss path.
+                When provided, the callback result becomes the direct `GPTModel.forward` result.
+                The callback receives the context object using the `context` keyword.
+                `output_weight` is populated with the shared embedding/output weight when
+                embeddings and output weights are tied. For untied embeddings, `output_weight`
+                is None and the output projection weight remains available as
+                `output_layer.weight`.
```
cspades: Doesn't this over-specify what this abstract output process does? Why is this information relevant again?

cspades:
/ok to test [e9da270](https://github.com/NVIDIA/Megatron-LM/pull/5739/commits/e9da270f93f3167e2cab12b584ba9ca316fe1770)

cspades:
Please ping @NVIDIA/mcore-oncall for all future assistance, we rotate.

chengcuiping:
Thanks for approving the new HEAD for testing and for the guidance. I’ll use the rotating MCore on-call team for future CI and review follow-ups.

megatron/core/models/gpt/gpt_model.py
chengcuiping: You're right. Those details describe the current callback invocation mechanics and are useful to downstream processors that reuse the output projection, but they over-specify the abstract `output_processor` contract.

I removed the context keyword and tied/untied `output_weight` details from the public docstring and narrowed the test change accordingly. The PR is now focused on the actual contract mismatch: an output processor may return an arbitrary result, and `GPTModel.forward()` propagates that result directly.

chengcuiping:
@NVIDIA/mcore-oncall, the review feedback has been addressed and the branch was amended to new HEAD `701bab31b`. Could you please run `/ok to test 701bab31b` for the new HEAD and review the PR when convenient? Thank you.

janEbert:
Just to note, oncall is not a required review. @NVIDIA/gpt, @NVIDIA/core-adlr, and @NVIDIA/core-nemo are the required review groups.

janEbert:
/claude fix Also update the type from `Tensor` to `Any` here: https://github.com/chengcuiping/Megatron-LM/blob/701bab31b9631866166c9592c65f2180d068e3ee/megatron/core/models/common/model_chunk_schedule_plan.py#L354

chengcuiping:
Thanks @janEbert for the clarification, and for applying the schedule-plan type fix and re-triggering CI on `2015cfc`. I appreciate the help.

I’ll wait for the checks on the updated HEAD and then follow up with the required review groups.

jaredcasper (approved):

ericharper (approved):

cuichenx (approved):

PR #5652, opened 2026-07-04
Add microbatch context helper

wujingyue:
## Summary
- Add `FsdpContext` with `is_last_microbatch` state for experimental FSDP.
- Add `microbatch(module, is_last)` to scope microbatch state across FSDP contexts.
- Add `FsdpModule.context` and root ownership tracking for lazily initialized context state.
- Expose only `fully_shard` and `microbatch` from the experimental package root.
- Add coverage for shared child/root contexts and unwrapped-parent child contexts.

## Testing
- `PATH=".venv/bin:$PATH" BASE_REF=main CHECK_ONLY=true SKIP_DOCS=false bash tools/autoformat.sh`
- `git diff --check`
- `rg -n "is_first_microbatch|is_first\\b" megatron/core/distributed/fsdp/src/megatron_fsdp/experimental tests/unit_tests/distributed/megatron_fsdp/test_context.py tests/unit_tests/distributed/megatron_fsdp/test_experimental_fully_shard.py` (no matches)
- `/opt/venv/bin/python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/megatron_fsdp/test_context.py tests/unit_tests/distributed/megatron_fsdp/test_experimental_fully_shard.py --tb=short --disable-warnings -rN` (`13 passed`)

PR #5513, opened 2026-06-26
Add forward all-gather overlap

wujingyue:
## Summary

Implements part of http://nv/mfsdp-schedule-design

- Add an `FsdpContext` that lazily owns a CUDA all-gather stream for one FSDP subtree.
- Run parameter unshard/all-gather on that context stream while compute remains on the default stream.
- Delay release of unsharded parameter storage so child-unit forward all-gathers can overlap default-stream GEMMs under a shared root context.
- Add context-sharing and memory-bound tests for nested FSDP units, plus a profiler test that asserts the forward child pipeline gives `num_children - 1` all-gather/compute overlaps.

## Scope
- This PR intentionally covers forward all-gather overlap only.
- Backward all-gathers are present in the profiled iteration, but they do not overlap compute in this PR because gradient reduction is still launched synchronously in `post_backward` before autograd reaches the next module’s `pre_backward` all-gather.
- The delayed gradient-reduction follow-up PR will address backward overlap.
- Reduce-scatter overlap is not included here.

## Testing
- `python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/megatron_fsdp/test_context.py tests/unit_tests/distributed/megatron_fsdp/test_experimental_fully_shard.py --tb=short --disable-warnings -rN`
- `git diff --check`

PR #5743, opened 2026-07-10
Support HSDP deferred DP-outer gradient reduction

### LINKED ISSUE #5616: MFSDP v2 should support HSDP and HFSDP placements

Implementation detail:

Expected outcome:

Development PRs:
- #5652

Achyuthan-S:
## Summary

Adds HSDP DP-outer gradient synchronization to the experimental Megatron-FSDP path, building on the microbatch context from #5652. Partially addresses #5616.

- Relax the `main_grad`/`main_weight` placement guard so `main_grad` may rest at a DP-outer-`Partial` accumulation state (`Partial` on an axis where `main_weight` is `Replicate`).
- Gate the gradient reduction on `is_last_microbatch`: reduce-scatter DP-inner and accumulate into `main_grad` every backward; reduce DP-outer and install the sharded parameter gradients only on the last microbatch.
- Compose the reduction one mesh axis at a time so a 2-D DP mesh can reduce-scatter DP-inner and then reduce DP-outer.
- Thread `is_last_microbatch` from the FSDP context through `post_backward`.
- Add a 2-D-mesh HSDP parity test driven via `microbatch(...)`.

## Scope and design notes (draft — feedback welcome)

- **HSDP only**: parameters and optimizer stay DP-outer-replicated (`[Replicate, Flat]`); gradients rest `[Partial, Flat]`. HFSDP (optimizer sharded across DP-outer) is a follow-up.
- The last-microbatch DP-outer reduction produces a per-step reduced-gradient buffer for the optimizer; a persistent/reused buffer could replace it.
- The gradient placement's DP-outer `Partial` reduce op must match the collective op (AVG). Symmetric-memory reduction across more than one axis is not yet supported (raises `NotImplementedError`).
- Single stream; overlapping DP-outer communication on a separate stream is orthogonal (see #5416 / #5513).

## Testing

Run on a 2×2 mesh (4 ranks):

python -m torch.distributed.run --standalone --nproc-per-node 4 -m pytest -q
  tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py

Covers HSDP parity against single-rank SGD for `num_microbatches` in {1, 3}; the existing 1-D FSDP path is unchanged.
cc @wujingyue @ahmadki


megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -314,3 +393,24 @@ def _get_parameter_owner(module: nn.Module, name: str) -> tuple[nn.Module, str]:
...
+    """
+    if len(grad_placements) != len(weight_placements):
+        return False
+    for grad_placement, weight_placement in zip(grad_placements, weight_placements):
+        if grad_placement == weight_placement:
+            continue
+        if isinstance(grad_placement, Partial) and isinstance(weight_placement, Replicate):
+            continue
+        return False
+    return True
```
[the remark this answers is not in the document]
Achyuthan-S: Restricted _grad_placements_reduce_to_weight to at most one deferred (Partial→Replicate) axis, matching the single-axis redistribute. Multi-axis deferral is a follow-up.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -275,27 +358,23 @@ def has_grad(parameters: Iterable[nn.Parameter]) -> bool:
                 grads, mesh=self.mesh, placements=[Partial(partial_op)] * self.mesh.ndim
             )
 
+        if self._defers_grad_reduction:
+            self._accumulate_deferred_grad(partial_grad, partial_op, is_last_microbatch)
+            for parameter in self.unsharded_parameters:
+                parameter.grad = None
+            return
```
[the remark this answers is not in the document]
Achyuthan-S: _reduced_grad is now cleared at the first microbatch of each step, so the previous step's buffer is released after zero_grad().

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -168,6 +183,74 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
...
+
+    outer_size = 2
+    inner_size = world_size // outer_size
+    mesh = init_device_mesh(
+        device.type, (outer_size, inner_size), mesh_dim_names=("dp_outer", "dp_inner")
+    )
+    torch.manual_seed(1234)
+    baseline = TinyModel().to(device)
+    model = TinyModel().to(device)
+    model.load_state_dict(baseline.state_dict())
```
wujingyue, line 224: Probably not needed because this test doesn't checkpoint.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -168,6 +183,74 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
     )
 
 
+@pytest.mark.parametrize("num_microbatches", [1, 3])
+def test_hsdp_losses_match_baseline(distributed_setup, num_microbatches):
```
wujingyue: In a separate test, can you do something like https://github.com/NVIDIA/Megatron-LM/pull/5513/changes#diff-47a8426b38ba29f0992fc9a06b4d2364c36ff9e0df43b1738bc7a95d310f610eR341-R354 to verify the dp-outer allgather/allreduce is only executed during first/last microbatch? That test doesn't need to verify numerics -- just count events.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -168,6 +183,74 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
...
+    mesh = init_device_mesh(
+        device.type, (outer_size, inner_size), mesh_dim_names=("dp_outer", "dp_inner")
+    )
+    torch.manual_seed(1234)
+    baseline = TinyModel().to(device)
+    model = TinyModel().to(device)
+    model.load_state_dict(baseline.state_dict())
+
+    fully_shard(model.fc1, mesh=mesh, placements=_hsdp_placements())
+    fully_shard(model.fc2, mesh=mesh, placements=_hsdp_placements())
```
wujingyue: fully_shard model as well so we exercise overlapping. Root FsdpModules can't overlap each other.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -151,13 +151,22 @@ def __init__(
                 "main_grad is built from main_weight tensor shapes on the same mesh, "
                 "and DBuffer layouts are deterministic from those shapes and mesh size."
             )
-            if self.main_grad.placements != self.main_weight.placements:
+            if not _grad_placements_reduce_to_weight(
```
wujingyue: Sounds a bit over-defensive. DBuffer.redistribute will error out when the given placements are not supported, so this extra check (while can be generated cheaply) doesn't add much value.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -151,13 +151,22 @@ def __init__(
...
+                    "Replicate (HSDP DP-outer accumulation). "
                     f"Got main_grad placements {self.main_grad.placements} and "
                     f"main_weight placements {self.main_weight.placements}."
                 )
+            # When main_grad placements differ from main_weight, main_grad rests at a
+            # DP-outer Partial accumulation state: each backward reduce-scatters DP-inner
+            # into it, and the deferred DP-outer reduction runs on the last microbatch.
+            self._defers_grad_reduction = self.main_grad.placements != self.main_weight.placements
+            self._grad_accumulating = False
+            self._reduced_grad: DBuffer | None = None
```
wujingyue: I'm not a fan of states. Can these be done in post-backward? E.g. 
```py
def post_backward:
  if is_last_microbatch:
    main_grad = main_grad.redistribute(main_weight.placements)
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -245,8 +254,86 @@ def release_unsharded_storage(self) -> None:
         # so keep the shared storage-release path.
         self._unsharded_model_weight.release_storage()
 
-    def reduce_gradients(self) -> None:
-        """Reduce full local gradients into sharded parameter gradients."""
+    def _reduce_partial_grad(
+        self, partial_grad: DBuffer, partial_op: dist.ReduceOp.RedOpType, *, out: "DBuffer | None"
+    ) -> DBuffer:
+        """Reduce an all-Partial gradient buffer to ``main_grad``'s placements.
```
wujingyue: What's this method for? For HSDP or HFSDP, main_grad->main_weight changes only one axis.

wujingyue (changes requested):
Thanks for the PR!

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Achyuthan-S: This one isn't about checkpointing — the seed makes every rank initialize the same weights. Without it each rank builds a different random TinyModel, so the FSDP-sharded parameter is assembled from mismatched per-rank inits and the all-gathered weight is inconsistent across ranks, which breaks the parity check. test_fully_shard_losses_match_baseline seeds for the same reason. I'd keep it unless you'd prefer initializing on one rank and broadcasting.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Achyuthan-S: Will add a separate profiler-based test that counts the DP-outer collective and asserts it only fires on the first/last microbatch (no numerics), following the #5513 pattern. I'll add it once the reduction structure below is settled, since it asserts on that path.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Achyuthan-S: Will fully_shard(model) as well so fc1/fc2 are children of a root unit and the overlap path is exercised, instead of two independent roots.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Agreed on cutting the state. The reason it's there: main_grad rests DP-outer-Partial across microbatches, so it can't be installed as a .grad DTensor until the last-microbatch reduce, and the per-microbatch accumulate/reset needed a signal separate from the sharded-grad path.

 To match your sketch I'd keep main_grad at its Partial placement, reduce-scatter DP-inner into it every backward, and at is_last_microbatch do redistribute(main_weight.placements) in post_backward and install the sharded grads there — dropping _defers_grad_reduction and _reduced_grad. The piece that still needs a decision is cross-microbatch accumulation/reset into main_grad: the existing has_sharded_grads path assumes sharded grads are set every backward, which we can't do while main_grad is Partial. Do you want a single "reset on the first microbatch after a last one" bit, or would you rather the reset be derived another way? I'll follow whichever keeps it stateless.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Will remove the guard and _grad_placements_reduce_to_weight and let DBuffer.redistribute raise on unsupported placements.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: In the gated design the inner reduce-scatter ([Partial,Partial]→[Partial,Flat]) and the last-microbatch outer reduce ([Partial,Flat]→[Replicate,Flat]) are each single-axis, so the multi-axis composition isn't needed — it's left over from an earlier every-backward version. I'll drop _reduce_partial_grad and use a direct redistribute.

Achyuthan-S:
> Thanks for the PR!

Anytime @wujingyue , you can always tag me in the future for any other issues , I find them very interesting and love solving them .

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
wujingyue: You are right! Let's keep it as is.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue: > it can't be installed as a .grad DTensor until the last-microbatch reduce

I don't follow this. main_grad is merely the storage backing `.grad`. It seems OK to redistribute `main_grad` from DP-outer-Partial to DP-outer-Replicate on the last microbatch and then immediately reset `.grad` according to the new main_grad. What am I missing?

Achyuthan-S: @wujingyue , Oh yes that makes sense you're right — I conflated backing .grad with accumulating. We don't need to touch .grad until the last microbatch: accumulate the DP-inner-reduced grad into main_grad (still DP-outer-Partial) every backward, and only at is_last_microbatch redistribute main_grad to main_weight.placements and set .grad from the result. No get_dtensor on a Partial buffer — that was my actual concern and it's a non-issue this way.

One detail to confirm: main_grad (Partial) and the Replicate buffer that redistribute produces are separate, so optimizer.zero_grad() clears .grad but not main_grad, which would otherwise carry across steps. To stay stateless I'd zero main_grad after the last-microbatch redistribute (and once at init), keeping per-microbatch accumulation a plain +=. Sound right, or would you reset it differently?

wujingyue: There's set_to_none, which makes a difference.

When set_to_none=True, zero_grad sets grads to None and main_grad will be [treated as garbage](https://github.com/NVIDIA/Megatron-LM/blob/48a887fecd01674346f724dcf44f417f455989f0/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py#L302) by reduce_gradients. 

When set_to_none=False, zero_grad zeros main_grad and keeps grads as views into main_grad. 

Either way, you may want to test both. set_to_none=True is the default. set_to_none=False is often used with cuda graph. Both are important.

Achyuthan-S: One structural check before I rewrite it: to reuse the zero_grad/set_to_none reset you described, main_grad has to back .grad every backward, so it should rest at main_weight.placements ([Replicate, Flat]) — a Partial buffer can't be a .grad view via get_dtensor. 

So per backward I'd reduce-scatter DP-inner into main_grad and accumulate through the existing has_sharded_grads path (which gives both set_to_none modes for free), and at is_last all-reduce the DP-outer (Replicate) axis in place so .grad becomes the final gradient. main_grad then transiently holds DP-outer-unreduced data between microbatches — exposed via .grad but not consumed until optimizer.step() after the last microbatch. 

Does that shape match your intent, or did you want main_grad to actually rest at [Partial, Flat] and redistribute at is_last?

wujingyue: By definition of HSDP, main_grad should be [Partial, Flat] between microbatches and [Replicate, Flat] after the last microbatch. It should back `.grad` unless `.grad` is None (e.g. set_to_none=True).

Achyuthan-S: @wujingyue  Reworked to keep main_grad honestly [Partial, Flat] between microbatches and finalized to [Replicate, Flat] on the last microbatch (dac14f228). To let it back .grad while it's Partial, I extended DBuffer.get_dtensor to emit a torch Partial placement — that keeps the existing zero_grad/has_sharded_grads accumulation working for both set_to_none modes with no added state. A finalized main_grad (placements == main_weight) marks a completed step, so the next reduction resets it to the accumulation placement.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
```diff
@@ -449,7 +449,10 @@ def get_dtensor(self, index: int) -> DTensor:
             elif isinstance(placement, Flat):
                 torch_placements.append(dist_tensor.Shard(0))
             elif isinstance(placement, Partial):
-                raise ValueError("Partial DBuffer placements cannot be represented as DTensor.")
+                # main_grad backs .grad while it rests DP-outer-Partial between
+                # microbatches, so a Partial placement must round-trip to a DTensor.
+                reduce_op = "avg" if placement.reduce_op == dist.ReduceOp.AVG else "sum"
```
wujingyue: I understood Partial takes a str, but can that be converted from dist.ReduceOp directly instead of using a switch that might miss cases for the future?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -268,42 +294,54 @@ def has_grad(parameters: Iterable[nn.Parameter]) -> bool:
...
+        deferred_axes = [
+            axis
+            for axis, placement in enumerate(self.main_weight.placements)
+            if isinstance(placement, Replicate)
+        ]
+        if deferred_axes and self._symm_mem_pool is not None:
+            raise NotImplementedError(
+                "Symmetric-memory gradient reduction does not support deferred DP-outer axes."
+            )
+        accumulation_placements = tuple(
```
wujingyue: Is this (or should this be) different from main_grad.placements?

wujingyue: Got it -- `self.main_grad` is a DBuffer that can be in different states, `[P,F]` or `[R,F]`. 

I think we can simplify by storing the initial main_grad_placements to `self` around [here](https://github.com/NVIDIA/Megatron-LM/blob/82e9dc69c9e6f8c27681f2cb6856a188187edf6b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py#L77-L79). Then we can simply use that instead of recomputing `accumulation_placements`.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -268,42 +294,54 @@ def has_grad(parameters: Iterable[nn.Parameter]) -> bool:
...
+            for index, sharded_parameter in enumerate(self.sharded_parameters):
+                sharded_parameter.grad = self.main_grad.get_dtensor(index)
+
+        if is_last_microbatch:
+            for axis in deferred_axes:
+                dist.all_reduce(
+                    self.main_grad.local_buffer,
+                    op=dist.ReduceOp.AVG,
+                    group=self.mesh.get_group(axis),
+                )
```
wujingyue: How about something like `self.main_grad = self.main_grad.redistribute(self.main_weight.placements)`? I understand this would create a new DBuffer, which is suboptimal. But given this is after the last microbatch the overhead may not affect the peak memory. Also, for HFSDP, the op will be reduce_scatter so we can't simply reuse the local_buffer.

wujingyue (commented):
LGTM overall!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
Achyuthan-S: Went with explicit avg/sum and a raise on any other op — an unhandled op now fails loudly instead of being silently mislabeled as "sum", which was the actual risk. I looked at deriving the string from the ReduceOp directly, but that mapping isn't stable across torch versions, so I kept it explicit.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Done — stored the initial gradient placement as self._accumulation_placements and reduce/reset against it instead of recomputing from main_weight each call.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Done — the finalize is now self.main_grad = self.main_grad.redistribute(self.main_weight.placements). That covers the HFSDP reduce-scatter case (and the buffer-size change) and uses the Partial's own reduce op, so the hardcoded AVG is gone. One note for HFSDP: since redistribute shrinks the buffer there, the next step's accumulation will need a re-allocated [Partial, Flat] buffer rather than the in-place placement reset — I'll handle that when HFSDP lands.

Achyuthan-S:
@wujingyue pushed the three review fixes as 6570526 (rebased onto the CI merge). copy-pr-bot de-authorized CI for the new tip — needs another /ok to test when you get a chance. The earlier run on 434a4d9 (the HSDP rework) passed; this just re-validates the small review changes.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -268,42 +297,44 @@ def has_grad(parameters: Iterable[nn.Parameter]) -> bool:
...
-                self.main_grad.local_buffer.div_(grad_divisor)
+        # main_grad accumulates DP-outer-Partial (Partial where main_weight is
+        # Replicate); symmetric memory does not support that deferral yet.
+        if (
+            self._accumulation_placements != self.main_weight.placements
+            and self._symm_mem_pool is not None
+        ):
+            raise NotImplementedError(
+                "Symmetric-memory gradient reduction does not support deferred DP-outer axes."
+            )
```
wujingyue: I understand the intention here but this check is probably fine to remove for now. Just do the DP-outer allreduce/reduce_scatter without symmetric memory.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -268,42 +297,44 @@ def has_grad(parameters: Iterable[nn.Parameter]) -> bool:
...
-            if has_sharded_grads:
-                self.main_grad.local_buffer.add_(reduced_grad.local_buffer)
-            else:
-                self.main_grad.local_buffer.copy_(reduced_grad.local_buffer)
-
-        if not has_sharded_grads:
-            for index, parameter in enumerate(self.sharded_parameters):
-                parameter.grad = self.main_grad.get_dtensor(index)
+            self.main_grad.local_buffer.copy_(reduced.local_buffer)
+            for index, sharded_parameter in enumerate(self.sharded_parameters):
```
wujingyue: Could make it an inner helper method to DRY

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -268,42 +297,44 @@ def has_grad(parameters: Iterable[nn.Parameter]) -> bool:
...
+            and self._symm_mem_pool is not None
+        ):
+            raise NotImplementedError(
+                "Symmetric-memory gradient reduction does not support deferred DP-outer axes."
+            )
+
+        # A finalized main_grad (placements == main_weight) means the previous step
+        # is done; reset it to the DP-outer-Partial accumulation placement.
+        if self.main_grad.placements == self.main_weight.placements:
+            self.main_grad.placements = self._accumulation_placements
```
wujingyue: Understood. 

Consider the following sketch: 
```py
  # This can only happen at the first microbatch.
  self.main_grad = self.main_grad.redistribute(self._accumulation_placements)
  reset sharded_parameters' grads if they are not None
  # ^^ should merely be a metadata change for HSDP. For HFSDP in the future, this should probably be `self.main_grad = a new empty or zeroed DBuffer` depending on has_sharded_grads. 

# At this point, self.main_grad.placements == self._accumulation_placements
the old code: https://github.com/NVIDIA/Megatron-LM/pull/5743/changes#diff-318602037e8dca86bd2f6aefe51cb128938ac95d1f01feff501dc9e83520a3b4L278-L306

your new code for `is_last_microbatch`: https://github.com/NVIDIA/Megatron-LM/pull/5743/changes#diff-318602037e8dca86bd2f6aefe51cb128938ac95d1f01feff501dc9e83520a3b4R332-R337
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Removed — the DP-outer finalize is a plain redistribute (no symmetric memory), so there was nothing to guard.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Done — pulled the grad install into _install_sharded_grads() and call it from the reset, the overwrite branch, and the finalize.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Adopted the structure — reset when placements != self._accumulation_placements (only happens on the first microbatch), refresh the sharded grads if they're set, and the HFSDP note. One correction: redistribute can't do the [Replicate,Flat] -> [Partial,Flat] relabel — Replicate -> Partial isn't one of its four transitions, so it raises. Since the buffer is overwritten right after, the HSDP reset is a direct self.main_grad.placements = self._accumulation_placements; HFSDP gets the fresh buffer you described.

wujingyue: > One correction: redistribute can't do the [Replicate,Flat] -> [Partial,Flat] relabel — Replicate -> Partial isn't one of its four transitions, so it raises

I'd add this as the fifth transition. `placements` is meant to be read-only so I'm not a fan of overwriting it. I should have made it a read-only property.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -268,42 +303,37 @@ def has_grad(parameters: Iterable[nn.Parameter]) -> bool:
...
+        # A non-accumulation main_grad means the previous step finalized it; this
+        # only happens on the first microbatch. Reset it to the DP-outer-Partial
+        # accumulation placement -- a metadata relabel for HSDP since the buffer is
+        # overwritten below. (HFSDP will instead need a fresh buffer here, since its
+        # finalized buffer was reduce-scattered and is smaller.)
+        if self.main_grad.placements != self._accumulation_placements:
+            self.main_grad.placements = self._accumulation_placements
+            if has_grad(self.sharded_parameters):
+                self._install_sharded_grads()
+        reduced = self._reduce_grad_axis(partial_grad, self._accumulation_placements)
```
wujingyue:
```suggestion
        reduced_grad = self._reduce_grad_axis(partial_grad, self._accumulation_placements)
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -268,42 +303,37 @@ def has_grad(parameters: Iterable[nn.Parameter]) -> bool:
...
-        can_reduce_into_main_grad = (
-            not has_sharded_grads and partial_grad.dtype == self.main_grad.dtype
-        )
-        reduce_axis = changed_mesh_axis(partial_grad.placements, self.main_grad.placements)
-        if reduce_axis is None:
-            raise RuntimeError("FSDP gradient reduction requires a changed placement axis.")
-        grad_divisor = self.mesh.size(reduce_axis) if partial_op == dist.ReduceOp.SUM else 1
-        if self._symm_mem_pool is not None:
-            partial_grad.rendezvous(reduce_axis)
-        if can_reduce_into_main_grad:
```
wujingyue, line 312: Is this optimization removed accidentally?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -245,8 +241,47 @@ def release_unsharded_storage(self) -> None:
         # so keep the shared storage-release path.
         self._unsharded_model_weight.release_storage()
 
-    def reduce_gradients(self) -> None:
-        """Reduce full local gradients into sharded parameter gradients."""
+    def _reduce_grad_axis(
```
wujingyue:
```suggestion
    def _reduce_partial_gradients(
```
Maybe? "Over its one changed axis" is an implementation detail that probably shouldn't appear in the name or even in the comment below.

wujingyue (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Renamed. It only appears in the accumulate branch now (reduced_grad); the overwrite branch reduces straight into main_grad via out= — see the optimization thread.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Yes — the restructure dropped it. Restored: the overwrite branch (set_to_none=True) reduces straight into main_grad via _reduce_partial_gradients(..., out=self.main_grad), no temporary + copy. The accumulate branch (set_to_none=False) still needs a temporary, since reduce_scatter overwrites its output and we're summing into the running gradient.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Renamed to _reduce_partial_gradients and dropped the "one changed axis" wording from the name and docstring.

Achyuthan-S:
@wujingyue Heads up before I push a merge — main moved ~14 commits since your approval, and one of them re-architected the exact path this PR touches: reduce_gradients was split into allocate_partial_grad_buffer / copy_gradients_to_partial_buffer / reduce_partial_gradients, driven from the new reduce_scatter_stream overlap in _reduce_gradient_groups. The branch no longer merges cleanly.

The merge is clean everywhere except that method — the guard relaxation, _accumulation_placements, and the DBuffer changes (get_dtensor Partial + the Replicate -> Partial redistribute) all merged fine. What's left is re-expressing the HSDP deferral on top of the new reduce_partial_gradients: a reset prologue (main_grad.placements != accumulation -> redistribute back) and an is_last_microbatch finalize epilogue (DP-outer reduce + install sharded grads), plus threading is_last_microbatch through the stream-overlapped path. Conveniently, the new reduce_partial_gradients already carries the can_reduce_into_main_grad optimization, so that part converges.

One semantic I'd want your read on: the DP-outer finalize (all_reduce for HSDP) would now run on reduce_scatter_stream. It's serialized after the last microbatch's reduce so it's functionally correct, but it's a new interaction with the overlap path.

How would you like to handle it — should I push the merge with HSDP re-expressed on the new substrate and re-trigger CI for a fresh review, or would you rather take the update / have the overlap-refactor author weigh in since it's your team's new code? Happy to do the re-integration; just don't want to step on the stream work.

(Your four approval comments are handled locally: the Replicate -> Partial fifth transition + not mutating .placements merged cleanly; the reduced_grad rename, the reduce-into-main_grad restore, and the _reduce_grad_axis rename are largely subsumed by the new reduce_partial_gradients.)

wujingyue:
> the overlap-refactor author

That's me. 

> should I push the merge with HSDP re-expressed on the new substrate and re-trigger CI for a fresh review

Yes. I'll take another look after the merge.

Achyuthan-S:
@wujingyue --- Merged main and re-expressed the HSDP deferral on the new pipeline as discussed: reset prologue + is_last_microbatch finalize folded into reduce_partial_gradients, is_last threaded through _reduce_gradient_groups. The DP-outer finalize runs on reduce_scatter_stream, serialized after the last microbatch's reduce. Ready for /ok to test and your re-look.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
```diff
@@ -417,6 +420,36 @@ def scatter(
...
+        if not isinstance(self.placements[axis], Replicate):
+            raise ValueError(
+                f"replicate_to_partial() requires Replicate placement on axis {axis!r}."
+            )
+
+        placements = list(self.placements)
+        placements[axis] = new_placement
+        _validate_placements(placements)
+        out = self._create_or_validate_out(placements, out)
+        out.local_buffer.copy_(self.local_buffer)
```
wujingyue: This looks wrong. It shouldn't copy data when going from replicate to partial.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -296,37 +300,42 @@ def has_grad(parameters: Iterable[nn.Parameter]) -> bool:
...
-        reduced = self._reduce_grad_axis(partial_grad, self._accumulation_placements)
 
-        # set_to_none=True cleared the sharded grads -> overwrite main_grad and
-        # install .grad; else they are views into a zeroed main_grad -> accumulate.
-        if has_grad(self.sharded_parameters):
-            self.main_grad.local_buffer.add_(reduced.local_buffer)
+        # zero_grad(set_to_none=True) clears sharded parameter grads, so this
+        # backward can reduce directly into main_grad. zero_grad(set_to_none=False)
+        # leaves sharded grads installed, so this backward accumulates into main_grad.
+        has_sharded_grads = has_grad(self.sharded_parameters)
```
wujingyue: Move this up to dedup with line 309?

wujingyue (changes requested):

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -168,6 +183,77 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
     )
 
 
+@pytest.mark.parametrize("set_to_none", [True, False])
+@pytest.mark.parametrize("num_microbatches", [1, 3])
+def test_hsdp_losses_match_baseline(distributed_setup, num_microbatches, set_to_none):
```
wujingyue, line 199: I believe I asked for another test verifying HSDP triggers the extra allreduce. Could you add that?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
Achyuthan-S: Right — replicate_to_partial now reinterprets the local buffer instead of copying it. Replicate and Partial share the same local layout, so it returns a from_local relabel (value-preserving for AVG). The reset doesn't need the old data anyway: the buffer is already zeroed for set_to_none=False or overwritten for set_to_none=True. Dropped the copy and the SUM rescale.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Done — hoisted has_sharded_grads above the reset and reused it there.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Achyuthan-S: Added test_hsdp_defers_dp_outer_allreduce_to_last_microbatch: profiles a 3-microbatch step and asserts the DP-outer all-reduce fires at all (plain DP never issues one) and that the reduce-scatter count is exactly num_microbatches x the all-reduce count — i.e. DP-inner reduce-scatters every microbatch, DP-outer all-reduces only on the last.

Achyuthan-S:
One addition since the above: the test now uses MultiChildModel with fully_shard(model), so the child units reduce under a shared root context (the overlap path) rather than as independent roots — that also covers your earlier "fully_shard(model) so we exercise overlapping" note.

Achyuthan-S:
Done in the parity test too — test_hsdp_losses_match_baseline now uses MultiChildModel with fully_shard(model), so the child layers reduce under a shared root context rather than as independent roots.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
```diff
@@ -417,6 +420,37 @@ def scatter(
...
+        if not isinstance(new_placement, Partial):
+            raise ValueError(f"replicate_to_partial() requires a Partial target on axis {axis!r}.")
+        if not isinstance(self.placements[axis], Replicate):
+            raise ValueError(
+                f"replicate_to_partial() requires Replicate placement on axis {axis!r}."
+            )
+        if new_placement.reduce_op != dist.ReduceOp.AVG:
+            raise NotImplementedError(
+                f"replicate_to_partial() supports AVG only, got {new_placement.reduce_op!r}."
+            )
```
wujingyue: Why?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
```diff
@@ -417,6 +420,37 @@ def scatter(
...
+        if out is not None:
+            raise NotImplementedError("replicate_to_partial() does not support an out buffer.")
+
+        placements = list(self.placements)
+        placements[axis] = new_placement
```
wujingyue: Nit: new_placements = self.placements[:axis] + (new_placement,) + self.placements[axis + 1:]

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
```diff
@@ -417,6 +420,37 @@ def scatter(
...
+        return DBuffer.from_local(
+            self.local_buffer, self.mesh, placements, self.layout.tensor_shapes
+        )
```
wujingyue: It might be more readable to have these inlined to `redistribute` and remove `replicate_to_partial`. `replicate_to_partial` is not a conventional communication op as `allgather` and `reduce_scatter`, so it doesn't seem worth adding a separate method for it.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -168,6 +183,154 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
...
+        event
+        for event in cuda_events
+        if "nccl" in event.name.lower()
+        and ("reducescatter" in event.name.lower() or "reduce_scatter" in event.name.lower())
+    ]
+    all_reduce_events = [
+        event
+        for event in cuda_events
+        if "nccl" in event.name.lower() and "allreduce" in event.name.lower()
+    ]
```
wujingyue: Nit: while there's https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html, it might be worth creating a helper to extract communication events from `prof` and using it from the overlap test as well. 

The signature could be
```
def collect_communication_events(prof: Profile, regex) -> list[Event]:
```

Either in this PR or I can do it separately.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -168,6 +183,154 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
...
+    # HSDP must trigger the DP-outer all-reduce that plain DP never issues.
+    assert all_reduce_events, [event.name for event in cuda_events]
```
wujingyue: assert len(all_reduce_events) == num_children + 1

to be slightly stricter?

wujingyue (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
Achyuthan-S: Replicate -> Partial is a value-preserving relabel only for AVG (the mean of identical per-rank locals is that value). SUM reduces to value x axis_size, so a correct SUM relabel would need a 1/axis_size rescale -- it couldn't be zero-copy. No caller needs SUM, so I raise loudly rather than silently mislabel. Now inlined into redistribute (see below).

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
Achyuthan-S: Gone with the inlining below -- redistribute already has the full new_placements tuple, so there's nothing to rebuild.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
Achyuthan-S: Done -- inlined the Replicate -> Partial relabel into redistribute's dispatch and removed replicate_to_partial. Agreed it isn't a conventional communication op, so it doesn't warrant its own method.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Achyuthan-S: Added _nccl_events(cuda_events, *name_fragments) and routed both this test and the overlap test through it. Kept name fragments rather than a regex so the exact substring matching is preserved; happy to switch to the regex signature if you'd rather.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Achyuthan-S: Done -- assert len(all_reduce_events) == num_children + 1 (one per child layer plus the root bias group), alongside the reduce-scatter == num_microbatches x all-reduce check.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
wujingyue: That's right -- thanks for clarification. 

SUM is OK when the content is zero or garbage -- actually a common case before the first microbatch. Let's think more about that in a separate PR when it's needed.

wujingyue:
Could you add an `event.activity_type == "kernel"` check in `_nccl_events`? In the current PyTorch profiler trace, the actual NCCL device launch has `activity_type="kernel"`, while `nccl:_reduce_scatter_base` is a `gpu_user_annotation` that is also reported as a CUDA event. The annotation currently causes reduce-scatter to be counted twice. Filtering by activity type avoids counting non-kernels and is more robust than matching the `nccl:` name prefix. I verified the focused 4-GPU test passes with this check.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -111,6 +126,16 @@ def _events_overlap(first, second) -> bool:
     )
 
 
+def _nccl_events(cuda_events, *name_fragments):
+    """CUDA NCCL events whose name contains any of ``name_fragments`` (case-insensitive)."""
+    return [
+        event
+        for event in cuda_events
+        if "nccl" in event.name.lower()
```
wujingyue, line 134:
```suggestion
        if "nccl" in event.name.lower() and event.activity_type == "kernel"
```

wujingyue: I let github apply the change but it failed to sign off the commit, causing DSO failures. I think you'll have to fix it yourself.

ericharper (approved):

yaoyu-33 (approved):

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Achyuthan-S: Fixed — signed it off and ran black (it wanted each `and` on its own line). Kept your activity_type == "kernel" filter; agreed it's the robust fix since the gpu_user_annotation was double-counting the reduce-scatters.

Achyuthan-S:
@wujingyue @ericharper Signed off and black-formatted the activity_type change in 96764cdda — DCO is green now. The force-push cleared CI authorization and disabled auto-merge, so it needs one more /ok to test 96764cdda and auto-merge re-enabled (or a manual merge once green). Thanks!

jiemingz (approved):

Issue #5675, opened 2026-07-06
Automatic Checkpoint Converter to HybridModel


Phlip79:
As we are migrating from `GPTModel` to `HybridModel`, it would make the process easier for users if `GPTModel` checkpoints can be converted automatically before training begins.

We will introduce a `--use-legacy-model` flag:

- HybridModel ckpt -> HybridModel training
- GPTModel ckpt + `--use-legacy-model` -> GPTModel training
- GPTModel ckpt -> HybridModel training

PR #5753, opened 2026-07-11
Add automatic GPT-to-Hybrid checkpoint conversion at training-load time

Chasing1020:
Implements the auto-migration flow requested in issue #5675 so that GPTModel checkpoints can be loaded into a HybridModel training run without a manual offline conversion step.

Three load-time scenarios are covered:
  1. HybridModel checkpoint into HybridModel training: no conversion.
  2. GPTModel checkpoint with --use-legacy-model: keep training as GPTModel (the Hybrid entrypoint routes the config back to gpt_config_from_args, so the runtime is GPTModel and the conversion branch never fires).
  3. GPTModel checkpoint into HybridModel training: rank 0 rewrites the source distributed checkpoint into hybrid layout in a temp directory, then every rank re-loads it collectively via DCP.

The pure state-dict transformation logic is factored into a new package megatron.core.models.hybrid.conversion, imported by both the offline CLI at tools/checkpoint/gpt_hybrid_conversion.py and the in-memory auto-conversion hook in megatron/training/checkpointing.py, so the two entrypoints share one source of truth.

SSM (M) layers are initialized from scratch on rank 0 following the same distribution as MambaMixer.__init__, persisted to the temporary dist checkpoint, and then reshared consistently across TP/PP/FSDP by DCP on reload. Attention and MLP tensors round-trip bit-for-bit.

Guardrails: auto-conversion requires torch_dist source format, --finetune, and --no-load-optim (SSM weights are fresh, so the source optimizer state is not meaningful). GPT-incompatible source features (MLA, MTP, gated_delta_net, heterogeneous layer specs, interleaved dense/MoE) are rejected up front by the shared compatibility validators.

Other changes:
  * New --use-legacy-model and --auto-convert-checkpoint-dir flags.
  * write_latest_iteration_marker now raises when it is handed a non-iter_ directory instead of silently no-oping.
  * Fix a stale --use-legacy-models reference in the Elastification hybrid entrypoint.
  * Unit and multi-rank tests covering pattern parsing, mapping, validators, SSM init shapes, end-to-end round trip, and the rank-zero-only conversion path.


guihong-nv:
Hi @Chasing1020 , thank you so much for this contribution. This is great work! We really appreciate the effort you put into it. However, we're already working on this feature internally as part of a broader checkpoint-conversion effort for hybrid models, and our implementation needs to align with some internal design constraints that are hard to coordinate through an external PR. So we have to close this PR in favor of the internal work. Apologies that this wasn't communicated earlier on #5675.

 Thanks again for the high-quality contribution, and we'd be glad to see more PRs from you!

PR #5756, opened 2026-07-12
Validate num_attention_heads is divisible by num_query_groups in TransformerConfig

### LINKED ISSUE #5755: [BUG] TransformerConfig accepts num_attention_heads not divisible by num_query_groups, then crashes with a cryptic shape error inside attention
**Describe the bug**

`TransformerConfig.__post_init__` validates `num_query_groups` against `tensor_model_parallel_size`, but never checks that `num_attention_heads` is divisible by `num_query_groups` — a hard requirement of grouped-query attention (the per-group head count is computed with floor division in `attention.py`). An invalid GQA configuration therefore passes validation and only fails much later, in the middle of the first forward pass, with a shape error that gives no hint that `num_query_groups` is the problem.

Non-positive values are also accepted (`num_query_groups=0` currently passes validation because `0 % tp_size == 0`).

Tag @NVIDIA/mcore-oncall

```python
from megatron.core.transformer.transformer_config import TransformerConfig

# 32 % 5 != 0 -- should be rejected, but constructs fine:
config = TransformerConfig(
    num_layers=1, hidden_size=64, num_attention_heads=32, num_query_groups=5
)
```

Building a GPT model from this config and running one forward pass (single GPU, local layer spec, tested on current `main`, 48a887fec) crashes with:

```
  File ".../megatron/core/transformer/attention.py", line 1829, in get_query_key_value_tensors
    mixed_qkv = mixed_qkv.view(*new_tensor_shape)
RuntimeError: shape '[8, 1, 5, 16]' is invalid for input of size 672
```

**Expected behavior**

A clear `ValueError` at config-construction time, consistent with the neighboring validations in `__post_init__`, e.g.:

```
ValueError: num_attention_heads (32) must be a multiple of num_query_groups (5).
```

**Additional context**

- The gap exists since GQA support was introduced (6dd0c7bd7, 2023-07).
- Users who pick `num_query_groups` manually for GQA hit this easily; the distance between the mistake (config) and the failure (a `view()` deep inside attention) makes it hard to diagnose.
- I have a fix ready (validation + unit tests) and will link a draft PR to this issue.

huthvincent:

# What does this PR do?

Rejects invalid grouped-query-attention configurations at config-construction time instead of letting them crash with a cryptic `view()` shape error deep inside `attention.py`.

`TransformerConfig.__post_init__` now raises a clear `ValueError` when:
- `num_query_groups` is not positive (previously `num_query_groups=0` passed validation because `0 % tp_size == 0`), or
- `num_attention_heads` is not divisible by `num_query_groups` (hard GQA requirement; the per-group head count is computed with floor division).

Without this check, e.g. `num_attention_heads=32, num_query_groups=5` constructs fine and then fails on the first forward pass with:

```
RuntimeError: shape '[8, 1, 5, 16]' is invalid for input of size 672
```

Also updates one existing unit test (`test_trtllm_single_device_converter.py`) that used `num_query_groups=0` as a placeholder; it now uses `num_query_groups=num_attn_heads` (plain MHA, matching the mocked weight layout).

All existing in-tree configs and tests satisfy the new check (verified by grepping every `num_query_groups` usage in `megatron/` and `tests/`).

## Issue tracking

Linked issue: Fixes #5755

### Pre-checks

- [x] I have added relevant unit tests (`tests/unit_tests/transformer/test_transformer_config.py`, 5 cases: valid GQA, default groups, non-divisible, groups > heads, non-positive)
- [ ] I have added relevant functional tests (not applicable — config validation)
- [ ] I have added relevant documentation (error message is self-documenting)
- [x] I have run the autoformatter on the changed files (black 24.10, isort)

**Local verification** (single GPU):

```
$ torchrun --nproc-per-node 1 -m pytest -q tests/unit_tests/transformer/test_transformer_config.py
5 passed
$ torchrun --nproc-per-node 1 -m pytest -q tests/unit_tests/export/trtllm/test_trtllm_single_device_converter.py
5 passed
```


cspades:
@huthvincent Want me to run CI/CD?

huthvincent:
@cspades Yes, please, thank you! CI would be much appreciated. All three PRs (#5756, #5759, #5761) are small, self-contained bug fixes; happy to have CI run on them whenever convenient.

cspades:
/ok to test [f6f5f53](https://github.com/NVIDIA/Megatron-LM/pull/5756/commits/f6f5f53684f9e53e4c348835e4b75c1309d87909)

huthvincent:
@cspades Sorry for the extra round. Right after your `/ok to test`, I pushed a small follow-up commit (`dd8db013c`) that adds the required NVIDIA copyright headers to the changed files so the `copyright-check` job passes (my new test file used the old header format, and the trtllm test file I touched had no header). Since that moved the PR head, the earlier approval is now tied to the previous commit and CI did not start on the new one. Could you re-run `/ok to test dd8db013c` when convenient? Thank you, and sorry again for the churn.

cspades:
/ok to test [dd8db01](https://github.com/NVIDIA/Megatron-LM/pull/5756/commits/dd8db013c893ed3ba60810630272c4864fcbe908)

huthvincent:
@cspades Update on this one. CI caught a real regression in my earlier commit: the `num_query_groups` positivity check I added rejected `num_query_groups == 0`, which is a valid sentinel (equal to `num_attention_heads`) relied upon by many minimal configs, so it broke a number of distributed tests. I have pushed `87ebbf113`, which enforces only the divisibility requirement and only when `num_query_groups > 0`, reverts the unrelated test change, and adds coverage for the zero/minimal case. Verified locally against the configs that were failing. Sorry for the churn again. Could you re-run `/ok to test 87ebbf113` when you have a chance? Thank you for your patience.

guihong-nv:
Thanks for the fix. The divisibility check looks good. I think one blocking edge case remains.
The current guard only runs when `num_query_groups > 0`, so zero and negative values still pass validation. In particular, the new test allows `num_attention_heads=2, num_query_groups=0`, but zero is not actually normalized to `num_attention_heads`; only None is. Attention initialization later divides by `num_query_groups`, so that configuration still fails downstream rather than at config construction.

huthvincent:
@guihong-nv Great catch, thank you. You are right that `0` was not normalized and would divide by zero during attention initialization. I pushed `15a230d43`, which treats `num_query_groups == 0` the same as `None` (both normalize to `num_attention_heads`), so `num_attention_heads=2, num_query_groups=0` now becomes a valid MHA config instead of failing downstream, while minimal configs (where `num_attention_heads` defaults to 0) stay valid. Zero and negative values are now rejected at construction whenever attention heads are present. I verified locally that the TRT-LLM converter and inference paths are unaffected, since they resolve to the same effective value, and updated the tests accordingly.

huthvincent:
@cspades Sorry to bother you again. I pushed `15a230d43` to address @guihong-nv's review feedback on the `num_query_groups == 0` edge case. Could you re-run `/ok to test 15a230d43` when you have a chance, so CI can validate it? Thank you.

cspades:
/ok to test [15a230d](https://github.com/NVIDIA/Megatron-LM/pull/5756/commits/15a230d433ed7e8b4bc8fbcab5d8870dfc0ad36c)

cspades:
@huthvincent You can ping @NVIDIA/mcore-oncall and they'll help you!

Phlip79 (approved):

zhongbozhu (approved):

jaredcasper (approved):

huthvincent:
Thanks @Phlip79, @zhongbozhu and @jaredcasper for the reviews, and @Phlip79 for taking the rebase on.

Is there anything left needed from me here? It still shows `blocked`, and on the current head (`2fa33b5`) the unit-test and integration matrices show as cancelled rather than completed — I cannot tell from outside whether that is what is holding it, or just a superseded run. The branch is also about 17 commits behind `main` now, so I am happy to rebase or to re-run anything that would help.

huthvincent:
@wujingyue — I've rebased this onto current `main`. Two things worth flagging.

**@Phlip79 — this replays over the rebase commit you pushed on 08-25.** Same content, and your `Signed-off-by` is still on the commit, but I'm the committer now. Sorry about that — I didn't see a way to rebase and keep it.

The conflict was confined to `tests/unit_tests/transformer/test_transformer_config.py`, where upstream had added `sequence_packing` tests in the same region as the class this PR adds. The two blocks are independent, so it was a keep-both resolution — upstream's first, then ours. The production change in `megatron/core/transformer/transformer_config.py` is `+13 −1` and identical line-for-line to what you reviewed; the added test lines are identical too, just sitting after upstream's new block instead of the old one. Whole PR is still `+57 −1`, now 0 behind `main`, and the three approvals are still on it.

The rebase did invalidate the `/ok to test` that named `2fa33b5`. The sha you'd need is `4915c835fb21797ca7ea2981dcb317b20b48d4f9`. That's the only thing I can see from outside — please tell me if something else is needed.

megatron/core/transformer/transformer_config.py
```diff
@@ -1556,9 +1556,21 @@ def __post_init__(self):
         if self.kv_channels is None:
             self.kv_channels = self.hidden_size // self.num_attention_heads
 
-        if self.num_query_groups is None:
+        # num_query_groups == 0 is treated the same as None (both mean "use num_attention_heads").
```
wujingyue: Why? Isn't this simply a user configuration error? The div-by-zero error message is pretty clear about it.

megatron/core/transformer/transformer_config.py
```diff
@@ -1556,9 +1556,21 @@ def __post_init__(self):
...
+        # This keeps minimal configs valid (num_attention_heads itself defaults to 0) and ensures
+        # a real attention config never reaches attention initialization with a zero query-group
+        # count, which would otherwise divide by zero downstream.
+        if self.num_query_groups is None or self.num_query_groups == 0:
             self.num_query_groups = self.num_attention_heads
 
+        if self.num_attention_heads > 0 and (
+            self.num_query_groups <= 0 or self.num_attention_heads % self.num_query_groups != 0
+        ):
```
wujingyue: Can you try your best to simplify? This looks way more complicated than necessary.

huthvincent:
Thanks — I've acted on the second one, but on the first I think I have to push back, with a case from the tree.

**On `num_query_groups == 0`.** I tried deleting that block and it turns your CI red: `0` is already used as a sentinel meaning "same as `num_attention_heads`", and `tests/unit_tests/export/trtllm/test_trtllm_single_device_converter.py:28` constructs `TransformerConfig(num_attention_heads=2, num_query_groups=0, ...)` directly. Without the normalisation that config raises instead of building.

This is also the path this PR already took, earlier in the thread. My first version added a positivity check that rejected `0`, and CI caught it — it broke a number of distributed tests, because `0` was relied on as that sentinel (2026-07-14). @guihong-nv then pointed out the opposite hole: `0` was *not* being normalised, so `num_attention_heads=2, num_query_groups=0` was accepted at config time and then divided by zero during attention init (2026-07-15). Normalising `0` like `None` is what resolved both, and it's what the three approvals are on.

So I'd rather keep `0` accepted — but you're right that the reason was invisible, so it's now two lines of comment naming the test instead of four lines of prose.

If you'd still prefer `0` to be a hard error, I'm happy to do that instead — it just means updating the trtllm export test in this PR, and I didn't want to change an unrelated test without asking first.

**On simplifying.** Cut what I could: the comment is 4 lines → 2, and production is `+11 −1`. The rest I believe is load-bearing rather than defensive:

- `num_attention_heads > 0` — it defaults to `0` in the minimal configs many non-attention tests build, where `num_query_groups` also ends up `0`, so without the guard `0 % 0` raises `ZeroDivisionError` on configs that are valid today.
- `num_query_groups <= 0` — after normalisation only negatives can reach it, and `4 % -1 == 0` in Python, so a negative would otherwise pass the divisibility test. This is the half @guihong-nv asked for.

New head is `5aecbe5a0` — sorry for the churn, that's two rebases and a fixup today.

huthvincent:
Small correction to my first bullet above — I named the wrong error, and the clause it justifies is the one you're most likely to ask about next.

Removing the `num_attention_heads > 0` guard does **not** give a `ZeroDivisionError`. `num_query_groups <= 0` is the left operand of the `or`, so it short-circuits and `0 % 0` is never evaluated. What a zero-head minimal config would actually hit is the new `ValueError` itself: `num_query_groups (0) must be a positive divisor of num_attention_heads (0)`. The guard is still needed — just for that reason, not the one I gave.

For completeness, since it bears on your first comment: `ZeroDivisionError` on an explicit `num_query_groups=0` needs *both* the normalisation and the `<= 0` clause gone. With the sentinel in place, `0` becomes `num_attention_heads` before the check, so `32 % 0` is unreachable. That two-line variant is the shape your comment described and it's still on the table — it just needs the trtllm converter test changed with it.

Phlip79:
Please resolve https://github.com/NVIDIA/Megatron-LM/pull/5756#discussion_r3873045805 before we merge your PR.

wujingyue (changes requested):
Can you address the remaining comments? Then we can merge this. Thanks for the PR!

megatron/core/transformer/transformer_config.py
huthvincent: Agreed, dropped. `num_query_groups=0` is no longer special-cased — it now hits the modulo in `__post_init__` rather than being normalised to `num_attention_heads`. The only place in tree that passed `0` to mean "same as `num_attention_heads`" now says that explicitly (`tests/unit_tests/export/trtllm/test_trtllm_single_device_converter.py`).

megatron/core/transformer/transformer_config.py
huthvincent: Simplified to one condition:

```python
if self.num_attention_heads > 0 and self.num_attention_heads % self.num_query_groups != 0:
```

The `> 0` guard stays because the minimal configs many non-attention tests build leave `num_attention_heads` at `0`; without it `TransformerConfig(num_layers=1, kv_channels=1)` raises `ZeroDivisionError`. Dropping the `<= 0` clause does mean a negative `num_query_groups` that divides `num_attention_heads` is accepted again, as it is on `main` today.

The `Signed-off-by` from the 08-25 rebase is still on the commit and the code under it changed with this push.

guihong-nv:
/ok to test [d7cbac9](https://github.com/NVIDIA/Megatron-LM/pull/5756/commits/d7cbac9fe4f8f0e9f4033d36566188e7354f9afd)

megatron/core/transformer/transformer_config.py
```diff
@@ -1586,6 +1586,12 @@ def __post_init__(self):
         if self.num_query_groups is None:
             self.num_query_groups = self.num_attention_heads
 
+        if self.num_attention_heads > 0 and self.num_attention_heads % self.num_query_groups != 0:
```
wujingyue:
```suggestion
        if self.num_attention_heads % self.num_query_groups != 0:
```

wujingyue: Is this equivalent and simpler?

wujingyue (changes requested):

megatron/core/transformer/transformer_config.py
huthvincent: Simpler, yes — but not equivalent. `num_attention_heads` is `field(default=0, …)` (line 151) and `num_query_groups` normalises to it just above, so a config that names no attention heads reaches this line with both at `0` and your version evaluates `0 % 0`. Applying the suggestion verbatim and running this PR's own test file gives `1 failed, 29 passed` — the failure is `test_minimal_config_without_attention_heads_is_allowed`, and `TransformerConfig(num_layers=1, kv_channels=1)` raises `ZeroDivisionError: integer modulo by zero`.

I've put that reason in a code comment above the guard.

tests/unit_tests/export/trtllm/test_trtllm_single_device_converter.py
```diff
@@ -25,7 +27,7 @@ def test_get_model_weights_converter(self, mocker):
         model_config = TransformerConfig(
             num_layers=num_layers,
             num_attention_heads=num_attn_heads,
-            num_query_groups=0,
+            num_query_groups=num_attn_heads,
```
wujingyue: Why is this changed?

megatron/core/transformer/transformer_config.py
wujingyue: Got it -- I missed the coupling between num_attention_heads and num_query_groups. 

Is it more readable to say
```suggestion
        if self.num_query_groups > 0 and self.num_attention_heads % self.num_query_groups != 0:
```
All you need is to prevent a div-by-zero. It's fine for the numerator to be zero.

wujingyue (changes requested):

megatron/core/transformer/transformer_config.py
huthvincent: Taken. Two consequences worth naming: `num_query_groups=0` with heads present is accepted again — as it is on `main` — which is the case guihong-nv asked to close in July; and with `0` accepted the trtllm converter test needs no change, so I've reverted it. The guard reads as what it is now, so the comment came out with it.

tests/unit_tests/export/trtllm/test_trtllm_single_device_converter.py
huthvincent: It isn't any more — reverted. It was only needed while `num_query_groups=0` raised; with the guard on the divisor it doesn't, so the file is untouched.

wujingyue (approved):

PR #5761, opened 2026-07-12
Enter both autocast and fp8 contexts in combined 1F1B execution

### LINKED ISSUE #5760: [BUG] combined_1f1b: `with context_manager and outer_fp8_context:` never enters the autocast context
**Describe the bug**

`megatron/core/pipeline_parallel/combined_1f1b.py` wraps the combined forward/backward model-chunk execution with:

```python
with context_manager and outer_fp8_context:  # autocast context and delayed fp8 context
```

`and` is a boolean operator, not context-manager composition: since a context-manager object is always truthy, the expression evaluates to `outer_fp8_context` alone, and `context_manager` (the `torch.autocast` context built from `config.autocast_dtype` when `config.enable_autocast=True`) is **never entered**. The comment documents the intent of entering both.

As a result, with `enable_autocast=True` and `overlap_moe_expert_parallel_comm=True` (the path that uses the combined 1F1B schedule), the requested autocasting is silently not applied around the layer compute of the combined pass. There is no error or warning; the model simply runs without autocast.

Tag @NVIDIA/mcore-oncall

The semantics are demonstrable in isolation:

```python
class Probe:
    def __init__(self):
        self.entered = False
    def __enter__(self):
        self.entered = True
    def __exit__(self, *a):
        return False

autocast_ctx, fp8_ctx = Probe(), Probe()
with autocast_ctx and fp8_ctx:   # the pattern in combined_1f1b.py
    pass
print(autocast_ctx.entered, fp8_ctx.entered)  # False True
```

**Expected behavior**

Both contexts entered, as the inline comment states:

```python
with context_manager, outer_fp8_context:
```

which prints `True True` in the snippet above.

**Additional context**

- Introduced in ae1c88296 ("feat(MoE): Support Expert Parallel A2A Overlapping - (02)").
- This is the only occurrence of the pattern in the code base (checked with a grep for `with <name> and <name>` across `megatron/`).
- One-line fix; I will link a draft PR to this issue.

huthvincent:

# What does this PR do?

One-line fix in `megatron/core/pipeline_parallel/combined_1f1b.py`:

```diff
- with context_manager and outer_fp8_context:  # autocast context and delayed fp8 context
+ with context_manager, outer_fp8_context:  # autocast context and delayed fp8 context
```

`and` returns one of its operands (here always `outer_fp8_context`, since context managers are truthy), so the autocast context built from `config.enable_autocast` / `config.autocast_dtype` was never entered around the combined forward/backward chunk execution. The comma form enters both, matching the inline comment's stated intent.

Impact: with `enable_autocast=True` and the combined 1F1B schedule (`overlap_moe_expert_parallel_comm=True`), autocasting was silently skipped for the layer compute of the combined pass.

## Issue tracking

Linked issue: Fixes #5760

### Pre-checks

- [ ] I have added relevant unit tests (no new test: the change restores documented intent at Python-semantics level; the demonstration snippet is in the linked issue, and the path is exercised by the existing combined-1F1B functional tests)
- [ ] I have added relevant functional tests (covered by existing ones)
- [x] I have added proper typing to my code (no signature changes)
- [ ] I have added relevant documentation (not applicable)
- [x] I have run the autoformatter on the changed files (black 24.10, isort)


cspades:
/ok to test [f5a48d6](https://github.com/NVIDIA/Megatron-LM/pull/5761/commits/f5a48d67ece12ea93491d6dfd5465ce8c061ae6b)

cspades:
@Wohox could you double-check this bug-fix? Curious why it was written like that.

megatron/core/pipeline_parallel/combined_1f1b.py
```diff
@@ -457,7 +457,7 @@ def forward_backward_step():
 
     b_grad = b_output_tensor_grad[0] if b_model else None
     # combined forward and backward model chunk execution of two micro-batches
-    with context_manager and outer_fp8_context:  # autocast context and delayed fp8 context
+    with context_manager, outer_fp8_context:  # autocast context and delayed fp8 context
```
yashaswikarnati, line 454: good catch, could we add to the unit test enable_autocast=True so this path gets exercised

yashaswikarnati (approved):
approving with a small comment to enable this flag in the unit test

huthvincent:
@yashaswikarnati Thanks for the review. I looked into this and there is a catch worth flagging.

`test_fsdp_1f1b_overlap` compares the overlap path against `fsdp_train_step` with `torch.equal`, and `fsdp_train_step` calls `model(**data)` without any autocast context. So simply setting `enable_autocast=True` would compare an autocast run against a non-autocast run, and the bit exact assertion would fail. Before this fix the two happened to agree precisely because line 460 was dropping autocast, which is how the bug stayed hidden.

To exercise the path properly I think the reference step needs the same autocast context, for example wrapping `model(**data)` in `torch.autocast("cuda", dtype=config.autocast_dtype)` when `config.enable_autocast` is set. I am happy to do that.

One caveat: I only have a single GPU without TransformerEngine locally, so I cannot run the a2a_overlap suite (it needs EP=4 and TE >= 2.3.0) and would be relying on CI to validate. Would you like me to go ahead with that approach, or would you prefer to handle the test change on your side?

huthvincent:
Thanks for the patience here, and sorry for the long gap on my side.

Re-checked against `main` today: the line is still `with context_manager and outer_fp8_context:` (now `combined_1f1b.py:454`), and `fsdp_train_step` in `tests/unit_tests/a2a_overlap/utils.py` still calls `model(**data)` with no autocast context, so the situation is the same as in July — turning `enable_autocast=True` on for the overlap side alone would compare an autocast run against a non-autocast one and the `torch.equal` assertion would fail. The test passed before this fix precisely because line 460 was dropping autocast.

Happy to go ahead with the version I proposed: wrap the reference forward in the same autocast context when the config asks for it, so both sides are comparable, and then add an `enable_autocast=True` case. Sketch:

```python
def fsdp_train_step(model, optimizer, data, config=None):
    """One standard forward-backward-optimizer step through FSDP. Return scalar loss."""
    from megatron.core.transformer.module import float16_to_fp32

    optimizer.zero_grad()
    if config is not None and config.enable_autocast:
        with torch.autocast("cuda", dtype=config.autocast_dtype):
            loss = model(**data)
    else:
        loss = model(**data)
    loss = float16_to_fp32(loss).sum()
    loss.backward()
    optimizer.step()
    return loss.detach().clone()
```

One caveat I should repeat: I cannot run the `a2a_overlap` suite locally — it needs EP=4 and TransformerEngine >= 2.3.0, and I have a single-GPU box without TE — so I would be relying on your CI to validate the test change. If you would rather own the test yourself, that is completely fine and I will leave the PR as the one-line fix.

Either way the branch is now well behind `main`; say the word and I will rebase it. And whenever it is convenient, `codeowners-approval` is still red on this one.

wujingyue:
> approving with a small comment to enable this flag in the unit test

I assume this is still relevant.

huthvincent:
Still relevant, yes. But I don't think enabling that flag can be the regression test, and answering it turned up a prior question about my own fix that I'd rather ask than guess at.

**Why the flag alone won't do it.** `fsdp_train_step` (`tests/unit_tests/a2a_overlap/utils.py:369`) takes no `config` and its forward is a bare `loss = model(**data)`, so the reference side can't enter autocast by any config route, while the comparison is bit-exact (`torch.equal` at `test_fsdp_1f1b_overlap.py:175` and `utils.py:387`). And under the bug the flag changes no number on the overlap side either: the autocast block that *is* entered, `with context_manager:` at `combined_1f1b.py:380`, wraps only `build_schedule_plan`, which constructs the plan and runs no tensor math. So that test stays green with the bug and could only move once the fix is in — the wrong way round for a regression test.

**The prior question.** With the tuple form, line 454 puts `TransformerModelChunkSchedulePlan.run()` under `torch.autocast`, and `run()` executes backward (`post_process.backward`, `pre_dispatch_computation.backward_dw`, `pre_process.backward`); at the third call site (`combined_1f1b.py:121`, `f_model=None`) it's backward-only. The non-overlap path deliberately doesn't do that — `schedules.py:513` wraps only `forward_step_func`, `forward_step_calc_loss` is outside it, and `backward_step` has no context — and torch's own docs say *"autocast should wrap only the forward pass(es)… Backward passes under autocast are not recommended."* I can't wrap forward only here, because this schedule interleaves them by design (`run()`'s docstring at `model_chunk_schedule_plan.py:518`: *"Phase 1: forward_layer[0] + backward_layer[3]"*).

So which do you want?

1. What this PR does — honour the comment, autocast over the whole combined region including backward.
2. `with outer_fp8_context:` and drop *"autocast context and"* from the comment — i.e. this schedule
   deliberately doesn't autocast. Then `enable_autocast` silently does nothing on this path, which probably
   deserves a validation error rather than silence.

Either way the line is a defect as it stands: `and` returns its right operand and both operands are always truthy (`torch.autocast` defines neither `__bool__` nor `__len__`), so `with` only ever sees `outer_fp8_context`. Happy to push whichever you prefer.

Once it's settled the test is easy and needs no numerics — assert `torch.is_autocast_enabled("cuda")` inside `run()`, at the call site that carries a forward. I'd rather write it against the semantics you want than against my guess.

One caveat I should be straight about: **I can't run this suite.** No transformer_engine here, and `test_fsdp_1f1b_overlap.py` is `flaky_in_dev` and needs ≥ 4 GPUs, so it skips on my box rather than telling me anything. Everything above is read out of the source.

YangFei1990:
@huthvincent Please stop using pure agent generated responses, which makes it very hard to communicate.

The request is very simple, we want to have a unit test to execute this code path, as the original syntax error should be able to be captured very easily.

megatron/core/pipeline_parallel/combined_1f1b.py
huthvincent: Added — `tests/unit_tests/pipeline_parallel/test_combined_1f1b_autocast.py`. It fails on `main` and passes with this change, with a control for `enable_autocast=False`.

YangFei1990 (approved):

jaredcasper (approved):

Phlip79 (approved):

yaoyu-33 (approved):

Issue #5763, opened 2026-07-12
[Bug] validate_args() eagerly probes the CUDA device and crashes GPU-less launcher/driver processes (Found no NVIDIA driver)


littlemex:
**Describe the bug**

`megatron/training/arguments.py::validate_args()` — pure argument validation — eagerly queries the CUDA device in three places that only MoE + tensor/context parallelism reach. When `validate_args` runs in a process that does not (yet) own a GPU, these probes raise `RuntimeError: Found no NVIDIA driver on your system` during argument parsing, before any training starts.

Under `torchrun` this is masked: `validate_args` runs inside `initialize_megatron` in a process that is about to become a GPU rank (`torch.cuda.set_device(args.local_rank)` follows), so it happens to have a device. It breaks when arguments are validated in a **GPU-less** process — e.g. a launcher/driver that parses and validates args before any device is assigned. This is common in Ray-based orchestration, where a CPU-only driver validates args once and the GPU worker actors consume the already-validated args; there the job dies at argument validation with a confusing "no driver" error.

The three probes (current `main`):

1. `--moe-grouped-gemm` compute-capability check — [arguments.py L1296-1298](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/arguments.py#L1296-L1298): `dc = torch.cuda.get_device_capability(); assert dc[0] >= 8`.
2. TP/CP>1 `CUDA_DEVICE_MAX_CONNECTIONS` note — [arguments.py L1371-1372](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/arguments.py#L1371-L1372): `if (TP>1 or CP>1) and get_device_arch_version() < 10:`.
3. FSDP arch check — [arguments.py L1402](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/arguments.py#L1402): `if args.use_torch_fsdp2 or args.use_megatron_fsdp and get_device_arch_version() >= 10:`.

`get_device_arch_version()` ([common_utils.py L478-480](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/utils/common_utils.py#L478-L480)) does `torch.cuda.get_device_properties(torch.device("cuda:0")).major`, so it hard-requires a device.

**Steps/Code to reproduce bug**

On a GPU-less process (`torch.cuda.is_available() == False`), with Megatron importable — these are the exact calls `validate_args` makes for a MoE + TP/CP>1 config:

```python
import torch
[...]

PR #5764, opened 2026-07-12
fix: guard CUDA device probes in validate_args for GPU-less processes

littlemex:
# What does this PR do?

Guards the three eager CUDA device probes in `validate_args()` so it no longer crashes when run on a GPU-less process. `validate_args` queries the device in three MoE + TP/CP>1 paths (`torch.cuda.get_device_capability()` for `--moe-grouped-gemm`; `get_device_arch_version()` twice for the `CUDA_DEVICE_MAX_CONNECTIONS` / FSDP-arch notes); on a process with no GPU these raise `RuntimeError: Found no NVIDIA driver` during pure argument validation. This is masked under `torchrun` (the validating rank already owns a GPU) but breaks any setup that validates args in a GPU-less launcher/driver before device assignment (e.g. a Ray driver that validates once; GPU worker actors only `set_args` and never re-validate).

## Issue tracking

Linked issue: https://github.com/NVIDIA/Megatron-LM/issues/5763

## The change

`megatron/training/arguments.py` only. Gate each probe on `torch.cuda.is_available()`, with the feature flag checked **first** so `is_available()` is only reached for configs that actually probe the device:

```python
    if args.moe_grouped_gemm and torch.cuda.is_available():   # (1) was: if args.moe_grouped_gemm:
        dc = torch.cuda.get_device_capability()
        assert dc[0] >= 8, "Unsupported compute capability for GroupedGEMM kernels."

    # (2) when GPU-less, skip the arch/CUDA_DEVICE_MAX_CONNECTIONS check with a warning
    if (args.tensor_model_parallel_size > 1 or args.context_parallel_size > 1) \
        and not torch.cuda.is_available():
        warnings.warn(..., stacklevel=2)                      # see honesty note below
    elif (args.tensor_model_parallel_size > 1 or args.context_parallel_size > 1) \
        and get_device_arch_version() < 10:
        ...  # unchanged

    # (3) preserve the original precedence, made explicit: fsdp2 OR (mfsdp AND arch>=10)
    if args.use_torch_fsdp2 or (
        args.use_megatron_fsdp and torch.cuda.is_available() and get_device_arch_version() >= 10
    ):
        ...
```

Three deliberate choices:
- **Lazy `is_available()`, feature-flag first.** I do *not* hoist `torch.cuda.is_available()` to the top of the function: that would call it on every `validate_args()` — including TP=1 / no-MoE / no-FSDP configs that never touched the device before — initializing the CUDA runtime and making the process fork-unsafe. Placing the flag first means `is_available()` is called only for the exact configs that already probed the device, so the common path's CUDA-init timing is unchanged.
- **Warn on every GPU-less skip** (all three probes): each probe emits a `warnings.warn` when it is skipped for lack of a device, so the skip is never silent. Probe 2's message is **conditioned on the FSDP/overlap flags** (all plain args, knowable without a device) to mirror the device-side guidance in the `elif` branch: for Torch-FSDP2/Megatron-FSDP it says `CUDA_DEVICE_MAX_CONNECTIONS` should **not** be 1, for `overlap_moe_expert_parallel_comm` it says 1-or-larger, and otherwise `=1` — rather than blindly recommending `=1`, which would contradict upstream's own advice for FSDP configs.
- **Preserve precedence** (probe 3): the original `A or B and C` parses as `A or (B and C)`; I keep that exact semantics with explicit parentheses while inserting the guard, rather than threading a new condition through an ambiguous boolean.

**Honest scope (important).** These checks are *skipped*, not truly *deferred*, when validation runs device-less. They are re-run only if a process that owns the device also calls `validate_args`. Under `torchrun` that always happens (every rank validates). Under a validate-once launcher (e.g. a Ray driver whose GPU workers only `set_args`) they are enforced nowhere, so all three skip-warnings are the operator's only signal to configure the GPU workers out of band. A fuller fix would extract these three device-dependent checks into a `validate_device_args(args)` that GPU ranks re-run after device init; I kept this PR to the minimal crash fix and am happy to do that extraction if maintainers prefer.

Why this layer: these are eager device queries inside pure argument validation, which should not require a live GPU. Guarding the probes (rather than removing the checks) keeps all GPU-process behavior byte-for-byte identical and only changes the GPU-less case from a hard crash to validate-with-warning.

Alternatives considered:
- **Make `get_device_arch_version()` GPU-less-safe** (return a sentinel when `not is_available()`): changes a shared utility's contract and forces callers to reason about a magic sentinel; the `get_device_capability()` probe still needs its own guard, so it does not centralize the fix (only two callers, both in `validate_args`). Rejected.
- **Defer the whole block to device-init** (move these checks into `initialize_megatron` after `set_device`): cleaner but a larger refactor with more blast radius, and it moves a user-facing validation error later in startup. Out of scope; the guard is compatible with such a refactor later.
- **Require callers to always validate on a GPU process**: rejected — `validate_args` is pure argument validation with no documented contract requiring a GPU rank; requiring one rules out reasonable launcher/driver designs.

### Pre-checks

- [ ] I have added relevant unit tests — N/A: the repo has no unit test that invokes `megatron.training.arguments.validate_args` (it has 100+ prior asserts and assumes a torchrun-style rank), so a focused unit test is not the established pattern for this function. Verified instead by the functional check below.
- [x] I have added relevant functional tests — verified on hardware (see below).
- [x] I have added proper typing to my code — the change adds no new typed surface (a local bool and two guarded conditions); matches the surrounding style.
- [x] I have added relevant documentation — inline comments explain the non-obvious *why* (deferral, precedence); no external docs describe these probes.
- [x] I have run the [autoformatter](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) — note `tools/autoformat.sh` targets `megatron/core` and `tests/`; this change is in `megatron/training`, which the formatter does not cover. The added lines match the surrounding style (the existing `if (...) \` line-continuation in this same block).

Verification (deterministic; the probes are pure device queries):

- **GPU-less process** (`is_available() == False`): before the change, `get_device_capability()` and `get_device_arch_version()` raise `RuntimeError: Found no NVIDIA driver`. After: probe 1 is skipped; probe 2 emits the `warnings.warn` and defers; probe 3 preserves precedence (`use_torch_fsdp2=True` still enters the branch). No crash.
- **GPU process** (H200 → `get_device_capability() == (9, 0)`, `get_device_arch_version() == 9`): all three conditions evaluate exactly as before — the compute-capability assert runs and the `arch < 10` branch is entered, no warning emitted. GPU behavior is unchanged.
- **End-to-end**: hit originally from a Ray-based MoE run (Qwen3-30B-A3B, TP=2/CP=2) whose GPU-less driver validated args; with the guard, `validate_args` completes on the driver and the run proceeds to a full training step on the GPU workers.

Blast radius: low. On any GPU-owning process the three conditions are unchanged (the added `is_available()` is `True`); only the GPU-less case changes, from crash to validate-and-defer. No public API change; the `torchrun` path is unaffected; dense / single-GPU configs never reach these paths.

## Code review

Happy to adjust. The change is small and scoped to one function in `megatron/training/arguments.py`.


megatron/training/arguments.py
```diff
@@ -1369,6 +1381,29 @@ def validate_args(args, defaults={}):
     # disable async_tensor_model_parallel_allreduce when
     # model parallel memory optimization is enabled
     if (args.tensor_model_parallel_size > 1 or args.context_parallel_size > 1) \
+        and not torch.cuda.is_available():
```
cspades: Could just not warn at all if no CUDA: https://github.com/NVIDIA/Megatron-LM/issues/5763#issuecomment-4960925268

I think? Trying to think of a situation where a user would want to know this even when CUDA isn't even initialized.

TL;DR
```
if torch.cuda.is_available() and get_device_arch_version() < 10: # warn
```

littlemex:
### FSDP high-priority-stream arch gate: intentionally left unchanged

I noticed the pre-existing precedence issue in this block:

```python
if args.use_torch_fsdp2 or args.use_megatron_fsdp and get_device_arch_version() >= 10:
```

`and` binds tighter than `or`, so the `>= 10` gate applies only to Megatron-FSDP, not Torch-FSDP2 (dates back to `bc804914d`).

**Out of scope here — I did not change this logic.** It's already handled by the reviewer-approved PR #5730; touching it would just conflict. It's also benign: `high_priority_stream_groups` only sets an NCCL scheduling-priority flag, with no effect on correctness.

My change to this block is limited to the crash fix — a `torch.cuda.is_available()` guard on the `use_megatron_fsdp` side. Precedence is left to #5730.

cspades:
/ok to test [423bb73](https://github.com/NVIDIA/Megatron-LM/pull/5764/commits/423bb73c2efc5c79a9c0a2c5d551d32729a6bd09)

PR #5794, opened 2026-07-14
Unset NCCL overrides for MFSDP v2 tests

wujingyue:
## Summary
- Leave the unit-test runner NCCL memory-saving overrides unchanged for the general CI path.
- Clear `NCCL_MAX_NCHANNELS` and `NCCL_NVLS_ENABLE` in the MFSDP v2 `distributed_setup` fixture before tests initialize `init_device_mesh` / NCCL communicators. Would have caught #5770 earlier.

## Why
`test_symmetric_memory.py` and `test_fully_shard.py` include tests that are sensitive to NCCL algorithm and channel choices. Clearing the CI launcher overrides in the MFSDP v2 fixture makes this bucket exercise NCCL default behavior, which better matches realistic runs.

## Testing
- `bash -n tests/unit_tests/run_ci_test.sh`
- `python -m py_compile tests/unit_tests/distributed/mfsdp_v2/conftest.py`
- `git diff --check`
- `env NCCL_MAX_NCHANNELS=1 NCCL_NVLS_ENABLE=0 uv run --no-sync python -m torch.distributed.run --standalone --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/mfsdp_v2/test_context.py --tb=short` (`3 passed`)
- Verified with `NCCL_DEBUG=INFO` that the MFSDP v2 fixture path clears both vars before communicator initialization: control run used `1 coll channels`; fixture/pytest path used `4 coll channels` and did not report the env overrides.
- CI requested; expected known failure: `test_overlaps_all_gather_and_compute`.

### DIFF
--- a/tests/unit_tests/distributed/mfsdp_v2/conftest.py
+++ b/tests/unit_tests/distributed/mfsdp_v2/conftest.py
@@ -21,6 +21,12 @@ class DistributedSetup:
 @pytest.fixture(scope="function")
 def distributed_setup() -> Iterator[DistributedSetup]:
     """Read torchrun rank state and set up this rank's local device."""
+    # Some MFSDP v2 tests are sensitive to NCCL algorithm/channel choices. Clear
+    # CI launcher overrides before init_device_mesh initializes NCCL communicators
+    # so this bucket uses NCCL settings closer to production.
+    os.environ.pop("NCCL_MAX_NCHANNELS", None)
+    os.environ.pop("NCCL_NVLS_ENABLE", None)
+
     if "RANK" not in os.environ or "WORLD_SIZE" not in os.environ:
         pytest.skip("Not running under torchrun. Use torchrun to run this test file.")
 


tests/unit_tests/run_ci_test.sh
```diff
@@ -145,9 +145,17 @@ DISTRIBUTED_ARGS=(
...
+# (notably test_symmetric_memory.py and test_fully_shard.py) are sensitive to
+# NCCL algorithm/channel choices, so clear these overrides for that bucket and
+# exercise NCCL's default behavior instead.
+if [[ "$BUCKET" == tests/unit_tests/distributed/mfsdp_v2/* ]]; then
+    unset NCCL_MAX_NCHANNELS
+    unset NCCL_NVLS_ENABLE
+else
+    export NCCL_MAX_NCHANNELS=1
+    export NCCL_NVLS_ENABLE=0
+fi
```
ko3n1g: any option to move this down to the conftest level?

cspades: FYI @wujingyue these ENV variables fixed NCCL hangs likely due to SM contention / overuse for FSDP V1 (among other algorithms in MCore). If you are confident that V2 will never experience this contention, then this is fine.

Agree with @ko3n1g to set this at the Python conftest level. We shouldn't be programming test bucket logic into the main CI launch script.

wujingyue: > any option to move this down to the conftest level?

Done. I got the wrong impression that fixtures are too late for NCCL configs -- not really.

ko3n1g (approved):
if you'd like to fully remove it from the shell (and add it to the top-level conftest.py), that'd be awesome. approving nevertheless

tests/unit_tests/run_ci_test.sh
wujingyue: > these ENV variables fixed NCCL hangs likely due to SM contention / overuse for FSDP V1 (among other algorithms in MCore). 

Good to know, thanks @cspades for the context! If you happen to have a pointer to the NCCL hang, I'd love to add it as a code comment so we don't lose the context over time.

That said, I'm happy to use whatever production uses. My main goal is for the NCCL settings in unit tests to match production as closely as possible. I don't think we use
```sh
export NCCL_MAX_NCHANNELS=1
export NCCL_NVLS_ENABLE=0
```
in production, since they would slow down communication quite a bit.

cspades: I don't have any logging messages unfortunately, basically how I discovered this was that I experienced a hang when running `torchrun -m pytest` that I don't experience with CI/CD, and I checked the `run_ci_test.sh` for differences between how the CI runs tests vs. my own CLI command, and these 2 ENV variables fixed it.

The SM contention reasoning is from looking up what these two variables do. They were apparently added a long time ago.

wujingyue:
> if you'd like to fully remove it from the shell (and add it to the top-level conftest.py), that'd be awesome. approving nevertheless

Thanks for https://github.com/NVIDIA/Megatron-LM/pull/5826! I realized I misread this comment -- it's about top-level conftest.py not mfsdp_v2/conftest.py.

Issue #4468, opened 2026-04-24
DeepSeek-V4 training support


sbhavani:
> This issue tracks DeepSeek-V4 training support in Megatron Core.

DeepSeek-V4 extends DeepSeek-V3/V3.2 with hybrid compressed attention, mHC, updated routing, Muon-based training recipes, FP4 QAT, and million-token context training support. DeepSeek-V4 also includes DeepSeek-V4-Flash-Vision-Exp, an experimental multimodal variant that extends the V4-Flash architecture with visual modules and visual-understanding capabilities.

**Overall status:** Core model architecture, packed-sequence training, context parallelism, Muon, and the primary fusion paths are available in `dev`. Current work is focused on CUDA Graph coverage, long-context and memory optimization, low-precision modules, and other performance optimizations. The newly released DeepSeek-V4-Flash-Vision-Exp is not yet supported in Megatron Core.

**Status:** ✅ available / merged / resolved · 🚧 in progress / open / draft · 📋 planned / pending validation

> Status represents end-to-end capability readiness. A merged supporting PR does not necessarily mean the entire capability is complete.

Last updated: 2026-09-18

### Core functionality

| Capability | Status | Summary |
|---|:---:|---|
| Model architecture | ✅ | Hybrid CSA/HCA attention, hash routing, ClampedSwiGLU, mHC, MTP, and HybridModel support |
| Packed sequence / THD | ✅ | End-to-end variable-length training and HybridEP/DeepEP dispatcher support |
| Context parallelism | ✅ | DSv4 hybrid-attention context-parallel support |
| Muon optimizer | ✅ | Muon/AdamW routing, distributed parameter layouts, and FP8 primary-weight gather |
| Multimodal / vision | 🚧 | DeepSeek-V4-Flash-Vision-Exp has been released; Megatron Core training support is not yet available |
| Recipes and examples | 🚧 | Existing recipes and examples are available; broader platform validation remains active |

### Optimization work

| Area | Status | Current focus |
|---|:---:|---|
| Long-context optimization | 🚧 | Communication overlap, load balancing, CP layouts, and packed-CP efficiency |
| Memory optimization | 🚧 | Optimizer-state offload, module-level memory optimization, and long-context memory optimization |
| Low-precision modules | 🚧 | BF16/MXFP8 BatchedGEMM, FP8 indexer integration, and FP4 QAT |
| CUDA Graphs | 🚧 | Whole-block THD capture |
[...]

PR #5799, opened 2026-07-14
Guard MLA YARN RoPE against cos/sin dim mismatch

CodersAcademy006:
# What does this PR do ?

Assert that `cos`/`sin` have `emb_dim` as their last dimension in the fused MLA
YARN RoPE forward passes, so a mismatched rotary cache fails loudly instead of
reading out of bounds.

## Motivation

The Q and KV Triton kernels in `fused_mla_yarn_rope_apply.py` index `COS`/`SIN`
with a stride of `emb_dim` (`tl.load(COS + token_idx * emb_dim + ...)`), but
`forward()` never checked that the supplied `cos`/`sin` actually have `emb_dim`
as their last dimension. When a caller passes a shorter `cos`/`sin` — e.g. a
rotary cache whose width was misconfigured upstream — the kernel reads past the
end of the buffer and returns arbitrary garbage, which surfaces as
nondeterministic NaNs far downstream and is expensive to localize.

This was hit in practice during DeepSeek-V4 SFT with `apply_rope_fusion=True`
(nondeterministic NaN at iteration 2). The addition of a one-line shape guard
was suggested in the DeepSeek-V4 tracking issue.

Related: #4468

## Change

In both `ApplyMLARotaryEmbQ.forward` and `ApplyMLARotaryEmbKV.forward`, next to
the existing `cos.is_contiguous()` / `sin.is_contiguous()` asserts:

```python
assert cos.shape[-1] == emb_dim
assert sin.shape[-1] == emb_dim
```

The guard runs before the kernel launch, so the misconfiguration raises an
`AssertionError` at the call site instead of producing OOB reads.

## Pre-checks

### DIFF
--- a/megatron/core/fusions/fused_mla_yarn_rope_apply.py
+++ b/megatron/core/fusions/fused_mla_yarn_rope_apply.py
@@ -251,6 +251,8 @@ class ApplyMLARotaryEmbQ(torch.autograd.Function):
         assert q.stride(-1) == 1
         assert cos.is_contiguous()
         assert sin.is_contiguous()
+        assert cos.shape[-1] == emb_dim
+        assert sin.shape[-1] == emb_dim
         assert headdim == qk_head_dim + emb_dim
         assert emb_dim % 4 == 0
 
@@ -628,6 +630,8 @@ class ApplyMLARotaryEmbKV(torch.autograd.Function):
         assert k_pos_emb.stride(-1) == 1
         assert cos.is_contiguous()
         assert sin.is_contiguous()
+        assert cos.shape[-1] == emb_dim
+        assert sin.shape[-1] == emb_dim
         assert emb_dim % 4 == 0
 
         o_key = kv.new_empty(total_seqlen, nheads, emb_dim + k_dim)


CodersAcademy006:
This adds the `cos.shape[-1] == emb_dim` guard suggested in #4468 to turn the out-of-bounds cos/sin read in the fused MLA YARN-RoPE kernels into a loud shape error at the call site, rather than nondeterministic NaNs surfacing far downstream.

CodersAcademy006:
cc @sbhavani (author of #4468) — this implements the one-line `cos.shape[-1] == emb_dim` guard @Meirtz suggested there for `fused_mla_rope`.

Phlip79:
This looks like a bug fix. Were you able to reproduce the bug yourself? Did you create an issue for the bug?

sbhavani:
Related: https://github.com/NVIDIA/Megatron-LM/issues/4468#issuecomment-4669454722

CodersAcademy006:
No, I didn't personally reproduce this. The underlying OOB read is already reproduced and root-caused in #4468, see @Meirtz's comment: https://github.com/NVIDIA/Megatron-LM/issues/4468#issuecomment-4669454722 (H100 and GB300, via verl and Megatron-Bridge). This PR implements exactly the one-line guard Meirtz proposed there (`fused_mla_rope_inplace` could assert `cos.shape[-1] == emb_dim`). The actual misconfig root cause was on the Megatron-Bridge side, already fixed in NVIDIA-NeMo/Megatron-Bridge#4271; this guard just turns the OOB read into a loud assert here instead of a silent downstream NaN.

CodersAcademy006:
@sbhavani this carries the `Final Review` label and has been quiet since 27 July, so flagging it rather than letting it age out. Phlip79's question was answered then: I did not reproduce it myself, the out-of-bounds cos/sin read is already reproduced and root-caused in #4468, by Meirtz on H100 and again on GB300, and this is the one-line `cos.shape[-1] == emb_dim` guard suggested there, which turns a silent OOB read in the fused MLA YARN-RoPE path into a loud shape error at the call site.

It has never had CI run on it, since copy-pr-bot has been the only other voice on the thread since 14 July. If it needs a `/ok to test` to move, I would appreciate one, and if you would rather the guard live inside the kernel wrapper than at the call site I will move it.

PR #5808, opened 2026-07-14
Fix MegatronFSDP root module hook dispatch

### LINKED ISSUE #5789: [BUG] `MegatronFSDP.forward` calls `self.module.forward(...)` directly, bypassing the wrapped module's registered forward hooks
**Describe the bug**

Package: `megatron-fsdp` (PyPI). Affected version: 0.5.0 (latest published release at time of writing). Upstream source: NVIDIA/Megatron-LM, `megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py`. Symbol: `MegatronFSDP.forward`. Location: `megatron_fsdp.py` (near line 1459 in the installed 0.5.0 wheel).

There are wrong results or crashes for the root FSDP unit's own (shallow / root-owned) parameters: they are read from storage that FSDP freed after initialization, because the root parameter-gather hook never runs. Child FSDP units gather normally, so the failure is localized to root-owned parameters and can look like selective garbage/NaNs rather than a total failure.

The wrapper invokes the wrapped module's `forward` method directly:

```python
output = self.module.forward(*inputs, **kwargs)
```

Calling `.forward(...)` bypasses `nn.Module.__call__`, which is what dispatches registered `forward_pre_hooks` / `forward_hooks`. Megatron-FSDP registers a pre-forward hook on the root module to gather its parameters before user code runs; skipping `__call__` means that hook never fires, so root-owned parameters remain detached from / pointing at storage that FSDP freed after init.

- Wrap a module that holds parameters directly on the root (not only inside child submodules that become their own FSDP units) with `MegatronFSDP`.
- Run a forward pass. With the direct `.forward(...)` call the root parameter-gather pre-hook is skipped and the root-owned parameters are read from freed/ungathered storage (wrong values or an error). Switching to `self.module(...)` runs the hook and the forward produces correct results.
- A tighter check: register a `forward_pre_hook` on `self.module` and assert it is invoked exactly once per `MegatronFSDP.__call__`; it is not invoked with the `.forward(...)` call and is invoked with `self.module(...)`.

**Expected behavior**

The root module's registered pre-forward hook should fire so root-owned parameters are gathered before user code runs. The fix is to invoke the wrapped module through `nn.Module.__call__` so its registered hooks run:

```python
output = self.module(*inputs, **kwargs)
```

**Additional context**

Affected version: 0.5.0 (latest published release at time of writing).

wujingyue:
## Summary

- Route `MegatronFSDP.forward` through `nn.Module.__call__` so wrapped-module forward hooks run.
- Add a regression test covering root-owned parameters under `optim_grads_params` sharding.

## Root Cause

`MegatronFSDP.forward` called `self.module.forward(...)` directly, bypassing the wrapped module’s registered forward pre-hooks. For root-owned parameters, that skipped the FSDP gather hook and left the root forward reading ungathered/freed parameter storage.

Fixes #5789

## Validation

- `python -m py_compile megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py tests/unit_tests/distributed/mfsdp_v1/test_mfsdp_fully_shard.py`
- `uv run --no-sync python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/mfsdp_v1/test_mfsdp_fully_shard.py::TestMegatronFsdpFullyShard::test_root_module_forward_uses_gathered_parameters`

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
@@ -1487,7 +1487,7 @@ class MegatronFSDP(torch.nn.Module):
         self._replace_param_with_raw_if_needed()
         with torch.autograd.profiler.record_function("CustomFSDP.forward"):
             # Call the forward pass of the wrapped module.
-            output = self.module.forward(*inputs, **kwargs)
+            output = self.module(*inputs, **kwargs)
             return output
 
 
--- a/tests/unit_tests/distributed/mfsdp_v1/test_mfsdp_fully_shard.py
+++ b/tests/unit_tests/distributed/mfsdp_v1/test_mfsdp_fully_shard.py
@@ -136,6 +136,23 @@ class ToyTransformer(torch.nn.Module):
         return x
 
 
+class RootParamModel(torch.nn.Module):
+    """Toy model with parameters owned directly by the root module."""
+
+    def __init__(self):
+        super().__init__()
+        self.weight = torch.nn.Parameter(torch.empty(DIM_SIZE, DIM_SIZE))
+        self.bias = torch.nn.Parameter(torch.empty(DIM_SIZE))
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        torch.nn.init.kaiming_uniform_(self.weight, a=math.sqrt(5))
+        torch.nn.init.zeros_(self.bias)
+
+    def forward(self, x):
+        return torch.nn.functional.linear(x, self.weight, self.bias)
+
+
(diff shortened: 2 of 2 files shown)


cspades (approved):
I think this is indeed a bug. https://github.com/NVIDIA/Megatron-LM/issues/5789#issuecomment-4972428185

Originally I tested TorchTitan and BioNeMo (HuggingFace) models with `fully_shard`. I wonder if every single root module I tested never owned any direct parameters, similar to MCore models.

cc @shjwudp

deepakn94 (approved):

Phlip79 (approved):

Issue #5643, opened 2026-07-03
Runtime error when enabling CP (context parallelism) during SFT of DeepSeek V4


gaojingwei:
I am performing full-parameter SFT on DeepSeek V4 using Megatron-Bridge framework. The training runs normally with both packed and non-packed data when context parallelism (CP) is disabled (i.e., --context-parallel-size 1).

However, when I enable CP with --context-parallel-size 2, both packed and non-packed data fail with errors.

Error logs:

1.  Non-packed data error:
```
[rank11]: File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1832, in inner
[rank11]:     result = forward_call(*args, **kwargs)
[rank11]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank11]: File "/opt/Megatron-Bridge/3rdparty/Megatron-LM/megatron/core/transformer/experimental_attention_variant/deepseek_v4_hybrid_attention.py", line 393, in forward
[rank11]:     rot_part_out = apply_rotary_pos_emb(
[rank11]:                    ^^^^^^^^^^^^^^^^^^^^^
[rank11]: File "/opt/Megatron-Bridge/3rdparty/Megatron-LM/megatron/core/models/common/embeddings/rope_utils.py", line 381, in apply_rotary_pos_emb
[rank11]:     return _apply_rotary_pos_emb_bshd(
[rank11]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank11]: File "/opt/Megatron-Bridge/3rdparty/Megatron-LM/megatron/core/models/common/embeddings/rope_utils.py", line 147, in _apply_rotary_pos_emb_bshd
[rank11]:     t = (t * cos_) + (_rotate_half(t, rotary_interleaved) * sin_)
[rank11]:          ~~^~~~~~
[rank11]: RuntimeError: The size of tensor a (848) must match the size of tensor b (424) at non-singleton dimension 0
```
2. Packed data error:
```
0: /opt/pytorch/pytorch/aten/src/ATen/native/cuda/IndexKernelUtils.cu:16: vectorized_gather_kernel: block: [1687,0,0], thread: [1,0,0] Assertion `ind >=0 && ind < ind_dim_size && "vectorized gather kernel index out of bounds"` failed.
[...]

PR #5809, opened 2026-07-14
Fix zero-loss indexer gradients for DSv4 CP

kunlunl:
## Summary

- Keep the existing DSv4 THD CP indexer-loss autograd path connected during training when `dsa_indexer_loss_coeff=0`.
- Add CP2 and CP4 regression coverage requiring every indexer parameter to receive a non-`None`, exactly zero gradient.

## Root cause

The ratio-4 THD CP path treated a zero indexer-loss coefficient as disabling the loss branch entirely. That left indexer parameters unused and their gradients as `None`, which breaks overlapped DDP gradient-ready bookkeeping. The fix preserves graph connectivity while the scaled loss and resulting gradients remain exactly zero.

This addresses the zero-loss failure reported in #5643.

### DIFF
--- a/megatron/core/transformer/experimental_attention_variant/csa.py
+++ b/megatron/core/transformer/experimental_attention_variant/csa.py
@@ -2448,9 +2448,7 @@ class CompressedSparseAttention(MegatronModule):
         # Final indices address these source rows directly. This avoids a
         # second per-sequence packed KV layout and lets torch.cat own backward.
         kv_full_thd = torch.cat((boundary_kv, kv_local, compressed_kv_rank_major), dim=0)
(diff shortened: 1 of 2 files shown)


kunlunl:
/ok to test [c664879](https://github.com/NVIDIA/Megatron-LM/pull/5809/commits/c6648795a4d6f3b9058102134269a3a6fe5c2046)

hxbai (approved):

kunlunl:
/ok to test [41dfe91](https://github.com/NVIDIA/Megatron-LM/pull/5809/commits/41dfe919f7045e0ab6e76cf7339f96f9bc9170b1)

PR #5824, opened 2026-07-15
Re-enable group filtering in RL

tdene:

# What does this PR do?

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

### DIFF
--- a/megatron/rl/agent/api.py
+++ b/megatron/rl/agent/api.py
@@ -219,11 +219,6 @@ class _GranularityConfig(NamedTuple):
         assert not (
             request.submission_granularity == "B" and request.consumption_granularity == "G"
         ), "Batch submission with group consumption is not supported."
-        assert not request.filter_groups_with_same_reward, (
-            "filter_groups_with_same_reward is not currently supported: dropped groups "
-            "are not regenerated, so non-streaming callers receive fewer groups than "
-            "requested and batch-order consumers stall on incomplete batches."
-        )
 
 
 class _SubmissionGate:
@@ -239,10 +234,10 @@ class _SubmissionGate:
         self._submission = submission
         self._release_on = RELEASE_STATE_BY_SUBMISSION[submission]
         self.capacity = capacity
-        # Observability counters, updated only on the configured submission
-        # granularity (the only path that touches the semaphore). `held`
-        # counts slots currently held; `prepare_blocked_seconds` accumulates
-        # time stage_prepare spent waiting on the semaphore.
+        # Observability counters, updated only on the configured submission granularity
+        # (the only path that touches the semaphore). `held` counts slots currently held;
+        # `prepare_blocked_seconds` accumulates time submitters
+        # (stage_prepare and stage_filter regeneration) spent waiting on the semaphore.
         self.held = 0
         self.prepare_blocked_seconds = 0.0
         self.acquire_calls = 0
@@ -288,6 +283,13 @@ class _InferredItem(NamedTuple):
     inferred_at: float = 0.0
 
 
+class _AssembledGroup(NamedTuple):
+    """One complete group flowing from assemble to filter."""
+
+    group: RolloutGroup
+    assembled_at: float = 0.0
+
+
 class _RolloutPipeline:
     """Per-call orchestrator for grouped rollout generation."""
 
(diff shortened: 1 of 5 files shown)


megatron/rl/agent/api.py
```diff
@@ -344,12 +351,41 @@ def __init__(
...
         self.assemble_queue_dwell: list[float] = []
+        self.filter_queue_dwell: list[float] = []
         self.output_queue_dwell: list[float] = []
         self.prepared_count = 0
         self.inferred_count = 0
         self.assembled_count = 0
+        self.filtered_count = 0
         self.yielded_count = 0
 
+    async def _submit_group(self, *, group_id: int, batch_id: int, index_in_batch: int) -> None:
```
lauradang: _submit_group is vague given how many queues we have.  Maybe rename to _submit_group_to_infer_queue?

megatron/rl/agent/api.py
```diff
@@ -439,34 +469,83 @@ async def stage_assemble(self) -> None:
...
+                    # never reaches: like its in-flight count, its slot carries
+                    # over to the replacement (no release here, no fresh "G"
+                    # acquire in _submit_group) and frees when the replacement
+                    # is ultimately consumed. Releasing here and re-acquiring
+                    # instead could deadlock: with the gate fully held, the
+                    # freed slot can be won by stage_prepare's FIFO-earlier
+                    # waiter, parking regeneration forever while batch-order
+                    # consumption waits on the very replacement it must yield.
+                    try:
+                        await self._regenerate_group(group)
```
lauradang: Should we have a guard for retry limit? Actually, not sure if retry limit is the solution since we are always drawing completely new prompts anyway for regeneration. But if rollouts per group = 1, then there'd be no variance, which could just continue to regenerating forever. Maybe we should just have a guard for this case: rollouts per group = 1

megatron/rl/agent/api.py
```diff
@@ -439,34 +469,83 @@ async def stage_assemble(self) -> None:
...
                             index_in_batch=first.item.index_in_batch,
-                        )
+                        ),
+                        assembled_at=time.monotonic(),
                     )
+                )
+        finally:
+            self.filter_queue.shutdown()
+
+    async def stage_filter(self) -> None:
```
lauradang: I am skeptical if we need a filter queue at all for this feature? I am thinking we can just check for the should drop condition in stage_assemble and use _submit_group to just put the item back on the infer queue. The functional necessity of having another queue is so that we can run the operations asynchronously, but the heavy work of inference is already offloaoded to the infer_queue. The filter queue is very lightweight itself (just a few conditionals), so I don't see the point of having it async.

lauradang (changes requested):

megatron/rl/agent/api.py
tdene: I ended up with a deadlock doing putting things back onto the `infer_queue`!

I will elaborate in a few hours, I'm just in the middle of something right now and wanted to quickly comment.

tdene: Okay, writeup.

You're suggesting that we put it back on the infer queue through `_submit_group`. In G/B submission modes, we don't have to worry about the semaphore at all this way, no release & reacquire, because we are still holding the exact same semaphore item.    
But in R submission mode, we've released the semaphore and have to reacquire it in `_submit_group`. This kind of release & reacquire is what always kills us. In this case, it means that `stage_assemble` is stuck awaiting `_submit_group`, which is stuck trying to acquire sufficient semaphores. Because when we entered `stage_assemble` we only had 1 spare semaphore slot, from the final rollout in the group; all the rest have already been re-acquired. So now we have to wait for K-1 semaphore slots, and we block the whole system until we obtain them.    
We don't deadlock, we just introduce a significant delay.

If instead of your suggestion, we just release & reacquire period, we have basically the same problem: `stage_assemble` has to wait to reacquire a semaphore. For G/G G/B B/B, if we don't do what you said but something worse, we actually end up with an infinite deadlock.

Ultimately the problem is that regenerating groups means there is a mode under which we will have to wait for a semaphore, no matter how we design the system. If we do not spawn a side-task, then one of our current tasks gets stalled instead, and that's really bad.

megatron/rl/agent/api.py
tdene: So this actually gets handled thoroughly by #6480. I'd like to defer addressing this until that PR, and leave this version with no limit guard.

megatron/rl/agent/api.py
tdene: Done!

megatron/rl/agent/api.py
lauradang: The new change is much better! thanks

lauradang (approved):

PR #5828, opened 2026-07-15
test(clip_grads): add regression coverage for get_grad_norm_fp32

### LINKED ISSUE #5827: Add regression test coverage for get_grad_norm_fp32 (norm_type paths + empty grads_for_norm)

`tests/unit_tests/optimizer/test_clip_grads.py` contains exactly one test,
covering only the `OptimizerConfig.grad_norm_skip_threshold` default. There
is no test coverage for `get_grad_norm_fp32` itself — none of the three
norm_type branches (`2.0`, `inf`, generic p-norm) are exercised, and there
is no regression test for the empty `grads_for_norm` crash fixed in #5530.

Add tests to `test_clip_grads.py` covering:
- `get_grad_norm_fp32([], norm_type=norm_type)` for `norm_type in [2.0, inf, 1.0]`
  returns `0.0` without raising (regression test for #5529 / #5530)
- `get_grad_norm_fp32(grads, norm_type=norm_type)` for the same three
  norm types on a small set of real gradients, checked against the
  equivalent `torch.norm`-based computation

Could test only the empty-list case since that's what #5530 fixed, but
covering the non-empty path too gives a baseline so a future change to
any branch doesn't silently break the others.

**Additional context**

Follow-up to #5529 / #5530.

Mattral:
## Summary

Follow-up to #5530. `test_clip_grads.py` had no test coverage for
`get_grad_norm_fp32` itself — this adds regression tests for all three
norm_type branches, including the empty-`grads_for_norm` case that #5530 fixed.

## Changes

**`tests/unit_tests/optimizer/test_clip_grads.py`**
- `test_empty_grads_returns_zero` (parametrized over `2.0`, `inf`, `1.0`):
  regression test for the crash fixed in #5530
- `test_nonempty_grads_matches_reference` (same parametrization): baseline
  correctness check against a plain `torch.norm` computation

## Testing

- Verified locally via linting/formatting (`black`, `isort`, `pylint`, `ruff` —
  all clean via `tools/autoformat.sh`)
- Could not execute pytest locally (no GPU available in this dev environment,
  and every test in this suite requires `torch.distributed` initialization);
  relying on CI's GPU runners for actual execution
- No changes to `megatron/core/optimizer/clip_grads.py` — test-only PR

## Checklist
- [x] Commits signed off (`git commit -s`)
- [x] Rebased on `main`
- [x] Test-only change, no production code touched

Fixes #5827


guihong-nv:
@Mattral please fix the failing tests.

Mattral:
Thanks for the ping 

I'll investigate the unit test failure (want to confirm whether it's related to this change or a pre-existing issue in the suite) and push a fix by Monday.

Mattral:
@guihong-nv . Found the root cause: the test assumed a single rank, but CI runs unit tests with `--nproc-per-node 8`. Every rank seeds identically and computes the same local grads, so the SUM-based all_reduce (`grad_stats_parallel_group=None` → WORLD group) correctly multiplies the result by `world_size`, the `inf`-norm case uses MAX instead, which is why only `[2.0]` and `[1.0]` failed. Pushed a fix that accounts for `world_size` in the test's expected value, watching CI now to confirm.

Mattral:
Hi @guihong-nv 

following up here. The fix landed last week (test now accounts for world_size in the SUM vs MAX all-reduce comparison). I notice a couple of checks are still pending 

let me know if those need an approval on your end, or if there's anything else you'd like adjusted before another look.

guihong-nv:
@Mattral may you please resolve the conflict here?

Mattral:
@guihong-nv  
Resolved, kept both: my get_grad_norm_fp32 regression tests and your TestCountZerosFp32GtpPadding class, merged cleanly with no overlap in what they cover. Ready for another look.

Issue #5790, opened 2026-07-13
[BUG] Load-bearing per-`Parameter` state (`_megatron_fsdp_model`, `_is_shared`) is stored as plain Python attributes that do not survive post-wrap parameter rebuilds


yuhezhang-ai:
**Describe the bug**

Package: `megatron-fsdp` (PyPI). Affected version: 0.5.0 (latest published release at time of writing). Upstream source: NVIDIA/Megatron-LM, `megatron/core/distributed/fsdp/src/megatron_fsdp/`. Symbols: `MegatronFSDP.__init__` and `ParamAndGradBuffer._init_optimizer_named_parameters` (stamp the attributes), `fully_shard.fully_shard_optimizer` (reads `_megatron_fsdp_model`), `MegatronFSDP._process_post_backward_gradients` / `_grad_acc` (read `_is_shared`), `ParamAndGradBuffer.update_main_grads` (reads `orig_param`). Locations (installed 0.5.0 wheel): `megatron_fsdp.py` lines ~349-351 (stamp back-ref), `param_and_grad_buffer.py` line ~2923 (stamp `orig_param`), `fully_shard.py` line ~479 (back-ref read), `megatron_fsdp.py` line ~701 (shared-param skip) and line ~619 (`_grad_acc` copy), `param_and_grad_buffer.py` line ~2946 (`orig_param` read).

The defect is a fragile per-`Parameter` state design. `MegatronFSDP.__init__` stamps three load-bearing pieces of state onto each distributed `Parameter` as ordinary Python attributes, exactly once:

```python
# megatron_fsdp.py, MegatronFSDP.__init__
self._replace_param_with_distributed_if_needed()
for param in self.module.parameters():
    setattr(param, "_megatron_fsdp_model", self)   # + __fsdp_param__, and _is_shared via _reestablish_shared_weights
```

`fully_shard_optimizer` recovers the owning model from that attribute:

```python
# fully_shard.py, fully_shard_optimizer
first_mfsdp_param = optimizer.param_groups[0][next(iter(optimizer.param_groups[0]))][0]
if not getattr(first_mfsdp_param, "_megatron_fsdp_model", None):
    raise ValueError(
        "[MegatronFSDP fully_shard_optimizer()] Could not retrieve a reference to "
        f"MegatronFSDP from the first registered Parameter: {first_mfsdp_param} ..."
    )
```

[...]

PR #5836, opened 2026-07-16
fix(fsdp): preserve parameter state across rebuilds

chengcuiping:
<html><head></head><body><p>Addresses #5790</p><h2>Summary</h2><p>Megatron-FSDP V1 owns stable raw/distributed <code inline="">Parameter</code> identities and the model/main-weight buffers that back them. A post-wrap operation such as <code inline="">load_state_dict(assign=True)</code> may replace module <code inline="">Parameter</code> objects. Re-registering Python identities alone is unsafe because reinstalling the cached MFSDP parameters can silently restore stale buffer values.</p><p>This change adds one narrow V1 reconciliation point. Newly assigned Parameters are temporary value sources. After complete validation and data-parallel group consensus, their values are copied into MFSDP-owned storage and the original MFSDP-owned Parameter identities are restored.</p><p>This is not complete <code inline="">assign=True</code> identity or property preservation. It is a value-reconciliation contract for a specific pre-forward, pre-optimizer lifecycle window.</p><h2>Supported call sequence</h2><ol><li><p>Wrap the model with Megatron-FSDP V1.</p></li><li><p>Perform same-layout <code inline="">load_state_dict(assign=True)</code> or an equivalent replacement.</p></li><li><p>Restore the final expected alias topology, such as embedding/LM-head tying.</p></li><li><p>Call <code inline="">reregister_parameters()</code> collectively on the data-parallel group.</p></li><li><p>Only then enter the first forward or call <code inline="">fully_shard_optimizer()</code>.</p></li></ol><p>Validation requires the exact FQN set and alias topology, plus matching shape, dtype, device, layout, stride, tensor subclass, DTensor mesh, and placements.</p><h2>Ownership and failure semantics</h2><ul><li><p>External Parameters are value sources only.</p></li><li><p>MFSDP-owned raw/distributed Parameter identities, maps, hooks, and authoritative buffers remain owned by MFSDP.</p></li><li><p>Validation reaches group consensus before mutation. If any rank fails validation, or ranks produce different replacement-FQN plans, all ranks fail before copy.</p></li><li><p>A validation failure guarantees zero mutation and leaves the wrapper reusable.</p></li><li><p>An unexpected error after copying begins is not rolled back. The wrapper enters a terminal closed state; later reconciliation, forward, and optimizer initialization fail closed.</p></li><li><p>The reconciliation window closes on the first forward attempt and when optimizer initialization begins, even if either attempt later raises.</p></li></ul><h2>Exact support matrix</h2>
Path | Status
-- | --
V1 optim_grads_params, DP=2/TP=1, same-layout FP32 replacement and final re-tie | Tested and supported
V1 optim_grads_params, BF16 model and FP32 main weights | Tested and supported
V1 no_shard, optim, and optim_grads reconciliation | Unsupported; tested to fail before mutation
Ordinary fully-shard operation under all four V1 strategies | Existing smoke passes; this is not reconciliation coverage
Delayed-wgrad reconciliation | Unsupported; rejection guard runtime tested
TE FP8/blockwise/transpose/high-precision special storage | Unsupported and not validated for reconciliation
HSDP helper storage | Unsupported and not validated for reconciliation
NeMo Automodel | Expected integration point only; not validated end-to-end and only applicable when the final Parameters satisfy the same-layout contract
TorchTitan | Not validated; no support claim
MFSDP V2 | Out of scope and unchanged

<h2>Local validation</h2><ul><li><p>The constant-7 DP=2/TP=1 oracle passed twice independently. For the selected inputs, each relevant logit is expected to be <code inline="">8 × 7 × 7 = 392</code>; 392 is not the sum of all logits.</p></li><li><p>The non-uniform tied-embedding plus separate-projection oracle passed twice independently, checking exact rank slices, buffer item/FQN mapping, full reconstruction, forward/loss, pre-step gradients, and the SGD update against an unwrapped reference.</p></li><li><p>The BF16-model/FP32-main-weight reconciliation oracle passed.</p></li><li><p>Full DP=2/TP=1 lifecycle suite: each rank <code inline="">24 passed, 1 deselected</code>.</p></li><li><p>Tensor-parallel detection and annotation suites: <code inline="">14 passed</code>.</p></li><li><p>Existing four-strategy fully-shard smoke: each rank <code inline="">4 passed</code>.</p></li><li><p>Four-GPU TP=2/DP=2 non-regression: every rank <code inline="">1 passed</code>.</p></li><li><p><code inline="">py_compile</code>, <code inline="">git diff --check</code>, isort, Ruff, and Black 25.1.0 using repository-compatible formatting options passed.</p></li></ul><pre><code class="language-text">python -m torch.distributed.run --standalone --nproc-per-node=2 -m pytest -q tests/unit_tests/distributed/mfsdp_v1/test_issue_5790_parameter_lifecycle.py -k 'not tp2_dp2'
python -m pytest -q tests/unit_tests/distributed/mfsdp_v1/test_mcore_tensor_parallelism_detect.py tests/unit_tests/distributed/mfsdp_v1/test_annotation.py
python -m torch.distributed.run --standalone --nproc-per-node=2 -m pytest -q tests/unit_tests/distributed/mfsdp_v1/test_mfsdp_fully_shard.py::TestMegatronFsdpFullyShard::test_fully_shard_ez
python -m torch.distributed.run --standalone --nproc-per-node=4 -m pytest -q tests/unit_tests/distributed/mfsdp_v1/test_issue_5790_parameter_lifecycle.py::test_tp2_dp2_tied_replacement_gradient_local_global_shape
</code></pre><p>Project CI has not run yet because the NVIDIA runner validation gate is still pending.</p><h2>Out of scope</h2><p>This change does not migrate optimizer state, modify MFSDP V2, or fix the separate torch-native TP <code inline="">_grad_acc</code> placement mismatch. The TP=2/DP=2 test is a non-regression guard only.


wujingyue (changes requested):
This change needs a thoughtful design doc as per https://github.com/NVIDIA/Megatron-LM/issues/5790#issuecomment-5009326371

chengcuiping:
Thanks @wujingyue — agreed. The current PR is too implementation-heavy to review before the ownership and lifecycle contract is settled. I’ll move #5836 back to draft and pause further implementation changes.

I’ll open a standalone `[Design Doc]` issue linked to #5790 and #5836. The document will treat the following as open decisions rather than assuming the current implementation is the answer:

1. Whether replacing module-installed `Parameter` objects after MFSDP wrapping should be supported at all, and within which lifecycle boundary.
2. Whether any solution belongs in V1, should be V2-first, or requires different mechanisms for the two ownership models.
3. Whether supported replacement should use a state-dict-specific automatic reset, a general explicit reconciliation API, or another holder/registry design.

It will define the ownership invariants, compare the alternatives and their failure modes, specify aliasing and authoritative-buffer semantics, and cover optimizer, collective, memory, and compatibility implications. I’ll also document NVIDIA-NeMo/Automodel#2986 as the current downstream workaround, including its validated scope and limitations, rather than treating it as proof that the upstream design is settled.

The design review will determine whether the present implementation should be retained, substantially reduced, split, or replaced. Unless you prefer another format, I’ll follow the repository’s standalone design-issue pattern and link it here before requesting another code review.

chengcuiping:
The design document is now available at #6087. I’ll keep #5836 in draft and avoid further implementation changes until we align on the design direction.

chengcuiping:
Closing this draft following the decision in #5790 not to pursue a general V1 solution. The narrower V2 assign=True issue was handled separately in #6203. Thanks for the design feedback and review.

PR #5841, opened 2026-07-16
[Dev] Support mHC selective recompute with CUDA graphs under EP a2a overlap

jingqiny-99:
### What

Hyper-connection layers carry an `n`-stream residual, so the aggregate that feeds
attention is the largest activation a layer holds. Discarding it and recomputing at
backward is worth ~20 GiB at 32 layers, but Megatron rejected the combination outright
— `explicit group replay is eager-only`, so mHC recompute and any CUDA graph were
mutually exclusive under EP a2a overlap. This admits two graph scopes and fixes the
cross-stream use-after-free that blocked interleaved pipelines.

### Why the graph scope matters

A captured backward dereferences the device addresses it observed at capture time. The
eager recompute path rebinds the logical output's storage to the recomputed storage,
which every eager consumer follows — but a replayed graph never consults the
`TensorImpl`, so rebinding cannot reach it.

- **Attention-only TE split.** The producer writes its aggregate straight into the
  graph's own static input through an arena slot. That buffer has to exist regardless,
  so the recompute costs no extra memory. `validate_output()` fails loudly if a kernel
  ignores `out=` and allocates its own result, which is otherwise silent.
- **Full-iteration capture.** The recompute runs inside the same graph, so addresses
  are consistent by construction and no slot is needed. It requires dropout to be zero:
  the recompute-time RNG rewind cannot run under stream capture.

Two invariants carry the safety, both enforced at runtime and both loud:
`MHCRecomputeArenaSlot.validate_output` asserts the producer direct-wrote the captured
address, and `TECudaGraphHelper._validate_mhc_static_hidden_inputs` asserts that static
inputs sharing an address have disjoint `[fwd, bwd]` liveness windows — turning TE's
`_reuse_graph_input_output_buffers` aliasing assumption into a post-capture check
rather than a comment.

The only configuration the split validator refuses is cross-attention, whose captured
`context` output exceeds the arity `_replay_mhc_attention_consumer` accepts; rejecting
it converts a mid-capture `RuntimeError` into a config-time message. Everything else —
precision recipes, activation offloading, delayed weight-gradient compute, MoE with or
without EP overlap, fused or unfused input layernorm, THD — is left to the guards that
already own it.

### The use-after-free

mHC post-processing ran inside the communication-stream combine node, so a recompute
subgraph was allocated on the compute stream and read from the communication stream —
a window the caching allocator cannot track. It surfaced as grad norms near 1e8 under
PP4/VPP2 and reproduced on pure upstream. Giving the post-processing its own
compute-stream schedule node removes the cross-stream read at its source; the bounded
hold ring that used to mask it is gone, along with the peak memory it cost.

The move is unconditional rather than opt-in: the communication-stream placement is the
defect, so keeping it behind a flag would only preserve a known-bad path.

### Measured — DeepSeek-V3 proxy, 32 layers, 64 x GB200

| configuration | peak | throughput |
| --- | --- | --- |
| full-iteration, no recompute | 123.6 GiB | 689 TFLOP/s/GPU |
| full-iteration, mHC recompute | 103.7 GiB | 688 TFLOP/s/GPU |
| attention split, no recompute | 119.7 GiB | 957 TFLOP/s/GPU |
| attention split, mHC recompute | 99.6 GiB | 897 TFLOP/s/GPU |

Recompute is worth 19.7-19.8 GiB for 0-6% throughput, reproduced across three
independent runs. The gap between the two graph scopes is the dispatcher rather than
the graph: HybridEP cannot be captured, so full-iteration falls back to a padded
alltoall, and matching the dispatcher closes it (688.6 vs 691.7 TFLOP/s/GPU).

### The optimizer-state reload fix

`Defer the optimizer-state reload out of a graph capture` is independent of mHC. Since
#5923 the reload forks the H2D stream inside the captured region while its join sits in
the optimizer step, which a full-iteration graph excludes — so the capture ends with a
forked stream that never rejoined. Any configuration combining full-iteration capture
with `offload_optimizer_states` fails on it, hyper connections or not. Nothing is lost
by deferring: a replayed graph runs no Python, so the early reload cannot fire on a
replay iteration either way.

### Testing

Unit coverage spans the capability guards for both graph scopes, the arena address and
direct-write contracts, the MTP layer under mHC (including that `release_state` clears
the recompute manager on the inner layer rather than the wrapper), the schedule-node
ordering, and a capture that calls the optimizer-state reload.

Two functional cases are added, both with golden values from 50-iteration
`dev`/`dgx_h100` runs, both in the `mr` and `mr-github` scopes:

- `gpt/gpt3_mcore_te_tp2_pp2_mhc_attn_cudagraph` — dense TP2/PP2 with the
  attention-only capture. It is the only case in the suite pairing `--te-rng-tracker`
  with `--recompute-granularity` at non-zero dropout, so it is what pins the RNG change
  below.
- `gpt/gpt3_mcore_te_tp2_pp2_ep4_mhc_overlap_attn_cudagraph` — TP2/PP2/DP2/EP4 with
  VPP=2 and EP a2a overlap. The dense case never builds the `mhc_post` node, so it does
  not reach the change this PR is built around; this one does, in the interleaved shape
  where the use-after-free actually bit. Its loss falls from 10.93 to 10.77 with no
  excursion, against the ~1e8 grad norms the defect produced from the first iteration.

Because this PR can shift loss curves, it needs the **`Run functional tests`** label
(`scope=mr-github`, `n_repeat=5`) rather than `Run tests`, so the golden values are
actually compared.

### Two behaviour changes that reach beyond mHC

Both are fixes rather than regressions, both can move an existing loss curve, and both
are documented next to the code as well as here.

**RNG snapshots under a graph-safe tracker.** `_get_all_rng_states` /
`_set_all_rng_states` now snapshot state *contents* instead of handing back live
generator handles: the handles made restore a no-op, so a checkpointed region could not
rewind its dropout offsets and recompute drew a different mask than its forward. These
helpers back `CheckpointFunction`, every `CheckpointWithoutOutput` user and
`_fork_rng()`, so the affected set is any `--te-rng-tracker` (or `use_cudagraphable_rng`)
run with recompute and non-zero dropout, not just mHC. The snapshot now carries an
explicit `cloned`/`live` tag, because both `clone_state()` and `graphsafe_get_state()`
return `torch.Generator` and the kind cannot be recovered by inspecting the value;
restoring an eager snapshot under capture raises rather than silently orphaning the
tracker's generators.

*Private API:* `_get_all_rng_states()` returns a 4-tuple rather than a 3-tuple and
`_set_all_rng_states()` takes a matching `kind`. Every in-repo caller splats the tuple,
but an out-of-tree fork that unpacks three values will fail at the first checkpointed
forward.

**fp32 `grad_h` in the torch h-aggregate backward.** `_torch_h_aggregate_bwd` upcasts
both operands before the product, matching the cuTile kernel and
`NativeHAggregateInto.backward`. `torch.sum` already accumulates a bf16 input in fp32,
so the cost being removed is the rounding of each `go*x` product to bf16 ahead of a
reduction spanning the whole hidden dimension. The function is shared with the
*pre-existing* `FusedHAggregate.backward`, so `--use-fused-mhc` runs on a cuTile-less
stack — including ordinary Triton-only builds — get a different, less noisy `grad_h`.

**Which existing recipes these affect: none.** The functional suite was swept for both.
Ten cases set `--te-rng-tracker` or `use_cudagraphable_rng`; exactly one also sets
`--recompute-granularity`, and that is the case added here. No case passes
`--use-fused-mhc` at all, so no golden values cover that path in either direction —
`test_torch_fallback_grad_h_upcasts_before_the_product` pins it at the unit level
instead. No pre-existing golden values needed refreshing.

### One shared-path change: the batch_p2p_sync skip

`P2PCommunicator._communicate` ends with a `torch.cuda.synchronize()` guarded by
`batch_p2p_comm and batch_p2p_sync` (both default `True`). Full-iteration capture
records the pipeline schedule, so that line is reached during capture; it is now
skipped while a stream is capturing.

This was measured on 2 GPUs rather than argued from the API contract: under capture
`torch.cuda.synchronize()` raises `operation failed due to a previous error during
capture`, while `batch_isend_irecv` and the `req.wait()` above it both capture cleanly.
So the skip is both necessary and sufficient, and no captured pipeline can carry that
workaround in any form, since a host-side device sync cannot be recorded into a graph.
Skipping is safe rather than merely convenient because the workaround targets a bug in
older PyTorch, and capturing NCCL point-to-point work requires a far newer stack than
that. The reasoning is deliberately a version argument and not a stream-ordering one:
the race it guards is cross-rank, so the skip must not be carried to a non-captured
path.


megatron/core/tensor_parallel/random.py
```diff
@@ -809,11 +904,55 @@ def discard_all_outputs(self) -> None:
...
+        not need another autograd-hook protocol change.
+        """
+        from megatron.core.transformer.mhc_recompute import MHCRecomputePhase
+
+        phase = MHCRecomputePhase(phase)
+        if self._recomputed:
+            return
+        if not self._outputs_discarded:
+            raise RuntimeError("CheckpointManager.recompute_until() requires discarded outputs.")
+        self.mhc_arena.validate_addresses()
```
buptzyb, line 943: Is `recompute_until` designed to be an mhc-specific helper? If so, should we name it `mhc_xxx`? If not, should we wrap `self.mhc_arena.validate_addresses()` in a condition?

megatron/core/tensor_parallel/random.py
```diff
@@ -822,13 +961,11 @@ def recompute_now(self) -> None:
...
-        if not self._outputs_discarded:
-            raise RuntimeError("CheckpointManager.recompute_now() requires discarded outputs.")
-        for ckpt in self.checkpoints:
-            ckpt._recompute(None)
-        self._recomputed = True
+        from megatron.core.transformer.mhc_recompute import MHCRecomputePhase
+
+        # Guards live in recompute_until; duplicating them here lets the two
+        # error paths drift apart.
+        self.recompute_until(MHCRecomputePhase.BEFORE_ATTN_BWD)
```
buptzyb, line 975: Same as above. The `CheckpointManager` involves so much MHC code without a condition. Like in this line, we make `recompute_now` always recompute until `MHCRecomputePhase.BEFORE_ATTN_BWD`. It looks like `CheckpointManager` is an MHC-specific helper and only works together with MHC.

jingqiny-99: Agree, I renamed the `CheckpointManager` to `MHCCheckpointManager`, and kept `recompute_until` unchanged.

megatron/core/transformer/transformer_layer.py
```diff
@@ -2309,6 +2468,176 @@ def _forward_post_mlp_with_fused_hyper_connection(
...
+                raise ValueError("hidden_states must be positional or keyword, not both")
+            hidden_states = args[0]
+        else:
+            hidden_states = kwargs.pop("hidden_states")
+
+        mhc_recompute_manager = getattr(self, '_mhc_recompute_manager', None)
+        output_slot = None
+        if mhc_recompute_manager is not None:
+            static_hidden_input = self.get_te_cuda_graph_static_hidden_input()
+            output_slot = mhc_recompute_manager.mhc_arena.bind_external_slot(
```
buptzyb, line 2498: In fact, with attention cudagraph, we cannot actually release the mhc output tensor even with mhc recompute, because we should leave one copy as the graph static input buffer. What we do here is to let the checkpointed mhc directly write to the static buffer so that we can **avoid a memcpy**, and this is the only benefit. Am I understanding it correctly?

jingqiny-99: Yes on the memcpy, and yes we can't release it — but the memcpy isn't the only thing we get out of it.

The reason we need this at all is that `CheckpointWithoutOutput` doesn't copy on recompute, it rebinds: `resize_(0)` the output, then `share_storage` the tensor onto the newly recomputed allocation, so the data_ptr moves from one address to another address, without moving data. That's fine under `full_iteration` capture since producer, recompute, and consumer all live in the same captured pool. But with a partial graph it isn't. 

So we either copy back into the static buffer after recompute, or write into it directly. Copy-back costs the D2D plus one extra live `[s, b, C]` during recompute (recomputed output + static buffer). Direct-write has no. 

I think that the partial graph scope is a memory/time tradeoff either way, and direct-write is what lets mhc recompute join that tradeoff at all, and it improves both sides a little.

buptzyb: > Copy-back costs the D2D plus one extra live [s, b, C] during recompute (recomputed output + static buffer)

Yes... But nit I think the two live [s, b, C] won't affect peak memory because the peak allocation time has passed  when we reach this recomputation, right?

jingqiny-99: I'm not sure about this, but in my opinion, the peak should be when the last block of a chunk is doing its recompute? The overhead is roughly `recompute_layer_num × [s, b, C]`, which is indeed not much.

But I think what we're really discussing is whether this bit of gain is worth the cost of writing this dirty code. I think even without `direct-write`, we'd still have to write the copy-back ourselves. AI estimated that changing to `copy` can save about 130 lines. 

If you feel this implementation is too dirty, I'm fine with changing it to `copy`.

buptzyb: Maybe in the future we can think of a general arena class for cudagraph input buffer direct-write. For now, I'm good with it.

buptzyb (approved):

megatron/core/transformer/transformer_layer.py
```diff
@@ -2009,6 +2067,107 @@ def forward(self, *args, **kwargs):
...
+            return (attention_output,)
+        return attention_output, output_bias
+
+    def _te_cuda_graph_capture(self, *args, **kwargs):
+        """Capture only the attention consumer for the split mHC path."""
+        if not self._uses_mhc_recompute_attn_cuda_graph_split():
+            return super()._te_cuda_graph_capture(*args, **kwargs)
+
+        self._reconstruct_packed_seq_params_from_kwargs(kwargs)
+        return self._forward_mhc_attention_cuda_graph_consumer(*args, **kwargs)
```
Wohox, line 2123: **[CRITICAL Correctness] Reject attention activation offloading for the mHC attention-only graph split**

This override bypasses the parent `_te_cuda_graph_capture()` implementation and therefore drops its `off_interface.backward_record()` and `forward_record()` hooks. Those hooks provide the synchronization edges between the TE graph stream and the H2D/D2H activation-offload streams.

However, the current configuration validation still allows attention-scope offloading modules such as `qkv_linear`, `core_attn`, and `attn_proj` together with `recompute_modules=[mhc]` and `cuda_graph_modules=[attn]`. The configuration is therefore accepted even though the split capture no longer supports those modules safely, potentially allowing offload/reload to race the captured attention forward or backward.

Until the split path implements the equivalent stream synchronization, please fail closed in argument or `TransformerConfig` validation when this mHC attention-only graph split is combined with any attention-scope offloading module. Please also add a configuration test covering the rejected combination.

Wohox (approved):
LGTM, thanks for this work!
(Please check the act offloading compatibility issue before merging~

megatron/core/transformer/transformer_layer.py
jingqiny-99: Agree. I have rejected mHC attn split + activation offloading

hxbai (approved):
LGTM

PR #5843, opened 2026-07-16
refactor: extract and split common logic between GDN & GDN2

xuantengh:

# What does this PR do?

This is the first of several PRs to refactor the GDN module in order to support different variants of gated delta networks (GDN / GDN2) and different backend implementations (FLA, FlashInfer, FlashQLA, etc.). This PR extracts the parameter initialization logic into `_GDNBase` in `common.py` so that any GDN variant can inherit this base class to avoid code duplication. Also, it also keeps the original `gated_delta_net.py` file to hold back compatibility.

This [wandb link](https://wandb.ai/nvidia/megatron-gdn2/workspace?nw=nwuserxuantengh) shows the refactor produces exactly the same output (lm loss).

#5392 will be the next step to abstract away the CP communication logic.


xuantengh:
/ok to test [c2cbf04](https://github.com/NVIDIA/Megatron-LM/pull/5843/commits/c2cbf048cc7245ba73f18ff641c0bfb77dc3e298)

xuantengh:
/ok to test [aaf8e93](https://github.com/NVIDIA/Megatron-LM/pull/5843/commits/aaf8e93a4887767240ae9c896c66afbc4d54db74)

xuantengh:
@yuzhongw-nvidia may I get your comments once you have a chance?

megatron/core/ssm/gated_delta_net/common.py
```diff
@@ -0,0 +1,871 @@
...
+        submodules: GatedDeltaNetSubmodules,
+        layer_number: int = None,
+        bias: bool = False,
+        conv_bias: bool = False,
+        conv_init: float | None = None,
+        use_qk_l2norm: bool = True,
+        A_init_range: tuple[float, float] = (1, 16),
+        pg_collection: ProcessGroupCollection = None,
+        cp_comm_type: str | None = None,
+        name: str | None = None,
```
guihong-nv, line 121: `cp_comm_type` was inserted before `name`, changing the documented public constructor order from `..., pg_collection, name` to `..., pg_collection, cp_comm_type, name`. Existing positional callers will silently lose the module name because it binds to the ignored argument, potentially disabling name-based quantization matching. Move `cp_comm_type` after `name` or make it keyword-only.

megatron/core/ssm/gated_delta_net/gdn.py
```diff
@@ -0,0 +1,278 @@
...
+
+        self.dt_bias_dim = self.num_v_heads_local_tp
+        self.a_log_dim = self.num_v_heads_local_tp
+
+        if self.config.deterministic_mode:
+            self.gated_delta_rule = torch_chunk_gated_delta_rule
+        else:
+            self.gated_delta_rule = chunk_gated_delta_rule
+
+    def forward(
```
guihong-nv, line 64: `forward()` now promises `tuple[Tensor, Tensor]`, but the default `bias=False` path returns `None` as the second element. We should use `tuple[torch.Tensor, torch.Tensor | None]`; the new forward/backward test should also assert the expected bias value.

AAnoosheh (approved):

megatron/core/models/gpt/experimental_attention_variant_module_specs.py
```diff
@@ -4,7 +4,7 @@
 
 from megatron.core.fusions.fused_bias_dropout import get_bias_dropout_add
 from megatron.core.models.backends import BackendSpecProvider
-from megatron.core.ssm.gated_delta_net import GatedDeltaNet, GatedDeltaNetSubmodules
+from megatron.core.ssm.gdn import GatedDeltaNet, GatedDeltaNetSubmodules
```
yuzhongw-nvidia: I'd suggest renaming the `gdn` folder to `gated_delta_net` while removing the `gated_delta_net.py` file. In this way, path changes like this can be avoided, and the compatibility could be better.

xuantengh:
/ok to test [5fc5bd5](https://github.com/NVIDIA/Megatron-LM/pull/5843/commits/5fc5bd52de02597c311dd8f8aa3dccaa7a32408d)

yuzhongw-nvidia (approved):
LGTM, thanks.

Phlip79 (approved):

megatron/core/ssm/gated_delta_net/common.py
```diff
@@ -951,17 +772,17 @@ def tensor_a2a_hp2cp(
 # Torch native gated delta rule
 ####################
 def torch_chunk_gated_delta_rule(
-    query,
-    key,
-    value,
+    q,
```
jiemingz, line 802: why is this change needed?

megatron/core/ssm/gated_delta_net/gdn.py
```diff
@@ -0,0 +1,278 @@
...
+            (self.qk_dim_local_tp * 2 + self.v_dim_local_tp) // self.cp_size,  # qkv
+            self.v_dim_local_tp // self.cp_size,  # gate (z)
+            self.num_value_heads // self.tp_size // self.cp_size,  # beta
+            self.num_value_heads // self.tp_size // self.cp_size,  # alpha
+        )
+
+        self.dt_bias_dim = self.num_v_heads_local_tp
+        self.a_log_dim = self.num_v_heads_local_tp
+
+        if self.config.deterministic_mode:
```
jiemingz, line 59: can we add test coverage for gdn deterministic mode?

megatron/core/ssm/gated_delta_net/common.py
xuantengh: To make the parameter name match the GDN & GDN2 forward interface from FLA:

https://github.com/fla-org/flash-linear-attention/blob/2e38c1fab332174d056928feaf29f8c5fd5ac550/fla/ops/gated_delta_rule/chunk.py#L379-L397

https://github.com/fla-org/flash-linear-attention/blob/2e38c1fab332174d056928feaf29f8c5fd5ac550/fla/ops/gdn2/chunk.py#L194-L215

We'll pass these input parameters using dict as kwargs to the unified `self.gated_delta_rule` during forward after integrating GDN2 to support dynamic argument passed for different modes.

guihong-nv (approved):

megatron/core/ssm/gated_delta_net/gdn.py
xuantengh: Added.

megatron/core/ssm/gated_delta_net/common.py
```diff
@@ -243,289 +254,90 @@ def __init__(
 
         self.reset_parameters()
 
+    def _setup_variant_attrs(self):
```
Mellonta, line 284: Why doesn't the base class own the shared query/key/value/z projection path, with each variant declaring only its additional sections?

megatron/core/ssm/gated_delta_net/common.py
jiemingz: okay, maybe that should be enforced as an interface? or documented somewhere; or else the change feels arbitrary

megatron/core/ssm/gated_delta_net/common.py
```diff
@@ -575,13 +398,18 @@ def _prepare_qkv_for_gated_delta_rule(self, qkv, gate, beta, alpha, batch, seq_l
...
-        beta = beta.contiguous()
-        alpha = alpha.contiguous()
+        gate_feats = tuple(t.contiguous() for t in gate_feats)
 
-        return query, key, value, gate, beta, alpha
+        return query, key, value, gate, *gate_feats
 
     @jit_fuser
-    def _compute_g_and_beta(self, A_log_local_cp, dt_bias_local_cp, alpha, beta):
+    def _compute_g_and_beta(
```
Mellonta, line 433: This seems to be GDN-specific, shall we move it to the subclass?

megatron/core/ssm/gated_delta_net/common.py
```diff
@@ -100,11 +110,14 @@ def __init__(
             pg_collection: The required process groups to use for tensor model parallel and context
                 parallel.
             name (str | None): module instance name passed top-down from its paranet module
+            cp_comm_type (Optional[str]): Accepted for TransformerLayer compatibility and
+                ignored; GDN implements context parallelism with its own all-to-alls rather
+                than the attention CP communication schemes.
         """
-
         if not HAVE_FLA:
```
Mellonta, line 141: How do you plan to support "different backend implementations"? Here we hardcoded FLA. Should backend selection and dependency validation be encapsulated behind a backend interface?

xuantengh: Yeah, [FlashInfer](https://bench.flashinfer.ai/docs/op-types/gdn) and [FlashQLA](https://github.com/QwenLM/FlashQLA/tree/main/flash_qla/ops/gated_delta_rule) provide CuTe DSL and TileLang GDN interfaces, respectively. We may leave this as another PR once we've decided to integrate these backends?

megatron/core/ssm/gated_delta_net/common.py
xuantengh: It'll be shared between GDN and GDN2.

megatron/core/ssm/gated_delta_net/common.py
xuantengh: Make the `in_proj_dim` into 2 parts: `in_proj_qkvg_dim` (common QKV and gate, shared across variants, set in `_GDNBase.__init__`) and `in_proj_extra_dim` (varies across variants, set in subclass `_setup_variant_attrs`).

megatron/core/ssm/gated_delta_net/common.py
xuantengh: Provides a interface protocol.

xuantengh:
/ok to test [44bc4d3](https://github.com/NVIDIA/Megatron-LM/pull/5843/commits/44bc4d394d07eaa5070fda84e6cebd7967e050b7)

Mellonta (approved):

jiemingz (approved):

xuantengh:
/ok to test [112fc4a](https://github.com/NVIDIA/Megatron-LM/pull/5843/commits/112fc4a1a8a4ae94e2fc65298e73e2d5181f95d0)

PR #5845, opened 2026-07-16
Fix silent expert-weight drop in tools/checkpoint MoE conversion

### LINKED ISSUE #5844: tools/checkpoint HF>mcore converter silently drops all MoE expert weights (Mixtral)
## Summary

The legacy checkpoint converter (`tools/checkpoint/convert.py`) silently produces a
**numerically broken** checkpoint for MoE models: every expert's `linear_fc1`/`linear_fc2`
weight is left at its (uninitialized) construction-time value instead of the loaded weights.
There is no error and no warning — `save_checkpoint` reports success — so the corruption is
only observable when the converted checkpoint is loaded and run.

Reproduced end-to-end with `--loader mixtral_hf --saver mcore` on `core_v0.18.0`; the root
cause is still present on `main` (checked at `82e9dc6`).

## Root cause

`MegatronCheckpointSaverBase.receive_lm` (`tools/checkpoint/saver_base.py`) builds the
per-layer params dict for MoE with **stacked** keys:

```python
if self.margs.num_experts:
    params_dict.update({
        "mlp_fc1_weight" : mlp_l0_weight[ep_rank][tp_rank],   # shape [num_local_experts, ffn//tp, hidden]
        "mlp_fc2_weight" : mlp_l1_weight[ep_rank][tp_rank]
    })
```

But `get_model_schema` returns `CoreMoETESchema` for any MoE model
(`tools/checkpoint/schema_core.py`), whose layer schema uses **per-expert** keys:

```python
**{f"mlp_fc1_weight.{i}" : f"mlp.experts.local_experts.{i}.linear_fc1.weight" for i in range(num_local_experts)},
**{f"mlp_fc2_weight.{i}" : f"mlp.experts.local_experts.{i}.linear_fc2.weight" for i in range(num_local_experts)},
```

`ModelSchema._set` (`tools/checkpoint/schema_base.py`) copies only the schema keys that are
present in the params dict:

```python
@classmethod
def _set(cls, schema, model, params):
    for k, m in schema.items():
        if k in params:
            cls._set_deep_tensor(model, m, params[k])
```

`"mlp_fc1_weight.0" … "mlp_fc1_weight.{N-1}"` are never in `params` (which only has the stacked
`"mlp_fc1_weight"`), so **no expert weight is ever copied**, and the stacked entry is silently
ignored. Because the saver runs with `--no-initialization`, the experts keep whatever
`torch.empty` produced. `get_model_schema` also has no grouped-GEMM branch (it hard-asserts
`transformer_impl == "transformer_engine"`), so there is no config under which the stacked key
is correct.

**Confirmed by direct weight comparison:** with the stacked key, a TP1 conversion's expert
`linear_fc1`/`linear_fc2` tensors differ from the HF source by O(1) (uninitialized memory);
with the per-expert-key fix they are bit-identical (`max|delta| = 0.0`) to `cat([w1, w3])` /
`w2` for every sampled expert/layer.

## Impact / scope

- Any MoE conversion through the `tools/checkpoint` saver (`saver_mcore` / `saver_base`) —
  observed with the `mixtral_hf` loader.
- Silent: no crash, no warning. Converted checkpoint loads fine and only reveals the problem
  at inference/finetune time (NaN or garbage loss because the expert FFNs are random).
- Not caught by CI because there is no test that converts an HF MoE checkpoint and then loads
  and runs it.

(Modern MoE import paths in Megatron-Bridge are unaffected — they use per-architecture bridges,
not this legacy saver.)

## Secondary gap in the same flow

`md.qkv_bias` is unset in `loader_mixtral_hf.py`. `saver_base.receive_lm` reads
`self.md.qkv_bias` unconditionally, but the mixtral loader's metadata block never sets it
(it sets `md.linear_bias`, `md.norm_has_bias`), so the saver dies with `AttributeError`.
`loader_base.py` sets `md.qkv_bias = margs.add_qkv_bias`; the mixtral loader should mirror it
(False for Mixtral).

(On `core_v0.18.0` the MoE model build also needed additional data-parallel / expert-tensor-parallel
fake process groups in the loader and saver. `main` has since refactored the saver's group init,
so that part may already be resolved on `main`; not included here.)

## Fix

A PR follows with two changes: unstack the expert weights by local-expert index so the keys
match `CoreMoETESchema`, plus the one-line `md.qkv_bias` fix. Longer term, a converter
round-trip test for an MoE model (convert HF -> mcore -> load -> compare a few expert tensors)
would prevent silent regressions of this class.

janbernloehr:
Fixes #5844.

## What

Two fixes to make `tools/checkpoint/convert.py --loader mixtral_hf --saver mcore` produce a
correct checkpoint for MoE (Mixtral) models:

1. **`tools/checkpoint/saver_base.py`** — emit **per-expert** MoE weight keys so they match
   `CoreMoETESchema`. Previously the saver passed a single stacked `mlp_fc1_weight` /
   `mlp_fc2_weight`, which `ModelSchema._set` never matches against the schema's
   `mlp_fc1_weight.{i}` keys, so every expert weight was silently left uninitialized.
2. **`tools/checkpoint/loader_mixtral_hf.py`** — set `md.qkv_bias` (read unconditionally by
   `saver_base.receive_lm`; the mixtral loader never set it, causing `AttributeError`).

See #5844 for the full root-cause analysis.

## Why

`get_model_schema` returns `CoreMoETESchema` for all MoE models, and that schema uses
per-expert keys (`mlp.experts.local_experts.{i}.linear_fc1.weight`). The saver's stacked key
never matches, so `_set` copies nothing and the experts keep their `--no-initialization`
values. The result is a checkpoint that loads without error but has random expert FFNs. There
is no config where the stacked key is correct (the schema selection hard-asserts TE and has no
grouped-GEMM branch).

## Testing

- **Weight equivalence (direct proof).** Converted `Mixtral-8x7B` HF -> mcore at TP=1/PP=1/EP=1
  and compared the converted tensors against the HF source: token embedding, router
  (`block_sparse_moe.gate`), and expert `linear_fc1` (= `cat([w1, w3])`) / `linear_fc2` (= `w2`)
  for experts 0 and 7 at layers 0, 15, 31. **All 16 sampled tensors are bit-identical
  (`max|delta| = 0.0`).** Before the fix these expert tensors are uninitialized garbage; after
  the fix they exactly reproduce the source, confirming correct values, expert ordering, and
  swiglu concat order.
- **End-to-end.** Converted at TP=4/PP=2/EP=1 and ran validation inference. Before the fix:
  `RuntimeError: found NaN in local forward loss` on the last pipeline stage. After the fix:
  finite, stable validation loss ~1.95 (PPL ~7.05), consistent across H100 (amd64), GB200-NVL
  and GB300-NVL (arm64), for both bf16 and fp8-hybrid.
- The `else` (dense) branch is unchanged, so non-MoE conversions are unaffected.

## Notes / scope

- Only the expert **weights** are exercised here; Mixtral has `add_bias_linear=False` so the
  expert-bias branch (also stacked) is not hit, and `CoreMoETESchema` does not currently define
  per-expert bias keys. Left as-is.
- On `core_v0.18.0` the MoE model build also needed additional data-parallel / expert-tensor-parallel
  fake process groups in the loader and saver; `main`'s saver refactor may already cover that, so
  it is not part of this PR (orthogonal to this correctness fix).

### DIFF
--- a/tools/checkpoint/loader_mixtral_hf.py
+++ b/tools/checkpoint/loader_mixtral_hf.py
@@ -264,6 +264,7 @@ def _load_checkpoint(queue, args):
     md.output_layer = margs.untie_embeddings_and_output_weights
     md.position_embedding_type = margs.position_embedding_type
     md.linear_bias = margs.add_bias_linear
+    md.qkv_bias = margs.add_qkv_bias
     md.norm_has_bias = False
     md.swiglu = margs.swiglu
     md.previous_tensor_parallel_size = margs.tensor_model_parallel_size
--- a/tools/checkpoint/saver_base.py
+++ b/tools/checkpoint/saver_base.py
@@ -471,10 +471,15 @@ class MegatronCheckpointSaverBase:
                             "mlp_norm_weight" : post_norm_weight
                         }
                         if self.margs.num_experts:
-                            params_dict.update({
-                                "mlp_fc1_weight" : mlp_l0_weight[ep_rank][tp_rank],
-                                "mlp_fc2_weight" : mlp_l1_weight[ep_rank][tp_rank]
-                            })
+                            # CoreMoETESchema registers per-expert keys
+                            # ("mlp_fc1_weight.{i}" -> mlp.experts.local_experts.{i}.linear_fc1.weight),
+                            # so the stacked expert tensor must be split by local-expert index. A
+                            # single stacked "mlp_fc1_weight" key never matches the schema, leaving
+                            # every expert weight uninitialized.
+                            num_local_experts = self.margs.num_experts // self.args.target_expert_parallel_size
+                            for expert_idx in range(num_local_experts):
+                                params_dict[f"mlp_fc1_weight.{expert_idx}"] = mlp_l0_weight[ep_rank][tp_rank][expert_idx]
+                                params_dict[f"mlp_fc2_weight.{expert_idx}"] = mlp_l1_weight[ep_rank][tp_rank][expert_idx]
                         else:
                             params_dict.update({
                                 "mlp_fc1_weight" : mlp_l0_weight[tp_rank],


janbernloehr:
Duplicate of https://github.com/NVIDIA/Megatron-LM/pull/6151

PR #5857, opened 2026-07-16
Disable grad norm check for optimizer CG

### LINKED ISSUE #5889: optimizer_cuda_graph=True crashes during graph capture when grad_norm_skip_threshold uses default float('inf')
**Environment:**
1.Megatron-LM: core_r0.18.0 && Transformer Engine: 2.14

**Summary:**
When --optimizer-cuda-graph is enabled, training crashes during the CUDA graph capture step (warmup step) with:
torch.AcceleratorError: CUDA error: operation not permitted when stream is capturing
The error occurs at the grad_norm > grad_norm_skip_threshold comparison in ChainedOptimizer.step().

**Root Cause**
Two commits interact badly:
3d87bfc1b — introduced OptimizerCudaGraphWrapper, which wraps the entire ChainedOptimizer.step() inside a torch.cuda.graph() capture context on a non-default stream.
180131620 — added the grad_norm > grad_norm_skip_threshold check inside ChainedOptimizer.step().

During CUDA graph capture, grad_norm returned by get_grad_norm_fp32() is a 0-dim CUDA tensor (not a Python float) when Transformer Engine is installed, because multi_tensor_scale_tensor_impl is not None and total_norm.pow(...) returns a tensor:

[clip_grads.py#L139-L142](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/optimizer/clip_grads.py#L139-L142)

if multi_tensor_scale_tensor_impl is not None:
    total_norm = total_norm.pow(1.0 / norm_type)   # ← 0-dim CUDA tensor
else:
    total_norm = total_norm.item() ** (1.0 / norm_type)  # ← Python float
Then at the comparison:

[optimizer.py#L1482](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/optimizer/optimizer.py#L1482)

if grad_norm > optimizer.config.grad_norm_skip_threshold:
tensor > float('inf') produces a BoolTensor, and Python's if calls .item() → cudaDeviceSynchronize() — which is not permitted while the stream is capturing.

This happens even with the default value float('inf') where the comparison is always False.

vasunvidia:

Fixes https://github.com/NVIDIA/Megatron-LM/issues/5889

# What does this PR do?

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


Phlip79 (approved):

megatron/core/optimizer/optimizer.py
```diff
@@ -1730,7 +1730,8 @@ def step(self):
 
             grad_norm_skip_threshold = optimizer.config.grad_norm_skip_threshold
             if (
-                main_params
+                not optimizer.config.optimizer_cuda_graph
+                and main_params
                 and math.isfinite(grad_norm_skip_threshold)
                 and grad_norm > grad_norm_skip_threshold
```
janEbert: This silently removes the gradient threshold check when using optimizer CG. Can't we simply convert the `grad_norm` to a float for the comparison, like the following?

```patch
                 main_params
                 and math.isfinite(grad_norm_skip_threshold)
-                and grad_norm > grad_norm_skip_threshold
+                and grad_norm.item() > grad_norm_skip_threshold
```

janEbert (changes requested):

megatron/core/optimizer/optimizer.py
RayXL97: maybe calls .item() → cudaDeviceSynchronize() — which is not permitted while the stream is capturing？

janEbert: You're right, this doesn't work. Unless we find a better option, I'd suggest that we error out if both CG and a skip threshold are active instead of silently changing behavior.

jaredcasper: Any updates here?

vasunvidia: I updated the PR to disallow skip threshold with optimizer CUDA graph.

janEbert (approved):

jaredcasper (approved):

janEbert:
@gautham-kollu I think you need to DCO-sign https://github.com/NVIDIA/Megatron-LM/pull/5857/commits/2c5c5fab369693bf85784d4b28a9974fadbf8386.

ko3n1g:
/ok to test [8aa582d](https://github.com/NVIDIA/Megatron-LM/pull/5857/commits/8aa582dda2493068d35935d76aa3470c53b93391)

ko3n1g:
/ok to test [e2cbabf](https://github.com/NVIDIA/Megatron-LM/pull/5857/commits/e2cbabffffb7993e99098fec2e9b5423c55d7beb)

janEbert:
Hey guys, I think we should stop retrying and let @gautham-kollu [fix DCO](https://github.com/NVIDIA/Megatron-LM/pull/5857#issuecomment-5661566172) before continuing.

PR #5813, opened 2026-07-14
Add initial MCore integration for MFSDP v2

wujingyue:
## Summary

Related to #5614.

This PR adds the initial MCore integration path for experimental MFSDP v2:

- expose `FullyShardedDataParallel` as the MFSDP adapter alias selected by `MFSDP_VERSION=2`
- wrap configured FSDP unit modules before the root module for MFSDP v2
- route MFSDP v2 optimizer setup through MCore `DistributedOptimizer`
- use MFSDP v2 DTensor parameter/gradient shards directly instead of V1 param-and-grad buffers
- disable the V1 startup parameter all-gather knob when `MFSDP_VERSION=2`, since V2 gathers parameters from module hooks
- filter empty local DTensor shards from optimizer param groups while TransformerEngine FusedAdam handling is fixed upstream
- fail fast for unsupported V2 configuration paths, including loss scaling, FP16, V1-only FSDP tuning options, HSDP, FP8/FP4, and unsupported optimizer features
- add distributed unit coverage against an unsharded MCore mixed-precision optimizer reference
- route the H100 unit-test recipe’s MFSDP v2 bucket through `MFSDP_VERSION=2`

## Notes

MFSDP v2 optimizer checkpointing is still intentionally unsupported in this initial path. See #5534.

The empty-shard optimizer filtering is tied to NVIDIA/TransformerEngine#3207.

## Validation

Ran a 5-iteration `pretrain_hybrid.py` DP=2 smoke with the same mock-data/model settings for no FSDP, MFSDP v1, and MFSDP v2. The MFSDP commands pass `--ckpt-format fsdp_dtensor` only because current Megatron-FSDP argument validation requires it; this smoke does not save or load checkpoints.

```bash
COMMON_ARGS=(
  pretrain_hybrid.py
  --mock-data
  --tokenizer-type NullTokenizer
  --vocab-size 1024
  --null-tokenizer-eod-id 0
  --null-tokenizer-pad-id 1
  --bf16
  --no-gradient-accumulation-fusion
  --tensor-model-parallel-size 1
  --pipeline-model-parallel-size 1
  --context-parallel-size 1
  --expert-model-parallel-size 1
  --num-layers 2
  --hidden-size 64
  --ffn-hidden-size 128
  --num-attention-heads 4
  --group-query-attention
  --num-query-groups 4
  --mamba-num-heads 8
  --mamba-head-dim 8
  --mamba-num-groups 8
  --hybrid-layer-pattern 'M*'
[...]

PR #5865, opened 2026-07-17
Add MFSDP v2 MCore integration with FullyShardedOptimizer

### LINKED ISSUE #5614: Integrate MFSDP v2 with Megatron Core training
## Goal

## Motivation

Initial models to try:

- Nemotron.

## Background

Suggested switch:

```bash
USE_MFSDP_V2=1
```

## Intended outcome

## Initial limitations

- FP8.

## Notes

wujingyue:
Closes #5614

This PR integrates MFSDP v2 with MCore using a dedicated `FullyShardedOptimizer`, which is my preferred optimizer integration for MFSDP v2.

This keeps the existing `DistributedOptimizer` focused on DDP and MFSDP v1. `DistributedOptimizer` has strong assumptions about DDP API/classes such as DistributedDataParallel and ParamAndGradBuffer. While MFSDP v1 reuses DistributedOptimizer, that reuse is shallow: most of the DDP-specific implementation is bypassed via the `if use_megatron_fsdp` branches, creating unnecessary mental burden for readers trying to understand which code paths are enabled for MFSDP and which are not.

See http://nv/mfsdp-optimizer-design for more details on the design.

## Summary

- Integrate experimental MFSDP v2 with the MCore DDP/optimizer construction path.
- Add `FullyShardedOptimizer` for MFSDP v2-owned sharded parameters and gradients.
- Route MFSDP v2 optimizer construction to `FullyShardedOptimizer` instead of `DistributedOptimizer`.
- Add MFSDP v2 optimizer support validation in `FullyShardedOptimizer._validate_config()`.
- Keep MFSDP v2 optimizer checkpointing explicitly unsupported in the new wrapper for now.
- Treat `reload_model_params()` as a no-op because model loads should write into persistent `main_weight`-backed sharded parameters.
- Select MFSDP v1 or v2 from `DistributedDataParallelConfig.megatron_fsdp_version`; `--megatron-fsdp-version {1,2}` exposes the selector in training, and no `MFSDP_VERSION` environment-variable dependency remains.

## Validation

- `UV_NO_SYNC=1 uv run isort megatron/core/optimizer/fully_sharded_optimizer.py`
- `UV_NO_SYNC=1 uv run python -m py_compile megatron/core/optimizer/fully_sharded_optimizer.py`
- `git diff --check`
- `env -u MFSDP_VERSION CUDA_VISIBLE_DEVICES=0,1 UV_NO_SYNC=1 uv run python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/mfsdp_v2/test_mcore_adapter.py` (`2 passed`)
- 5-iteration, DP=2 `pretrain_hybrid.py` mock-data smoke using `--use-megatron-fsdp --megatron-fsdp-version 2 --use-distributed-optimizer --ckpt-format fsdp_dtensor` (`MFSDP_VERSION` unset; completed successfully)

### `pretrain_hybrid.py` smoke result

Ran a 5-iteration, DP=2 mock-data smoke for no FSDP, MFSDP v1, and MFSDP v2. The MFSDP v2 run used `--megatron-fsdp-version 2`, `--use-distributed-optimizer`, and `--ckpt-format fsdp_dtensor`.

Logged rank-0 `lm loss`:

| iteration | no FSDP | MFSDP v1 | MFSDP v2 |
| ---: | ---: | ---: | ---: |
| 1 | 6.945961 | 6.945961 | 6.945961 |
| 2 | 6.946891 | 6.946891 | 6.946891 |
| 3 | 6.958145 | 6.958147 | 6.958147 |
| 4 | 6.924651 | 6.924606 | 6.924606 |
| 5 | 6.908551 | 6.908510 | 6.908510 |

No skipped or NaN iterations were reported in any run.


megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -112,7 +119,7 @@ def __init__(
         config: TransformerConfig,
         ddp_config: DistributedDataParallelConfig,
         module: torch.nn.Module,
-        fsdp_unit_modules: Optional[List[torch.nn.Module]] = None,
+        fsdp_unit_modules: Optional[List[Type[torch.nn.Module]]] = None,
```
skyw, line 124: nit: I usually use different names if it is not an object. also explicit list instead of relying on plural form. e.g. `fsap_unit_module_cls_list`. This prevents hard to spot typo like fsdp_unit_module if also used in the code, also prevent the confusion is it a module instance or supposed to use it to initialize module.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -483,6 +490,198 @@ def sync_rng_states_across_tp_group(self):
...
+        ddp_config: DistributedDataParallelConfig,
+        module: torch.nn.Module,
+        fsdp_unit_modules: Optional[List[Type[torch.nn.Module]]] = None,
+        disable_bucketing: bool = False,
+        device: Optional[torch.device] = None,
+        pg_collection: Optional[ProcessGroupCollection] = None,
+    ):
+        if not HAVE_MEGATRON_FSDP:
+            raise IMPORT_MEGATRON_FSDP_ERROR
+        if pg_collection is None:
```
skyw, line 530: I'd say at this point, new code should stop providing the fallback path and make `pg_collection` mandatory.
cc @ericharper

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -483,6 +490,198 @@ def sync_rng_states_across_tp_group(self):
...
+            pg_collection = ProcessGroupCollection.use_mpu_process_groups()
+        self._validate_config(config, ddp_config, module, pg_collection, disable_bucketing)
+
+        if has_config_logger_enabled(config):
+            log_config_to_disk(config, locals(), prefix=type(self).__name__)
+
+        self.ddp_config = ddp_config
```
skyw, line 540: Suggestion: Although common, I still advice avoid and `self.config = config` type of code. Code below (line 524) already extract argument. only need to store necessary attribute in self.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -483,6 +490,198 @@ def sync_rng_states_across_tp_group(self):
...
+
+        if fsdp_unit_modules is None:
+            fsdp_unit_modules = [TransformerLayer, MoETransformerLayer, MambaLayer]
+
+        log_single_rank(
+            logger, logging.INFO, f'Setting up DistributedDataParallel with config {ddp_config}'
```
skyw: Nit: Avoid f string in log, especially large object.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -483,6 +490,198 @@ def sync_rng_states_across_tp_group(self):
...
+            grad_comm_dtype=ddp_config.megatron_fsdp_grad_comm_dtype,
+        )
+        log_single_rank(
+            logger,
+            logging.INFO,
+            f'Setting up Megatron-FSDP MixedPrecisionPolicy with config {self.mp_policy}',
+        )
+
+        if device is None:
+            device = next(module.parameters()).device
```
skyw: Q: Does it allow device to be different than what module.parameters() is on? what is the use?

skyw:
@greptileai review with strictness 1

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -483,6 +490,198 @@ def sync_rng_states_across_tp_group(self):
...
+                    submodule,
+                    mesh=mesh,
+                    placements=placements,
+                    mixed_precision_policy=self.mp_policy,
+                )
+        fully_shard(module, mesh=mesh, placements=placements, mixed_precision_policy=self.mp_policy)
+        super().__init__(config=config, module=module)
+
+    @staticmethod
+    def _validate_config(
```
skyw, line 582: Q: What's the reason to make a staticmethod private (w/ leading _)?

megatron/core/optimizer/__init__.py
```diff
@@ -753,6 +758,9 @@ def _get_megatron_emerging_optimizer(
         eopt_name = bare_name
         use_layer_wise = True
 
+    if isinstance(model_chunks[0], FullyShardedDataParallelV2):
+        raise ValueError("MFSDP v2 with emerging optimizers is not currently validated.")
```
skyw: Should this be NotImplementedError?

tests/unit_tests/distributed/mfsdp_v2/test_mcore_adapter.py
```diff
@@ -0,0 +1,187 @@
...
+                data_parallel_sharding_strategy="optim_grads_params",
+                megatron_fsdp_main_params_dtype=torch.float32,
+                megatron_fsdp_main_grads_dtype=torch.float32,
+                fsdp_all_gather_in_start_param_sync=False,
+            ),
+            module=model,
+            fsdp_unit_modules=[TransformerLayer],
+        )
+
+        assert isinstance(wrapped.module, FsdpModule)
```
skyw, line 83: use test module assert, e.g self.assertTrue instead of debug assert in test.

tests/unit_tests/distributed/mfsdp_v2/test_mcore_adapter.py
```diff
@@ -0,0 +1,187 @@
...
+                microbatch_losses.append(loss.detach())
+            success, _, _ = optimizer.step()
+            assert success
+            losses.append(torch.stack(microbatch_losses).mean())
+
+        losses = torch.stack(losses)
+        reference_losses = torch.stack(reference_losses)
+        assert torch.isfinite(losses).all()
+        assert torch.isfinite(reference_losses).all()
+        torch.testing.assert_close(losses, reference_losses, rtol=1e-3, atol=1e-3)
```
skyw: set atol to 0 if loss is not supposed to be 0.

skyw:
> Reviews (1): Last reviewed commit: ["Apply Claude fix for PR #5865"](https://github.com/nvidia/megatron-lm/commit/189a2df58a37dfa329e3f8296cd88c99d01eabfc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46023937)

@greptileai did you post results?

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: Agreed. I didn't change this because I want MFSDP v2 <> MCore integration to be a drop-in replacement for v1.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: I go with least permission by default. public/private and static/instance are orthogonal, so being private and static shouldn't be a problem.

tests/unit_tests/distributed/mfsdp_v2/test_mcore_adapter.py
wujingyue: TestMcoreAdapter is a plain pytest test class, not a unittest.TestCase, so it has no self.assertTrue. Plain assert is pytest’s normal idiom and gives better assertion introspection.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: Done. We only use `.dtype` from `device` so I'll hard-code "cuda" when device is None.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: Done

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: Done

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: Agreed.

This one is again for compatibility. https://github.com/NVIDIA/Megatron-LM/blob/58bf14e9e68915a5e0c7d70451620e6e713c07d5/megatron/core/optimizer/__init__.py#L1081 reads it.

megatron/core/optimizer/__init__.py
wujingyue: Done

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
skyw: Valid point. maybe document somewhere?

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
skyw: Subsequent code only uses a few attributes of ddp_config. But changing it seems out of scope of this PR.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
skyw: staticmethod usually implies FullyShardedDataParallelV2._validate_config(...) is a valid use.
Remind me why it is staticmethod at first place?

skyw (approved):

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: Oh, I forgot to replace `self._validate_config` with `FullyShardedDataParallelV2._validate_config()`, which I think has been confusing you. 

An alternative would be to make it a module-level function, e.g. `def _validate_config_for_mfsdp_v2()`, but that feels unnecessarily verbose compared to keeping it attached to the class.

wujingyue: > Oh, I forgot to replace self._validate_config with FullyShardedDataParallelV2._validate_config(), which I think has been confusing you.

Done.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -483,6 +491,207 @@ def sync_rng_states_across_tp_group(self):
...
+    def __init__(
+        self,
+        config: TransformerConfig,
+        ddp_config: DistributedDataParallelConfig,
+        module: torch.nn.Module,
+        fsdp_unit_modules: Optional[List[Type[torch.nn.Module]]] = None,
+        disable_bucketing: bool = False,
+        device: Optional[torch.device] = None,
+        pg_collection: Optional[ProcessGroupCollection] = None,
+    ):
```
Phlip79, line 507: Docstring?

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -483,6 +491,207 @@ def sync_rng_states_across_tp_group(self):
...
+
+    @staticmethod
+    def _validate_config(
+        config: TransformerConfig,
+        ddp_config: DistributedDataParallelConfig,
+        module: torch.nn.Module,
+        pg_collection: ProcessGroupCollection,
+        disable_bucketing: bool,
+    ) -> None:
+        """Reject features outside the initial MFSDP v2 integration scope."""
```
Phlip79: Docstring here too?

Phlip79 (approved):

asolergi-nv (approved):

megatron/core/optimizer/__init__.py
```diff
@@ -1046,6 +1055,10 @@ def get_megatron_optimizer(
 
     log_single_rank(logger, logging.INFO, f'Setting up optimizer with config {config}')
 
+    if is_mfsdp_v2:
+        if not config.use_distributed_optimizer:
+            raise ValueError("MFSDP v2 currently requires use_distributed_optimizer=True.")
```
jaredcasper: This is confusing. So I need to set --use-distributed-optimizer, but then it doesn't use DistributedOptimizer, but uses FSDP's version of distributed optimizer??

wujingyue: Thanks -- I'll fix that to require use_distributed_optimizer=False. This line was an artifact from a previous attempt (#5813) trying to actually use DistributedOptimizer.

wujingyue: Done. I also did a minor refactor in megatron/core/optimizer/__init__.py to reduce indentation.

jaredcasper (approved):

guihong-nv (approved):

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -0,0 +1,126 @@
...
+        super().__init__(optimizer, config, grad_scaler, init_state_fn)
+        self.model_chunks = model_chunks
+        self.ddp_config = self.model_chunks[0].ddp_config
+        for model_chunk in self.model_chunks:
+            if self.ddp_config != model_chunk.ddp_config:
+                raise ValueError("All MFSDP v2 model chunks must share the same ddp_config.")
+        self.is_stub_optimizer = optimizer is None
+
+    @staticmethod
+    def _validate_config(config: OptimizerConfig, model_chunks: List[MegatronModule]) -> None:
```
wujingyue, line 60: From @guihong-nv 

> I saw some models start to train the model with hybrid optimizer, like linear layers using Muon, other layers using AdamW. Do we currently support such hybrid configs?

I haven't tested it. Which OptimizerConfig reflects that? I could add a guard here so people don't enter uncharted area.

guihong-nv: I think we can skip the hybrid optimizer test for this PR. The hybrid optimizer is mainly supported by https://github.com/NVIDIA-NeMo/Emerging-Optimizers

yaoyu-33 (approved):

PR #5869, opened 2026-07-18
[Dev] Support full activation recompute for EP A2A overlap (VPP-stage recompute)

Wohox:

> Main-branch counterpart of this PR: NVIDIA/Megatron-LM#5920

# What does this PR do ?

Adds full activation recompute support for the EP A2A overlap path
(`--overlap-moe-expert-parallel-comm`), which previously hard-asserted
`recompute_granularity=full` off. Recompute is done at **VPP-stage / model-chunk**
granularity: only the input tensor of each VPP stage is retained across the
forward→backward gap, and the whole stage forward is recomputed at the start of
its backward. When `pipeline_model_parallel_size == 1` the VPP stage is the full
model chunk; when PP > 1 each VPP stage is a model chunk.

## Design

- The stage's initial forward runs under `no_grad` (no activation autograd graph
  retained); `pre_process` / `post_process` keep their graphs. Only the stage
  input tensor (`pre_process` output) survives the forward→backward gap, so the
  activation memory of every in-flight microbatch drops to a single input tensor.
- At the start of the stage backward, `recompute_vpp_stage_forward()` re-runs the
  whole stage forward with grad enabled by re-invoking the **same** `ScheduleNode`
  objects in forward order, so each node repopulates its own
  `inputs` / `output` / `detached` / `layer_state`. The activation→node mapping is
  therefore implicit and needs no bookkeeping.
- The A2A communication issued during this recompute is **exposed** (not
  overlapped) by design; the normal forward and normal backward A2A overlap is
  fully preserved.
- RNG states are snapshotted after `pre_process` and replayed for the recompute.
  MTP rolls `input_ids` / `position_ids` / `padding_mask` in place during the
  forward, so these are snapshotted and restored before the recompute to avoid a
  double roll.

## Constraints (asserted)

Under `overlap_moe_expert_parallel_comm` + `recompute_granularity=full`:
- `recompute_method` / `recompute_num_layers` must be `None` (the whole stage is
  recomputed as a single unit).
- `attention_dropout == 0` and `hidden_dropout == 0` — the forward RNG stream is
  interleaved across microbatches in the 1F1B schedule and cannot yet be
  faithfully replayed for nonzero dropout.
- Not yet supported (asserted off, left as follow-up work): Megatron FSDP, CUDA
  graphs (`cuda_graph_impl != "none"`), and `moe_paged_stash`.

## Test results
Verified with a DeepSeek-V3-proxy deterministic recipe (unfused attention,
`deterministic_mode`, `--no-force-balance`) on `TP1/PP4/VPP2EP16`:
Screenshot 2026-07-21 at 13 42 36
Screenshot 2026-07-21 at 15 51 28
Screenshot 2026-07-21 at 15 51 54

- Overlap + VPP-stage recompute is **bitwise-identical to standard Megatron
  `recompute_granularity=full`** (no overlap).

### Pre-checks
- [x] I have added relevant documentation (inline comments / docstrings)


megatron/core/models/common/model_chunk_schedule_plan.py
```diff
@@ -462,6 +516,92 @@ def release_state(self):
...
+            cs.mtp_hidden_states = None
+            self._recompute_state_snapshot = None
+
+        # Enable grad on the layer nodes' forward so the recompute builds the graph.
+        for layer_plan in self._transformer_layers:
+            layer_plan.set_forward_no_grad(False)
+
+        # Replay the forward RNG stream so dropout / rng-forked ops reproduce the
+        # initial forward. _fork_rng() restores the ambient RNG state on exit.
+        with _fork_rng():
```
lhb8125, line 667: **[CRITICAL Precision]** This recompute path replays the layer forwards as ordinary FP8 forwards. Unlike `te_checkpoint`, it never enters `activation_recompute_forward(..., recompute_phase=False/True)`. Transformer Engine uses those phases to snapshot and restore FP8 metadata (`scale`, `amax_history`, and first-module state); without them, the no-grad pass and recompute can both update metadata and the recompute may run with different scales, silently changing activations or gradients. Please either add the TE activation-recompute contexts around the initial and recompute forwards (and validate FP4 separately), or reject FP8/FP4 for this mode until supported.

megatron/core/models/common/model_chunk_schedule_plan.py
```diff
@@ -514,6 +657,12 @@ def run(
                 pre_backward(b_schedule_plan.vp_stage)
                 b_schedule_plan.record_current_stream()
 
+            # VPP-stage full recompute: rebuild the whole stage's layer forward
+            # graphs from the retained stage input before any layer backward runs.
+            # The A2A issued here is exposed (not overlapped) by design. No-op unless
+            # recompute is on.
+            b_schedule_plan.recompute_model_chunk_schedule_plan()
```
lhb8125, line 746: **[IMPORTANT Memory]** Please run `post_process.backward()` before rebuilding the stage. On the last PP stage, the current order materializes every recomputed transformer activation while output-projection, logits, and post-process saved tensors are still alive. With a large vocabulary this can create an avoidable peak or OOM and undercut the memory benefit. `post_process.backward()` produces `b_grad` independently of the recompute, so it can release the post-process graph first and then recompute the stage.

megatron/core/models/gpt/fine_grained_callables.py
```diff
@@ -309,6 +309,21 @@ def __init__(
...
+        """Release the forward activation tensors held by this node while keeping the
+        node reusable for a later recompute forward.
+
+        Used by the VPP-stage full recompute path (EP A2A overlap): after the initial
+        (no-grad) forward, the layer node's activation tensors are freed so that only
+        the stage input tensor survives the forward->backward gap. The same node
+        object is later re-run (with grad enabled) to rebuild ``inputs``/``output``/
+        ``detached`` before the backward pass.
+        """
+        self.inputs = None
```
lhb8125, line 377: **[IMPORTANT Memory]** Clearing only the node fields does not leave only the stage input. Each MoE dispatcher still holds initial-forward tensors such as `probs`, `routing_map`, and `reversed_local_input_permutation_mapping`; Flex managers have equivalent token-probability, index, and map caches. Those objects are per layer, so retained metadata accumulates across the stage until recompute overwrites it. Please add an explicit transient-forward-state reset to the dispatcher or communication manager and call it after the initial forward has completed and synchronized; recompute will repopulate the state needed for backward.

megatron/core/models/common/model_chunk_schedule_plan.py
Wohox: Added an assert rejecting only delayed-scaling FP8 on this path — it's the sole recipe that double-updates the persistent `amax_history`. Current-scaling recipes (mxfp8/blockwise) re-derive identical scales on the recompute; blockwise verified bitwise-identical to recompute-off on H100.

megatron/core/models/common/model_chunk_schedule_plan.py
Wohox: Done

megatron/core/models/gpt/fine_grained_callables.py
Wohox: Added a new API (`reset_transient_forward_state()`) on the alltoall + flex dispatchers to release the references to these transient forward tensors, called from the recompute-path activation release.

haloradar:
Hi~ I wonder whether the interleaved schedule and overlap of the recompute comm(dispatch/combine) with 1f1b is planed to be supported?

Wohox:
> Hi~ I wonder whether the interleaved schedule and overlap of the recompute comm(dispatch/combine) with 1f1b is planed to be supported?

I don't have plans for now. This is hard to support, because the recompute comm needs another layer's(also another microbatch) forward/backward to overlap with, there is simply no such candidate.

lhb8125 (approved):
Thanks!

Wohox:
/ok to test [830b30a](https://github.com/NVIDIA/Megatron-LM/pull/5869/commits/830b30a42e180c1663bf4a4b011a79ca04351bde)

Wohox:
/ok to test [42658d9](https://github.com/NVIDIA/Megatron-LM/pull/5869/commits/42658d9dc99a9df4c3d240218389469e4dfa3a1c)

PR #5883, opened 2026-07-20
Avoid per-step host-to-device copy in the CSA compressed-index path

zkyue:
### What
`CompressedSparseAttention` builds the compressed top-k index tensor on every forward step with

```python
torch.where(valid, topk_indices_compressed + offset, torch.tensor(-1, device=x.device))
```

The `torch.tensor(-1, device=x.device)` term allocates a fresh device scalar on each call, which lowers to a synchronizing host->device copy on the hot forward path.

### Change
Use `torch.where`'s scalar overload (already used elsewhere in this file):

```python
torch.where(valid, topk_indices_compressed + offset, -1)
```

### Why it's safe
`topk_indices_compressed + offset` is int64; the scalar sentinel keeps that dtype and every element value. Verified elementwise-identical on random inputs -- the selected indices are bitwise-unchanged. Only the per-step host->device copy and its implicit stream synchronize are removed; the original single select kernel is unchanged (no extra device->device passes).

### Impact
Removes one synchronizing H2D copy per forward step. Most noticeable at high step counts and under CUDA-graph capture, where a mid-path device-scalar allocation forces a stream break.

One of two small, independent host-side sync-reduction fixes in the experimental sparse-attention variants.


yaox12 (approved):

zkyue:
@yaox12 Thanks for the approval! The `linting` failure on bc07616 is fixed in afa259f5e — black wanted the multi-line `torch.where` collapsed to a single line (verified locally against the CI's pinned linters: black/isort/pylint/ruff all clean now).

Since the push changed the head SHA, the earlier `/ok to test` no longer applies — could you re-authorize CI on afa259f5e when you get a chance? Thanks!

zkyue:
Correcting my comment above: I had the root cause wrong, and the real one is on my side rather than in CI.

`cicd-mbridge-testing` is not what ejected this PR. The PR was removed from the merge queue at 15:26:59 on 2026-07-30, 49 s after being enqueued, while that job did not fail until 18:15:57 — 2 h 49 min later. It also does not gate the queue: on 08-01/08-02, ten PRs merged normally while their merge-group run showed `cicd-mbridge-testing` red.

The actual cause was DCO, on the merge-group commit (`3172f5dd`):

- the squash sets the commit author to my account's primary address, `kaiyue.zhou@z.ai`;
- the `Signed-off-by:` trailer was copied verbatim from my original commit, which used my older `51477259+zkyue@users.noreply.github.com` address;
- the DCO App therefore reported `action_required` ("There is one commit incorrectly signed off") at 15:26:28, and the queue ejected the PR with `reason: failed_checks`.

DCO passes on the PR branch itself, where author and sign-off are both the noreply address and agree — so none of the four CI rounds could have surfaced this. The mismatch only comes into existence at the moment the queue squashes.

I have re-signed the commit so that author and sign-off both use the address matching my account (`d2fa670` → `e56596d1`). The push carries no code change: the tree hash is identical (`10d7cd88…`), `git patch-id` is unchanged (`766e7102…`), `git diff d2fa670 e56596d1` is empty, and the original author date is preserved. NVIDIA CI is not authorized on the new SHA yet, so those checks are currently pending.

One upstream note, since this will affect other external contributors: the `DCO_merge_group` placeholder job in `cicd-main.yml` is intended to keep DCO from re-running inside the queue, but the DCO App posts its own check — also named `DCO` — on merge-group commits, and that is the one the queue acts on. Any contributor whose sign-off address differs from their account's primary address gets ejected in under a minute, with every check on the PR green and nothing in the PR's own CI pointing at the cause.

PR #5888, opened 2026-07-20
Add standard MLite local validation

kunlunl:
## Summary

- add a local test entry for the standard 8-GPU H100 environment
- organize MLite tests by model, primitive, runtime, and examples
- include the MLite model and parallel fixes exposed while consolidating the test workflow


ISEEKYAN:
The proposed model/primitive/runtime/examples split is a useful first step, but the test tree is still too flat. I would prefer tests to mirror the production package hierarchy.
For example, FSDP2 lives under primitive/optimizers/fsdp2, so its tests should live under tests/unit/primitive/optimizers/fsdp2/ and be split by source responsibility (config, wrap, optimizer, adamw, grad_clip) instead of being combined in test_fsdp2_unit.py.
The same rule should apply to GPU and smoke coverage. Tests owned by FSDP2 can remain under the mirrored FSDP2 path, while genuine cross-layer scenarios—such as FSDP2 combined with runtime checkpointing—should be placed under an explicit integration/workflow path based on the components’ nearest common scope.
This would make test ownership and discovery much clearer and prevent unit/primitive from becoming another flat collection of large test files. I suggest establishing this layout before adding the new discovery runner, since the runner already discovers recursively and therefore does not require a flat structure.

like:
```
tests/
├── unit/
│   ├── primitive/
│   │   ├── optimizers/
│   │   │   └── fsdp2/
│   │   │       ├── test_config.py
│   │   │       ├── test_wrap.py
│   │   │       ├── test_optimizer.py
│   │   │       ├── test_adamw.py
│   │   │       └── test_grad_clip.py
│   │   └── parallel/
│   │       ├── test_state.py
│   │       ├── test_pipeline.py
│   │       └── test_context_parallel.py
│   └── runtime/
│       └── checkpoint/
│           └── test_checkpoint.py
├── smoke/
│   └── primitive/
│       └── optimizers/
│           └── fsdp2/
│               ├── test_offload.py
│               └── test_distributed.py
└── integration/
    └── runtime/
        └── checkpoint/
            └── test_fsdp2_resume.py
```

ISEEKYAN (approved):

FDecaYed (approved):

PR #5451, opened 2026-06-23
Optimize memory usage of partial CUDA graphs

jiemingz:
- [x] I, the PR author, have personally reviewed every line of this PR.

# What does this PR do ?

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact @NVIDIA/mcore-oncall.

## Issue tracking

For PRs from open-source community contributors:

- **New features**: a linked issue is **required**. Please open a [feature request](https://github.com/NVIDIA/Megatron-LM/issues/new?template=feature_request.md) and reference it here before submitting the PR.
- **Small updates (bug fixes, minor improvements)**: a linked issue is **recommended** and will accelerate the PR review process.

Linked issue: 

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment @NVIDIA/mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

Once all required reviewers have approved, the `Approved` label is applied **automatically**.

### Merge

Any member of [mcore-engineers](https://github.com/orgs/NVIDIA/teams/mcore-engineers) will be able to merge your PR.

PR #5893, opened 2026-07-20
Disable TE CUDA graph parameter gradient clones

buptzyb:

# What does this PR do?

Disable Transformer Engine parameter-gradient clones for MCore CUDA graphs with Transformer Engine 2.19.0 and later.

Transformer Engine's `clone_param_grads_on_return` option keeps parameter-gradient outputs stable across later graph replays. MCore consumes parameter gradients through its accumulation hooks during each replay, before those static buffers can be overwritten, so retaining cloned gradients is unnecessary and extends their lifetime.

This change passes `clone_param_grads_on_return=False` when the installed Transformer Engine version is at least 2.19.0. Older Transformer Engine versions remain supported because MCore does not pass the new keyword to them.

## Issue tracking

Related to NVIDIA/TransformerEngine#2937.

### Pre-checks

### Validation

- Existing `TECudaGraphHelper` coverage now checks the final `make_graphed_callables` kwargs for both sides of the TE 2.19.0 version boundary.
- `tests/unit_tests/transformer/test_cuda_graphs.py` passed in the MCore CI container on eight H100 ranks: 67 passed, 3 skipped, and 20 deselected per active rank.
- Black, isort, pylint, ruff, Python compilation, and `git diff --check` passed.

### Code review

Feel free to message or comment @NVIDIA/mcore-oncall to help accelerate review once CI is green.

### DIFF
--- a/megatron/core/transformer/cuda_graphs.py
+++ b/megatron/core/transformer/cuda_graphs.py
@@ -2707,6 +2707,10 @@ class TECudaGraphHelper:
                 # Starting from TE 2.7.0, make_graphed_callables() optimizes the graph memory usage
                 # by reusing input/output data buffers between graphs.
                 kwargs['_reuse_graph_input_output_buffers'] = True
+            if is_te_min_version("2.19.0"):
+                # MCore consumes parameter gradients through its accumulation hooks during each
+                # replay, before TE can overwrite the static buffers on a later replay.
+                kwargs['clone_param_grads_on_return'] = False
 
             if sample_kwargs:
                 kwargs['sample_kwargs'] = sample_kwargs
--- a/megatron/core/transformer/cuda_graphs.py
+++ b/megatron/core/transformer/cuda_graphs.py
@@ -104,6 +104,14 @@ def get_gtp_runner_streams() -> List[torch.cuda.Stream]:
     return _GTP_RUNNER_STREAMS
 
 
+def _set_te_param_grad_clone_policy(kwargs: Dict[str, Any]) -> None:
+    """Disable TE parameter-gradient clones when the public option is available."""
+    if 'clone_param_grads_on_return' in inspect.signature(make_graphed_callables).parameters:
+        # MCore consumes parameter gradients through its accumulation hooks during each replay,
+        # before TE can overwrite the static gradient buffers on a later replay.
+        kwargs['clone_param_grads_on_return'] = False
+
+
 def _set_skip_fp8_weight_update_tensor(skip: bool) -> None:
     """Toggle TE's FP8 "skip weight refresh" flag between microbatches.
 
@@ -2437,6 +2445,7 @@ class TECudaGraphHelper:
                 # Starting from TE 2.7.0, make_graphed_callables() optimizes the graph memory usage
                 # by reusing input/output data buffers between graphs.
                 kwargs['_reuse_graph_input_output_buffers'] = True
+            _set_te_param_grad_clone_policy(kwargs)
 
             if sample_kwargs:
                 kwargs['sample_kwargs'] = sample_kwargs
--- a/tests/unit_tests/transformer/test_cuda_graphs.py
+++ b/tests/unit_tests/transformer/test_cuda_graphs.py
@@ -1230,6 +1230,10 @@ class TestTECudaGraphHelper:
         assert (
             'sample_kwargs' in make_graphed_callables_kwargs
         ), "sample_kwargs should be present in make_graphed_callables_kwargs for TE >= 1.10.0"
+        if is_te_min_version("2.19.0"):
+            assert make_graphed_callables_kwargs['clone_param_grads_on_return'] is False
+        else:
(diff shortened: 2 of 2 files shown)


pggPL:
@buptzyb I talked to mbridge folks and if we merge this PR and TE PR before next releases everything should be fine. I think it is good idea to merge this PR first and then merge TE one to avoid problems in case of delays in review of this PR.

cspades:
FYI https://github.com/pytorch/pytorch/issues/181723

From what I understand, `AccumulateGrad` uses the first BWD gradient as the accumulation buffer in `param.grad`, so if that buffer is a CG grad buffer, then `AccumulateGrad` will initialize `param.grad` to the CG buffer, which will get overridden by the gradient of the 2nd microbatch during CG reply, which then accumulates on itself.

Megatron (DDP, FSDP) use `param.main_grad` for accumulation so `param.grad` being overridden or even dereferenced is fine. @buptzyb correct me if I am wrong.

cspades:
Also this is highly related to @jiemingz 's PR that removed the wgrad clone: https://github.com/NVIDIA/Megatron-LM/pull/5451/changes#diff-a3f9fb509ad46e999a7d8e1b6850c40d2ca0a85344c749096008fefebf51ab08L690

The difference is here we are adding back the clone on the TE side for the vanilla PyTorch use case of `param.grad`, but removing it for Megatron since `param.main_grad` exists.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -104,6 +104,14 @@ def get_gtp_runner_streams() -> List[torch.cuda.Stream]:
     return _GTP_RUNNER_STREAMS
 
 
+def _set_te_param_grad_clone_policy(kwargs: Dict[str, Any]) -> None:
```
jiemingz: did the TE change get tied to a TE version? It would be nice if we could do a version check here

buptzyb: According to @pggPL 's suggestion above, we'd better merge this PR on the megatron side first, then add the knob on the TE side. So there is no certain TE version now.

pggPL: I've heard that @cspades is responsible for MCore-TE integration now, so @cspades what is your opinon?
From TE point it does not matter, so you can ignore my opinion.

cspades: TE should go in first followed by a version or attribute guard for this argument, not the other way around. There is a delay of 2 per cycle when TE is updated, so no NeMo repository will feel this change immediately.

@buptzyb can you make sure the perf and convergence is the same when turning off the clone? A basic TE CG run will do, unless a unit or functional test already covers this. Then we can merge the TE PR, and then the MCore PR.

buptzyb: @cspades do you want the test on mcore side or TE side? On TE side there are convergence tests "clone on vs eager" and "clone off vs eager", but there is no perf measurement. On mcore side there are functional tests that compare TE CG convergence & perf to golden. The perf tolerance is 5%. But this is not a direct clone on vs off test.

cspades: That is indeed a bit hard to test, I was initially just thinking a casual test, but even that will be hard to detect, a better test would likely be to check the `id` of the accumulated gradient but DDP and FSDP hooks will delete `param.grad` so would need to intercept the grad clone and then compare with the CG output grad...

...forget it, can you just casually make sure that DDP and FSDP on the MCore side do not have a copy/clone operation in the Nsys profile?

cspades: BTW, this is not necessarily blocking merge, just due diligence. Worse case after merge we can revert and retry the PR, or we can fix-forward.

I suppose I must ask this question to reduce the probability that our TE CG perf will be damaged. 😅

buptzyb: @cspades I can try to dump nsys reports of DDP with clone on/off to compare. But for FSDP, as you know, we didn't merge the v1 CG PR, and v2 CG is also not ready, right?

cspades: Yes, I completely forgot. But in any case, the TE PR can be merged, and this PR just needs a version and/or argument guard and it can be merged as well.

buptzyb: Got it. @pggPL so, could you help merge the TE PR first? Thanks!

cspades (approved):

megatron/core/transformer/cuda_graphs.py
cspades: Merged on TE side, I approved this one, signature inspection guard looks great. @jiemingz the task is done, could you review as well?

buptzyb: @jiemingz do you think we need a TE version or just signature inspection? This feature is likely to go with TE 2.19.0.

jiemingz (approved):

yaoyu-33 (approved):

PR #5895, opened 2026-07-20
Add offline Megatron-FSDP (fsdp_dtensor) to torch_dist checkpoint converter

ahmadki:

# What does this PR do?

Adds an **offline, CPU-only, model-free converter** that rewrites a Megatron-FSDP
`fsdp_dtensor` checkpoint as a native `torch_dist` checkpoint, so a classic
(non-FSDP) N-D-parallel job — DDP/TP/PP/EP/CP/VPP — can resume from a model that
was trained with Megatron-FSDP, restoring **weights and the full
distributed-optimizer state**, reshardable into any target parallel layout on
load.

It is the inverse of the existing `convert-torch-dist-to-fsdp-dtensor` command
and lives in the same tool:

```bash
python tools/checkpoint/checkpoint_inspector.py \
    convert-fsdp-dtensor-to-torch-dist \
    /path/to/in/iter_XXXXXXX  /path/to/out/iter_XXXXXXX
```

Because PyTorch DCP records each tensor's full global shape in metadata, a plain
single-rank load fully gathers every tensor regardless of the TP/PP/EP/FSDP
sharding it was written with — so the converter is parallelism-agnostic and needs
no GPU and no model instantiation. For checkpoints larger than one host's RAM it
can be launched under `torchrun --nproc_per_node N`, which shards the tensor items
across processes by transform-group (peak host memory ≈ 1/N) and cooperatively
writes a single store identical to the one-process output.

## Issue tracking

# Detailed description

## What the converter does

Per tensor/key, it inverts the forward split and emits bare mcore keys plus
`common.pt` / `metadata.json`:

- **Strips** the Megatron-FSDP wrapper prefixes (`model.module.`, and any run of
  `module.` wrappers).
- **Merges** the SwiGLU `_w`/`_v` fc1 pairs back into a single fused fc1 tensor
  (auto-detected from the `_w`/`_v` tags; `--swiglu-modules` restricts merging to
  named module scopes, e.g. VLM `language_model` only).
- **Re-stacks** per-expert weights into mcore's grouped
  `...mlp.experts.experts.linear_fc{1,2}` axis-0 tensor — handling **both**
  grouped-GEMM (`TEGroupedMLP`, `linear_fc1.weight{idx}`) and non-grouped
  (`SequentialMLP`, `local_experts.{idx}.linear_fc1.weight`) source layouts.
- **Stacks** per-layer decoder tensors into the `(num_layers, *param)` layout for
  homogeneous blocks (auto-detected; `--stack-layers`/`--non-homogeneous-layers`
  override).
- **Renames** Multi-Token-Prediction keys (`mtp_model_layer` → `transformer_layer`).
- **Splits** the fused Gated-DeltaNet and Mamba-2 `in_proj`/`conv1d` projections
  into the per-component sub-keys (`query/key/value/z/beta/alpha` for GDN;
  `z/x/B/C/dt` for Mamba-2) that a classic `torch_dist` load expects.
- **Reconstructs the distributed-optimizer state**: reverses the optimizer-key
  remap (`optimizer.state.<subkey>.<param>`), synthesizes the fp32 master weights
  Megatron-FSDP stores implicitly (the fp32 params *are* the masters) and
  downcasts the model section to the training compute dtype (bf16/fp16), and
  rebuilds standard `param_groups` from Megatron-FSDP's per-parameter
  `param_to_group_meta` map.
- **Emits the `fully_reshardable` content metadata** so the produced checkpoint
  loads as the one distributed-optimizer format whose on-disk layout is
  per-parameter and model-shaped — matching what `fsdp_dtensor` holds.
- **Fails loudly** (`NotImplementedError`) on structures it cannot faithfully
  invert (e.g. an un-indexed stacked-layer buffer), rather than silently emitting
  a corrupt checkpoint.

## Resuming a classic job from the converted checkpoint

To resume with optimizer state, the classic job must set:

```
--ckpt-format torch_dist --use-distributed-optimizer \
--dist-ckpt-optim-fully-reshardable \
--dist-ckpt-strictness log_all \
--no-load-rng
```

- `--dist-ckpt-optim-fully-reshardable` — required only when resuming with
  optimizer state; selects the per-parameter, model-shaped optimizer format the
  tool emits. (Use `--no-optimizer` at convert time for a weights-only checkpoint
  if optimizer resume is not needed.)
- `--dist-ckpt-strictness log_all` — the converted checkpoint omits TransformerEngine
  `_extra_state` blobs (bf16-safe); a non-`assume_ok_unexpected` strictness drops
  those keys from the load template instead of erroring.
- `--no-load-rng` — RNG state is not round-tripped.

## Files in this PR

| Area | Path | Notes |
|---|---|---|
| Converter | `tools/checkpoint/checkpoint_inspector.py` (+923/-13) | New `convert-fsdp-dtensor-to-torch-dist` subcommand + `reverse_convert_checkpoint` and the pure transform/sharding helpers |
| Unit tests | `tests/unit_tests/tools/checkpoint/test_reverse_convert.py` | ~60 CPU-only, pytest-collected tests of every pure transform, asserting tensor values |
| Round-trip | `tests/unit_tests/tools/checkpoint/test_reverse_convert_roundtrip.py` | Standalone GPU round-trip script (not pytest-collected) |
| User guide | `docs/user-guide/features/megatron_fsdp.md` (+26) | New "Converting `fsdp_dtensor` → N-D Parallel (`torch_dist`)" section |
| Validation harness | `tools/checkpoint/fsdp_dtensor_to_torch_dist_validation/` | Manual GPU harness: `common.sh`, `run_all.sh`, `validate_{resume,reshard,source_sharding}.sh`, `validate_bitexact.py`, `models/*.sh`, README |
| Skill | `skills/mcore-reverse-checkpoint-conversion/SKILL.md` | Contributor guide for using/extending the converter and harness |

# Supported architectures

Dense, SwiGLU, MoE (grouped-GEMM **and** SequentialMLP experts, including
aux-loss-free routing — its fp32 `router.expert_bias` buffer is preserved, not
downcast), Multi-Latent Attention, Multi-Token Prediction, Gated-DeltaNet, and
Mamba-2 hybrids.

# How this was validated

**Automated (CI, CPU):** `test_reverse_convert.py` — ~60 unit tests covering
prefix stripping, optimizer-key reversal, SwiGLU merge, grouped & SequentialMLP
expert re-stacking, layer stacking, homogeneity detection, MTP rename, GDN &
Mamba-2 splits, `param_groups` rebuild, the fp32-keep fence, the unsupported-scope
fence, and the multi-rank output-group grouping. These assert **actual tensor
values** (e.g. concatenating the split parts reproduces the fused blob), not just
shapes/keys. Verified: 60 tests pass on CPU in the dev container, and the
file runs in CI via the `tests/unit_tests/**` bucket.

> Coverage boundary: CI gates the **pure leaf transforms**. The orchestrator
> `reverse_convert_checkpoint` itself — fp32-master synthesis + bf16 downcast,
> `content_metadata` emission, `param_groups` re-nesting, the classify/skip loop —
> and the multi-rank shard-assignment (`_assign_tensor_keys_to_rank`) are **not**
> CI-gated; they are exercised by the GPU harness below.

**Manual (dev container, GPU):**
- **8-family GPU harness** (`fsdp_dtensor_to_torch_dist_validation/`): trains a
  small model with Megatron-FSDP, converts, and resumes a classic job — checking
  the model loads at the right iteration and the first resumed `lm loss` matches
  the FSDP loss within bf16 tolerance with the LR exact. Covers dense, dense+SwiGLU,
  dense+FP8, MoE(gated), MoE(grouped), MoE+MLA+MTP, MTP, and GDN-hybrid; plus
  load-side TP/PP/EP resharding and source-side (multi-GPU FSDP) sharding drivers,
  and a per-tensor bit-exact diff.

> Note on the harness: the shell drivers use `set -uo pipefail` (not `set -e`) and
> end in an `echo`, so their exit code is not a reliable pass/fail signal — the
> real evidence is the **human-read loss/LR continuity** in the logs, which is the
> documented PASS criterion.

# What is NOT covered / known limitations

- RNG state, rerun-state, and all `_extra_state`
  (including FP8 amax history: already discarded in the `fsdp_dtensor` source).
  FP8 resume re-initializes amax and therefore tracks ~1% looser than bf16
  tolerance — expected.

## Pre-checks

- [x] I have added relevant unit tests (`test_reverse_convert.py`, CPU, CI-collected)
- [x] I have added relevant documentation (user guide + skill)
- [x] I have run `tools/autoformat.sh` on my PR 

# Notes for reviewers

- The reverse converter is inserted whole after `convert_torch_dist_to_fsdp_dtensor`
  in `checkpoint_inspector.py`; no pre-existing function bodies were modified (only
  the import block, reordered by `isort`).


tools/checkpoint/fsdp_dtensor_to_torch_dist_validation/README.md
```diff
@@ -0,0 +1,272 @@
+# Reverse checkpoint converter — validation harness
+
+End-to-end validation for the offline **`fsdp_dtensor` → `torch_dist`** reverse
+checkpoint converter
+(`tools/checkpoint/checkpoint_inspector.py convert-fsdp-dtensor-to-torch-dist`).
```
wujingyue, line 5: Would it be better to put this in a separate binary, such as convert_dtensor_to_torch_dist?
1. We can certainly reuse methods from checkpoint_inspector, but it feels odd for an “inspector” tool to also perform format conversion.
2. I’m also not sure fsdp belongs in the name. The implementation appears to be more general than FSDP, or at least could be extended to support DTensor more broadly.

tools/checkpoint/fsdp_dtensor_to_torch_dist_validation/validate_bitexact.py
wujingyue, line 1: Can we make these validation scripts unit tests? Otherwise they won't be protected by CI and can be broken anytime. Either in https://github.com/NVIDIA/Megatron-LM/blob/cfb116f28875cb69868b2c3f509082349443ef91/tests/unit_tests/test_checkpointing.py or tests/unit_tests/checkpoint/...

wujingyue (commented):
Thanks for the PR! 

I only left a few high-level comments. I'll focus on checkpoint_inspector.py next.

tools/checkpoint/checkpoint_inspector.py
wujingyue, line 1: Can you try your best to https://google.github.io/eng-practices/review/developer/small-cls.html even this file? For example, can swiglu, mamba, gdn be added separately? Not saying these features are not important but separating them out can help reviewers understand the code structure much better.

tools/checkpoint/fsdp_dtensor_to_torch_dist_validation/README.md
wujingyue, line 1: This page covers validation, which is helpful.

To start reviewing the implementation, though, I'm looking for the core APIs and their contract. I'm sure this will also be more clear after you split the PR.

tools/checkpoint/fsdp_dtensor_to_torch_dist_validation/validate_bitexact.py
ahmadki: This is something I considered, but there are two issues that made settle for the scripts approach.
First, the scripts do 2 type of validations:

- Resume: Train with FSDP to 100 iterations, convert iterations 60 and 80 to `torch_dist`, resume training from the converted checkpoints (train for 3 iterations). We expect the loss to be within a certain tolerance (BF16) from FSDP loss and the LR to be exact. 
    - Technically, this validation is three parts, (source, target): (single rank, single rank) & (single rank, shared) & (sharded, single rank). We can probably trim these to only (single rank, shared).
- Per tensor diff: We load the converted tensor to a real mode, then re-save them. We expect the loaded and saved checkpoints to be bit-exact. Basically, we are not mismatching or omitting tensors.

Converting these to tests would mean:

1. Formalizing a passing rubric for the resume part. Is BF16 tolerance suitable or not ? What about FP8 where scaling statistics are not in the checkpoint ? Are 3 iterations post-load enough or should we train for longer ?
2. Saving, conversion, loading are launched as sub-processes. I need to look how to code them into pytests as we don't have similar patterns in the CI. Do we have the GPU budget to run these tests or can we cache golden values ?

tools/checkpoint/fsdp_dtensor_to_torch_dist_validation/README.md
ahmadki: Valid point, but rather than a mechanical split, I think we need to re-write the entire `tools/checkpoint` module. It doesn't make sense for checkpoint conversion to live under an "inspection" model.
I'll share a proposal privately.

tools/checkpoint/checkpoint_inspector.py
ahmadki: Yes. My goal with the PR was to provide as a complete of feature as I can, that is also validate to the best of my abilities (even the optimizer state conversion was outside the scope of the issue)
If you think this is feature complete, I can split it into smaller PRs that are easier to review individually.

wujingyue:
@ahmadki and I discussed this PR offline. Given the size of the PR, we probably need a design doc before we refine the PR. The doc should ideally cover:
* **Scope** (e.g., resume training, precision handling, etc.).
* **API**
* **High-level architecture**
* **Necessary complexity.** For example, a per-module adapter seems difficult to maintain and should probably be avoided unless there's a strong reason. Also, can MXFP8 stats be recomputed from the checkpoint, given that it stores high-precision values?
* **Testing plan**, for example:
  * Round-trip to `torch_dist` with bit-exact verification.
  * End-to-end conversion: **FSDP → DTensor → MCore load → torch_dist → MCore save → torch_dist**, followed by a bit-exact comparison of the two `torch_dist` checkpoints.
* **Rough timeline.**
  * This should ultimately be a transitional solution that we intend to throw away, so we probably don't want to over-invest in the implementation.
  * Explain how this proposal aligns with the longer-term plan of using DTensor everywhere.

Am I missing anything?

ahmadki:
I think most of the appeared complexity is coming the from the validation module. The conversion logic is limited to one file, it's not "simple", but also not very "complex".
Since we agree that this PR is feature complete, or we are not missing important models. I'm working on a plan that:

1. Refactor `checkpoint_inspector.py`. The conversion really shouldn't be living there, but since `torch_dist` -> `fsdp_dtensor` was already implemented there, I kept it. Now the file has ballooned to 1k+ lines.
2. Validations through pytests. I still don't have an answer how this will be done though, see https://github.com/NVIDIA/Megatron-LM/pull/5895#discussion_r3618153444
3. Smaller, incremental PRs.

PR #5898, opened 2026-07-20
fix: exclude bucket-end padding from optimizer checkpoints

cuichenx:
# What does this PR do?

## The problem

Saving a `dp_reshardable` distributed-optimizer checkpoint could fail whenever a bucket carried DP/alignment padding and a rank's final shard mixed two different kinds of padding: valid intra-bucket padding that belongs in the checkpoint, and runtime bucket-end padding that does not. Because the shard was allowed to run all the way to the physical bucket end, it extended past the bucket's declared *unpadded* global size. PyTorch distributed checkpointing then rejected the global save plan and left an incomplete checkpoint behind.

We first hit this in a Nemotron Super PP=2 run, where a shared-embedding bucket boundary produces exactly this geometry — but the bug is not specific to PP=2. Any topology that lands a bucket boundary in the same place is affected.

## The fix

Clip each rank's terminal padding shard to the global unpadded bucket boundary instead of to the physical bucket end. This emits only the valid checkpoint-layout padding and drops the runtime-only bucket tail, yielding a save plan that PyTorch distributed checkpointing accepts for any topology with this bucket geometry.

## Testing

A new DP=8 regression test pins the affected padded/unpadded bucket sizes and the terminal shard geometry, then performs a full distributed-checkpoint save to confirm the plan validates.

<details>
<summary>Concrete failing example, with the numbers worked out</summary>

With PP=2, Nemotron Super isolates its shared embedding copy in its own optimizer bucket. That shifts the *preceding* bucket's boundary, and in the failing run the dense DP size was 8.

Two distinct kinds of padding meet at that boundary:

1. **Checkpoint-layout padding (64-element parameter alignment).** The last real optimizer-state chunk ends at global offset 4,256. Parameter starts must be aligned to multiples of 64, so the next aligned position is:

   ```text
   ceil(4256 / 64) * 64 = 4288
   ```

   Hence `gbuf_world_numel_unpadded = 4288`. The range `[4256, 4288)` is a 32-element alignment gap that *is* part of the logical checkpoint layout.

2. **Runtime bucket-end padding (DP/128 alignment).** The physical bucket end is aligned to `lcm(DP, 128)`:

   ```text
   lcm(8, 128) = 128
   ceil(4288 / 128) * 128 = 4352
   ```

   Hence `gbuf_world_numel = 4352`. The range `[4288, 4352)` is 64 elements of runtime-only bucket-end padding that must **not** be checkpointed.

Each DP rank owns an equal physical slice:

```text
gbuf_local_numel = gbuf_world_numel / DP
                 = 4352 / 8
                 = 544
```

For DP rank 7, the part of its slice that lies inside the logical checkpoint is:

```text
gbuf_local_numel_unpadded
    = clamp(gbuf_world_numel_unpadded - rank * gbuf_local_numel,
            0,
            gbuf_local_numel)
    = clamp(4288 - 7 * 544, 0, 544)
    = 480
```

Rank 7 starts at global offset `7 * 544 = 3808`, and its last real optimizer-state chunk ends at local offset `4256 - 3808 = 448`. So the correct terminal checkpoint padding is `[448, 480)` — 32 elements.

The old code instead extended that padding to `gbuf_local_numel = 544`, giving `[448, 544)` — 96 elements. Globally that shard spanned `[4256, 4352)`, overshooting the declared checkpoint shape of 4,288 by 64 elements. It hit `param`, `exp_avg`, and `exp_avg_sq` alike, and caused PyTorch distributed checkpointing to reject the save plan with `ValueError: Failed to validate global plan`.

This PR uses `gbuf_local_numel_unpadded` as the terminal padding endpoint, so it emits only the valid 32-element checkpoint-layout gap and excludes the 64-element runtime bucket tail.

</details>

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

Any member of [mcore-engineers](https://github.com/orgs/NVIDIA/teams/mcore-engineers) will be able to merge this PR.

### DIFF
--- a/megatron/core/optimizer/distrib_optimizer.py
+++ b/megatron/core/optimizer/distrib_optimizer.py
@@ -1856,6 +1856,13 @@ class DistributedOptimizer(MixedPrecisionOptimizer):
                     assert gbuf_world_numel_unpadded <= gbuf_world_numel
                     assert gbuf_world_numel % data_parallel_world_size == 0
                     gbuf_local_numel = gbuf_world_numel // data_parallel_world_size
+                    gbuf_local_numel_unpadded = max(
+                        0,
+                        min(
+                            gbuf_local_numel,
+                            gbuf_world_numel_unpadded - data_parallel_rank * gbuf_local_numel,
+                        ),
+                    )
 
                     sharded_bucket_key = (
                         f'optimizer.distributed.dp_group_idx_{self.data_parallel_group_idx}'
@@ -1870,8 +1877,9 @@ class DistributedOptimizer(MixedPrecisionOptimizer):
                     all_pad_tensors = {}
                     for i in range(-1, len(bucket_state)):
                         if i == len(bucket_state) - 1:
-                            # Potential padding at the end
-                            next_param_start = gbuf_local_numel
+                            # Potential intra-param padding at the end. Exclude the
+                            # bucket-end padding beyond the global unpadded boundary.
+                            next_param_start = gbuf_local_numel_unpadded
                         else:
                             next_param_start = bucket_state[i + 1]['gbuf_local_start']
                         if i == -1:
--- a/megatron/core/optimizer/distrib_optimizer.py
+++ b/megatron/core/optimizer/distrib_optimizer.py
@@ -1856,6 +1856,9 @@ class DistributedOptimizer(MixedPrecisionOptimizer):
                     assert gbuf_world_numel_unpadded <= gbuf_world_numel
                     assert gbuf_world_numel % data_parallel_world_size == 0
                     gbuf_local_numel = gbuf_world_numel // data_parallel_world_size
+                    # Number of elements in this rank's DP-local bucket shard that fall within the
+                    # global unpadded range. This is the full local size on earlier ranks, partial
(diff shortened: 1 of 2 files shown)


zhongbozhu:
Can you put a detailed description in the PR of how did you find this bug and what does the error look like if we don't fix it?

zhongbozhu (approved):
LGTM

Phlip79 (approved):

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -1856,6 +1856,13 @@ def sharded_param_state_dp_reshardable(
                     assert gbuf_world_numel_unpadded <= gbuf_world_numel
                     assert gbuf_world_numel % data_parallel_world_size == 0
                     gbuf_local_numel = gbuf_world_numel // data_parallel_world_size
+                    gbuf_local_numel_unpadded = max(
```
deepakn94, line 1862: Can you add a comment here to describe what `gbuf_local_numel_unpadded` keeps track of?

deepakn94:
Can you add a concrete example to this PR description of this bug's manifestation?

PR #5916, opened 2026-07-21
Fix gradient-norm undercounting when using EP and TP

philipcmonk:

# What does this PR do?

This PR fixes correctness bugs surrounding expert-parallel training when combined with tensor parallelism and/or expert-tensor parallelism.

There are two related issues, both applying only to MoE models:

- Common case: if EP > 1 and TP > 1, then the gradient norm calculations omit some experts, so gradient clipping applies less than expected.  This could reduce stability during warmup and in case of gradient spikes.  This usually applies only if `--moe-grouped-gemm` is set.

- Uncommon case: if EP = 1 and ETP != TP (which would be unusual), then the gradients themselves are not correctly reduced across the expert data-parallel group.  If ETP > TP, then unrelated gradients will mix (probably with other columns of the same expert).  If TP > ETP, then each expert will only receive a portion of the gradient it should, and that portion will be different on different ranks. This will cause ranks to diverge -- they will think the experts are replicated across the EDP group, but they are not.

In both cases, the logging for the parameter norms, gradient norms, and count_zeros will also be incorrect.

These come from three main causes:
- param.allreduce=False is used to indicate that a parameter is subject to the expert topology (EP/ETP/EDP) instead of the normal topology (TP/CP/DP). It was only being set if EP > 1, however it should also be set if ETP != TP, since that still implies that EDP != DP.

- `param_is_not_tensor_parallel_duplicate` assumed the parameter was subject to the normal topology.  It should check the fixed param.allreduce flag, and if it's subject to the expert topology, use the ETP group instead of the TP group.

- `TEGroupedLinear` did not mark experts with param.tensor_model_parallel when it should have (to match other linear layers).

Besides fixing these, this PR also attaches expert_tp_group to the optimizer alongside the existing tp_group.  We need this at all the call-sites for param_is_not_tensor_parallel_duplicate.

Finally, we add these tests which were useful in reproducing and fixing the bugs:
- Set the params and gradients to all 1s and verify the calculated L2 norms are sqrt(num_params) under a few different parallelisms.
- In the same condition, run gradient clipping and assert each gradient is now `max_norm / sqrt(num_parameters)`.
- To verify the correct reduction of gradients, for an expert linear layer, with EP=1, set the gradients to 2**rank, then call finish_grad_sync from DDP. Assert the expected gradients come out of this. The rank-specific values are necessary to distinguish the case where gradients were being mixed with other ranks when ETP!=TP.

### Pre-checks

- [X] I have added relevant functional tests


zhongbozhu:
Have we added shared expert to your test coverage, might be better to have one (unit test or functional test).

philipcmonk:
It was not included, but it's easy to add to the param-norm and gradient clipping tests.  Done in ba58aa0

AAnoosheh (approved):

yaoyu-33 (approved):

megatron/core/extensions/transformer_engine.py
```diff
@@ -86,6 +86,25 @@
...
 
 
+def _set_expert_parameter_attributes(
+    module: torch.nn.Module, parallel_mode: Optional[str], use_expert_groups: bool
+) -> None:
+    """Set expert gradient-reduction and tensor-partition metadata."""
+    for name, param in module.named_parameters(recurse=False):
+        param.allreduce = not use_expert_groups
+
+        is_weight = name == "weight" or (name.startswith("weight") and name[6:].isdigit())
```
deepakn94: Uh oh, `name[6:]` makes me nervous.

philipcmonk: Because the name might change to a different format?  Is the concern with examining the name at all or more specific to this predicate?

philipcmonk: If the former, we could push the responsibility to the caller to identify which parameters are weights and which are biases.  TELinear, TERowParallelLinear, and TEColumnParallelLinear already get `module.weight_names` and `module.bias_names` from TE.  TEGroupedLinear does not, but we could scope the name inspection to that one class.

philipcmonk: [This](https://github.com/NVIDIA/Megatron-LM/commit/dbd30c1ca6664dc2e518650c1f61de741eca00fb) is what that would look like.  A bit more verbose, but maybe safer in case names change.  If an unrecognized name appears, it'll fail here and force you to label it.

deepakn94: Right, "[6:]" seems a bit clunky. Not sure if I love the alternative so maybe we just make this a regex and call it a day?

megatron/core/extensions/transformer_engine.py
```diff
@@ -86,6 +86,25 @@
 _TE_CONFIG_TYPE_KEY = "transformer_engine_config_type"
 
 
+def _set_expert_parameter_attributes(
+    module: torch.nn.Module, parallel_mode: Optional[str], use_expert_groups: bool
+) -> None:
+    """Set expert gradient-reduction and tensor-partition metadata."""
+    for name, param in module.named_parameters(recurse=False):
```
deepakn94, line 109: Can you add a comment here as to the reason behind this logic? And also add to docstring above (in particular, I'm not sure what `use_expert_groups` is)?

deepakn94: Maybe `use_edp_pgs` or something is better. Feel free to workshop.

megatron/core/optimizer/__init__.py
```diff
@@ -686,11 +685,12 @@ def init_state_fn(opt, config=None):
         setattr(optimizer, 'grad_stats_parallel_group', model_parallel_group)
 
     if pg_collection is None or not hasattr(pg_collection, 'tp'):
-        tp_group = parallel_state.get_tensor_model_parallel_group()
-    else:
-        tp_group = pg_collection.tp
-    # TODO(M4): plumb tp_group through optimizer constructors so this setattr disappears.
+        pg_collection = ProcessGroupCollection.use_mpu_process_groups()
```
deepakn94, line 688: Seems this is deprecated now? https://nvidia.slack.com/archives/C0434FDLPQV/p1784768461209949?thread_ts=1784765369.532389&cid=C0434FDLPQV

philipcmonk: I interpreted `use_mpu_process_groups` to be deprecated in the same way as `parallel_state.get_tensor_model_parallel_group`, since it's just a wrapper around that.  So this would not be a new usage but preserving existing behavior.

Eventually we should remove this remove this compatibility case altogether, but this is a really common function, and it looks like there are production uses that don't specify pg_collection, including [training.py](https://github.com/NVIDIA/Megatron-LM/blob/602fad039a9617ad3d66e1a19f0066f14ebb537c/megatron/training/training.py#L2088) and [unit tests](https://github.com/NVIDIA/Megatron-LM/blob/602fad039a9617ad3d66e1a19f0066f14ebb537c/tests/unit_tests/test_optimizer.py#L805).

megatron/core/extensions/transformer_engine.py
philipcmonk: I added a comment, hopefully it helps?  And I renamed to `use_expert_pgs` because it means we're subject to all the expert PGs: EP, ETP, and EDP.

megatron/core/extensions/transformer_engine.py
philipcmonk: I changed it to a regex

megatron/core/optimizer/__init__.py
deepakn94: Let's keep this as is, per Slack thread.

megatron/core/tensor_parallel/layers.py
```diff
@@ -89,11 +89,16 @@
     dist_reduce_scatter_func = torch.distributed._reduce_scatter_base
 
 
-def param_is_not_tensor_parallel_duplicate(param, tp_group=None):
-    """Returns true if the passed-in parameter is not a duplicate parameter
-    on another TP rank."""
+def param_is_not_tensor_parallel_duplicate(param, tp_group=None, expert_tp_group=None):
+    """Return whether a parameter contributes a unique model-parallel shard.
```
deepakn94: "contributes to a"?

megatron/core/tensor_parallel/layers.py
```diff
@@ -89,11 +89,16 @@
...
+
+    Parameters reduced over expert data parallel groups use the expert tensor-parallel
+    group for duplicate filtering. Other parameters use the regular tensor-parallel group.
+    """
     if hasattr(param, "tensor_model_parallel") and param.tensor_model_parallel:
         return True
+    if not getattr(param, "allreduce", True) and expert_tp_group is not None:
```
deepakn94, line 104: Comment for this?

megatron/core/tensor_parallel/layers.py
```diff
@@ -1269,7 +1278,11 @@ def __init__(
                 set_tensor_model_parallel_attributes(
                     tensor=self.weight, is_parallel=True, dim=1, stride=stride
                 )
-        setattr(self.weight, "allreduce", not (self.is_expert and self.expert_parallel))
+        use_expert_groups = self.is_expert and (
```
deepakn94: Should this be `use_expert_pgs` for consistency?

tests/functional_tests/test_cases/moe/gpt3_moe_mcore_te_tp4_ep2_etp2_pp2_resume_torch_dist_dist_optimizer/golden_values_dev_dgx_h100.json
```diff
@@ -7,424 +7,424 @@
...
-            "42": 10.33113,
-            "43": 10.31542,
-            "44": 10.29012,
-            "45": 10.30282,
-            "46": 10.2657,
-            "47": 10.25567,
-            "48": 10.20709,
-            "49": 10.21058,
-            "50": 10.21068,
+            "4": 10.9172,
```
deepakn94, line 10: Cool, so this is one functional test which was affected by this bug through incorrect clipping?

deepakn94 (approved):

tests/functional_tests/test_cases/moe/gpt3_moe_mcore_te_tp4_ep2_etp2_pp2_resume_torch_dist_dist_optimizer/golden_values_dev_dgx_h100.json
philipcmonk: Correct

PR #5917, opened 2026-07-21
Add context-parallel causal conv1d support

Mellonta:
## Summary

- Add a context-parallel wrapper for causal conv1d over contiguous sequence shards.
- No full activation copies are introduced.

## Tests
```
python -m torch.distributed.run --nproc-per-node 4 -m pytest -q \
    tests/unit_tests/ssm/test_causal_conv1d.py
```

### DIFF
--- a/megatron/core/ssm/causal_conv1d.py
+++ b/megatron/core/ssm/causal_conv1d.py
@@ -1,0 +1,90 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+"""Causal convolution over contiguous context-parallel sequence shards."""
+
+import torch
+
+from megatron.core.tensor_parallel.mappings import all_to_all
+
+try:
+    from causal_conv1d import causal_conv1d_fn
+except ImportError:
+    causal_conv1d_fn = None
+
+
+def _exchange_initial_states(
+    x: torch.Tensor, state_len: int, cp_group: torch.distributed.ProcessGroup
+) -> torch.Tensor | None:
+    """Exchange the preceding rank's tail as the local convolution state.
+
+    All ranks participate in a differentiable ring exchange. Rank 0 zeros the
+    wrapped tail to preserve the global causal boundary.
+    """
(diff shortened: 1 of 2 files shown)


kvareddy (approved):

megatron/core/ssm/causal_conv1d.py
```diff
@@ -0,0 +1,90 @@
...
+"""Causal convolution over contiguous context-parallel sequence shards."""
+
+import torch
+
+from megatron.core.tensor_parallel.mappings import all_to_all
+
+try:
+    from causal_conv1d import causal_conv1d_fn
+except ImportError:
+    causal_conv1d_fn = None
```
Phlip79, line 12: Why is this in a try/except?

tests/unit_tests/ssm/test_causal_conv1d.py
```diff
@@ -0,0 +1,73 @@
...
+from megatron.core import parallel_state
+from megatron.core.ssm import causal_conv1d as causal_conv1d_module
+from tests.unit_tests.test_utilities import Utils
+
+try:
+    from causal_conv1d import causal_conv1d_fn
+
+    HAVE_CAUSAL_CONV1D = True
+except ImportError:
+    HAVE_CAUSAL_CONV1D = False
```
Phlip79, line 16: Same here

megatron/core/ssm/causal_conv1d.py
Mellonta: I kept the design consistent with existing files (like megatron/core/ssm/mamba_mixer.py). I assume the original rationale was to give people the option to use Megatron's Triton implementation instead of this library.

Phlip79 (approved):

megatron/core/ssm/causal_conv1d.py
Phlip79: Got it, thanks. I don't like this but there is a precedent.

Issue #5732, opened 2026-07-09
Design Discussion For Parameter-Level Initialization/Optimization Configurability


chawkins-nvidia:
## High Level Problem Description

P0: I want to assign each parameter its own initialization and its own optimizer. 

P0: I want to be able to configure the values passed to the initialization function and optimizer. 

P1: I want the initialization function and optimizer constructor to have access to training process information like model width, model depth, parameter layer index, training duration, etc. If necessary I can supply all of this by managing my external config.

## Extended Problem Description

Principled scaling schemes such as ABC-parameterizations (muP) or norm-constrained optimizers (Muon/Shampoo/Scion/etc) have spawned large research areas to support stable and performant optimization across scales. A general trend in both optimization research and large-scale training is to enable per-parameter optimizers and per-parameter initializations. In optimal recipes, both initialization and optimizer are functions of layer type and layer characteristics. 

These recipes are hard to express in Megatron-LM because per-parameter overrides are not exposed through [OptimizerConfig](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/optimizer/optimizer_config.py#L139C7-L139C22) but rather through internal functions [see Emerging Optimizers Examples](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/megatron/core/optimizer/emerging_optimizers.py). Per-parameter overrides need to be directly configurable from training entrypoints [link](https://github.com/NVIDIA/Megatron-LM/blob/cf2f07d7b1315c96c05554c670c43207c6783e5e/pretrain_gpt.py) because Megatron-LM does not expose any detailed configuration override options for the optimizer through TransformerConfig.

The same issue is present for initialization. Granular initialization options are not accessible in `TransformerConfig` [link](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/transformer_config.py). Recent changes introduced [muP scaled inits](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/transformer_config.py#L39) which enable one scaling init recipe, but not flexible granular recipes.

See also https://github.com/NVIDIA/Megatron-LM/pull/4381 for great existing work trying to get these scaling ideas into Megatron-LM.

## Motivating Examples

[...]

PR #5925, opened 2026-07-21
Draft: Add YAML-configurable optimizer group overrides

chawkins-nvidia:
Add named optimizer instances, constructor kwargs, and parameter groups to OptimizerConfig overrides. Preserve their identities across checkpoint save/load and model-parallel rank alignment.

# What does this PR do?
Enable configurable per-parameter optimizers.

## Issue tracking

Related to: #5732 in that this provides a sample optimizer configuration approach to enable discussion.

https://github.com/NVIDIA/Megatron-LM/issues/5732

### Pre-checks

- [X] I have added relevant functional tests

## Summary

- add `OptimizerConfig.overrides_config` as a structured optimizer recipe accepting either a YAML path or an in-memory mapping
- model the recipe after quantization configuration: named optimizer instances own named `param_groups`, enabled glob `matchers` select one exact `(optimizer, param_group)` target, and an explicit `default` handles unmatched parameters
- forward arbitrary alias-level `kwargs` to emerging-optimizer constructors and arbitrary fields in each named parameter group to the optimizer parameter group
- create one runtime optimizer per alias and expert bucket, so one alias can own many parameter groups while two aliases of the same optimizer type can have different constructor arguments
- reject conflicting matcher targets, protected fields, unknown constructor kwargs, and constructor-only Muon settings placed on parameter groups
- retain the legacy standard and MuP override composition path with deprecation warnings; serialized recipes and legacy overrides are mutually exclusive
- reuse the existing emerging-optimizer construction machinery for configured Adam, Muon, SOAP, and Lion chains
- include optimizer alias, named parameter group, constructor kwargs, and group kwargs in stable checkpoint group identity and reorder matching

## Configuration Example and Explanation

```yaml
optimizers:
  adam_main: 
    type: adam
    kwargs: {}
    param_groups: 
      default:
        eps: 1.0e-12

  muon_three:
    type: muon
    kwargs:
      num_ns_steps: 3
    param_groups:
      default:
        newon_scale: 0.5
      secondary: # possible to have multiple named parameter groups under an optimizer, enables muP-style partitioning
        newon_scale: 0.6

  muon_five:
    type: muon
    kwargs:
      num_ns_steps: 5
    param_groups:
      default:
        newon_scale: 0.75

default: # need to assign a default optimizer and parameter group for parameters with no matches
  optimizer: adam_main
  param_group: default

matchers:
  fc1:
    type: glob
    pattern: "*fc1.weight"
    optimizer: muon_three
    param_group: default
    enabled: true
  fc2:
    type: glob
    pattern: "*fc2.weight"
    optimizer: muon_five
    param_group: default
    enabled: true
  fc4:
    type: glob
    pattern: "*fc4.weight"
    optimizer: muon_three
    param_group: secondary
    enabled: true
```

`muon_three` is one optimizer instance with two parameter groups. `muon_five` is a second Muon instance because its instance-wide Newton-Schulz setting differs.

Alias-level `kwargs` are construction-time settings. For a registered emerging optimizer, a newly supported constructor kwarg can therefore be configured without adding another `OptimizerConfig` field; the selected optimizer signature validates it. Parameter-group fields are written directly under `param_groups.<name>`. Muon settings that control the entire instance, such as `num_ns_steps`, `coefficient_type`, and `tp_mode`, are registered as constructor-only and produce a targeted error if placed on a parameter group. Built-in Adam and SGD continue to use their typed `OptimizerConfig` construction surface.

Multiple enabled matchers may match a parameter only if they select the same exact target. Otherwise configuration fails instead of silently composing incompatible groups.

## Checkpoint resume and model parallelism

- optimizer group identity covers every declared `ParamGroupOverride` field, including the optimizer alias, named group, and nested constructor and group kwargs
- nested mappings, sequences, sets, scalars, and `None` are canonicalized deterministically
- ordinary and distributed optimizer checkpoint matching use the same identity
- parameter groups are aligned across ranks before optimizer construction, including empty rank-local groups needed by pipeline and expert layouts
- checkpoint coverage rebuilds a four-instance Adam/Muon/Muon/Lion chain, restores model and optimizer state, verifies the nested state exactly, and performs the next update at TP1/PP1 and TP2/PP2

### DIFF
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -3,6 +3,7 @@ import copy
 import logging
 import warnings
 from collections import defaultdict
+from collections.abc import Mapping
 from dataclasses import astuple
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
@@ -51,6 +52,7 @@ from megatron.core import parallel_state
 from megatron.core.optimizer.cpu_offloading.hybrid_optimizer import HybridDeviceOptimizer
 from megatron.core.optimizer_param_scheduler import (
     ParamGroupOverride,
+    canonicalize_optimizer_config_value,
     combine_param_group_overrides,
     param_group_override_to_tuple,
 )
@@ -81,6 +83,9 @@ from .optimizer import (
 from .optimizer_config import (
     AdamOptimizerConfig,
     OptimizerConfig,
+    OptimizerInstanceConfig,
+    OptimizerOverrideRecipe,
+    OptimizerParamGroupTarget,
     ParamKey,
     ParamPredicate,
     ParamWithNamePredicate,
@@ -88,6 +93,13 @@ from .optimizer_config import (
 )
 
 logger = logging.getLogger(__name__)
+
+try:
+    import yaml
+
+    HAVE_YAML = True
+except ImportError:
+    HAVE_YAML = False
 
 
 def get_standard_config_overrides(config: OptimizerConfig) -> Dict[ParamKey, ParamGroupOverride]:
@@ -99,6 +111,13 @@ def get_standard_config_overrides(config: OptimizerConfig) -> Dict[ParamKey, Par
     Returns:
         Dict[ParamKey, ParamGroupOverride]: standard config overrides.
     """
+    warnings.warn(
+        "get_standard_config_overrides is deprecated and superseded by "
+        "OptimizerConfig.overrides_config.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+
     config_overrides: Optional[Dict[ParamKey, ParamGroupOverride]] = {}
     # First, figure out how we are going to do wd skipping. The two main approaches are:
     #  1. The classic megatron approach of skipping all len 1 and bias parameters.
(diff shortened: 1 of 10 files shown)


megatron/core/optimizer/__init__.py
```diff
@@ -128,6 +147,245 @@ def get_standard_config_overrides(config: OptimizerConfig) -> Dict[ParamKey, Par
...
+    if unknown_recipe_fields:
+        raise ValueError(
+            f"overrides_config has unsupported fields: {sorted(unknown_recipe_fields)}"
+        )
+
+    optimizers = as_mapping(recipe.get("optimizers", {}), "overrides_config.optimizers")
+    if not optimizers:
+        raise ValueError("overrides_config.optimizers must not be empty")
+    matchers = as_mapping(recipe.get("matchers", {}), "overrides_config.matchers")
+    protected_param_group_fields = {
```
chawkins-nvidia, line 249: Capitalize this constant

megatron/core/optimizer/optimizer_config.py
```diff
@@ -135,6 +136,78 @@ def matches(self, param: torch.nn.Parameter, param_name: str) -> bool:
...
+
+    optimizers: Dict[str, OptimizerInstanceConfig]
+    matchers: Dict[ParamKey, OptimizerParamGroupTarget]
+    default: OptimizerParamGroupTarget
+
+    def resolve(self, target: OptimizerParamGroupTarget) -> Dict[str, Any]:
+        """Return checkpoint-stable metadata for a configured parameter group."""
+        optimizer = self.optimizers[target.optimizer]
+        raw_group = optimizer.param_groups[target.param_group]
+        named_override_fields = {
```
chawkins-nvidia, line 168: Capitalize constant

megatron/training/training.py
```diff
@@ -1950,12 +2061,17 @@ def get_megatron_optimizer_config(args: Any) -> OptimizerConfig:
...
-    # Construct the appropriate config_overrides object. This default handles many cases, but
-    #  can be added to as needed by the user, or replaced entirely with a custom override.
-    config_overrides = get_standard_config_overrides(config=config)
+    # The serialized recipe and legacy standard overrides are mutually exclusive. Leave recipe
+    # parsing to get_megatron_optimizer so passing both APIs can be detected there.
+    config_overrides = (
+        None
+        if config.overrides_config is not None
+        else get_standard_config_overrides(config=config)
+    )
```
janEbert, line 2070: This is the only "real" change in the file, right? Or is there another place here that you modified semantically?

megatron/core/optimizer/__init__.py
```diff
@@ -128,6 +147,245 @@ def get_standard_config_overrides(config: OptimizerConfig) -> Dict[ParamKey, Par
...
+    Returns:
+        Union[Dict[ParamKey, ParamGroupOverride], OptimizerOverrideRecipe]: Legacy programmatic
+            overrides or the parsed optimizer recipe.
+    """
+
+    def as_mapping(value: Any, path: str) -> Mapping[str, Any]:
+        if isinstance(value, Mapping):
+            return value
+        if hasattr(value, "__dict__"):
+            return vars(value)
```
janEbert, line 209: Could also add an additional case for `value.__slots__`.

chawkins-nvidia:
Tagging @NVIDIA/mcore-emerging-optimizers for input

PR #5928, opened 2026-07-21
fix(moe): exclude padding tokens from dropless HybridEP routing

FDecaYed:
## What

Exclude padding-token rows from the routing probabilities and sparse routing map only for dropless HybridEP. Dispatchers that still require fixed route counts keep their existing behavior.

This is a self-contained correctness prerequisite for the HybridModel-only DSv4 merge effort.

## Provenance

- Refactors the DSv4 feature set captured in #5795.
- Ports the focused fix from #5542 to main.
- Original implementation by @xiaoyao0115; DSv4 integration work by @hxbai and the source authors.
- GPTModel integration is intentionally out of scope.

## Testing

- uv run isort --check-only megatron/core/transformer/moe/router.py tests/unit_tests/transformer/moe/test_routers.py
- uv run ruff check megatron/core/transformer/moe/router.py tests/unit_tests/transformer/moe/test_routers.py
- uv run python -m compileall -q megatron/core/transformer/moe/router.py tests/unit_tests/transformer/moe/test_routers.py
- CUDA unit tests require the project CUDA/Triton environment and are left to CI.

### DIFF
--- a/megatron/core/transformer/moe/router.py
+++ b/megatron/core/transformer/moe/router.py
@@ -784,6 +784,20 @@ class TopKRouter(Router):
                 router_replay=self.router_replay,
             )
 
+        # Dropless HybridEP consumes the sparse routing map directly, so exclude padding
+        # rows before dispatch. Other dispatchers retain their existing fixed-route
+        # assumptions until they support sparse routing maps end to end.
+        use_dropless_hybridep = (
+            self.config.moe_token_dispatcher_type == "flex"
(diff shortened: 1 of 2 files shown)


FDecaYed:
Closing this draft until its real dependency round. Its logical successor is P10 (HybridEP packed-token correctness), to be rebuilt after #5008, P4, and P7 merge. The publication diff will use the current #5008 flag name and will reference tracking PR #5795.

PR #5929, opened 2026-07-21
Selective-FP32 runtime and native-FP32 optimizer checkpoint lifecycle

FDecaYed:
# Selective-FP32 runtime and native-FP32 optimizer checkpoint lifecycle

## Summary

Preserve explicitly marked FP32 parameters while converting the rest of a
module, provide an FP8-disabled execution context, and make optimizer
checkpoint mapping correct for interleaved low-precision and native-FP32
parameters.

## Scope and non-goals

- Add the selective-FP32 marking, conversion, and execution primitives.
- Preserve native-FP32 parameters in optimizer sharded state dictionaries.
- Do not add mHC, DSv4 attention, model construction, or recipe wiring.

## Provenance

This local recut is reconstructed from the frozen `main` baseline
`bb5647a9bdd0`, not a replay of historical commits. It references the DSv4
tracking PR #5795 and consolidates the complete contracts from #5929 and
#5930. Original DSv4 integration credit: @hxbai in #5795.

## Dependencies

None beyond the frozen baseline used for this local recut. This is a reusable
prerequisite outside the nine-PR DSv4-specific series.

## Tests

- Selective marking, FP16/BF16 conversion, and the FP8-disabled context.
- Interleaved BF16/FP32/BF16 optimizer-state mapping and Adam state keys.

## Publication

Publish in the first prerequisite round, rebased onto then-current `main`, so
the GitHub changeset contains only this slice. This is not a stacked-review PR.

### DIFF
--- a/megatron/core/optimizer/optimizer.py
+++ b/megatron/core/optimizer/optimizer.py
@@ -7,7 +7,6 @@ import logging
 import math
 import warnings
 from abc import ABC, abstractmethod
-from itertools import chain
 from logging import getLogger
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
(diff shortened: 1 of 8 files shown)


Phlip79 (approved):

megatron/core/optimizer/optimizer.py
```diff
@@ -1019,25 +1019,51 @@ def sharded_state_dict(
...
+        # param in that order so the ids line up even when both kinds are present.
+        native_fp32_param_ids = {
+            id(param) for param in chain.from_iterable(self.fp32_from_fp32_groups)
+        }
+
+        def model_params_in_optimizer_order():
+            for inner_group, float16_group in zip(self.optimizer.param_groups, self.float16_groups):
+                float16_params = iter(float16_group)
+                for param in inner_group['params']:
+                    yield (param if id(param) in native_fp32_param_ids else next(float16_params))
```
janEbert: The inner for-loop needs to guard each iteration on whether `param.requires_grad`, according to the logic that assigns to `self.float16_groups` and `self.fp32_from_fp32_groups`, right? Otherwise we may iterate `float16_params` too early and for unrelated parameters.
I think the design is a bit fragile with the `zip` potentially ending early and the `next` throwing a hard-to-understand error.

janEbert (changes requested):

megatron/core/optimizer/optimizer.py
FDecaYed: `requires_grad` guard would still be unsafe here because the FP32 main copies are detached and therefore also have `requires_grad=False`.

I replaced the positional iterator with an explicit FP32-main-parameter-ID to model-parameter map. Frozen and native-FP32 parameters remain mapped to themselves, while the `fp32_from_fp16` checkpoint IDs are now restricted to actual main copies. I also added a regression covering interleaved trainable BF16 / frozen BF16 / native FP32 / trainable BF16 parameters and verifying the resulting optimizer state IDs.

janEbert: Thanks! Also, adding `strict=True` is great.

janEbert (approved):

asolergi-nv (approved):

deepakn94 (approved):

PR #4894, opened 2026-05-20
[dev] [DeepSeek-v4] Part 4: Fusion Kernels for DSv4 Hybrid Attention

hxbai:
# What does this PR do ?
https://github.com/NVIDIA/cudnn-frontend release v1.24.0 added the fusion kernels for DeepSeek-V4 Hybrid Attention.

This PR integrated these kernel into MCore.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact @NVIDIA/mcore-oncall.

## Issue tracking

For PRs from open-source community contributors:

- **New features**: a linked issue is **required**. Please open a [feature request](https://github.com/NVIDIA/Megatron-LM/issues/new?template=feature_request.md) and reference it here before submitting the PR.
- **Small updates (bug fixes, minor improvements)**: a linked issue is **recommended** and will accelerate the PR review process.

Linked issue: 

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment @NVIDIA/mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

Once all required reviewers have approved, the `Approved` label is applied **automatically**.

### Merge

Any member of [mcore-engineers](https://github.com/orgs/NVIDIA/teams/mcore-engineers) will be able to merge your PR.

<details>
[...]

PR #5930, opened 2026-07-21
fix(optimizer): checkpoint native FP32 parameters in mixed precision

FDecaYed:
## What

Fix sharded optimizer-state mapping when mixed-precision parameter groups interleave BF16/FP16 parameters with parameters that remain natively FP32. State ids now follow the inner optimizer order, while FP32 master copies are paired only with converted low-precision parameters.

This is a generic checkpointing correctness fix and is detached from the DSv4 stack. It does not depend on the parameter-marking PR; DSv4 call sites can consume both independently.

## Provenance

- Extracted from the high-precision DSv4 follow-up in #5795 (commit 081eebc4).
- The eventual precision-sensitive consumers originate in the CSA work from #4458 and #4894.
- Original DSv4 integration by @hxbai and the source authors.

## Testing

- uv run isort --check-only on both changed Python files
- uv run ruff check on both changed Python files
- uv run python -m compileall -q on both changed Python files
- Focused CUDA checkpoint test is included; local execution requires the project CUDA/Torch environment and is left to CI.

### DIFF
--- a/megatron/core/optimizer/optimizer.py
+++ b/megatron/core/optimizer/optimizer.py
@@ -7,7 +7,6 @@ import logging
 import math
 import warnings
 from abc import ABC, abstractmethod
-from itertools import chain
 from logging import getLogger
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


FDecaYed:
Closing this standalone boundary because it is absorbed into P1. P1 combines selective-FP32 runtime and native-FP32 optimizer checkpoint correctness and will be published by updating #5929. Tracking PR: #5795.

PR #2811, opened 2026-01-05
Optimizer State and Master Weight Offloading

hxbai:
# What does this PR do ?
dev branch: #2760 

### Summary

This PR introduces optimizer state offloading to CPU for the `DistributedOptimizer`, enabling significant GPU memory savings during training by temporarily moving optimizer states (exp_avg, exp_avg_sq) and master weights to CPU memory when not in use.

### Motivation

During the forward and backward passes, optimizer states occupy GPU memory but are not actively used. For large models, these states can consume a substantial portion of GPU memory. By offloading optimizer states to CPU after `optimizer.step()` and reloading them before the next step, we can reclaim this GPU memory for other operations like activation checkpointing or larger batch sizes.

### Performance & Memory Savings

**Memory Savings:**
- On **DeepSeek-V3**, this feature saves **15-20GB** of GPU memory (with 0.1-0.2s/iter overhead on GB200)

**Comparison with Optimizer CPU Offloading:**

| Aspect | CPU Offloading (ZeRO-Offload style) | State Offloading (This PR) |
|--------|-------------------------------------|----------------------------|
| **Where optimizer runs** | CPU | GPU |
| **D2H/H2D frequency** | Every step (gradients + params) | Every step (states only) |
| **Compute location** | Adam step on CPU | Adam step on GPU |
| **Best for** | Memory-constrained, bandwidth-limited | High-bandwidth interconnects (NVLink, GB200) |

More details:
- With higher H2D/D2H bandwidth such as **GB200**, the state offloading has significantly less overhead
- Async transfer overlapping
- Optimizer step still runs on GPU, avoiding CPU compute bottleneck
- Pinned memory enables maximum PCIe/NVLink bandwidth utilization
- Currently requires TE FusedAdam optimizer

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @megatron-oncall.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
        C1[Expert Review] --> C2[Final Review]
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [ ] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
[...]

PR #5933, opened 2026-07-21
feat(optimizer): add distributed optimizer state offloader

FDecaYed:
## What

Add the reusable distributed-optimizer state offloader:

- asynchronously copy TE FusedAdam moments and primary weights to pinned CPU buffers;
- release and reallocate GPU storage around the offload window;
- reload on a dedicated CUDA stream and synchronize before the optimizer step;
- cover initialization, early primary-weight offload, repeated cycles, memory release, and numerical parity.

This PR contains only the optimizer-side implementation. A small stacked follow-up owns the training CLI and lifecycle call sites, so optimizer reviewers do not need to review training-loop changes here. This component is otherwise independent of DSv4.

## Provenance

- Refreshes the dev implementation from #2987 and its earlier main port #2811.
- Refactors the same feature captured in #5795 (commit 152877ba).
- Original implementation by @hxbai, @yaox12, and the source authors.
- GPTModel is unrelated and intentionally absent.

## Testing

- uv run isort --check-only on all changed Python files
- uv run ruff check on all changed Python files
- uv run python -m compileall -q on all changed Python files
- Focused CUDA tests are included; local execution requires TE FusedAdam and CUDA and is left to CI.

## Stack follow-up

- #5934 contains the next logical slice and temporarily targets this PR mirror.

### DIFF
--- a/megatron/core/optimizer/cpu_offloading/README.md
+++ b/megatron/core/optimizer/cpu_offloading/README.md
@@ -1,6 +1,10 @@
-## How to use ?
+# Optimizer CPU Offloading
 
-Add these flags to enable optimizer cpu offload in MCore.
+MCore provides two mutually exclusive options for reducing optimizer GPU memory usage by leveraging CPU memory. They cannot be combined.
+
+## Option 1: `--optimizer-cpu-offload`
+
+Runs the entire optimizer step on CPU. Gradients are copied to CPU, the optimizer updates parameters there, and the updated parameters are copied back to GPU.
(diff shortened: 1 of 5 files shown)


FDecaYed:
Closing this draft outside the DSv4 merge path. If optimizer-state offload is pursued, this PR can be reopened and rebuilt as one independent end-to-end feature absorbing #5934, with lifecycle tests. It is not a dependency or child of #5795.

PR #5935, opened 2026-07-21
feat(recompute): coordinate checkpoint-without-output recomputation

FDecaYed:
## What

Extend CheckpointWithoutOutput so it can safely coordinate several sequential checkpointed operations:

- preserve mixed tensor and non-tensor arguments through recompute;
- return None gradients for metadata arguments;
- add a CheckpointManager that discards all managed outputs and registers one ordered recompute hook;
- preserve CUDA-graph capture behavior;
- add focused unit coverage for argument restoration, hook ordering, storage reuse, gradients, and edge cases.

This reusable recompute primitive is the root of the HybridModel mHC lane, but contains no mHC, model, or GPTModel code itself.

## Provenance

- Reconstructs the checkpoint-management portion of the mHC work from #2943 and the current main proposal #4531.
- Refactors the same implementation captured in #5795 (commit 07c0d088).
- Original implementation by the #2943/#4531 authors; DSv4 adaptation by @hxbai.

## Testing

- uv run isort --check-only on both changed Python files
- uv run ruff check on both changed Python files
- uv run python -m compileall -q on both changed Python files
- Focused tests are included; local execution requires the project Torch runtime and is left to CI.

## Stack follow-up

- #5936 contains the next logical slice and temporarily targets this PR mirror.

### DIFF
--- a/megatron/core/tensor_parallel/random.py
+++ b/megatron/core/tensor_parallel/random.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
 # Parts of the code here are adapted from PyTorch
 # repo: https://github.com/pytorch/pytorch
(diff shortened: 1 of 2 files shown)


FDecaYed:
Closing this draft because the generic checkpoint-without-output/mHC contract is supplied by external prerequisite #4531. The later P2 PR will be a focused HybridModel/fused-mHC delta after #4531 and P1 merge. Tracking PR: #5795.

PR #5939, opened 2026-07-21
Extend dynamic inference asynchronous scheduling support

lmcafee-nvidia:
## Summary

Extend dynamic inference asynchronous scheduling across the remaining request and output features needed by the primary inference path. The controller keeps one pending forward while CPU request work is reordered around GPU sampling and forward execution.

## Architecture

- Route asynchronous iterations through phase orders matched to the current work:
  - no-overlap: sample or verify the previous output, resolve lifecycle changes, prepare and admit requests, then launch the next forward
  - overlap: prepare the next decode step, sample the previous output, launch the next forward, then resolve the previous output
  - MTP overlap: verify drafts, rewind rejected state, prepare, launch the next forward, then resolve
- Keep the first admitted request as a primer-only forward and preserve one base-model forward per controller iteration.
- Keep expert-parallel ranks in a consistent MTP-before-base-forward order, including ranks with dummy work.
- Add single-bank Mamba handling by deferring admission until pending state is safe to reuse.
- Support chunked prefill and KV/Mamba prefix caching within asynchronous scheduling.
- Use the existing sampling implementations for greedy and non-greedy sampling with Torch and FlashInfer backends.
- Preserve sampled log probabilities, top-n log probabilities, and stop-word completion behavior across request compaction.
- Expose exactly two scheduling modes: `legacy` and `async`.

Routing replay remains explicitly unsupported by this PR.


sidsingh-nvidia:
The name `_run_async_sched_prefill_step` is misleading (controller L2437). It can fire for both prefills and decodes 
1. admit a new request - actual prefill;
2. transition an in-flight prefill → decode (num_prefill_requests > 0 is set from the previous prefill step) — produces decode output;
3. cold start (not _async_sched_logits.is_valid) - launches a primer (can be a prefill)

The real common thread is that all three flush the in-flight forward and issue the next one after resolve/admit.

Recommendation.

1. Rename the schedule to reflect what it is. I recommend using `async_sched_flush` since the one invariant in this schedule is that we flush out all GPU work from the past turn by sampling and resolving it. 
2. Suggest breaking down `_run_async_sched_prefill_step` into two methods, which can be called from the engine. 
  - `async_sched_flush_and_prepare() `— L2459–2497 (sample/MTP → resolve → prepare/commit); returns the resolved output.
  - `async_sched_post_flush_forward() `— L2506–2526 (issue exactly one base forward: real / primer / dummy).

Engine flow becomes `async_sched_flush_and_prepare()` →  `admit()` → `async_sched_post_flush_forward()`. This removes the current callback inversion where schedule_waiting_requests is passed into the controller and called back out.
3. Unify the dummy forward. Route both dummy cases through a single controller dummy-forward primitive called from the engine, with an arg (e.g. run_mtp_dummy: bool) to skip the MTP dummy:
  - flush schedule, drained-to-zero → run_mtp_dummy=False (the drain already issued the real MTP collectives this iteration, so only the base half is left);
  - idle rank (local_pending == 0) → run_mtp_dummy=True (no drain ran, so both MTP and base must be faked).
  
  I have left a separate comment about the "5 counter restore hack", which we can avoid with a simple fix.

megatron/core/inference/text_generation_controllers/text_generation_controller.py
```diff
@@ -1996,12 +2270,47 @@ def _run_async_sched_forward(
...
         return forward_done_event
 
+    def _run_dummy_async_sched_base_step(self) -> Optional[torch.cuda.Event]:
+        """Run the base-forward half of an async EP step after local work finishes.
+
+        Returns:
+            Optional[torch.cuda.Event]: Event marking dummy base-forward completion.
+        """
+        context = self.inference_wrapped_model.inference_context
+        saved_counters = (
```
sidsingh-nvidia: This save/restore is a workaround: reset(preserve_prefix_cache=True) only preserves the counters when prefix caching is on.  The idle-rank dummy_forward hits the same bug silently (resets step_count despite its "keep monotonic for logging" comment). 

Fix at the root: a preserve_counters flag for context.reset that clears everything except the step counters and maybe also the prefix cache?. This will eliminate the need for this hacky restore.  After that we should unify the dummy forwards into a single function again, with a skip_mtp flag for the flusing schedule.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1624,10 +1623,48 @@ def schedule_waiting_requests(self):
...
+            return self._cg_admission_check(req, candidate)
+        return self._matches_cg_admission(candidate)
+
+    def _has_async_sched_prefill_work(self) -> bool:
+        """Return whether this step should use async prefill ordering.
+
+        Returns:
+            bool: Whether prefill is active or the queue head is currently admissible.
         """
+        if self.context.num_prefill_requests > 0:
```
sidsingh-nvidia: Just confirming that this condition is hit when we are transitioning from prefill -> decode, right? If so can we leave a comment here?

sidsingh-nvidia (changes requested):

megatron/core/inference/text_generation_controllers/text_generation_controller.py
lmcafee-nvidia: Addressed the counter issue at the root. `DynamicInferenceContext.reset()` now has an explicit `preserve_counters` option independent of prefix-cache preservation, both dummy reset paths use it, and the manual save/restore workaround is removed. The dummy-forward unification is intentionally left for a separate change.

megatron/core/inference/engines/dynamic_engine.py
lmcafee-nvidia: Added a comment at the selector. It covers the prefill-to-decode transition and the other cases that require complete lifecycle bookkeeping: paused requests and insufficient KV capacity.

lmcafee-nvidia:
@sidsingh-nvidia Thanks for the detailed review.

The naming concern is addressed by renaming the schedules according to their actual execution property: `no_overlap` and `overlap`. The no-overlap path also handles lifecycle fallbacks beyond prefill, so I avoided `flush`; it consumes the pending output but launches the successor forward without globally flushing GPU work.

I'd prefer to keep the phase orchestration in the controller. Sampling/MTP, pending-logit state, request resolution, events, and forward submission constitute one controller step. Splitting that step around engine admission would expose partially completed controller state across calls and make the ordering contract harder to preserve. The synchronous admission callback keeps engine-owned queue mutation at one explicit point without moving controller state into the engine.

I agree that the two async dummy cases can share one controller primitive. I'll consolidate those using a `run_mtp_dummy` argument while preserving the existing phase ordering.

The counter workaround mentioned at the end has also been removed in `7164358b4`: counter preservation is now explicit and independent of prefix-cache preservation.

sidsingh-nvidia:
Follow ups -
1. It looks like the non-overlap schedule is triggered for decodes too - specifically decode steps that follow right after prefill. Can we put this in comments in the code.
2. A corollary of 1 is that logging of is_decode_only is broken - need to make a call on whether to fix it in this PR or not
3. We need to confirm that async scheduling does not break context.is_decode_only() - as it is used in many places in the code.

sidsingh-nvidia (approved):
Approving with minor nit - we should explicitly print whether a given step is prefill or decode in the logs. It would be unintuitive to parse from the tupled states.

kvareddy (approved):

megatron/training/arguments.py
```diff
@@ -2012,13 +2012,11 @@ def _add_inference_args(parser):
                             'is requested but the package is not installed.')
     group.add_argument('--inference-dynamic-batching-async-sched-mode',
                        type=str, default='legacy',
-                       choices=['legacy', 'serial', 'overlap'],
```
jiemingz, line 2150: technically this is not strictly backwards compatible, though im not sure how much we care about back compat for inference.

jiemingz (approved):

megatron/training/arguments.py
asolergi-nv: Yes, we should at least leave a comment. Is `async` a (improved) rename of `overlap`?

asolergi-nv (approved):

shanmugamr1992 (approved):

megatron/training/arguments.py
lmcafee-nvidia: yes, `async` is a rename of `overlap`.

and @jiemingz , I believe inference has not being enforcing backwards compatibility for a while. also, all of this async stuff has been going in in the last couple weeks, so it's relatively short-lived.

PR #5938, opened 2026-07-21
feat(moe): add hash-based token routing

FDecaYed:
## What

Add model-neutral Hash MoE routing:

- persistent token-to-expert lookup tables for the leading MoE layers;
- softmax, sigmoid, and sqrt-softplus combination scores;
- forced benchmark routing override and inference fallback;
- input_ids plumbing through TransformerBlock, TransformerLayer, recompute, partial CUDA graphs, and MoELayer;
- validation for vocabulary size, PP placement, MTP exclusion, and unsupported EP overlap;
- focused router and real TransformerLayer/MoE coverage.

This is an independent root. It intentionally stops before HybridModel/HybridStack, training CLI, attention, Clamp/SwiGLU, mHC, and recipe wiring.

## Stack

- #5945 passes token IDs through HybridModel to hash-routed TransformerLayer instances.
- A separate small training/config bridge supplies the padded vocabulary size without adding a duplicate CLI.

## Provenance

- Reconstructs the Hash routing portion of #4481.
- Includes the forced-routing correction from #5130.
- Refactors the same work captured in #5795 (commits 156bb6cd and 21944429).
- Original implementation by the #4481/#5130 authors; DSv4 adaptation by @hxbai.
- Reconstructed against current main; feature cutoff: 2026-07-21.
- GPTModel integration is intentionally excluded.

## Testing

- isort --check-only, ruff, Black --check, Python compilation, and git diff --check passed for the slice.
- The 8-rank focused pytest was attempted locally; collection is blocked on this macOS host by missing Triton/CUDA (CUDA device count is zero), not by a test assertion.

PR #5945, opened 2026-07-21
feat(hybrid): pass token ids to hash-routed layers

FDecaYed:
## What

Teach HybridModel/HybridStack to supply token IDs only to TransformerLayer consumers that use Hash MoE routing.

- pass `input_ids` through the HybridModel forward path;
- preserve the strict call signature for non-transformer layers;
- keep selective/full recompute and local/TE CUDA-graph execution covered;
- add focused HybridModel coverage without introducing GPTModel support.

This is a child of #5938. Its GitHub Files changed view contains only the HybridModel adapter and its focused test.

## Provenance

- Reconstructs the HybridModel integration portion of #4481.
- Refactors the same frozen feature set captured in #5795.
- Original Hash routing implementation by the #4481 authors; DSv4 adaptation by @hxbai.
- Reconstructed against the #5938 merge mirror; feature cutoff: 2026-07-21.

## Testing

- isort --check-only, ruff, Black --check, Python compilation, and git diff --check passed for the slice.
- Focused CUDA/Torch execution is delegated to PR CI because this local macOS host has no CUDA devices.

### DIFF
--- a/megatron/core/models/hybrid/hybrid_block.py
+++ b/megatron/core/models/hybrid/hybrid_block.py
@@ -232,6 +232,7 @@ class HybridStack(MegatronModule):
         inference_params: Optional[BaseInferenceContext] = None,
         packed_seq_params: Optional[PackedSeqParams] = None,
         padding_mask=None,
+        input_ids: Optional[Tensor] = None,
     ):
         """
         Forward function of the HybridStack class.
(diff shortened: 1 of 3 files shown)


FDecaYed:
Closing this stacked adapter because it is absorbed into P4: one complete HybridModel-only Hash MoE feature rebuilt through #5938 after P2 merges. Tracking PR: #5795.

PR #5949, opened 2026-07-21
Move FSDP model weight sync to optimizer post-step

wujingyue:
## Summary

- Move experimental FSDP main-weight to model-weight synchronization from forward hooks to the optimizer post-step hook, so it runs once per optimizer step instead of once per microbatch.
- Add `precision_aware=True` for optimizers, such as FusedAdam, that accept mixed-precision gradients without temporary casts.
- Initialize compute weights at FSDP setup and add multi-step synchronization coverage.

## Validation

- `python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/mfsdp_v2/test_optimizer.py`
- Focused FSDP post-step synchronization, next-forward freshness, and mixed-precision Adam parity tests.
- Black, isort, Ruff, and `git diff --check`.

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
@@ -68,7 +68,11 @@ def fully_shard(
 
 @contextmanager
 def microbatch(module: nn.Module, is_last: bool) -> Iterator[None]:
-    """Scope FSDP state to one microbatch.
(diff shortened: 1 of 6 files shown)


shjwudp (approved):

wujingyue (commented):
From @shjwudp : for drop-in replacement and potentially CPU offloading, we may want to sync in hooks instead of post optimizer step. Revisit this when there's a clearer use case.

Phlip79 (approved):

jiemingz (approved):

PR #5950, opened 2026-07-21
Fix inference overflow into dummy blocks

tdene:

# What does this PR do?

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -2355,10 +2355,10 @@ def initialize_attention_state(
                 self._cpu_mha_cu_kv_seq_lengths[real_bs]
             )
 
-        # Block table: [0:real_bs] real, [real_bs:padded_bs] = -1 sentinel.
+        # Block table: [0:real_bs] real, [real_bs:padded_bs] = dummy block.
         self._cpu_mha_block_table[:real_bs] = request_to_kv_block_ids_view[:real_bs]
         if real_bs < padded_bs:
-            self._cpu_mha_block_table[real_bs:padded_bs] = -1
+            self._cpu_mha_block_table[real_bs:padded_bs] = self.kv_block_allocator.dummy_block_idx
```
tdene, line 2369: `-1` and `dummy_block_idx` are equivalent for book-keeping, but not for the kernels.

sidsingh-nvidia: With -1s, do we write into incorrect blocks?

tdene: The kernel reads the "-1" as a literal value, instead of a relative list index, and tries to access a memory location using that -1 in the formula.

Sometimes that memory location works and we access a random unintended location. Other times it IMAs.

That's what we saw in MRL at least.

sidsingh-nvidia (approved):

megatron/core/inference/contexts/dynamic_context.py
sidsingh-nvidia: Do we need a similar fix for mamba?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -577,6 +577,20 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
...
+            self.kv_block_allocator.active_count < min_usable_block_count
+            and torch.distributed.get_rank() == 0
+        ):
+            warnings.warn(
+                f"The active KV block pool ({self.kv_block_allocator.active_count} blocks) "
+                f"is too small for one request generating up to the engine's sequence "
+                f"budget ({min_usable_block_count} blocks for max_sequence_length="
+                f"{self.max_sequence_length}); such a request would pause forever. "
+                f"Increase buffer_size_gb or reduce max_sequence_length."
+            )
```
santhnm2: I think we should throw an error here instead of a warning

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -577,6 +577,20 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
         if self.num_speculative_tokens > 0:
             self.max_kv_block_count += 1
 
+        # The pool must be sufficient to hold `max_sequence_length` + 1 blocks.
+        min_usable_block_count = max((self.max_sequence_length - 1) // self.block_size_tokens, 1)
```
santhnm2: Why isn't this `(self.max_sequence_length // self.block_size_tokens) - 1`? In other words, why is the -1 coming before the `// self.block_size_tokens` instead of after? Also why do we need the max with 1?

megatron/core/inference/contexts/dynamic_context.py
santhnm2: Mamba doesn't have a dummy block index, we just modified the kernel to ignore padding values (specified by -1)

megatron/core/inference/contexts/dynamic_context.py
tdene: I've reworked the logic so that this is no longer relevant.

megatron/core/inference/contexts/dynamic_context.py
tdene: Take a look now.

I don't know why I made the earlier version of the PR so complicated. I took weeks-old rebased code, corrected it five times, and it ended up being a mess.

The original version of this branch was solving a similar issue to #4114. But #4114 has since been merged, so most of that logic is not necessary today.

The current version is really all we need.  We check against the active pool (which correctly does not include the dummy block). That prevents issues if we have a non-0 paused allocation.

santhnm2 (approved):

shanmugamr1992 (approved):

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -2363,10 +2363,10 @@ def initialize_attention_state(
                 self._cpu_mha_cu_kv_seq_lengths[real_bs]
             )
 
-        # Block table: [0:real_bs] real, [real_bs:padded_bs] = -1 sentinel.
+        # Block table: [0:real_bs] real, [real_bs:padded_bs] = dummy block.
         self._cpu_mha_block_table[:real_bs] = request_to_kv_block_ids_view[:real_bs]
         if real_bs < padded_bs:
-            self._cpu_mha_block_table[real_bs:padded_bs] = -1
+            self._cpu_mha_block_table[real_bs:padded_bs] = self.kv_block_allocator.dummy_block_idx
```
lmcafee-nvidia, line 2369: could we add a direct test for the safe padding value? for example, run one real request in a four-request graph bucket and check that the other three block_table rows contain `dummy_block_idx` instead of -1. this would catch the exact unsafe value this line is fixing.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -1121,12 +1121,15 @@ def _add_request(
             request.add_event_error_nontransient(TokenOverflowError(request_id))
 
         # Check that the KV cache has enough blocks for this request's max sequence length.
+        # Blocks are granted to a running request only from the active pool.
         max_request_tokens = (
             len(request.prompt_tokens) + request.sampling_params.num_tokens_to_generate
         )
+        if request.sampling_params.num_tokens_to_generate > 0:
+            # Decode pre-allocates its next block within num_speculative_tokens of a boundary.
+            max_request_tokens += self.context.num_speculative_tokens
```
lmcafee-nvidia: could we count only the tokens that are actually stored? for example, with an 8-token prompt, 247 output tokens, two speculative tokens, and one 256-token block, the model stores 8 + 246 + 2 = 256 tokens; the 247th output token is produced afterward and is not stored. this code counts 8 + 247 + 2 = 257 and rejects a request that fits, so could we fix the calculation and make this boundary test run to completion?

megatron/core/inference/contexts/dynamic_context.py
tdene: Done!

megatron/core/inference/engines/dynamic_engine.py
tdene: Good catch, thanks! I should be doing this now.

lmcafee-nvidia (approved):

Issue #5656, opened 2026-07-05
MFSDP v2 composability with expert parallelism


wujingyue:
MFSDP v2 should be composable with expert parallelism (EP).

For MoE layers, expert parameters are EP-owned and commonly also fully sharded. In the current non-experimental Megatron-FSDP path, EP composability is supported by representing each local expert's weight as a separate parameter. FSDP then shards that expert's 2D weight tensor, rather than sharding a grouped expert axis.

One example setup is 128 GPUs with:
- expert parallelism: EP=64
- fully sharded data parallelism: FSDP=2

In this example, each EP rank owns the experts assigned to that EP rank, and the FSDP group shards each local expert's parameters. If an expert weight is represented as a 2D tensor `[h_in, h_out]` or `[h_out, h_in]`, FSDP can shard dimension 0 of that per-expert tensor. This avoids treating the EP-local expert axis itself as the FSDP shard axis.

An open design question for MFSDP v2 is whether to keep the current per-expert-parameter representation or move toward grouped expert weights represented as a single contiguous 3D tensor, e.g. `[E_local, h_in, h_out]` or `[E_local, h_out, h_in]`. There is a general performance preference for the grouped contiguous representation: TransformerEngine engineers have found it tricky to guarantee that multiple independent `torch.Parameter` objects are contiguous in memory, which disables or complicates some performance optimizations.

wujingyue:
I'll start with TE's single-grouped expert weights and with FSDP=2,EP=4 config. Will update how it goes...

PR #5956, opened 2026-07-22
Divide expert gradients by the expert-parallel size

wujingyue:
Addresses #5656.

## What

Megatron-FSDP v2's `fully_shard` does **not** compose correctly with expert parallelism today: expert gradients come out `ep_size` times too large. This adds a `grad_divisor` argument that fixes it, plus the unit test that found it.

> An earlier revision of this PR claimed no library change was needed. That was wrong — see the measurements below.

## Why expert gradients are over-counted

A rank's experts process tokens routed to them from every rank in the expert-parallel group, and the backward pass routes those tokens' gradients back. So a rank's expert gradient already sums over `ep_size` ranks' data before any collective runs.

mFSDP then reduce-scatters with `ReduceOp.AVG` over the mesh it was handed. For experts that mesh is expert-DP (size `dp_size`), so it divides by `dp_size` when `world_size` ranks contributed — leaving expert gradients `ep_size` times too large. Dense parameters are unaffected: no alltoall, one contribution per mesh rank, and the full DP mesh is already the right size.

MCore DDP applies exactly this correction as `expert_gradient_scaling_factor = expt_dp / dp` (`distributed_data_parallel.py:208`), and Megatron-FSDP v1 carries it end to end. v2 had no way to express it, because the reduce op is chosen internally.

## Test

`tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py` checks that an `EP=4` MoE `HybridModel` sharded with mFSDP v2 — experts over the expert-DP sub-mesh, dense params over the full DP mesh — consuming its `1/dp` shard of a global batch reproduces a single full-batch `EP=1` reference.

The reference has no distributed logic at all: it runs the whole batch on every rank with size-1 process groups, so its gradients are the exact full-batch gradients and need no reduction. That makes it an independent oracle for the alltoall dispatch, the sharding, and the gradient reduction together.

On 8×H100 (`EP=4`, `EDP=2`, `TP=PP=CP=1`) the model matches the reference to `max_abs=1.192093e-07` — `2^-23`, one ULP of float32. With `grad_divisor=1`, i.e. the pre-fix path, the same comparison peaks at `max_abs=1.578331e-03`: a 13,240× separation, so the test is a real regression guard.

The explicit `rtol=1e-3` / `atol=0` are dropped in favour of `assert_close`'s float32 defaults (`rtol=1.3e-6`, `atol=1e-5`).

### DIFF
--- a/tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
+++ b/tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
@@ -1,0 +1,233 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+"""Megatron-FSDP v2 composed with expert parallelism through a real MCore HybridModel.
+
+Checks that an ``EP=4`` transformer-MoE ``HybridModel`` (an attention layer + a MoE layer)
+sharded with mFSDP v2 (experts over the expert-DP sub-mesh, dense params over the full DP
+mesh), consuming its ``1/dp`` shard of a global batch, reproduces a single **full-batch
+``EP=1`` reference**.
+
+The reference processes the whole global batch on every rank, so its gradients are
+identical across ranks and need no reduction -- it has no distributed logic. The model's
+gradients are reduced only by mFSDP. So this independently validates EP all-to-all
+dispatch, FSDP sharding, and the gradient reduction/scaling: a broken reduction (or a
+missing expert-grad scaling factor) would diverge from full-batch training.
+
+Both models are built from explicit ``ProcessGroupCollection``s (no global
+``parallel_state`` / ``initialize_model_parallel``): the reference with a size-1 ``ep``
+group (all experts local), the model with the 2-way ``ep`` group. Model shapes and the
+``(ep, dp)`` split are test-local so different tests can vary them.
+"""
+
+import pytest
+import torch
+import torch.distributed as dist
+from torch.distributed.device_mesh import init_device_mesh
+
+from megatron.core.distributed.fsdp.src.megatron_fsdp.experimental import (
+    Flat,
+    Placements,
+    fully_shard,
+)
+from megatron.core.models.hybrid.hybrid_layer_specs import hybrid_stack_spec
+from megatron.core.models.hybrid.hybrid_model import HybridModel
+from megatron.core.process_groups_config import ProcessGroupCollection
+from megatron.core.transformer.enums import AttnBackend
+from megatron.core.transformer.transformer_config import TransformerConfig
+from megatron.core.transformer.transformer_layer import MoETransformerLayer
+
+_FLAT_SHARD = Placements(dp_axes=[0], parameter=[Flat()], gradient=[Flat()], optimizer=[Flat()])
+
+
+def _transformer_config(
+    num_layers: int, num_experts: int, ep_size: int, hidden: int, ffn_hidden: int
+) -> TransformerConfig:
+    return TransformerConfig(
+        num_layers=num_layers,
+        hidden_size=hidden,
+        num_attention_heads=4,
+        num_moe_experts=num_experts,
+        expert_model_parallel_size=ep_size,
+        moe_token_dispatcher_type="alltoall",
+        moe_router_topk=2,
+        moe_aux_loss_coeff=0.0,
+        moe_grouped_gemm=True,
+        moe_ffn_hidden_size=ffn_hidden,
+        add_bias_linear=False,
+        gradient_accumulation_fusion=False,
+        use_cpu_initialization=True,
+        params_dtype=torch.float32,
+        hidden_dropout=0.0,
+        attention_dropout=0.0,
+        # unfused (native-PyTorch) attention: flash/fused don't support the fp32 params we use.
+        attention_backend=AttnBackend.unfused,
+    )
+
+
+def _build_process_group_collection(
+    one: dist.ProcessGroup,
+    dp: dist.ProcessGroup,
+    ep: dist.ProcessGroup,
+    expert_dp: dist.ProcessGroup,
+) -> ProcessGroupCollection:
+    """A ProcessGroupCollection for a TP=PP=CP=1 MoE model. `one` is a size-1 group for the
+    trivial TP/PP/CP axes; dp, ep, expert_dp are the data-, expert-, and expert-data-parallel
+    groups.
+    """
+    return ProcessGroupCollection(
+        tp=one,
+        expt_tp=one,
+        cp=one,
+        pp=one,
+        tp_cp=one,
+        tp_dp_cp=dp,
+        ep=ep,
+        tp_ep=ep,
+        expt_dp=expert_dp,
+        dp=dp,
+        dp_cp=dp,
+        embd=None,
+        pos_embd=None,
+    )
+
+
+def _build_hybrid_model(
+    config: TransformerConfig,
+    pg_collection: ProcessGroupCollection,
+    vocab: int,
+    seq: int,
+    pattern: str,
+) -> HybridModel:
+    return HybridModel(
+        config=config,
+        hybrid_stack_spec=hybrid_stack_spec,
+        vocab_size=vocab,
+        max_sequence_length=seq,
+        hybrid_layer_pattern=pattern,
+        pg_collection=pg_collection,
+    ).cuda()
+
+
+def _train(
+    model: torch.nn.Module,
+    ids: torch.Tensor,
+    pos: torch.Tensor,
+    mask: torch.Tensor | None,
+    target: torch.Tensor,
+    loss_reduce_group: dist.ProcessGroup | None = None,
+) -> list[torch.Tensor]:
+    """Run 5 SGD steps; return the per-step losses (globally averaged if loss_reduce_group given)."""
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.02, foreach=False)
+    losses = []
+    for _ in range(5):
+        optimizer.zero_grad()
+        loss = torch.nn.functional.mse_loss(
+            model(input_ids=ids, position_ids=pos, attention_mask=mask), target
+        )
+        loss.backward()
+        optimizer.step()
+        # The model sees a shard, so average the loss across ranks for the global loss.
+        loss = loss.detach()
+        if loss_reduce_group is not None:
+            dist.all_reduce(loss, op=dist.ReduceOp.AVG, group=loss_reduce_group)
+        losses.append(loss)
+    return losses
+
+
+def test_ep_fsdp_matches_fullbatch_reference(distributed_setup):
+    """EP=4 + mFSDP on 1/dp-sharded data reproduces single full-batch EP=1 training."""
+    device = distributed_setup.device
+    world_size, rank = distributed_setup.world_size, distributed_setup.rank
+
+    num_experts, ep_size = 8, 4
+    # hidden=64 with 4 heads -> head_dim=16, large enough for the attention backend.
+    hidden, ffn_hidden, vocab, seq, b_local = 64, 128, 128, 8, 2
+    # HybridModel builds one layer per pattern symbol, so "*E" is a two-layer stack: "*" a
+    # self-attention-only layer, "E" a MoE layer -- i.e. a real transformer-MoE block
+    # (attention then MoE, two residuals; equivalent to a Mixtral-style layer). This exercises
+    # mFSDP sharding the dense attention params over full DP alongside EP-sharded experts.
+    layer_pattern = "*E"
+    num_layers = len(layer_pattern)
+    if world_size % ep_size != 0 or num_experts % ep_size != 0:
+        pytest.skip(f"world_size {world_size} is incompatible with EP={ep_size}.")
+    edp_size = world_size // ep_size
+    global_batch = world_size * b_local  # one shard per rank
+
+    # Process groups (no global parallel_state). world_mesh: the full DP group; moe_mesh: the
+    # ep (ep_size-way) and expert-DP (edp_size-way) groups for the EP=4 model. Meshes also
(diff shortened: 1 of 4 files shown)


tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
```diff
@@ -0,0 +1,231 @@
...
+        tp_cp=one,
+        tp_dp_cp=dp,
+        ep=ep,
+        tp_ep=ep,
+        expt_dp=expert_dp,
+        dp=dp,
+        dp_cp=dp,
+        embd=None,
+        pos_embd=None,
+    )
```
wujingyue, line 92: @Connor-XY FYI, I had to `=one` many times when constructing a ProcessGroupCollection. Seems avoidable. 

(I tagged you because apparently you are trying to get rid of parallel_state in favor of ProcessGroupCollection)

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
```diff
@@ -0,0 +1,231 @@
...
+            for local, global_ in enumerate(model_layer.mlp.local_expert_indices):
+                getattr(model_fc, f"weight{local}").data.copy_(
+                    getattr(reference_fc, f"weight{global_}").data
+                )
+
+    # Shard the model: experts over the expert-DP sub-mesh, dense params over the full DP mesh.
+    for decoder_layer in model.decoder.layers:
+        if isinstance(decoder_layer, MoETransformerLayer):
+            fully_shard(decoder_layer.mlp.experts, mesh=moe_mesh["dp"], placements=_FLAT_SHARD)
+    fully_shard(model, mesh=world_mesh, placements=_FLAT_SHARD)
```
Autumn1998: In MCore's terminology, `world_mesh` here is the DP mesh, and the `moe_mesh["dp"]` used
for the experts is the EDP mesh. Is this naming going to be confusing?

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
```diff
@@ -0,0 +1,231 @@
...
+            reference_fc = getattr(reference_layer.mlp.experts, fc)
+            for local, global_ in enumerate(model_layer.mlp.local_expert_indices):
+                getattr(model_fc, f"weight{local}").data.copy_(
+                    getattr(reference_fc, f"weight{global_}").data
+                )
+
+    # Shard the model: experts over the expert-DP sub-mesh, dense params over the full DP mesh.
+    for decoder_layer in model.decoder.layers:
+        if isinstance(decoder_layer, MoETransformerLayer):
+            fully_shard(decoder_layer.mlp.experts, mesh=moe_mesh["dp"], placements=_FLAT_SHARD)
```
Autumn1998: Side question: do we support using ZeRO-1 for the MoE part and ZeRO-3 for the dense part?
If so, is there a corresponding unit test?

wujingyue: I haven't tested ZeRO-1 at all. 

The placements passed into fully_shard should be 
```
model_weight=[Replicate]
main_grad=[Partial]
main_weight=[Shard]
```

The implementation should probably reuse what's added by #5743 and #5972. I don't expect more than 10 lines of code change in: https://github.com/NVIDIA/Megatron-LM/blob/076f61fc7be6dd0d8a7a95f99cc66dd775ac5c42/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py#L287. But I could certainly be wrong. 

Would you like to give it a shot?

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
wujingyue: Do you prefer to call them `dp_mesh` and `edp_mesh`? 

There's a difference between "mesh" and "process group". A mesh is a container of process groups -- recall the `DeviceMesh.from_group` API. So I'd call a ProcessGroup `dp_group` or `edp_group` etc but keep calling mesh by their functionality (e.g. dense_mesh or moe_mesh). 

What do you think?

wujingyue: Question for @Autumn1998 and @shjwudp: is this the sharding you care about for QWen 3.5? Basically, MoE are sharded on DP=4 and EP=2, and the rest on DP=8.

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
```diff
@@ -0,0 +1,231 @@
...
+        optimizer.zero_grad()
+        loss = torch.nn.functional.mse_loss(
+            model(input_ids=ids, position_ids=pos, attention_mask=mask), target
+        )
+        loss.backward()
+        optimizer.step()
+        # The model sees a shard, so average the loss across ranks for the global loss.
+        loss = loss.detach()
+        if loss_reduce_group is not None:
+            dist.all_reduce(loss, op=dist.ReduceOp.AVG, group=loss_reduce_group)
```
wujingyue, line 133: Verify this is the right order. Should we all_reduce before loss.backward()?

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
```diff
@@ -0,0 +1,231 @@
...
+) -> list[torch.Tensor]:
+    """Run 5 SGD steps; return the per-step losses (globally averaged if loss_reduce_group given)."""
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.02, foreach=False)
+    losses = []
+    for _ in range(5):
+        optimizer.zero_grad()
+        loss = torch.nn.functional.mse_loss(
+            model(input_ids=ids, position_ids=pos, attention_mask=mask), target
+        )
+        loss.backward()
```
wujingyue, line 128: Double check: gradients may have to be divided by ep_size.

shjwudp (approved):

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
Autumn1998: Ok, I will try no-shard/zero-1/zero-2

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
Autumn1998: I mean, the name "dp" in ep_group, expert_dp_group = moe_mesh.get_group("ep"), moe_mesh.get_group("dp"), maybe use "edp" is more clear?

Autumn1998: In the Qwen models, we are more focused on large models like the 397B class, which are very hard to fit within 8 ranks. Overall, we prefer using larger EP settings, such as EP4-EDP2 or EP8-EDP1. The latter in particular may introduce some corner cases.

wujingyue: Certainly, will do -- thanks!

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
wujingyue: @shjwudp The order seems correct as is. TorchTitan's training loop runs the following three steps in order: 
1. https://github.com/pytorch/torchtitan/blob/1c40dd26a91b0489249011a53d1da69061b03d38/torchtitan/trainer.py#L800. compute the local loss
2. https://github.com/pytorch/torchtitan/blob/1c40dd26a91b0489249011a53d1da69061b03d38/torchtitan/trainer.py#L845. compute the cross-rank averaged loss from the local loss
3. https://github.com/pytorch/torchtitan/blob/1c40dd26a91b0489249011a53d1da69061b03d38/torchtitan/trainer.py#L874. print the cross-rank averaged loss

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
wujingyue: Thanks -- this turns out to be significant. I'll update the PR with the fix soon...

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
shjwudp: My understanding is that this steps eventually determines the final metric loss. Assuming the loss values involved in the computation are numerically correct, it should be sufficient to inspect only the first step.

TorchTitan may compute the loss locally on a per-token basis. If we want to follow the per-token loss formulation more precisely, the calculation should be:

```python
local_loss_sum = sum(
    (per_token_loss * loss_mask).sum()
    for microbatch in accumulation_window
)
loss_for_backward = local_loss_sum * dp_world_size / global_tokens
loss_for_backward.backward()
```

Here, `per_token_loss` is the output of `cross_entropy(reduction="none")`.

wujingyue: Are we talking about the same thing? I was trying to figure out whether the loss should be averaged **across ranks** before backward. As a data point, TorchTitan's training loop says no.

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
wujingyue: Done

Phlip79 (approved):

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
shjwudp: I think no need additional all-reduce of the loss is necessary for training correctness, since gradient averaging already provides the required effects.

If we assume that each DP rank receives an equivalent input, with no imbalance in the number of tokens, TorchTitan's approach is valid.

However, if the number of input tokens differs across DP ranks, we need to compute the loss on a per-token basis.

jiemingz (approved):

tests/unit_tests/distributed/mfsdp_v2/test_expert_parallel.py
wujingyue: > However, if the number of input tokens differs across DP ranks, we need to compute the loss on a per-token basis.

The local loss in this test is already locally normalized: mse_loss defaults to reduction="mean", averaging over the local batch × sequence × vocab elements. Every DP rank has the same shape and there is no loss mask, so averaging gradients across ranks is equivalent to computing the mean over the full global batch.

However, I agree that if the user computes the local loss as the sum we will need explicit global-token normalization. 

Am I understanding this correctly?

chtruong814:
@wujingyue Some unit tests were failing in the merge queue now.  Would you please take a look?
https://github.com/NVIDIA/Megatron-LM/actions/runs/30503354295/job/90766469959

wujingyue:
> **Edited twice.** This replaces my original analysis, which was wrong — it attributed the failures to NCCL reduction-order round-off and suggested loosening the test's tolerance. **Don't do that**; the assertion is catching a real bug. A later revision also wrongly claimed only `num_microbatches=3` was affected; corrected below. Earlier text is in this comment's edit history.

**Update on the mfsdp_v2 failures**

Comparing the sharded model's gradients against the baseline's at the end of step 0, before any optimizer step (every rank sees identical data, so a correct reduction must reproduce the baseline gradient exactly), with two spare 1-element all-reduces added to perturb timing:

| condition | gradients |
|---|---|
| unperturbed | exact — 48/48 comparisons zero, test passes 0/4 |
| + two 1-element all-reduces | **12%–125% wrong**, test fails 8/8 |

Worst observed: `layers.0.weight` off by 124.9%, `bias` by 67.8%. Magnitudes and the set of affected parameters vary run to run. The loss assertion only saw a ~1e-3 shadow of this because `lr=0.05` damps a one-step gradient error.

Both `num_microbatches` settings fail, `3` far more often than `1`. Counting failures of `test_hsdp_losses_match_baseline` across my runs: the PR arm gave 4 `[1-*]` and 3 `[3-*]`; a 32-run sweep gave 1 `[1-*]` and 35 `[3-*]`. The original CI failure here included `[1-True]` and `[1-False]`. So this is not specific to the deferred DP-outer accumulation path — with `num_microbatches=1` that path isn't exercised at all. More reductions per step just means more chances to hit it.

**It's the reduce-scatter stream.** Moving only `reduce_partial_gradients` to the current stream in `_reduce_gradient_groups` — leaving `partial_grad`'s allocation and both `wait_stream` edges untouched — makes it clean. So `main_grad` is written on `reduce_scatter_stream` and read somewhere without correct ordering.

I have **not** identified the missing ordering edge. Three hypotheses tested and refuted: `partial_grad` buffer lifetime (alloc and free are both on `reduce_scatter_stream`, so it's fine); autograd accumulating into `main_grad` (it doesn't — `copy_gradients_to_partial_buffer` reads the unsharded params); and the post-backward join resolving `current_stream()` on the autograd engine thread (adding a default-stream join neither fixes it nor is safe under CUDA graph capture).

Unperturbed mFSDP v2 gave bit-exact gradients in all my runs, so I have no evidence normal training is affected — but the perturbation needed to expose this is tiny.

PR 5956's source diff is a no-op for HSDP (at the default `grad_divisor=1` the new expression is identical to the old, and `test_fully_shard.py` never passes `grad_divisor`), this reproduces on clean main, and process groups are healthy throughout (598 known-answer collectives, 0 wrong). This looks like a pre-existing mFSDP v2 bug that 5956 surfaces by perturbing timing, not something 5956 introduces.

Reproducer: insert two `dist.all_reduce` calls on the mesh dim groups at the top of `test_hsdp_losses_match_baseline`, run the `mfsdp_v2` bucket on 8 GPUs. Fails 8/8. Ran on H200; CI is H100.

PR #5963, opened 2026-07-22
Fix fused MLA Q-RoPE in-place load/store races

### LINKED ISSUE #4640: Flaky: TestFusedApplyMLARope::test_forward_backward_for_q[thd] backward mismatch exceeds bf16 tolerances
### Summary

`tests/unit_tests/fusions/test_mla_yarn_rope_apply.py::TestFusedApplyMLARope::test_forward_backward_for_q[thd]` is intermittently failing in CI with a backward-pass numerical mismatch that exceeds bf16 tolerances.

### Observed failure

CI run: https://github.com/NVIDIA/Megatron-LM/actions/runs/25408640123 (job `tests/unit_tests/**/*.py - latest`, ID `74525633131`)

```
FAILED tests/unit_tests/fusions/test_mla_yarn_rope_apply.py::TestFusedApplyMLARope::test_forward_backward_for_q[thd]
E   AssertionError: Mismatch in bwd: Tensor-likes are not close!
E   Mismatched elements: 31 / 786432 (0.0%)
E   Greatest absolute difference: 3.015625 at index (104, 29, 170) (up to 0.05 allowed)
E   Greatest relative difference: 33.509803771972656 at index (104, 28, 166) (up to 0.02 allowed)
```

The assertion at `tests/unit_tests/fusions/test_mla_yarn_rope_apply.py:111` compares the backward gradient of the reference `apply_rotary_pos_emb` against the fused `fused_apply_mla_rope_for_q` in bf16, `thd` packed-sequence layout, `cu_seqlens=[0, 27, 54, 99, 128]`.

### Why this is non-deterministic

- Only 31 / 786,432 elements (~0.004%) exceed tolerance.
- bf16 tolerances: `atol=5e-2`, `rtol=2e-2`.
- The same merge-queue commit passed this job on rerun in workflow [25415024224](https://github.com/NVIDIA/Megatron-LM/actions/runs/25415024224).
- The fused MLA YARN RoPE backward kernel produces small numerical drift in the `thd` (packed-sequence) path that occasionally exceeds bf16 tolerances at outlier indices.

### Owning code

- Test: `tests/unit_tests/fusions/test_mla_yarn_rope_apply.py`
- Kernel: `megatron/core/fusions/fused_mla_yarn_rope_apply.py`
- Introduced in MR `!2949` (perf(mla, experimental): MLA RoPE fusion and YARN embedding cache); removed experimental tag in #2233.

### Mitigation

Marked `flaky_in_dev` in #4639 as a stop-gap. The underlying numerical drift in the fused backward kernel for `thd` should be investigated and tightened so the test can be re-enabled.

Sunt-ing:

# What does this PR do?

Issue #4640 reports intermittent BF16 gradient mismatches from `fused_apply_mla_rope_for_q` with packed THD input. The investigation found the same in-place load/store race in both Q-RoPE kernels. Forward converts interleaved values to split layout, while backward converts split-layout gradients back to interleaved layout. With multiple warps in one Triton program, one warp can overwrite a source element before another warp has loaded it.

This change adds one unconditional `tl.debug_barrier()` after the source loads and before the first overlapping store in each kernel. It restores the original upstream multi-warp autotune space and keeps the existing `flaky_in_dev` marker. THD scan and autotune-key changes remain separate performance work.

## Issue tracking

Fixes #4640

## Reproduction and regression

Run the fusion test file in the standard Megatron Core development environment with one CUDA GPU.

```bash
uv run python -m torch.distributed.run \
  --nproc-per-node 1 \
  -m pytest -q \
  --experimental \
  tests/unit_tests/fusions/test_mla_yarn_rope_apply.py
```

The new regression first checks a `BLOCK_H=1, num_warps=1` control against an independent PyTorch RoPE oracle for the tested configuration. It then compares known failing multi-warp configurations with that control using exact equality.

- Forward uses `BLOCK_H=1` with 8 warps and 200 replays for each layout.
- Backward uses `BLOCK_H=2` with 4 warps and 100 replays for each layout.

Removing only the forward barrier makes both forward layout cases fail. Removing only the backward barrier makes both backward layout cases fail. All four cases pass with both barriers present. Their test bodies take about 2.4 seconds combined on H200, and the slowest takes 1.43 seconds.

The same test file exercises the normal `fused_apply_mla_rope_for_q` entry point with both SBHD and packed THD input and runs autograd backward. The full file passed on the same H200 with two adjacent Triton versions.

- Triton 3.5.1 completed 11 tests in 23.86 seconds.
- Triton 3.5.0 completed 11 tests in 23.52 seconds.

### Supplemental public-entry stress

A separate BF16 packed THD stress run covered Q projection, `fused_apply_mla_rope_for_q`, and autograd backward from a scalar loss. It used 32 local heads and `cu_seqlens=[0,27,54,99,128]` on one H200. The runtime was Python 3.12.3, PyTorch `2.11.0a0+eb65b36914.nv26.02`, CUDA 13.1, driver 590.48.01, and Triton 3.5.1.

A single-warp control was validated against the PyTorch oracle before each forced multi-warp candidate was replayed 1,000 times.

| Observable                                 | Multi-warp configuration |        Before |       Fixed | Remove only its barrier |
| ------------------------------------------ | ------------------------ | ------------: | ----------: | ----------------------: |
| forward output                             | `BLOCK_H=1`, 8 warps   | 749 incorrect | 0 incorrect |           719 incorrect |
| backward Q and projection-weight gradients | `BLOCK_H=2`, 4 warps   |   8 incorrect | 0 incorrect |             4 incorrect |

The fixed candidates also completed 1,000 replays per direction without a mismatch on Triton 3.5.0. This stress run supplements the checked-in regression and is not required to execute it.

## Performance impact

The barrier-only benchmark forced identical block sizes and warp counts before and after the fix. Six rounds ran in alternating order on one H200. Each variant used 200 warmup launches and 2,000 timed samples per round. The table reports the pooled median across 12,000 timed samples per variant.

SBHD shapes are ordered as sequence length, batch size, local heads, and head dimension. THD shapes are ordered as token count, local heads, and head dimension.

| Kernel / layout | Input               | Before median | Fixed median |  Delta |
| --------------- | ------------------- | ------------: | -----------: | -----: |
| forward / SBHD  | `[4096,8,32,256]` |   868.032 µs |  852.960 µs | -1.74% |
| forward / THD   | `[32768,32,256]`  |  1606.688 µs | 1605.632 µs | -0.07% |
| backward / SBHD | `[4096,8,32,192]` |   322.624 µs |  322.624 µs |  0.00% |
| backward / THD  | `[32768,32,192]`  |   421.344 µs |  423.200 µs | +0.44% |

No material regression was observed. The negative forward deltas are not treated as a speedup claim.

### Pre-checks

Black, isort, Pylint, Python compilation, and `git diff --check` pass for both changed files.


Sunt-ing:
cc @ko3n1g @Victarry for visibility on #4640

Butterfingrz:
### Fix selection

Adding `tl.debug_barrier()` is sufficient to fix the race by ensuring that all loads complete before stores begin. It adds only one `bar.sync 0` at the PTX level, with negligible overhead.

We nevertheless use `num_warps=1` for all backward autotune configurations because `_get_thd_token_idx`, which runs inside `rotary_bwd_q_kernel`, is sensitive to multi-warp execution on packed THD inputs: its uniform scan, address calculations, and control flow execute across more warps. Single-warp execution therefore removes the race and can substantially reduce THD indexing overhead, making it the simplest robust fix. The cold-L2 results below compare the pre-fix, barrier, and single-warp variants.

Butterfingrz:
### SBHD cold-L2

The input shape is `[sequence_length, batch_size, local_heads, 192]`.

| SBHD input `[S, B, H, D]` | Pre-fix median | Barrier median | Single-warp median |          Latency Δ |
| :------------------------ | -------------: | -------------: | -----------------: | -----------------: |
| `[4096, 8, 16, 192]`      |      85.856 µs |      85.872 µs |          85.024 µs | -0.832 µs (-0.97%) |
| `[4096, 8, 32, 192]`      |     165.312 µs |     165.312 µs |         163.872 µs | -1.440 µs (-0.87%) |
| `[32768, 2, 32, 192]`     |     323.936 µs |     323.936 µs |         321.728 µs | -2.208 µs (-0.68%) |
| `[131072, 1, 32, 192]`    |     640.864 µs |     640.864 µs |         636.960 µs | -3.904 µs (-0.61%) |
| `[4096, 1, 128, 192]`     |      85.600 µs |      85.600 µs |          84.960 µs | -0.640 µs (-0.75%) |

### THD cold-L2

In the THD table, `T` is the number of local packed tokens processed by the current CP rank. The input shape is `[T_local, N, H, D]`.

| THD workload                            | Local input `[T, N, H, D]` | Pre-fix median | Barrier median | Single-warp median |             Latency Δ |
| :-------------------------------------- | :------------------------- | -------------: | -------------: | -----------------: | --------------------: |
| Single sequence, `S=131072`, `CP=1`     | `[131072, 1, 32, 192]`     |     641.056 µs |     641.040 µs |         636.896 µs |    -4.160 µs (-0.65%) |
| Issue #4640, ragged, `CP=1`             | `[128, 4, 32, 192]`        |       6.752 µs |       6.752 µs |           6.944 µs |    +0.192 µs (+2.84%) |
| Issue #4640, padded, `CP=2`, rank 0     | `[68, 4, 32, 192]`         |       6.784 µs |       6.784 µs |           7.008 µs |    +0.224 µs (+3.30%) |
| Issue #4640, padded, `CP=2`, rank 1     | `[68, 4, 32, 192]`         |       6.752 µs |       6.752 µs |           6.976 µs |    +0.224 µs (+3.32%) |
| Alternating `N=32` scan, `CP=2`, rank 0 | `[16384, 32, 32, 192]`     |     554.112 µs |     554.528 µs |         153.952 µs | -400.160 µs (-72.22%) |
| Alternating `N=32` scan, `CP=2`, rank 1 | `[16384, 32, 32, 192]`     |     551.360 µs |     551.696 µs |         153.408 µs | -397.952 µs (-72.18%) |

`Latency Δ = Single-warp − Pre-fix`; negative values indicate a speedup. THD coverage includes a single 131072-token sequence, ragged `(27, 27, 45, 29)`, alternating `(512, 1536) × 16`, and padded `CP=2` `(28, 28, 48, 32)` workloads across both CP ranks. All cases are token-packed and run with matching `cu_seqlens`, local `grad`/`dO`, and `cos`/`sin` tensors using `BLOCK_H=4`.

The test environment matches the PR description: NVIDIA H200, BF16, PyTorch 2.9.1+cu128, and Triton 3.5.1

guihong-nv:
Thanks for digging into this. The diagnosis matches what I see reading the kernel, but I don't think `num_warps=1` is the right mechanism, and the fix looks incomplete. A few things before this lands:

**1. `rotary_fwd_q_kernel` has the same hazard and is untouched.**

The forward kernel does the mirror-image in-place conversion over the identical byte range: loads at `x_1_off`/`x_2_off` (stride 2), stores at `x_left_off`/`x_right_off`, no fence between them, default `num_warps`. If the WAR argument holds for backward it holds here verbatim. Since `_test_fused_apply_mla_rope_for_q` also asserts on the forward output and this PR removes `flaky_in_dev` from it, I don't think we can call #4640 fixed while forward is left as-is. Either fix both, or please explain why forward is safe.

**2. `num_warps=1` isn't a correctness guarantee, and the new comment claims one.**

"Keep each program single-warp so every source load is issued before any overlapping destination store" doesn't follow from `num_warps=1`. What actually makes it work is LLVM refusing to reorder may-alias memory ops, plus lockstep lanes in non-divergent code. Both hold today; neither is a contract, and nothing in the test would catch a codegen change that broke either.

Please make the ordering explicit with `tl.debug_barrier()` between the loads and the stores in both Q kernels. The accesses are provably intra-CTA (programs partition `(token, head)` disjointly), so a single `bar.sync` is sufficient. Your own cold-L2 table is the strongest argument for it: the barrier variant is median-identical to pre-fix in nearly every row, while single-warp costs 2.8-3.3% on the small ragged/padded THD shapes.

**3. Please split the warp-count change out.**

The -72% on the alternating `N=32` scan isn't the race fix, it's `_get_thd_token_idx`'s serial O(seq_num) scan being re-executed by every warp. `num_warps=1` masks that rather than fixing it; the real fix is to stop recomputing a per-program linear scan. Bundling the two means a correctness patch also ships a measurable regression on V2-Lite-scale shapes. I'd prefer: barrier for correctness here, warp count + THD indexing as a follow-up with its own numbers.

**4. The new test.**

- It launches with `num_warps=config.num_warps`, which is always 1, so it never exercises the configuration that actually fails. The only real regression guard is `assert config.num_warps == 1`; the 8 compilations and 800 launches around it can't fail if that assert passes. With a barrier in place, launching at `num_warps=4` and asserting exactness would have real teeth.
- `BLOCK_H=64`/`128` at one warp with `half=32` puts roughly 128 elements per lane per tensor in registers, plus fp32 intermediates. In production the autotuner prunes configs that raise `OutOfResources`; this test launches them directly, so it will error rather than skip on any GPU with a tighter budget than the H200 you tested on. Could you post `n_regs` and spill counts for those two configs?
- It reaches into `.fn` and `.configs`, neither of which is stable Triton API, and it doesn't guard on the import having succeeded: `rotary_bwd_q_kernel` can be `None` from the bare `except`, and the inner `import triton` raises instead of skipping.
- No coverage of the forward direction or of a `cu_seqlens` path; only SBHD with `batch_size=1`.

**Smaller items:**

- Worth considering building the interleaved tile in registers (`tl.join`/`tl.interleave` + reshape) and issuing one contiguous `emb_dim`-wide store, so each thread writes the addresses it read and the hazard disappears by construction with four warps intact.
- `key=["emb_dim", "head_num"]` doesn't include whether `cu_seqlens_q` is `None`, so the SBHD-chosen `BLOCK_H` is reused for THD. Pre-existing, but it undercuts "all shapes selected BLOCK_H=4" and is probably part of why only `[thd]` got flagged flaky.
- I'd keep `flaky_in_dev` for a release and drop it after a soak rather than in the same PR.
- The perf tables say H20Z; the environment block says H200.
- `d40b015` is titled "Fix fused MLA RoPE backward synchronization" but the net diff contains no synchronization. Worth squashing with an accurate message. There's also an open duplicate in #5981 with the same title as `84b224a` - probably close one, and call out the co-authorship since the two commits carry different sign-offs.

Sunt-ing:
Hi @guihong-nv, I added one unconditional CTA barrier to each Q kernel and restored the original autotune space. I kept the existing layout mapping instead of using `tl.join`/`tl.interleave`, whose barrier-free correctness would require proving same-thread load/store ownership. THD indexing and autotune-key changes remain separate performance work. For the inspected H200 specializations under Triton 3.5.1, each barrier lowered to one `bar.sync 0`.

The regression covers SBHD and THD using reproducible multi-warp cases: forward uses `BLOCK_H=1` with 8 warps as a race-stress configuration, while backward uses `BLOCK_H=2` with 4 warps. Each is compared exactly with a PyTorch-validated single-warp control. Removing only the corresponding barrier produced mismatches in both layouts and directions; the fix passed on Triton 3.5.0 and 3.5.1. Across the four measured H200 cases, the largest positive pooled-median latency delta was +0.44%.

Sunt-ing:
Hi @guihong-nv, thanks for your high-quality comment! IMHO I've solved all of them. Do you have any update on the review?

Phlip79:
@Sunt-ing can you please resolve conflicts?

Sunt-ing:
Thanks for your notification @Phlip79! I've resolved conflicts. All H200 validation passed.

Sunt-ing:
@Phlip79 Any update on the review?

Sunt-ing:
Hi @guihong-nv @Phlip79, do you have any update on the review?

hxbai (approved):
LGTM

Phlip79 (approved):

jaredcasper (approved):

PR #5972, opened 2026-07-22
Support HFSDP deferred DP-outer gradient reduction

Achyuthan-S:
# Summary

Adds HFSDP (optimizer sharded across DP-outer) to the experimental Megatron-FSDP path, building directly on the HSDP deferred DP-outer reduction from #5743. Continues #5616.

HFSDP differs from HSDP in exactly one placement: the optimizer shards across DP-outer as well, so `optimizer=[Flat, Flat]` instead of `[Replicate, Flat]`. `parameter=[Replicate, Flat]` and `gradient=[Partial, Flat]` are unchanged.

The consequences fall out of that one change:

- **Finalize — no change.** `main_grad.redistribute(main_weight.placements)` becomes `[Partial, Flat] -> [Flat, Flat]` (`Partial -> Flat` = reduce-scatter on the DP-outer axis), which is already a supported `redistribute` transition. HSDP's all-reduce and HFSDP's reduce-scatter share the same generic finalize.
- **`sync_model_weight` — no change.** `main_weight[Flat, Flat] -> model_weight[Replicate, Flat]` (`Flat -> Replicate` = all-gather) reconstructs the DP-outer-sharded optimizer weights for compute via the existing path.
- **Reset — the one code change.** After finalize, `main_grad` rests at the reduce-scattered `[Flat, Flat]` optimizer placement, which is smaller than the `[Partial, Flat]` accumulation layout, so it cannot be relabeled back in place. The reset now allocates a fresh `[Partial, Flat]` accumulation buffer for that case (zeroed only when accumulating, i.e. `set_to_none=False`), while HSDP keeps the zero-copy `Replicate -> Partial` relabel.

# Testing

Run on a 2x2 mesh (4 ranks):

```
uv run python -m torch.distributed.run --nproc-per-node=4 \
  -m pytest tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py -k hfsdp
```

- `test_hfsdp_losses_match_baseline`: HFSDP parity against single-rank SGD for `num_microbatches` in `{1, 3}` and both `set_to_none` modes.
- `test_hfsdp_reduce_scatters_dp_outer_on_last_microbatch`: event counts confirm HFSDP never all-reduces and reduce-scatters `(num_microbatches + 1)` times per parameter group (DP-inner every microbatch plus DP-outer on the last).

The HSDP and plain-DP paths are unchanged.

cc @wujingyue @Autumn1998


megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -313,11 +330,25 @@ def has_grad(parameters: tuple[nn.Parameter, ...]) -> bool:
...
+        # on DP-outer), so relabel it in place; HFSDP's finalize reduce-scattered
+        # DP-outer to a smaller optimizer-sharded buffer, so allocate a fresh one
+        # (zeroed when we accumulate into it, i.e. sharded grads are still set).
         if self.main_grad.placements != self._accumulation_placements:
-            self.main_grad = self.main_grad.redistribute(self._accumulation_placements)
+            reset_axis = changed_mesh_axis(
+                self.main_grad.placements, self._accumulation_placements
+            )
+            if isinstance(self.main_grad.placements[reset_axis], Replicate):
+                self.main_grad = self.main_grad.redistribute(self._accumulation_placements)
```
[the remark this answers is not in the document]
Achyuthan-S: Added assert reset_axis is not None — we're inside the placements != accumulation guard so an axis always changed, and it matches how gather_axis/reduce_axis narrow changed_mesh_axis elsewhere in this file.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -253,6 +253,22 @@ def _install_sharded_grads(self) -> None:
         for index, sharded_parameter in enumerate(self.sharded_parameters):
             sharded_parameter.grad = self.main_grad.get_dtensor(index)
 
+    def _new_accumulation_grad(self) -> DBuffer:
```
wujingyue:
```suggestion
    def _allocate_accumulation_grad(self) -> DBuffer:
```
to be more precise

wujingyue: Also, given this function is short and called from only one site, is it better to just inline it?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -313,11 +330,26 @@ def has_grad(parameters: tuple[nn.Parameter, ...]) -> bool:
...
+        # DP-outer to a smaller optimizer-sharded buffer, so allocate a fresh one
+        # (zeroed when we accumulate into it, i.e. sharded grads are still set).
         if self.main_grad.placements != self._accumulation_placements:
-            self.main_grad = self.main_grad.redistribute(self._accumulation_placements)
+            reset_axis = changed_mesh_axis(
+                self.main_grad.placements, self._accumulation_placements
+            )
+            assert reset_axis is not None  # the placements differ, so an axis changed
+            if isinstance(self.main_grad.placements[reset_axis], Replicate):
+                self.main_grad = self.main_grad.redistribute(self._accumulation_placements)
```
wujingyue, line 388:
```suggestion
                // HSDP. Replicate=>Partial changes only metadata and reuses the tensor.
                self.main_grad = self.main_grad.redistribute(self._accumulation_placements)
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -313,11 +330,26 @@ def has_grad(parameters: tuple[nn.Parameter, ...]) -> bool:
...
+            else:
+                # Zero the fresh buffer only when we accumulate into it
```
wujingyue:
```suggestion
                # HFSDP. main_grad is sharded so not large enough to hold the accumulation. Therefore, we re-allocate. 
                # Zero the fresh buffer only when we accumulate into it
```

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -274,6 +287,82 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
...
+    is a reduce-scatter (not an all-reduce) that finalizes main_grad to the
+    optimizer's [Flat, Flat] placement, shrinking the buffer; the next step's reset
+    therefore allocates a fresh [Partial, Flat] accumulation buffer. Every rank
+    sees identical data, so the averaged gradient equals the single-rank gradient
+    and losses must match. Both ``zero_grad`` modes are covered.
+    """
+    rank = distributed_setup.rank
+    world_size = distributed_setup.world_size
+    device = distributed_setup.device
+    if world_size < 4 or world_size % 2 != 0:
```
```suggestion
    if world_size % 2 != 0:
```
world_size=2 is fine in the sense that it should still run and pass the convergence test, even though its communication pattern is too simple to be called HFSDP.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -274,6 +287,82 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
...
+
+    outer_size = 2
+    inner_size = world_size // outer_size
+    mesh = init_device_mesh(
+        device.type, (outer_size, inner_size), mesh_dim_names=("dp_outer", "dp_inner")
+    )
+    torch.manual_seed(1234)
+    dim = 8
+    baseline = MultiChildModel(dim=dim, num_children=2).to(device)
+    model = MultiChildModel(dim=dim, num_children=2).to(device)
```
wujingyue, line 393: Nit: .to(device) is probably unnecessary. 

fully_shard puts weights on GPU, so it's not needed for `model`. `baseline` is fine training on CPU unless that really causes numerics to diverge.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -340,6 +429,70 @@ def train_one_step() -> None:
...
+    train_one_step()
+    torch.cuda.synchronize(device)
+
+    with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA]) as prof:
+        train_one_step()
+        torch.cuda.synchronize(device)
+
+    cuda_events = [event for event in prof.events() if event.device_type.name == "CUDA"]
+    reduce_scatter_events = _nccl_events(cuda_events, "reducescatter", "reduce_scatter")
+    all_reduce_events = _nccl_events(cuda_events, "allreduce")
```
wujingyue: Can you rebase to main? I replaced `_nccl_events` with something else.

wujingyue (approved):
Great work! Thank you!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Inlined it into the reset as you suggested — helper's gone.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Added.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Achyuthan-S: Added — merged it with the set_to_none note right above the re-allocation.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Achyuthan-S: Relaxed to if world_size % 2 != 0 on the parity test — world_size=2 gives a 2x1 mesh that still exercises the DP-outer reduce-scatter and the fresh-buffer reset and converges, as you noted. I kept the event-count test at < 4 though: with inner_size=1 the DP-inner reduce-scatter is a size-1 no-op that emits no kernel, so its exact reduce-scatter count would be wrong.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Achyuthan-S: Kept baseline's — it shares the on-device x, and torch.randn on CPU vs GPU produces different values, so a CPU baseline would break parity rather than just perturb it. model's is redundant (fully_shard moves it to GPU); I left it for symmetry with the HSDP parity test.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Achyuthan-S: Rebased onto main and switched the event-count test to collect_linked_kernels. (The rebase auto-merged without a conflict but left the old _nccl_events calls dangling since the removal was in a different spot, so I fixed those up.)

yaoyu-33 (approved):

jiemingz (approved):

guihong-nv:
/ok to test [7ca8b70](https://github.com/NVIDIA/Megatron-LM/pull/5972/commits/7ca8b70d9698772644b70ce002aa4c937214eee2)

wujingyue:
https://github.com/NVIDIA/Megatron-LM/actions/runs/30526560328/job/90899263607?pr=5972 shows some real errors:
```
  FAILED tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py::test_hfsdp_losses_match_baseline[1-True] - AssertionError: HFSDP losses did not match baseline losses.
  FAILED tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py::test_hfsdp_losses_match_baseline[1-False] - AssertionError: HFSDP losses did not match baseline losses.
  FAILED tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py::test_hfsdp_losses_match_baseline[3-True] - AssertionError: HFSDP losses did not match baseline losses.
  FAILED tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py::test_hfsdp_losses_match_baseline[3-False] - AssertionError: HFSDP losses did not match baseline losses.

```

Could be a tight threshold, but I'd be cautious.

Achyuthan-S:
Hi @wujingyue , I just saw it , yes you are right , im on it 
Will check and report asap.

Achyuthan-S:
Thanks for catching this @wujingyue — you're right to be cautious, and it's a real failure, not just tolerance. Every rank sees identical data and the mesh is 2x4 (both powers of 2), so the AVG reductions are bit-exact — HSDP does the same averaging and passes all 4 within the same default tolerance. HFSDP failing all 4, including num_microbatches=1, points to a genuine scale or layout discrepancy rather than float noise.

This is my miss — earlier /ok to test runs hit "problem deploying to test" and never actually executed, so the HFSDP parity was never validated on GPU until this run. I own the fix.

I traced the HFSDP-specific paths statically (the [Partial,Flat]->[Flat,Flat] finalize reduce-scatter, the [Flat,Flat] optimizer sharding, and the [Flat,Flat]->[Replicate,Flat] sync all-gather) and they each look correct on paper, so I want the actual loss magnitude to tell scale-vs-layout apart before I patch anything. I'll reproduce on a 2x1 mesh (2 GPUs exercise the full DP-outer machinery), expose the numbers, land the fix, and report back with a clean run.

Achyuthan-S:
Update: root cause understood , thanks @wujingyue — the finalize self.main_grad = self.main_grad.redistribute(self.main_weight.placements) frees the old main_grad on a different stream than the reduce-scatter that's still reading it, so the allocator can reuse that storage mid-read -> corrupted gradients. Matches my static trace (the reduction is bit-exact correct; it's purely the stream/allocator race), and HFSDP exposes it because the reset allocates a fresh [Partial, Flat] buffer each step.

I'll implement the buffer-lifetime fix here and verify the parity on GPU + CI. One steer so it matches your stream model: I'm planning to allocate the fresh accumulation buffer on the reduce_scatter stream so alloc/collective/free stay single-stream (per the blog you linked), rather than record_stream — does that match the direction you had in mind, or are you already mid-fix and I should rebase on yours?

wujingyue:
I'm mid-fix; will update you on Monday.

Achyuthan-S:
@wujingyue — I see #6187 is approved (clean fix — exactly the cross-stream main_grad allocation). Once it lands on main I'll rebase #5972 on it, resolve the #6041 conflict in parameter_group.py, and confirm the HFSDP parity goes green in the mfsdp_v2 bucket. Just flagging I'm ready to turn it around quickly whenever the fix is in — happy to rebase and test.

wujingyue:
Thanks for waiting! Indeed, I got several dependency PRs approved. I'll let you know this PR is ready to go. There might be quite some rebase work for this PR, and I hope agents can do it reasonably well. If you need help, give me the access to your fork and I'm happy to fix the PR for you!

Achyuthan-S:
Thanks @wujingyue — really appreciate you lining up the dependency PRs and keeping #5972 moving. Sounds good: once you signal the deps are in, I'll rebase onto main and get the HFSDP parity green in the mfsdp_v2 bucket. I'll take the first pass myself, and if the substrate churn makes it gnarly I'll ping you and gladly take you up on the offer.

wujingyue:
All dependency PRs are merged to main. Can you rebase?

Achyuthan-S:
@wujingyue , Rebased onto main. Re-expressed the HFSDP reset on the new substrate and adapted the tests to fully_shard_context / microbatch(context). The fresh [Partial, Flat] accumulation buffer now allocates inside the reduce_scatter stream context, so it inherits your #6187 fix and should be race-free. Ready for /ok to test.

Achyuthan-S:
@wujingyue — heads up, the last run died on infra, not the change: cicd-container-build (aws-h100) hit the 24h timeout and was cancelled, so the unit-test buckets (including mfsdp_v2) were all skipped and Nemo_CICD_Test failed as a cascade. The code-level checks (linting, DCO, copyright, codeowners) all passed. Could you re-run the failed jobs? Or I can rebase onto current main to trigger a fresh run (the branch is a bit behind base now) — whichever you prefer. Just want the mfsdp_v2 bucket to actually execute so we confirm the HFSDP parity is green.

Achyuthan-S:
@wujingyue — CI ran cleanly this time (container built), and the results narrow it down: test_hsdp_losses_match_baseline (all 4) and test_hfsdp_reduce_scatters_dp_outer_on_last_microbatch both PASS, but test_hfsdp_losses_match_baseline still fails all 4 (same signature as before #6187). So the collective pattern is right, but the HFSDP numerics are still corrupted after the race fix.

My guess: HFSDP's reset re-allocates the finalized [Flat,Flat] buffer (where HSDP relabels in place), so it frees the finalized buffer — which the optimizer read on the compute stream — at the next reset on the reduce_scatter stream, i.e. a cross-stream free your __init__ fix doesn't cover. But I can't confirm scale-vs-layout without the loss magnitude (the test's custom msg hides it), and I don't want to guess-patch a lifetime bug in the reduce_scatter substrate.

I'll gladly take you up on the fork-access offer if you want to debug it directly on GPU — or I can push a diagnostic that exposes the magnitude first, whichever you prefer.

wujingyue:
> My guess: HFSDP's reset re-allocates the finalized [Flat,Flat] buffer (where HSDP relabels in place), so it frees the finalized buffer

Why is it a problem? The free happens in the same stream as the re-allocation.

> I'll gladly take you up on the fork-access offer

Please do that. I can also push small cleanups on your behalf. 

> I can push a diagnostic that exposes the magnitude first. 

I'll let you try that too. 

Reproducing such race conditions is indeed hard. Can you try the small `all_reduce` perturbation probe mentioned in the description of https://github.com/NVIDIA/Megatron-LM/pull/6187? It made the repro much more reliable on 4xA100, and could help HFSDP as well.

Achyuthan-S:
@wujingyue , you're right — the free and the realloc are both on the reduce_scatter stream, so scratch that.

Fork access is on now, please go ahead. I'll push the #6187 perturbation probe plus a magnitude diagnostic (dropping the custom msg so torch prints the numeric diff) so the next run gives you a reliable repro and the actual numbers — I'll ping you once it's up so we don't push over each other.

One data point meanwhile: HFSDP fails all 4 including both num_microbatches=1 cases ([1-True], [1-False]), whereas your #6187 writeup notes the HSDP repro never failed at num_microbatches=1.

Achyuthan-S:
Pushed — the #6187 perturbation probe (per-axis all_reduce after mesh construction) plus a magnitude diagnostic ([HFSDP-DIAG] prints the sharded/baseline losses and their abs diff, and I dropped the custom assert msg so torch prints the numeric mismatch). 

Both are marked temporary/do-not-merge and I'll revert them before this is ready. Whenever you get a chance: /ok to test. Fork access is on, so feel free to push over this if you want to poke at it directly.

wujingyue:
@Achyuthan-S — reproduced on 4xH100, and it's not a race. Pushed a fix on top of your commits.

The HFSDP losses are bit-identical across all five steps (`1.2920293` x5), while the optimizer weight advances every step and the gradient never moves — forward is reading a stale compute weight.

HSDP's optimizer placement `[Replicate, Flat]` equals its parameter placement, so `main_weight` and `model_weight` alias and the optimizer writes the compute weight directly. HFSDP's `[Flat, Flat]` makes them distinct, so `sync_model_weight_from_main_weight()` is required — and it only runs from the `step_post_hook` that `fully_shard_optimizer()` registers. The test used a plain `torch.optim.SGD`, so it never ran.

Fix is one line:

```python
fully_shard_optimizer(optimizer)
```

HFSDP then matches the baseline exactly (absdiff `0.0` at every step). HSDP untouched. I also dropped the probe and the magnitude diagnostic, since the failure is deterministic.

This needs #6336 first — without it `fully_shard_optimizer` deadlocks with >1 parameter group, since its post-hook iterates a `set` of `FsdpParameterGroup` (hashed by `id()`), so ranks issue the per-group collectives in different orders. I'll merge this after #6336 lands.

Achyuthan-S:
@wujingyue Nice — that's a clean root cause, and thanks for reproducing it on H100.
 
The identical losses across all five steps make sense now: with HFSDP's [Flat, Flat] optimizer placement distinct from the [Replicate, Flat] parameter placement, main_weight and model_weight are separate buffers, so a plain SGD never fires the step_post_hook that runs sync_model_weight_from_main_weight() — the compute weight stays frozen at init while the optimizer advances. HSDP aliasing those two placements is exactly why it passed unwrapped. So it was the test harness, not the reduction, and not a race .

fully_shard_optimizer(optimizer) is the right fix. Happy to wait on #6336 for the post-hook ordering before this merges — and glad to pick up anything you'd like on either PR.

wujingyue:
Hey @chtruong814, the PR has passed CI except the approval check which I guess is due to https://github.com/NVIDIA/Megatron-LM/pull/6529. What would you recommend me to do next?

wujingyue:
@Achyuthan-S the merge queue said this probably needs a rebase. I wanted to do it for you but found I no longer have the push access to your branch.

wujingyue:
> @Achyuthan-S the merge queue said this probably needs a rebase. I wanted to do it for you but found I no longer have the push access to your branch.

Never mind. claude-fix seemed to work!

Achyuthan-S:
@wujingyue Fork access is on my side — "Allow edits by maintainers" is enabled, so you and the CI bot should both be able to push; happy to re-toggle it if it's still blocking you.
I checked the claude-fix merge commit against parameter_group.py — it looks right: it keeps the HSDP/HFSDP reset branch and hoists the new allocation_stream assert above it so it guards both arms, with the HFSDP re-alloc path unchanged. 

If the merge queue still wants the branch brought up to date and it's easier for me to rebase onto main, just say the word — I've held off to avoid dismissing the approvals with a force-push.

PR #5976, opened 2026-07-22
Fix Megatron FSDP optimizer under-update

### LINKED ISSUE #5974: [BUG] Megatron-FSDP fully_shard optimizer under-applies the weight update vs DDP/FSDP2 (same gradient → ~1.3–2× smaller step)
### Summary

With `megatron_fsdp.fully_shard(model, optimizer, zero_dp_strategy=3)`, given the **same input gradient**, the realized weight update is materially **smaller** than a reference `torch.optim.AdamW` step on DDP-averaged gradients. Training diverges from the DDP/FSDP2 reference starting at the first optimizer step — loss descends more slowly and grad-norm stays systematically higher. Both ZeRO-2 (`zero_dp_strategy=2`) and ZeRO-3 (`=3`) are affected.

This is silent: nothing errors and loss still decreases, so a user gets under-trained results with no signal.

### Environment

* `megatron-fsdp` 0.5.0. Reproduced on a 0.5.0-based fork that adds only unrelated `make_fsdp_dtensor` / `MegatronFSDP.forward` / `update_main_grads` local-shape fixes (issues NVIDIA/Megatron-LM#5786/NVIDIA/Megatron-LM#5787/NVIDIA/Megatron-LM#5789) — the `fully_shard_optimizer` / main-weight update path is unchanged from 0.5.0, so this is a 0.5.0 bug.
* torch 2.12, python 3.12, 2×–8× H100, NCCL.

### Minimal repro (torch + `megatron_fsdp` only)

`torchrun --standalone --nproc_per_node=2 repro.py` (reproduces at dp=2; also dp=4/8, and with `DTYPE=float32`).

Reference arm = replicated model + `all_reduce(AVG)` grads + `torch.optim.AdamW` (exact DDP). mfsdp arm = same model/init/data, `fully_shard(zero_dp_strategy=3)`, same AdamW. The realized update is measured backend-agnostically as the change in the model's function on a fixed probe input, `‖model_after_step(z) − model_before_step(z)‖` — the compute param can't be read directly under `fully_shard` (it's a transient flat-buffer view with `placements=None` that raises `CUDA error: invalid argument` on `.clone()`; a hook on `AdamW.step` also sees a zero delta on `optimizer.param_groups`, i.e. the real update goes through the distributed main-weight path).

<details><summary>repro.py</summary>

```python
#!/usr/bin/env python3
# Minimal torch-only repro of the Megatron-FSDP optimizer UNDER-UPDATE bug.
#
# No NeMo, no HuggingFace, no transformers, no from_pretrained. Pure torch +
# megatron_fsdp.
#
# CLAIM: given the SAME gradient, megatron_fsdp.fully_shard(model, optimizer,
# zero_dp_strategy=3)'s optimizer path applies a MATERIALLY SMALLER weight update
# than a reference torch AdamW (same LR/betas/eps, weight_decay=0). Step-0 loss is
# IDENTICAL across the two arms (same init + same data + same forward => same
# gradient, by determinism), but after the first optimizer.step() Megatron-FSDP's
# realized update is ~Nx smaller, so the loss trajectory diverges from step 1 on.
#
# HOW WE MEASURE THE UPDATE ("direct", not just a loss proxy): we cannot read the
# model's compute weight directly under Megatron-FSDP -- after fully_shard the
# param object (e.g. `2.weight`) is a transient *plain-tensor view into
# Megatron-FSDP's internal flat buffer* (placements=None) that is invalidated on
# reshard; `.clone()`/`.to_local()` on it raises `CUDA error: invalid argument`.
# So we measure the update the model ACTUALLY applied via a fixed probe input z:
# ||model_after_step(z) - model_before_step(z)|| -- the change in the function the
# model computes. A smaller weight update => smaller functional change. This is a
# faithful, backend-agnostic measure of the realized optimizer step (forward-only,
# so it works identically for the reference and for Megatron-FSDP).
#
# VERSIONS:
#   * megatron_fsdp from fork  github.com/yuhezhang-ai/Megatron-LM @ 920e1d24
#     (branch fix/mfsdp-0.5.0-local-shape, subdirectory megatron/core/distributed/fsdp/src),
#     installed via `uv sync --locked` inside canonical container nemo_26.06.00.sqsh.
#   * torch 2.12 (nv container), python 3.12.
#
# LAUNCH (dp = nproc):
#   torchrun --standalone --nproc_per_node=2 minimal_optimizer_underupdate_repro.py
#   (also try --nproc_per_node=8; set DTYPE=float32 to test precision-independence)
#
# OBSERVED (fork @ 920e1d24, torch 2.12, tiny 3-Linear model, lr=1e-2):
#   step-0 loss IDENTICAL to 0.00e+00 across arms (same gradient), then:
#     dp=2 bf16 : reference ||Df||=2.796  mfsdp ||Df||=2.063  -> mfsdp 1.36x smaller
#     dp=4 bf16 : reference ||Df||=2.683  mfsdp ||Df||=1.989  -> mfsdp 1.35x smaller
#     dp=8 bf16 : reference ||Df||=2.801  mfsdp ||Df||=2.067  -> mfsdp 1.35x smaller
#     dp=8 fp32 : reference ||Df||=2.799  mfsdp ||Df||=2.083  -> mfsdp 1.34x smaller
#   The under-update is consistent across dp size AND precision-independent (bf16==fp32),
#   confirming it is the optimizer step, not a reduction/precision effect. mfsdp's loss
#   diverges upward (under-trains) from step 1 onward. Reproduces at dp=2 (2 GPUs); no
#   large scale or real model needed. (On the real Qwen3-0.6B the ratio is ~2.2x; the
#   tiny model shows the same qualitative bug at a smaller magnitude.)
#
# REFERENCE ARM: replicated model + explicit all-reduce(AVG) of grads across dp
# (exact DDP semantics) stepped with plain torch.optim.AdamW.
# MFSDP ARM:     same model / same init / same data, wrapped with
# megatron_fsdp.fully_shard(zero_dp_strategy=3), same torch.optim.AdamW.
#
# ROOT CAUSE (what this isolates): for an identical input gradient, Megatron-FSDP's
# fully_shard_optimizer applies a smaller weight update to the model's compute
# weights than plain torch AdamW; its distributed main-weight update path
# under-applies the step, so training diverges from the DDP/FSDP2 reference.

import os

import torch
import torch.nn as nn
from torch.distributed.device_mesh import init_device_mesh

from megatron_fsdp import fully_shard

HIDDEN = 128
STEPS = 6
SEED = 1234
DTYPE = {"bfloat16": torch.bfloat16, "float32": torch.float32}[os.environ.get("DTYPE", "bfloat16")]
LR = 1e-2


def build_model() -> nn.Sequential:
    torch.manual_seed(SEED)  # identical init across arms
    return nn.Sequential(
        nn.Linear(HIDDEN, HIDDEN),
        nn.GELU(),
        nn.Linear(HIDDEN, HIDDEN),
        nn.GELU(),
        nn.Linear(HIDDEN, HIDDEN),
    )


def fixed_batch(rank: int, device) -> tuple[torch.Tensor, torch.Tensor]:
    g = torch.Generator().manual_seed(4242 + rank)  # per-rank fixed data, shuffle off
    x = torch.randn(8, HIDDEN, generator=g).to(device=device, dtype=DTYPE)
    y = torch.randn(8, HIDDEN, generator=g).to(device=device, dtype=DTYPE)
    return x, y


def probe_input(device) -> torch.Tensor:
    g = torch.Generator().manual_seed(777)  # shared, fixed probe (same on all ranks)
    return torch.randn(16, HIDDEN, generator=g).to(device=device, dtype=DTYPE)


def ref_grad_norm(params) -> float:
    # Reference grads are replicated (full) after all-reduce(AVG); norm is local.
    sq = torch.zeros((), dtype=torch.float64, device="cuda")
    for p in params:
        if p.grad is not None:
            sq = sq + p.grad.double().pow(2).sum()
    return float(sq.sqrt().item())


def run_arm(kind: str, mesh, rank: int):
    local_rank = int(os.environ["LOCAL_RANK"])
    device = torch.device("cuda", local_rank)
    model = build_model().to(device=device, dtype=DTYPE)
    opt = torch.optim.AdamW(model.parameters(), lr=LR, betas=(0.9, 0.999), eps=1e-8, weight_decay=0.0)

    if kind == "mfsdp":
        model, opt = fully_shard(
            model,
            opt,
            device_mesh=mesh,
            dp_shard_dim="dp",
            tp_dim="tp",
            zero_dp_strategy=3,
            fsdp_unit_modules=[nn.Linear],
        )

    x, y = fixed_batch(rank, device)
    z = probe_input(device)
    per_step = []  # (loss, ref_grad_norm_or_nan)
    probe = []  # model(z) at the START of each step (before this step's update)

    for _step in range(STEPS):
        with torch.no_grad():
            probe.append(model(z).detach().float().clone())  # current function on probe z

        out = model(x)
        loss = (out.float() - y.float()).pow(2).mean()
        opt.zero_grad()
        loss.backward()

        if kind == "ref" and torch.distributed.is_initialized():
            for p in model.parameters():  # DDP semantics: average grads across dp
                if p.grad is not None:
                    torch.distributed.all_reduce(p.grad, op=torch.distributed.ReduceOp.AVG)
            gn = ref_grad_norm(list(model.parameters()))
        else:
            gn = float("nan")  # mfsdp keeps grads in internal buffers; use loss to show same grad

        per_step.append((float(loss.item()), gn))
        opt.step()

    # Realized update applied by step-0's optimizer.step, measured as the change in
    # the model's function on the fixed probe between step-0's and step-1's forward.
    df0 = float((probe[1] - probe[0]).norm().item())
    return per_step, df0


def main():
    torch.distributed.init_process_group(backend="nccl")
    rank = torch.distributed.get_rank()
    world = torch.distributed.get_world_size()
    torch.cuda.set_device(int(os.environ["LOCAL_RANK"]))
    # Megatron-FSDP requires named dp + tp mesh dims (tp=1 here, dp-only sharding).
    mesh = init_device_mesh("cuda", (world, 1), mesh_dim_names=("dp", "tp"))

    ref, df_ref = run_arm("ref", mesh, rank)
    mf, df_mf = run_arm("mfsdp", mesh, rank)

    if rank == 0:
        print(f"\n=== dp={world}  dtype={DTYPE}  lr={LR}  (fork megatron_fsdp @ 920e1d24) ===")
        print(f"{'step':>4} | {'ref loss':>10} {'ref gnorm':>10} | {'mfsdp loss':>11}")
        for s in range(STEPS):
            print(f"{s:>4} | {ref[s][0]:>10.5f} {ref[s][1]:>10.5f} | {mf[s][0]:>11.5f}")
        loss0_diff = abs(ref[0][0] - mf[0][0])
        print(
            f"\nstep-0 loss: ref={ref[0][0]:.6f}  mfsdp={mf[0][0]:.6f}  (identical to {loss0_diff:.2e}) "
            f"-> SAME init + SAME data + SAME forward => SAME input gradient"
        )
        ratio = df_ref / df_mf if df_mf > 0 else float("inf")
        print(
            f"step-0 realized update ||Df(z)||: reference={df_ref:.6e}  mfsdp={df_mf:.6e}  "
            f"(mfsdp {ratio:.2f}x smaller)"
        )
        loss1 = (mf[1][0] - ref[1][0])
        print(f"step-1 loss gap: mfsdp {mf[1][0]:.5f} vs ref {ref[1][0]:.5f}  (mfsdp under-trains by {loss1:+.4f})")
        reproduced = ratio > 1.15 or loss1 > 0.02
        print(
            f"\n[VERDICT] {'REPRODUCED' if reproduced else 'NOT reproduced'}: SAME input gradient "
            f"(step-0 loss identical), but reference update ||Df||={df_ref:.4e} vs "
            f"mfsdp ||Df||={df_mf:.4e} (mfsdp {ratio:.2f}x smaller) "
            f"-> Megatron-FSDP under-applies the optimizer step."
        )
    torch.distributed.destroy_process_group()


if __name__ == "__main__":
    main()
```

</details>

### Observed

```
step-0 loss IDENTICAL across arms (same gradient), then:
  dp=2 bf16 : reference ‖Δf‖=2.796  mfsdp ‖Δf‖=2.063  → mfsdp 1.36× smaller
  dp=4 bf16 : reference ‖Δf‖=2.683  mfsdp ‖Δf‖=1.989  → mfsdp 1.35× smaller
  dp=8 bf16 : reference ‖Δf‖=2.801  mfsdp ‖Δf‖=2.067  → mfsdp 1.35× smaller
  dp=8 fp32 : reference ‖Δf‖=2.799  mfsdp ‖Δf‖=2.083  → mfsdp 1.34× smaller
```

Consistent across dp size and **precision-independent** (bf16 ≈ fp32), so it is the optimizer step, not a reduction/precision effect. On a real model (Qwen3-0.6B, 8×H100) the under-update is ~2.2×, and mfsdp's grad-norm diverges **+42%** from FSDP2 and DDP (which agree to <0.5%) by step 1; both ZeRO stages behave identically.

### Root cause (isolated to)

For an identical input gradient, `fully_shard_optimizer`'s distributed main-weight update path applies a smaller step to the model's compute weights than plain `AdamW`. Ruled out by controlled A/B: gradient scaling (grads are correctly averaged by 1/dp; step-0 grads identical), gradient clipping, overlap flags, and master-weight precision (`preserve_fp32_weights` on/off both diverge identically). Happy to help instrument the main-weight update path further.

wujingyue:
## Summary

- make Megatron FSDP optimizer-state preinitialization numerically inert
- preserve learning-rate and weight-decay settings while resetting both per-parameter and per-group step counters
- add behavioral regression coverage for Adam, AdamW, and Transformer Engine FusedAdam with ZeRO-2 and ZeRO-3

## Root cause

`fully_shard_optimizer()` performs a synthetic zero-gradient optimizer step to materialize state before loading DCP checkpoints. A zero gradient alone does not make that step inert:

- Adam-style optimizers advance their step counters independently of the learning rate. Leaving the synthetic step counted makes the first real update use step 2 bias correction, producing the approximately 1.34x under-update reported in #5974.
- Decoupled weight decay can update parameters when the learning rate is nonzero.
- Coupled weight decay is added to the effective gradient before moment accumulation. Setting only the learning rate to zero prevents the parameter update but can still populate `exp_avg` and `exp_avg_sq` with synthetic decay state.

The initialization step now temporarily sets both learning rate and weight decay to zero, restores the caller's optimizer settings after state allocation, and resets all initialized step counters to zero. This keeps the newly allocated state equivalent to an optimizer that has not yet trained.

## Design note

This initialization is similar to PyTorch DCP's `_init_optim_state`, which also materializes lazy optimizer state with a synthetic zero-gradient step. MFSDP v1 currently performs it in `fully_shard_optimizer()`, where fresh training may follow, so the step must be numerically inert.

An open follow-up is whether initialization should instead move to checkpoint loading, as in the MFSDP v2 checkpoint API proposed in #6024. In that flow, the checkpoint immediately overwrites the synthetic optimizer state.

## User impact

The first and subsequent Megatron FSDP optimizer updates now use the expected Adam bias correction and unmodified moment state. DCP optimizer state remains eagerly initialized for checkpoint loading.

## Testing

- `python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/mfsdp_v1/test_mfsdp_fully_shard.py::TestMegatronFsdpFullyShard::test_optimizer_loss_curve_matches_reference`
  - passes for Adam, AdamW, and FusedAdam with ZeRO-2 and ZeRO-3
  - verified that the AdamW/ZeRO-2 case fails against the pre-fix implementation
- `python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/mfsdp_v1/test_mfsdp_fully_shard.py::TestMegatronFsdpFullyShard::test_fully_shard_ez`
- Black, isort, pylint, and ruff via `tools/autoformat.sh`

Fixes #5974


tests/unit_tests/distributed/mfsdp_v1/test_mfsdp_fully_shard.py
```diff
@@ -748,6 +748,41 @@ def test_fully_shard_ez(self, shard_strategy):
...
+        """Optimizer state pre-initialization must not consume a training step."""
+        if optimizer_type == "fused_adam" and not HAVE_TE_FUSED_ADAM:
+            pytest.skip("Transformer Engine FusedAdam is not available")
+
+        torch.manual_seed(1234)
+        model = RootParamModel().cuda()
+        model_input = torch.randn(DIM_SIZE, DIM_SIZE, device="cuda")
+        expected_output = model(model_input).detach().clone()
+
+        mfsdp_model = fully_shard_model(
```
akoumpa: Thanks for the quick fix @wujingyue .

For sparsely activated models, like MoE, I've seen during early training, `step` being missing for some parameters (e.g., expert they didn't get any update), so I'm wondering whether it would make sense to include such a case here in the pytest.

wujingyue: I'll try to follow up in a separate PR. Initial prototyping with MoE hit some bugs that I can explain so far.

shjwudp (approved):

Phlip79 (approved):

deepakn94 (approved):

Issue #5968, opened 2026-07-22
[ENHANCEMENT] Fused forward+backward CUDA kernel for the CSA/HCA `Compressor` gated pooling (THD path)


zkyue:
**Is your feature request related to a problem? Please describe.**

The `Compressor` module used by the CSA/HCA experimental attention variants
(`megatron/core/transformer/experimental_attention_variant/csa.py`, `dev` branch)
implements its gated softmax pooling in eager PyTorch. In `_forward_thd`, the
region between the two projection GEMMs and the RMSNorm — gather-index build,
gather, `+ APE`, overlap-window transform (`coff == 2`), fp32 softmax over the
window, gated weighted sum — decomposes into ~39 forward and ~51 backward
kernel launches per call (measured at `compress_ratio = 4`), materializes
`(total_comp, 2*ratio, 1, head_dim)` window intermediates (also saved for
autograd), and its backward pays an `indexing_backward` + radix sort for the
gather. Per CSA/HCA layer this costs hundreds of microseconds per microbatch
in both directions at the sizes below, and wall-clock time sits well above the
summed kernel durations, i.e. a large share is launch/host overhead.

**Describe the solution you'd like**

An additive fused fast path for exactly that region: one forward kernel and one
backward kernel written in the CuTe Python DSL (`nvidia-cutlass-dsl`, JIT-compiled
at first call), keeping the existing eager implementation as the semantic
reference and fallback. We have this working and validated on the
configurations reported below, and we would like to contribute it as a PR if
there is interest.

Properties of the fused kernels:

- Covers the THD packed path (`packed_seq_params.qkv_format == 'thd'`), both
  `coff == 2` (overlapping, `compress_ratio == 4`) and `coff == 1`
  (non-overlapping), with the same trailing-token semantics as the eager code
  (per-segment `seqlen % ratio` tail dropped, first-window `-inf`/zero fill at
  segment boundaries). Segment lengths need not be multiples of `ratio`
  (a ragged pack is part of the correctness runs below). Current
  implementation limit: int32 offsets, i.e. `total_tokens * coff * head_dim <
  2**31`.
- All arithmetic in fp32 with a single final bf16 rounding (the eager path
  currently rounds the softmax weights to bf16 and multiplies in bf16 before
  the fp32-accumulated sum — see numerics below).
- Backward is atomic-free for `dKV`/`dScore`: every consumed input element
  belongs to exactly one pooling window, so gradient stores are disjoint
[...]

Issue #427, opened 2023-07-21
[QUESTION] Does the zero-out operation conduct during the model initialization and affect reproducibility?


zxgx:
**Your question**

As indicated in the annotation, embedding weight in the last stage would be zeroed out. 
I doubt that this zero-out operation is not conducted, as the two conditions in the `if` statement in L87 are contradictory.
But I assume that, finally after the model initialization, the two embedding weights in `self.language_model.embedding` and `self.word_embeddings` would be added up by the following `all_reduce` anyway, so the two weight matrices would still be the same, am I right? 
If so, I assume this problem may lead to some reproducibilitiy issue since the following `all_reduce` changes the initial embedding weight imported from other models.

Just close this issue if I'm wrong 😄 

https://github.com/NVIDIA/Megatron-LM/blob/6ef5bdc9fbe030c45d9f7a19e6c42d39f844aaa7/megatron/model/module.py#L85-L89

```[tasklist]
### Tasks
```

jaredcasper:
This is for the T5 model where you have an embedding layer as part of the first stage in the decoder, which isn't the first stage of the overall pipeline. In that case, and when using the embedding weights as the weights for the output layer, there are three copies of the embedding weights: at the beginning of the encoder, beginning of the decoder, and the output layer.

This code isn't run for GPT or Bert models.

Does that answer the question?

zxgx:
That's true! I must ignore some setup for encoder&decoder models because I follow scripts in `examples/sc21` to read the source code. 😵 

Thanks for your answer!

PR #5984, opened 2026-07-23
Dispatch the CSA Compressor gated pooling to the cudnn-frontend fused kernels (THD)

zkyue:
### What (reworked per review)

Addresses #5968. Reworked as discussed with @hxbai: the fused kernels themselves have
moved to cudnn-frontend — NVIDIA/cudnn-frontend#427 (**merged** into `develop` at
`b950af1`, 2026-07-30) adds them behind the FE-OSS
`APIBase` pattern (`cudnn.csa.compressor`, `CSACompressorForward`/`Backward` +
high-level wrappers) — and this PR is now the **thin Megatron-side dispatch** only:

- `Compressor._forward_thd` (THD packed, non-pre-grouped path) tries the fused fast
  path for the gated-softmax pooling region — gather-index build, gather, `+ APE`,
  overlap-window transform (`coff == 2`), fp32 softmax, gated weighted sum — and keeps
  the existing eager implementation as the semantic reference and fallback for every
  unsupported configuration. One fused compute kernel per direction (plus a small
  `dAPE` zero-init in backward) instead of ~40 forward / ~50 backward eager launches.
- The dispatch is additive: the eager region is unchanged (re-indented only), SBHD and
  the pre-grouped CP-prep input path are untouched.

Compared to the previous revision of this PR, the ~1000-line kernel module and the
benchmark harness are gone (they live in cudnn-frontend #427, harness included in its
test/bench assets); what remains is a ~240-line dispatch shim, the `csa.py` edit, and a
trimmed test file: **+732 / −37** vs `dev` (previously +1930 / −37).

### Dependency

- **cudnn-frontend with the CSA compressor API** provides the kernels:
  NVIDIA/cudnn-frontend#427, **merged** into `develop` at `b950af1` (2026-07-30). The
  fused path activates when the installed cudnn-frontend provides `cudnn.csa` — i.e. a
  source install of `develop` >= `b950af1`, or the upcoming `nvidia-cudnn-frontend` 1.27
  pip release (expected early August) — otherwise the dispatch silently keeps the eager
  path. This includes the frontend's `nvidia-cutlass-dsl` runtime dependency (the
  frontend's `cutedsl` extra). **No hard dependency is added to Megatron**: availability
  is probed at dispatch time by importing the concrete entry points
  (`cudnn.csa.compressor.csa_compressor_forward_wrapper` / `..._backward_wrapper`) —
  capability detection, not version comparison. `nvidia-cudnn-frontend` installs that
  predate the API simply lack `cudnn.csa`, and any import failure (missing package,
  missing DSL extra, partial install) silently keeps the eager path; `csa.py`'s import
  chain never fails because of the frontend.
- Autograd wiring stays on the Megatron side (a `torch.autograd.Function` around the
  frontend's forward/backward wrappers); the frontend APIs are pure
  kernels-plus-validation.

### Dispatch / gating (unchanged semantics vs the previous revision)

The fast path engages only for: THD packed non-pre-grouped path, `compress_ratio == 4`
(`coff == 2`, the production CSA/HCA configuration), bf16 `kv`/`score`, fp32 `ape`,
compute capability **10.0** (the frontend's validated envelope), int32 flat offsets
(`total_tokens * coff * head_dim < 2**31`). Everything else — SBHD, the pre-grouped
CP-prep path, `compress_ratio == 128` (functionally supported by the frontend kernels,
stays on eager until tuned), missing/old frontend, other devices — keeps the eager
implementation. `MCORE_CSA_FUSED_COMPRESSOR=0` disables the dispatch entirely. The
dispatch also keeps the eager path under `torch.use_deterministic_algorithms(True)`
(`dAPE` is accumulated with fp32 atomics in the fused backward) and under
`torch.compile` tracing (the frontend launch path takes raw pointers; eager lets the
compiler fuse the region itself). CUDA graphs: capture-compatible after a one-step
eager warmup per `(ratio, head_dim, coff)` configuration (a first call that would JIT
under capture raises loudly, per the frontend); the dispatch passes the caller's static
`fixed_total_comp` capacity through without device synchronization.

### Numerics

The numerics contract is the frontend kernels' (measured and tested in
NVIDIA/cudnn-frontend#427; original analysis in #5968): all arithmetic fp32 with a
single final bf16 rounding, `mul.rn`/`fma.rn` pinned in PTX. vs an fp32-intermediate
eager reference: `dKV`/`dScore` **bit-identical**, forward within one bf16 rounding
step. Not bit-identical to the current eager region (which rounds softmax weights to
bf16 and multiplies in bf16) but at least as accurate against an fp64 oracle on every
tested output. Forward, `dKV`, `dScore` are bitwise run-to-run deterministic; `dAPE` is
not (fp32 atomics) — hence the deterministic-mode fallback above. Incoming gradients on
static-capacity padding rows are ignored (they are tail padding, not consumed
downstream); never-consumed input elements get exact zeros, matching autograd. These
gates are re-verified here at the dispatch level (see Tests).

### Performance

The kernels this PR dispatches to are the ones measured in **NVIDIA/cudnn-frontend#427**
(same B200 methodology as before; the port also picked up two additional optimization
commits there — forward 32-bit vectorized access and backward kernel-side zero-writes —
so the numbers are *better* than this PR's previous revision). From #427, isolated GPU
kernel time (nsys, sum of kernel durations, 50 iterations after 20 warmup; THD packs of
8192-token sequences, `ratio = 4`, `coff = 2`, bf16, vs the verbatim eager region of
`Compressor._forward_thd`):

| THD pack | head_dim | eager fwd | fused fwd | fwd | eager bwd | fused bwd | bwd |
|---|---|---|---|---|---|---|---|
| 1×8192 | 128 | 117.8 µs | 4.5 µs | **26.5×** | 187.2 µs | 12.8 µs | **14.6×** |
| 3×8192 | 128 | 229.8 µs | 10.0 µs | **23.0×** | 352.7 µs | 22.2 µs | **15.9×** |
| 1×8192 | 512 | 263.3 µs | 12.4 µs | **21.2×** | 425.0 µs | 22.8 µs | **18.6×** |
| 3×8192 | 512 | 664.3 µs | 35.0 µs | **19.0×** | 1155.8 µs | 66.0 µs | **17.5×** |

Wall-clock numbers, the hardware-ceiling audit (DRAM traffic within 1% of algorithmic
bytes), and per-commit bitwise gates are in #427's description.

### Tests

`tests/unit_tests/transformer/experimental_attention_variant/test_csa_fused_compressor.py`
— trimmed to the Megatron-side wiring (kernel-level coverage — zero-write ownership,
CUDA-graph capture/replay, determinism, `coff == 1`, fp64 oracle — lives in #427's test
suite):

- numerics of the dispatched fused region vs the eager region, with the same gates as
  before (`dKV`/`dScore` bitwise vs an fp32-intermediate reference; forward within one
  bf16 rounding step; tolerance vs verbatim upstream numerics) over ragged THD packs
  including segments shorter than `ratio`;
- `fixed_total_comp` static-capacity padding through the dispatch (forward semantics +
  ignored padding-row gradients);
- dispatch gating and eager fallback: kill switch, **missing/old cudnn-frontend** (no
  `cudnn.csa` → silent eager, verified bitwise-identical to the kill-switch path at the
  `Compressor._forward_thd` level), deterministic mode (dispatch falls back; enabling
  deterministic mode between forward and backward raises loudly in the frontend),
  `ratio == 128`, non-bf16, unexpected layout, empty output;
- `Compressor._forward_thd` integration: fused engages (spy), matches eager, gradients
  flow to inputs and `ape`.

The module keeps `pytestmark = pytest.mark.launch_on_gb200`; without CUDA, without a
cudnn-frontend that provides `cudnn.csa`, or off CC 10.0 every test skips cleanly (the
GB200 CI lane needs `nvidia-cudnn-frontend` >= 1.27 — the first release containing
#427 — plus its `cutedsl` extra to actually exercise the fused path; until then the
tests skip rather than fail).

Full runs on 1×B200 with cudnn-frontend `develop` @ `b950af1` (the #427 merge)
installed from source: the new test file
8 passed; the related suites (`test_attention_variant_csa.py`,
`test_csa_cp_layout_kernels.py`, `test_csa_cp_utils.py`) with the fused dispatch live:
123 passed, 1 skipped (needs ≥2 ranks), 0 failed. With no cudnn-frontend installed:
8 skipped, related suites unaffected (eager path is untouched).

### Notes

- The fast path stays **on by default** for the supported configuration (kill switch
  `MCORE_CSA_FUSED_COMPRESSOR=0`), as in the previous revision — happy to flip to
  opt-in if you prefer.
- NVIDIA/cudnn-frontend#427 has **merged** (`develop` @ `b950af1`, 2026-07-30), so
  the ordering constraint from the previous revision is satisfied; this PR remains a
  no-op (eager everywhere) against any cudnn-frontend that predates the CSA API.
- Follow-ups tracked on the frontend side: deterministic `dAPE` reduction for the
  ratio=4 backward, wider arch enablement after validation. (The dedicated
  `compress_ratio == 128` kernels, previously listed here, landed in #427 itself.)

### DIFF
--- a/megatron/core/transformer/experimental_attention_variant/csa_fused_compressor.py
+++ b/megatron/core/transformer/experimental_attention_variant/csa_fused_compressor.py
@@ -27,16 +27,29 @@ bit-identical to the eager region (which rounds the softmax weights to bf16 and
 multiplies in bf16) but at least as accurate against an fp64 oracle; forward, ``dKV``
 and ``dScore`` are bitwise run-to-run deterministic, ``dAPE`` is accumulated with fp32
 atomics and is not — the dispatch therefore keeps the eager path under
-``torch.use_deterministic_algorithms(True)``. Measurements and numerics analysis:
-cudnn-frontend PR #427 and https://github.com/NVIDIA/Megatron-LM/issues/5968.
+``torch.use_deterministic_algorithms(True)``. The two ratio families carry different
+contracts against that fp32-intermediate eager reference: at ``compress_ratio == 4``
+``dKV``/``dScore`` are bit-identical to it, while at ``compress_ratio == 128`` they are
+faithful within the r128 gate tolerances (differing elements <= max(1, 0.1%), max_abs
+<= 1.6e-2 on the frontend gate's documented input distribution). Keeping the softmax
+weights in fp32 also makes the fused output measurably closer to an fp64 oracle than the
+region it replaces: on a 4096-token pack the fused forward's max absolute error against
+that oracle is 1.4-2.6x smaller than the eager region's across ``compress_ratio in
+{4, 128}`` x ``coff in {1, 2}`` x ``head_dim in {128, 512}``, and matches an
+fp32-intermediate eager reference (what remains is the single final bf16 rounding).
+Measurements and numerics analysis: cudnn-frontend PR #427 and
+https://github.com/NVIDIA/Megatron-LM/issues/5968.
 
-Dispatch gating (initial; everything else keeps eager):
-  - ``MCORE_CSA_FUSED_COMPRESSOR=0`` kill switch (disables the dispatch entirely);
+Dispatch gating (everything else keeps eager):
+  - the caller's ``use_fused_dsa_kernels(config)`` decision, i.e. the same switch that
+    gates the other optional CSA/DSA fused kernels (``Compressor.use_fused_compressor``);
   - cudnn-frontend with the CSA compressor API importable, CUDA device with compute
     capability 10.0 (the frontend's validated envelope);
-  - ``compress_ratio == 4`` / ``coff == 2`` (the production CSA/HCA configuration;
-    ``compress_ratio == 128`` stays on eager pending tuning), bf16 ``kv``/``score``,
-    fp32 ``ape``, int32 flat offsets (``total_tokens * coff * head_dim < 2**31``);
+  - ``compress_ratio in {4, 128}`` and ``coff in {1, 2}`` (the frontend's validated
+    envelope; ``Compressor`` itself only produces ``(4, 2)`` and ``(128, 1)``), with
+    ``compress_ratio == 128`` additionally restricted to ``head_dim in {128, 512}``
+    (the r128 kernels' validated head dims); bf16 ``kv``/``score``, fp32 ``ape``, int32
+    flat offsets (``total_tokens * coff * head_dim < 2**31``);
   - eager under deterministic mode (``dAPE`` atomics, above) and under
     ``torch.compile`` tracing (the frontend launch path takes raw pointers; eager lets
     the compiler fuse the region itself).
@@ -48,17 +61,31 @@ configuration before capture (a first call that would JIT under capture raises a
 static ``total_comp`` capacity through, so no device synchronization is introduced.
 """
 
-import os
+import logging
 from typing import Optional
 
 import torch
 
-_ENV_ENABLE = "MCORE_CSA_FUSED_COMPRESSOR"
+logger = logging.getLogger(__name__)
 
 # The compute capability the frontend kernels are validated for (its ``check_support``
 # raises on anything else); the dispatch pre-filters so other devices keep eager
 # silently. Mirrors ``cudnn.csa.compressor``'s envelope — widen together with it.
 _SUPPORTED_COMPUTE_CAPABILITY = (10, 0)
+
+# Mirrors ``cudnn.csa.compressor``'s validated envelope: ``ratio in {4, 128}`` x
+# ``coff in {1, 2}``. ``Compressor`` currently only produces (4, 2) and (128, 1) --
+# ``overlap`` is derived from ``compress_ratio`` -- but the gate follows the kernels
+# rather than that derivation, so a future overlap-policy change keeps the fast path
+# instead of silently falling back to eager. Widen together with the frontend.
+_SUPPORTED_RATIOS = frozenset({4, 128})
+_SUPPORTED_COFF = frozenset({1, 2})
+# ratio 128 is served by the frontend's dedicated r128 kernels, validated for these
+# head dims.
+_R128_HEAD_DIMS = frozenset({128, 512})
+
+# One-shot guard so a missing frontend warns once per process, not once per layer-call.
+_warned_missing_frontend = False
 
 _UNINITIALIZED = object()
 # Lazily resolved ``cudnn.csa.compressor`` module: ``_UNINITIALIZED`` -> not probed yet,
@@ -96,15 +123,13 @@ def _get_frontend():
     return _frontend
 
 
-def _dispatch_enabled() -> bool:
-    """Return True unless the fused compressor is disabled via environment variable."""
-    return os.environ.get(_ENV_ENABLE, "1").lower() not in ("0", "false", "off")
-
-
 def fused_compressor_available(device: Optional[torch.device] = None) -> bool:
-    """Return True when the fused kernels can run: frontend importable + supported device."""
-    if _get_frontend() is None:
-        return False
+    """Return True when the fused kernels can run: supported device + frontend importable.
+
+    The device is checked first so the missing-frontend warning below is only emitted
+    where the kernels could actually have run (compute capability 10.0); on every other
+    device the eager path is the expected outcome, not a misconfiguration.
+    """
     try:
         if not torch.cuda.is_available():
             return False
@@ -120,7 +145,22 @@ def fused_compressor_available(device: Optional[torch.device] = None) -> bool:
     if supported is None:
         supported = torch.cuda.get_device_capability(index) == _SUPPORTED_COMPUTE_CAPABILITY
         _DEVICE_SUPPORTED_CACHE[index] = supported
-    return supported
+    if not supported:
+        return False
+    if _get_frontend() is None:
+        global _warned_missing_frontend
+        if not _warned_missing_frontend:
+            _warned_missing_frontend = True
+            logger.warning(
+                "CSA fused compressor is enabled and this device is supported, but "
+                "cudnn-frontend's CSA compressor API is unavailable (%s: %s); keeping the "
+                "eager compressor. Install nvidia-cudnn-frontend (with the 'cutedsl' "
+                "extra) to enable the fused path.",
+                type(_frontend_error).__name__,
+                _frontend_error,
+            )
+        return False
+    return True
 
 
 class _CompressThdFused(torch.autograd.Function):
@@ -188,6 +228,7 @@ def maybe_compress_thd_fused(
     ratio: int,
     head_dim: int,
     coff: int,
+    enabled: bool = True,
 ) -> Optional[torch.Tensor]:
     """Dispatch helper for ``Compressor._forward_thd``: fused result or None (use eager).
 
@@ -197,13 +238,15 @@ def maybe_compress_thd_fused(
     ``check_support`` re-validates the envelope and raises (instead of falling back) on
     anything the gates below let through.
     """
-    if not _dispatch_enabled():
+    if not enabled:
         return None
     if kv.device.type != "cuda":
         return None
     if not fused_compressor_available(kv.device):
         return None
-    if ratio != 4 or coff != 2:
+    if ratio not in _SUPPORTED_RATIOS or coff not in _SUPPORTED_COFF:
+        return None
+    if ratio == 128 and head_dim not in _R128_HEAD_DIMS:
         return None
     if kv.dtype != torch.bfloat16 or score.dtype != torch.bfloat16:
         return None
--- a/megatron/core/transformer/experimental_attention_variant/csa_fused_compressor.py
+++ b/megatron/core/transformer/experimental_attention_variant/csa_fused_compressor.py
@@ -1,0 +1,241 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+"""Dispatch shim for the fused CSA/HCA ``Compressor`` gated-pooling kernels.
+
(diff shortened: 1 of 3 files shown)


hxbai:
Hi @zkyue Thanks for your contribution!

I think it is better to put the kernels in cudnn-frontend like other CSA/HCA fused kernels. Is it convenient for you to do so, or do you need our help to port the kernel?

zkyue:
@hxbai Sure, happy to do that — the kernels are self-contained CuTe DSL and should map cleanly onto the cudnn-frontend python API structure (we've contributed there before). Plan: I'll open a PR against cudnn-frontend porting the fwd+bwd kernels (APIBase-style wrapper, THD semantics, tests), then rework this PR into a thin dispatch that uses the cudnn-frontend implementation when available, keeping the eager fallback. Will link the kernel PR here once it's up.

zkyue:
@hxbai The cudnn-frontend kernel PR is up: NVIDIA/cudnn-frontend#427 (fwd+bwd kernels, APIBase-style wrapper, THD semantics, 40 tests; outputs bit-identical to the kernels in this PR). Once it lands I'll rework this PR into the thin dispatch that uses the cudnn-frontend implementation when available, keeping the eager fallback.

zkyue:
cudnn-frontend #427 merged this morning (`develop` @ `b950af1`), so this PR is now reworked to the thin dispatch as discussed: the kernel module and the benchmark harness live upstream in NVIDIA/cudnn-frontend#427, and what remains here is the ~240-line dispatch shim, the `csa.py` edit, and the trimmed test file — **+732/−37** vs `dev` (previously +1930/−37).

Force-pushed `82175427e` → `e073b2822` (rebased onto current `dev`); title and PR body updated accordingly. Gates re-run against merged `develop` @ `b950af1`: dispatch tests **8 passed** (8 skipped when no cudnn-frontend with `cudnn.csa` is on the path — silent eager fallback, no hard dependency), related suites **123 passed / 1 skipped**, pinned lint clean.

megatron/core/transformer/experimental_attention_variant/csa_fused_compressor.py
```diff
@@ -0,0 +1,241 @@
...
+    ``check_support`` re-validates the envelope and raises (instead of falling back) on
+    anything the gates below let through.
+    """
+    if not _dispatch_enabled():
+        return None
+    if kv.device.type != "cuda":
+        return None
+    if not fused_compressor_available(kv.device):
+        return None
+    if ratio != 4 or coff != 2:
```
hxbai: There are kernels supporting `ratio==128 and coff == 1` now, right?

megatron/core/transformer/experimental_attention_variant/csa_fused_compressor.py
```diff
@@ -0,0 +1,241 @@
...
+        # A partially installed frontend or DSL can raise more than ImportError.
+        except Exception as e:  # pylint: disable=broad-except
+            _frontend = None
+            _frontend_error = e
+    return _frontend
+
+
+def _dispatch_enabled() -> bool:
+    """Return True unless the fused compressor is disabled via environment variable."""
+    return os.environ.get(_ENV_ENABLE, "1").lower() not in ("0", "false", "off")
```
hxbai: Could you reuse the arg like here https://github.com/zkyue/Megatron-LM/blob/e073b2822b183859a692636f666e1356689ba33c/megatron/core/transformer/experimental_attention_variant/csa.py#L1613 rather than use a new envvar?

megatron/core/transformer/experimental_attention_variant/csa_fused_compressor.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def fused_compressor_available(device: Optional[torch.device] = None) -> bool:
```
hxbai, line 126: need a warning if the lib is not available

zkyue:
@hxbai Thanks for the review — all three addressed and pushed. Everything below was measured on a B200 (CC 10.0, cudnn-frontend at #427, CUDA 13).

### 1. ratio 128 / coff 1

Yes — and I widened the gate to the frontend's full validated envelope rather than adding just that one pair: `compress_ratio in {4, 128}` x `coff in {1, 2}`, with ratio 128 additionally restricted to `head_dim in {128, 512}` (the r128 kernels' validated head dims). `Compressor` only produces `(4, 2)` and `(128, 1)` today because `overlap` is derived from `compress_ratio`, but the dispatch now follows the kernels rather than that derivation — if the overlap policy ever changes, the fast path keeps working instead of silently falling back. All four combinations are covered by the unit test and were measured.

**Speed.** The region captured into CUDA graphs, both sides symmetrically (forward+backward as one replay). This removes the per-op launch overhead from eager and fused alike, which is the fairest single basis for these launch-bound shapes:

| ratio 128 / coff 1 | eager | fused | speedup |
|---|---|---|---|
| 1 x 8192, d=128 | 244.9 us | 20.7 us | **11.8x** |
| 3 x 8192, d=128 | 327.7 us | 30.7 us | **10.7x** |
| 1 x 8192, d=512 | 370.6 us | 33.2 us | **11.2x** |
| 3 x 8192, d=512 | 635.0 us | 68.1 us | **9.3x** |
| 1 x 131072, d=128 | 923.1 us | 88.2 us | **10.5x** |

Isolated kernel time (nsys, sum of kernel durations per iteration, 50 iters after 20 warmup, backward includes the dAPE zero-fill) — same methodology as the ratio-4 table in `docs/fe-oss-apis/csa.md`:

| ratio 128 / coff 1, 1 x 8192 | eager fwd | fused fwd | eager bwd | fused bwd |
|---|---|---|---|---|
| d=128 | 77.5 us | 5.7 us (**13.6x**) | 124.0 us | 9.4 us (**13.2x**) |
| d=512 | 116.2 us | 8.9 us (**13.1x**) | 192.2 us | 20.0 us (**9.6x**) |

> **Corrected 2026-07-31.** The backward columns above first read 203.2 / 9.4 us (21.6x) and 314.0 / 19.9 us (15.8x). That run timed the eager backward with the forward pass *inside* the timed region (and with input clones), while the fused side timed only the backward wrapper — an asymmetric comparison on my side. The table now shows a re-measurement with the forward outside the timed region, matching the `csa.md` methodology. The forward columns are unchanged, and the CUDA-graph and per-call tables are unaffected (those come from the frontend's own harness, which captures both sides symmetrically). The corrected figures also restore internal consistency: the eager kernel sum, 77.5 + 124.0 = 201.5 us, is now below the eager total-graph replay of 244.9 us, as it has to be. Kernel counts, for reference: eager 18 forward / 16 backward kernel types, fused 1 / 2 — the second fused backward kernel is the dAPE zero-fill.

Per-call wall clock (CUDA events, launch overhead included) is 5.0-5.3x at 1x8192/d128 and 5.8-11.1x at 1x131072/d128. The three tables are different measurement bases and are not comparable to one another.

As a check on the harness, I re-ran the published ratio-4 / coff-2 configuration on the same box: it reproduces the per-call and graph tables in `csa.md` within +/-2% on every cell except `fused fwd-graph` at 1x8192/d128 (10.8 -> 13.8 us), the smallest quantity in the set and the one the doc already flags as jittery.

The `# compress_ratio 128 stays on eager ... not yet a wall-clock win at production sizes` note in the test was stale and is gone.

**Numerics.** Worth stating explicitly, since the two ratio families differ: at ratio 4, dKV/dScore are bit-identical to the fp32-intermediate eager autograd; at ratio 128 it is a tolerance contract (frontend gate: differing elements <= max(1, 0.1%), max_abs <= 1.6e-2). Both are now in the module docstring. In the other direction, keeping the softmax weights in fp32 makes the fused path *more* accurate than the region it replaces — max absolute error against an fp64 oracle on a 4096-token pack:

| | fused | eager region | fp32-weight eager |
|---|---|---|---|
| ratio 4, coff 2, d=128 | 7.79e-3 | 1.69e-2 | 7.79e-3 |
| ratio 128, coff 1, d=128 | 2.49e-3 | 4.36e-3 | 2.49e-3 |

Across all eight (ratio, coff, head_dim) combinations the fused error is 1.4-2.6x smaller than the eager region's and matches the fp32-intermediate reference — what remains is the single final bf16 rounding.

### 2. Config arg instead of the env var

Done — `MCORE_CSA_FUSED_COMPRESSOR` and `_dispatch_enabled()` are gone. `Compressor.__init__` now sets `self.use_fused_compressor = use_fused_dsa_kernels(config)`, and the dispatch takes it as `enabled=`.

One consequence worth confirming, because it changes the default: `use_fused_dsa_kernels` requires `dsa_kernel_backend != "none"`, and that field defaults to `"none"` and is documented as the *ordinary-DSA* kernel backend selector. So the fused compressor is now off unless a DSA kernel backend is selected, and selecting `"tilelang"` would enable the cudnn compressor. If you would rather it were independent of that choice, I can add a dedicated field instead — happy either way, just say which.

### 3. Warning when the library is unavailable

Added, with two refinements so it stays signal rather than noise: the device-capability check now runs *before* the frontend probe, so the warning only fires where the kernels could actually have run (CC 10.0) instead of on every H100/A100 forward; and it is one-shot per process, with the underlying import error included.

### Validation

`tests/unit_tests/.../test_csa_fused_compressor.py` passes 8/8 on B200, including new coverage that all four (ratio, coff) combinations dispatch and that out-of-envelope head dims and ratios still fall back. Linters clean at the pinned versions (black 24.4.2, isort 5.13.2, pylint 10.00/10).

Note that the push changes the head SHA, so the CI authorization on `e073b282` no longer applies — a fresh `/ok to test` would be needed when you get a chance.

hxbai (approved):
LGTM

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/5984/commits/bdfcca9e523f94290581058240fa23f8be6ba35d

PR #5985, opened 2026-07-23
Pre-Fetch DP-Outer AG post-optimizer step

### LINKED ISSUE #4130: Megatron-FSDP Pre-Fetch DP-Outer AG post-optimizer step.

- Megatron-FSDP currently all-gathers fully-sharded model weights and partially-sharded model weights subsequently on the first microbatch after an optimizer step when using HFSDP. This can be pipelined to spread the communication cost over multiple FSDP communication groups and IB/NVL domains, because the DP-Inner AG is only dependent on the DP-Outer AG:
  - DP-Outer AG Unit 2 + 1 & DP-Inner AG Unit 1
  - DP-Outer AG Unit 3 & DP-Inner AG Unit 2
  - DP-Outer AG Unit 4 & DP-Inner AG Unit 3
  - etc.

- Simply pre-fetch the buckets 2 FSDP units downstream for DP-Outer AG, and fetch the buckets 1 FSDP unit downstream for DP-Inner AG, during Megatron-FSDP's AG hooks. May improve performance for GradAccum=1.

- There may be communication performance considerations when implementing this strategy. Likely should be toggled by an argument for further testing.

**Additional context**

Randl:

# What does this PR do?

Pipeline HFSDP parameter gathers: DP-Outer gathers are prefetched one more unit ahead to speed up subsequent DP-Inner gathers.

This PR is better reviewed commit-by-commit. First commit refactors some common code to prepare the additional functionality. The second commit adds the pipeline functionality. The third commit adds relevant flags to all necessary places.

I've run some small-scale experiments locally, with Llama 3 8b for 50 iterations, and seen a measurable speedup of ~5%. However, for fused fp8 attentions, I also noticed some increase in loss dispersion (none in deterministic mode).

## Issue tracking
Fixes https://github.com/NVIDIA/Megatron-LM/issues/4130
cc @cspades 

### Pre-checks


cspades:
@Randl Really clean code, promise will review when I have time! This is something Selvaraj and I had come up with a long while ago, completely forgot about it.

There's also MFSDP V2 you can try contributing this too, since V1 is in maintenance mode due to code quality, there'll be more reviewers who have bandwidth to review, V1 requires some caution on my side, need to do all the normal profiling (understanding how the pre-fetch affects a variety of parallelism combinations, and make sure that no memory allocations have appeared though I think the AG pipeline should help do that for you) and tests. 😩

rapatel:
@Randl Thanks for the PR! You mentioned "with Llama 3 8b for 50 iterations ... seen a measurable speedup of ~5%" - could you share the number of gradient accumulation (GA) steps in this experiment? I believe a smaller GA would show a higher end-to-end speedup with this overlap. Thanks!

wujingyue:
Thanks for the PR! I'm looking into whether (and how) to apply a similar idea in MFSDP v2 (#5616).

> DP-Outer gathers are prefetched one more unit ahead to speed up subsequent DP-Inner gathers.

Where does the speedup come from? My understanding is that, even before this PR, DP-outer and DP-inner allgathers are launched on different streams, so DP-outer AG_{i+1} and DP-inner AG_i should already be able to overlap naturally.

Does this PR increase the amount of overlap in practice (e.g. by making overlap more likely or starting work earlier), or is there another mechanism behind the speedup?

Randl:
> could you share the number of gradient accumulation (GA) steps in this experiment? I believe a smaller GA would show a higher end-to-end speedup with this overlap. 

@rapatel Unfortunately, I do not have access to the machine I ran on for some time. I think I used GA=1, but I'm not 100% sure. Real speedups probably will be more modest

> Where does the speedup come from? 

@wujingyue If I understand correctly, it comes from both increasing the amount of overlap in practice by scheduling an extra outer unit and better dependency granularity (https://github.com/NVIDIA/Megatron-LM/pull/5985/changes#diff-da62f73a7a6a4ac7815ed316a147ba348a7915e35a2f4885ceaf1678e5f650fbR4648-R4652). Possibly some profiling could shed more light on the exact mechanism of the speedup.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -4348,6 +4350,15 @@ def reset(self, preserve_non_fsdp_units: bool = True):
...
                 self.wait_bucket_ready(bucket_id, bwd)
 
+        # A prefetched DP-Outer gather may not have reached DP-Inner if execution
+        # stopped early. It must finish before the optimizer updates its buffer.
+        synchronized_events = set()
+        for event in self.outer_bucket_ready_events.values():
+            if id(event) not in synchronized_events:
+                event.synchronize()
+                synchronized_events.add(id(event))
+        self.outer_bucket_ready_events.clear()
```
cspades: Quick update @Randl this is not CG-safe, just no host-device synchronizations allowed, and IIRC `reset` here is called within the full-CG capture scope, perhaps in `finalize_model_grads` or something.

Still no time at the moment, mind giving me a quick update on why we need to update `synchronized_events` here? Can this Python variable be avoided, i.e. can we not simply just wait on the DP-Outer AG event instead, which is CG capturable?

rapatel: Would this suffice:

```
        # A prefetched DP-Outer gather may not have reached DP-Inner if execution
        # stopped early. It must finish before the optimizer updates its buffer.
        # Use a device-side wait (not event.synchronize()): host sync is illegal on
        # events recorded during CUDA graph capture / full-iteration CUDA graphs.
        if self.outer_bucket_ready_events:
            current_stream = torch.cuda.current_stream()
            synchronized_events = set()
            for event in self.outer_bucket_ready_events.values():
                if id(event) not in synchronized_events:
                    current_stream.wait_event(event)
                    synchronized_events.add(id(event))
            self.outer_bucket_ready_events.clear()
```

wujingyue:
image

Here's my understanding of what the runtime schedule looks like **without** prefetching i+2. Nothing seems to prevent inner `AG_i` from overlapping outer `AG_{i+1}`. But I'm more than happy to be proven wrong! 

Note: AO = allocation of allgather output and DO = deallocation.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
rapatel: [hfsdp-prefetch-cg-fix.patch](https://github.com/user-attachments/files/30358873/hfsdp-prefetch-cg-fix.patch)

Applying this patch fixes the CG issue. I tested this using DSv3-MXFP8 with GA=8. Measured ~1.7% speedup.

Randl: I've applied the patch to my branch

wujingyue:
> Here's my understanding of what the runtime schedule looks like **without** prefetching i+2. Nothing seems to prevent inner `AG_i` from overlapping outer `AG_{i+1}`. But I'm more than happy to be proven wrong!
> 
> Note: AO = allocation of allgather output and DO = deallocation.

Recording an offline discussion with @cspades: without this PR, there's an additional indirect dependency from the `inner AG_{i}` to the outer `AG_{i+1}` through **buffer allocation**. That dependency prevents the two operations from overlapping.

rapatel (approved):
LGTM.

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -4439,73 +4599,34 @@ def all_gather_params(
             # the maximum limit of 2 buffers allocated at any point in time.
             self.buffer.ddp_config.fsdp_double_buffer
             and no_fsdp_units
-        ):
+        )
+        pipeline_hfsdp_gathers = (
+            self.buffer.ddp_config.hfsdp_param_gather_overlap
+            and self.buffer.dist_index.use_hybrid_fsdp
+            and self.buffer.ddp_config.outer_dp_sharding_strategy != "no_shard"
```
wujingyue: The public docstrings/README say this flag is effective only with `outer_dp_sharding_strategy='optim'`, but this predicate enables it for every non-`no_shard` outer strategy. Is the wider scope intentional?

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
```diff
@@ -4439,73 +4599,34 @@ def all_gather_params(
...
-                prefetch_all_gather_size = (
-                    sum(
-                        [
-                            parameter_groups[i].model_weight_buffer.bucket_index.size
-                            for i in ag_buckets
-                        ]
-                    )
-                    - base_all_gather_size
+                    inner_prefetch_units = min(1, max(0, 2 - len(double_buf_units)))
+                ag_buckets = self._extend_by_fsdp_units(
```
wujingyue: Enabling this flag also replaces the existing byte/element-budget prefetch policy with exactly one subsequent FSDP unit: `suggested_AG_prefetch_size` is not used on this path. That is a material DP-Inner scheduling and memory-policy change, not only the advertised extra DP-Outer prefetch. Is the fixed one-unit inner window required for the pipeline?

wujingyue (commented):
LGTM otherwise

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
Randl: That's equivalent for now but fair comment, I'll fix it

megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py
Randl: Yeah I think you're right

Randl:
Anything needed from me at this point?

wujingyue (approved):

guihong-nv (approved):

Phlip79 (approved):

rapatel:
FYI - We have a change under review (#6773) that lets outer_dp_sharding_strategy='optim' even when DP-inner strategy leaves the weights replicated (zero-1/zero-2 inner). With that, the inner weight buffer (I believe it's called `inner_dp_wbuf.data`) is the full replicated bucket rather than this rank's shard, and the gather has to target inner_dp_wbuf.get_shard_from_local_buffer() instead. There's nothing to to change here.. I'll likely wait for this to land and pick it up on rebase. Just want to put it on the radar. Thanks.

Randl:
Rebased on main. Still need approval from
@NVIDIA/core-adlr @NVIDIA/core-nemo  @NVIDIA/data-parallelism

wujingyue:
@Randl I pinged @NVIDIA/core-adlr internally. Other approvals have been granted. Thanks for waiting!

jaredcasper (approved):

PR #5988, opened 2026-07-23
Use explicit process groups for dataloader checkpoints

yashaswikarnati:
## Summary

- pass the checkpoint caller's tensor-, pipeline-, and data-parallel groups into external dataloader state saving
- select the single TP0/PP0 writer and run save barriers through those explicit module groups, while retaining MPU fallback behavior
- keep ranks whose `save_state()` returns `None` in both data-parallel barriers and skip only their file write

## Motivation

Heterogeneous training can use module-specific process groups that differ from the global MPU groups. Dataloader checkpoint saving needs to use the caller's groups both to select writers and to keep collectives scoped to the correct data-parallel group.

## Validation

- `isort` on both changed Python files
- Black check on the changed test file
- Ruff, `compileall`, and `git diff --check`
- OCI-HSG Slurm job `5570068`: `2 passed, 15 deselected`

### DIFF
--- a/megatron/training/checkpointing.py
+++ b/megatron/training/checkpointing.py
@@ -577,8 +577,15 @@ def save_checkpoint(iteration, model, optimizer, opt_param_scheduler, num_floati
     checkpoint_name = get_checkpoint_name(save_dir, iteration, release=release, pipeline_parallel=pipeline_parallel,
         tensor_rank=tensor_rank, pipeline_rank=pipeline_rank, expert_parallel=expert_parallel, expert_rank=expert_rank, return_base_dir=return_base_dir)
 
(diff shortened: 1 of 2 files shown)


deepakn94 (approved):

yashaswikarnati:
/ok to test [6c5cd75](https://github.com/NVIDIA/Megatron-LM/pull/5988/commits/6c5cd7539f3110d6797ee31aa27e0d10ea8331a7)

cuichenx (approved):

yashaswikarnati:
/ok to test [9dff2ac](https://github.com/NVIDIA/Megatron-LM/pull/5988/commits/9dff2ac5c4bed71e5e9f7a7ebc15e3ef708fa14f)

PR #5822, opened 2026-07-15
Make LRU prefix caching eviction policy only evict child blocks

santhnm2:
- [x] I, the PR author, have personally reviewed every line of this PR.

# What does this PR do?
The LRU prefix eviction policy only considered timestamps when selecting block to evict, but this could lead to parent blocks being evicted and leaving holes in the prefix chain. Instead, this PR ensures that parent <-> child relationships are taken into account by the LRU eviction policy so that only leaf node blocks are evicted. This PR also makes ref-count updates happen BEFORE allocation rather than AFTER allocation so that pending matched blocks are not accidentally evicted because their ref-counts were 0.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact @NVIDIA/mcore-oncall.

## Issue tracking

For PRs from open-source community contributors:

- **New features**: a linked issue is **required**. Please open a [feature request](https://github.com/NVIDIA/Megatron-LM/issues/new?template=feature_request.md) and reference it here before submitting the PR.
- **Small updates (bug fixes, minor improvements)**: a linked issue is **recommended** and will accelerate the PR review process.

Linked issue: 

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment @NVIDIA/mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

[...]

PR #5995, opened 2026-07-23
Fix LRU block accounting for resumed requests

lmcafee-nvidia:
## What does this PR do?

- Extracts the allocator's existing free-plus-evictable LRU capacity calculation
  so `is_memory_available()` and paused-request resume admission share one source
  of truth.
- Accounts for physical allocation using cumulative new-block demand separately
  from active-buffer block demand.
- Allows requests that need no new block to resume when the raw free pool is
  empty.
- Relies on the existing parent-safe LRU eviction path from #5822 when allocating
  blocks for resumed requests.
- Adds `run_prefix_cache_lru_resume_repro.sh`, an end-to-end reproducer that
  launches through `gpt_dynamic_inference.py`.

## Why?

`resume_paused_requests()` capped a count of resumable requests with
`total_avail`, which counts only raw free blocks. That both mixed request and
block units and ignored registered, ref-zero prefix-cache blocks that the LRU
allocator can evict. With a saturated raw pool, this could leave no active
request and trigger:

```text
active_request_count == 0 with no hidden chunked prefill
```

Evictability remains defined solely by the allocator: a cached block must have
`ref_count == 0` and a registered hash.

## Tests

DFW Slurm job `14286405`, one node / eight GPUs:

- `tools/autoformat.sh` check mode: Black, isort, pylint, and Ruff passed.
- Focused allocator availability regression passed on all ranks.
- Both new resume regressions passed on all ranks.
- Full `test_kv_block_allocator.py`: 19 passed per rank.
- Full `test_dynamic_prefix_caching.py`: 40 passed per rank.

DFW Slurm job `14288437`, one H100, identical end-to-end launcher on both
revisions:

- `main` at `51e915a404268f2f6eec37b81b0d286065fa0f4c`: exited 1 on request 3
  with the expected assertion after reporting `KV 2/4 blocks cached
  (2 evictable)`.
- This branch at `4f4af3f8953723c1bfca13efdf2593535804ad7a`: exited 0, completed all
  four requests in eight steps, and wrote the result JSON.


megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -3356,53 +3357,113 @@ def release_memory_blocks_from_request_indexes(self, request_indexes) -> None:
             sa._intermediate_block_ids_cpu[request_indexes] = -1
             sa._eos_cache_block_id_cpu[request_indexes] = -1
 
+    def _get_paused_request_count_within_block_budget(self) -> int:
+        """Count the left-most paused requests whose blocks fit the paused budget."""
+        if self.paused_request_count == 0:
```
kvareddy: I guess we can also check for block_budget  == 0

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -3356,53 +3357,113 @@ def release_memory_blocks_from_request_indexes(self, request_indexes) -> None:
...
+        request references account for its current allocator reference count.
+        """
+        suffix_request_count = request_end_idx - request_start_idx
+        releasable_counts = [0] * (suffix_request_count + 1)
+        if suffix_request_count == 0:
+            return releasable_counts
+
+        if not self.enable_prefix_caching:
+            releasable_count = 0
+            for suffix_count, request_idx in enumerate(
```
kvareddy: Can we avoid the loop here?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -3356,53 +3357,113 @@ def release_memory_blocks_from_request_indexes(self, request_indexes) -> None:
...
+            ):
+                block_ids = self.request_to_kv_block_ids[request_idx]
+                releasable_count += int(torch.count_nonzero(block_ids >= 0).item())
+                releasable_counts[suffix_count] = releasable_count
+            return releasable_counts
+
+        selected_reference_counts: dict[int, int] = {}
+        releasable_count = 0
+        block_ref_counts = self.kv_block_allocator.block_ref_counts
+        for suffix_count, request_idx in enumerate(
```
kvareddy: Can we avoid the nested for loop?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -3445,53 +3508,59 @@ def evict_overflow_paused_requests(
...
+            self.request_last_kv_block_offset[: self.paused_request_count]
+            >= self.block_size_tokens - 1 - self.num_speculative_tokens
+        )
+        overflow_needs_new_block = needs_new_block[
+            retained_paused_request_count : self.paused_request_count
+        ]
+        releasable_block_counts = self._get_releasable_block_counts(
+            retained_paused_request_count, self.paused_request_count
+        )
+        block_count_avail = self.kv_block_allocator.get_allocatable_count()
```
kvareddy, line 3581: Eviction should happen only when block_count_avail is zero, right?

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -3445,53 +3508,59 @@ def evict_overflow_paused_requests(
...
+
+        # Preserve current ordering: evict the smallest right-most suffix whose
+        # physical releases let every remaining overflow request resume.
+        evict_request_count = None
+        for candidate_evict_count in range(overflow_paused_request_count + 1):
+            survivor_count = overflow_paused_request_count - candidate_evict_count
+            if survivor_count > allowed_to_resume:
+                continue
+            survivor_new_block_count = int(overflow_needs_new_block[:survivor_count].sum().item())
```
kvareddy: I think we need to get rid of the continue check above and update survivor_new_block_count = min(allowed to resume, ......)

kvareddy: code is fine. ignore this comment

kvareddy (approved):

shanmugamr1992 (approved):

jaredcasper (approved):

asolergi-nv (approved):

PR #6015, opened 2026-07-24
[Dev] Align decoupled Muon FP8 parameter-gather paths with main

Wohox:

## Summary

This PR is now the `dev`-branch alignment counterpart of the decoupled Muon FP8
parameter-gather work in #5479 on `main`. It keeps the original #6015 LayerWise
checkpoint/param-gather fixes and ports the subsequent implementation cleanups and
performance work so the two branches use the same primitives and validation boundaries.

The refactor also incorporates the batched MXFP8 copy-back optimization from #6094.
`megatron/training/training.py` remains unchanged; the shared behavior lives in the DDP,
FP8, and optimizer layers that own it.

### FP8 parameter staging and copy-back

- Reuse the batched MXFP8 copy-back path from #6094, invoking the quantizer directly
  instead of routing every parameter through `aten::copy_` / `__torch_dispatch__`.
- Stage LayerWise FP32 masters directly into one flat BF16 transport tensor instead of
  allocating one BF16 tensor per parameter and flattening the temporary copies.
- Use one batched helper for BF16 -> MXFP8/blockwise copy-back, while preserving the
  single-parameter compatibility wrapper.
- Centralize consumption and clearing of Transformer Engine's high-precision
  initialization value in `pop_high_precision_init_val`, shared by DDP, Distributed
  Optimizer, LayerWise Optimizer, Float16 Optimizer, and Megatron FSDP call sites.
- Keep the Megatron FSDP fallback behavior independent: it reuses the shared helper but
  does not opt into the LayerWise gather topology.

### Validation and routing

- Detect only the LayerWise-supported parameter storage types (plain MXFP8 and
  blockwise FP8) instead of treating every generic Float8/GroupedTensor representation
  as supported.
- Reject `--moe-single-grouped-weight` for the compact LayerWise FP8 gather because the
  batched copy-back does not support Transformer Engine GroupedTensor storage.
- Preserve native FP32 routing and mixed BF16/FP8 bucket handling.
- Correct the overlap assertion for chunked optimizer-state/master-weight offload: a
  child-level pre-forward requirement does not imply that every bucket owned by that
  child has already been dispatched.

### Existing #6015 fixes retained

- Handle empty/padding-only `dp_reshardable` bucket shards in the decoupled compact
  LayerWise configuration.
- Avoid slot-0 assumptions in LayerWise parameter gather when another rank owns the
  first real parameter.
- Keep the FP8 parameter-gather tests non-vacuous and exercise deferred overlap paths.

## Checkpoint redesign boundary

The existing checkpoint implementation is retained by this PR; the cleaner
padding-free redesign is intentionally not implemented here.

The follow-up design has two relevant rank-local cases (Distributed Optimizer ownership
is already resolved before these buffers are constructed, so an "unowned bucket" case
is not part of the design):

1. A shard wholly beyond `numel_unpadded` is trailing DP-alignment padding and should be
   omitted from save and load.
2. A shard wholly inside an intra-parameter alignment gap is still inside the current
   dense bucket checkpoint extent. It should be omitted by making the coverage contract
   padding-aware, not by synthesizing checkpoint tensors. The change must preserve
   explicit bucket identity and construct the corresponding load request without relying
   on list position, so multi-bucket checkpoints round-trip by value.

This must be a scoped coverage rule for declared padding regions; globally disabling DCP
coverage validation would also accept missing optimizer state and is not safe.

## Test plan

H100:

- `tests/unit_tests/test_muon_decouple_fp8_param_gather.py`: **8 passed, 9 skipped**
  per rank.
- `tests/unit_tests/test_fp8_utils.py`: **2 passed, 3 skipped** per rank.
- `tests/unit_tests/distributed/test_fp8_param.py`: **10 passed, 12 skipped** per rank.
- `tests/unit_tests/distributed/mfsdp_v1/test_mfsdp_fully_shard.py`:
  **197 passed, 125 skipped, 6 xfailed** per rank. The six results are expected XFAILs,
  not failures introduced by this PR.
- `tests/unit_tests/dist_checkpointing/test_layer_wise_optimizer.py`:
  **292 passed, 161 skipped** per rank.

GB200:

- `tests/unit_tests/test_fp8_utils.py`: **5 passed** per rank, including the real MXFP8
  batched copy-back path.
- The full Muon file exposes two pre-existing test/baseline issues: chunked-offload
  numerical drift also reproduces on the pre-refactor `b9086f1a7` baseline, and the
  pending-gather test calls `finish_grad_sync()` after manually waiting the parameter
  gather handle. Neither is caused by this refactor.

Formatting and static checks:

- Black check passed for all changed Python files.
- `tools/autoformat.sh` scope passed Black, isort, pylint, and ruff.
- Copyright checks passed.
- `git diff --check` passed.

## Related

- Main alignment PR: #5479
- Batched MXFP8 copy-back: #6094


Wohox:
/ok to test [7db50e2](https://github.com/NVIDIA/Megatron-LM/pull/6015/commits/7db50e27081ffa62dd7ddef3276b4694bc746f6c)

megatron/core/dist_checkpointing/strategies/torch.py
```diff
@@ -403,6 +411,99 @@ def _restore_dict_types(x: Union[dict, list, Any], keys_template: Union[dict, li
...
+
+    regular_metadata = dataclasses.replace(
+        metadata,
+        state_dict_metadata={
+            key: value
+            for key, value in metadata.state_dict_metadata.items()
+            if key not in sparse_keys
+        },
+    )
+    if not _validate_global_plan(global_plan, regular_metadata):
```
kunlunl, line 500: This code assumes that `_validate_global_plan()` returns a boolean. But MCore declares torch>=2.6.0, and torch 2.12 changed this private helper to return a list of validation errors, where an empty list means the plan is valid. So on torch 2.12+, a valid plan returns [] and is incorrectly rejected here, causing any dp_reshardable save that reaches this manifest-aware validation to fail. Conversely, a non-empty error list is treated as success by this condition.

kunlunl (approved):

PR #6021, opened 2026-07-24
fix(cuda-graphs): MB-928 align DDP initialization with capture stream

svcnemo-autobot:
## Background

Full-iteration CUDA graph capture can fail when DDP-retained AccumulateGrad nodes reference a different initialization stream.

## What changed

- Use the shared capture stream only for `full_iteration` CUDA graphs.
- Preserve a dedicated DDP initialization stream for `none`, `local`, and `transformer_engine`.
- Cover both stream-selection paths and the retained-node regression.

## Details

Both DDP construction paths select the shared stream only for full-iteration capture. Other implementations retain the side-stream behavior required by Transformer Engine FP8 gradient setup.

## Tested

- `tools/autoformat.sh`
- `test_dist_utils.py` — 68 passed
- Exact failing H100 CI image: `test_fp8_param.py` — 9 passed, 10 skipped
- Exact failing H100 CI image: `test_full_cuda_graph.py` — 2 passed per rank

The separate GB200 failure was a launcher API read timeout before tests started.


cuichenx (approved):

yaoyu-33 (approved):

megatron/training/models/dist_utils.py
```diff
@@ -293,10 +293,9 @@ def _ddp_wrap(
         if not ddp_config.overlap_grad_reduce:
             ddp_config.bucket_size = None
 
-    # DDP initialization is required to be on a side-stream for the full-iteration CUDA graph.
-    #  this side-stream may be nested if being called from within the get_model function, but it
-    #  is here in case someone wants to use this directly outside of get_model.
-    ddp_stream = torch.cuda.Stream()
+    # DDP initialization must use the full-iteration capture stream so its retained
+    # AccumulateGrad nodes do not reference a different, non-capturing stream.
+    ddp_stream = get_shared_capture_stream()
```
lxb007981: Should we still create a cuda stream when `cuda_graph_impl` is `"none"`, `"local"` or `"transformer_engine"`?

jiemingz (approved):

megatron/training/models/dist_utils.py
```diff
@@ -294,11 +294,15 @@ def _ddp_wrap(
...
-    #  this side-stream may be nested if being called from within the get_model function, but it
-    #  is here in case someone wants to use this directly outside of get_model.
-    ddp_stream = torch.cuda.Stream()
+    if get_model_config(model[0]).cuda_graph_impl == "full_iteration":
+        # DDP initialization must use the full-iteration capture stream so its retained
+        # AccumulateGrad nodes do not reference a different, non-capturing stream.
+        ddp_stream = get_shared_capture_stream()
+    else:
+        # Preserve a dedicated initialization stream for all other implementations.
+        ddp_stream = torch.cuda.Stream()
```
lxb007981, line 303: Would it make sense to use:

- `full_iteration`: the shared capture stream
- `local` / `transformer_engine`: a dedicated side stream
- `none`: the default stream

Without CUDA graphs, forward/backward normally run on the default stream. Constructing DDP on a separate stream can associate its retained `AccumulateGrad` nodes with that stream, while incoming gradients are produced on the default stream, resulting in Pytorch warnings like:

```
The AccumulateGrad node's stream does not match the stream of the node that produced the incoming gradient.
```

The FP8 failures seen at `5caed0b` do not demonstrate that `none` requires a side stream. The initial TE CUDA-graph test failed during capture and leaked TE’s temporary class-level monkey-patch wrapper; the subsequent `main_grad` failures appear to be cascading failures from that leaked state.

Issue #5534, opened 2026-06-29
Support torch distributed checkpointing for MFSDP v2


wujingyue:
## Goal

Support saving and loading MFSDP v2 checkpoints using PyTorch Distributed Checkpoint (DCP).

## Background

MFSDP v2 introduces a per-module `fully_shard()` path with sharded parameter, gradient, and optimizer state layouts. Checkpointing needs to preserve those layouts well enough to resume training correctly.

The initial target should be the PyTorch Distributed Checkpoint format. Support for Megatron-LM's `torch_dist` checkpoint format should be tracked separately and is likely lower ROI for now.

## Non-goals

- Support for Megatron-LM's `torch_dist` checkpoint format.
- Checkpoint format conversion between DCP and `torch_dist`.

## Related issues and PRs

The items below are for the existing Megatron-FSDP implementation rather than MFSDP v2. They are listed only as references for prior checkpointing issues, fixes, and test coverage.

- #4129
- #4910
- #5265
- #4623
- #4753
- #4843
- #4748
- #3510
- #2805
- #4553

LucccaH:
Hi @ahmadki , just checking if there are any updates on the progress of this issue.

We’d like to start evaluating mFSDP v2 in our training stack, but complete checkpoint save/load support is currently one of the main blockers, especially for reliably resuming existing training jobs.

Do you have an updated status or a rough ETA for the checkpoint support? Thanks!

PR #6024, opened 2026-07-24
Add PyTorch DCP checkpoint save/load for MFSDP v2

ahmadki:

# What does this PR do?

Add PyTorch Distributed Checkpoint (DCP) save/load helpers for the experimental
Megatron-FSDP (v2) path.

After `fully_shard()`, a module's parameters rest as `DTensor` views over the
optimizer main-weight buffers, and the optimizer moments (`exp_avg`/`exp_avg_sq`)
are `DTensor`s on the same mesh, so they can be checkpointed with DCP. However, a
`FsdpParameterGroup` packs several parameters into one flat buffer with
least-common-multiple row padding, so a parameter's per-rank shard does **not**
tile like torch's canonical `Shard(0)`. A naked `dcp.save(model.state_dict())`
would therefore compute wrong global offsets and silently corrupt the checkpoint.

This PR adds `megatron_fsdp/experimental/checkpoint.py` with:

- `save_dcp_checkpoint` / `load_dcp_checkpoint` — attach true per-shard chunk
  metadata via `preprocess_state_dict_for_uneven_dtensor` (reused from the stable
  Megatron-FSDP path) before `dcp.save` / `dcp.load`.
- `materialize_optimizer_state` — allocate optimizer state slots so the in-place
  DCP load has destinations to write into.
- `resync_compute_weights` — refresh mixed-precision compute weights from the
  loaded main weights after load.

The functions are exported from the experimental package. This is the
foundational library piece for DCP checkpointing of Megatron-FSDP v2; follow-up
PRs wire it into `FullyShardedOptimizer` and the training loop.

## Issue tracking

Linked issue: Related to #5534

### Pre-checks


tests/unit_tests/distributed/mfsdp_v2/test_checkpoint.py
wujingyue, line 1: test_checkpoint.py

We don't plan to online load/save other formats at this moment.

tests/unit_tests/distributed/mfsdp_v2/test_dcp_checkpoint.py
```diff
@@ -0,0 +1,154 @@
...
+from torch import nn
+from torch.distributed.device_mesh import init_device_mesh
+from torch.distributed.tensor import DTensor
+
+from megatron.core.distributed.fsdp.src.megatron_fsdp.experimental import (
+    Flat,
+    Placements,
+    fully_shard,
+    fully_shard_optimizer,
+    load_dcp_checkpoint,
```
wujingyue:
```suggestion
    load_checkpoint,
```

tests/unit_tests/distributed/mfsdp_v2/test_dcp_checkpoint.py
```diff
@@ -0,0 +1,154 @@
...
+            else:
+                assert expected == actual, f"optim[{index}][{key}] scalar mismatch"
+
+    return local_nonempty
+
+
+# (param_dtype, mixed_precision_policy): plain fp32, and bf16 compute with fp32 master weights.
+_CONFIGS = {
+    "fp32": (torch.float32, None),
+    "bf16_fp32_master": (torch.bfloat16, MixedPrecisionPolicy(main_params_dtype=torch.float32)),
```
wujingyue: main_weight is by default fp32, so I don't think we need to pass in MixedPrecisionPolicy

tests/unit_tests/distributed/mfsdp_v2/test_dcp_checkpoint.py
```diff
@@ -0,0 +1,154 @@
...
+@pytest.mark.parametrize("config", list(_CONFIGS), ids=list(_CONFIGS))
+def test_dcp_roundtrip_flat_dp(distributed_setup, shared_checkpoint_dir, config):
+    """Saving then loading a flat-DP sharded model+optimizer restores state bit-exactly.
+
+    The fc1 group packs ``weight (16, 8)`` and ``bias (16,)`` into one flat buffer, so per-rank
+    shards do not tile like canonical ``Shard(0)`` (e.g. one rank owns no bias rows). This
+    exercises the uneven-DTensor metadata path in :func:`save_dcp_checkpoint`.
+    """
+    if distributed_setup.world_size < 2:
+        pytest.skip("This test requires at least 2 ranks.")
```
wujingyue: Really?

tests/unit_tests/distributed/mfsdp_v2/conftest.py
```diff
@@ -50,3 +51,32 @@ def distributed_setup() -> Iterator[DistributedSetup]:
...
+    if not dist.is_initialized():
+        return
+    if device.type == "cuda":
+        dist.barrier(device_ids=[device.index])
+    else:
+        dist.barrier()
+
+
+@pytest.fixture(scope="function")
+def shared_checkpoint_dir(distributed_setup, request) -> Iterator[str]:
```
wujingyue: Could we keep `shared_checkpoint_dir` and its `_barrier` helper in `test_dcp_checkpoint.py`? They are specific to this checkpoint roundtrip test and do not need to be exposed to every mFSDP v2 test through `conftest.py`.

tests/unit_tests/distributed/mfsdp_v2/test_dcp_checkpoint.py
```diff
@@ -0,0 +1,154 @@
...
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.fc2(torch.relu(self.fc1(x)))
+
+
+def _flat_placements() -> Placements:
+    return Placements(dp_axes=[0], parameter=[Flat()], gradient=[Flat()], optimizer=[Flat()])
+
+
+def _build_sharded(seed, mesh, device, *, param_dtype, mp_policy):
+    torch.manual_seed(seed)
```
wujingyue: Nit: we could directly initialize the weights to be different rather than **indirectly** via the seed. So the intention is more clear.

tests/unit_tests/distributed/mfsdp_v2/test_dcp_checkpoint.py
```diff
@@ -0,0 +1,154 @@
...
+    ((model(x) - target) ** 2).mean().backward()
+    optimizer.step()
+
+
+def _to_local(value):
+    return value.to_local() if isinstance(value, DTensor) else value
+
+
+def _snapshot(model, optimizer):
+    model_snap = {
```
wujingyue:
```suggestion
    model_snapshot = {
```

tests/unit_tests/distributed/mfsdp_v2/test_dcp_checkpoint.py
```diff
@@ -0,0 +1,154 @@
...
+        key: _to_local(value).detach().clone()
+        for key, value in model.state_dict().items()
+        if not key.endswith("_extra_state")
```
wujingyue: Why?

tests/unit_tests/distributed/mfsdp_v2/test_dcp_checkpoint.py
```diff
@@ -0,0 +1,154 @@
...
+    optimizer.zero_grad()
+    ((model(x) - target) ** 2).mean().backward()
+    optimizer.step()
+
+
+def _to_local(value):
+    return value.to_local() if isinstance(value, DTensor) else value
+
+
+def _snapshot(model, optimizer):
```
wujingyue:
```suggestion
def _snapshot_local_state(model, optimizer):
```

tests/unit_tests/distributed/mfsdp_v2/test_dcp_checkpoint.py
```diff
@@ -0,0 +1,154 @@
...
+    optim_snap = {}
+    for index, state in optimizer.state_dict()["state"].items():
+        optim_snap[index] = {
+            key: (_to_local(value).detach().clone() if torch.is_tensor(value) else value)
+            for key, value in state.items()
+        }
+    return model_snap, optim_snap
+
+
+def _assert_matches(snapshot, model, optimizer):
```
wujingyue:
```suggestion
def _assert_match_snapshot(model, optimizer, model_snapshot, optimizer_snapshot):
```

wujingyue: Nit: split into `_assert_model_matches_snapshot` and `_assert_optimizer_matches_snapshot`? This way, it's clear that `model` and `optimizer` don't interact during matching.

wujingyue: Type annotate and code-comment the return value. It's not obvious to readers.

wujingyue (commented):
Thanks for the PR! 

Here's the first batch of reviews focusing on the test. Will go through checkpoint.py soon...

tests/unit_tests/distributed/mfsdp_v2/conftest.py
```diff
@@ -50,3 +51,32 @@ def distributed_setup() -> Iterator[DistributedSetup]:
...
+    """A filesystem directory shared across ranks for DCP save/load.
+
+    All ranks run on one node and share the same path, derived deterministically from the test
+    name so no cross-rank broadcast is needed (the process group is not yet initialized when this
+    fixture runs). Rank 0 removes the tree on teardown.
+    """
+    root = os.environ.get("MFSDP2_DCP_TMP", "/tmp/mfsdp2_dcp_tests")
```
wujingyue: Could we reuse the existing `tmp_path_dist_ckpt` fixture and `TempNamedDir` here instead of introducing `MFSDP2_DCP_TMP`? They already provide a rank-shared checkpoint path, synchronization, and cleanup. That would also let us remove `shared_checkpoint_dir`, `_barrier`, and the new `shutil` import.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+    ``_extra_state`` holds opaque per-module bytes (for example FP8 scale history) that FSDP
+    does not manage as DTensors. Following the stable Megatron-FSDP DCP path, these are not
+    checkpointed here.
+    """
+    return {
+        key: value for key, value in model_state_dict.items() if not key.endswith("_extra_state")
+    }
+
+
+def _build_state_dict(model: torch.nn.Module, optimizer: Optional[torch.optim.Optimizer]) -> dict:
```
wujingyue:
```suggestion
def _build_state_dict(model: torch.nn.Module, optimizer: torch.optim.Optimizer | None) -> dict:
```
https://typing.python.org/en/latest/reference/best_practices.html#shorthand-syntax

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    for module in _iter_fsdp_modules(model):
+        for parameter_group in module.parameter_groups:
+            parameter_group.sync_model_weight_from_main_weight()
+
+
+def save_dcp_checkpoint(
+    model: torch.nn.Module,
+    optimizer: Optional[torch.optim.Optimizer],
+    checkpoint_id: Union[str, os.PathLike],
```
wujingyue:
```suggestion
    checkpoint_id: str | os.PathLike,
```
ditto

wujingyue:
```suggestion
    checkpoint_dir: Union[str, os.PathLike],
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+) -> None:
+    """Save a ``fully_shard``-wrapped model (and optionally its optimizer) as a DCP checkpoint.
+
+    Args:
+        model: A module tree that has been sharded with :func:`fully_shard`.
+        optimizer: Optimizer stepping the sharded parameters, or ``None`` to save only weights.
+        checkpoint_id: Destination directory for the DCP checkpoint.
+    """
+    state_dict = _annotate_uneven_dtensors(_build_state_dict(model, optimizer))
+    dcp.save(state_dict, checkpoint_id=str(checkpoint_id))
```
wujingyue:
```suggestion
    dcp.save(state_dict, checkpoint_id=checkpoint_id)
```
Apparently, dcp.save accepts Pathlike as well.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+def _build_state_dict(model: torch.nn.Module, optimizer: Optional[torch.optim.Optimizer]) -> dict:
+    state_dict = {"model": _strip_extra_state(model.state_dict())}
+    if optimizer is not None:
+        state_dict["optimizer"] = optimizer.state_dict()
+    return state_dict
+
+
+def _annotate_uneven_dtensors(state_dict: dict) -> dict:
+    """Attach uneven-DTensor chunk metadata to the model and optimizer sub-dicts in place."""
+    preprocess_state_dict_for_uneven_dtensor(state_dict["model"])
```
wujingyue: Non-blocking Q: Have you tried to understand what this function does? I failed... It's obviously from v1 and it's unclear to me what metadata is attached and why. The implementation also feels quite spaghetti. We may need to rewrite it from first principles for **just** what v2 needs; otherwise it'll soon become legacy that nobody wants to touch.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+    does not manage as DTensors. Following the stable Megatron-FSDP DCP path, these are not
+    checkpointed here.
+    """
+    return {
+        key: value for key, value in model_state_dict.items() if not key.endswith("_extra_state")
+    }
+
+
+def _build_state_dict(model: torch.nn.Module, optimizer: Optional[torch.optim.Optimizer]) -> dict:
+    state_dict = {"model": _strip_extra_state(model.state_dict())}
```
wujingyue: If code here already strips extra state, does it still make sense for the test to filter?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+    if optimizer is not None:
+        state_dict["optimizer"] = optimizer.state_dict()
```
wujingyue: I'd remove this function and let the caller handle `model` and `optimizer` separately. This way, it's clear to the readers that `model`'s state and `optimizer`'s state don't interact. 

Right before dcp.save, we pack both dicts into one **just** because we want to save them to the same path.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+                # A sharded parameter advertises the FSDP gradient dtype via ``grad_dtype``, which
+                # can differ from the (main-weight) parameter dtype under mixed precision. Match it
+                # so the temporary grad is consistent with the parameter's grad contract.
+                grad_dtype = getattr(param, "grad_dtype", param.dtype)
+                param.grad = torch.zeros_like(param, dtype=grad_dtype)
+    optimizer.step()
+    optimizer.zero_grad(set_to_none=True)
+
+
+def resync_compute_weights(model: torch.nn.Module) -> None:
```
wujingyue:
```suggestion
def sync_model_weight_from_main_weight(model: torch.nn.Module) -> None:
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def _strip_extra_state(model_state_dict: dict) -> dict:
+    """Drop TransformerEngine ``_extra_state`` entries.
+
+    ``_extra_state`` holds opaque per-module bytes (for example FP8 scale history) that FSDP
+    does not manage as DTensors. Following the stable Megatron-FSDP DCP path, these are not
+    checkpointed here.
+    """
+    return {
```
wujingyue: Could `_strip_extra_state` mutate `model_state_dict` in place and return `None`, matching the stable mFSDP path? `model.state_dict()` gives us a fresh mapping, and deleting keys in place preserves its `OrderedDict` metadata, which this comprehension drops. `_build_state_dict` can first assign `model.state_dict()`, call `_strip_extra_state(model_state_dict)`, and then store that mapping.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+    state_dict = {"model": _strip_extra_state(model.state_dict())}
+    if optimizer is not None:
+        state_dict["optimizer"] = optimizer.state_dict()
+    return state_dict
+
+
+def _annotate_uneven_dtensors(state_dict: dict) -> dict:
+    """Attach uneven-DTensor chunk metadata to the model and optimizer sub-dicts in place."""
+    preprocess_state_dict_for_uneven_dtensor(state_dict["model"])
+    if "optimizer" in state_dict:
```
wujingyue: Could we use `torch.distributed.checkpoint.state_dict.get_state_dict()` / `set_state_dict()` instead of maintaining custom optimizer-state materialization and raw optimizer state handling? `get_state_dict()` already initializes empty optimizer state, maps optimizer parameter IDs to stable FQNs, and preserves the DTensors; we can still apply `preprocess_state_dict_for_uneven_dtensor()` to the returned dictionaries. `set_state_dict()` would then replace the direct model/optimizer `load_state_dict()` calls. This follows the standard DCP flow and avoids duplicating optimizer-specific initialization logic.

wujingyue (commented):
The code structure looks good! My comments are mostly "why is X needed/complicated and can we simply" and improving readability. I hope most of them can be easily fixed by agents.

tests/unit_tests/distributed/mfsdp_v2/test_dcp_checkpoint.py
ahmadki: the whole point of the test is the *uneven* packing across
the DP mesh, which only exists with >=2 ranks. on 1 rank it becomes a trivial single-shard roundtrip.
Will add clarification to the skip message

wujingyue: > on 1 rank it becomes a trivial single-shard roundtrip.

I knew that but 1-rank shouldn't make the test fail either. Correct? So I'd remove the skip for fewer conditions and more coverage. Even a user with a single GPU can run a sanity check.

tests/unit_tests/distributed/mfsdp_v2/test_dcp_checkpoint.py
ahmadki: The filter mirrored `_strip_extra_state` in `checkpoint.py`, which is gone now

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
ahmadki: The function computes `ChunkStorageMetadata(offsets, sizes)` for every DTensor in the state dict, the rank's shard offset within the full tensor plus its local size. And it attaches it to `dtensor._local_tensor` with:

1. `__create_chunk_list__`: used for load/reshard planning
2. `__create_write_items__` : used for save planning

DCP's `DefaultSavePlanner` looks for those tags and when present, uses them instead of its built-in even `Shard(0)` offset formula.

We do need this logic in v2, a naked `dcp.save` would have a silently corrupted checkpoint. This is inherent to the flat packed DBuffer layout.

v1 implementation is generic, it walks an opaque state dict, handles `_StridedShard`, and the part that reads as "spaghetti" recovers each shard's offset at runtime via an `all_gather_object` of local shard shapes.
v2 does not need that gather, `GlobalLayout`/`DBuffer` already know each tensor's global element offset and this rank's owned row range (`_get_owned_range`), so the `ChunkStorageMetadata` can be computed without it

The re-write is on my todo list, but it's outside the scope of this PR.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
ahmadki: Done.
`_build_state_dict` is removed and `save_checkpoint`/`load_checkpoint` now build  `model_state_dict` and `optimizer_state_dict` independently and only pack them into:
```
{"model": ..., "optimizer": ...}
```
at the `dcp.save`/`dcp.load` call

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
ahmadki: Done for the most part. We still need a custom `_init_optimizer_state` though (explained in the function's docstring)

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
wujingyue: > The re-write is on my todo list, but it's outside the scope of this PR.

👍

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    if optimizer.state:
+        return
+    for group in optimizer.param_groups:
+        for param in group["params"]:
+            if param.grad is None:
+                grad_dtype = getattr(param, "grad_dtype", None) or param.dtype
+                param.grad = torch.zeros_like(param, dtype=grad_dtype)
+    optimizer.step()
+    optimizer.zero_grad(set_to_none=True)
```
wujingyue:
```suggestion
    optimizer.zero_grad()
```
set_to_none=True is the default. So I'd remove it unless there's a specific reason, e.g., we can't support set_to_none=False.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,138 @@
...
+    set_optimizer_state_dict,
+)
+
+from ..uneven_dtensor import preprocess_state_dict_for_uneven_dtensor
+from .module import FsdpModule
+
+__all__ = ["save_checkpoint", "load_checkpoint", "sync_model_weight_from_main_weight"]
+
+
+def sync_model_weight_from_main_weight(model: torch.nn.Module) -> None:
```
wujingyue:
```suggestion
def _sync_model_weight_from_main_weight(model: torch.nn.Module) -> None:
```

Do we need to expose it? Doesn't seem to be used elsewhere.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -0,0 +1,138 @@
...
+            for parameter_group in module.parameter_groups:
+                parameter_group.sync_model_weight_from_main_weight()
+
+
+def _init_optimizer_state(optimizer: torch.optim.Optimizer) -> None:
+    """Allocate optimizer state so a DCP load has DTensors to fill.
+
+    :func:`get_optimizer_state_dict` initializes empty optimizer state via torch's
+    ``_init_optim_state``, but that assigns a parameter-dtype gradient. A Megatron-FSDP sharded
+    parameter advertises the FSDP gradient dtype through ``grad_dtype``, which differs from the
```
wujingyue, line 71: Can we file a bug to PyTorch? https://github.com/pytorch/pytorch/blob/190d2809ecade0a89ea011ce255f9b1f1216ddfc/torch/distributed/checkpoint/state_dict.py#L649 should read grad_dtype, which will make this method unnecessary.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
wujingyue: Depending on private attributes like `_local_tensor` and `__create_chunk_list__` creates maintenance problems. We should push PyTorch to support uneven splits better so we don't have to hack around. 

Non-blocking AI: can you file an issue to PyTorch? Let's try to make the repro torch only -- no FSDP code. I believe there's a general limitation around uneven splits.

wujingyue (approved):
Wonderful!

tests/unit_tests/distributed/mfsdp_v2/test_checkpoint.py
```diff
@@ -0,0 +1,168 @@
...
+    """Assert the optimizer's local state equals the snapshot."""
+    current = optimizer.state_dict()["state"]
+    assert optimizer_snapshot.keys() == current.keys()
+    for index, expected_state in optimizer_snapshot.items():
+        for key, expected in expected_state.items():
+            actual = current[index][key]
+            if torch.is_tensor(expected):
+                actual = _to_local(actual)
+                assert expected.shape == actual.shape, f"optim[{index}][{key}] shape mismatch"
+                assert torch.equal(expected, actual), f"optim[{index}][{key}] value mismatch"
```
wujingyue: https://docs.pytorch.org/docs/2.13/testing.html#torch.testing.assert_close with rtol=atol=0?

tests/unit_tests/distributed/mfsdp_v2/test_checkpoint.py
```diff
@@ -0,0 +1,168 @@
...
+    optimizer: torch.optim.Optimizer, optimizer_snapshot: dict[int, dict]
+) -> None:
+    """Assert the optimizer's local state equals the snapshot."""
+    current = optimizer.state_dict()["state"]
+    assert optimizer_snapshot.keys() == current.keys()
+    for index, expected_state in optimizer_snapshot.items():
+        for key, expected in expected_state.items():
+            actual = current[index][key]
+            if torch.is_tensor(expected):
+                actual = _to_local(actual)
```
wujingyue: Is it possible to test DTensor, not just the local Tensor? For example, DTensors have global shapes that are important to get correct in the checkpoint.

wujingyue (approved):

wujingyue:
@ahmadki Once you've addressed the remaining comments, please ping the `#megatron-core-developments` channel for a final review. Most approvers haven't gotten into the habit of checking GitHub notifications yet.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
ahmadki: I will resolve this conversation and take the AI as a side task

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
ahmadki: Field a bug to PyTorch: https://github.com/pytorch/pytorch/issues/191918
And left a TODO to delete the function once this is resolved from PyTorch's side.

Phlip79 (approved):

jaredcasper (approved):

wujingyue:
CI failure is due to a conflict with my recently merged #6190: `fully_shard()` now requires `fully_shard_context()`, but the new checkpoint test calls it outside that context. Both cases fail before exercising DCP.

Proposed fix: #6353 wraps those calls in `fully_shard_context(device=device)`. Verified with the targeted 2-rank test (fp32 and bf16 pass).

PR #5833, opened 2026-07-15
Add encoder prefetch for heterogeneous MIMO training

yashaswikarnati:
## What changed

- add opt-in, configurable-depth prefetch for frozen MIMO encoder features with any compatible batch iterator.
- keep completed features in a bounded FIFO and cap unfinished encoder GPU work at one batch.
- overlap CPU generation of the next batch with the current GPU encode using one staged CPU batch.
- retain encoder-rank `input_ids` on CPU, removing an unnecessary H2D copy and pre-bridge GPU-to-host synchronization.
- keep the trainable projection on the schedule stream and inside autograd.

## Validation

- final cleanup bundle: Black, isort, Ruff, and 76 focused/surrounding MIMO tests passed on CW DFW `14223654`
- final 8-GPU heterogeneous mock smoke completed two iterations with finite loss/gradients and zero skipped/NaN iterations on CW DFW `14223688`; all eight rank logs were checked

PR #6039, opened 2026-07-25
Handle checkpoint and text-only batches in MIMO encoder prefetch

yashaswikarnati:
## What changed

- expose `EncoderPrefetchLoader.save_state() -> None` so checkpointing can treat encoder lookahead as derived state while the canonical language-loader cursor remains authoritative.
- allow text-only MIMO microbatches to pass through encoder prefetch without invoking the image encoder.
- guard CUDA stream and debug-timing bookkeeping when a batch produces no encoder features.

## Why

Encoder prefetch from #5833 can wrap a checkpointable data iterator and can receive image-sparse batches from heterogeneous multimodal datasets. The outer wrapper must participate in the checkpoint interface without persisting its speculative queue, and batches without images must preserve schedule progress without manufacturing encoder work.

The model still validates missing inputs when matching modality tokens are present; this only permits genuinely text-only batches.

## Validation

- `python3 -m py_compile examples/mimo/training/encoder_prefetch.py tests/unit_tests/models/mimo/test_mimo_encoder_prefetch.py`
- OCI-HSG 5-node Mistral VLM smoke job `5590459`: 200 iterations completed with both patches in the staging branch.
- focused regression coverage added for non-authoritative prefetch state and text-only batch pass-through.

Local pytest was not rerun on the Codex host because that host does not provide `uv` or Torch; CI should run the focused tests in the supported container.

### DIFF
--- a/examples/mimo/training/encoder_prefetch.py
+++ b/examples/mimo/training/encoder_prefetch.py
@@ -278,8 +278,8 @@ class EncoderPrefetchLoader:
                 self._condition.notify_all()
                 terminate = source_exhausted or source_error is not None
             if self._debug:
-                assert encode_start is not None
-                _log_producer_debug(batch_id, data_fetch_ms, encode_start, completion_event)
+                if encode_start is not None:
+                    _log_producer_debug(batch_id, data_fetch_ms, encode_start, completion_event)
                 self._drain_encoder_wait_timings()
                 self._drain_projection_timings()
             if terminate:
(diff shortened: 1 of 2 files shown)


yashaswikarnati:
/ok to test [d16427a](https://github.com/NVIDIA/Megatron-LM/pull/6039/commits/d16427a4610e690aa67fdd8656617c3149bcf9c1)

sancha (approved):

PR #6041, opened 2026-07-25
Rebind MFSDP v2 sharded grads once per gradient reduction

wujingyue:
## What

`FsdpParameterGroup.reduce_partial_gradients` (Megatron-FSDP v2 / experimental) installed the sharded-parameter `.grad` DTensors at up to three points per call, via an `_install_sharded_grads()` helper: after relabeling `main_grad` to the accumulation placement, after a fresh reduce into `main_grad`, and after the last-microbatch finalize.

This collapses them into a single unconditional rebind at the end of the method and inlines the (now single-caller) helper.

## Why it's safe

Within `reduce_partial_gradients`, `.grad` is read only for *presence* (`has_grad`, at the top); its values are consumed by the optimizer after the reduction completes. `main_grad` reaches its final identity by the end of the method (it is only reassigned by the two `redistribute` calls). So a single rebind at the end keeps every sharded grad pointed at the current `main_grad` buffer, with identical behavior — including cross-microbatch accumulate-vs-overwrite, since each microbatch still ends with `.grad` bound.

## Scope

Behavior-preserving refactor; no golden-value or API changes. Existing `tests/unit_tests/distributed/mfsdp_v2/` cover plain-DP and HSDP gradient reduction and accumulation.

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
@@ -250,12 +250,6 @@ class FsdpParameterGroup:
         # so keep the shared storage-release path.
         self._unsharded_model_weight.release_storage()
 
-    def _install_sharded_grads(self) -> None:
-        """Point each sharded parameter's grad at main_grad's current DTensor view."""
-        assert self.main_grad is not None
-        for index, sharded_parameter in enumerate(self.sharded_parameters):
-            sharded_parameter.grad = self.main_grad.get_dtensor(index)
-
     def allocate_partial_grad_buffer(self) -> DBuffer:
         """Allocate the unreduced reduce-scatter input buffer."""
         assert self.main_grad is not None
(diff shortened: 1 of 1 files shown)


wujingyue:
cc @Achyuthan-S Looks good to you?

Achyuthan-S:
Hey @wujingyue , looks good to me — traced it and it's behavior-preserving. The three rebinds were each only needed because main_grad gets reassigned by the reset/finalize redistributes; since .grad's values aren't read inside reduce_partial_gradients (only presence at the top) and the accumulation writes main_grad.local_buffer directly, a single unconditional rebind at the end lands .grad on the final main_grad identically — across set_to_none {True,False} and single/multi microbatch.

One coordination note: this overlaps with #5972 (HFSDP), which also touches this method — it adds a fresh-buffer reset branch for HFSDP's reduce-scattered finalize. They'll conflict textually, so whoever lands second rebases. This actually simplifies the HFSDP side (the fresh-buffer reset won't need its own rebind — your end rebind covers it), and I checked the single rebind works with the fresh buffer. Happy to rebase HFSDP on top once this merges.

shjwudp (approved):

Phlip79 (approved):

deepakn94 (approved):

PR #6043, opened 2026-07-25
Support RL eval-only runs

tdene:
- [x] I, the PR author, have personally reviewed every line of this PR.

# What does this PR do?
**IN SUMMARY, this confusion stems from the fact that** the `train` method does more than just training for RL, and the `--skip-train` flag is designed to only skip training (not the entire `train` method). In order to skip the actual `train` method altogether, we require `--train-iters 0` as well.

----

Allow RL eval-only runs to be run by adding `--skip-train --train-iters 0`. This is separate from the functionality of `--skip-train`, which performs no-training RL runs.

Currently in `main` there are two modes of operation for RL: normal, and `--skip-train` which skips only gradient updates but nothing else in the RL "training" loop (e.g. all the inference, refit, context-switching, still happens).

After this PR, there are three modes of operation for RL: normal, `--skip-train`, and this newly added `--skip-train --train-iters 0` which skips the entire RL "training" loop, instead doing only evaluation.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact @NVIDIA/mcore-oncall.

## Issue tracking

For PRs from open-source community contributors:

- **New features**: a linked issue is **required**. Please open a [feature request](https://github.com/NVIDIA/Megatron-LM/issues/new?template=feature_request.md) and reference it here before submitting the PR.
- **Small updates (bug fixes, minor improvements)**: a linked issue is **recommended** and will accelerate the PR review process.

Linked issue: 

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment @NVIDIA/mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

[...]

PR #6044, opened 2026-07-25
Miscellaneous fixes to the RL loop

tdene:

# What does this PR do?

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


megatron/rl/rl_utils.py
```diff
@@ -2061,7 +2113,18 @@ def evaluate_and_print_results_rl(
                         tb_writer.add_scalar(k, v, iteration)
             wandb_writer = get_wandb_writer()
             if wandb_writer:
-                wandb_writer.log(eval_metrics, step=iteration)
+                if args.do_train:
```
lauradang, line 2210: You probably need to explicitly check for the skip train flag as well:

```python
if args.do_train and not args.skip_train:
```

**Explanation:**
RL setup sets `do_train` solely from `train_iters`, without considering `skip_train`:

```python
if args.perform_rl_step:
    # we don't need to build any dataloaders for RL training
    train_dataloader = None
    valid_dataloaders = None
    test_dataloader = None
    do_train = (args.train_iters or 0) > 0
```

That value is then broadcast and stored in `args.do_train`:

```python
flags = torch.tensor(
    [int(do_train), int(do_valid), int(do_test)],
    dtype=torch.long,
    device='cuda',
)

torch.distributed.broadcast(flags, 0)

args.do_train = getattr(args, "do_train", False) or flags[0].item()
```

Later, `skip_train` independently prevents the gradient update:

```python
if args.skip_train:
    # RL inference-only mode: skip gradient updates, just collect rollouts.
    loss_dict = {}
    skipped_iter = 0
    should_checkpoint = False
    should_exit = False
    exit_code = 0
    grad_norm = 0.0
    num_zeros_in_grad = 0
    max_attention_logit = None
else:
    # ... train_step(...)
```

Therefore these values can coexist in RL inference-only mode:

```python
args.perform_rl_step = True
args.skip_train = True
args.train_iters > 0
args.do_train = True
```

As a result, this new condition still takes the training branch and logs with an explicit checkpoint iteration:

```python
if args.do_train:
    wandb_writer.log(eval_metrics, step=iteration)
else:
    # Intended eval-only handling
```

Could this instead key off `skip_train`, for example:

```python
if args.do_train and not args.skip_train:
    wandb_writer.log(eval_metrics, step=iteration)
else:
    # Eval-only handling
```

A regression test with `do_train=True` and `skip_train=True` would cover the inference-only configuration that currently bypasses the new handling.

megatron/rl/rl_utils.py
```diff
@@ -551,9 +557,40 @@ def get_agent(args, parallel_generation_tasks: int | None = None):
 _INFERENCE_INTERFACE = None
 
 
+def _eager_train_comm_warmup():
```
lauradang, line 564: Should we add some timing logs for this warmup just to ensure it isn't causing overhead (even if someone were to put a bug here in the future, we could catch it)?

megatron/rl/sequence_packing_utils.py
```diff
@@ -606,9 +606,11 @@ def compute_packed_inference_logprobs_stats(
         # poisons the mismatch stats with spurious |p_old - 1| terms.
         mask = mask & filled_mask.to(mask.device)
 
-    # Ensure shapes match
     if mask.shape != old_logprobs.shape:
-        return
+        raise RuntimeError(
```
lauradang, line 610: What's the reason for raising an error here vs. just passing? I agree just passing wasn't ideal without at least a warning log though.

lauradang (changes requested):

megatron/rl/rl_utils.py
tdene: Take a brief look at #6043.

In short, `do_train` determines whether we enter the `train()` function at all, and `skip_train` determines whether we skip the training logic. In RL, the `train()` function has a lot of logic that is not actually doing training, so we want to be able to enter `train()` but then skip the training parts of it.

The three valid modes are:
 * Normal training: `do_train=True`, `skip_train=False`.
 * Full RL runs that skip the training: `do_train=True`, `skip_train=True`.
 * Eval: `do_train=False`, `skip_train=True`.

So `do_train=True`, `skip_train=True` is allowed.

I did add a regression test to pin this down, but this is intentional.

megatron/rl/rl_utils.py
tdene: Absolutely, thanks!

megatron/rl/sequence_packing_utils.py
tdene: If there's a mismatch here, the importance sampling correction formula is offset, and we train on garbage data. This is meant to catch one of those subtle failures that is impossible to diagnose if it ever happens.

megatron/rl/rl_utils.py
lauradang: pretty confusing for the reader, but i'd say it's out of scope for this PR. maybe something to improve on later

lauradang (approved):

chtruong814:
@tdene 
This seems like a real failure in the merge queue
https://github.com/NVIDIA/Megatron-LM/actions/runs/31746288088/job/94606116649

Perhaps due to the change that's in front of it?
https://github.com/NVIDIA/Megatron-LM/pull/6125

So rather than re-adding this.  Maybe wait until the one in front is merged.

tdene:
> @tdene This seems like a real failure in the merge queue https://github.com/NVIDIA/Megatron-LM/actions/runs/31746288088/job/94606116649
> 
> Perhaps due to the change that's in front of it? #6125
> 
> So rather than re-adding this. Maybe wait until the one in front is merged.

Thanks for catching this!

tdene:
@chtruong814 all the failures appear on `766267172432.dkr.ecr.us-east-2` during `cicd-container-build` and say `denied: The subject of your artifact has the maximum number of artifacts (100) referring to it.`

chtruong814:
@tdene Yeah we noticed the issue with the GB300 testing.  We're skipping that for now.  Re-running your CI

PR #6048, opened 2026-07-26
Enforce that the number of optimizer shards used in layout computation is the same used during the training iteration

### LINKED ISSUE #6390: [feat] GTP+Muon

deepakn94:
# What does this PR do ?

Everything here is scoped to `--num-distributed-optimizer-instances > 1`. With a single instance, nothing in this PR changes behaviour.

With multiple instances, the optimizer shards each bucket over the **intra-instance** data-parallel group rather than the full `dp_cp`, because that is the group DDP reduce-scatters gradients over. Two places got that wrong:

1. `DistributedOptimizer` layouts were sized by the full `dp_cp`, so the layout recorded more shards than the collective used.
2. `LayerWiseDistributedOptimizer` shards over the full `dp_cp` unconditionally, and has no notion of instances at all.

The first is not a live failure — it was caught during the GTP merge and already fixed; this PR re-implements that fix differently. The second is a live bug.

## Background on (1)

`DistributedOptimizer._build_model_gbuf_range` splits each bucket into shards and gives rank `r` the r-th one, while the reduce-scatter and all-gather run over the buffer's `data_parallel_group`. With multiple optimizer instances, that group is the intra-instance group. When the layout is sized by the full `dp_cp` instead, the trailing shards of every bucket are owned by no rank: those parameters are never updated, and they drop out of grad-norm, num-zeros and params-norm, which sum over owned shards only.

`gpt3_mcore_te_tp2_pp1_resume_torch_dist_multi_dist_optimizer_instances` caught this during the GTP merge, and 19f8482766 fixed it by having core stop reading the layout's count and re-derive it from the group. `main` is green.

## What changes here

**Reverts 19f8482766, and takes the other side of the trade.** The layout's `num_optimizer_shards` stays authoritative; callers must size layouts by the group the optimizer shards over, and an assertion in `_build_model_gbuf_range` enforces it. A caller that gets it wrong fails loudly at startup, instead of silently dropping parameters. `test_optimizer_shards_cover_every_param` from that commit is kept, since it asserts the invariant without assuming where the count comes from.

**Brings the two training-side callers in line.** `wrap_model_chunks_with_ddp` (training.py) and `_ddp_wrap` (models/dist_utils.py) pass `intra_dp_cp` / `intra_expt_dp`. `intra_dp_cp` is `dp_cp` with a single optimizer instance, so no branch on the instance count is needed. DDP's own fallback layout at `distributed_data_parallel.py:143` already used these groups. The two layout helpers in `test_param_and_grad_buffer.py` pass `with_context_parallel=True, partial_data_parallel=True`, to request the same intra DP x CP group.

**Rejects multiple instances for LayerWiseDistributedOptimizer.** LayerWise assigns whole params to ranks of the full `dp_cp` group and all-gathers over that same group, so with more than one instance, a rank is asked to update params whose gradients DDP reduce-scattered to a different rank. Nothing rejected the combination, so muon with multiple instances trained on partial gradients silently. This adds an assertion. Supporting the combination properly would mean moving LayerWise onto the intra groups throughout, which is left as follow-up. With the assertion in place, `intra_dp_cp` is `dp_cp` whenever LayerWise runs, so the layout sizing above needs no special case for it.

LayerWise also now reads `dp_cp` and `expt_dp` once in `__init__` as `self.dp_cp` / `self.expt_dp`, instead of reaching into `pg_collection` at each of eleven call sites across sharding, all-gather and broadcast.

## Testing

- [x] `test_optimizer_shards_cover_every_param` (test added in `19f8482766`).
- [x] Muon unit tests that construct `LayerWiseDistributedOptimizer` directly.
- [ ] Functional tests that construct multiple distributed optimizer instances.

## Contribution process

### DIFF
--- a/megatron/core/distributed/param_and_grad_buffer.py
+++ b/megatron/core/distributed/param_and_grad_buffer.py
@@ -1071,6 +1071,7 @@ class _ParamAndGradBuffer:
             param_layout = _compute_default_per_buffer_param_layout(self.params, bucket_size)
         self.param_index_map = param_layout.param_index_map
         self.bucket_indices = param_layout.bucket_indices
+        self.num_optimizer_shards = param_layout.num_optimizer_shards
         per_bucket_numel_unpadded = param_layout.per_bucket_numel_unpadded
 
         # Check if this buffer contains NVFP4 params.
--- a/megatron/core/optimizer/distrib_optimizer.py
+++ b/megatron/core/optimizer/distrib_optimizer.py
@@ -211,6 +211,18 @@ class DistributedOptimizer(MixedPrecisionOptimizer):
         data_parallel_rank = param_and_grad_buffer.data_parallel_group.rank()
         data_parallel_world_size = param_and_grad_buffer.data_parallel_group.size()
 
+        # The layout records how many shards it was built for. That count has to match the group
+        # the reduce-scatter and all-gather run over, which is the intra-instance group when
+        # there are several optimizer instances. If the layout was sized by a larger group, the
+        # trailing shards of every bucket belong to no rank: those params are never updated and
+        # drop out of grad-norm, num-zeros and params-norm, which sum over owned shards only.
+        num_optimizer_shards = param_and_grad_buffer.num_optimizer_shards
+        assert num_optimizer_shards is None or num_optimizer_shards == data_parallel_world_size, (
+            f"Parameter layout was built for {num_optimizer_shards} optimizer shards but the "
+            f"buffer's data-parallel group has {data_parallel_world_size} ranks. Size the layout "
+            f"by the group the optimizer shards over."
+        )
+
         bucket = param_and_grad_buffer.buckets[bucket_index]
         gbuf_size = bucket.grad_data.numel()
         assert (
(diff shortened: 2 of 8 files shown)


megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1071,6 +1071,7 @@ def __init__(
             param_layout = _compute_default_per_buffer_param_layout(self.params, bucket_size)
         self.param_index_map = param_layout.param_index_map
         self.bucket_indices = param_layout.bucket_indices
+        self.num_optimizer_shards = param_layout.num_optimizer_shards
```
deepakn94, line 1074: Is this what we do with the `LayerWiseDistributedOptimizer` as well?

deepakn94:
/ok to test [75ee560](https://github.com/NVIDIA/Megatron-LM/pull/6048/commits/75ee560efe3a3946efa09d116d535ff84f15005f)

deepakn94:
/ok to test [dc4ad85](https://github.com/NVIDIA/Megatron-LM/pull/6048/commits/dc4ad85eb829936add1747c82b6fb3f1de7a3943)

yashaswikarnati (approved):

fanshiqing (approved):

kvareddy (approved):

jiemingz (approved):

yaoyu-33 (approved):

zhongbozhu (approved):
LGTM

Issue #5786, opened 2026-07-13
[BUG] `make_fsdp_dtensor` views a rank-local buffer with the DTensor global trailing shape (rowwise-TP deadlock)


yuhezhang-ai:
**Describe the bug**

Package: `megatron-fsdp` (PyPI). Affected version: 0.5.0 (latest published release at time of writing). Upstream source: NVIDIA/Megatron-LM, `megatron/core/distributed/fsdp/src/megatron_fsdp/param_and_grad_buffer.py`. Symbol: `make_fsdp_dtensor` (module-level function). Location: `param_and_grad_buffer.py`, in the flat-buffer view block (near line 4726 in the installed 0.5.0 wheel).

There is a deterministic hang/deadlock at Megatron-FSDP initialization when combining Megatron-FSDP with Torch-native tensor parallelism (`DTensor`) that shards a parameter rowwise (i.e. a non-leading dimension differs per TP rank).

`make_fsdp_dtensor` builds the view shape for a rank-local flat buffer from the parameter's *global* trailing dimensions:

```python
if len(orig_param.shape) > 1:
    local_shape = (-1, *orig_param.shape[1:])
else:
    local_shape = (-1,)
```

For a `DTensor`, `orig_param.shape` is the global shape. Under rowwise TP the trailing dimensions differ per rank, so `orig_param.shape[1:]` is not the shape of the local shard held by this rank. The rank-local flat buffer only contains the rank-local shard, so the reshape/view against the global trailing shape is inconsistent across ranks: some DP/CP shards can perform the view while others cannot. The ranks that fail (or take a different branch) desynchronize from the uneven-DTensor metadata collectives that the remaining ranks enter, and the job deadlocks.

**Steps/Code to reproduce bug**

* Build a small model whose parameter is sharded rowwise across a TP group (e.g. a `Linear` sharded so a non-leading dimension differs per rank), wrapped as a Torch-native `DTensor`.
* Wrap it with Megatron-FSDP with `dp_size > 1` so the shard group is nontrivial.
* On any TP rank where the local trailing shape differs from the global trailing shape, `make_fsdp_dtensor` produces a view inconsistent with peer ranks, and the subsequent uneven-DTensor metadata collective deadlocks. With the `to_local().shape` fix, all ranks view their own shard and initialization completes.

**Expected behavior**

[...]

PR #6052, opened 2026-07-26
Add tensor-parallel composability

wujingyue:
## What

Adds data-parallel + tensor-parallel (DP+TP) composability to the experimental
Megatron-FSDP path (`megatron/core/distributed/fsdp/src/megatron_fsdp/experimental`),
implementing phases 1–3 of the design in
`docs/discussions/mfsdp-v2-tp-composability-plan.md` (included here).

Optimizer state and checkpoints now see full-mesh DTensors carrying both DP and
TP placements, while compute keeps ordinary TP-local parameters and FSDP never
gathers across the TP axis.

## Design

- **DBuffer** owns only the data-parallel submesh (storage, layouts,
  collectives) and constructs DP-only DTensors.
- **FsdpParameterGroup** owns the full user-provided mesh. Its `get_dtensor`
  lifts a DBuffer's DP-only DTensor onto the full mesh, splicing the
  tensor-parallel placement onto the single non-DP axis and scaling the sharded
  dimension to the global shape. This is the only place that maps DP mesh axes
  to full-mesh axes, recovering the correspondence from the two meshes' dim names.
- **TP placement is inferred** from each parameter's own MCore / Transformer
  Engine attributes (`tensor_model_parallel` / `partition_dim`), so callers pass
  no separate TP layout. The emitted DTensor placement tuple follows the
  user-provided mesh-axis order, so `[dp, tp]` and `[tp, dp]` place the same
  parameter symmetrically.
- `mesh_utils.build_dp_mesh` derives the DP submesh from the full mesh.

## Tests

`tests/unit_tests/distributed/mfsdp_v2/test_tensor_parallel.py` (4 ranks, DP=2 × TP=2):
- DP-only DBuffers, full-mesh placements in both axis orders, global
  shape/stride, and optimizer-state metadata.
- A loss-curve parity test: a real `ColumnParallelLinear` → GELU →
  `RowParallelLinear` MLP wrapped in `fully_shard` tracks an unsharded
  single-replica reference. Process groups are passed explicitly via
  `ProcessGroupCollection` (no global `parallel_state` mutation).

Requires a 4-GPU job.

## Out of scope (follow-ups)

TE/MCore adapter breadth, DCP resharding across a changed DP degree, HSDP+TP
composition, and strided (`partition_stride > 1`) TP layouts, which are rejected
for now.

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
@@ -468,7 +468,7 @@ class DBuffer:
         should build the full-mesh DTensor at the ``FsdpParameterGroup`` layer,
         which owns the full user-provided mesh.
         """
-        torch_placements = tuple(to_dtensor_placement(placement) for placement in self.placements)
+        dtensor_placements = tuple(to_dtensor_placement(placement) for placement in self.placements)
 
         local_tensor = self.get_local_tensor(index)
         tensor_shape = self.layout.tensor_shapes[index]
@@ -477,7 +477,7 @@ class DBuffer:
         return DTensor.from_local(
             local_tensor=local_tensor,
             device_mesh=self.mesh,
-            placements=torch_placements,
+            placements=dtensor_placements,
             run_check=False,
             shape=tensor_shape,
             stride=local_tensor.stride(),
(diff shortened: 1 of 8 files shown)


wujingyue:
Cross post: https://github.com/NVIDIA/Megatron-LM/issues/5786#issuecomment-5275123588

Consider supporting parallelize_module, PyTorch's official wrapper-based tensor parallel solution.

PR #6058, opened 2026-07-27
[dev] Load-balanced context-parallel DSA indexer

jingqiny-99:
## Summary

Adds load-balanced context-parallel indexing for DSv4 packed THD inputs. Each rank scores complementary head/tail query chunks with the existing compact fused indexer, then restores selected indices and sparse-loss predictions to contiguous query ownership. Eligible eager packs use the balanced path; ineligible eager packs retain the ordinary CP path.

Selection and scoring preserve the configured precision, including MXFP8. Delayed-scaling selection uses a stateless nonquantized projection to avoid extra updates to the shared projection's amax history. Existing fused row-limit guards remain in force.

## Configuration and CUDA graphs

- `--dsa-cp-balance-indexer` is the feature opt-in and remains disabled by default.
- Dynamic-pack graph routing is enabled automatically when balanced indexing is combined with `--cuda-graph-impl transformer_engine`, an attention-capturing graph scope, and `--sequence-packing-scheduler dp_balanced`. The mode is derived from these settings.
- This graph path requires fixed CP and a fixed per-rank capacity from `--max-seqlen-per-dp-cp-rank`. It refreshes fixed-address route metadata for each model chunk and microbatch slot so captured graphs can replay different eligible pack layouts.
- PP/VPP additionally requires `--cuda-graph-dynamic-microbatches` to keep graph input slots distinct while forwards remain live.

See the [design and validation guide](https://github.com/jingqiny-99/Megatron-LM/blob/609d7c041ebcc381d07939ee1b1e92652470157a/docs/developer/dsv4_cp_balanced_indexer.md) for routing, precision, and coverage details.

## Follow-up

[#7206: Support balanced DSA partial graphs with EP 1F1B overlap and PP/VPP](https://github.com/NVIDIA/Megatron-LM/pull/7206) adds EP 1F1B overlap to balanced-indexer partial CUDA graphs for GPTModel and DSv4 HybridModel, including PP1, ordinary PP, and interleaved VPP. Full-iteration CUDA graphs are outside that follow-up's scope.

### DIFF
--- a/megatron/core/transformer/experimental_attention_variant/cp_balanced_indexer.py
+++ b/megatron/core/transformer/experimental_attention_variant/cp_balanced_indexer.py
@@ -1,0 +1,570 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+"""Load-balanced context-parallel DSA indexer.
+
+Under context parallelism the DSA indexer uses a contiguous sequence split, so the causal
+per-query cost ``visible_k(p) = min((p + 1) // ratio, comp_len)`` grows with position and later
+CP ranks become stragglers (the layout-level max/min work ratio is ``2 * cp_size - 1``). This
+module removes that imbalance without a new kernel: it splits the global sequence into
+``2 * cp_size`` near-equal chunks and gives CP rank ``r`` chunk ``r`` (a low-position "head",
+light) and chunk ``2 * cp_size - 1 - r`` (a high-position "tail", heavy), so
+``work(head) + work(tail)`` is ~constant across ranks. Each chunk is scored by a per-chunk call to
+``compute_cp_indexer_topk`` at the chunk's true global offset, so multi-sequence packs, the
+``use_fused`` flag, and non-even per-rank lengths are all handled exactly as in the reference path.
+
+``balanced_compute_cp_indexer_topk`` is a drop-in replacement for
+``csa_cp_utils.compute_cp_indexer_topk``: it returns the top-k in the same contiguous
+``[l_local, topk]`` layout, so the downstream index-building and sparse attention are unchanged.
+"""
+import torch
+import torch.distributed as dist
+
+from megatron.core.transformer.experimental_attention_variant.dsa import rotate_activation
+from megatron.core.utils import nvtx_range_pop, nvtx_range_push
+
+
+def _all_gather_rows(x, l_local, cp_group, cp_size):
+    """AllGather x[l_local, D] across the CP group -> [cp_size*l_local, D] (global-row order)."""
+    D = x.shape[-1]
+    x2 = x.reshape(l_local, D).contiguous()
+    g = torch.empty((cp_size * l_local, D), dtype=x2.dtype, device=x2.device)
+    dist.all_gather_into_tensor(g, x2, group=cp_group)
+    return g
+
+
+# Static all-to-all metadata and staging buffers, cached per (group, cp_size, l_local[, dtype,
+# width]). The chunk map is a fixed permutation: chunk ``c``'s rows are OWNED by rank ``c // 2``
+# (contiguous layout) and COMPUTED by rank ``proc(c) = c if c < cp_size else 2*cp_size - 1 - c``.
+# Dispatch moves each owned half-chunk to its computing rank; combine moves each computed top-k
+# chunk back to its owner. Payload is ~l_local rows per rank per direction (vs. cp_size*l_local
+# for AllGather, or an S-row zero-padded reduce_scatter for the combine). All split sizes are
+# static for a fixed pack size, so ``all_to_all_single`` is CUDA-graph capturable, and the cached
+# staging buffers keep the allocator pool static (no expandable-segments cuMem churn).
+_A2A_META: dict = {}
+_A2A_BUF: dict = {}
+
+
+def _group_key(cp_group):
+    """Stable cache key for a process group.
+
+    ``id()`` alone can be recycled if a group is destroyed and a new one is allocated at
+    the same address; c10d's ``group_name`` is unique per created group, so prefer it.
+    """
+    return getattr(cp_group, "group_name", None) or id(cp_group)
+
+
+def _a2a_meta(cp_group, cp_size, l_local):
+    key = (_group_key(cp_group), cp_size, l_local)
+    meta = _A2A_META.get(key)
+    if meta is not None:
+        return meta
+    N, nch = cp_size, 2 * cp_size
+    S = N * l_local
+    r = cp_group.rank()
+    bounds = [(k * S) // nch for k in range(nch + 1)]
+    size = [bounds[k + 1] - bounds[k] for k in range(nch)]
+
+    def proc(c):  # computing rank of chunk c (head for c < N, tail otherwise)
+        return c if c < N else nch - 1 - c
+
+    def own(c):  # owning rank of chunk c in the contiguous layout
+        return c // 2
+
+    head_c, tail_c = r, nch - 1 - r
+
+    # Convention on both sides: rows for one peer are ordered by ascending chunk id, and
+    # all_to_all_single orders the buffers by peer rank. head_c < tail_c and
+    # own(head_c) <= own(tail_c) always hold, so [head | tail] is already peer-ordered on the
+    # receive side of the dispatch and on the send side of the combine. The owner's local pair
+    # (2r, 2r+1) needs a swap iff proc(2r) > proc(2r+1) (true for 2r >= N).
+    d_in = [0] * N
+    d_in[proc(2 * r)] += size[2 * r]
+    d_in[proc(2 * r + 1)] += size[2 * r + 1]
+    d_out = [0] * N
+    d_out[own(head_c)] += size[head_c]
+    d_out[own(tail_c)] += size[tail_c]
+    c_in = [0] * N
+    c_in[own(head_c)] += size[head_c]
+    c_in[own(tail_c)] += size[tail_c]
+    c_out = [0] * N
+    c_out[proc(2 * r)] += size[2 * r]
+    c_out[proc(2 * r + 1)] += size[2 * r + 1]
+    swap_pair = proc(2 * r) > proc(2 * r + 1)
+
+    meta = {
+        "d_in": d_in,
+        "d_out": d_out,
+        "c_in": c_in,
+        "c_out": c_out,
+        "swap_pair": swap_pair,
+        "s0": size[2 * r],
+        "s1": size[2 * r + 1],
+        "sh": size[head_c],
+        "st": size[tail_c],
+    }
+    _A2A_META[key] = meta
+    return meta
+
+
+def _a2a_buf(tag, rows, width, dtype, dev, cp_group):
+    key = (tag, _group_key(cp_group), rows, width, dtype)
+    buf = _A2A_BUF.get(key)
+    if buf is None:
+        buf = torch.empty((rows, width), dtype=dtype, device=dev)
+        _A2A_BUF[key] = buf
+    return buf
+
+
+def _alltoall_dispatch(x, meta, cp_group, tag):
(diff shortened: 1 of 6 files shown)


jingqiny-99:
# Benchmark + kernel breakdown: DSv4 proxy, 256K @ CP16, partial CUDA graph (GB200)

**Setup**: DSv4 proxy — 4 layers + 1 MTP layer, `--csa-compress-ratios [0,128,4,0,0]` (one CSA/indexer layer, one HCA layer, window layers elsewhere), MLA (hidden 4096, 64 heads), MoE 128 experts topk-6 hybridEP EP4 + shared expert, hyper-connections (4 streams), mxfp8. CP16 (`--cp-comm-type a2a`, contiguous CP partition), varlen THD, one 262144-token sequence per 256K pack (exact fill). 16× GB200 (4 nodes). Partial CUDA graph: `--cuda-graph-impl transformer_engine --cuda-graph-modules attn`. 30 train iters; steady state = iters 12–29 excluding the periodic manual-gc iters (spikes appear at the same iters in both arms). nsys on all 16 ranks; per-module numbers read from the NVTX rows of the timelines.

## End-to-end (steady state, partial CG)

|  | baseline | balanced | Δ |
|---|---|---|---|
| TFLOP/s/GPU | 773–781 | **827–840** | **+7.4%** |
| iter time (ms) | 808–819 | **753–763** | **−6.9%** |
| peak alloc (GB) | 129.7 | 129.8 | +0.1 |

Eager reference (same config, no CUDA graph, two independent rounds each): baseline 830 / 823 ms (760 / 767 TF), balanced 795 / 789 ms (794 / 797 TF) → −4.2% wall. The gain is larger under CG because graph replay removes the launch/CPU exposure that partially hides the baseline's straggler wait.

## Per-layer forward breakdown (ms per call, from nsys timelines)

|  | baseline | balanced |
|---|---|---|
| indexer fwd kernel | 28.53 | **8.93** |
| indexer (`indexer_total` range) | 35.09 | **18.33** |
| CSA layer attn total (`attn_csa_total`) | 39.08 | **22.32** |
| balance overhead (dispatch + combine) | 0 | 2.84 |

- The indexer forward dominates the CSA layer (35 of 39 ms baseline); everything outside the indexer (HCA attention, MoE) is identical between arms.
- Balanced execution cuts the indexer fwd kernel time 28.5 → 8.9 ms: uniform S/2N-row call shapes with causally-tight, quantized K widths (K-prefix slice), plus removal of the straggler wait.
- The balance tax is 2.84 ms per iter (A2A dispatch + combine), against a >16 ms saving on the indexer range.

## Loss

iter-12 lm loss (same seed/data/init per pair; all runs start at 12.586 ±0.001 at iters 1–2; indexer loss coeff 0, MoE aux coeff 0):

| run | baseline | balanced | Δ (base − bal) |
|---|---|---|---|
| eager, round 1 | 11.68539 | 11.67069 | +0.0147 |
| eager, round 2 | 11.68983 | 11.67009 | +0.0197 |
| partial CG | 11.66948 | 11.66970 | **−0.0002 (aligned)** |

30-iter horizon under CG: the two curves track each other within 1.6e-3 throughout and converge — iter 20: 9.821200 vs 9.821811 (Δ6e-4); iter 30: **8.846773 vs 8.846653 (Δ1.2e-4)**. Balanced computation is numerically equivalent to the baseline over the full horizon.

Note on the eager gap: the eager baseline sits 0.015–0.020 *above* the balanced band and is also noisier run-to-run (base-to-base Δ4.4e-3 vs bal-to-bal Δ6e-4), while under CUDA-graph static buffers the baseline rejoins the band exactly. Together with alignment controls at CP32/CP64 (baseline call rows 8192/4096: base↔bal Δ ≤ 7.5e-4) this points at the ≥16384-row fused indexer top-k call path being sensitive to allocator-reused memory — i.e. the discrepancy is a baseline-side artifact, not introduced by this PR. Will be filed as a separate issue.

jingqiny-99:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/6058/commits/fc5c68bf1fd17ce6c988b2e50ef8ff16b182ae5d

megatron/core/transformer/experimental_attention_variant/csa.py
```diff
@@ -1792,6 +1792,16 @@ def __init__(
             )
         else:
             self.indexer = None
+        if (
+            config.dsa_cp_balance_indexer
+            and self.compress_ratio == 4
+            and not config.csa_dense_mode
+            and self.indexer is None
```
kunlunl, line 1799: When can compress_ratio == 4 but self.indexer is None? Is it when the module spec doesn’t provide an indexer?

megatron/core/transformer/experimental_attention_variant/csa.py
```diff
@@ -2627,6 +2637,32 @@ def _forward_thd_cp(
                     raise RuntimeError(
                         f"DSv4 THD CP indexer expects bsz=1, got {indexer_x.shape[1]}."
                     )
+                nvtx_range_push("indexer_total")
+                # Switch 1: enable per-pack balanced routing via the config flag.
+                # CP<=1 has nothing to balance.
+                use_balance = self.config.dsa_cp_balance_indexer and cp_size > 1
+                graph_dynamic_packs = bool(
+                    getattr(self.config, "dsa_cp_balance_indexer_graph_dynamic_packs", False)
```
kunlunl, line 2646: Do we need a separate flag for dynamic packs, or can this be inferred from balanced indexer + TE attention CUDA graphs + the dp_balanced scheduler?

megatron/core/transformer/experimental_attention_variant/cp_balanced_indexer.py
```diff
@@ -0,0 +1,2200 @@
...
+            w_rows.reshape(sz, n_heads),
+            kv_rows,
+            cu_seqlens,
+            cu_seqlens_compressed,
+            0,
+            ratio,
+            topk,
+            softmax_scale,
+            max_seqlen_q=mq,
+            use_fused=True,
```
kunlunl, line 1666: balanced_compute_cp_indexer_topk() accepts use_fused, but the balanced path always passes use_fused=True. Since this path is fused-only, should we remove the argument or fail explicitly when it is False?

megatron/core/transformer/experimental_attention_variant/cp_balanced_indexer.py
```diff
@@ -0,0 +1,2200 @@
...
+            "balanced CP indexer: stale or foreign zigzag plan (K-slice ends "
+            f"{plan.get('k_end_head')}/{plan.get('k_end_tail')} vs {k_rows_total} "
+            f"K rows, plan ratio {plan.get('_ratio')} vs {ratio})."
+        )
+    # NOTE: k_h / k_t must stay prefix VIEWS of the full gathered buffer. The
+    # packed layouts may declare per-sequence K ranges past the slice end
+    # (mkv_* is a capacity, k_end_* the true causal need); reads past the
+    # slice land on valid full-buffer memory only while these are views — a
+    # .contiguous() here would turn them into real out-of-bounds reads.
+    k_h = (
```
kunlunl: Is it safe to pass a sliced K tensor while cu_seqlens_kv may still reference rows beyond k.shape[0]? This seems to rely on data outside the tensor’s logical bounds remaining accessible in the backing storage. Could we keep the full K tensor and only tighten max_seqlen_kv instead?

megatron/core/transformer/experimental_attention_variant/csa_utils/cp_utils.py
```diff
@@ -330,9 +382,12 @@ def compute_cp_indexer_topk(
...
-        cu_seqlens_q, cu_seqlens_compressed, global_start, l_local
-    )
+    if prebuilt_layout is not None:
+        cu_q_topk, cu_k_topk, q_causal_offsets = prebuilt_layout
+    else:
+        cu_q_topk, cu_k_topk, q_causal_offsets = _build_cp_indexer_layout(
+            cu_seqlens_q, cu_seqlens_compressed, global_start, l_local
+        )
 
     if not use_fused:
```
kunlunl, line 399: Could we fail fast when synthetic_layout=True and use_fused=False? The doc says this combination silently mis-masks, but it currently proceeds without an error.

jingqiny-99: Sure, I'll fix it in next commit.

megatron/core/transformer/experimental_attention_variant/csa.py
jingqiny-99: Yes, for example, if the user uses some customized --spec flags. I added this to pass the strict review.

megatron/core/transformer/experimental_attention_variant/csa.py
jingqiny-99: Agree. I think it is better to infer from config than to make a separate flag. I'll fix it in the next commit.

megatron/core/transformer/experimental_attention_variant/cp_balanced_indexer.py
jingqiny-99: That is a mistake. I'll remove the useless argument.

megatron/core/transformer/experimental_attention_variant/cp_balanced_indexer.py
jingqiny-99: Agree, I'll fix it soon.

megatron/core/models/hybrid/hybrid_model.py
```diff
@@ -530,6 +535,14 @@ def forward(
                 decoder_input = tensor_parallel.scatter_to_sequence_parallel_region(
                     decoder_input, group=self.pg_collection.tp
                 )
+            if padding_mask is not None and self.config.sequence_parallel:
```
kunlunl, line 565: Is this unrelated to the balanced indexer change, but an existing bug discovered and fixed while testing this PR?

pretrain_hybrid.py
```diff
@@ -95,6 +99,8 @@ def get_batch(data_iterator, vp_stage=None):
 
     args = get_args()
     config = core_transformer_config_from_args(args)
+    balance_indexer = getattr(config, "dsa_cp_balance_indexer", False)
```
kunlunl: This becomes too DSv4-specific (I think pretrain_hybrid.py is supposed to be general) ...

megatron/core/models/hybrid/hybrid_model.py
jingqiny-99: Yes, and maybe it can stand as an independent PR to be merged further.

tests/unit_tests/transformer/experimental_attention_variant/test_dsv4_hybrid_attention_cp.py
kunlunl, line 871: Could we add a balanced-indexer variant of this CP-vs-full-reference test? It would be useful to cover an eligible ratio-4 packed input with dsa_cp_balance_indexer=True on CP2/CP4, and compare the attention output, hidden gradient, and parameter gradients against the CP1 reference. The test input should ensure that every sequence length is divisible by 2 * CP; otherwise, it would fall back to the original path and not exercise the balanced indexer.

kunlunl (approved):

megatron/core/models/hybrid/hybrid_model.py
kunlunl: It's fine. Just out of curious.

tests/unit_tests/transformer/experimental_attention_variant/test_dsv4_hybrid_attention_cp.py
jingqiny-99: OK! I'll add the test

pretrain_hybrid.py
jingqiny-99: Agree.. I'll try a better way to do this.

hxbai (approved):
LGTM

PR #6064, opened 2026-07-27
perf(inference): env-gated decode-path optimizations for MoE inference (Qwen3-30B-A3B, +36%)

shanmugamr1992:
## Summary

Sixteen independently env-gated optimizations to the dynamic-batching decode path, plus a FlashAttention-generation selector, developed against Qwen3-30B-A3B on one 4×GB200 node (TP1/PP1/EP4/ETP1, BS256/OSL1024 gsm8k).

Cumulatively they move throughput from ~22.4k to **30,482 tok/s** — from **65.9% to 89.7%** of a vLLM DP4+EP baseline (33,994.5 tok/s) on this workload.

That headline is the steady-state iterations of one arm, TPOT 8.40 ms/tok, and the arm it is measured against reproduces to 0.19% across two runs on the same node (30,341 / 30,399 with the last change off).

**Everything defaults to off.** The only exception is `MCORE_FUSE_FC1_ACT`, which env-gates behaviour that was previously hard-coded on. With no environment variables set, this PR is inert.

### MoE kernel path

| Gate | What it does | Δ |
|---|---|---:|
| `MCORE_MOE_GEMM_TUNE` | Per-GEMM tiles for FC1/FC2, whose differing N cannot share one config | +4.3% |
| `MCORE_ROUTER_FUSED_TOPK` | Fused softmax + top-k router, 4 kernels → 1 | +3.9% |
| `MCORE_MOE_FUSED_COUNT` | Fold token count and zero-fill into the indirection-table build | +3.0% |
| `MCORE_NVLS_RS_BF16` | Reduce the MoE combine in bf16 instead of fp32 | +2.5% |
| `MCORE_MOE_FUSED_ALIGN` | Merge init / prefix-sum / fill into one kernel | +1.9% |
| `MCORE_MOE_SUM_FAST` | Predicate the topk-reduction locality test into the load mask | +1.8% |
| `MCORE_MOE_FUSED_SCATTER` | Fold the scatter into the same kernel | +0.5% |

`MCORE_NVLS_RS_BF16` is the newest and is worth explaining, because the copy it removes is not the point. The `ep_rsv` symmetric-memory buffer was allocated fp32, which cost three things at once: the combine reduce-scatter moved **twice the NVLink bytes it needed to**, the MoE wrote fp32 through `_moe_sum`'s `out=`, and every layer paid a cast kernel converting back to the bf16 residual stream. Allocating it bf16 removes all three — `_moe_sum`'s `tl.store` casts on the way in for free, and `output.to(bfloat16)` becomes a no-op returning its argument. Confirmed structurally as well as by throughput: `bfloat16_copy_kernel` instances fall 98% and the per-step copy count drops by exactly 48, one per layer.

The precision cost is smaller than "reduce in bf16" implies: `multimem.ld_reduce` **accumulates in f32 either way**, and the flag only selects whether the operands it loads are `f32` or `bf16x2`. This is also where vLLM already is (`ncclDevKernel_ReduceScatter_Sum_bf16`).

### Attention

| Gate | What it does | Δ |
|---|---|---:|
| `MCORE_FLASHINFER_DECODE` | Route paged decode attention to flashinfer's trtllm-gen `fmhaSm100f` kernel | +2.6% |
| `MCORE_FUSED_QK_NORM` | Fuse the two per-head q/k RMSNorm launches in self-attention into one kernel | +2.9% |
| `MCORE_GROUPED_QK_NORM` | Read the decode query from the grouped QKV output, removing a full-query copy per layer | +1.5% |
| `MCORE_FLASHINFER_PDL` | Programmatic Dependent Launch for that kernel | +0.3% |

Megatron's decode path runs FA2's `flash_attn_with_kvcache`, whose kernels predate Blackwell. A matched per-category budget against vLLM put attention **0.409 ms/step behind at an identical launch count**, which rules out anything Megatron does around the call and points at the kernel generation — vLLM reaches `fmhaSm100f` through flashinfer, and flashinfer 0.6.14 is already installed in the same venv as flash-attn, so this adds no dependency.

Measured in isolation before integrating (`dev/moe_fused/harness_attn.py`, under graph replay so the number is device time and not Python): **24% faster than FA2 at every KV length tested** — 59.6→45.1 µs at 512, 105.8→80.7 at 1024, 199.3→151.9 at 2048.

The integration is small because the kernel takes **exactly the paged layout Megatron already has** (`kv_layout="NHD"`), and because the trtllm-gen entry point needs no host-side `plan()` call, so it captures in the decode graph. Page size is irrelevant to it (256/128/64 within 0.1 µs), so Megatron's 256-token pages stay — FA2 is in fact the stricter of the two, *requiring* pages be a multiple of 256.

It is guarded narrowly: paged KV only, no attention sinks, no sliding window, no speculative decoding, bf16/fp16 only. Each exclusion is a case not verified against FA2, where a wrong answer would be a silent accuracy bug rather than a crash. The 128 MiB scratch buffer must be zeroed outside graph capture, so the path **declines with a one-shot message** rather than record a 128 MiB memset into the graph and replay it every step.

`MCORE_FLASHINFER_PDL` is reported at +0.3% because two arms both landed above the non-PDL number (29,888.2 / 29,898.3 vs 29,802.9) — reproducible, but close enough to the 0.5% iteration spread that it should be read as "not harmful and probably small-positive", not as a measured win.

`MCORE_GROUPED_QK_NORM` removes a copy a profile had shown for four sessions without naming: a 2.8 µs strided `elementwise_kernel` per layer per step, between the QKV GEMM and the q/k norm. It is the `[sq, b, ng, np/ng * hn] → [sq, b, np, hn]` query reshape, and it cannot be a view — merging the group axis with the head axis requires the group stride to equal `(np/ng) * hn`, and it is `(np/ng + 2) * hn`, because each group's k and v head sit between consecutive groups' q heads. For this model that is 1024 required against 1280 actual, so `reshape` silently materializes the entire query. The arithmetic is checkable in ten seconds on a laptop with no GPU, which is what settled it after four cluster sessions spent suspecting the split instead — see `MCORE_QKV_SPLIT_VIEWS` below, whose exactly-neutral result was itself the answer, since both split branches feed the same reshape.

The fix costs nothing because the fused q/k norm already allocates its output, so it can absorb the repack on the way through: read the query with the two strides the grouped layout needs, write the contiguous result the rest of attention wants. Same values in the same order, so it is **bit-identical**, and it removes 48 launches and 0.122 ms of device time per step — confirmed by diffing two traces, where the copy bucket fell from 59 launches / 0.164 ms to 11 / 0.042 ms while every other bucket held still within 0.012 ms.

That kernel was then re-shaped. One CTA per 128-wide row put 9,216 single-warp CTAs on the device to move 4.5 MB, reaching ~570 GB/s; eight rows per CTA at eight warps measures 8.22 → 6.15 µs, bit-identical, and is worth a further +0.35% end-to-end (`MCORE_QK_NORM_ROWS` / `MCORE_QK_NORM_WARPS`, both tunable). A harness variant of the same kernel at the same shape reaches 4.12 µs where the shipped one does not; three explanations were tested and rejected, so it ships as measured with the discrepancy recorded rather than guessed at.

### Norm fusion

| Gate | What it does | Δ |
|---|---|---:|
| `MCORE_FUSED_ADD_NORM_QKV` | Fuse `mlp_bda`'s residual add with the *next* layer's QKV input RMSNorm, across 47 of 48 block boundaries | +2.4% |
| `MCORE_FUSED_ADD_NORM` | Fuse the `self_attn_bda` residual add with the standalone pre-MLP RMSNorm | +1.4% |

Both normalize a small row and are launch/latency bound rather than bandwidth bound, so both are gated to decode-sized token counts (`≤256` by default, tunable per gate); above that the reference path is faster and is used. They **overdeliver against their own microbenchmarks** — `MCORE_FUSED_ADD_NORM_QKV` returns +2.4% where its device-time arithmetic predicts ~0.5% — because the norms sit on the serial per-layer dependency chain, so removing one launch also removes one CUDA-graph node and one host dispatch gap, ×47.

`MCORE_FUSED_ADD_NORM_QKV` is worth a note on how it was found: the matching fusion was previously ruled unreachable because "the norm lives inside TE's `LayerNormLinear`". That was true of the class name and false of the **TP=1 code path**, where `InferenceLayerNormColumnParallelLinear.forward` runs the norm as its own kernel launch immediately before the GEMM. Reading the actual forward, rather than trusting the ruling, was the whole difference.

### Host path

| Gate | What it does | Δ |
|---|---|---:|
| `MCORE_INFER_FAST_POST_PROCESS` | Decode fast path in `post_process_requests` (7.8× on that call) | +1.0% |
| `MCORE_INFER_INCR_ATTN_STATE` | Incremental attention-state rebuild keyed on a request-layout version counter (3.4×) | +0.9% |
| `MCORE_INFER_VEC_UPDATE_REQS` | Vectorized per-request bookkeeping (1.9× on `update_requests`) | +0.9% |

Each percentage is against a same-session reference measured immediately before that change. They are **not additive** — run-to-run drift between sessions is up to ±1.6%, which is why every A/B was run in-session, and gates that attack the same resource compose sub-additively. One clean example: `CUDA_DEVICE_MAX_CONNECTIONS=8` was worth +0.4–0.9% before `MCORE_NVLS_RS_BF16` and is worth **exactly nothing after it** (29,346.3 vs 29,354.2), because both were relieving contention on the same fp32 collective. Read the tables as a ranking, not a sum; the 88.0% figure is the only cumulative number here that was actually measured.

### FlashAttention generation

`MCORE_FLASH_ATTN_VERSION` pins the generation, because `flash_attention_version` has no CLI flag. Pinning **2** is worth **+3.5%** over the FA4 default on this workload, and that finding survived a rebase: FA4 hardcoded `num_splits=1` when first tested, upstream has since made it auto (`num_splits=0`), and re-testing shows FA4 **still 3.07% slower** (27,716 vs 28,594). So split-KV was not the reason it lost.

The reason is routing, and it may deserve a fix in `main` independent of this PR. mcore's auto-preference is FA4 → FA3 → FA2, but **FA4 has no dedicated decode kernel on this path**: `flash_decode_and_prefill` routes FA4 decode into `flash_attn4_varlen_func`, the varlen *prefill* interface called with `max_seqlen_q=1`, while FA2 goes to `flash_attn_with_kvcache`, the purpose-built flash-decoding kernel. So the default order silently costs decode ~3.5% wherever FA4 is installed. Config still wins over the env var, and the default resolution order is unchanged by this PR. With `MCORE_FLASHINFER_DECODE=1` this selector applies only to prefill.

## Where the remaining 12% is, and where it is not

The most useful thing this section can say is that **its previous version was wrong**, and how. It reported attention at 1.51 ms against a ~1.89 ms bf16 bandwidth floor and concluded "at the bound — not a tuning lever". A 24%-faster kernel for the same math on the same bytes says otherwise. The error was fitting a roofline to one kernel and treating the fit as a property of the *problem*; a bandwidth floor bounds what a perfect kernel can do, and says nothing about whether the current one is that kernel.

The current decomposition comes from a matched per-category budget between the two engines' traces, windowed to steady-state decode, comparing **device-side durations and launch counts only** — never wall or idle, which `--cuda-graph-trace=node` inflates unequally between an engine that replays a whole-step graph and one that launches eagerly:

| | vLLM | mcore | delta |
|---|---|---|---|
| GPU busy, interval union | 6.99 ms | 7.48 ms | +0.49 |
| exposed gap (step − busy) | ~0.54 ms | ~1.43 ms | **+0.89** |

Two thirds of the gap was **packing, not work**. Only 0.18 ms of it was extra kernel time: the expert GEMM is at parity (+0.06 ms) and mcore's collectives are half a millisecond *faster* than vLLM's. Attention was the one real work gap, and `MCORE_FLASHINFER_DECODE` addresses it.

What remains, in order:

- **Nothing overlaps.** mcore's sum-of-durations and its interval union differ by 0.6%; vLLM hides 4.8% of its work behind other work. mcore's ~1 ms of collectives is fully serialized against compute. Matching vLLM's overlap ratio is worth ~0.3 ms and is a scheduling change, not a kernel.
- **The gaps are launch-shaped.** 855 of ~900 per-step gaps are sub-microsecond and sum to ~0.45 ms, against 1027 launches to vLLM's 810. A calibration ballast that adds no-op nodes to the decode graph prices one serialized node at **1.34 µs**, so every removed launch pays twice — its device time and its gap.
- **96 kernels/step vLLM does not have at all**: 82 split-K reduce launches (0.165 ms) from cuBLASLt's choice on the skinny decode GEMMs, and ~48 unattributed QKV-sized copies between the QKV GEMM and the norm.

Two things are explicitly **not** levers, both established by measurement rather than assumption:

- **Host-side work.** One gate's fast-path engagement was raised from 0.1% to 81% of calls, removing ~204 µs/step of real host work, and **throughput did not move** — that work was already fully hidden behind GPU execution by async scheduling. Earlier profile-derived claims of ~18% recoverable host idle were substantially a profiling artifact; measuring `cudaGraphLaunch` with `perf_counter` rather than under nsys confirmed the cost nsys attributes to it is largely tracing overhead.
- **EP arrival skew.** Median cross-rank spread of step-entry timestamps is 24 µs, measured directly with `perf_counter_ns` across the four same-node ranks.

## Correctness

Correctness differs by gate class, so it is stated separately rather than claimed uniformly.

**MoE kernel gates are bit-exact** against the unfused path: `max_abs` and `max_rel` both 0.0 at 128/256/384/512 tokens. Bit-exactness is structural rather than incidental — `BLOCK_SIZE_K` is unchanged, so the fp32 K-reduction order is identical, and masked slots in the fast topk reduction add an exact 0.0. End-to-end, temperature-0 completions are byte-identical with each of these gates on versus off.

**`MCORE_NVLS_RS_BF16` changes numerics and is the one gate whose default-off status is load-bearing.** The cross-rank sum's operands become bf16 while its accumulation stays f32 in hardware; greedy decode is coherent on all three probes and iteration spread is 0.4%. Three coherence probes are not an accuracy evaluation, so enabling it by default should follow lm-eval or equivalent.

**`MCORE_GROUPED_QK_NORM` is bit-identical** to the reshape-then-normalize path, for both query and key, and end-to-end greedy completions are identical character for character. Bit-exactness is the right bar for it rather than a tolerance: the change is an addressing change, and a head-permutation bug would produce individually plausible values in every head and pass a tolerance check.

**`MCORE_FLASHINFER_DECODE` matches FA2 to bf16 tolerance** (max relative difference ~3e-3 across KV lengths 512/2048, checked in the harness before any e2e run), and greedy decode is coherent end-to-end. Both compute exact attention; the difference is accumulation order.

**Host gates each ship a `*_VERIFY=1` mode** that runs both paths and compares every derived tensor — length buffers, cumsums, block table, metadata tensors, the whole GPU bookkeeping buffer, and the address/shape/dtype binding of every view — over hundreds of consecutive decode steps. Zero mismatches, and temperature-0 completions are byte-identical. `MCORE_INFER_FAST_POST_PROCESS` is additionally safe by construction rather than by testing: the fast path **declines any step on which a request finishes**, so the request-termination state machine is unreachable from it.

**The norm fusions are not bit-exact — they are accurate to one bf16 ulp**, because TE's internal rsqrt and reduction order differ from the fused kernel's. The residual add *is* bit-exact (fp32 accumulate then round reproduces torch's bf16 add); the norm is bf16-ulp (`max_rel ≤ 7.9e-3`). On temperature-0 coherence, two of three prompts are byte-identical to gate-off and the third diverges at a low-confidence token choice, with both continuations fluent and factually correct. All are heavily guarded — decode only, RMSNorm only, TP=1 for the QKV variant, no bias/dropout/recompute/offload, no cross-attention, `not layernorm_zero_centered_gamma` — and fall back to the untouched reference path whenever a guard fails, so training and every non-matching shape remain byte-identical.

## Measured negatives retained as gates

Two gates are here because the negative result is worth being able to re-run after a rebase, not because they help:

- `MCORE_MOE_FP8_WEIGHTS` — weight-only fp8 expert weights **lose** performance: the tile has to be widened to bf16 in registers before the dot product, and that convert costs more than the halved weight traffic saves. Quantizing activations too (w8a8) does win, substantially, but is not iso-precision with the bf16 baseline this PR is measured against, so it is not wired up.
- `MCORE_QKV_SPLIT_VIEWS` — splitting QKV with `torch.split` views instead of TE's `SplitAlongDim`, tested as the suspected source of the per-layer QKV-sized copy. Exactly neutral (29,307 vs 29,313). The copy turned out to be the reshape that follows both branches rather than the split, so this arm's neutrality was the diagnostic answer — see `MCORE_GROUPED_QK_NORM` above.

## Diagnostics included

Env-gated, default-off instrumentation ships alongside the optimizations, because the measurements above are not reproducible without it: `host_phase_timing.py` (Python-side phase self-time), `step_gpu_timing.py` (CUDA-event block-graph timing and per-rank step-entry timestamps for skew), `graph_launch_timing.py` (`cudaGraphLaunch` host cost without a profiler attached, which is how the host-idle claim above was retired), `fusion_diag.py` (one-shot report of *why* a decode fusion did or did not engage — this is what found `MCORE_FUSED_ADD_NORM` silently inert), `dispatch_ballast.py` (no-op graph nodes, to price one launch), `copy_trace.py` (dispatch-level naming of copy-like ops), and `floor_ablation.py` (component ablation for chain-delta attribution).

`floor_ablation.py`'s two NVLS gates are retained as a **documented dead end** rather than a working tool, and its docstring says so: `multimem_all_gatherv_3tensor` produces the routing map that sizes all downstream grouped-GEMM work, so skipping it leaves the experts running against warmup's denser prefill map and doing *more* work — removing both collectives measures as 13.5% *slower*. Subtraction cannot size a component that decides how much work its successors do.


shanmugamr1992:
## Superseded by a 17-PR split

This PR was too large to review as one change. It has been decomposed into 16 shippable slices plus one optional instrumentation PR, each rebased onto current `main`, each carrying its own measured delta and a self-contained description. I am leaving this open as the campaign record; **please review the split PRs instead.**

The 16 shippable slices together are **+35.80%** end-to-end throughput on Qwen3-30B-A3B decode (23,265 -> 31,593 tok/s).

**MoE grouped-GEMM and dispatch**

- #6452 — fuse SiLU(gate)*up into the FC1 GEMM epilogue — `+14.81%`
- #6453 — build the MoE indirection table in one kernel — `+7.21%`
- #6454 — per-GEMM tile configs for the decode expert GEMMs — `+6.77%`
- #6455 — predicate the locality test in the top-k gather — `+1.99%`

**MoE router and combine**

- #6456 — fused softmax + top-k router selection — `+4.75%`
- #6457 — fold the CUDA-graph padding sentinel into the route mask — `+1.47%`
- #6458 — reduce the MoE combine in bf16 — `+2.42%`

**Attention: QK norm and decode kernel**

- #6459 — fuse the per-head Q and K RMSNorms — `+5.17%`
- #6460 — read the decode Q/K norm input from the grouped layout — `+1.18%`
- #6461 — pin FlashAttention 2 on the decode path — `-0.35%` (inside noise floor)
- #6462 — Blackwell-native decode attention via FlashInfer — `+1.46%`

**Residual + RMSNorm fusion**

- #6463 — fuse the residual add with the pre-MLP RMSNorm — `+3.54%`
- #6464 — fuse mlp_bda's residual add with the next layer's QKV norm — `+0.98%`

**Host path**

- #6465 — incremental attention-state update — `-0.42%` (inside noise floor)
- #6466 — eliminate dead whole-tensor work in decode bookkeeping — `-0.07%` (inside noise floor)
- #6467 — decode fast path for post_process_requests — `+0.09%` (inside noise floor)

**Optional, land last or not at all**

- #6468 — opt-in instrumentation modules used to find all of the above. No optimization of its own.

### How the numbers were produced

Each delta is leave-one-out: the full stack with exactly that change's environment gate off, against the interpolated full-stack reference at that point in the run. All 20 arms ran back-to-back in a single Slurm allocation on one OCI GB200 NVL72 node (4 GPUs, TP1/PP1/EP4/ETP1, BF16, 256 gsm8k requests at BS256, OSL1024, CUDA graphs on), so node and thermal state are held constant. Run-to-run sigma was 0.42% across three full-stack repeats; anything under 0.84% is not distinguishable from zero by this campaign.

Deltas are marginal contributions with the rest of the stack already in, and they do not sum to the total — these changes shorten the same decode step and compose sub-additively. Two caveats worth reading in place: #6452 measures much larger in-stack than it did standalone (an interaction with the tile configs in #6454), and the three host-path slices land inside the noise floor because async scheduling hides host-side savings that were real when they were first measured.

Before measuring, the stack tip was checked against this PR's head under the same protocol and they agreed to within run-to-run variance, so the split preserves behavior.

PR #6075, opened 2026-07-27
fix(training): count THD seqlen stats once under interleaved pipelining

ilml:
With virtual pipeline parallelism the schedule invokes forward_step once per (micro-batch, model chunk), and every chunk observes an identical micro-batch through its own data iterator. Since #4103 each of those calls recorded the same cu_seqlens into the THD FLOPs accumulator, while consume_seqlen_stats_in_iteration() deduplicates only by TP*CP*PP — so reported FLOPs/TFLOPs were inflated by exactly the virtual-pipeline size whenever THD + VPP were combined (the whole-model formula already covers all args.num_layers).

Record stats only from virtual stage 0 (None and 0 both denote the primary/only chunk), mirroring the equivalent fix in Megatron-Bridge (NVIDIA-NeMo/Megatron-Bridge#4750). Adds unit tests covering the VPP gating, including that legitimate per-micro-batch accumulation on stage 0 is preserved.

# What does this PR do?

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge

### DIFF
--- a/megatron/training/training.py
+++ b/megatron/training/training.py
@@ -299,7 +299,7 @@ def print_datetime(string, override_timestamp=None):
     print_rank_0(f'[{string}] datetime: {time_str} ')
 
 
-def update_seqlen_stats_from_cu_seqlens(cu_seqlens, vp_stage=None):
+def update_seqlen_stats_from_cu_seqlens(cu_seqlens):
     """Add ``sum(L_i)`` and ``sum(L_i ** 2)`` from one micro-batch's REAL ``cu_seqlens``.
 
     Args:
@@ -308,14 +308,14 @@ def update_seqlen_stats_from_cu_seqlens(cu_seqlens, vp_stage=None):
             ``cu_seqlens`` rather than ``cu_seqlens_padded`` so the FLOPs
             metric reports useful work only, not work on CP-alignment or
             end-of-sequence padding tokens.
-        vp_stage: The calling model chunk's virtual pipeline stage, or ``None``
-            when interleaved pipelining is off. Under interleaved (virtual)
-            pipeline parallelism the forward step runs once per virtual model
-            chunk for the same logical micro-batch, and every chunk observes
-            identical ``cu_seqlens``. Only virtual stage 0 contributes stats so
-            model chunking does not multiply the whole-model FLOPs estimate
-            (the formula already covers all ``args.num_layers``); ``None`` and
-            ``0`` both denote the primary/only chunk.
+
+    Call this ONCE per logical micro-batch. Under interleaved (virtual) pipeline
+    parallelism the forward step runs once per virtual model chunk for the same
+    micro-batch and every chunk observes identical ``cu_seqlens``, so callers
+    must record only from the primary chunk (``vp_stage`` ``None`` or ``0``) --
+    the whole-model FLOPs formula already covers all ``args.num_layers``, so
+    recording per chunk would multiply the estimate by the virtual-pipeline
+    size. See the call sites in ``pretrain_gpt.py`` / ``pretrain_hybrid.py``.
 
     Every rank in the same data-parallel group sees the same ``cu_seqlens`` (it is
     broadcast across TP/CP/PP). The per-micro-batch reduction stays on device --
@@ -325,8 +325,6 @@ def update_seqlen_stats_from_cu_seqlens(cu_seqlens, vp_stage=None):
     flag at ``False`` and pay zero collective cost.
     """
     global _seqlen_stats_in_iteration, _seqlen_stats_active
-    if vp_stage not in (None, 0):
-        return
     if cu_seqlens is None or cu_seqlens.numel() < 2:
         return
     # Pin the accumulator to the current CUDA device when available so the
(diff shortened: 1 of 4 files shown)


megatron/training/training.py
```diff
@@ -317,6 +325,8 @@ def update_seqlen_stats_from_cu_seqlens(cu_seqlens):
     flag at ``False`` and pay zero collective cost.
     """
     global _seqlen_stats_in_iteration, _seqlen_stats_active
+    if vp_stage not in (None, 0):
```
zhongbozhu: That's still not the correct fix because ideally each DP rank's input sequence needs to be all-reduce once per log-interval so that it won't impact the training performance 

The right fix is in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4511

zhongbozhu (approved):

zhongbozhu (changes requested):
Please take a look at the Mbridge PR

megatron/training/training.py
ilml: fixed

pretrain_gpt.py
```diff
@@ -315,8 +315,10 @@ def forward_step(data_iterator, model: GPTModel, return_schedule_plan: bool = Fa
         if cu_seqlens_padded is not None:
             cu_seqlens_padded = cu_seqlens_padded.squeeze(0)
         # Use real (unpadded) cu_seqlens to feed the FLOPs accounting: varlen
-        # attention only computes work for real tokens within each chunk.
-        update_seqlen_stats_from_cu_seqlens(cu_seqlens)
+        # attention only computes work for real tokens within each chunk. With
+        # interleaved pipelining every virtual chunk re-runs forward_step on the
+        # same micro-batch, so only virtual stage 0 records the stats.
+        update_seqlen_stats_from_cu_seqlens(cu_seqlens, vp_stage=vp_stage)
```
zhongbozhu: How about other multimodal training entry points like mimo, pretrain_vlm, etc. 

On the other hand, can we think of a way that doesn't change the API to apply this change, otherwise we need to update all the training entry points.

megatron/training/training.py
```diff
@@ -355,36 +365,45 @@ def consume_seqlen_stats_in_iteration() -> Tuple[Optional[float], Optional[float
...
-        dedup = tp_size * cp_size * pp_size
-    else:
-        # No model-parallel state -> treat as a single rank, no reduction.
-        # This is the standalone unit-test path; production always initializes mpu.
-        dedup = 1
+        dp_group = mpu.get_data_parallel_group(with_context_parallel=False)
+        # The group size is a global topology property, so every rank takes the
+        # same branch -- the skip can never leave a peer waiting in a collective.
+        if dp_group.size() > 1:
+            torch.distributed.all_reduce(t, group=dp_group)
```
zhongbozhu, line 414: if we have dynamic CP merged, this won't be needed right? how do we differentiate between regular path without seq packing scheduler versus the one with seq packing scheduler

megatron/training/training.py
```diff
@@ -355,36 +365,45 @@ def consume_seqlen_stats_in_iteration() -> Tuple[Optional[float], Optional[float
...
     if not _seqlen_stats_active:
         # BSHD path: never allocated the tensor; tell the caller to use the
         # closed-form defaults.
         return None, None
     t = _seqlen_stats_in_iteration
+    # Without distributed / model-parallel state there is a single rank and
+    # nothing to reduce. That is the standalone unit-test path; production
+    # always initializes mpu.
     if torch.distributed.is_initialized() and mpu.model_parallel_is_initialized():
-        torch.distributed.all_reduce(t)
```
zhongbozhu, line 372: why are we deleting this?

megatron/training/training.py
```diff
@@ -355,36 +365,45 @@ def consume_seqlen_stats_in_iteration() -> Tuple[Optional[float], Optional[float
...
-
-    All ranks within one DP group accumulated identical values (``cu_seqlens`` is
-    replicated across TP/CP/PP); the world all-reduce therefore overcounts by a
-    factor of ``TP * CP * PP``, which we divide out.
+    Sync cost: at most ONE all-reduce of a 2-element ``float64`` tensor and ONE
+    host sync (``tolist()``). Skipped entirely when the flag is ``False`` -- the
+    BSHD path costs zero -- and the collective is skipped when the reduction
+    group holds a single rank.
+
+    Reduction scope: the SUM runs over the PURE data-parallel group, context
```
zhongbozhu, line 388: How often are we running this all-reduce? We can run it only once per log-interval but it doesn't seem to be the case now.

pretrain_gpt.py
ilml: fixed

megatron/training/training.py
ilml: yes, dcp won't need it, but i think the regular path still needs it. 
we can use _seqlen_stats_are_global in 5907 to differentiate them.

megatron/training/training.py
ilml: its just mimicing mbridge's changes

megatron/training/training.py
ilml: once per training iteration, yes we can do log-interval as well. the payload is very tiny anyway.

PR #6081, opened 2026-07-28
[Dev] Add pipeline-parallel support to multimodal training in multimodal_dev code path

BestJuly:

# What does this PR do?

Adds pipeline-parallel (PP) support to the `examples/multimodal_dev` training path and its Qwen3.5-VL model, which previously built the full model on every rank and rejected `--pipeline-model-parallel-size > 1` outright.

Rebased onto current `dev`, which now includes #6315 (Qwen3.5-VL migrated to `HybridModel`). The PP plumbing is threaded through `HybridModel`'s native `pre_process` / `post_process` / `vp_stage` parameters.

## Issue tracking

Linked issue: N/A (example-only change, no `megatron/core` modifications)

## Summary

**Model / stage splitting**
- `MultimodalModel` accepts `pre_process` / `post_process` / `vp_stage` and forwards them to the wrapped `HybridModel`; the vision encoder is built on the **first PP stage only**.
- Surfaces `share_embeddings_and_output_weights` and `shared_embedding_or_output_weight()` on the outer module. `finalize_model_grads._allreduce_embedding_grad` resolves the module with `get_attr_wrapped_model(model_module, 'pre_process', return_model_obj=True)`, which now stops at `MultimodalModel`, and then reads *both* names off it. `Float16Module` proxies neither, so without this the tied-embedding / MTP grad all-reduce would `AttributeError`.
- `post_process` is deliberately **not** stored on the wrapper. Nothing reads it there, and storing it would make `get_attr_wrapped_model(model, 'post_process', return_model_obj=True)` — used by `schedules.clear_embedding_activation_buffer` — resolve to `MultimodalModel` instead of the `HybridModel` that owns `embedding_activation_buffer`.
- `Qwen35VLModel` skips vision-encoder construction when `pre_process=False`.
- `pretrain_multimodal.model_provider` forwards the stage flags to the registry factory, and the `PP > 1` guard is removed now that stage splitting is actually wired through.

**forward_step**
- Drops `pixel_values` off the first stage; keeps `image_grid_thw` on every stage, since MRoPE freqs are computed per stage inside `HybridModel`.
- Only CP-splits `loss_mask` on the last stage, where the pipeline scheduler invokes the loss closure.

**Static sequence length under PP**
- BSHD batches pad to `args.seq_length` when PP > 1. The non-packed branch of `pack_or_pad_batch` otherwise pads each microbatch to the per-microbatch maximum sample length, which does not match the PP scheduler's static P2P buffers. PP = 1 keeps the previous per-microbatch padding.
- Two guards make the resulting invariants explicit instead of failing deep in a P2P send:
  - `--seq-length` must be divisible by the CP/SP alignment factor under PP > 1. `pack_or_pad_batch` rounds *up* to that factor while `schedules.get_tensor_shapes` sizes the static buffers with floor division, so an unaligned length would mismatch them.
  - No sample may exceed `--seq-length` under PP > 1, since the batch is padded to a static length there.
- `--use-packed-sequence` with PP > 1 sets `config.variable_seq_lengths`, so THD activations (`[total_padded_tokens, 1, H]`) negotiate their shapes at each send/recv. The allgather-dispatcher check is repeated locally because the field is assigned after `__post_init__` has run. PP = 1 has no send/recv, so neither the flag nor the dispatcher restriction is imposed there.

**Launcher**
- `PP > 1` defaults `USE_FSDP` to 0 (FSDP and PP are mutually exclusive on the standard path), but an **explicit** `USE_FSDP=1` with `PP > 1` is now rejected rather than silently downgraded — otherwise the run proceeds without `--use-megatron-fsdp`, `--data-parallel-sharding-strategy`, `--init-model-with-meta-device` and `--ckpt-format fsdp_dtensor`, and could be mistaken for the FSDP+PP validation the caller asked for. The was-set probe uses `${USE_FSDP:+x}` so an empty `USE_FSDP=` is treated as unset, matching the `${USE_FSDP:-1}` default beside it.
- Adds `SAVE_CHECKPOINT=0` to skip `--save` / `--save-interval`. With the defaults `TRAIN_ITERS=500` and `SAVE_INTERVAL=500`, every plain run writes a full checkpoint at the end whether or not it wanted one.

## Review feedback addressed

**PP tests were invisible to CI.** The suite required a world size of exactly 2, but the unit-test bucket runs `torchrun --nproc_per_node 8`, so every test in the module skipped and the bucket reported green with zero PP coverage. The fixture now initialises `PP=PP_SIZE` and lets the remaining ranks form data-parallel replicas, so any world size that is a multiple of `PP_SIZE` executes the suite. Point-to-point sends target this rank's own pipeline group instead of the hard-coded global ranks 0 and 1, which are no longer the stage peers once the world is larger than `PP_SIZE`.

To keep this from regressing silently, `examples/multimodal_dev/tests/conftest.py` gains a bucket-level guard that fails the session if a module listed in `_MUST_RUN_MODULES` skipped while the world size was large enough to run it.

**Tied embeddings were only checked for shape, not behaviour.** The previous assertions verified that `share_embeddings_and_output_weights` and `shared_embedding_or_output_weight` merely *exist*, which would still pass with a wrong flag, a wrong parameter, or no cross-stage reduction at all. `test_tied_embedding_grads_sync_across_stages` now builds a tied model, runs a two-stage forward/backward, drives the real `_allreduce_word_embedding_grads` path, and asserts the resulting gradient equals the sum of both stages' contributions.

**Launcher mode rewriting.** Covered above; `examples/multimodal_dev/tests/test_run_script_fsdp_pp.py` exercises the unset, empty, explicit-`0`, explicit-`1` and `PP=1` cases against the script's dry-run path.

### Re-verified after the rebase onto `HybridModel`

Full `examples/multimodal_dev` CI bucket, single node, 8×H100, run the way CI runs it (`torch.distributed.run --nproc_per_node 8`, markers `not experimental and not flaky_in_dev`):

**76 passed.**

The PP tests are reported as PASSED rather than skipped at world size 8, which is the acceptance criterion requested in review. PP=1 vs PP=2 loss parity on identical weights is bit-identical:

```
[fp32]     PP=1 4.88800621 | PP=2 4.88800621 | rel 0.000E+00
[bfloat16] PP=1 4.87177610 | PP=2 4.87177610 | rel 0.000E+00
```

`test_pp_support.py` covers stage construction (vision encoder on stage 0 only), flag threading into `HybridModel`, a real two-stage forward + backward including the vision-encoder gradient path, the tied-embedding gradient synchronisation described above, and the PP=1 vs PP=2 parity check.

### Verified before the rebase, not yet re-run

These were run against the pre-rebase revision (when the decoder was still `GPTModel`) and have **not** been repeated on the current `HybridModel` base:

- End-to-end PP = 2 on mock data, MoE (16 experts, top-2, grouped GEMM, alltoall) + MTP (1 layer) + vision encoder through the real 1F1B scheduler: 10 iterations, no NaN.
- TP = 2 + `--sequence-parallel` + PP = 2, 3 iterations.
- THD + PP = 2 through the real 1F1B scheduler with microbatches packing to 48, 30 and 61 tokens against `--seq-length 64`, so no microbatch matches the static P2P buffer size. The same workload with `variable_seq_lengths` left unset hangs, which is the failure mode the change removes.
- Both `__main__` guards firing with their intended messages, and not firing at PP = 1.

**Not covered** (called out for reviewers): VPP, PP > 2, CP > 1 or EP > 1 combined with PP, GatedDeltaNet + PP, checkpoint resharding across PP, multi-node, and MTP combined with PP > 1 in the unit suite (the launcher defaults `MTP_NUM_LAYERS=1`, so the MTP branch of `_get_shared_word_embedding_weight` is reachable in a default run but is exercised only by the end-to-end run above, not by a unit test).

### Pre-checks

### DIFF
--- a/examples/multimodal_dev/scripts/run_qwen35_vl.sh
+++ b/examples/multimodal_dev/scripts/run_qwen35_vl.sh
@@ -6,13 +6,17 @@
 #   ./examples/multimodal_dev/scripts/run_qwen35_vl.sh
 #
 # Environment variables:
-#   MODEL_VARIANT: proxy (default), 0.8b, 2b, 4b, 9b, 27b, 35b_a3b, 122b_a10b, 397b_a17b, 35b_a3b_light
+#   MODEL_VARIANT: proxy (default), pp_proxy, 0.8b, 2b, 4b, 9b, 27b, 35b_a3b, 122b_a10b, 397b_a17b, 35b_a3b_light
 #   CKPT_LOAD: path to a pre-converted checkpoint to load (enables --load + --finetune)
 #   CKPT_FORMAT: checkpoint format override (e.g. torch_dist); auto-detected when empty
-#   TP, EP, PP: parallelism sizes (PP must stay 1; multimodal_dev does not
-#               support pipeline parallelism)
+#   TP, EP, PP: parallelism sizes (PP>1 forces USE_FSDP=0)
 #   MBS, GBS: micro/global batch sizes
 #   NUM_LAYERS, NUM_EXPERTS: override for proxy testing
+#   DATASET_PROVIDER: cord_v2 (default) or mock
+#   TOKENIZER_TYPE: HuggingFaceTokenizer (default) or NullTokenizer
+#   VOCAB_SIZE: required for NullTokenizer
+#   MTP_NUM_LAYERS: set to 0 to disable MTP (default 1)
+#   SAVE_CHECKPOINT: set to 0 to skip --save/--save-interval
 #   FORCE_LOAD_BALANCING: set to 1 to enable --moe-router-force-load-balancing
 #                         (perf / mock-data only; OFF for real finetuning)
 #   LAUNCHER: torchrun (default) or python
@@ -44,6 +48,16 @@ LAUNCHER=${LAUNCHER:-torchrun}
 
 MODEL_VARIANT=${MODEL_VARIANT:-proxy}
 VISION_NUM_LAYERS=${VISION_NUM_LAYERS:-}
+
+# PP proxy-run knobs: let a tiny PP job run offline against mock data
+# without a HF tokenizer/processor download, and without MTP or
+# checkpoint writes.
+DATASET_PROVIDER=${DATASET_PROVIDER:-cord_v2}
+TOKENIZER_TYPE=${TOKENIZER_TYPE:-HuggingFaceTokenizer}
+VOCAB_SIZE=${VOCAB_SIZE:-}
+MTP_NUM_LAYERS=${MTP_NUM_LAYERS:-1}
+MTP_LOSS_SCALING_FACTOR=${MTP_LOSS_SCALING_FACTOR:-0.1}
+SAVE_CHECKPOINT=${SAVE_CHECKPOINT:-1}
 
 # Batch sizes
 MBS=${MBS:-2}
@@ -105,6 +119,18 @@ case "$MODEL_VARIANT" in
         NUM_QUERY_GROUPS=2
         LINEAR_NUM_VALUE_HEADS=64
         VISION_NUM_LAYERS=${VISION_NUM_LAYERS:-2}
+        ;;
+    pp_proxy)
+        # Minimal variant for pipeline-parallel smoke / parity runs.
+        NUM_LAYERS=${NUM_LAYERS:-4}
+        NUM_EXPERTS=${NUM_EXPERTS:-4}
+        HIDDEN_SIZE=${HIDDEN_SIZE:-512}
+        FFN_HIDDEN_SIZE=${FFN_HIDDEN_SIZE:-2048}
+        NUM_ATTN_HEADS=${NUM_ATTN_HEADS:-2}
+        NUM_QUERY_GROUPS=${NUM_QUERY_GROUPS:-1}
+        LINEAR_NUM_KEY_HEADS=${LINEAR_NUM_KEY_HEADS:-2}
+        LINEAR_NUM_VALUE_HEADS=${LINEAR_NUM_VALUE_HEADS:-4}
+        VISION_NUM_LAYERS=${VISION_NUM_LAYERS:-1}
         ;;
     9b)
         NUM_LAYERS=${NUM_LAYERS:-32}
@@ -177,6 +203,8 @@ case "$MODEL_VARIANT" in
         VISION_NUM_LAYERS=${VISION_NUM_LAYERS:-27}
         ;;
 esac
+# Upstream default for every variant except the PP proxy, which shrinks it.
+LINEAR_NUM_KEY_HEADS=${LINEAR_NUM_KEY_HEADS:-16}
 
 # Fail fast on inconsistent expert-parallelism configuration. Dense variants
 # (NUM_EXPERTS=0) do not emit any --num-experts / MoE args, so forwarding
@@ -260,13 +288,18 @@ TRAINING_ARGS=(
     --enable-experimental
     --manual-gc
     --manual-gc-interval 50
-    --mtp-num-layers 1
-    --mtp-loss-scaling-factor 0.1
     --sft
     --use-flash-attn
     # --attention-backend flash
     --calculate-per-token-loss
 )
+# MTP_NUM_LAYERS=0 disables MTP (useful to isolate PP from MTP behaviour).
+if [ "$MTP_NUM_LAYERS" -gt 0 ]; then
+    TRAINING_ARGS+=(
+        --mtp-num-layers "$MTP_NUM_LAYERS"
+        --mtp-loss-scaling-factor "$MTP_LOSS_SCALING_FACTOR"
+    )
+fi
 
 PROFILE_ARGS=()
 NSYS_CMD=()
@@ -296,9 +329,7 @@ fi
 SAVE_INTERVAL=${SAVE_INTERVAL:-500}
 EVAL_AND_LOGGING_ARGS=(
     --log-interval 1
-    --save-interval "$SAVE_INTERVAL"
     --eval-interval 500
-    --save "$CHECKPOINT_STORE_PATH"
     --eval-iters 10
     --tensorboard-dir "$TENSORBOARD_LOGS_PATH"
     --wandb-project "$WANDB_PROJECT"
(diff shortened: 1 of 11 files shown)


FDecaYed:
With #6083 now in `dev`, the proposed landing order is #6315 first and this PR second, so the PP support is implemented and tested against the final HybridModel path.

Before rebasing after #6315 lands, could you address the remaining PP/launcher coverage gaps? In particular, the standard multimodal unit bucket now exists, but `test_pp_support.py` still skips unless the overall world size is exactly 2 ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6081#discussion_r3800614035)); the tied-embedding path is only checked structurally rather than behaviorally ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6081#discussion_r3800615282)); and the launcher still overrides an explicit `USE_FSDP=1` request when PP is enabled ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6081#discussion_r3757076172)). After #6315 lands, please rebase onto the then-current `dev`, adapt the PP wiring/tests to HybridModel, and ping us for exact-head review/CI.

examples/multimodal_dev/tests/test_pp_support.py
```diff
@@ -0,0 +1,409 @@
...
+
+@pytest.fixture(scope="module", autouse=True)
+def _init_pp():
+    """Initialise PP=2 model-parallel groups for the whole module."""
+    # Guard before initialising: Utils.initialize_model_parallel raises on a
+    # world size that is not a multiple of PP_SIZE, and this fixture is
+    # autouse, so checking any later would turn the intended skip into a
+    # collection-time error under the documented 1-rank invocation.
+    if Utils.world_size != PP_SIZE:
+        pytest.skip(f"needs exactly {PP_SIZE} ranks (PP={PP_SIZE})")
```
FDecaYed, line 145: This causes the entire PP test module to be skipped by the standard unit-test environment, which launches eight ranks. Now that `examples/multimodal_dev` is collected by the dedicated CI bucket, a green run can still provide zero PP coverage because every test here requires the global world size to equal 2 exactly. Please make these tests execute in the real CI topology—for example, support any world size divisible by `PP_SIZE` (PP=2 with the remaining ranks forming DP groups), or place the module in a genuinely two-rank job. The important acceptance criterion is that CI reports these PP tests as executed rather than skipped; please add/adjust the bucket-level assertion or CI configuration so this cannot regress silently.

examples/multimodal_dev/tests/test_pp_support.py
```diff
@@ -0,0 +1,409 @@
...
+    assert model.language_model.post_process is is_last
+    assert model.language_model.vp_stage == 0
+
+    # Only the first stage builds an embedding; only the last an output layer.
+    assert hasattr(model.language_model, "embedding") is is_first
+    assert hasattr(model.language_model, "output_layer") is is_last
+
+    # finalize_model_grads inspects the OUTER module for these two.
+    assert hasattr(model, "share_embeddings_and_output_weights")
+    assert callable(model.shared_embedding_or_output_weight)
```
FDecaYed, line 197: These assertions only prove that the wrapper exposes the two attributes inspected by `finalize_model_grads`; they do not exercise the behavior that is easy to break when the embedding and output layer live on different pipeline stages. Please add a tied-embedding case that performs forward/backward across PP=2, invokes the relevant final-gradient synchronization path, and verifies that the first-stage embedding and last-stage output weight receive the expected synchronized gradients. Otherwise an implementation with a callable delegate but incorrect sharing flags, parameter selection, or cross-stage synchronization would still pass this test.

examples/multimodal_dev/scripts/run_qwen35_vl.sh
```diff
@@ -459,6 +466,11 @@ fi
 
 # --- FSDP ---
 USE_FSDP=${USE_FSDP:-1}
+# FSDP and PP are mutually exclusive on Megatron's standard path.
+if [ "$PP" -gt 1 ] && [ "$USE_FSDP" -eq 1 ]; then
+    echo "[run_qwen35_vl] PP=${PP} > 1 -> forcing USE_FSDP=0"
+    USE_FSDP=0
```
FDecaYed, line 472: Please do not rewrite an explicitly requested execution mode here. If a caller sets `USE_FSDP=1` together with `PP>1`, this branch proceeds with a non-FSDP run after changing checkpoint format, optimizer/sharding, and initialization behavior; the log message does not prevent the job from being mistaken for the requested FSDP+PP validation. Please distinguish an unset variable from an explicit `USE_FSDP=1`: choose a PP-compatible default when it is unset, but fail early with a clear unsupported-combination error when the caller explicitly requests both. A small launcher test should cover the unset/default case, explicit `USE_FSDP=0`, and explicit incompatible `USE_FSDP=1`.

examples/multimodal_dev/tests/test_pp_support.py
BestJuly: Fixed in bd95f9ab9. The fixture now initialises `PP=PP_SIZE` and lets the remaining
ranks form DP replicas, so any world size that is a multiple of `PP_SIZE` runs the
suite — option (a) from your suggestion, no separate 2-rank job. This also required
moving the P2P calls off the hard-coded global ranks 0/1, which stop being the stage
peers once the world is larger than `PP_SIZE`.

For the "cannot regress silently" part I added a bucket-level assertion rather than
trusting the fixture: `conftest.py` lists modules that must execute, and
`pytest_sessionfinish` fails the session if one skipped while `WORLD_SIZE` was large
enough to run it.

Verified the way CI runs it (`--nproc_per_node 8`, dev container, 1×8 H100):
**76 passed**, PP tests reported as PASSED rather than skipped. PP=1 vs PP=2 parity is
bit-identical (`rel 0.000E+00`, fp32 and bf16).

examples/multimodal_dev/tests/test_pp_support.py
BestJuly: Replaced by `test_tied_embedding_grads_sync_across_stages`
(bd95f9ab9), which builds with `share_embeddings_and_output_weights=True`, asserts the
flag *value* and that the wrapper returns the stage-appropriate tensor (`is` the
embedding weight on the first stage, `is` the output-layer weight on the last), runs a
real PP=2 forward/backward, then drives
`finalize_model_grads._allreduce_word_embedding_grads` and checks the resulting
gradient equals the sum of both stages' pre-reduction gradients (plus a non-zero peer
contribution, so it can't pass trivially).

examples/multimodal_dev/scripts/run_qwen35_vl.sh
BestJuly: Fixed in bd95f9ab9. `USE_FSDP` unset → defaults to 0 so plain PP smoke runs still
work; `USE_FSDP=0` → runs without FSDP; `USE_FSDP=1` with `PP>1` → exits 1 with an
unsupported-combination error instead of downgrading.

Two details: the was-set probe uses `${USE_FSDP:+x}`, so `USE_FSDP=` (empty) is
treated as unset rather than hard-failing with a message claiming FSDP was requested.
And I corrected the flag list in the error text — `--use-distributed-optimizer` is
emitted unconditionally from `MODEL_PARALLEL_ARGS` and survives a downgrade; what is
actually lost is `--use-megatron-fsdp`, `--data-parallel-sharding-strategy`,
`--init-model-with-meta-device` and `--ckpt-format fsdp_dtensor`.

Launcher test added as requested (`tests/test_run_script_fsdp_pp.py`): unset, empty,
explicit `0`, explicit `1`, and `PP=1`. Drives the script's `DRY_RUN` path, so no GPU
needed.

FDecaYed (approved):

PR #6083, opened 2026-07-28
[Dev] ci: guard examples/multimodal_dev with a unit-test bucket

BestJuly:

# What does this PR do?

Runs the `examples/multimodal_dev/tests/` suite as its own H100 unit-test bucket, and repairs the two things that stopped that suite from being a usable guard.

`examples/multimodal_dev` tracks `megatron.core` APIs closely — `rope_utils`, `GPTModel`, `TransformerConfig`, the GPT layer specs — but it lives outside `tests/unit_tests`, so nothing in CI exercises it. Core changes have therefore landed without anyone noticing they broke the example.

**That is not hypothetical: it is broken on `dev` today.** The fused-mRoPE change (#5962) made `apply_rotary_pos_emb` read `config.mrope_section` and `config.apply_rope_fusion` directly, and `test_mrope_parity.py::test_vision_rope_wrapper_forwards_max_seqlen_to_thd` drives that code with a `SimpleNamespace` stub written before those fields existed:

```
AttributeError: 'types.SimpleNamespace' object has no attribute 'mrope_section'
megatron/core/models/common/embeddings/rope_utils.py:93
```

Measured on `dev` @ `9304a2564`, 1×8×H100: **1 failed, 45 passed** (identical at 1 rank and at 8).

Three commits:

1. **Repair the stale config stub.** The stub now carries `mrope_section=None, apply_rope_fusion=False`, pinning the test to the unfused legacy THD path it was written for. No change to `megatron/core`.

2. **Make the standalone harnesses real tests.** Four of the eight files (`test_cp_correctness.py`, `test_cp_thd_correctness.py`, `test_thd_correctness.py`, `test_vision_patch_merger_parity.py`) were `argparse` + `main()` scripts with no `test_*` functions — pytest imported them and collected nothing, so the BSHD-vs-THD, CP-vs-CP=1 and patch-merger parity checks they implement only ever ran by hand. Each comparison body is factored into a reusable function that both a pytest entry point and the original `main()` call, so the CLI and the tests cannot drift; a per-module `DEFAULTS` dict backs both the argparse defaults and the test parameters. CP sizes that do not divide the world size `skip` rather than fail, and every module leaves the model-parallel groups torn down, so the files work at any `--nproc-per-node` and share one pytest process. The `__main__` entry points are unchanged in behaviour and were re-verified after the refactor. Net: **+7 executing tests**.

3. **Wire up the bucket.** New `test_case: [examples/multimodal_dev/tests/**/*.py]` in `tests/test_utils/recipes/h100/unit-tests.yaml`, plus a local `conftest.py`. The bucket is collected outside `tests/unit_tests/`, so `tests/unit_tests/conftest.py` does not apply to it; the local conftest supplies exactly what `tests/unit_tests/run_ci_test.sh` depends on — the `--experimental` option its second pytest pass passes, and the `pytest_sessionfinish` hook that maps pytest's "no tests collected" (exit 5) to success.

### Scope of the bucket

`environment: [dev]`, `tag: [latest]`, deliberately:

- `tag: latest` only — the legacy ref pins an older mcore *and* an older copy of this example, so the pairing this bucket guards does not exist there.
- `environment: dev` only — validated against the dev container; `lts` is unverified and an unverified entry risks a red bucket on older TE. Widening is a one-line change once it has been observed green.

gb200 is not wired up: selection there is marker-driven (`launch_on_gb200`) and no file in this suite carries the marker, so the bucket would launch an empty job.

### Validation (1×8 H100)

| | Result |
|---|---|
| `dev` @ `9304a2564`, baseline | 1 failed, 45 passed |
| This branch, 8 ranks | **53 passed** |

The exact CI entry point was then run with the tree bind-mounted at `/opt/megatron-lm` so the script's hardcoded paths resolve as they do in CI:

```
bash tests/unit_tests/run_ci_test.sh --tag latest --environment dev \
  --bucket "examples/multimodal_dev/tests/**/*.py" --log-dir /tmp/ci-logs
```

```
53 passed, 64 warnings in 40.11s                  # pass 1: -m 'not experimental and not flaky_in_dev'
collected 53 items / 53 deselected / 0 selected   # pass 2: --experimental
RUN_CI_TEST_EXIT=0
```

Pass 2 is what the new conftest exists for: `--experimental` is accepted and the resulting exit code 5 is mapped to 0. Recipe resolution was checked through `recipe_parser.load_workloads(scope='unit-tests', model='unit-tests', test_case='examples/multimodal_dev/tests/**/*.py', environment='dev', tag='latest')`, which returns the workload plus its `mcore-pyt-dev` build dependency like every other bucket, and `find_test_cases.py` returns no `--ignore` args (no child buckets). Bucket wall clock ~40 s.

## Issue tracking

Linked issue: none. This is a test/CI change with no runtime impact.

### Pre-checks

- [ ] I have added relevant functional tests — n/a, this PR *is* test/CI wiring
- [x] I have added proper typing to my code — matches the surrounding suite's style
- [x] I have added relevant documentation — the recipe entry carries a comment explaining the `dev`/`latest` scoping, and each converted module documents its pytest invocation
- [x] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR — a no-op here: it only considers `.py` files under `megatron/core` and `tests/`, and this PR's only `tests/` change is a YAML


FDecaYed (approved):
Looks good to me. The new bucket is covered by the exact-head CI run.

PR #6094, opened 2026-07-28
Perf: skip per-param copy_ dispatch in the MXFP8 param copy-back

### LINKED ISSUE #6095: [perf] Reducing exposed cast(bf16->mxfp8) overheads for GTP+fp8-param+reuse-grad-buf
Reducing exposed BF16→MXFP8 casts, possible solutions:
- **Skip per-param dispatch**. Call update_quantized directly instead of dst.copy_. Bitwise identical. Targets the gap only (#6094).
- **CUDA-graph the copy-back**. Host cost → ~0, but attacks the same gap as the fix above. Medium risk.
- **Fused grouped cast via TE group_quantize**. Needs an output-taking TE binding upstream. High risk, cross-repo.

### LINKED ISSUE #6136: [feat] Scale efficiency
Testing / tracking / fixing GTP's scalability at large scale

fanshiqing:

# What does this PR do?
Under --reuse-grad-buf-for-mxfp8-param-ag the param all-gather carries BF16, so _post_param_sync re-quantizes each param afterwards. It did so via dst.copy_(src), which reaches the TE cast only through the aten::copy_ op, __torch_dispatch__ and quantize_. Take the whole bucket and call the quantizer directly instead. Values are bitwise identical; the cast kernels are still one per param. Grouped params keep the per-member path

## Per-iteration data workflow (per-bucket)

image

## Issue tracking
https://github.com/NVIDIA/Megatron-LM/issues/6095

## Results

Before fix:
image

After fix: 
image

### DIFF
--- a/megatron/core/fp8_utils.py
+++ b/megatron/core/fp8_utils.py
@@ -226,6 +226,46 @@ def copy_tensor_to_quantized_param(param: torch.Tensor, src: torch.Tensor) -> No
     dst.copy_(src.view(dst.shape))
 
 
+def copy_tensors_to_quantized_params(params: List[torch.Tensor], srcs: List[torch.Tensor]) -> None:
+    """List form of :func:`copy_tensor_to_quantized_param`, for a whole bucket of params.
+
+    Same values, minus the per-param ``copy_`` and tensor-subclass dispatch: the quantizer is
+    resolved up front and called directly. Cast kernels are unchanged, one per param. Worth it
+    because those casts are small and issuing them is expensive, and under
+    --reuse-grad-buf-for-mxfp8-param-ag they run inside the forward pass.
+
+    Args:
+        params: quantized model params to write into.
+        srcs: high-precision source values, one per param, in the same order.
+    """
+    if len(params) == 0:
+        return
+
+    srcs_to_cast = []
+    dsts_to_cast = []
+    quantizers = []
+    for param, src in zip(params, srcs):
+        dst = _unwrap_parameter_data(param)
+        quantizer = (
+            None
+            if is_grouped_tensor_with_quantized_storage(dst)
+            else getattr(dst, "_quantizer", None)
+        )
+        if quantizer is None:
+            # Grouped storage quantizes per member; a missing quantizer has to be built. Both
+            # cases are handled by the single-param path.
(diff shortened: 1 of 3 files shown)


jaredcasper (approved):

megatron/core/fp8_utils.py
```diff
@@ -226,6 +226,46 @@ def copy_tensor_to_quantized_param(param: torch.Tensor, src: torch.Tensor) -> No
     dst.copy_(src.view(dst.shape))
 
 
+def copy_tensors_to_quantized_params(params: List[torch.Tensor], srcs: List[torch.Tensor]) -> None:
```
zhongbozhu, line 229: Nit: `copy_multi_tensors_to_quantized_params` might be better name

zhongbozhu (approved):
LGTM

Phlip79 (approved):
Please add testing. Conditionally approving.

megatron/core/fp8_utils.py
kunlunl, line 193: Can the new `copy_tensors_to_quantized_param` fully replace this one? If it does, directly remove this function may be a cleaner way?

kunlunl (approved):

megatron/core/fp8_utils.py
fanshiqing: But the current name is already plural on both sides — `copy_tensors_to_quantized_params` vs `copy_tensor_to_quantized_param` — which is the minimal, honest distinction: same operation, list form.

And `multi_tensor_*` is a term of art here, and it means **fused**. In mcore it always denotes an apex/TE multi-tensor-applier kernel: `multi_tensor_applier`, `multi_tensor_scale`, etc.

Thus I prefer the current one :P

megatron/core/fp8_utils.py
fanshiqing: `copy_tensors_to_quantized_params` delegates to `copy_tensor_to_quantized_param` for two cases it deliberately doesn't handle: grouped quantized storage (which quantizes per expert member) and a destination with no quantizer of its own.

If you want one function, the way to do it is to fold the grouped/no-quantizer handling into the list version rather than keeping a delegate. But I suggest keep both, it does not hurt

PR #6017, opened 2026-07-24
Fix inter-iteration activation retention in HyperConnectionTransforme…

htesd:
## What does this PR do?

  `HyperConnectionTransformerLayer._forward_post_mlp_with_fused_hyper_connection` never
  releases `self.mlp_norm_manager` after calling `group_offload()`, unlike the base class
  `TransformerLayer._forward_post_mlp`, which sets `self.mlp_norm_manager = None` right
  after use.

  Because `FineGrainedActivationOffloadingInterface.__init__` stores the wrapped tensor
  unconditionally (`self.tensor = tensor`, even when `offload=False`), the manager left
  attached to the layer module keeps a reference to one microbatch's pre-MLP-norm input
  until the next forward pass overwrites it. The memory is retained across the idle window
  between iterations.

  The effect is strongly amplified with `recompute_granularity='full'`: the retained tensor
  is the one produced during the recomputation pass, and its autograd history keeps the
  layer's entire recomputation subgraph alive (stream-expanded residual buffers, h_post/BDA
  outputs, input grads). Holding one `(s, b, h)` tensor ends up pinning `(s, b, n*h)`-sized
  buffers several times over.

  ## Measured impact

  2-layer GPT config with hyper connections (`num_residual_streams=4`), bf16, mbs=1, mock
  data, `recompute_granularity='full'`, single GPU (H200), measuring
  `torch.cuda.memory_allocated()` in the idle window between iterations:
  | metric | before | after |
  |---|---|---|
  | idle floor between iterations, seq 8192 | 29.30 GB | 28.63 GB (−0.67 GB) |
  | floor growth from seq 4096 to 8192 | +0.61 GB | +0.27 GB |

  Extrapolated to a 9-layer pipeline stage at 16k sequence length, the retention is roughly
  7 GB of wasted per-GPU memory.

  The one-line fix mirrors the existing cleanup in the base class `_forward_post_mlp`.

  A possible follow-up hardening (not included here, to keep this minimal): make
  `FineGrainedActivationOffloadingInterface` not store the tensor when `offload=False`.

PR #6117, opened 2026-07-29
Fix cross-iteration gradient retention in CheckpointFunction.backward

htesd:
## What does this PR do?

`CheckpointFunction.backward` in `megatron/core/tensor_parallel/random.py` recomputes
the forward pass using detached segment inputs, runs backward on the recomputed graph,
and collects the input gradients from the detached leaves:

`grads = tuple(inp.grad if isinstance(inp, torch.Tensor) else inp for inp in detached_inputs)`

The detached leaves can remain reachable after `backward` returns when the recomputed
graph retains references to them. In that case, their `.grad` fields also keep the
input-gradient storage alive longer than required.

This change clears each detached tensor's `.grad` after collecting the returned
gradients. The `grads` tuple retains the gradient tensors, so the gradients returned by
`CheckpointFunction.backward` are unchanged. Nested checkpointing is unaffected: each
call only touches its own `detached_inputs`.

## Measured impact

On a single H200, a 2-layer GPT configuration with `num_residual_streams=4`, bf16,
micro-batch size 1, mock data, and `recompute_granularity='full'` produced the
following allocator floor, in otherwise identical runs differing only by this change:

| metric | before | after |
|---|---:|---:|
| Between-iteration floor, sequence length 8192 | 28.63 GB | 28.36 GB |

This is a reduction of 0.27 GB for the measured configuration.

Weakref instrumentation confirmed that the detached recomputation inputs and their
`.grad` storage remained reachable during the between-iteration interval before this
change, and that clearing those `.grad` fields manually released the same allocator
blocks. With this change, the gradient references are cleared immediately after
collection.

Assuming linear scaling with checkpointed segment count and sequence length, the
measured result corresponds to an estimated reduction of approximately 2.4 GB for a
9-layer pipeline stage at sequence length 16384. This figure is an extrapolation, not a
direct measurement.

### DIFF
--- a/megatron/core/tensor_parallel/random.py
+++ b/megatron/core/tensor_parallel/random.py
@@ -632,6 +632,15 @@ class CheckpointFunction(torch.autograd.Function):
         torch.autograd.backward(outputs, args)
         grads = tuple(inp.grad if isinstance(inp, torch.Tensor) else inp for inp in detached_inputs)
 
+        # Release the recomputation-leaf gradients: they were harvested into
+        # `grads` above, but the detached leaves can outlive this call (e.g.
+        # kept alive by saved-variable references from the recomputation
+        # subgraph), retaining a stale copy of the segment's input gradients
+        # across iterations.
+        for inp in detached_inputs:
+            if isinstance(inp, torch.Tensor):
+                inp.grad = None
+
         _unset_checkpointing()
         return (None, None) + grads
 


htesd:
For the record: this PR was **not** closed by me on purpose. My fork left the upstream fork network on Aug 5 (repo turned private for unrelated internal use), and GitHub auto-closes open PRs from a detached fork, attributing the close to the repo owner — hence the misleading "htesd closed this". It took #6017 down in the same second.

Re-opened as #6339 from a clean fork, on current `dev` (`c76ff61`), byte-identical change. Sorry for the noise.

PR #6132, opened 2026-07-29
fix: refresh TE fused MLP DDP hooks dynamically

nvegesna-netizen:
# What does this PR do?

Refreshes the Transformer Engine fused MLP's forwarded descendant pre-hooks at
call time so Megatron distributed-data-parallel parameter-gather hooks continue
to run when DDP enables or replaces them after the fused implementation is
constructed.

`TEFusedMLP` lazily creates its fused implementation during the first forward.
The previous implementation captured descendant hook objects only once at that
point. Distributed optimizer training can temporarily disable its parameter
gather hooks for the first iteration and enable them afterward, leaving the
fused implementation with an empty or stale snapshot. Subsequent forwards then
skip the current DDP hooks.

This change:

- caches only the descendant module identities whose calls are skipped by the
  fused implementation;
- reads their current pre-hook registries on every fused invocation;
- avoids replaying hooks registered on `TEFusedMLP` itself, since normal
  `torch.nn.Module.__call__` already executes those hooks;
- retains the existing rejection of descendant backward hooks and
  input-modifying pre-hooks; and
- adds regression coverage for hooks added, removed, and replaced after fused
  implementation construction.

## Validation

- Focused hook regression on 8 GB200 GPUs: both tests passed on every rank.
- Distributed-optimizer parameter-shard ownership test with two independent
  4-rank jobs: both parameterizations passed on both nodes.
- 50-step Llama 3 8B NVFP4 training on 8 GB200 GPUs with the Transformer Engine
  op fuser: completed successfully.
- Repository-pinned Black 24.4.2, isort 5.13.2, and Ruff 0.9.10 checks passed for
  both changed files.


nvegesna-netizen:
@NVIDIA/mcore-oncall can this be reviewed.

janEbert:
@CarlosGomes98 could you also take a look at this since you previously worked on the fused op hook code?

nvegesna-netizen:
@NVIDIA/mcore-oncall any update?

nvegesna-netizen:
Thanks — both points are valid.

The lifecycle documentation in this PR now clarifies that the descendant set is captured at build time, descendant post-forward hooks are captured at build time, and descendant backward hooks are revalidated only after `_reset_fused_impl()`.

The TP=1 `TEFusedMLPWithGroupedLinear` ordering issue is addressed separately in #6654. That follow-up uses phase-specific forwarding rather than registering the generic forwarder on the normalization sequence: FC1 and its descendants run their current pre-forward hooks before normalization, while FC2 and the remaining descendants run theirs at the main fused boundary. Descendant post-forward hooks and backward-hook validation remain on the main boundary exactly once. This avoids replaying FC2 early or invoking arbitrary descendant hooks twice.

#6654 includes deterministic red/green event-order coverage and a bounded distributed-training compatibility validation. It remains stacked on this PR and will be rebased after this PR lands.

CarlosGomes98:
Nice! LGTM

CarlosGomes98 (approved):

Phlip79 (approved):

jaredcasper (approved):

PR #6135, opened 2026-07-29
Support FP4 in 1F1B A2A overlap

dingqingy-nv:

# What does this PR do?

Adds FP4 quantization-context handling to the combined 1F1B schedule used for expert-parallel all-to-all overlap.

The schedule previously entered an inner context only for non-delayed FP8 recipes. NVFP4 layers therefore executed the fine-grained forward nodes outside `get_fp4_context()`, unlike the standard `TransformerBlock` path.

This change generalizes the schedule helper to select the appropriate FP8 or FP4 context and applies it to every fine-grained forward node. MTP layers continue to run outside the FP4 context, matching the existing policy while FP4 MTP awaits numerical validation.

This is based on Selvaraj Anandaraj's original implementation:
https://github.com/sanandaraj5597/Megatron-LM/commit/5367074e4ea0504275d43dac5d0cc8569aa90f56

It also incorporates the follow-up `nullcontext` correction:
https://github.com/sanandaraj5597/Megatron-LM/commit/5eee9144605f96abccb20a2a52e16b162f5d0859

## Issue tracking

Linked issue: N/A (small compatibility fix)

### Pre-checks

## Tests

- Added focused unit coverage for selecting the FP4 context on transformer layers.
- Added coverage confirming MTP remains outside the FP4 context.
- `isort==5.13.2 --check-only --diff` (changed files)
- `black==24.4.2 --check` (changed files)
- `ruff==0.9.10 check` (changed files)
- `python3 -m py_compile` (changed files)

GPU pytest was not run locally because the locked development environment depends on Linux CUDA packages unavailable on macOS; the focused tests are included for Linux/CUDA CI.

### DIFF
--- a/megatron/core/models/common/model_chunk_schedule_plan.py
+++ b/megatron/core/models/common/model_chunk_schedule_plan.py
@@ -1,14 +1,10 @@
 # Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
 
-from contextlib import nullcontext
 from typing import Any, Callable, Optional
 
 import torch
 from torch import Tensor
 
-from megatron.core.enums import Fp8Recipe
-from megatron.core.fp4_utils import get_fp4_context
-from megatron.core.fp8_utils import get_fp8_context
 from megatron.core.pipeline_parallel.utils import (
     AbstractSchedulePlan,
     NoopScheduleNode,
@@ -222,20 +218,7 @@ class TransformerLayerSchedulePlan:
 
     def get_low_precision_context(self):
         """Get the low-precision context for the transformer layer."""
-        from megatron.core.transformer.multi_token_prediction import MultiTokenPredictionLayer
-
-        use_inner_fp8_context = (
-            self.layer.config.fp8 and self.layer.config.fp8_recipe != Fp8Recipe.delayed
-        )
-        use_inner_fp4_context = self.layer.config.fp4 and not isinstance(
-            self.layer, MultiTokenPredictionLayer
-        )
-
-        if use_inner_fp8_context:
-            return get_fp8_context(self.layer.config, self.layer.layer_number - 1)
-        if use_inner_fp4_context:
-            return get_fp4_context(self.layer.config, self.layer.layer_number - 1)
-        return nullcontext()
+        return self.layer.get_inner_quantization_context()
 
     @staticmethod
     def run(f_layer, b_layer, f_input=None, b_grad=None, is_last_layer_in_bwd=False):
--- a/megatron/core/models/common/model_chunk_schedule_plan.py
+++ b/megatron/core/models/common/model_chunk_schedule_plan.py
@@ -7,6 +7,7 @@ import torch
 from torch import Tensor
 
 from megatron.core.enums import Fp8Recipe
+from megatron.core.fp4_utils import get_fp4_context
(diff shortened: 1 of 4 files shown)


Phlip79 (approved):

yashaswikarnati (approved):

megatron/core/models/common/model_chunk_schedule_plan.py
```diff
@@ -219,19 +220,23 @@ def set_fsdp_reshard_hooks(self, post_forward_hook, post_backward_hook):
...
+
         use_inner_fp8_context = (
             self.layer.config.fp8 and self.layer.config.fp8_recipe != Fp8Recipe.delayed
         )
-        return (
-            get_fp8_context(self.layer.config, self.layer.layer_number - 1)
-            if use_inner_fp8_context
-            else nullcontext()
+        use_inner_fp4_context = self.layer.config.fp4 and not isinstance(
+            self.layer, MultiTokenPredictionLayer
```
jaredcasper: Why is the "MTP is not in fp4" recipe baked into the code here? Shouldn't this be part of a higher level recipe or the MTPLayer itself?

dingqingy-nv: good point, thanks, updated

Connor-XY (approved):

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -1251,6 +1251,14 @@ def __init__(
         )
         self.offload_context = nullcontext()
 
+    def get_inner_quantization_context(self) -> AbstractContextManager:
+        """Return the quantization context for fine-grained MTP execution."""
+        if self.config.fp8 and self.config.fp8_recipe != Fp8Recipe.delayed:
+            return get_fp8_context(self.config)
+
+        # FP4 in MTP layers still needs numerical validation.
```
jaredcasper, line 1259: Can this be checked in the recipe? If the recipe/layer configs? If the configuration calls for fp4 MTP we should throw an error or warn and do it. Doesn't feel like we should silently ignore the configuration here (i.e. hardcoding in the policy that MTP is in full precision, not fp4.)

dingqingy-nv: Silently ignoring fp4 in MTP is the behavior in non 1f1b overlap path too (see [here](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/multi_token_prediction.py#L1607))

I agree it's better to have a recipe level flag, but I feel this is orthogonal to this PR's scope.

Could we keep it this way in this PR, and create a follow-up PR for an explicit flag `mtp_use_fp4` (default to false) for both 1f1b and non 1f1b path?

yaoyu-33 (approved):

megatron/core/transformer/multi_token_prediction.py
malay-nagda: gentle ping @jaredcasper

bdubauski:
@dingqingy-nv @erhoo82 can this PR be merged in time for 26.08.01 patch? It is a dependency for another PR that targets NVFP4 perf tuning for QWen3.

jaredcasper (approved):

PR #6137, opened 2026-07-30
Support no-shard, ZeRO-1, and ZeRO-2 in MFSDP v2

Autumn1998:
## What does this PR do?

Extend MFSDP v2 beyond its existing `optim_grads_params` (ZeRO-3) layout to support these data-parallel strategies:

| Strategy | Parameters | Gradients | Optimizer state |
| --- | --- | --- | --- |
| `no_shard` | replicated | averaged partial | replicated |
| `optim` (ZeRO-1) | replicated | averaged partial | sharded |
| `optim_grads` (ZeRO-2) | replicated | sharded | sharded |

The implementation handles no-op placement changes, preserves partial gradients across non-final microbatches, and only applies the required gradient reduction when the layout changes. It also refreshes compute weights after an optimizer step in deterministic optimizer-parameter order.

## Validation

- Two-rank SGD parity against an unsharded baseline for all four strategies (`no_shard`, `optim`, `optim_grads`, and existing `optim_grads_params`) with one and three microbatches.
- Two-rank repeated-backward gradient-accumulation test.
- `ruff check`.


megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
```diff
@@ -109,13 +109,13 @@ def step_post_hook(
             set_grad(parameter, original_grad)
         casted_grads.clear()
 
-        fsdp_parameter_groups: set[FsdpParameterGroup] = set()
+        fsdp_parameter_groups: dict[FsdpParameterGroup, None] = {}
```
Autumn1998: Because the current `sync_model_weight_from_main_weight` includes all-gather communication in zero-1 and is placed in the optimizer’s post hook, we need to maintain ordered communication across all ranks. Therefore, we must use an ordered list instead of a set. If we later move the communication in `sync_model_weight_from_main_weight` to the first microbatch, this change may no longer be necessary.

wujingyue: Apparently, this PR doesn't try to use values of the dict at all. What am I missing?

wujingyue (commented):
Is this ready for review? It's shown as draft currently.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
```diff
@@ -109,13 +109,13 @@ def step_post_hook(
...
         for optimizer_group in hooked_optimizer.param_groups:
             for parameter in optimizer_group["params"]:
                 parameter_group = get_containing_parameter_group(parameter)
                 if parameter_group is None:
                     continue
-                fsdp_parameter_groups.add(parameter_group)
+                fsdp_parameter_groups[parameter_group] = None
 
         for parameter_group in fsdp_parameter_groups:
             parameter_group.sync_model_weight_from_main_weight()
```
Autumn1998: In the optimizer post hook, we convert the main weights into model weights. In ZeRO-1, the optimizer-facing weights are stored as `flat` shards, while the model weights are `replicated`, so there is an all-gather here. We would like to move this part into the first microbatch so that we can overlap between different layers. @wujingyue Do we have a corresponding design change for this? 🤔

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -289,15 +290,31 @@ def reduce_partial_gradients(
...
+        finalize_axis = changed_mesh_axis(self.main_grad.placements, self.main_weight.placements)
+        if finalize_axis == self.mesh.ndim - 1 and isinstance(
+            self.main_grad.placements[finalize_axis], Partial
+        ):
+            with self._symmetric_memory_context():
+                partial_grad = partial_grad.cast(self.main_grad.dtype)
+            if not is_last_microbatch:
+                self.main_grad.local_buffer.add_(partial_grad.local_buffer)
+                return
+            partial_grad.local_buffer.add_(self.main_grad.local_buffer)
```
Autumn1998: For no_shard and optim, MFSDP v2 currently accumulates `partial gradients` into `main_grad` and performs the final RS on the last microbatch.
However, many kernels fuse gradient accumulation by writing directly into main_grad, bypassing FSDP’s explicit accumulation step. The current implementation is not compatible with this fused path.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -289,15 +290,31 @@ def reduce_partial_gradients(
...
+            if self._symm_mem_pool is not None:
+                partial_grad.rendezvous(finalize_axis)
+            optimizer_grad = partial_grad.redistribute(self.main_weight.placements)
+            if partial_grad.placements[finalize_axis].reduce_op == dist.ReduceOp.SUM:
+                optimizer_grad.local_buffer.div_(self.mesh.size(finalize_axis))
+            self._install_sharded_grads(optimizer_grad)
```
Autumn1998: Here, we install `optimizer_grad` instead of `main_grad` because `main_grad` has a Partial placement, while the optimizer requires a Flat shard. 
The accumulation-gradient layout differs from the optimizer-gradient layout, so we may need a separate accumulation_grad buffer.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
wujingyue: Good question. sync_main_weight_to_model_weight, invoked by the post-optimizer-step hook, does a cast followed by a redistribute today: https://github.com/NVIDIA/Megatron-LM/blob/77c87725e71c9f5022ac289071b16941876ca858/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py#L209-L211. 

Do we prefer to
1. keep cast in post-optimizer-step but redistribute (usually an allgather) in the first microbatch, or
2. keep both in the first microbatch, or
3. keep both in post-optimizer-step as is today? 

My gut feeling is (1), because: 
1. compute, which is on the critical path, should be batched for shorter latency
2. communication, which should be hidden behind compute, should be pipelined and exposed as little as possible

As the **last resort**, we could also introduce a user knob.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue: Yes, I'm aware of this unimplemented optimization. I believe it can be added by allocating reduce_scatter input before calling the fused backward+copy_in kernel. In fact, for exactly this extension, I chose to launch backward and copy_in on the same stream: http://nv/mfsdp-schedule-design => alternatives considered => FSDP2.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue: https://github.com/NVIDIA/Megatron-LM/pull/6041 should make this unnecessary.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
Autumn1998: 1. Different ranks may iterate over the same set in different orders. A dict preserves insertion order, so it produces the same order when optimizer parameters are ordered consistently across ranks.

  2. This matters because the ZeRO-1 post-optimizer hook performs all-gather communication for each parameter group. Different iteration orders could cause ranks to launch collectives in different sequences, potentially resulting in a deadlock.

  3. A set cannot guarantee iteration order because it is implemented as a hash table. Its order depends on object hashes and the table’s internal layout. The dict is therefore used as an ordered set: its values are intentionally unused, while its keys deduplicate parameter groups and preserve their first-insertion order.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
Autumn1998: I also think option (1) is the better choice. In addition, we should find a clean way to determine whether it is the first microbatch. Perhaps it would make sense to first open a PR specifically to identify the “first microbatch.” 🤔

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: let me refine this PR with #6041  🤔

Autumn1998:
> Is this ready for review? It's shown as draft currently.

It’s still a draft; I’ll include #6041 and revise it.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
wujingyue: > identify the "first microbatch."

Remember https://github.com/NVIDIA/Megatron-LM/pull/5652 tried to add `is_first_microbatch` but got pivoted to `is_last_microbatch` because FSDP2 doesn't annotate first microbatch? 

However, given it's only about redistribution, maybe something like https://github.com/NVIDIA/Megatron-LM/blob/ff6b92a7c3711c41f8169663614cd865233468ca/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py#L322-L323 will work without the user having to annotate?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
wujingyue: Aha, I missed that `dict` preserves insertion order and `set` doesn't.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue: #6041 has been merged FYI

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
Autumn1998: We actually don’t need the user to specify the first microbatch; we can just determine it ourselves within the param group.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
```diff
@@ -109,13 +109,13 @@ def step_post_hook(
...
         casted_grads.clear()
 
-        fsdp_parameter_groups: set[FsdpParameterGroup] = set()
+        fsdp_parameter_groups: dict[FsdpParameterGroup, None] = {}
         for optimizer_group in hooked_optimizer.param_groups:
             for parameter in optimizer_group["params"]:
                 parameter_group = get_containing_parameter_group(parameter)
                 if parameter_group is None:
                     continue
-                fsdp_parameter_groups.add(parameter_group)
```
wujingyue: I'd do this instead:

```
if parameter_group in synced_parameter_groups:
    continue
synced_parameter_groups.add(parameter_group)
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: This is still fairly subtle. Take ZeRO-1 as an example. During gradient accumulation, main_grad uses a Partial placement. On the last microbatch, it is finalized with a reduce-scatter and becomes the optimizer-facing gradient with a Flat placement. Therefore, the lifecycle contains two transitions: Partial -> Flat on the last microbatch, followed by Flat -> Partial on the first microbatch of the next accumulation cycle.

 These transitions should be handled differently. Partial -> Flat requires a real reduce-scatter and allocates a new output DBuffer. In contrast, Flat ->  Partial should not perform communication: the Flat gradient belongs to the previous optimizer step and can be discarded after the optimizer has consumed  it, so we can directly allocate a fresh Partial accumulation buffer.

 The storage dependencies must still be maintained correctly. We need a record stream:
```
Partial buffer A
          │
          │ enqueue Reduce-Scatter (asynchronous)
          ▼
    Flat buffer B

    self.main_grad = B
          │
          └── A loses its last reference and is released by the allocator
                        │
                        ▼
                 A's storage is reused
                        │
                        ▼
           Reduce-Scatter is still reading A → corrupted gradients
```
Since main_grad is replaced by a new DBuffer every global batch—either returned by redistribute() or allocated directly—could its storage address become  unstable? Would CUDA graph capture or fused gradient accumulation require this address to remain stable across iterations?

Autumn1998: I think an alternative approach is to create an accumulate grad (which uses partial/replicate placement) and separate it from the main grad (which is the flat shard), but let them share the same storage — for example, the main grad would be a view of the accumulate grad. @wujingyue What do you think?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
Autumn1998: I still little confusing: how does this ensure that the order in which communication is issued on each rank is the same?

wujingyue: synced_parameter_groups is **only** used for deduplication and is **never** traversed. As long as the parameter order in `optimizer.param_groups` is uniform, we are good.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue: > Reduce-Scatter is still reading A → corrupted gradients

I doubt this. All these operations in the above diagram are enqueued to the same, reduce-scatter stream. So it should be safe to free after last use. https://dev-discuss.pytorch.org/t/fsdp-cudacachingallocator-an-outsider-newb-perspective/1486#p-2441-v1-single-stream-no-nonsense-cudacachingallocator-3

> We need a record stream

Try our best not to. The [whole blogpost](https://dev-discuss.pytorch.org/t/fsdp-cudacachingallocator-an-outsider-newb-perspective/1486) is pretty much to teach people not to `record_stream` for FSDP.

wujingyue: > I doubt this. 

I take this back. I think you might have found the root cause of https://github.com/NVIDIA/Megatron-LM/pull/5956#issuecomment-5136138642. https://github.com/NVIDIA/Megatron-LM/blob/83392004c81b6e477002ae84d4f0b1271518c1d1/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py#L345 deallocates the old `main_grad` on a different stream than it was allocated, which invites race conditions. I'll figure out a way forward ASAP. 

cc @Achyuthan-S as well for the HFSDP CI failure he's investigating

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -44,6 +44,7 @@ class FsdpParameterGroup:
     mesh: DeviceMesh
     dtype: torch.dtype
     requires_grad: bool
+    is_first_microbatch: bool
```
wujingyue: Not a big fan of states. Can we infer it from other fields, e.g., placements?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -283,71 +281,58 @@ def reduce_partial_gradients(
...
+                if self.is_first_microbatch:
+                    self.main_grad.local_buffer.copy_(reduced_grad.local_buffer)
+                else:
+                    self.main_grad.local_buffer.add_(reduced_grad.local_buffer)
 
+        optimizer_grad = self.main_grad
         if is_last_microbatch:
-            # Finalize the deferred DP-outer reduction (all-reduce for HSDP,
-            # reduce-scatter for HFSDP) before binding the sharded parameter grads.
-            self.main_grad = self.main_grad.redistribute(self.main_weight.placements)
```
wujingyue, line 438: I understand why you don't like this because of https://github.com/NVIDIA/Megatron-LM/pull/6137#discussion_r3688414858. Does https://github.com/NVIDIA/Megatron-LM/pull/6187#event-28807447539 change your mind?

wujingyue (commented):
Thanks for the PR! I left some comments to address before diving into details.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: Yes, now we can simplify the code.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: That should work. However, if we want to move the ZeRO-1 unshard from the optimizer post-hook to the first forward, wouldn’t we still need an is_first_microbatch flag? 🤔

wujingyue: For ZeRO-1, the placements should be:
```
main_grad.placements=[Partial]
main_weight.placements=[Shard]
model_weight.placements=[Replicate]
```
The post-optimizer-step hook syncs main_weight to model_weight without redistribution so model_weight will be [Shard]. The first microbatch detects `model_weight.placements` differs from what's specified and therefore allgathers `model_weight`. 

How does that sound?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -150,6 +150,19 @@ def _flat_placements() -> Placements:
     return Placements(dp_axes=[0], parameter=[Flat()], gradient=[Flat()], optimizer=[Flat()])
 
 
+def _strategy_placements(sharding_strategy: str) -> Placements:
```
wujingyue: The `sharding_strategy` string seems to be an unnecessary indirection. Can we just create several functions, e.g., `def _zero_1_placements`(), and parametrize the placements with `ids=`?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -349,26 +347,35 @@ def reduce_partial_gradients(
         has_sharded_grads = self._has_sharded_grads()
 
         # A non-accumulation main_grad means the previous step finalized it; this
-        # only happens on the first microbatch. Redistribute it back to the
-        # DP-outer-Partial accumulation placement -- a metadata relabel for HSDP,
-        # and a fresh reduce-scattered buffer for HFSDP in the future.
+        # only happens on the first microbatch. Start with fresh accumulation storage
+        # rather than trying to redistribute a Flat optimizer shard back to Partial.
         if self.main_grad.placements != self._accumulation_placements:
```
wujingyue: The changes starting from this line are a bit too convoluted. Can you try your best to simplify before I look more closely? Try to minimize the number of conditions.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -174,9 +187,14 @@ def _mb(num_bytes: int) -> str:
 _GEMM_OP_NAME_SUBSTRING = "aten::mm"
 
 
+@pytest.mark.parametrize(
+    "sharding_strategy", ["no_shard", "optim", "optim_grads", "optim_grads_params"]
+)
 @pytest.mark.parametrize("num_microbatches", [1, 3])
-def test_fully_shard_sgd_losses_match_baseline(distributed_setup, num_microbatches):
-    """Minimal per-module FSDP training should match single-rank SGD."""
+def test_fully_shard_sgd_losses_match_baseline(
```
wujingyue, line 197: Note to myself: may want to separate this to test_hybrid.py. This file is getting huge.

wujingyue, line 197: It's good to have this mini-convergence test. In addition to that, we may also want to test memory and overlapping etc to prevent https://github.com/NVIDIA/Megatron-LM/pull/6357

(They can be added in separate PRs.)

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: fixed now

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: I saw that you made changes to this part. Do you still think it’s too complex and needs further simplification?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: The current ZeRO-1 probably still doesn’t have overlap, because the AG is still in the post-optimizer hook. Should we change that in this PR? 🤔

wujingyue: > Should we change that in this PR?

Up to you. 

I'd do it if it's not too complicated. The PR is currently +68/-48 so we can accept some more changes.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
```diff
@@ -109,16 +109,21 @@ def step_post_hook(
...
+                if parameter_group in seen_parameter_groups:
+                    continue
+                seen_parameter_groups.add(parameter_group)
 
-        for parameter_group in fsdp_parameter_groups:
-            parameter_group.sync_model_weight_from_main_weight()
+                context = parameter_group.get_context()
+                allgather_stream = context.allgather_stream
+                allgather_stream.wait_stream(context.current_stream())
```
wujingyue: Why does casting need to be done on allgather_stream? Removing this would avoid having to expose `get_context`.

wujingyue: @Autumn1998 I haven't reviewed your other changes in this PR yet. If the PR is ready to review again, click the "re-request review" icon near my username so I know!

Autumn1998: In regular ZeRO-3, the cast here converts a flat FP32 main weight into a flat BF16 model weight. Essentially, it is:
`main_weight.local_buffer.copy_(model_weight.local_buffer)`
So this is an in-place replacement: main_weight is not replaced out of place.
However, in ZeRO-1, our model weight is replicated. That means that, after casting, the buffer size of the casted main weight differs from the buffer size of the model weight. Therefore, we need to use a flat model weight to replace the replicated model weight, and then restore it to a replicated weight through all-gather (AG) in the first microbatch.
```
ZeRO-1
────────────────────────────────────────────────────
Cast result: Flat BF16 tensor
[current-rank shard only]
          │
          │ out-of-place replacement
          ▼
Model weight: Replicated tensor
[ shard 0 | shard 1 | ... | shard N-1 ]
          │
          │ AG during the first microbatch
          ▼
Restored replicated model weight
[ valid 0 | valid 1 | ... | valid N-1 ]
```
During this process, we need to create a new flat model weight. It is allocated on the default stream, but its later AG operation clearly needs to run on the AG stream. As a result, we would need to record the AG stream. To avoid that, we should allocate this flat model weight directly on the AG stream.

Alternatively, we could support a special property for Replicate in dbuffer—say, flat_dirty. This would mean that, although the tensor is replicated, the data corresponding to the current rank’s flat shard is valid, while the data for all other shards is stale. We could then provide an in-place AG operation to restore this Replicate tensor.

```
All operations are in place
══════════════════════════════════════════════════════════════════

(1) In-place cast/write into the current rank's shard

Flat FP32 main-weight shard
[ local FP32 data ]
          │
          │ cast + copy_ (in-place)
          ▼
Replicate model weight (flat_dirty = true)
[ dirty | dirty | local BF16 valid | dirty | dirty ]

The Replicate buffer is retained.
No new Flat model-weight buffer is allocated.
No out-of-place replacement occurs.


(2) In-place all-gather before / during the first microbatch

Replicate model weight (flat_dirty = true)
[ dirty | dirty | local valid | dirty | dirty ]
          │
          │ inplace_all_gather()
          ▼
Replicate model weight (flat_dirty = false)
[ valid | valid | valid | valid | valid ]

Same tensor object.
Same underlying buffer.
Only its contents and logical state change.
```

wujingyue: > Therefore, we need to use a flat model weight to replace the replicated model weight, and then restore it to a replicated weight through all-gather (AG) in the first microbatch.

Makes sense. I think this creates another reason for separating cast (done in optimizer) and allgather (done in pre-forward of the first microbatch) in this PR. This way, we can remove this with-allgather-stream scope and instead allocate `model_weight` on the allgather stream. 

(I thought you already separated them given you renamed `sync` to `cast. But I could be missing something.)

Autumn1998: Yep, I’ve separated them. However, we may still need to perform the cast in ag_stream, because the cast in zero-1 creates a new DBuffer.

```
  Main Stream                    AG Path (Host Submission / GPU Execution)
  ───────────                    ─────────────────────────────────────────

  FP32 main_weight
         │
         │ cast
         ▼
  BF16 Flat A  ────────────────► B = DBuffer(...)          [Host allocation]
                                 all_gather(A → B)          [Host submission]
                                        │
                                        ├── GPU reads A
                                        └── GPU writes B
                                 self.model_weight = B      [Host assignment]
                                        │
                                        ▼
                                 Last reference to A dropped
                                 A's memory may be reused
                                        │
                                        ▼
                                 ⚠ GPU may still be reading A

```

wujingyue: > because the cast in zero-1 creates a new DBuffer.

That's right. Let me think about it...

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -220,6 +225,15 @@ def __init__(
         self._switch_to_sharded_parameters()
         self._unsharded_model_weight.release_storage()
 
+    def get_context(self) -> "FsdpContext":
```
wujingyue: I'd move the with-all-gather-stream scope inside ParameterGroup.cast_main_weight_to_model_weight so we don't have to expose `get_context` outside ParameterGroup. 

https://en.wikipedia.org/wiki/Single-responsibility_principle

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -242,23 +256,32 @@ def _switch_to_unsharded_parameters(self) -> None:
...
+        """Cast optimizer weights and restore the configured model-weight placements."""
+        self.cast_main_weight_to_model_weight()
+        self.restore_model_weight()
+
+    def cast_main_weight_to_model_weight(self) -> None:
+        """Cast optimizer weights to the model-weight dtype."""
         if self.main_weight is self.model_weight:
             return
 
         if self.main_weight.placements == self.model_weight.placements:
```
wujingyue: Cast is supposed to be only about dtype. Can we avoid having this placement check here? 

Also, `DBuffer.cast` already shortcuts: https://github.com/NVIDIA/Megatron-LM/blob/8971d19dda0f2bda87783353dde93a5134422a96/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py#L284. So I suspect we can remove the entire method and just call `self.model_weight = self.main_weight.cast(self.model_weight.dtype)` instead.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -242,23 +256,32 @@ def _switch_to_unsharded_parameters(self) -> None:
...
 
-        # main_weight is typically the higher-precision optimizer dtype, while
-        # model_weight is the lower-precision compute dtype. Cast before redistributing
-        # so cross-rank communication moves the smaller compute-dtype payload.
-        self.main_weight.cast(self.model_weight.dtype).redistribute(
-            self.model_weight.placements, out=self.model_weight
-        )
+        self.model_weight = self.main_weight.cast(self.model_weight.dtype)
+
+    def restore_model_weight(self) -> None:
```
wujingyue: Not worth a helper. Code is more readable when we inline this. 

Bottom line: optimize for readability, not conciseness.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -242,23 +256,32 @@ def _switch_to_unsharded_parameters(self) -> None:
             self._set_module_parameter(fsdp_parameter.fqns, fsdp_parameter.unsharded)
 
     def sync_model_weight_from_main_weight(self) -> None:
-        """Refresh compute weights from optimizer weights."""
+        """Cast optimizer weights and restore the configured model-weight placements."""
+        self.cast_main_weight_to_model_weight()
+        self.restore_model_weight()
```
wujingyue: Should we move communication into the first microbatch? Again, it's quite easy to miss this because the helper name `restore_model_weight` doesn't even say redistribution/communication -- another reason to get rid of the helper.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
wujingyue: I don't see an *easy* way around so I'm fine with getting the stream from the containing context in this PR. 

Do you like the idea of attaching to a DBuffer the stream from which it's allocated? It violates https://en.wikipedia.org/wiki/Single-responsibility_principle but may be good for
1. error checking. We can check `DBuffer.__del__` happens on the same stream because it's so easy to break that contract. 
2. allocating the replacement on the same stream as we are trying to do here.

Autumn1998: I am fine with adding streams to dbuffer 🤔

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: fixed, moved this scope into the  ParameterGroup

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: Yes, that is correct. This function has been simplified.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: While testing, I found an issue here. If we want to place this AG on the symm context, we need to allocate a new dbuffer in the symm context and call rendezvous(). Simply moving redistribute() to the symm context is insufficient, because the required rendezvous() step would be missing.

Therefore, I kept this helper and renamed it to better reflect its purpose.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: Yes, we do not need the AG here, fixed

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/optimizer.py
wujingyue: > We can check DBuffer.__del__ happens on the same stream because it's so easy to break that contract.

This doesn't work. The current stream when calling `__del__` doesn't matter because CUDACachingAllocator remembers the stream and will deallocate with respect to that stream anyway. In fact, the teardown happens when the current stream is the default stream, which is fine. The key instead is to make sure all uses the buffer are **joined back** to the allocation stream (e.g. via wait_event) before `__del__`. This is much harder to check. 

However, we can still add some partial checks in the user code of DBuffer, e.g., checking the current stream is the same as the allocation stream right before `self.main_grads = DBuffer(...)`. But the original idea of checking things in `DBuffer.__del__` would generate too many false positives.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue:
```suggestion
    def _get_context(self) -> "FsdpContext":
```
I don't want to expose it for now. I'm also in general nervous about back edges.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -258,24 +275,48 @@ def _switch_to_unsharded_parameters(self) -> None:
...
+            restored_model_weight = DBuffer(
+                mesh=self.mesh,
+                placements=placements,
+                tensor_shapes=self.model_weight.layout.tensor_shapes,
+                dtype=self.model_weight.dtype,
+                device=self.model_weight.device,
+            )
+        gather_axis = changed_mesh_axis(self.model_weight.placements, placements)
+        assert gather_axis is not None
+        restored_model_weight.rendezvous(gather_axis)
```
wujingyue: Can this be done inside DBuffer's constructor?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -286,12 +327,10 @@ def unshard_parameters(self) -> None:
...
         )
-        if gather_axis is None:
-            raise RuntimeError("FSDP parameter unshard requires a changed placement axis.")
         with torch.autograd._unsafe_preserve_version_counter(
             self._unsharded_model_weight.local_buffer
         ):
-            if self._symm_mem_pool is not None:
+            if self._symm_mem_pool is not None and gather_axis is not None:
                 self._unsharded_model_weight.rendezvous(gather_axis)
             self.model_weight.redistribute(
```
wujingyue: Is this an unnecessary memcpy for non-FSDP, because `model_weight_placements=[Replicate()]`? I'll think about a way for doing this nicely...

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: Because rendezvous needs to know which mesh dimension it operates on, performing it in the constructor would require iterating over every possible subgroup of the mesh. It would be cleaner to handle it in the redistribute AG instead.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: Yes , do we need to fix this issue in this PR? I don’t have a good idea for now

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +514,71 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+    with fully_shard_context(device=device):
+        fully_shard(model, mesh=mesh, placements=placements)
+
+    (parameter_group,) = model.parameter_groups
+    (fsdp_parameter,) = parameter_group.fsdp_parameters
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.01, foreach=False)
+    fully_shard_optimizer(optimizer)
+
+    # Warm up.
+    with torch.no_grad():
```
wujingyue: Why?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +514,71 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+        for _ in range(2):
+            model(x)
```
wujingyue: we don't warm up the backward?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +514,71 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+    )
+    with fully_shard_context(device=device):
+        fully_shard(model, mesh=mesh, placements=placements)
+
+    (parameter_group,) = model.parameter_groups
+    (fsdp_parameter,) = parameter_group.fsdp_parameters
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.01, foreach=False)
+    fully_shard_optimizer(optimizer)
+
+    # Warm up.
```
wujingyue: Why do we even need a warmup?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +514,71 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+    fp32_size = torch.empty((), dtype=torch.float32).element_size()
+    bf16_size = torch.empty((), dtype=dtype).element_size()
+    theoretical_optimizer_size = 2 * shard_numel * fp32_size
+    # Replicated weight replaces its Flat shard, plus the retained output.
+    theoretical_forward_growth = (
+        dim * dim - shard_numel
+    ) * bf16_size + num_tokens * dim * bf16_size
+
+    assert actual_optimizer_size == theoretical_optimizer_size
+    actual_forward_growth = torch.cuda.memory_allocated(device) - allocated_before_forward
```
wujingyue: Don't we want to test the peak?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -624,7 +711,14 @@ def test_overlaps_communication_and_compute(distributed_setup, use_symmetric_mem
...
     dim = 16384
     num_children = 4
+    num_microbatches = 3
     dtype = torch.bfloat16
+    placements = Placements(
+        dp_axes=[0],
+        parameter=[Flat() if strategy == "zero3" else Replicate()],
+        gradient=[Partial(dist.ReduceOp.AVG) if strategy == "zero1" else Flat()],
+        optimizer=[Flat()],
+    )
```
wujingyue: https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

Try to create some module-level Placements or add some module-level functions (e.g. _zero2_placements) that create these Placements.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -594,6 +680,7 @@ def test_root_backward_returns_to_resting_memory(distributed_setup):
     )
 
 
+@pytest.mark.parametrize("strategy", ["zero1", "zero2", "zero3"])
```
wujingyue: In the other test, you use "optim", "optim_grad", and "optim_grad_param". Can we at least be consistent? I slightly prefer zero* because they are more commonly used outside the project.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -653,50 +747,77 @@ def test_overlaps_communication_and_compute(distributed_setup, use_symmetric_mem
...
+                model(x).sum().backward()
+        optimizer.step()
 
     train_one_iteration()
     torch.cuda.synchronize(device)
 
     with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA]) as prof:
         train_one_iteration()
+        with torch.no_grad():
+            model(x)
```
wujingyue: Why?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -653,50 +747,77 @@ def test_overlaps_communication_and_compute(distributed_setup, use_symmetric_mem
...
+            num_microbatches * (num_children - 1),
+        ),
+        "zero3": (
+            (2 * num_microbatches + 1) * num_children,
+            num_microbatches * num_children,
+            (2 * num_microbatches + 1) * (num_children - 1),
+            num_microbatches * (num_children - 1),
+        ),
+    }
+    expected_ag, expected_rs, expected_ag_overlap, expected_rs_overlap = expected[strategy]
```
wujingyue: https://google.github.io/styleguide/pyguide.html#316-naming

avoid abbreviation

wujingyue (commented):
I cleaned up the non-test part, and it now looks good to me. 

Tests need a bit more work. Can you give it a shot? Other than that, LGTM!

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: Fix: add `_zero0_placements()`, `_zero1_placements()`, and `_zero2_placements()`, and use `_flat_placements()` for the ZeRO-3 strategy.``

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: Sure, used  `_zerox_placements()` instead.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: Sorry, a separate warm-up is unnecessary here because the following `model(x).sum().backward()` already serve that purpose:

Why warm up:
  - optimizer.step() lazily initializes the optimizer states, then we can verify that their sharded sizes match the theoretical values.
  - The initial forward/backward also initializes one-time CUDA/NCCL resources. We record the memory baseline after the optimizer step and then measure the next forward’s peak growth to verify that the ZeRO-1 all-gather introduces only the expected model-weight allocation.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: Removed

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: Removed

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: After the initial training step, the persistent memory used by the weights and optimizer states has stabilized. We then reset the peak-memory statistics and use the next forward pass to verify that the peak growth matches the expected value.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: For ZeRO-1, the all-gather is triggered only in the next training step after optimizer.step(). That is why an additional forward pass was originally added, to ensure that the ZeRO-1 all-gather was captured.

  I agree that this was confusing, so I changed the test to run two steps, with two microbatches per step. This covers all relevant paths, and each ZeRO strategy now has distinct expected collective counts.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: Fixed now. Thanks for review!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -259,43 +264,49 @@ def _switch_to_unsharded_parameters(self) -> None:
...
+        context = self._get_context()
+        # TODO: Retrieve the all-gather stream directly from self.model_weight after
+        # https://github.com/NVIDIA/Megatron-LM/pull/6441 merges.
+        allgather_stream = context.allgather_stream
+        current_stream = context.current_stream()
+        allgather_stream.wait_stream(current_stream)
+        with torch.cuda.stream(allgather_stream):
+            self.model_weight = self.main_weight.cast(self.model_weight.dtype)
+        # CUDA graph capture requires every forked stream to rejoin the capture
+        # stream before capture ends. Keep the eager path asynchronous.
```
Autumn1998: We create the new DBuffer for the model-weight cast on the all-gather side stream, which introduces a side stream into optimizer.step(). When optimizer.step() is captured independently, CUDA requires every participating side stream to rejoin the capture stream before capture ends. Otherwise, capture fails with cudaErrorStreamCaptureUnjoined. Therefore, this change adds a capture-only join.

  The behavior during CUDA graph capture is slightly unusual: although the cast is scheduled synchronously on a side stream, it is effectively serialized with both the preceding and subsequent work. We do not expect the cast to overlap with other work here, but this may still look confusing.

  @wujingyue , do you think this handling is reasonable?

wujingyue: I'd wait_stream unconditionally. 

We could also move the waits to the caller to reduce the number of waits and/or enable overlapping. But we would have to carefully cluster ParameterGroups by context, because an optimizer may optimize parameters from different FsdpContexts.

Autumn1998: As discussed previously, to avoid exposing the `_get_context()` outside the parameter group,  changed this to unconditionally wait.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +524,61 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+        fully_shard(model, mesh=mesh, placements=placements)
+
+    (parameter_group,) = model.parameter_groups
+    (fsdp_parameter,) = parameter_group.fsdp_parameters
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.01, foreach=False)
+    fully_shard_optimizer(optimizer)
+
+    # Initialize sharded Adam states.
+    model(x).sum().backward()
+    optimizer.step()
```
wujingyue: Why?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +524,61 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+    dtype = torch.bfloat16
+    x = torch.ones(num_tokens, dim, device=device, dtype=dtype)
+    model = ElementwiseModel(dim).to(device=device, dtype=dtype)
+    mesh = init_device_mesh(device.type, (world_size,))
+    placements = _zero1_placements()
+    with fully_shard_context(device=device):
+        fully_shard(model, mesh=mesh, placements=placements)
+
+    (parameter_group,) = model.parameter_groups
+    (fsdp_parameter,) = parameter_group.fsdp_parameters
```
wujingyue: Why?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +524,61 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.01, foreach=False)
+    fully_shard_optimizer(optimizer)
+
+    # Initialize sharded Adam states.
+    model(x).sum().backward()
+    optimizer.step()
+    assert parameter_group.model_weight.placements == (Flat(),)
+    torch.cuda.synchronize(device)
+    torch.cuda.empty_cache()
```
wujingyue: Why? Does it affect memory_allocated below?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +524,61 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+    allocated_before_forward = torch.cuda.memory_allocated(device)
+    torch.cuda.reset_peak_memory_stats(device)
+
+    # Restore replicated model weight.
+    with torch.no_grad():
+        output = model(x)
+    torch.cuda.synchronize(device)
+    assert parameter_group.model_weight.placements == (Replicate(),)
```
```suggestion
```
https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
wujingyue: > For ZeRO-1, the all-gather is triggered only in the next training step after optimizer.step().

Really? `ParameterGroup.__init__` creates `model_weight` as a quantized/casted `main_weight`. So the first step should trigger allgather already.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: In theory, it shouldn’t. Since we wanted to measure the memory growth of Zero-1 AG during a single forward pass, so cleared the cache beforehand to make the subsequent forward-pass memory change more consistent. However, it can indeed be removed here, since empty_cache() does not affect allocated memory.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: I found that PyTorch Adam lazily creates `exp_avg` and `exp_avg_sq` during the first `step()`. This step ensures that the sharded optimizer states exist before we check their sizes and record the memory baseline for next forward checking.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: removed

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: I was using fsdp_parameter as an indirect way to access the optimizer state. Since the entire model is managed by FSDP in this test, iterating over all optimizer states  maybe clearer

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -653,50 +750,86 @@ def test_overlaps_communication_and_compute(distributed_setup, use_symmetric_mem
...
+                    model(x).sum().backward()
+            optimizer.step()
 
-    train_one_iteration()
+    train_two_steps()
     torch.cuda.synchronize(device)
 
     with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA]) as prof:
-        train_one_iteration()
+        train_two_steps()
```
Autumn1998: @wujingyue  You’re right. The ZeRO-1 all-gather is also triggered in the first iteration because model_weight starts with the optimizer placement. I had a wrong assumption.

  However, with only one iteration and one microbatch, ZeRO-1 and ZeRO-2 have the same communication pattern. We need at least two microbatches to distinguish ZeRO-1, ZeRO-2, and ZeRO-3.

  I still suggest keeping the current two-step, two-microbatch setup. The two microbatches distinguish all three sharding strategies, while the second step also verifies the all-gather triggered after optimizer.step().

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -159,6 +165,15 @@ def __init__(
         else:
             self._symm_mem_pool = None
 
+        current_stream = torch.cuda.current_stream(allgather_stream.device)
+        allgather_stream.wait_stream(current_stream)
```
Autumn1998: Initialize the model weight on the all-gather stream; otherwise, its allocation stream defaults to the main stream.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
wujingyue: I realized I've also been using empty_cache unnecessarily in other tests. Created https://github.com/NVIDIA/Megatron-LM/pull/6551 to clean it up. PTAL!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue: Can you code-comment on the two waits? `allgather_stream.wait_stream` is needed because the cast needs to read from `main_weight`, allocated on the default stream. I'm not sure why `current_stream.wait_stream` is needed. 

Nit: if it helps, there's also the alternative to **only allocate** model_weight on `allgather_stream` and cast on the default using `cast(out=self.model_weight)`.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
wujingyue: Got it -- can you say that in the code comment? In general, code-comment why, not what. 

Also, I mentioned in the other comment that we could, several lines down, run the entire training step rather than just forward. That would make the test more realistic (people don't use FSDP to just run forward) and this artificial optimizer.step unnecessary.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +524,59 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+
+    (parameter_group,) = model.parameter_groups
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.01, foreach=False)
+    fully_shard_optimizer(optimizer)
+
+    # Initialize sharded Adam states.
+    model(x).sum().backward()
+    optimizer.step()
+    assert parameter_group.model_weight.placements == (Flat(),)
+    torch.cuda.synchronize(device)
```
wujingyue: Not sure if it's needed given #6551

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +524,59 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+    allocated_before_forward = torch.cuda.memory_allocated(device)
+    torch.cuda.reset_peak_memory_stats(device)
+
+    # Restore replicated model weight.
+    with torch.no_grad():
+        output = model(x)
```
wujingyue: Why don't we run backward and optimizer like https://github.com/NVIDIA/Megatron-LM/blob/e79110c67b6095326a98f210e8adc5afc905b07d/tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py#L1022?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -493,6 +524,59 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+    torch.cuda.synchronize(device)
```
wujingyue: ditto

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -150,6 +150,28 @@ def _flat_placements() -> Placements:
     return Placements(dp_axes=[0], parameter=[Flat()], gradient=[Flat()], optimizer=[Flat()])
 
 
+def _zero0_placements() -> Placements:
```
wujingyue: Nit: I haven't seen people say zero-0. Maybe just "no_shard" here.

wujingyue (commented):
LGTM otherwise

wujingyue (approved):

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: It seems that torch.cuda.memory_allocated() reads the allocator’s behavior on CPU and does not require device synchronization. The synchronization here was only defensive.  removed

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: This is similar to the previous approach: preallocate `model_weight` on the all-gather stream, then populate it from `main_weight`. However, when  their `dtypes` match, `DBuffer.cast() `returns `main_weight` directly, so `model_weight` and `main_weight` alias. Therefore, we need to allocate `main_weight` on the all-gather stream also

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: My goal here is to verify the exact memory deltas. Optimizer-state storage and forward memory growth are relatively straightforward to calculate independently. Calculating the exact peak of a complete forward-backward-optimizer step is more involved because it depends on the overlapping lifetimes of activations, autograd gradients, .... I will work through that calculation in a follow-up commit.

  `test_fully_shard_reduces_peak_training_memory` provides somewhat coarser limition: it only verifies each sharded strategy against an unsharded Sequential baseline . I have an idea for a controlled test that verifies the expected peak-memory ordering:

  no_shard > ZeRO-1 > ZeRO-2 > ZeRO-3

  I will experiment with that as well.

wujingyue: > test_fully_shard_reduces_peak_training_memory provides somewhat coarser limition: it only verifies each sharded strategy against an unsharded Sequential baseline

https://github.com/NVIDIA/Megatron-LM/pull/6551 hardens the test. PTAL!

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: The UT has been updated to run a complete forward–backward–optimizer step. It compares the observed peak memory against the theoretical peak, calculated as the maximum of the forward/backward peak and the optimizer-step peak, and asserts that the absolute difference remains below a small tolerance.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: commented and remove `current_stream.wait_stream`

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: fixed

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: removed

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
wujingyue: > However, when their dtypes match, DBuffer.cast() returns main_weight directly, so model_weight and main_weight alias. 

Yes

> Therefore, we need to allocate main_weight on the all-gather stream also

Do we have to? main_weight is allocated during `__init__` and persists. Even when `model_weight` is an alias of `main_weight`, `self.model_weight = ...` doesn't trigger deallocation because the underlying tensor is still held by `main_weight`. 

This looks fine to me. What's the failure pattern if we don't allocate main_weight on the all-gather stream? 

Bottom line: the contract we are trying to enforce is that any deletion of a tensor has to happen after all uses of the tensor are joined back to the allocation stream.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -654,6 +685,97 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+        optimizer_model_weight_growth
+        - optimizer_main_gradient_reduction
+        + optimizer_casted_gradient_size
+        + optimizer_state_size
+        + optimizer_temporary_size
+    )
+    theoretical_training_growth = max(theoretical_backward_growth, theoretical_optimizer_growth)
+
+    assert actual_optimizer_size == theoretical_optimizer_size
+    assert abs(actual_training_growth - theoretical_training_growth) < 1024**2
```
wujingyue: Awesome! I'll try to simplify this slightly after it's merged. Resting memory after initialization and peak memory after that can be tested separately, I think.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -755,6 +877,14 @@ def test_root_backward_returns_to_resting_memory(distributed_setup):
     )
 
 
+@pytest.mark.parametrize(
+    "strategy,placements_factory",
+    [
+        pytest.param("zero1", _zero1_placements, id="zero1"),
```
wujingyue: Why two "zero1"s?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
wujingyue: > the second step also verifies the all-gather triggered after optimizer.step().

I don't follow this. Before profiling, don't we already run a warmup that invokes `optimizer.step()`?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -1181,12 +1349,20 @@ def test_non_leaf_parameter_view_survives_storage_resize(distributed_setup):
...
+    # Ignoring common allocations, let W be the full BF16 model size, R be the
+    # world size, and T be one layer's temporary gradient buffers plus live
+    # activations. W covers all 16 layers, while T covers only the current layer's
+    # gradient buffers and small activations, so T < W in this test.
+    # 1. ZeRO-1 peak = max(W + T, 4W / R).
+    # 2. ZeRO-2 peak = max(W / R + T, 4W / R).
+    # At R = 2, the optimizer term dominates both, so their peaks may be equal;
+    # R >= 4 ensures that the ZeRO-1 peak is strictly greater than the ZeRO-2 peak.
+    if world_size < 4:
+        pytest.skip("This test requires at least 4 ranks for strict peak-memory ordering.")
```
wujingyue: I'd revert changes to this test. Given https://github.com/NVIDIA/Megatron-LM/pull/6551, we should probably replace this test with more accurate measurement, one for each strategy.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -814,50 +947,86 @@ def test_overlaps_communication_and_compute(distributed_setup, use_symmetric_mem
...
     allgather_kernels = collect_linked_kernels(prof, _ALL_GATHER_OP_NAME_SUBSTRING)
     reduce_scatter_kernels = collect_linked_kernels(prof, _REDUCE_SCATTER_OP_NAME_SUBSTRING)
-    # Each child layer does a forward and a backward all-gather and one
-    # reduce-scatter. Zero-CTA moves the all-gather to copy-engine memcpys, so it
-    # should not emit all-gather kernels.
-    expected_allgather_kernel_count = 0 if use_symmetric_memory else 2 * num_children
+    # ZeRO-1/2 all-gather once per optimizer step. ZeRO-1 reduces only the second
+    # microbatch, while ZeRO-2 reduces both. ZeRO-3 gathers for every forward and
+    # backward and reduces every microbatch.
+    expected_collective_counts = {
```
wujingyue: Nit: this is on the border of https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html. 

I'll probably un-parametrize it after this is merged.

wujingyue (approved):

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: change to just 1 step🤔

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: reverted

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
Autumn1998: fixed here

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: Previously, this was to ensure that the model weights would always have an AG stream as their allocation stream. It has now been changed.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/dbuffer.py
```diff
@@ -120,6 +120,13 @@ def device(self) -> torch.device:
         """Device of the local buffer."""
         return self.local_buffer.device
 
+    @property
+    def is_symmetric_memory(self) -> bool:
```
wujingyue, line 124: Really just a suggestion for the future: changes like this could land much faster if it's a separate PR.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -123,6 +125,10 @@ def __init__(
         model_weight_placements = tuple(placements.parameter)
         main_grad_placements = tuple(placements.gradient)
         main_weight_placements = tuple(placements.optimizer)
+        self._model_weight_placements = model_weight_placements
+        # main_grad rests here (DP-outer-Partial for HSDP) between microbatches and
+        # is finalized to main_weight's placements after the last microbatch.
+        self._accumulation_placements = main_grad_placements
```
wujingyue: Nit:
```suggestion
        self._main_grad_placements = main_grad_placements
```
so it's symmetric with `self._model_weight_placements`.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -158,6 +164,21 @@ def __init__(
             self._symm_mem_pool = symm_mem.get_mem_pool(self.main_weight.device)
         else:
             self._symm_mem_pool = None
+        if main_weight_dtype == self.dtype:
+            self.model_weight = self.main_weight
+        else:
+            # Allocate model_weight on the all-gather stream to record its allocation stream.
```
wujingyue: It could be me who added this comment but this is the wrong reason. 

I think the real reason is to enforce the contract that any deletion of `self.model_weight` (e.g. when self.model_weight is replaced with the allgathered version for zero-1 and zero-2) has to happen after its uses are joined back to the allocation stream. It's easier to enforce this when `self.model_weight` is allocated on the allgather stream instead of the default.

wujingyue (approved):
Looks great! 🚢

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
Autumn1998: fixed

jaredcasper (approved):

PR #6153, opened 2026-07-30
Fix FSDP activation recompute prefetch

shjwudp:
## What

- Track activation-recompute state per `FsdpModule` with an explicit
  `RESTING` / `FORWARD` / `BACKWARD` phase.
- Detect reentrant recomputation by checking whether forward executes inside an
  active autograd `GraphTask`.
- Disable forward-order all-gather prefetch during activation recomputation.
- Keep recomputed parameters materialized until their matching post-backward
  gradient reduction and reshard.
- Add regression coverage for both reentrant and non-reentrant activation
  checkpointing.

## Bug

Activation recomputation executes module forward hooks while autograd is already
running backward. Megatron FSDP v2 previously treated these recomputed forwards
as ordinary forward execution, so `FsdpModule.pre_forward()` continued using the
static forward order to prefetch the next module.

For two consecutive FSDP modules, the failure sequence is:

1. The later module is recomputed, completes backward, and is resharded.
2. Autograd then recomputes the earlier module.
3. The earlier module's forward prefetch unshards the later module again.
4. The later module's backward has already completed, so no later hook reshards
   it.
5. Backward finishes with the later module exposing unsharded parameters.

In the downstream optimizer path, this can leave stale gradients on empty local
DTensor shards. The next `zero_grad()` clears nonempty optimizer shards but
cannot reach those empty sharded parameter objects while the model exposes its
unsharded parameters. Gradient reduction then observes a mixture of set and
missing gradients and raises:

`FSDP sharded gradients must be either all set or all None.`

## Fix

Each FSDP module now owns its execution phase:

- An ordinary forward transitions `RESTING -> FORWARD -> RESTING`.
- The module's backward-pre hook sets `BACKWARD`, which identifies
  non-reentrant recomputation.
- Reentrant recomputation can start before the child backward-pre hook runs, so
  forward also checks `torch._C._current_graph_task_id()` to detect execution
  inside the active autograd `GraphTask`. This follows the PyTorch FSDP2 signal
  used in #6323.

For either recompute mode, forward-order prefetch and post-forward reshard are
suppressed. The recomputed parameters remain materialized for the immediately
following backward; `post_backward()` reduces gradients, reshards the module,
and returns its phase to `RESTING`.

## Regression test

The test activation-checkpoints two independently sharded linear layers and is
parameterized over `use_reentrant=False` and `use_reentrant=True`.

It verifies that:

- both layer weights are sharded `DTensor` objects after backward;
- the root and child modules return to `RESTING`;
- a second forward/backward resumes normal forward prefetch and again finishes
  fully resharded.

## Validation

- Post-merge 4-GPU non-profiler `test_fully_shard.py` suite in
  `mcore-ci-dev`: `25 passed, 3 deselected` on every rank.
- Focused 2-GPU activation-recompute regression: both `non_reentrant` and
  `reentrant` passed on every rank.
- Earlier downstream ND-parallel validation on this PR: `4 passed, 5 skipped`.
- Black 26.3.0, isort, Ruff, and `git diff --check` passed on the final changed
  files.

The three deselected cases are profiler parameterizations affected by the known
test-container incompatibility where `FunctionEvent.linked_correlation_id` is
unavailable; all non-profiler cases pass.


tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -189,6 +203,28 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
     )
 
 
+@pytest.mark.parametrize("use_reentrant", [True, False], ids=["reentrant", "non_reentrant"])
+def test_fully_shard_activation_recompute_reshards_parameters(distributed_setup, use_reentrant):
+    """Activation recomputation should leave every FSDP module resharded."""
+    world_size = distributed_setup.world_size
+    device = distributed_setup.device
+    if world_size < 2:
+        pytest.skip("This test requires at least 2 ranks.")
```
```suggestion
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -36,6 +36,9 @@ class FsdpContext:
     # unnecessary because it can be detected when ``model_weight``, after syncing
     # from ``main_weight``, has placements different from ``Placements.optimizer``.
     is_last_microbatch: bool
+    # True from the root pre-backward hook until autograd completes. Forward
+    # hooks use this to identify activation recomputation inside backward.
+    is_backward: bool
```
wujingyue: https://medium.com/@harshiljani2002/use-enums-over-booleans-to-clarify-code-intent-5656a282069f

e.g. forward, backward, and resting?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -243,9 +249,13 @@ def pre_forward(self) -> None:
         # issued afterwards, so it is free to run concurrently with this FsdpModule).
         current_stream.wait_event(self._unshard_event)
 
-        next_module = context.forward_order.next_item(self)
-        if next_module is not None:
-            next_module._unshard_parameter_groups()
+        # Activation recomputation runs forward hooks inside backward. Do not
+        # prefetch the next module in forward order: its backward may already
```
wujingyue, line 282: Is it just prefetching that we want to skip? Don't we want to skip https://github.com/NVIDIA/Megatron-LM/pull/6153/changes#diff-f21cfe22709b6d177ca822856ab5fd4461737aa5edd6b4c00a8551c678665849R246 as well?

wujingyue (commented):
Thanks for the PR!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
shjwudp: I believe only the forward prefetch needs to be skipped. Duplicate unsharding during both forward and backward passes is already prevented by the `_unshard_event` guard at [module.py#L268](https://github.com/shjwudp/Megatron-LM/blob/e639bb548d3be1282657fbad8e258779158278ba/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py#L268). (Also, the link you shared doesn't seem to open on my end.)

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
shjwudp: Thanks for the feedback. I’ve addressed this in commit 4e358d870.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -36,6 +52,10 @@ class FsdpContext:
     # unnecessary because it can be detected when ``model_weight``, after syncing
     # from ``main_weight``, has placements different from ``Placements.optimizer``.
     is_last_microbatch: bool
+    # BACKWARD from the root pre-backward hook until the autograd final callback
+    # runs. Forward hooks use this to identify activation recomputation inside
+    # backward.
+    phase: FsdpContextPhase
```
wujingyue: Should this be per-FsdpModule? I'm worried about cases like MoE 1F1B where there's no clear forward/backward boundary on the context level. The execution order is not always like `F0 F1 F2 ... Fn Bn Bn-1 ... B0`.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
wujingyue: That's right -- regardless of `_unshard_event` Prefetching has to be protected; otherwise we would be prefetching in the wrong order.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -203,6 +204,35 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
...
+    model = CheckpointedTinyModel(use_reentrant=use_reentrant).to(device)
+    fully_shard(model.fc1, mesh=mesh, placements=_flat_placements())
+    fully_shard(model.fc2, mesh=mesh, placements=_flat_placements())
+    fully_shard(model, mesh=mesh, placements=_flat_placements())
+
+    x = torch.randn(2, 8, device=device, requires_grad=True)
+    model(x).sum().backward()
+
+    assert isinstance(model.fc1.weight, DTensor)
+    assert isinstance(model.fc2.weight, DTensor)
```
wujingyue, line 261:
```suggestion
```
Not the responsibility of this test.

wujingyue (commented):
LGTM otherwise

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
shjwudp: In the MoE 1F1B scenario, a module actually processes multiple microbatches in "parallel". I understand this is more like a per-module-batch phase.

FsdpContextPhase can only handle regular forward-backward loop, for PP 1F1B or MoE 1F1B, I think we need external input to guide what exactly the schedule is. 🤔

wujingyue: Regardless of 1F1B, isn't it more straightforward and more robust to make this state per-FsdpModule? The only user of this state is FsdpModule and the pre/post forward/backward hooks are the natural places to update it.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
shjwudp: This assertion is a protective check. In the previous implementation, the prefetch of fc1 was not prevented. Therefore, in the sequence `fc2 backward -> fc1 recompute (also prefetch fc2) -> fc1 backward`, the result is that the `fc2.weight` is in  an unsharded Parameter instead of a sharded DTensor at the end of backward.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
shjwudp: Yeah, make sense, change phase owner to FsdpModule.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
wujingyue: Cool! Code-comment that so future readers know what these assertions try to capture.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -27,6 +28,14 @@
 from .placement import Placements
 
 
+class FsdpModulePhase(enum.Enum):
```
wujingyue: For access control, can we move this into `class FsdpModule` as an inner class? After that, you can simply name it `Phase`, not `FsdpModule`, because its parent class already defines the scope.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -228,6 +246,9 @@ def pre_forward(self) -> None:
         on the comm stream, so ``AG_{i+1}`` is launched before ``F_i`` finishes.
         """
         self._lazy_init_context()
+        is_recomputing = self._phase is FsdpModulePhase.BACKWARD
```
wujingyue: May want to
```
assert self._phase is not FORWARD
```
before this line

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -289,10 +319,16 @@ def _reshard_parameter_groups(self) -> None:
 
     def pre_backward(self) -> None:
         """Prepare full parameters and prefetch the next FsdpModule in backward order."""
+        self._phase = FsdpModulePhase.BACKWARD
         torch.cuda.nvtx.range_push(self._nvtx_label("backward"))
         context = self.context
         current_stream = context.current_stream()
         if self.is_root():
+            # Reentrant checkpointing recomputes a child before that child's own
+            # backward-pre hook runs, so mark the entire subtree from the root
```
wujingyue: Thanks for pointing this out -- I totally missed it. 

Can we remove this block and limit this PR to non-reentrant checkpointing for now? In non-reentrant mode, each module’s backward-pre hook runs before its recomputed forward, so the module-local phase is sufficient.

Reentrant checkpointing is harder because a child can be recomputed before its own backward-pre hook. The current root-wide phase update works around that, but complicates per-module phase ownership and interleaved schedules such as 1F1B.

I have some ideas for detecting reentrant execution using module.training together with torch.is_grad_enabled(), but they need more verification around ordinary no_grad() forwards and interleaved execution. I’d prefer to defer reentrant support rather than commit to the root-wide workaround here. Please keep the existing non-reentrant regression coverage and drop the reentrant case for now.

wujingyue (approved):
LGTM with comments

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -91,6 +105,9 @@ class FsdpModule:
     # ``None`` lets pre_forward enqueue an all-gather unless an earlier FsdpModule
     # already prefetched this module.
     _unshard_event: torch.cuda.Event | None
+    # The root backward-pre hook sets this to BACKWARD before activation
+    # recomputation can run. Module hooks otherwise own its transitions.
```
wujingyue: Needs an update

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
wujingyue: https://github.com/NVIDIA/Megatron-LM/pull/6309 is my current idea. PTAL even though we don't have to handle reentrant mode in this PR.

shjwudp: @wujingyue Thanks for your input! I now use precisely controlled per-module phases and have also incorporated the active-GraphTask check from [#6323](https://github.com/NVIDIA/Megatron-LM/pull/6323/changes) to detect reentrant recomputation that starts before the module backward-pre hook.

shjwudp:
Thanks @wujingyue for the detailed reviews and the reentrant checkpointing follow-up in #6323. The GraphTask insight was key to completing the fix cleanly. Really appreciate your help and collaboration! : )

wujingyue (approved):
> After the explicit FSDP construction-scope change in https://github.com/NVIDIA/Megatron-LM/pull/6190, this logic runs
after FsdpContext.ensure_finalized(), and the regression model is constructed inside fully_shard_context().

You may want to update the PR description. The above paragraph doesn't make sense to me any more after we move the phase tracking to module hooks.

Phlip79 (approved):

jaredcasper (approved):

PR #6155, opened 2026-07-30
[feat] NCCL Symmetric Memory Registration Support

### LINKED ISSUE #6540: [feat] GTP+Symmetric Memory Registration buffer

prajwal1210:
# What does this PR do?
Adds NCCL symmetric memory/user buffer registration support for GTP

## Summary

Symmetric memory registration allows for the use of faster LSA/GIN based and CE based collectives in NCCL to improve all-gather and reduce-scatter communication performance for GTP. To enable these new kernels, NCCL requirements state:

- All-gathers: Input and Output both need to backed by symmetric memory registration
- Reduce-scatter: Input needs to be backed by symmetric memory registration but output does not.

Based on these requirements, this MR enables symmetric memory backed tensors gated by two flags:
Flag | Scope
-- | --
--gtp-nccl-ub | dense GTP remat group
--egtp-nccl-ub | routed-expert (EGTP) group
These are independent of independent of `--use-nccl-ub` and off by default.

## Design

One `ncclMemAlloc`-backed `torch.cuda.MemPool` per GTP process group, registered once with
`register_mem_pool(pool, symm=True)`. PyTorch's ProcessGroupNCCL segment hook then auto-registers
every later allocation made under `gtp_mem_pool_ctx(group)`, so the context manager itself is
collective-free and capture-safe.

Every buffer either collective touches, where it comes from, and what puts it in the pool:

| Buffer | Where it comes from | What puts it in the pool |
|---|---|---|
| **AG send** — MXFP8 + `--fp8-param-gather` | the param's own TE quantized storage | pool context wrapping the TE constructor |
| **AG send** — BF16 + distributed optimizer | DDP `param_data` | `register_ddp_buffers_on_gtp_groups` |
| **AG send** — BF16, no distributed optimizer | the sliced shard's own storage | `_gtp_slice_one_param` |
| **AG recv** | `GTPWeightCache` ticket buffer | `_allocate_buffer(rs=False)` under the pool context |
| **RS send** | `RegisteredLifoPool`, padded full shape | `alloc()` under the pool context |
| **RS recv** | `GTPWeightCache` ticket buffer | nothing — `_graphed_alloc`, stays out of the window |

Screenshot 2026-07-30 at 10 09 39 AM

### Pre-checks

### DIFF
--- a/megatron/core/distributed/param_and_grad_buffer.py
+++ b/megatron/core/distributed/param_and_grad_buffer.py
@@ -1126,6 +1126,12 @@ class _ParamAndGradBuffer:
         self.extra_main_grads = []
         self.nccl_mem_pool = None
 
+        # Only the distributed optimizer has a param_data all-gather input to back with
+        # ncclMemAlloc; without it this would just move grad_data into the pool for no benefit.
+        _any_param_needs_nccl_mem = self.ddp_config.use_distributed_optimizer and any(
+            getattr(p, 'param_needs_nccl_mem', False) for p in self.params
+        )
+
         if self.nccl_ub:
             # If nccl_ub is True, use nccl_allocator to allocate memory for param_data/grad_data.
             nccl_allocator.init()
@@ -1145,6 +1151,14 @@ class _ParamAndGradBuffer:
             tmp_warmup_tensor = torch.zeros([1], device="cuda")
             torch.distributed.all_reduce(tmp_warmup_tensor, group=self.data_parallel_group)
             torch.distributed.barrier()
+        elif _any_param_needs_nccl_mem:
+            # Params requested ncclMemAlloc backing so this buffer can be window-registered
+            # on additional comm groups by the caller. Allocation-only: no DP group
+            # registration or warmup here — the caller handles those for its own group(s).
+            nccl_allocator.init()
+            pool = nccl_allocator.create_nccl_mem_pool(symmetric=True)
+            self.nccl_mem_pool = pool
+            mem_alloc_context = functools.partial(torch.cuda.use_mem_pool, pool)
         else:
             # If nccl_ub is False, mem_alloc_context is nullcontext.
             mem_alloc_context = nullcontext
--- a/megatron/core/distributed/param_and_grad_buffer.py
+++ b/megatron/core/distributed/param_and_grad_buffer.py
@@ -1126,12 +1126,6 @@ class _ParamAndGradBuffer:
         self.extra_main_grads = []
         self.nccl_mem_pool = None
 
-        # Only the distributed optimizer has a param_data all-gather input to back with
-        # ncclMemAlloc; without it this would just move grad_data into the pool for no benefit.
-        _any_param_needs_nccl_mem = self.ddp_config.use_distributed_optimizer and any(
-            getattr(p, 'param_needs_nccl_mem', False) for p in self.params
-        )
-
         if self.nccl_ub:
             # If nccl_ub is True, use nccl_allocator to allocate memory for param_data/grad_data.
             nccl_allocator.init()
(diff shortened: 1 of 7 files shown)


fanshiqing:
@prajwal1210 We also need to support SMR for GTP-bf16-a2a kernels supported in https://github.com/NVIDIA/Megatron-LM/pull/6200

megatron/core/tensor_parallel/gtp_symm.py
```diff
@@ -0,0 +1,270 @@
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""GTP symmetric-memory pools: the shared registration primitive.
```
fanshiqing, line 5: We need to add one new section in `generalized_tensor_parallelism.md` to introduce the key design and benefits of this feature.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -384,6 +391,13 @@ class GTPRematConfig:
     # DDP's 1/replicate scaling to yield the full (replicate x gtp) mean.
     calculate_per_token_loss: bool = False
 
+    # Back the AG output cache, RS send buffers and (under distopt) the DDP param buffer with
+    # ncclMemAlloc pools registered on the GTP / EGTP group. Independent of --use-nccl-ub (DP).
+    gtp_nccl_ub: bool = False
+    egtp_nccl_ub: bool = False
+    # Without distopt the slice-path shard keeps its own storage, so it goes to the symm pool.
+    use_distributed_optimizer: bool = False
```
fanshiqing, line 399: It's not a good idea to keep `use_distributed_optimizer` in `GTPRematConfig`. It is a orthogonal feature for GTP itself.

deepakn94:
> @prajwal1210 We also need to support SMR for GTP-bf16-a2a kernels supported in https://github.com/NVIDIA/Megatron-LM/pull/6200

For this, I believe we also need to register the output buffer as symmetric memory?

prajwal1210:
> > @prajwal1210 We also need to support SMR for GTP-bf16-a2a kernels supported in #6200
> 
> For this, I believe we also need to register the output buffer as symmetric memory?

This is using the current approach of calling nccl all_to_all and then doing the reduce ourselves? I don't know if nccl symmetric has a device-initiated all-to-all kernel. The symmetric RS kernels in NCCL only require inputs to be registered and not the output

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1151,6 +1145,16 @@ def __init__(
             tmp_warmup_tensor = torch.zeros([1], device="cuda")
             torch.distributed.all_reduce(tmp_warmup_tensor, group=self.data_parallel_group)
             torch.distributed.barrier()
+        elif self.ddp_config.use_distributed_optimizer and any(
+            getattr(p, 'param_needs_nccl_mem', False) for p in self.params
```
deepakn94, line 1149: Drop `param` from the attribute name? Is understood IMO.

megatron/core/distributed/param_and_grad_buffer.py
```diff
@@ -1151,6 +1145,16 @@ def __init__(
...
+        ):
+            # Params opted into ncclMemAlloc backing so the caller can window-register this
+            # buffer on other comm groups; it registers and warms them itself, so this branch
```
deepakn94, line 1152: Difference between this and previous if block?

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
deepakn94: Is this still relevant?

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1443,8 +1508,37 @@ def batched_all_gather_and_prefetch(self, **kwargs):
...
     def get_wgrad_tensor(self):
-        """Pool-allocate a wgrad scratch tensor of unsharded shape for the bwd GEMM."""
-        return _wgrad_pool_get(self._unsharded_shape, self.main_grad.dtype, self.device)
+        """Pool-allocate a wgrad scratch tensor of unsharded shape for the bwd GEMM.
+
+        Symm-RS path: return the [:unsharded] view of a persistent padded send buffer
+        from the LIFO pool. The bwd GEMM writes into the view; the RS sends the full
+        padded buffer directly (already window-registered), skipping F.pad.
+        """
+        if not getattr(self, "gtp_smr", False):
```
deepakn94, line 1517: Why do we need the getattr here?

megatron/core/distributed/param_and_grad_buffer.py
prajwal1210: This block just changes the allocation from regular torch allocator to using nccl Mem Alloc for the tensors. It does not register the pool on the DDP group - 
            mem_alloc_context = functools.partial(torch.cuda.use_mem_pool, pool)


The previous "if nccl-ub" block, does both - 
            mem_alloc_context = functools.partial(
                nccl_allocator.nccl_mem,
                pool,
                group=self.data_parallel_group,   --> Pool registered on the DDP group.
                symmetric=not self.ddp_config.disable_symmetric_registration,
            )

PR #6164, opened 2026-07-30
Fix tied parameters

### LINKED ISSUE #6139: Integrate NeMo Automodel Qwen3.5-35B-A3B recipe with MFSDP v2
## Summary

The NeMo Automodel Qwen3.5-35B-A3B VLM fine-tuning recipe currently uses PyTorch FSDP2 and needs a validated Megatron Core path using experimental MFSDP v2.

Reference recipe:
- https://github.com/NVIDIA-NeMo/Automodel/blob/main/examples/vlm_finetune/qwen3_5_moe/qwen3_5_35b.yaml

## Reproduction / starting point

The reference recipe runs with:

```bash
automodel examples/vlm_finetune/qwen3_5_moe/qwen3_5_35b.yaml --nproc-per-node 8
```

Use the same model and the closest equivalent data, parallelism, and checkpoint settings through the MFSDP v2 path.

wujingyue:
## What changed

- collect MFSDP v2 parameter registrations without deduplicating names, while representing each physical `Parameter` once
- retain every owning-module-relative FQN and use the complete tuple when swapping sharded and unsharded parameter replacements
- register one gradient hook per physical trainable parameter and track backward readiness with a resettable counter
- add a tied-embedding regression test covering two forward/backward cycles, balanced annotations, DBuffer size, and optimizer-visible parameter count

## Why

PyTorch's default parameter enumeration deduplicates tied parameters by object identity. MFSDP v2 instead collected registrations by FQN, so the same physical parameter could receive multiple buffer entries and replacement objects. That broke weight identity and prevented the ready-parameter set from reaching the duplicated trainable-parameter count during backward.

## Impact

MFSDP v2 now preserves tied parameter aliases while allocating, sharding, reducing, and exposing only one physical optimizer parameter.

## Validation

- `ruff check` on all changed Python files
- Black check on all changed Python files
- Python bytecode compilation on all changed Python files
- `python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/mfsdp_v2`
  - 53 passed
  - 11 skipped because they require 4-5 ranks

## Related issue

Related to #6139.

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
@@ -85,7 +85,7 @@ class FsdpModule:
     _name: str | None
     _parameter_groups: tuple[FsdpParameterGroup, ...]
     _context: FsdpContext | None
-    _ready_grad_parameters: set[nn.Parameter]
+    _num_ready_grad_parameters: int
     _num_trainable_parameters: int
     # Event recorded after this FsdpModule's full parameters are materialized.
     # ``None`` lets pre_forward enqueue an all-gather unless an earlier FsdpModule
@@ -119,7 +119,7 @@ class FsdpModule:
             for group_parameters in _group_parameters(owned_parameters)
         ]
         self._parameter_groups = tuple(parameter_groups)
-        self._ready_grad_parameters = set()
+        self._num_ready_grad_parameters = 0
         self._num_trainable_parameters = sum(
             len(group.sharded_parameters) for group in self._parameter_groups if group.requires_grad
         )
@@ -211,12 +211,12 @@ class FsdpModule:
             if not group.requires_grad:
                 continue
             for parameter in group.unsharded_parameters:
-                parameter.register_post_accumulate_grad_hook(self._make_grad_hook(parameter))
+                parameter.register_post_accumulate_grad_hook(self._make_grad_hook())
 
-    def _make_grad_hook(self, parameter: nn.Parameter) -> Callable[[nn.Parameter], None]:
+    def _make_grad_hook(self) -> Callable[[nn.Parameter], None]:
         def grad_hook(_parameter: nn.Parameter) -> None:
-            self._ready_grad_parameters.add(parameter)
-            if len(self._ready_grad_parameters) == self._num_trainable_parameters:
+            self._num_ready_grad_parameters += 1
+            if self._num_ready_grad_parameters == self._num_trainable_parameters:
                 self.post_backward()
 
         return grad_hook
@@ -229,7 +229,7 @@ class FsdpModule:
         """
         self._lazy_init_context()
         torch.cuda.nvtx.range_push(self._nvtx_label("forward"))
-        self._ready_grad_parameters.clear()
+        self._num_ready_grad_parameters = 0
         context = self.context
         allgather_stream = context.allgather_stream
         current_stream = context.current_stream()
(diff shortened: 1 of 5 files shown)


megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -361,7 +360,7 @@ def _collect_owned_parameters(root_module: nn.Module) -> dict[str, nn.Parameter]
     parameters: dict[str, nn.Parameter] = {}
 
     def visit(submodule: nn.Module, submodule_fqn: str) -> None:
-        direct_parameters = list(submodule.named_parameters(recurse=False))
+        direct_parameters = submodule.named_parameters(recurse=False, remove_duplicate=False)
```
shjwudp, line 363: It looks like tied parameters may be traversed multiple times. Would we create duplicate copies for those tied parameters? 🤔

wujingyue: Yes, this is intentional; otherwise, set_module_parameter will miss some references.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -85,7 +85,7 @@ class FsdpModule:
     _name: str | None
     _parameter_groups: tuple[FsdpParameterGroup, ...]
     _context: FsdpContext | None
-    _ready_grad_parameters: set[nn.Parameter]
```
wujingyue, line 88: I forgot why I used a set. A counter seems sufficient.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
shjwudp: Yeah, I feel like this would break the semantics of the tie parameter, changing it from a single parameter to multiple parameters that are updated separately. I understand that the tie parameter needs to accumulate gradients multiple times in one backward pass and perform grad reduce after the final accumulation.

wujingyue: I guess you are missing the fact that `remove_duplicate=False` returns a dict from potentially **multiple** strs to **the same Parameter object**. 

> changing it from a single parameter to multiple parameters that are updated separately

No, the code doesn't do that. 

> I understand that the tie parameter needs to accumulate gradients multiple times in one backward pass and perform grad reduce after the final accumulation.

Yes, the code does exactly that.

shjwudp: I see, but I think there's still a correctness issue here. When constructing the tied shared parameters, they should share the same underlying weight — however, looking at [parameter_group.py L166](https://github.com/NVIDIA/Megatron-LM/pull/6164/changes#diff-318602037e8dca86bd2f6aefe51cb128938ac95d1f01feff501dc9e83520a3b4L166), the sharded parameters for tied shared parameters are each derived from `main_weight.get_dtensor(index)`, which means they could end up backed by independent main_weight entries rather than sharing the same one.

wujingyue: I guess you missed 
```
        for index, (parameter, fqns) in enumerate(parameter_to_fqns.items()):
```
from https://github.com/NVIDIA/Megatron-LM/pull/6164/changes#diff-318602037e8dca86bd2f6aefe51cb128938ac95d1f01feff501dc9e83520a3b4R171

This loop never calls `get_dtensor(index)` twice for the same index. cc @shjwudp

shjwudp: Sorry for the miss. This mapping does work—looks great to me!

shjwudp (approved):
LGTM

jaredcasper (approved):

Phlip79 (approved):

PR #6167, opened 2026-07-31
fix: initialize dense local CUDA graph state

nvegesna-netizen:
## Problem

GraphableMegatronModule can create a local CUDA graph manager from its constructor. For an MLP-only graph scope, that path reads is_moe_layer before TransformerLayer initializes it, so dense layers fail during construction.

## Fix

Initialize the dense default before the base constructor while preserving the value already set by MoETransformerLayer. TransformerLayer still derives the final value from the constructed MLP later in initialization.

## Validation

- Added a focused unit regression that constructs a dense TransformerBlock with local MLP CUDA graph scope and verifies every layer receives a graph manager.


nvegesna-netizen:
@NVIDIA/mcore-oncall please review.

mathemakitten (approved):

yaoyu-33 (approved):

jaredcasper (approved):

PR #6168, opened 2026-07-31
fix: make local CUDA graph outputs pipeline-safe

nvegesna-netizen:
## Problem

PyTorch custom autograd Functions expose returned tensors as views. Local CUDA graph record and replay nodes can therefore expose a view as the last layer output, while pipeline output pseudo-deallocation requires a viewless tensor.

## Fix

Convert only last-layer record and replay outputs to viewless tensors when pipeline output deallocation is enabled. The conversion aliases the existing storage without an activation-sized copy and preserves the autograd graph.

## Validation

- Added a focused distributed CUDA regression covering record and replay output deallocation, custom backward propagation, and finite input gradients.


nvegesna-netizen:
@NVIDIA/mcore-oncall please review.

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1652,6 +1653,21 @@ def apply_cudagraph_record_metadata(self, args, kwargs, outputs):
         for t in self.get_tensors(outputs):
             _apply_cudagraph_buffer_metadata(t, is_output=True)
 
+    def _make_pipeline_output_viewless(self, output):
+        """Make a last-stage record or replay output safe for pipeline deallocation."""
+
+        if not (self.is_last_layer and self.deallocate_pipeline_outputs):
```
jiemingz, line 1661: lets also check  if `self.config.pipline_parallell_size > 1 ` here?

nvegesna-netizen: Thanks. I checked this against the multi-module cross-grid schedule and kept the existing deallocation guard. A module could have local PP size 1 while its output was still sent across a module boundary and pseudo-deallocated, so gating on `pipeline_model_parallel_size > 1` would have reintroduced the view assertion. I added a code comment documenting this.

jiemingz (approved):

ericharper (approved):

PR #6173, opened 2026-07-31
[Dev] Account for Qwen3.5 vision encoder FLOPs

BestJuly:
## Summary

- add Qwen3.5-VL vision encoder work to Megatron's model FLOPs and throughput metrics
- account for Conv3d patch projection, dense vision attention and MLP projections, bidirectional attention, and patch-merger projections
- aggregate actual `image_grid_thw` statistics across microbatches and distributed ranks so variable-resolution, multi-image, text-only and packed-sequence inputs use their real patch and attention sizes
- preserve decoder-only behavior for non-multimodal models
- validate the vision metadata once at model-construction time, so a misconfiguration fails before the dataloader is built instead of at iteration 1
- document the accounting contract and add exact-formula, validation, rerun, and distributed accumulator tests

## Root cause

The Qwen3.5-VL model factory already registered the vision architecture metadata, but the core `num_floating_point_operations` path did not consume it. Reported TFLOP/s therefore included only the language decoder.

## Review changes since the last push

Rebased onto current `dev` (`0cd11658f`), which includes #6315 (HybridModel migration). The three blockers from @FDecaYed's review are addressed:

**1. Rank-local flag gating a world `all_reduce`** ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6173#discussion_r3774097634))

Collective participation is now a global invariant: whether `consume_vision_model_flops_stats` enters the all-reduce is decided by `args.count_vision_model_flops` alone — a config value identical on every rank — never by whether this rank happened to call `update_*`. Ranks with nothing to report contribute a zero tensor, and the "did any rank report runtime data" flag is reduced inside that same collective (it rides in the stats tensor, so the protocol is still one all-reduce and one host sync per iteration).

Covered by `TestVisionFlopsAccumulatorDistributed`, including `test_differing_activity_across_ranks_still_deduplicates` (activity differs across ranks with TP>1; consumption completes and the dedup is still applied) and `test_invalid_grid_on_one_rank_raises_on_all_ranks`. The `TP*CP*PP` dedup, previously only asserted in a docstring, is now pinned by `TestVisionFlopsAccumulatorTopology` over the existing `_TOPOLOGY_8GPU_PARAMS` matrix.

**2. Nominal fallback conflating temporal patch depth with the runtime grid count** ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6173#discussion_r3774097525))

Took the "omit" option. The fallback is deleted outright — no `temporal_grid_size = args.vision_temporal_patch_size`, no `args.image_size` read from core code, no one-image-per-sequence assumption. The vision terms are driven exclusively by statistics a forward step actually reported; when none were reported the vision contribution is excluded and a one-time warning names the hook the entry point failed to call. `examples/multimodal_dev/arguments.py` is untouched again.

Tests cover a temporal extent different from `vision_temporal_patch_size`, multi-image microbatches, text-only microbatches, and the omit path.

**3. Rerun replays double-counting** ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6173#discussion_r3774097741))

The hook is now `record_vision_model_flops(args, model, batch)`, which skips when the rerun state machine is in `RERUNNING_IN_PLACE` — the only state that replays in-process; the checkpoint-restart states replay in a fresh process with an empty accumulator. `examples/multimodal_dev/tests/test_vision_flops_accounting.py::test_in_place_rerun_does_not_double_count` forces one replay and asserts the consumed statistics equal a single logical pass.

Also addressed from the earlier `/claude strict-review`: metadata validation and derived-scalar precomputation hoisted out of the per-iteration path into `validate_vision_flops_metadata`; grid value validation made device-agnostic (malformed rows accumulate into a reduced counter instead of being checked only on CPU tensors, so production CUDA grids get the same strictness at no extra host sync); merged-token count uses floor division to match the merger's `view(-1, merge_dim)`; the exact-formula test anchored to a hand-computed integer instead of transcribing the implementation; variant renamed `qwen35_vl_v2` → `qwen35_vl`; the `README.md` registry template no longer names a variant that would apply Qwen3.5-VL's formula to a different architecture.

## Test plan

Run in the CI container image on 1x and 8x H100:

- `pytest tests/unit_tests/test_num_floating_point_operations.py` — 75 passed, 23 skipped (skips are the >=2-rank and 8-rank cases)
- `torchrun --nproc_per_node=8 -m pytest tests/unit_tests/test_num_floating_point_operations.py -k 'Distributed or Topology'` — 23 passed, covering all 8 `(TP, CP, PP)` topologies for both accumulators
- `pytest examples/multimodal_dev/tests/test_vision_flops_accounting.py` — 7 passed
- `black==26.3.0`, `isort`, `ruff` clean on every line this PR adds

(The previous revision's test plan noted pytest had never executed, blocked on container registry credentials. That is resolved; the numbers above are from real runs.)

## Sequencing

@FDecaYed asked to land this after #6315 and #6081. #6315 is merged and is in this base; #6081 is still open. Happy to rebase again onto the then-current `dev` once it lands.

### DIFF
--- a/examples/multimodal_dev/forward_step.py
+++ b/examples/multimodal_dev/forward_step.py
@@ -18,6 +18,7 @@ from megatron.core.parallel_state import (
     get_tensor_model_parallel_src_rank,
 )
 from megatron.training import get_args
+from megatron.training.training import update_vision_model_flops_stats
 
 # -------------------------------------------------------------------
 # dtype <-> int mapping for cross-rank broadcast
@@ -400,6 +401,13 @@ def forward_step(data_iterator, model):
     if batch is None:
         return None, None
 
+    args = get_args()
+    image_grid_thw = batch.get("image_grid_thw", None)
+    if getattr(model, "training", True) and getattr(args, "count_vision_model_flops", False):
+        update_vision_model_flops_stats(
+            image_grid_thw, spatial_merge_size=args.vision_spatial_merge_size
+        )
+
     pixel_values = batch.get("pixel_values", None)
     if (
         pixel_values is not None
@@ -417,7 +425,7 @@ def forward_step(data_iterator, model):
         loss_mask=batch.get("loss_mask", None),
         padding_mask=batch.get("padding_mask", None),
         pixel_values=pixel_values,
-        image_grid_thw=batch.get("image_grid_thw", None),
+        image_grid_thw=image_grid_thw,
         packed_seq_params=batch.get("packed_seq_params", None),
     )
 
--- a/examples/multimodal_dev/forward_step.py
+++ b/examples/multimodal_dev/forward_step.py
@@ -17,7 +17,9 @@ from megatron.core.parallel_state import (
     get_tensor_model_parallel_rank,
     get_tensor_model_parallel_src_rank,
 )
+from megatron.core.rerun_state_machine import RerunState, get_rerun_state_machine
 from megatron.training import get_args
+from megatron.training.training import update_vision_model_flops_stats
 
 # -------------------------------------------------------------------
 # dtype <-> int mapping for cross-rank broadcast
@@ -393,9 +395,49 @@ def loss_func(loss_mask, output_tensor):
 # -------------------------------------------------------------------
 
 
+def record_vision_model_flops(args, model, batch):
+    """Report this microbatch's vision grids to the FLOPs accumulator.
+
+    Called from :func:`forward_step` BEFORE its ``batch is None``
+    early-return, so a rank whose data iterator ran dry still reports an
+    explicit zero (``grid_thw=None``) instead of silently dropping its share
+    of the global batch. Entering the collective in
+    ``consume_vision_model_flops_stats`` is keyed off
+    ``args.count_vision_model_flops`` alone, so a missed report here cannot
+    desync the job -- it would only skew the reported number.
+
+    Two executions of a microbatch must not both be counted:
+
+    * ``model.training`` excludes ``evaluate()``, which calls
+      ``model.eval()``. ``evaluate()`` reuses this same forward step but never
+      drains the accumulator, so anything it added would leak into the next
+      ``train()`` iteration.
+    * ``RerunState.RERUNNING_IN_PLACE`` excludes the rerun state machine's
+      in-place replay of the SAME microbatches (``--check-for-nan-in-loss`` /
+      ``--check-for-spiky-loss``), which re-executes with the model still in
+      train mode. The accepted execution is the initial run, whose
+      contribution is already in the accumulator; the replay must add nothing.
+      The other rerun states replay in a freshly restarted process, whose
+      accumulator starts empty.
+    """
+    if not getattr(args, "count_vision_model_flops", False):
+        return
+    if not getattr(model, "training", True):
+        return
+    if get_rerun_state_machine().state == RerunState.RERUNNING_IN_PLACE:
+        return
+    update_vision_model_flops_stats(
+        batch.get("image_grid_thw", None) if batch is not None else None,
+        spatial_merge_size=getattr(args, "vision_spatial_merge_size", None),
+    )
+
+
 def forward_step(data_iterator, model):
     """Forward step for multimodal_dev training."""
     batch = get_batch(data_iterator)
+
+    args = get_args()
+    record_vision_model_flops(args, model, batch)
 
     if batch is None:
         return None, None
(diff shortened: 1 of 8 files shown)


FDecaYed:
With #6083 now in `dev`, I suggest landing this after #6315 and #6081 so the FLOPs accounting is rebased onto the final HybridModel and PP-aware forward path.

Before the new exact-head CI, could you address the current correctness blockers: the rank-local active flag conditionally entering a world `all_reduce` ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6173#discussion_r3774097634)), the nominal fallback conflating temporal patch depth with the runtime temporal-grid count ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6173#discussion_r3774097525)), and duplicate accumulator updates when the rerun state machine replays a microbatch ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6173#discussion_r3774097741))? After #6315/#6081 land, please rebase onto the then-current `dev` and ping us for exact-head review/CI.

megatron/training/training.py
```diff
@@ -384,6 +396,101 @@ def consume_seqlen_stats_in_iteration() -> Tuple[Optional[float], Optional[float
...
+def consume_vision_model_flops_stats() -> Tuple[Optional[float], Optional[float], Optional[float]]:
+    """Read, reset, and globally reduce per-iteration vision shape stats.
+
+    Returns:
+        ``(total_patches, attention_seqlen_squared_sum, merged_tokens)`` for
+        the global batch. Returns ``(None, None, None)`` when no multimodal
+        forward step supplied a vision grid in this iteration.
+    """
+    global _vision_flops_stats_active, _vision_flops_stats_in_iteration
+    if not _vision_flops_stats_active:
```
FDecaYed, line 475: `_vision_flops_stats_active` is rank-local state, but it conditionally determines whether the rank enters a world `all_reduce` below. A rank can miss the update—for example through a stage-specific/no-batch path—while another rank records a vision grid; the inactive rank then returns here while its peers enter the collective, causing a distributed hang. Please make collective participation a global invariant whenever vision-FLOPs accounting is enabled: inactive ranks should contribute a zero tensor, and an active/has-runtime-data flag should be reduced as part of the same synchronized protocol before deciding whether to return `None`. Please add a distributed test in which activity differs across ranks and verify that consumption completes and produces the correctly deduplicated result.

megatron/training/training.py
```diff
@@ -494,8 +601,175 @@ def _dsv4_hybrid_self_attention_flops(
...
+                f"got image_size={image_size}, vision_patch_size={patch_size}"
+            )
+        grid_size = image_size // patch_size
+        if grid_size % spatial_merge_size != 0:
+            raise ValueError(
+                "vision patch grid must be divisible by vision_spatial_merge_size, "
+                f"got grid_size={grid_size}, "
+                f"spatial_merge_size={spatial_merge_size}"
+            )
+        temporal_grid_size = args.vision_temporal_patch_size
```
FDecaYed, line 712: `vision_temporal_patch_size` describes the temporal patch kernel/depth, not the resulting temporal grid count `T`. Using it directly as `temporal_grid_size` conflates patch size with the number of post-patch temporal positions (which depends on the input temporal extent). This fallback also derives spatial shape from the mock-data `image_size` knob and assumes one vision input per sample, so it gives incorrect FLOPs for videos, packed or multi-image samples, and text-only batches. Please avoid fabricating runtime grid statistics from these arguments: either require explicit nominal grid/count metadata with unambiguous semantics, or omit vision FLOPs when runtime `grid_thw` data is unavailable. Tests should include a temporal extent different from the temporal patch size as well as multi-image/text-only cases.

examples/multimodal_dev/forward_step.py
```diff
@@ -400,6 +401,13 @@ def forward_step(data_iterator, model):
     if batch is None:
         return None, None
 
+    args = get_args()
+    image_grid_thw = batch.get("image_grid_thw", None)
+    if getattr(model, "training", True) and getattr(args, "count_vision_model_flops", False):
+        update_vision_model_flops_stats(
```
FDecaYed, line 407: This updates process-global counters on every invocation of `forward_step`, but the counters are consumed only once by the outer training iteration. Megatron rerun/replay paths can invoke the forward step more than once for the same logical microbatch; there is no microbatch identity, rollback, or idempotence guard here, so a rerun double-counts vision patches while the logical batch/iteration accounting remains unchanged. Please integrate this with the rerun state so only the accepted execution contributes (or reset/replace the contribution when a microbatch is replayed), and add a regression test that forces one rerun and verifies the consumed vision statistics equal a single logical pass.

FDecaYed (approved):
Re-reviewed the current head. The collective participation, runtime-grid accounting, and replay-safe accumulation fixes address my findings. The corresponding distributed/topology and rerun regression tests passed in exact-head CI. LGTM.

PR #6174, opened 2026-07-31
[Dev] Fix FSDP forward prefetch in combined 1F1B

lhb8125:
## Summary

- restore the normal FSDP forward state before a genuine forward that combined 1F1B schedules inside another microbatch's backward window
- scope the state reset to the forward layer, leaving the independently scheduled backward layer in `PRE_BACKWARD`
- preserve `PRE_BACKWARD` when forward and backward refer to the same underlying midpoint layer
- add a two-microbatch regression test that checks state isolation, numerical parity, and final parameter parity

## Root cause

The Megatron-FSDP root pre-backward hook marks every submodule `PRE_BACKWARD`. This is correct for activation-recomputation forwards, where forward-order parameter prefetch and eager resharding must be disabled. Combined 1F1B also runs a genuine forward for the next microbatch during this window, so that forward incorrectly inherited `PRE_BACKWARD`; `_pre_forward_param_unshard` consequently disabled forward-order prefetch before the schedule midpoint.

This change clears the marker only for the layer about to execute the genuine forward. The peer backward layer keeps its recomputation semantics.

## Test plan

- `CHECK_ONLY=true BASE_REF=dev bash tools/autoformat.sh` (Black, isort, Pylint, and Ruff passed)
- `python tools/check_copyright.py <changed files>` (5/5 passed)
- Lyris GB200, 4 distributed ranks, SLURM `2544709`:
  `python -m torch.distributed.run --nproc-per-node=4 -m pytest -q tests/unit_tests/a2a_overlap/test_fsdp_1f1b_overlap.py::TestFSDP1F1BOverlap::test_fsdp_1f1b_genuine_forward_state`
  (passed on all ranks)
- DeepSeek-V3 full-model validation on 256 GB200 GPUs, TP1/PP1/EP64 and Megatron-FSDP DOI=2:
  full-iteration CUDA graph SLURM `2542009` and eager/Nsys SLURM `2542527` both completed 11 iterations; the eager trace confirmed forward prefetch was restored before the midpoint

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py
@@ -501,6 +501,21 @@ class MegatronFSDP(torch.nn.Module):
             # attribute.
             param.overwrite_main_grad = True
 
+    @torch.compiler.disable
+    def prepare_forward_module(self, module: nn.Module) -> None:
+        """Prepare an FSDP unit for a genuine forward scheduled during backward.
+
+        The root pre-backward hook marks every submodule ``PRE_BACKWARD`` so
+        activation-recomputation forwards do not issue forward-order prefetches
+        or eagerly reshard their parameters. The EP-overlap 1F1B schedule also
+        executes a genuine forward for another microbatch inside that backward
+        window. Clear the marker only for that forward layer; recomputation in
+        the independently scheduled backward layer remains ``PRE_BACKWARD``.
+        """
+        for submodule in module.modules():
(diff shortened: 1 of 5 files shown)


wujingyue:
> @lhb8125: 我在 fsdp v1 中发现了和 1f1b a2a overlap 一起打开时候的 overlap issue，提了一个 pr，我不确定这个问题在 v2 里有没有

I don't think v2 has this problem: https://github.com/NVIDIA/Megatron-LM/blob/fb24e8707c94aebe28fb919e4a6253b600ad0528/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py#L342

PR #3193, opened 2026-02-02
Add absorbed-mla

kunlunl:
# What does this PR do ?

PR for main https://github.com/NVIDIA/Megatron-LM/pull/3198

Implement MLA with matrix absorption:
- Absorbs K's up projection into Q: Q' = Q @ K_up_proj^T
- Applies V's up projection after core attention
- Core attention operates in MQA form with KV being single-head.

The absorption is mathematically equivalent to standard MLA but enables MQA-style attention which
can be more efficient for certain attention variants.

---

## 1. TL;DR

- **What**: This PR adds a new `AbsorbedMLASelfAttention` class that implements MLA with matrix absorption optimization, where K's up-projection is absorbed into Q before core attention, enabling MQA-style computation.
- **Why**: Matrix absorption allows MLA to operate in a more memory-efficient MQA form where KV becomes single-head, enabling more efficient implementation for certain attention variants.

---

## 2. Big Picture

```mermaid
graph TB
    subgraph "Before: Standard MLA"
        A1[hidden_states] --> B1[Q down proj]
        A1 --> C1[KV down proj]
        B1 --> D1[Q up proj]
        C1 --> E1[KV up proj<br/>joint K+V]
        D1 --> F1[RoPE on Q]
        E1 --> G1[RoPE on K]
        E1 --> H1[V extraction]
        F1 --> I1[Core Attention<br/>MHA: n heads for Q,K,V]
        G1 --> I1
        H1 --> I1
        I1 --> J1[linear_proj]
    end
    
    subgraph "After: Absorbed MLA"
        A2[hidden_states] --> B2[Q down proj]
        A2 --> C2[KV down proj]
        B2 --> D2[Q up proj]
        C2 --> E2[K up weight<br/>NOT applied to KV]
        D2 --> F2["Q absorbed = Q @ K_up^T"]
        E2 --> F2
        F2 --> G2[RoPE on Q_absorbed]
        C2 --> H2[Compressed KV<br/>single head]
        H2 --> I2[RoPE on K_pos]
        G2 --> J2["Core Attention<br/>MQA: n heads Q, 1 head KV"]
        I2 --> J2
        J2 --> K2["V up proj AFTER attention"]
        K2 --> L2[linear_proj]
    end
    
    style F2 fill:#90EE90
    style K2 fill:#90EE90
    style J2 fill:#FFD700
```

---

### 3.1 Problem Background

Standard MLA (Multi-Latent Attention) as implemented in `MLASelfAttention` works as follows:

```python
[...]

PR #6178, opened 2026-07-31
[Dev] Allow FP8/FP4 with AbsorbedMLA up-projection recompute

niyunsheng:
## Summary

`AbsorbedMLASelfAttention` asserted that quantization was unsupported whenever `"mla_up_proj"` selective recompute was enabled, blocking FP8 and FP4 training for any model that needs that recompute to fit in memory. The assertion arrived with the initial absorbed-MLA implementation (#3193 / #3198) as a not-yet-validated placeholder rather than a guard for a known failure.

The quantized replay is safe for the same reason it is in `MLASelfAttention`, which runs the identical `CheckpointWithoutOutput` pattern with no such assertion (`multi_latent_attention.py`):

- The checkpoint records the forward recipe and amax state (`ctx.fp8` /
  `ctx.fp8_recipe` under `activation_recompute_forward(recompute_phase=False)`) and replays under the recorded `fp8_autocast` with `recompute_phase=True`.
- The only quantized operation inside `qkv_up_proj_and_rope_apply` is the Q up projection — the same module set MLA already replays.
- The absorption einsum reads the K up-projection weight directly rather than through the checkpoint context. That weight is a persistent parameter: under
  `fp8_param_gather` it is remapped to TE quantized storage that outlives the backward pass (`param_and_grad_buffer.py`), and the only path that frees param storage is explicit weight CPU offload. So the forward and the replay observe the same values.

`DeepSeekV4HybridAttention` already runs the same pattern without the assertion.

## Test plan

Added `test_fp8_up_proj_recompute_parity`, which drives two identically-initialized modules — one recomputing the up projection, one not — over the same inputs under `get_fp8_context`, and requires **bitwise-identical** (`atol=0, rtol=0`) outputs and per-parameter gradients. Recompute is meant to be an exact replay, so any drift would indicate a different quantization scale or a stale weight; a tolerance-based comparison would hide exactly the failures this guards against.
A bf16 reference run guards the parity check against passing vacuously: the same weights go through an `fp8=False` module, and the FP8 output must track it closely and differ from it bitwise — which is only possible if quantization actually engaged.

Coverage: `tensorwise` / `mxfp8` / `blockwise` recipes x `sbhd` / `thd` layouts x combined / split K-V up-projection module specs.

Run on GB200 (mxfp8 needs Blackwell, so nothing is capability-skipped there):

```
pytest tests/unit_tests/transformer/experimental_attention_variant/test_absorbed_mla.py
====================== 76 passed, 558 warnings in 52.58s =======================
```

76 = 64 pre-existing `test_functionality` cases (unchanged, no BF16 regression) + 12 new FP8 cases, 0 skipped.


tests/unit_tests/transformer/experimental_attention_variant/test_absorbed_mla.py
```diff
@@ -432,3 +439,123 @@ def _calculate_tensor_similarity(x, y):
...
 
     Utils.destroy_model_parallel()
+
+
+# Hopper = SM 9.0, Blackwell = SM 10.0+. mxfp8 needs Blackwell.
+_IS_BLACKWELL = torch.cuda.is_available() and torch.cuda.get_device_capability()[0] >= 10
+
+_QUANT_RECIPES = [
+    pytest.param("tensorwise", id="fp8-tensorwise"),
+    pytest.param("delayed", id="fp8-delayed"),
```
kunlunl: Better to test blockwise scaling rather than delayed. Delayed scaling is almost not used in anywhere.

tests/unit_tests/transformer/experimental_attention_variant/test_absorbed_mla.py
```diff
@@ -432,3 +439,123 @@ def _calculate_tensor_similarity(x, y):
...
+    if state_dict is not None:
+        model.load_state_dict(state_dict)
+    return model
+
+
+@pytest.mark.skipif(not fp8_available, reason=reason_for_no_fp8)
+@pytest.mark.parametrize("fp8_recipe", _QUANT_RECIPES)
+@pytest.mark.parametrize("qkv_format", ['sbhd', 'thd'])
+@pytest.mark.parametrize("combined_kv_up_projection", [True, False])
+def test_fp8_up_proj_recompute_parity(
```
kunlunl: This test only tests fp8 recompute against fp8 no-recompute. Better to test fp8 recompute vs bf16 baseline as other tests.

niyunsheng:
> **Non-blocking suggestion (test coverage):** The code path also enables FP4 (`self.config.fp8 or self.config.fp4`) and the PR title says "FP8/FP4," but the test only exercises FP8 recipes; the test's guarantee is effectively FP8-scoped. FP4 shares the exact same code path and requires TE ≥2.7 + capable hardware, so deferring it is reasonable — worth a note in the test docstring or a follow-up.

Good catch — I will add `nvfp4` coverage since I have Blackwell available.

kunlunl (approved):

FDecaYed (approved):
LGTM

kunlunl:
/ok to test [ece6447](https://github.com/NVIDIA/MegatronLM/pull/6178/commits/ece6447f91f7548cd9d313eee97eed3d7c2f0c58)

niyunsheng:
The failing bucket `tests/unit_tests/transformer/**/*.py` - latest is unrelated to this PR: the only failure is [test_te_layers_batch_invariant.py::test_te_column_parallel_linear_batch_invariant_randomized](https://github.com/NVIDIA/Megatron-LM/actions/runs/30885006091/job/91934222450?pr=6178#step:4:11874) with RuntimeError. But this PR only touches `absorbed_mla.py` and its test file.

Could someone re-run the failed job (or re-trigger with `/ok to test`)? Thanks! @buptzyb @kunlunl @FDecaYed

buptzyb:
It failed again. Are you sure it's not related to your changes?

niyunsheng:
> It failed again. Are you sure it's not related to your changes?

@buptzyb CI is green now. Root cause was an import side effect: the test imported helpers from `tests.unit_tests.determinism.utils`, which transitively imports `torch.testing._internal.common_utils` and freezes torch backend global flags (`disable_global_flags()`), deterministically breaking `test_te_layers_batch_invariant.py` later in the same session. Fixed in `6648492` by inlining the two RNG helpers so `torch.testing._internal` is never loaded in this bucket.

PR #6186, opened 2026-08-01
Document MFSDP optimizer design

### LINKED ISSUE #5614: Integrate MFSDP v2 with Megatron Core training

wujingyue:
## Context

This is the design document that backs [#5865](https://github.com/NVIDIA/Megatron-LM/pull/5865).

## Summary

- Document optimizer ownership across standalone MFSDP, MFSDP v1, and MFSDP v2.
- Explain why FSDP needs optimizer-specific handling and compare hooks, subclassing, and composition.
- Recommend hooks for elementwise optimizers and composition for matrix optimizers.


Autumn1998 (approved):

yaoyu-33 (approved):

megatron/core/distributed/fsdp/src/docs/optimizer.md
jaredcasper, line 1: copyright header on this file?

jaredcasper (approved):

PR #6190, opened 2026-08-01
Replace lazy MFSDP context initialization with an explicit scope

### LINKED ISSUE #5616: MFSDP v2 should support HSDP and HFSDP placements

wujingyue:

# What does this PR do?

## Why

MFSDP v2 has lazily initialized its runtime context on the first forward to stay close to PyTorch FSDP2. That defers stream creation, root discovery, module naming, and prefetch-order construction until runtime, adding lifecycle checks and coupling context setup to forward hooks.

[#6187](https://github.com/NVIDIA/Megatron-LM/pull/6187) is a concrete example of the resulting complexity. Because the reduce-scatter stream is unavailable when `fully_shard()` constructs parameter groups, `main_grad` must be materialized lazily so its allocation occurs on the stream that later releases it. With an eagerly initialized context, stream-dependent state can instead be constructed against the intended stream up front.

## What changes

This PR introduces an explicit construction scope:

```python
with fully_shard_context(device=device) as context:
    fully_shard(module_a, mesh=mesh, placements=placements)
    fully_shard(module_b, mesh=mesh, placements=placements)
```

Every `fully_shard()` call must run inside `fully_shard_context()`. Calling it without an active context raises a hard error. Nested construction scopes are also rejected.

The context eagerly creates its communication streams. Enclosed FSDP modules naturally inherit that context and register with it. When the scope exits, the context discovers roots, assigns module names, and builds the static forward and backward prefetch orders. Forward execution before scope exit also raises a hard error.

`microbatch()` now takes the context directly, so last-microbatch state is context-scoped rather than discovered by walking from a root module.

## Other benefits

1. All FSDP modules constructed under the same `fully_shard_context()` participate in one cross-root overlap schedule. Users no longer need an artificial `fully_shard(root_model)` call merely to make independently sharded children share streams and overlap communication.
2. The explicit context provides a natural home for context-wide configuration and state, including flags such as symmetric-memory enablement.

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
@@ -17,13 +17,44 @@
 import dataclasses
 from collections.abc import Iterator
 from contextlib import contextmanager
+from contextvars import ContextVar
 
+import torch
 from torch import nn
 from torch.distributed import DeviceMesh
 
 from ..mixed_precision import MixedPrecisionPolicy
 from .module import FsdpContext, FsdpModule
 from .placement import MeshAxis, Placements
+
+_FSDP_CONTEXT = ContextVar[FsdpContext | None]("megatron_fsdp_context", default=None)
+
+
+@contextmanager
+def fully_shard_context(device: torch.device | None = None) -> Iterator[FsdpContext]:
+    """Construct FSDP modules that share runtime streams and prefetch orders.
+
+    Independent roots are ordered by their root-level ``fully_shard`` calls.
+    Construction must finish before any of the registered modules run forward.
+
+    Args:
+        device: CUDA device on which to create communication streams. Defaults to
+            the current CUDA device.
+    """
+    device = device or torch.device("cuda", torch.cuda.current_device())
+    if device.type != "cuda":
+        raise ValueError(f"fully_shard_context requires a CUDA device, got {device}.")
+
+    context = FsdpContext(device=device)
+    token = _FSDP_CONTEXT.set(context)
+    try:
+        yield context
+    except Exception:
+        raise
+    else:
+        context.finalize()
+    finally:
+        _FSDP_CONTEXT.reset(token)
 
 
 def fully_shard(
@@ -49,6 +80,9 @@ def fully_shard(
     """
     if isinstance(module, FsdpModule):
         raise ValueError("This module is already managed by FSDP.")
+    context = _FSDP_CONTEXT.get()
+    if context is None:
+        raise RuntimeError("fully_shard must run inside fully_shard_context.")
 
     placements = _normalize_placements(mesh, placements)
     mixed_precision_policy = mixed_precision_policy or MixedPrecisionPolicy()
@@ -58,6 +92,7 @@ def fully_shard(
         assert isinstance(module, FsdpModule)
         FsdpModule.__init__(
             module,
+            context=context,
             mesh=mesh,
             placements=placements,
             mixed_precision_policy=mixed_precision_policy,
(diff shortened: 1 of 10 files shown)


megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -28,36 +28,76 @@
...
+        visited: set[FsdpModule] = set()
+        root_modules: list[FsdpModule] = []
+        for module in reversed(self._registered_modules):
+            if module in visited:
+                continue
+            root_modules.append(module)
+            _collect_fsdp_modules_under(cast(nn.Module, module), visited)
+        root_modules.reverse()
+        for root_module in root_modules:
+            root_module._is_root = True
```
wujingyue: Note to myself: there's an easy way to compute root_modules without assuming the order of registered_modules

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
```diff
@@ -17,14 +17,48 @@
...
 
 from ..mixed_precision import MixedPrecisionPolicy
 from .module import FsdpContext, FsdpModule
 from .placement import MeshAxis, Placements
 
+_FSDP_CONTEXT = ContextVar[FsdpContext | None]("megatron_fsdp_context", default=None)
+
+
+@contextmanager
+def fully_shard_context(device: torch.device | None = None) -> Iterator[FsdpContext]:
```
Autumn1998, line 34: Example of silent misbehavior:

```
with fully_shard_context(device=device) as ctx1:
    fully_shard(model.inner, mesh=mesh, placements=placements)
with fully_shard_context(device=device) as ctx2:
    fully_shard(model, mesh=mesh, placements=placements)
# ctx2.finalize() silently includes model.inner in its forward_order,
# while model.inner._is_root=True from ctx1 still holds.

```
Maybe worth adding a guard or UT that detects when an FsdpModule being registered already belongs to another context, and raises explicitly.

Also, If we require all `fully_shard` calls to live under a single continuous fully_shard_context, does that make it harder to align with the FSDP2 API? (I recall the plan was to achieve compatibility via a wrapper)

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -238,7 +239,7 @@ def train(model, optimizer, log_prefix) -> list[torch.Tensor]:
 
             for microbatch_index, (microbatch_x, microbatch_target) in enumerate(microbatches):
                 is_last = microbatch_index == num_microbatches - 1
-                with microbatch(model, is_last=is_last):
+                with microbatch(context, is_last=is_last):
```
Autumn1998, line 268: If the context is already attached to the model, would it be cleaner to retrieve it via something like model.get_context()?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
wujingyue: > Maybe worth adding a guard or UT that detects when an FsdpModule being registered already belongs to another context, and raises explicitly.

Yep! Good idea.

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
wujingyue: If `model` is an `FsdpModule`, it already has `model.context`.

Otherwise, I’d prefer fewer places to track global state. Since this PR introduces `fully_shard_context`, we can reuse it for all knobs shared across `FsdpModule`s. Along those lines, perhaps `fully_shard_optimizer` should take `context` rather than `model` as well.

shjwudp (approved):

Phlip79 (approved):

jaredcasper (approved):

PR #6200, opened 2026-08-02
[feat] Add reduce-scatter-with-fp32-accumulation support for GTP

### LINKED ISSUE #6199: [feat] GTP + grad-reduce-in-fp32
## Requested Feature
When `--grad-reduce-in-bf16` is set, currently GTP reduceScatter will call plain **ring** `bf16 nccl-reduceScatter` by default.

We also need to support GTP with `reduce-scatter-with-fp32-accumulation` for better accuracy especially at large scale:

> bf16 reduceScatter --> bf16 all2 all + fp32 local reduction

## What it brings
- **Accuracy**: 
      - A ring reduce-scatter rounds the partial sum at every one of its N-1 hops, so BF16 gradient error grows as √N with the axis size
      - `reduce-scatter-with-fp32-accumulation` replaces it with an all-to-all plus **oneshot** local FP32 sum, killing that accumulation error entirely for the same bytes on the wire.

## What it costs
- **Memory**: one extra full-size scratch buffer to land the all-to-all in.
- **Reduction no longer overlaps transfer**: NCCL's ring folds the adds into the communication; here the entire FP32 sum is local work at `.wait()` time.

fanshiqing:

# What does this PR do?
Add `resuce-scatter-with-fp32-accumulation` support for GTP.

# Details
New flag `--gtp-remat-reduce-scatter-with-fp32-accumulation` (default OFF), independent of the DDP flag.

Shared primitive gains two optional arguments:
        * scale -- **applies the 1/gtp_remat mean to the FP32 sum** instead of pre-scaling the BF16 wgrad, which would round away exactly the precision this change exists to preserve.
        * all_to_all_output_tensor -- lets the caller supply the scratch, so GTP draws it from its `wgrad pool` instead of allocating one per in-flight reduce-scatter.

**Auto-bypassed at axis size <= 2**, where one addition rounds the same either way and the scratch buffer would be pure cost. The gate reads the per-chain group, so GTP_remat=8 x EGTP_remat=2 gets FP32 accumulation on the dense weights and the plain reduce-scatter on the experts.

 **Batched (grouped / routed-expert) all-to-alls are coalesced**, as the plain batched reduce-scatter already is. The manager cannot BE the  handle -- it waits only the NCCL work it collects, while each fp32-accum handle still owes a local FP32 sum -- so the sums are deferred behind it in one composite handle." is still too long

## Convergence Test

Model: 1/4 of NT3 Ultra
Parallelism: GTP16DP2EP16EGTP2_mb1gb32
Dataset: `1T-phase1var-moresft.json`
Training from scratch for ~10k steps.

Ablation experiments: w/ and w/o `--gtp-remat-reduce-scatter-with-fp32-accumulation`

**Results**: both lm loss and validation loss matches well; also match the loss curve from 3D parallelism.

image

## Issue tracking

Linked issue: Fixes #6199 

### Pre-checks

### DIFF
--- a/megatron/core/distributed/reduce_scatter_with_fp32_accumulation.py
+++ b/megatron/core/distributed/reduce_scatter_with_fp32_accumulation.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
 
 
-from typing import Any
+from typing import Any, Optional
 
 import torch
 
@@ -16,12 +16,14 @@ class _ReduceScatterWithFP32AccumulationWorkHandle:
         all_to_all_output_tensor: torch.Tensor,
         output_tensor: torch.Tensor,
         world_size: int,
+        scale: Optional[float] = None,
     ):
         """Initialize WorkHandle object."""
         self.all_to_all_handle = all_to_all_handle
         self.all_to_all_output_tensor = all_to_all_output_tensor
         self.output_tensor = output_tensor
         self.world_size = world_size
+        self.scale = scale
 
     def wait(self):
         """Wait until communication (and associated computation) is completed."""
@@ -35,8 +37,13 @@ class _ReduceScatterWithFP32AccumulationWorkHandle:
         )
         assert output_tensor_in_fp32.dtype == torch.float32
 
+        # Apply the mean/normalization factor while still in fp32; scaling the low-precision
+        # input beforehand would round away the precision this implementation preserves.
+        if self.scale is not None:
+            output_tensor_in_fp32 *= self.scale
+
         # Copy downcasted sum into output_tensor.
-        self.output_tensor.copy_(output_tensor_in_fp32)
+        self.output_tensor.copy_(output_tensor_in_fp32.view(self.output_tensor.shape))
 
 
 def reduce_scatter_with_fp32_accumulation(
@@ -45,6 +52,8 @@ def reduce_scatter_with_fp32_accumulation(
     op: torch.distributed.ReduceOp,
     group: torch.distributed.ProcessGroup,
     async_op: bool,
+    scale: Optional[float] = None,
+    all_to_all_output_tensor: Optional[torch.Tensor] = None,
 ):
     """Reduce-scatter with FP32 accumulation.
 
@@ -58,6 +67,11 @@ def reduce_scatter_with_fp32_accumulation(
         op (torch.distributed.ReduceOp): Only torch.distributed.ReduceOp.SUM is supported.
         group (torch.distributed.ProcessGroup): Process group to use for reduce-scatter.
         async_op (bool): Only False is supported right now.
+        scale (float, optional): Factor applied to the FP32 sum before the downcast. Use this
+            for a mean-reduction rather than pre-scaling input_tensor, which rounds on the way in.
+        all_to_all_output_tensor (torch.Tensor, optional): Caller-provided scratch matching
+            input_tensor's shape and dtype, for callers with their own buffer pool. Allocated
+            internally when omitted; must stay alive until .wait() returns.
     """
     # Make sure arguments conform to the implementation.
     assert op == torch.distributed.ReduceOp.SUM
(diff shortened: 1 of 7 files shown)


fanshiqing:
@claude strict-review

fanshiqing:
@claude strict-review

fanshiqing:
@claude review

megatron/core/distributed/reduce_scatter_with_fp32_accumulation.py
```diff
@@ -58,6 +67,11 @@ def reduce_scatter_with_fp32_accumulation(
         op (torch.distributed.ReduceOp): Only torch.distributed.ReduceOp.SUM is supported.
         group (torch.distributed.ProcessGroup): Process group to use for reduce-scatter.
         async_op (bool): Only False is supported right now.
+        scale (float, optional): Factor applied to the FP32 sum before the downcast. Use this
```
deepakn94: I don't love scale being added to the API here. Would ideally like to keep API compatibility with the PyTorch reduce-scatter function as much as possible.

I also don't think it provides any benefit: let's think of dividing by scale as essentially a bit-shift by log2(scale) bits. I don't think `cast_to_bf16(sum >> bit_shift)` and `cast_to_bf16(sum) >> bit_shift` are any different when `sum >> bit_shift` doesn't devolve into sub-normals (which is below `2^-126`)?

A compromise could me to also support a `torch.distributed.ReduceOp.AVG` op.

fanshiqing: Essentially you are right:
Pre-scaling by a **power of two** is an exact exponent decrement, so bf16(w*s) loses nothing and `Σ(w_i·s) == s·Σ(w_i)` bitwise. The scale argument buys nothing for any **realistic `gtp_remat_size`**. 

Let me  refine.

fanshiqing: Done~

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1519,13 +1545,23 @@ def _wait_reduce_scatter(self, finalize_grad=False):
...
-        # Release stashed wgrad inputs: UNGRAPHED buffers go back to the pool;
-        # GRAPHED just drops Python refs (addresses must stay stable for CG).
-        if getattr(self, "_wgrad_input_bufs", None) is not None:
+        self._release_comm_scratch()
+
+    def _release_comm_scratch(self, attrs=("_wgrad_input_bufs", "_rs_a2a_bufs")):
+        """Release the buffers a finished RS was reading.
+
+        Its wgrad inputs, and the fp32-accum all-to-all scratch (input to the deferred FP32 sum,
+        so only free once the handle has been waited). UNGRAPHED buffers go back to the pool;
```
deepakn94: Nit: "once the handle has been waited on".

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1586,6 +1656,42 @@ def _reduce_scatter(self, wgrads, async_op, nvtx_label=None):
...
         with rs_ctx:
+            # Size <= 2 gains nothing (one addition) and still costs the scratch, so it bypasses.
+            # self.group is per chain, so each axis decides on its own.
+            if GTP_CONFIG.reduce_scatter_with_fp32_accumulation and self.group.size() > 2:
+                nvtx_range_push(f"{nvtx_label}.gtp_rs_fp32accum")
+                outputs, sum_handles = [], []
+                if len(wgrads) > 1:
+                    # Batched: group the all-to-alls into one ncclGroupStart/End, exactly like
+                    # the plain batched RS below. Unlike that path we cannot hand the manager
+                    # back as the handle: the manager waits only the NCCL work it collects, and
```
deepakn94: "the manager waits only on the underlying NCCL collective to complete"

deepakn94 (approved):

yaoyu-33 (approved):

PR #6203, opened 2026-08-03
[FSDP] Reject post-wrap load_state_dict(assign=True) in MFSDP V2

### LINKED ISSUE #6087: [MFSDP V2] Post-wrap load_state_dict(assign=True) does not resume from the loaded checkpoint
## Observed behavior

At `bc67abde107ca0a7751208ae142d09ce602e45be`, post-wrap
`load_state_dict(checkpoint, assign=True)` returns a clean load. For the fixed
identity input and zero target, the pre-load weights have loss `0.390625`,
while the checkpoint weights have loss `0.44921875`. The measured loss after
the load is `0.390625000`, which matches the pre-load weights rather than the
checkpoint weights.

Verified with PyTorch 2.8.0+cu128, CUDA 12.8, and two NVIDIA
A100-SXM4-80GB GPUs. The complete stdout from each of two fresh DP=2 runs was:

```text
NCCL version 2.27.3+cuda12.9
load_result: missing_keys=[], unexpected_keys=[]
step | loss
   0 | 0.390625000
   1 | 0.390625000
   2 | 0.390625000
   3 | 0.390625000
   4 | 0.390625000
   5 | 0.390625000
   6 | 0.390625000
   7 | 0.390625000
   8 | 0.390625000
   9 | 0.390625000
  10 | 0.390625000
  11 | 0.390625000
  12 | 0.390625000
  13 | 0.390625000
  14 | 0.390625000
  15 | 0.390625000
  16 | 0.390625000
  17 | 0.390625000
  18 | 0.390625000
  19 | 0.390625000
  20 | 0.390625000
loss_decreased=False
```

This is a deterministic convex MSE problem: a bias-free linear layer receives
the identity matrix as input and is trained toward a zero target with SGD. The
checkpoint contains fixed nonzero weights and comes from the normal
`state_dict()` of an identically sharded model, with every value frozen by
`detach().clone()`. The optimizer is constructed only after the `assign=True`
load. The measured loss is identical at every one of the 20 steps: the model does not train, and it stays at the loss of its own pre-load weights.
Both fresh `torchrun` processes exited 0 and produced byte-identical stdout.

## Minimal reproduction

Save the following as `repro_issue_6087_assign_true_only.py` in the
Megatron-LM repository root:

```python
#!/usr/bin/env python3
"""Behavior-only reproduction for post-wrap load_state_dict(assign=True)."""

from __future__ import annotations

import os

import torch
import torch.distributed as dist
import torch.nn.functional as F
from torch import nn
from torch.distributed.device_mesh import init_device_mesh

from megatron.core.distributed.fsdp.src.megatron_fsdp.experimental import (
    Flat,
    Placements,
    fully_shard,
)


WORLD_SIZE = 2
SEED = 6087
STEPS = 20
LEARNING_RATE = 0.5


def build_model() -> nn.Linear:
    return nn.Linear(4, 4, bias=False)


def placements() -> Placements:
    return Placements(dp_axes=[0], parameter=[Flat()], gradient=[Flat()], optimizer=[Flat()])


def set_weight(model: nn.Linear, values: list[list[float]]) -> None:
    with torch.no_grad():
        model.weight.copy_(torch.tensor(values, dtype=torch.float32, device=model.weight.device))


def evaluate_loss(model: nn.Module, inputs: torch.Tensor, targets: torch.Tensor) -> float:
    with torch.no_grad():
        loss = F.mse_loss(model(inputs), targets)
    dist.all_reduce(loss, op=dist.ReduceOp.SUM)
    return (loss / dist.get_world_size()).item()


def main() -> None:
    if not torch.cuda.is_available():
        raise RuntimeError("CUDA is required")

    local_rank = int(os.environ["LOCAL_RANK"])
    torch.cuda.set_device(local_rank)
    dist.init_process_group("nccl")
    try:
        rank = dist.get_rank()
        world_size = dist.get_world_size()
        if world_size != WORLD_SIZE:
            raise RuntimeError(f"Expected world_size={WORLD_SIZE}, got {world_size}")

        torch.manual_seed(SEED)
        torch.use_deterministic_algorithms(True)
        torch.backends.cuda.matmul.allow_tf32 = False
        torch.backends.cudnn.allow_tf32 = False

        device = torch.device("cuda", local_rank)
        mesh = init_device_mesh("cuda", (world_size,))
        shard_placements = placements()
        inputs = torch.eye(4, dtype=torch.float32, device=device)
        targets = torch.zeros(4, 4, dtype=torch.float32, device=device)

        source_model = build_model().to(device)
        set_weight(
            source_model,
            [
                [1.00, -0.50, 0.25, -0.75],
                [-0.25, 0.75, -1.00, 0.50],
                [0.50, 0.25, -0.50, 1.00],
                [-0.75, 1.00, 0.50, -0.25],
            ],
        )
        fully_shard(source_model, mesh=mesh, placements=shard_placements)
        checkpoint = {
            name: value.detach().clone() for name, value in source_model.state_dict().items()
        }

        model = build_model().to(device)
        set_weight(
            model,
            [
                [0.25, -0.75, 0.50, -0.25],
                [0.50, 0.25, -0.75, 1.00],
                [-1.00, 0.50, 0.25, -0.50],
                [0.75, -0.25, 1.00, 0.50],
            ],
        )
        fully_shard(model, mesh=mesh, placements=shard_placements)
        load_result = model.load_state_dict(checkpoint, assign=True)

        optimizer = torch.optim.SGD(
            model.parameters(),
            lr=LEARNING_RATE,
            momentum=0.0,
            weight_decay=0.0,
            foreach=False,
        )

        losses = [evaluate_loss(model, inputs, targets)]
        for _ in range(STEPS):
            optimizer.zero_grad(set_to_none=True)
            loss = F.mse_loss(model(inputs), targets)
            loss.backward()
            optimizer.step()
            losses.append(evaluate_loss(model, inputs, targets))

        loss_decreased = losses[-1] < losses[0]
        if rank == 0:
            print(
                "load_result: "
                f"missing_keys={list(load_result.missing_keys)}, "
                f"unexpected_keys={list(load_result.unexpected_keys)}"
            )
            print("step | loss")
            for step, loss in enumerate(losses):
                print(f"{step:>4} | {loss:.9f}")
            print(f"loss_decreased={loss_decreased}", flush=True)

        dist.barrier()
    finally:
        if dist.is_initialized():
            dist.destroy_process_group()


if __name__ == "__main__":
    main()
```

Run from the repository root:

```bash
# CUBLAS_WORKSPACE_CONFIG is required by torch.use_deterministic_algorithms(True).
CUBLAS_WORKSPACE_CONFIG=:4096:8 NCCL_DEBUG=WARN \
torchrun --standalone --nproc-per-node=2 \
repro_issue_6087_assign_true_only.py
```

## Expected behavior

Post-wrap `load_state_dict(assign=True)` should make continuation training use
the loaded parameters, or it should be explicitly rejected during load with a
clear error. It should not return clean success and then leave the model neither loaded nor training.

chengcuiping:

## What does this PR do?

Rejects MFSDP V2 `load_state_dict(assign=True)` calls made after `fully_shard()`, before the affected `FsdpModule` parameters can be replaced.

Post-wrap `assign=True` currently returns successfully while the next MFSDP forward continues using the previous state. This change registers a load-state pre-hook on each `FsdpModule`, reads the assign metadata propagated by PyTorch, and raises a clear error directing callers to load before `fully_shard()` or to use the in-place `assign=False` path.

Regression coverage contains one behavior-level test that fully shards a model and verifies that loading a state dict with `assign=True` raises the documented error.

## Issue tracking

Linked issue: Closes #6087

## Testing

- **PASS** — focused `isort --check-only` for both changed Python files.
- **PASS** — focused `ruff check --no-fix` for both changed Python files.
- **PASS** — `git diff --check`.
- **BLOCKED (local environment)** — the focused DP=2 pytest run stopped during fixture setup on both ranks because the available PyTorch build does not provide `torch.distributed._symmetric_memory.set_backend`. The test body was not executed.
- **NOT RUN** — the complete DP=2 `test_checkpoint.py` file, because the focused test could not pass the fixture-setup gate. The canonical checkpoint suite is left to project CI.
- **BLOCKED (local environment)** — `BASE_REF=main CHECK_ONLY=true SKIP_DOCS=false bash tools/autoformat.sh` requires Black 26 and Pylint; the available environment has Black 25.1.0 and does not have Pylint. No dependencies were installed or upgraded.

> **Note:** the latest review follow-up only simplifies the regression test; it does not modify the production implementation.

### Pre-checks


megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -221,6 +222,24 @@ def grad_hook(_parameter: nn.Parameter) -> None:
...
+        _state_dict: dict[str, object],
+        _prefix: str,
+        local_metadata: dict[str, object],
+        _strict: bool,
+        _missing_keys: list[str],
+        _unexpected_keys: list[str],
+        _error_msgs: list[str],
+    ) -> None:
+        """Reject state-dict loads that replace parameters managed by FSDP."""
+        if local_metadata.get("assign_to_params_buffers", False):
```
wujingyue, line 293: Sounds fragile to assume the key is a particular string. Any alternatives?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
wujingyue, line 1: Should probably move to test_checkpoint.py after #6024 is in.

wujingyue (commented):
LGTM overall. Notify me after #6024 is merged!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
chengcuiping: You're right that this is still an internal-ish dependency. I checked for alternatives: the public pre-hook does not receive `assign` directly. PyTorch 2.8 threads it through `local_metadata["assign_to_params_buffers"]`, which `_load_from_state_dict` [[documents](https://github.com/pytorch/pytorch/blob/v2.8.0/torch/nn/modules/module.py#L2353-L2356)](https://github.com/pytorch/pytorch/blob/v2.8.0/torch/nn/modules/module.py#L2353-L2356) and consumes.

Overriding `load_state_dict` would miss parent-recursive loads, while detecting replacement after the load is too late. I couldn't find another pre-mutation signal. The structural alternative would be a broader guard on every MFSDP-owned parameter replacement. Would you prefer that direction, or is this metadata dependency acceptable?

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
chengcuiping: Agreed. I’ll keep this PR in draft, rebase after #6024 lands, move these load-state tests to `test_checkpoint.py`, and run the full DP=2 checkpoint test file before requesting review again.

chengcuiping:
Hi @wujingyue, #6024 is now merged, and I have rebased this PR onto it. The updated revision is `0d9f70abb34c985fd9635b977b3620285eb28971` and remains a single commit. I have marked the PR ready for review again.

Since your previous review, I moved the load-state coverage to `test_checkpoint.py` and changed the load-state hook registration to use an unbound static callback so that it does not retain the `FsdpModule`.

One design point still needs your confirmation. I re-audited the supported PyTorch 2.6–2.8 load paths, but could not find a public pre-mutation alternative to `local_metadata["assign_to_params_buffers"]`: the public pre-hook does not receive `assign`, overriding `load_state_dict()` misses recursive loads through an ordinary parent, and detecting replacement after the load is too late. The current revision therefore keeps the narrow metadata guard and documents that it is PyTorch’s internal propagation protocol. Would you be comfortable with this dependency, or would you prefer the broader structural guard on MFSDP-owned parameter replacement?

I also want to call out one gap against my earlier testing plan. I intended to run the complete DP=2 checkpoint test file before requesting review again, but the canonical pytest entry point is blocked before test execution because my local PyTorch 2.8 build does not provide `torch.distributed._symmetric_memory.set_backend`, which the current-main fixture requires. Supplemental DP=2 coverage passed on both ranks for all four load-state contracts, SGD parity, module retention, and the FP32 DCP roundtrip; the repository check-only lint pipeline and `compileall` also passed. Full details are recorded in the PR description.

Could a PR vetter approve the workflows for `0d9f70abb34c985fd9635b977b3620285eb28971` so the complete checkpoint suite can run in the repository CI environment?

tests/unit_tests/distributed/mfsdp_v2/test_checkpoint.py
```diff
@@ -39,6 +39,106 @@ def _flat_placements() -> Placements:
...
+
+def _assert_assign_true_error(error: RuntimeError) -> None:
+    message = str(error)
+    assert "load_state_dict(assign=True)" in message
+    assert "fully_shard()" in message
+    assert "assign=False" in message
+
+
+@pytest.mark.parametrize("load_from_parent", [False, True], ids=["direct", "parent"])
+def test_post_wrap_assign_true_load_raises(distributed_setup, load_from_parent: bool):
```
wujingyue: Agents tend to generate too many over-defensive rejection tests. Most of them don't even follow https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html

How about just **one** rejection test that tries to load a checkpoint with assign=True?

chengcuiping:
Agreed — I reduced this to a single behavior-level rejection test. It now creates a regular state dict, fully shards the target model, and verifies that `load_state_dict(..., assign=True)` raises the documented error.

I removed the direct/parent matrix, the post-failure state assertions, and the separate allowed-path cases. The production implementation is unchanged in this revision.

The focused DP=2 run remains blocked before the test body by the local fixture requiring `torch.distributed._symmetric_memory.set_backend`, which is unavailable in this environment, so the canonical run is left to CI.

tests/unit_tests/distributed/mfsdp_v2/test_checkpoint.py
```diff
@@ -39,6 +39,23 @@ def _flat_placements() -> Placements:
...
+    device = distributed_setup.device
+    mesh = init_device_mesh(device.type, (distributed_setup.world_size,))
+    checkpoint = nn.Linear(4, 4, bias=False, device=device).state_dict()
+    model = nn.Linear(4, 4, bias=False, device=device)
+    with fully_shard_context(device=device):
+        fully_shard(model, mesh=mesh, placements=_flat_placements())
+
+    with pytest.raises(
+        RuntimeError,
+        match=r"load_state_dict\(assign=True\) is not supported after fully_shard\(\)",
```
wujingyue:
```suggestion
        match=r"load_state_dict\(assign=True\)",
```

wujingyue (approved):
Looks great! Thanks for the PR

tests/unit_tests/distributed/mfsdp_v2/test_checkpoint.py
```diff
@@ -39,6 +39,23 @@ def _flat_placements() -> Placements:
     return Placements(dp_axes=[0], parameter=[Flat()], gradient=[Flat()], optimizer=[Flat()])
 
 
+def test_post_wrap_assign_true_load_raises(distributed_setup):
+    if distributed_setup.world_size < 2:
+        pytest.skip("This test requires at least 2 ranks.")
```
wujingyue:
```suggestion
```

wujingyue (approved):

tests/unit_tests/distributed/mfsdp_v2/test_checkpoint.py
chengcuiping: @wujingyue  Thanks! Applied the two final review suggestions in signed and verified commit `86d949102e2955d0d8cd0d449c5a59486fe196e5`:

- removed the redundant `world_size < 2` skip;
- narrowed the `pytest.raises` match to `load_state_dict\(assign=True\)`.
Focused `diff-check`, isort, Ruff, and Python compilation checks pass. Could you or a maintainer approve the workflows for this revision? Thanks.

ahmadki (approved):

PR #6122, opened 2026-07-29
Add configurable attention latent normalization epsilon

buptzyb:
(Dev PR: #6204)

## Summary

- add `attention_latent_norm_epsilon` to `MLATransformerConfig` with a matching `--attention-latent-norm-epsilon` argument, inheriting `layernorm_epsilon` when unset for backward compatibility
- apply the dedicated epsilon to primary query and key-value latent norms in standard, fused, absorbed, and inference-optimized MLA/DSA paths, independently of the RMSNorm/LayerNorm selection
- keep DSA indexer K normalization on its existing `dsa_indexer_k_norm_epsilon` control and cover distinct global, attention-latent, and indexer epsilons with resolver, module, and native-parity tests

## Test plan

- `tests/unit_tests/transformer/test_mla_qk_norm_config.py`: 6 passed
- `tests/unit_tests/transformer/test_multi_latent_attention.py`: H100 active shards 55 passed, 27 deselected
- `tests/unit_tests/transformer/experimental_attention_variant/test_absorbed_mla.py`: H100 active shards 64 passed
- DSA native parity: passed on H100 and GB200

PR #6204, opened 2026-08-03
[Dev] Add configurable attention latent normalization epsilon

buptzyb:
(Main PR: #6122)

## What

- Port Multi-Latent Attention support to `HybridModel` as the prerequisite already used on `main`.
- Add `MLATransformerConfig.attention_latent_norm_epsilon` and `--attention-latent-norm-epsilon`.
- Default the new option to `layernorm_epsilon` when it is unset.
- Apply the option to standard MLA/DSA, absorbed MLA, and the primary DSv4 Q/KV latent norms.

This is the dev-targeted counterpart of #6122. The main PR remains open and unchanged.

## Scope

The option only selects epsilon; `normalization` still selects RMSNorm versus LayerNorm. It does not change DSA indexer K normalization, CSA/HCA compressors, DSv4 post-up Q normalization, or the model-wide input/pre-MLP/final norms.

The commits are intentionally split into:

1. the main HybridModel/MLA prerequisite;
2. the generic MLA/DSA option;
3. dev-only DSv4 integration and parity coverage.

## Testing

- Black, isort, pylint, and ruff passed through `tools/autoformat.sh` in check-only mode.
- On one OCI-HSG node (4x GB200), using the CI-dev image:
  - `test_mla_qk_norm_config.py`: 6 passed per effective rank.
  - `test_absorbed_mla.py`: 64 passed per effective rank.
  - `test_dsv4_hybrid_native_parity.py`: 78 passed, 70 skipped per effective rank; fused DSv4 cases retain their existing GB200 skip condition, while the new unfused epsilon parity case passed.
- `test_multi_latent_attention.py` passed the two new epsilon cases, then hit the existing GB200/TE failure in `test_gpu_forward_thd_padded[yarn]` (`No dot product attention backend is available`). Latest `origin/dev` reproduced the same test and exception on the same node and image, so this is not a regression from this PR.

### DIFF
--- a/megatron/core/extensions/transformer_engine.py
+++ b/megatron/core/extensions/transformer_engine.py
@@ -1009,6 +1009,7 @@ class TELayerNormColumnParallelLinear(te.pytorch.LayerNormLinear):
         tp_group: Optional[torch.distributed.ProcessGroup] = None,
         stride: int = 1,
         name: str | None = None,
+        eps: float | None = None,
     ):
         """
         Args:
@@ -1117,7 +1118,7 @@ class TELayerNormColumnParallelLinear(te.pytorch.LayerNormLinear):
             super().__init__(
                 in_features=input_size,
                 out_features=output_size,
-                eps=self.config.layernorm_epsilon,
+                eps=self.config.layernorm_epsilon if eps is None else eps,
                 sequence_parallel=self.config.sequence_parallel,
                 fuse_wgrad_accumulation=self.config.gradient_accumulation_fusion,
                 tp_group=tp_group if torch.distributed.is_initialized() else None,
--- a/megatron/core/extensions/transformer_engine.py
+++ b/megatron/core/extensions/transformer_engine.py
@@ -1009,10 +1009,13 @@ class TELayerNormColumnParallelLinear(te.pytorch.LayerNormLinear):
         tp_group: Optional[torch.distributed.ProcessGroup] = None,
         stride: int = 1,
         name: str | None = None,
+        eps: float | None = None,
     ):
         """
         Args:
             name (str | None): module instance name passed top-down from its paranet module
+            eps (float | None): Epsilon for the fused layer norm. Defaults to
+                ``config.layernorm_epsilon`` when ``None``.
         """
         if not HAVE_TE:
             raise ImportError(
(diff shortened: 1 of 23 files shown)


megatron/core/extensions/transformer_engine.py
```diff
@@ -1009,6 +1009,7 @@ def __init__(
         tp_group: Optional[torch.distributed.ProcessGroup] = None,
         stride: int = 1,
         name: str | None = None,
+        eps: float | None = None,
```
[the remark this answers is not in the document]
buptzyb: Fixed in 2fb29d266. The eps parameter is now documented on both TELayerNormColumnParallelLinear and InferenceLayerNormColumnParallelLinear, including the None fallback to config.layernorm_epsilon.

tests/unit_tests/transformer/experimental_attention_variant/test_dsv4_hybrid_native_parity.py
```diff
@@ -1001,6 +1002,22 @@ def teardown_method(self):
         gc.collect()
         torch.cuda.empty_cache()
 
+    @pytest.mark.launch_on_gb200
+    @pytest.mark.parametrize(("backend", "use_fused_kernels"), _DSA_BACKENDS)
+    def test_attention_latent_norm_epsilon_matches_native_reference(
```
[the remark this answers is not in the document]
buptzyb: Fixed in e32b1f783. The redundant GB200-only wrapper was removed. The existing DSv4 parity matrix now asserts that Q/KV latent norms use attention_latent_norm_epsilon, compressor norms retain layernorm_epsilon, and the native query RMS norm receives the global epsilon.

megatron/core/transformer/experimental_attention_variant/absorbed_mla.py
```diff
@@ -266,7 +269,7 @@ def __init__(
         if self.config.q_lora_rank is None:
             # Not projecting query
             self.linear_q_proj = build_module(
-                submodules.linear_q_proj,
+                layer_classes["linear_q_proj"],
```
hxbai, line 269: Why does it need to override in the init func rather than replace the spec directly?

buptzyb: I don't know... This is carried over from #4452... This PR needs 4452's mla_qk_norm_config.py.

This is how my agent replies to your question:

> The init-time resolution is needed for the config-independent hybrid_stack_spec, since Q/KV norm placement depends on runtime config and backend, and the resolver also validates custom specs. Simply using the original field here would break the static Hybrid MLA/DSA path. I agree that resolving this into a config-aware copied spec before module construction would be cleaner, but that would be a broader #4452 refactor rather than part of the epsilon change.

hxbai: OK. I see.

hxbai (approved):

PR #6206, opened 2026-08-03
[dev](perf): DSA pre-attention: plan-driven scorer dispatch, host-built CP layout metadata

jingqiny-99:
## Summary

This PR makes DSA indexer scoring correct for general packed THD inputs and removes repeated layout construction from the packed context-parallel hot path.

The central semantic correction is that packed sequence cardinality is independent of context parallelism: CP=1 can contain multiple sequences, and CP>1 can contain either one or many. Scorer selection now follows the actual THD metadata and kernel preconditions instead of using CP size as a proxy for sequence count.

The final branch preserves existing configuration and backend-hook compatibility. It does **not** claim full CUDA Graph support for packed DSA+CP. Such configurations remain constructible and receive an experimental-support warning; capture/replay support is follow-up work.

### 1. Plan-driven general THD scoring

- Add a total scoring-plan resolver with explicit outcomes for plain causal scoring, packed single-segment scoring, general packed THD scoring, unfused bounds, and decline.
- Run multi-sequence CP=1 scoring as one cuDNN THD call using the original packed Q/K/W tensors and cu-seqlens.
- Run CP>1 general packed scoring from the same THD metadata, using per-sequence zigzag query segments and gathered-key reorder metadata.
- Keep the single-sequence specialization only when its structural preconditions hold.
- Pass explicit key-position state through the production dispatcher so unsupported layouts reach the plan's `DECLINE` outcome before a cuDNN launch.
- Emit one diagnostic when a valid layout intentionally uses the existing unfused bounds path.

This removes both invalid implications: CP=1 no longer means “one sequence”, and CP>1 no longer means “multiple sequences”.

### 2. Host-built packed-CP layout metadata

- Record compact host Q/K cu-seqlens during the existing THD route-prebuild stage.
- Build all-rank query positions and gathered-key reorder tables from those host integers.
- Construct the reorder as a direct inverse permutation instead of a full argsort.
- Copy each finished table to the target device with a direct tensor transfer.
- Memoize layouts on the per-microbatch `PackedSeqParams`, so later DSA layers reuse the first build.
- Use the computed Q/K output-cover flags in both `positions_and_reorder` memo sites, so equal geometry has an equal cache key.

Independent CPU goldens cover CP1/2/4, both/all ranks, multiple sequences, zero-length entries, all-empty packs, padding, distinct Q/K boundaries, and every Q/K cover-flag combination. A production-forward regression test also enters both memo sites and proves that the second site reuses the first result.

### 3. Compatibility and invariant hardening

- Keep the existing backend-hook kwargs `use_local_indexer_varlen` and `single_packed_thd_sequence` on their CP-only contract.
- Offer CP-independent layout facts as separate optional kwargs.
- Move callable-signature filtering into `megatron.core.utils.filter_kwargs_for_callable`; exact signatures, `**kwargs`, positional-only parameters, opaque callables, and the identity-safe signature cache have direct tests.
- Preserve the pre-existing opaque full-hook `varlen_is_plain_causal` contract.
- Preserve valid self-attention Q/K metadata aliases through THD padding and inherited metadata.
- Reject inconsistent scorer devices, dtypes, views, TP slices, and gradient arity before producing incorrect results.
- Pass configured `cp_comm_type` into attention layers for Dynamic CP even when configured CP is 1; the effective microbatch CP group may still be larger than 1.
- Keep that effective Dynamic CP group inside selective core-attention checkpoint replay, then restore the caller's group after both the original call and recomputation.
- Restore the non-optional `starts`/`ends` autograd contract after the caller's concrete-bounds guard.

### 4. CUDA Graph scope

Packed DSA+CP graph configurations are no longer rejected by an empirical hardware/version matrix. The final check is structural and warning-only:

- static packed DSA warns when configured CP > 1 and CUDA graphs are enabled;
- Dynamic CP packed DSA warns whenever CUDA graphs are enabled, including configured CP=1, because the effective per-microbatch group can be larger;
- the warning is independent of scorer backend and capture module scope;
- eager execution (`cuda_graph_impl="none"`) is unchanged.

The evidence is diagnostic, not an allow/deny table. On the validation GB200 image, cuDNN static CP2 eager execution passes, while local attention capture reaches graph warmup and fails when the cloned packed metadata is empty. Backend `none` was previously measured only for static CP2/local/attention and also failed during graph warmup. TileLang was unavailable in that image and remains unmeasured. No unmeasured combination is presented as known-good, and none is hard-rejected by this PR.

Full capture/replay support for changing packed layouts is intentionally deferred.

## Validation

All GPU work below ran on Lyris GB200 in the standard `gb200` partition (no backfill), using `dsv4-gb200-torch2604-nvrx060`.

- Final consolidated targeted suite: **164 passed** — config compatibility/warnings, scoring plans, hook ABI, generic callable filtering, independent host-layout goldens and cache lifetime, cross-site memo reuse, Dynamic-CP communication and selective-recompute group lifetime, explicit-position dispatch, plus real cuDNN CP2 parity on both zigzag ranks and three seeds against an independent PyTorch forward/backward reference.
- Profiler-free model probes:
  - static CP2 eager: pass across one- and multi-sequence boundary changes;
  - Dynamic CP configured CP2 / effective CP1 eager: pass;
  - Dynamic CP configured CP1 / effective CP2 eager: pass after the communication-propagation fix;
  - static CP2 local-attention graph: configuration constructs and warns, eager preflight runs, then capture fails at graph warmup on empty cloned metadata (known deferred boundary).
- Selected H2D A/B: direct transfer reduced helper latency by **3.0%–17.9%** across the measured CP2/4/8/16 layouts versus pin-then-transfer, with bit-identical outputs. This is a helper-level measurement, not an end-to-end iteration percentage.

Only the conclusion-relevant H2D nsys report was retained; correctness probes were run without profiling.

## Commit map

1. Define explicit indexer scoring plans.
2. Preserve optional backend-hook compatibility.
3. Preserve THD Q/K metadata aliases.
4. Prebuild packed CP layout metadata.
5. Support general packed THD indexer scoring.
6. Add the initial packed DSA+CP graph check.
7. Preserve inherited THD metadata aliases.
8. Harden packed cuDNN scorer invariants.
9. Refine graph and hook compatibility from review feedback.
10. Final review follow-up: remove empirical hard rejection, fix memo/H2D/dispatcher contracts, generalize the hook helper, cover Dynamic CP configured-CP1 and selective recomputation, and expand independent/GB200 tests.

### DIFF
--- a/megatron/core/transformer/cuda_graph_config.py
+++ b/megatron/core/transformer/cuda_graph_config.py
@@ -59,6 +59,110 @@ def validate_moe_cuda_graph_support(config) -> None:
     )
 
 
+PACKED_DSA_CP_CUDA_GRAPH_ERROR = (
+    "CUDA graph capture is not supported for this packed DSA context-parallel "
+    "configuration: the requested capture path cannot preserve or reconstruct the "
+    "per-sequence CP layout during warmup and replay. Full CUDA Graph support for "
+    "packed DSA+CP is deferred; set cuda_graph_impl='none'. Other unguarded capture "
+    "combinations remain experimental unless separately validated."
+)
+
+
+def is_packed_dsa_cp_cuda_graph_capture_unsupported(
+    *,
+    experimental_attention_variant: Optional[str],
+    dsa_kernel_backend: str,
+    sequence_packing_scheduler: Optional[str],
+    dynamic_context_parallel: bool,
+    context_parallel_size: int,
+    cuda_graph_impl: str,
+    cuda_graph_modules: Optional[List[CudaGraphModule]],
+    inference_cuda_graph_scope: InferenceCudaGraphScope,
+) -> bool:
+    """Return whether a packed DSA+CP graph combination is proven unsupported.
+
+    Keep this predicate limited to configurations covered by GPU evidence. The
+    static validation matrix exercised configured CP=2 and CP=4; the dynamic
+    matrix exercised configured CP=2 only. Combined local attention+MLP,
+    attention+Mamba, and TE attention+MLP scopes were measured only at static
+    CP=2. CP=3 failed during model construction before eager preflight or graph
+    capture, so it is not graph-specific evidence and remains outside this
+    guard. Dynamic CP with configured CP=1 is likewise not inferred from the
+    DP x CP pool because that neighboring topology did not reach capture.
+    Unmeasured MoE/Mamba partial scopes and full-iteration capture also remain
+    available until they are measured.
+    """
+
+    static_packing = sequence_packing_scheduler == "dp_balanced" and not dynamic_context_parallel
+    dynamic_packing = (
+        sequence_packing_scheduler == "default_dynamic_cp" and dynamic_context_parallel
+    )
+    if experimental_attention_variant != "dsa":
+        return False
+    if not (static_packing or dynamic_packing):
+        return False
+
+    # Capture scope execution uses membership checks, so repeated spellings such
+    # as ``attn,attn`` are the same effective scope as ``attn``.
+    modules = frozenset(cuda_graph_modules or [])
+    if dsa_kernel_backend == "cudnn":
+        if cuda_graph_impl == "local":
+            measured_scopes = set()
+            if static_packing and context_parallel_size in (2, 4):
+                measured_scopes.update(
+                    {
+                        frozenset(),
+                        frozenset({CudaGraphModule.attn}),
+                        frozenset({CudaGraphModule.mlp}),
+                    }
+                )
+                if context_parallel_size == 2:
+                    measured_scopes.update(
+                        {
+                            frozenset({CudaGraphModule.attn, CudaGraphModule.mlp}),
+                            frozenset({CudaGraphModule.attn, CudaGraphModule.mamba}),
+                        }
+                    )
+            elif dynamic_packing and context_parallel_size == 2:
+                measured_scopes.update(
+                    {
+                        frozenset(),
+                        frozenset({CudaGraphModule.attn}),
+                        frozenset({CudaGraphModule.mlp}),
+                    }
+                )
+
+            if modules not in measured_scopes:
+                return False
+            if not modules:
+                # ``block`` owns an inference-only TransformerBlock graph and
+                # skips the per-layer whole-scope manager exercised by the
+                # training validation matrix.
+                return inference_cuda_graph_scope == InferenceCudaGraphScope.layer
+            return True
+        if cuda_graph_impl == "transformer_engine":
+            measured_scopes = set()
+            if static_packing and context_parallel_size in (2, 4):
+                measured_scopes.update({frozenset(), frozenset({CudaGraphModule.attn})})
+                if context_parallel_size == 2:
+                    measured_scopes.add(frozenset({CudaGraphModule.attn, CudaGraphModule.mlp}))
+            elif dynamic_packing and context_parallel_size == 2:
+                measured_scopes.update({frozenset(), frozenset({CudaGraphModule.attn})})
+            return modules in measured_scopes
+
+    # The unfused scorer was measured only for static attention-only local
+    # capture. TileLang was unavailable in the validation image, and the
+    # remaining backend-none scopes were not measured, so keep them available.
+    if (
+        dsa_kernel_backend == "none"
+        and static_packing
+        and context_parallel_size == 2
+        and cuda_graph_impl == "local"
+    ):
+        return modules == {CudaGraphModule.attn}
+    return False
+
+
 def normalize_cuda_graph_modules(
     scopes: Optional[Union[str, CudaGraphModule, List[Union[str, CudaGraphModule]]]],
 ) -> Tuple[List[CudaGraphModule], List[Tuple[str, str, object]], bool]:
--- a/megatron/core/transformer/cuda_graph_config.py
+++ b/megatron/core/transformer/cuda_graph_config.py
@@ -59,108 +59,36 @@ def validate_moe_cuda_graph_support(config) -> None:
     )
 
 
-PACKED_DSA_CP_CUDA_GRAPH_ERROR = (
-    "CUDA graph capture is not supported for this packed DSA context-parallel "
-    "configuration: the requested capture path cannot preserve or reconstruct the "
-    "per-sequence CP layout during warmup and replay. Full CUDA Graph support for "
-    "packed DSA+CP is deferred; set cuda_graph_impl='none'. Other unguarded capture "
-    "combinations remain experimental unless separately validated."
+PACKED_DSA_CP_CUDA_GRAPH_WARNING = (
+    "CUDA graph capture for packed DSA with context parallelism is experimental and not "
+    "fully supported. Some capture paths may fail during warmup or replay; set "
+    "cuda_graph_impl='none' to use the supported eager path."
 )
 
 
-def is_packed_dsa_cp_cuda_graph_capture_unsupported(
+def is_packed_dsa_cp_cuda_graph_experimental(
     *,
     experimental_attention_variant: Optional[str],
-    dsa_kernel_backend: str,
     sequence_packing_scheduler: Optional[str],
     dynamic_context_parallel: bool,
     context_parallel_size: int,
     cuda_graph_impl: str,
-    cuda_graph_modules: Optional[List[CudaGraphModule]],
-    inference_cuda_graph_scope: InferenceCudaGraphScope,
 ) -> bool:
-    """Return whether a packed DSA+CP graph combination is proven unsupported.
+    """Whether packed DSA context parallelism is combined with CUDA graphs."""
 
-    Keep this predicate limited to configurations covered by GPU evidence. The
-    static validation matrix exercised configured CP=2 and CP=4; the dynamic
-    matrix exercised configured CP=2 only. Combined local attention+MLP,
-    attention+Mamba, and TE attention+MLP scopes were measured only at static
-    CP=2. CP=3 failed during model construction before eager preflight or graph
-    capture, so it is not graph-specific evidence and remains outside this
-    guard. Dynamic CP with configured CP=1 is likewise not inferred from the
-    DP x CP pool because that neighboring topology did not reach capture.
-    Unmeasured MoE/Mamba partial scopes and full-iteration capture also remain
-    available until they are measured.
-    """
-
-    static_packing = sequence_packing_scheduler == "dp_balanced" and not dynamic_context_parallel
+    static_packing = (
+        sequence_packing_scheduler == "dp_balanced"
+        and not dynamic_context_parallel
+        and context_parallel_size > 1
+    )
     dynamic_packing = (
         sequence_packing_scheduler == "default_dynamic_cp" and dynamic_context_parallel
     )
-    if experimental_attention_variant != "dsa":
-        return False
-    if not (static_packing or dynamic_packing):
-        return False
-
-    # Capture scope execution uses membership checks, so repeated spellings such
-    # as ``attn,attn`` are the same effective scope as ``attn``.
-    modules = frozenset(cuda_graph_modules or [])
-    if dsa_kernel_backend == "cudnn":
-        if cuda_graph_impl == "local":
-            measured_scopes = set()
-            if static_packing and context_parallel_size in (2, 4):
-                measured_scopes.update(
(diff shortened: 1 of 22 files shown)


megatron/core/transformer/experimental_attention_variant/dsa_cudnn_kernels.py
```diff
@@ -415,6 +415,36 @@ def _causal_seq_lens(q_positions: Tensor, ratio: int, sk: int) -> Tensor:
...
+    branch, which reaches the fused single-kernel ``indexer_forward_wrapper``
+    instead of the per-head scoring loop in
+    :func:`_compute_indexer_scores_chunk_with_global_rows`.
+
+    The flag carries the mask builder's structural knowledge, so we avoid the
+    per-forward host/device sync that a ``torch.equal`` bounds comparison would
+    force on this common path. Genuine packed/CP/custom-position varlen is left
+    untouched (``varlen_is_plain_causal`` is False there).
+    """
+    if cp_size == 1 and packed_seq_params is None and varlen_is_plain_causal:
```
buptzyb, line 490: Can we let it reach the fused scorer when `packed_seq_params` is not None?

jingqiny-99: Yes. The current head routes packed THD independently of CP size: single-sequence packs use the packed-segment scorer and multi-sequence packs use the general THD scorer, both fused when their metadata preconditions hold. `_normalize_plain_causal_bounds` remains limited to the non-packed case because the packed paths must retain cu-seqlens instead of dropping their bounds. This is covered by the scoring-plan matrix for CP=1/2/16 and real CP=1 multi-sequence forward/backward parity against an independent PyTorch reference.

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -2215,15 +2310,45 @@ def forward(
...
+                        (
+                            "positions_and_reorder",
+                            cp_size,
+                            cp_rank,
+                            local_len,
+                            local_len,
+                            local_len * cp_size,
+                            False,
+                            False,
+                        ),
```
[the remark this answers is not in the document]
jingqiny-99: They are conflict branches, not memo-key mismatch.

megatron/core/transformer/experimental_attention_variant/dsa_scoring_plan.py
```diff
@@ -0,0 +1,195 @@
...
+from enum import Enum
+from typing import Optional
+
+import torch
+
+
+class IndexerScoringPlan(Enum):
+    """How indexer scores and top-k are produced for one attention call."""
+
+    PLAIN_CAUSAL = "plain_causal"
```
buptzyb: Why "PLAIN_xxx" instead of "UNPACKED_xxx", to align with the following "PACKED_xxx"?

megatron/core/transformer/experimental_attention_variant/dsa_scoring_plan.py
```diff
@@ -0,0 +1,195 @@
...
+    PACKED_SINGLE_SEGMENTS = "packed_single_segments"
+    """One sequence per pack, scored as one or more causal query/key segments."""
+
+    PACKED_THD_GENERAL = "packed_thd_general"
+    """General packed THD scoring driven by query/key cu_seqlens metadata."""
+
+    UNFUSED_BOUNDS = "unfused_bounds"
+    """No fused kernel claims this layout; scoring falls back to the per-head loop."""
+
+    DECLINE = "decline"
```
buptzyb, line 45: Are we planning to eliminate unfsed bounds & decline cases gradually in the future?

megatron/core/transformer/experimental_attention_variant/dsa_kernels.py
```diff
@@ -74,6 +76,74 @@ def _log_declined_hook(config: TransformerConfig, hook_name: str, reason: str) -
     )
 
 
+def _hook_kwargs_accepting(fn, *, opaque_fallback_names: Tuple[str, ...] = (), **candidate_kwargs):
```
buptzyb: I think this is a very general helper. Should we move this to some files like utils.py?

megatron/core/transformer/experimental_attention_variant/dsa_layout.py
```diff
@@ -357,6 +452,161 @@ def build_packed_allgather_cp_query_positions_and_key_reorder(
...
+            )
+        half = seq_len // cp_size // 2
+        if half <= 0:
+            continue
+        spans.append((seq_start + cp_rank * half, half))
+        spans.append((seq_end - (cp_rank + 1) * half, half))
+    return spans
+
+
+def _host_to_device(values: torch.Tensor, device: torch.device) -> torch.Tensor:
```
buptzyb, line 486: Don't we even have a general h2d/d2h helper function in mcore as of 2026.9?

megatron/core/transformer/cuda_graph_config.py
```diff
@@ -59,6 +59,110 @@ def validate_moe_cuda_graph_support(config) -> None:
...
+
+            if modules not in measured_scopes:
+                return False
+            if not modules:
+                # ``block`` owns an inference-only TransformerBlock graph and
+                # skips the per-layer whole-scope manager exercised by the
+                # training validation matrix.
+                return inference_cuda_graph_scope == InferenceCudaGraphScope.layer
+            return True
+        if cuda_graph_impl == "transformer_engine":
```
buptzyb: elif

megatron/core/transformer/cuda_graph_config.py
```diff
@@ -59,6 +59,110 @@ def validate_moe_cuda_graph_support(config) -> None:
...
+    )
+    if experimental_attention_variant != "dsa":
+        return False
+    if not (static_packing or dynamic_packing):
+        return False
+
+    # Capture scope execution uses membership checks, so repeated spellings such
+    # as ``attn,attn`` are the same effective scope as ``attn``.
+    modules = frozenset(cuda_graph_modules or [])
+    if dsa_kernel_backend == "cudnn":
```
buptzyb: Are tilelang/none backends verified supported, or just not considered?

megatron/core/transformer/experimental_attention_variant/dsa_scoring_plan.py
jingqiny-99: In my opinion, yes.

megatron/core/transformer/experimental_attention_variant/dsa_kernels.py
jingqiny-99: Agree, and I'll move it immediately.

megatron/core/transformer/cuda_graph_config.py
jingqiny-99: OK!

megatron/core/transformer/cuda_graph_config.py
jingqiny-99: Tilelang and none have not been tested before, and I can test them now.

megatron/core/transformer/experimental_attention_variant/dsa_layout.py
jingqiny-99: I'm not sure, but I tried to find similar helpers by matching the name or asking agents like Claude or Codex, and found nothing T_T

buptzyb: So, may we also move this helper to somewhere general?

megatron/core/transformer/experimental_attention_variant/dsa_scoring_plan.py
buptzyb: So should we document what known cases would go to unfused bounds and decline for now?

megatron/core/transformer/experimental_attention_variant/dsa_layout.py
jingqiny-99: Agree. I will move it to a general place.

megatron/core/transformer/experimental_attention_variant/dsa_scoring_plan.py
jingqiny-99: OK! I'll summarize and document these cases.

megatron/core/utils.py
```diff
@@ -691,6 +762,11 @@ def get_model_config(model):
     return get_attr_wrapped_model(model, "config", allow_none=False)
 
 
+def move_host_tensor_to_device(values: torch.Tensor, device: torch.device) -> torch.Tensor:
```
buptzyb: Should we merge this helper with `maybe_move_tensor_to_cpu`? https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/moe/moe_utils.py#L1238

I think `maybe_move_tensor_to_cpu` is in a bad place. Moving it here would be better.

BTW looks like `maybe_move_tensor_to_cpu` is using `non_blocking` always. We may need a new argument `synchronization: bool = True`, which determines whether we add a stream sync after the non-blocking copy.

Ignore me if you think this is a bad idea.

megatron/core/transformer/transformer_layer.py
```diff
@@ -371,7 +371,9 @@ def __init__(
         )
 
         attention_optional_kwargs = {}
-        if config.context_parallel_size > 1 and config.cp_comm_type is not None:
+        if (
+            config.context_parallel_size > 1 or config.dynamic_context_parallel
```
buptzyb, line 375: I saw somewhere in `transformer_config.py` still reads `if config.context_parallel_size > 1: xxxx`. Should we align them to `if config.context_parallel_size > 1 or config.dynamic_context_parallel: xxxx` as well?

megatron/core/transformer/experimental_attention_variant/dsa_cudnn_kernels.py
```diff
@@ -2130,8 +2504,8 @@ def forward(
         sparse_loss: bool,
         calculate_per_token_loss: bool,
         d_v: int,
-        varlen_starts: Optional[Tensor],
-        varlen_ends: Optional[Tensor],
+        varlen_starts: Tensor,
```
buptzyb: Not optional anymore?

jingqiny-99: Sorry for this mistake... I have reverted this change

megatron/core/transformer/transformer_layer.py
jingqiny-99: Agree, I updated other checks to also cover dynamic_context_parallel.

megatron/core/utils.py
jingqiny-99: I think you are right, the `maybe_move_tensor_to_cpu` is in a strange place, I think it's good to move it here, and merge them in the future, because in my opinion, it may become a large helper (move tensor to arbitrary device and maybe do some synchronization if need), and it isn't transparent to its callers.

megatron/core/utils.py
```diff
@@ -691,6 +762,36 @@ def get_model_config(model):
...
+    """Move a host tensor to ``device``."""
+    return values.to(device, non_blocking=device.type == "cuda")
```
buptzyb, line 782: Do we need `values.pin_memory().to(device)` to avoid pageable H2D?

jingqiny-99: I think it is right, but in my test case, removing `pin_memory` will save copy time. So I will add an argument here to decide whether run `values.pin_memory()` here, and its default value will be False.

buptzyb: Is it because your `values` are already on pin? Non-pin h2d is dangerous since pagable copy is blocking and non-graphable IIRC.

jingqiny-99: soga..Thanks for teaching me, and sorry for my limited knowledge. I will pin the memory by default.

buptzyb: Happen to read an article about a hang issue caused by pagable memcpy: https://mp.weixin.qq.com/s/uOJALDKb588mFynaeJQJoA

buptzyb (approved):
I have to admit I cannot understand every single detail of this PR. But the code organization and functionality generally LGTM. One concern is the main branch is also making their refactoring to the dsa code (like #7211), so the main-dev divergency would be larger and larger. Beg @hxbai for a final review, thanks!

yaox12 (approved):

PR #6207, opened 2026-08-03
[fix] Give same-key GTP chain neighbours distinct gather buffers

### LINKED ISSUE #6046: [feat]GTP schedule optimization for MoE layer: one-block-ahead prefetch

When GTP/EGTP weight remat is enabled and the shard group spans nodes — the large-model
configuration it exists for — the routed-expert weights (`linear_fc1`/`linear_fc2`)
dominate the bytes re-gathered per MoE block: `num_experts × 2` weights per layer, versus
a handful for attention. The prefetch is only **one step ahead within a single chain**, so each
of **those all-gathers has just the preceding GEMM to hide behind, leaving inter-node transfer
exposed in every MoE block**.

Image

Image

One-block-ahead prefetch for grouped-expert weights:

- **Give fc1 and fc2 their own homogeneous chains**, so `next_w` links the same weight role
  across consecutive MoE layers — the gather gets a full MoE block of compute as runway.
- Keep them as two chains but one shared IB stream, so the all-gathers serialize instead
  of splitting bandwidth.
- **Double-buffer via a chain-position parity in the weight-cache key**: layers N and N+1 are
  live at once, so they must alternate between two buffers.
- **Eager path only**; when MoE is CUDA-graph captured, grouped weights stay in the GRAPHED
  chain so the cross-graph drain still finds them.

- Merging fc1+fc2 into one chain — leaves fc2 a GEMM behind.
- A separate IB stream per role — parallel gathers just split bandwidth.
- Deeper prefetch (N+2) — needs ≥3 buffers per key; one block is already enough runway.

**Additional context**

Related MR: https://github.com/NVIDIA/Megatron-LM/pull/6057

/

fanshiqing:

# What does this PR do?
Give same-key GTP chain neighbours distinct gather buffers to avoid potential data-race.

One-step-ahead prefetch keeps two chain-adjacent weights live at once, so they must not share a cache key. Under CUDA graphs the GTP_ungraphed chain collapses to `embedding + output_layer`, **which are same-shaped: embedding's consume gathers output_layer into the shared buffer while F.embedding is still reading it**.

```
[GTP_ungraphed chain]             
node_id | layer_id | dtype                | shape                | curr_weight_name                                                       | prev_weight_name
--------+----------+----------------------+----------------------+------------------------------------------------------------------------+-----------------------------------------------------------------------
      0 |        - | torch.bfloat16       | (131072, 8192)       | module.module.embedding.word_embeddings.weight                         | -
      1 |        - | torch.bfloat16       | (131072, 8192)       | module.module.output_layer.weight                                      | module.module.embedding.word_embeddings.weight
 ```

**Fix generally** rather than for that pair: `_ensure_distinct_buffer_from_prev` gives any weight its own buffer when its chain predecessor would claim the same one, so any chain a CUDA-graph partition leaves homogeneous is covered.

### Pre-checks

### DIFF
--- a/docs/api-guide/core/generalized_tensor_parallel.md
+++ b/docs/api-guide/core/generalized_tensor_parallel.md
@@ -342,7 +342,7 @@ The figure visualizes the per-class split from the list above: green = resolves 
 
 Two distinct pools with explicit lifecycle rules:
 
-- **`GTPWeightCache`** (AG/RS output buffers) — ticket-based, keyed on `(shape, dtype, fwd, expert_idx, reduce_scatter)`. Same-shape buffers across layers are shared. Tickets persistent; buffer allocated lazily on first `get()`; addresses stable across iterations for CG replay.
+- **`GTPWeightCache`** (AG/RS output buffers) — ticket-based, keyed on `(shape, dtype, fwd, expert_idx, reduce_scatter)`. Same-shape buffers across layers are shared, **except between chain neighbours** — one-step-ahead keeps `prev_w` and the current weight live at once, so `_ensure_distinct_buffer_from_prev` folds a parity bit into the key when the two would collide, at the cost of one extra buffer for the second of the pair. Normally inert (neighbours are different weight roles, hence different shapes); it fires when CG capture leaves two same-shaped weights adjacent — embedding + output_layer alone in the `UNGRAPHED` chain. Tickets persistent; buffer allocated lazily on first `get()`; addresses stable across iterations for CG replay.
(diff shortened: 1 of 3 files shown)


kvareddy (approved):

JF-D:
LGTM. This fix looks quite general. It seems we should add an optimization later that links the embedding and output layer to the expert ungraphed chain? This should improve overlap and reduce memory usage by allowing them to share the same large embedding buffer.

JF-D (approved):

Phlip79 (approved):
Thank you for updating the doc!

PR #6210, opened 2026-08-03
Apply black 26.3.0 formatting to four files it now reformats

deepakn94:
## Summary

#6180 bumped `black` from 24.4.2 to 26.3.0 but did not reformat the tree. Black 26 strips redundant parentheses around assignment targets and re-joins calls that black 24 had exploded, so files predating the bump no longer satisfy `black --check`.

`tools/autoformat.sh` lints only the files that differ from `main`, so a PR hits the violation only if it happens to touch such a file. That keeps the debt invisible here — and surfaces it all at once in downstream forks, whose diff against `main` is their entire delta. That is how these four surfaced.

This is purely the output of running the pinned black over them. No behavioural change.

- `megatron/core/models/vision/radio.py` — `(pos_H, pos_W) = ...` → `pos_H, pos_W = ...`
- `megatron/core/transformer/attention.py` — same paren-stripping on tuple unpacking
- `megatron/core/transformer/mlp.py` — re-join the `warnings.warn(...)` call
- `megatron/core/transformer/transformer_layer.py` — re-join the `assert ..., "EP overlap..."`

## Scope

117 files under `megatron/` are affected repo-wide. This PR is deliberately limited to the four that are blocking, to keep the diff reviewable and avoid conflicting with work in flight. Happy to widen it to the full set if you would rather clear the backlog in one pass — or to close this if the intent is to let files get reformatted as they are touched.

## Test plan

- [x] `black==26.3.0 --skip-magic-trailing-comma --skip-string-normalization --check` passes on all four
- [x] `isort==5.13.2`, `ruff~=0.9.0`, `pylint==3.2.6` clean on all four (pylint 10.00/10)
- [x] `python -m py_compile` succeeds on all four
- [x] CI status checks pass

### DIFF
--- a/megatron/core/models/vision/radio.py
+++ b/megatron/core/models/vision/radio.py
@@ -610,11 +610,11 @@ class RADIOViTModel(VisionModule):
             return pos_embed
 
         def aspect_ratio_select(pos_embed):
-            (pos_H, pos_W) = pos_embed.shape[-2:]
-            (input_H, input_W) = input_dims
+            pos_H, pos_W = pos_embed.shape[-2:]
+            input_H, input_W = input_dims
             if input_H == input_W:
                 return pos_embed
-            (crop_H, crop_W) = (pos_H, pos_W)
+            crop_H, crop_W = (pos_H, pos_W)
             if input_W < input_H:
                 crop_W = min(pos_W, math.ceil(pos_W * (input_W / input_H)))
             else:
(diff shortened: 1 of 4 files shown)


chtruong814 (approved):

santhnm2 (approved):

jaredcasper (approved):

yaoyu-33 (approved):

yaoyu-33 (approved):

deepakn94:
/ok to test [0581618](https://github.com/NVIDIA/Megatron-LM/pull/6210/commits/0581618f23fd6f4df9e038d8bae5c273baf96148)

PR #6214, opened 2026-08-03
Derive MFSDP v2 checkpoint chunk metadata from the parameter layout

ahmadki:
- [x] I, the PR author, have personally reviewed every line of this PR.

# What does this PR do?

Replace the generic uneven-DTensor helper in the experimental Megatron-FSDP DCP path with an analytic, collective-free equivalent. This answers the "we may need to rewrite it from first principles for just what v2 needs" comment on #6024.

The generic helper recovers each shard's offset with one `all_gather_object` per DTensor. This path does not need to ask: `GlobalLayout` already records every parameter's global element offset, and its `DBuffer` records this rank's owned element range, so each shard's chunk follows by arithmetic.

This adds an experimental `uneven_dtensor` module that derives each parameter's chunk from its parameter group's main-weight buffer, keys it by the FQN torch's DCP state-dict helpers use, and attaches DCP's `__create_chunk_list__` / `__create_write_items__` hooks to the model and optimizer state dicts in one pass. Per-parameter optimizer state is allocated as `zeros_like` of a sharded parameter, so it shares that parameter's chunk. Tied parameters are keyed per FQN, since the state dict carries an entry for each name. The stable Megatron-FSDP helper is untouched, so v1 is unaffected.

The described chunks are unchanged, including the offsets reported for empty shards, so checkpoints stay byte-identical to the ones the generic helper produces. This is a behavior-preserving refactor that removes one collective per DTensor from both save and load.

## Dependencies

Stacked on #6024, which adds the experimental DCP path this replaces the metadata step of.

## Testing

New `tests/unit_tests/distributed/mfsdp_v2/test_uneven_dtensor.py`: the analytic metadata equals what the gather-based helper computes, computing it issues no collective (the collectives are monkeypatched to raise), and every parameter's chunks tile its global shape exactly once. Runs on 1 and 2 ranks.

Byte-identity was also checked out-of-tree on 2× RTX 6000 Ada by saving the same trained model and optimizer through both paths and comparing the per-rank `.distcp` payloads by sha256 plus the `.metadata` `state_dict_metadata` / `planner_data` / `storage_data`, for fp32 and for bf16 with fp32 master weights: identical. A DP=2 → DP=1 reshard load was verified bit-exact.

## Issue tracking

Linked issue: Related to #5534

### Pre-checks

- [x] I have added relevant unit tests
- [ ] I have added relevant functional tests 
- [x] I have added proper typing to my code
- [x] I have added relevant documentation
- [x] I have run the autoformatter.sh on my PR

PR #6215, opened 2026-08-03
Support DCP checkpointing for the Megatron-FSDP v2 optimizer

ahmadki:

# What does this PR do?

Fill `FullyShardedOptimizer.sharded_state_dict` and `load_state_dict` (previously `NotImplementedError`) so a Megatron-FSDP v2 optimizer saves and loads its state as a PyTorch Distributed Checkpoint in the `fsdp_dtensor` format.

The layout mirrors `DistributedOptimizer.sharded_param_state_fsdp_dtensor`, so v1 and v2 write the same on-disk format:

```python
{"state": {fqn: {"exp_avg": DTensor, "exp_avg_sq": DTensor}},
 "param_to_group_meta": {fqn: {...group hyperparameters...}}}
```

That is also why the parameter names come from MCore's `get_global_unique_param_name` rather than from torch's `get_optimizer_state_dict`, whose names are the plain `named_parameters` ones rather than MCore's PP/EP-unique checkpoint names. The base optimizer (TE FusedAdam) tracks `step` per param group rather than per parameter, so `step` round-trips inside `param_to_group_meta`.

**Empty-local shards are the interesting part.** Megatron-FSDP v2 filters parameters whose *local* shard is empty out of the base optimizer, per rank (a TE FusedAdam workaround, [TransformerEngine#3207](https://github.com/NVIDIA/TransformerEngine/issues/3207)), so the surviving state and param groups differ across ranks. Meanwhile `preprocess_state_dict_for_uneven_dtensor` runs one `all_gather_object` *per DTensor* in sorted key order, so a rank-dependent DTensor keyspace deadlocks. `sharded_state_dict` therefore emits every trainable parameter: real state where the optimizer holds it, and empty-local `zeros_like` placeholders otherwise. A placeholder has the parameter's global shape and dtype but no local rows, so it contributes no data — the rank that owns a non-empty shard covers the tensor fully — and only keeps the keyspace identical everywhere. One `all_gather_object` over the sharded mesh dimensions distributes the per-parameter state keys so each placeholder carries exactly the owning rank's key set.

`param_to_group_meta` stays rank-local on purpose: it covers exactly the parameters this rank's optimizer owns, which is what both ends need. On save DCP unions the non-tensor entries written by all ranks, and on load each rank reads back the entries for the parameters it owns — including after a reshard, where a rank asks for parameters another rank saved.

`state_dict()` keeps raising: Megatron-FSDP always runs with `--ckpt-format fsdp_dtensor` (enforced in `validate_args`), so the plain accessor is unreachable, and forwarding the wrapped optimizer's state would only produce a checkpoint that silently drops the cross-rank resharding metadata.

Materializing the optimizer state before a load reuses `init_optimizer_state` from #6024 rather than carrying a second implementation, so that helper becomes public.

## Dependencies

This is stacked on #6024, which adds `megatron_fsdp/experimental/checkpoint.py`, whose `init_optimizer_state` this PR reuses (and makes public). Review only the `Support DCP checkpointing for the Megatron-FSDP v2 optimizer` commit; the rest of the branch is #6024.

## Testing

New `tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py` trains a source model, checkpoints the model and the optimizer the way the training loop does (`state_dict_for_save_checkpoint` + `sharded_state_dict` + the uneven-DTensor preprocessing), and loads into a zero-initialized model that has been trained a *different* number of steps, so a dropped or mis-serialized `step` — and the stale momentum that comes with it — fails the comparison instead of matching by construction. The optimizer state, the param groups (including `step`), and the model weights are all asserted bit-exact, at zero tolerance, including each DTensor's global shape, placements and device mesh rather than only this rank's local shard.

The test runs on a single rank too, where it degenerates to a single-shard sanity check. With two or more ranks it additionally asserts that some rank really had an empty-local shard, so the placeholder path is not covered vacuously.

Verified on 2× RTX 6000 Ada: passes on 2 ranks and on 1 rank, together with #6024's `test_checkpoint.py` and the neighbouring `test_mcore_adapter.py` / `test_optimizer.py` / `test_dbuffer.py` buckets.

## Issue tracking

Linked issue: Related to #5534

### Pre-checks


wujingyue (commented):
I'm sure it's a good PR! 

Should I wait for #6024 to be merged or could you change the diffbase to #6024 so I can review the delta? 

FYI, https://github.com/NVIDIA/Megatron-LM/blob/df2da78b39442a7642d15cc8c678d11353f20501/skills/mcore-split-pr/SKILL.md?plain=1#L84

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
```diff
@@ -63,7 +63,7 @@ def _sync_model_weight_from_main_weight(model: torch.nn.Module) -> None:
                 parameter_group.sync_model_weight_from_main_weight()
 
 
-def _init_optimizer_state(optimizer: torch.optim.Optimizer) -> None:
+def init_optimizer_state(optimizer: torch.optim.Optimizer) -> None:
```
wujingyue: How about moving this to optimizer.py? I understand it was added **for** checkpointing but the implementation depends on only optimizer.

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -55,6 +59,8 @@ def __init__(
             if self.ddp_config != model_chunk.ddp_config:
                 raise ValueError("All MFSDP v2 model chunks must share the same ddp_config.")
         self.is_stub_optimizer = optimizer is None
+        # Lazily built by _param_fqn.
+        self._param_to_fqn: dict[torch.nn.Parameter, str] | None = None
```
wujingyue, line 66: Will this work for a tied parameter, which is mapped to multiple fqns?

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -55,6 +59,8 @@ def __init__(
             if self.ddp_config != model_chunk.ddp_config:
                 raise ValueError("All MFSDP v2 model chunks must share the same ddp_config.")
         self.is_stub_optimizer = optimizer is None
+        # Lazily built by _param_fqn.
```
wujingyue, line 65: Why is it built lazily versus just there in the constructor?

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -86,24 +92,212 @@ def _validate_config(config: OptimizerConfig, model_chunks: List[MegatronModule]
...
+                ]
+        return []
+
+    def _gather_state_keys_by_fqn(self) -> dict[str, list[str]]:
+        """Map every parameter's FQN to the keys of its ``DTensor`` optimizer state entries.
+
+        The keys are read off live state (``exp_avg``/``exp_avg_sq`` for Adam) rather than
+        hard-coded, so any base optimizer works. They are gathered because a parameter that
+        this rank's optimizer filtered out is only described by the rank that owns a
+        non-empty shard of it, which is what lets every rank synthesize a placeholder with
```
wujingyue, line 177: > non-empty

still needs this filter given https://github.com/NVIDIA/TransformerEngine/issues/3207 is fixed? 

It appears that we are doing extra all_gather_object during checkpointing just to work around the optimizer-dislikes-empty-parameter bug?

wujingyue (commented):
Thanks for the PR! LGTM overall. I'll do another pass after I understand the restriction around empty parameters.

megatron/core/optimizer/fully_sharded_optimizer.py
ahmadki: Yes.
I also added `test_tied_parameter_roundtrip`. two Linears sharing a weight, sharded with MFSDP v2, checkpointed and restored.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/checkpoint.py
ahmadki: done

megatron/core/optimizer/fully_sharded_optimizer.py
ahmadki: `get_global_unique_param_name` scans every model chunk's `named_parameters()` per call and for a parameter inside a `TransformerLayer`, walks `model.modules()` again to recover the PP global layer number. Building the whole map is quadratic to the parameter count, and wasteful for a job that never reaches a checkpoint.
`DistributedOptimizer._param_name` builds its `param_to_name` the same way, on first use (`if not hasattr(self, "param_to_name")`, in distrib_optimizer.py)

If you'd still prefer it eager, it's a two-line change.

megatron/core/optimizer/fully_sharded_optimizer.py
ahmadki: Yes, the placeholders and `all_gather_object` are only because of the empty-parameter filter.

But the fix is not everywhere (not in LTS), which is fair to exclude for v2. But `tests/test_utils/recipes/h100/unit-tests.yaml` runs `tests/unit_tests/distributed/mfsdp_v2/**` under `environment: [lts, dev]`, so dropping the filter now would silently stop updating some shards on LTS.

Either ways, the filter lives in (`megatron/core/optimizer/__init__.py`), not in `FullyShardedOptimizer`. Removing it is a change to optimizer construction and is not something I want to do in a checkpoint PR.

megatron/core/optimizer/fully_sharded_optimizer.py
wujingyue: But dict can store only **one** str per parameter. What am I missing?

megatron/core/optimizer/fully_sharded_optimizer.py
wujingyue: > wasteful for a job that never reaches a checkpoint.

Got it!

> If you'd still prefer it eager, it's a two-line change.

I still prefer eager because I'd like to optimize for simplicity by default. Let's revisit when the building is indeed slow. My gut feeling is that for small runs the quadric algorithm is fast enough and for large production runs models need to be checkpointed anyway...

megatron/core/optimizer/fully_sharded_optimizer.py
wujingyue: > Yes, the placeholders and all_gather_object are only because of the empty-parameter filter.

Got it. That's good -- all_gather_object and code around it are transient hacks that we can get rid of once 3207 lands everywhere.

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -89,24 +95,228 @@ def _validate_config(config: OptimizerConfig, model_chunks: List[MegatronModule]
...
-        """Load optimizer state."""
-        raise NotImplementedError("MFSDP v2 optimizer checkpointing is not yet supported.")
+    def _param_fqn(self, param: torch.nn.Parameter) -> str:
+        """Return a model parameter's globally unique checkpoint name.
+
+        ``get_global_unique_param_name`` rescans every model chunk on each call and the
+        checkpoint path resolves each name several times, so the mapping is built once and
+        cached, mirroring :class:`DistributedOptimizer`'s ``param_to_name``.
+        """
+        if self._param_to_fqn is None:
```
wujingyue, line 116: Nit: If we insist on lazy build, we could cache one key at a time for simplicity and potentially better latecy. Something like:

```
if param in self._param_to_fqn:
    return self._param_to_fqn[param]
self._param_to_fqn[param] = ...
return ...
```

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -89,24 +95,228 @@ def _validate_config(config: OptimizerConfig, model_chunks: List[MegatronModule]
...
+        collectives do.
+
+        Raises:
+            NotImplementedError: If the parameters do not all share one device mesh, which
+                is what expert parallelism produces.
+        """
+        sharded_parameters = [
+            param for param in self._trainable_parameters() if isinstance(param, DTensor)
+        ]
+        if len({param.device_mesh for param in sharded_parameters}) > 1:
```
wujingyue, line 152: Isn't this already broken by DP mesh and EDP mesh being different? They are both 1D but have different sizes.

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -89,24 +95,228 @@ def _validate_config(config: OptimizerConfig, model_chunks: List[MegatronModule]
...
+
+        The keys are read off live state (``exp_avg``/``exp_avg_sq`` for Adam) rather than
+        hard-coded, so any base optimizer works. They are gathered because a parameter that
+        this rank's optimizer filtered out is only described by the rank that owns a
+        non-empty shard of it, which is what lets every rank synthesize a placeholder with
+        exactly the owning rank's keys. A parameter that has no state anywhere (nothing has
+        stepped yet) is simply absent.
+        """
+        state_keys_by_fqn = {
+            self._param_fqn(param): sorted(
```
wujingyue, line 182: Why sorted?

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -89,24 +95,228 @@ def _validate_config(config: OptimizerConfig, model_chunks: List[MegatronModule]
...
         self,
         model_sharded_state_dict: ShardedStateDict,
         is_loading: bool = False,
         metadata: Optional[dict] = None,
     ) -> ShardedStateDict:
-        """Build a sharded optimizer state dict."""
-        raise NotImplementedError("MFSDP v2 optimizer checkpointing is not yet supported.")
+        """Build the ``fsdp_dtensor`` sharded optimizer state dict.
+
+        The layout mirrors :meth:`DistributedOptimizer.sharded_param_state_fsdp_dtensor`, so
```
wujingyue, line 267: Possible to extract and reuse?

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -89,24 +95,228 @@ def _validate_config(config: OptimizerConfig, model_chunks: List[MegatronModule]
...
+                read from the wrapped optimizer directly.
+            is_loading: Whether the state dict will be filled by a load. If so, the optimizer
+                state is materialized first so the load has DTensors to write into.
+            metadata: Accepted for interface parity; the ``fsdp_dtensor`` format takes no
+                sharding options.
+
+        Returns:
+            The optimizer state dict, in the ``fsdp_dtensor`` format described above.
+        """
+        if is_loading:
```
wujingyue, line 296: Is `is_loading` required by the API? Otherwise, we could let the caller to invoke `init_optimizer_state`, saving one parameter.

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -89,24 +95,228 @@ def _validate_config(config: OptimizerConfig, model_chunks: List[MegatronModule]
...
+        state_keys_by_fqn = {
+            self._param_fqn(param): sorted(
+                key for key, value in state.items() if isinstance(value, DTensor)
+            )
+            for param, state in self.optimizer.state.items()
+        }
+        for group in self._shard_process_groups():
+            gathered = [None] * torch.distributed.get_world_size(group)
+            torch.distributed.all_gather_object(gathered, state_keys_by_fqn, group=group)
+            state_keys_by_fqn = {fqn: keys for rank in gathered for fqn, keys in rank.items()}
```
wujingyue, line 190: When two ranks both have keys for `fqn`, are they guaranteed to be identical?

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -89,24 +95,228 @@ def _validate_config(config: OptimizerConfig, model_chunks: List[MegatronModule]
...
+        }
+
+        packed_state: dict[str, Any] = {}
+        for param in self._trainable_parameters():
+            fqn = self._param_fqn(param)
+            if fqn in state_by_fqn:
+                packed_state[fqn] = state_by_fqn[fqn]
+            else:
+                # Filtered out of this rank's optimizer because its local shard is empty. The
+                # rank owning a non-empty shard saves the real state; this placeholder has the
```
wujingyue, line 312: Can you leave a code comment about removing these filters after https://github.com/NVIDIA/TransformerEngine/pull/3212 is propagated?

wujingyue (commented):
Non-test looks good to me. I'll review the test next!

tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py
```diff
@@ -0,0 +1,482 @@
...
+        bf16=True,
+        params_dtype=torch.bfloat16,
+        # Megatron-FSDP shards the optimizer state itself, so MFSDP v2 rejects MCore's
+        # distributed optimizer on the OptimizerConfig.
+        use_distributed_optimizer=False,
+        clip_grad=0.0,
+    )
+    optimizer = get_megatron_optimizer(optimizer_config, [model], use_gloo_process_groups=False)
+    assert isinstance(optimizer, FullyShardedOptimizer)
+    optimizer.reload_model_params()
```
wujingyue, line 100: Why?

tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py
```diff
@@ -0,0 +1,482 @@
...
+    return model, optimizer
+
+
+def _zero_parameters(module: torch.nn.Module) -> None:
+    """Zero every weight so a correct load has to overwrite them."""
+    with torch.no_grad():
```
wujingyue, line 106: [nn.init.zero](https://docs.pytorch.org/docs/2.13/nn.init.html#torch.nn.init.zeros_)

tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py
```diff
@@ -0,0 +1,482 @@
...
+from megatron.core.transformer.enums import AttnBackend
+from megatron.core.transformer.fsdp_dtensor_checkpoint import get_global_unique_param_name
+from megatron.core.transformer.module import MegatronModule
+from megatron.core.transformer.transformer_block import TransformerBlock
+from megatron.core.transformer.transformer_config import TransformerConfig
+from tests.unit_tests.dist_checkpointing import TempNamedDir
+from tests.unit_tests.test_utilities import Utils
+
+SOURCE_STEPS = 3
+DESTINATION_STEPS = 1
```
wujingyue: Code-comment?

tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py
```diff
@@ -0,0 +1,482 @@
...
+        empty-local shards out of the base optimizer per rank, so the placeholders are what
+        keep the DTensor keyspace identical across ranks.
+        """
+        config = _transformer_config()
+        world_size = torch.distributed.get_world_size()
+
+        source_model, source_optimizer = _build_model_and_optimizer(
+            config, self.pg_collection, zero_init=False
+        )
+        for _ in range(SOURCE_STEPS):
```
wujingyue, line 285: I'm not sure if this should be a **global** variable. How many steps seems to be a decision of each test. 

Bottom line: Optimize for readability, not for conciseness.

tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py
```diff
@@ -0,0 +1,482 @@
...
+        assert {group["step"] for group in param_group_snapshot} == {SOURCE_STEPS}
+
+        if world_size > 1:
+            # Confirm the placeholder path is not vacuous: at least one rank must have had an
+            # empty-local shard filtered out of its optimizer. Count empty local shards
+            # directly, which is the exact condition the filter tests.
+            empty_local = sum(
+                1
+                for param in source_model.parameters()
+                if param.requires_grad
```
wujingyue, line 300: Why?

tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py
```diff
@@ -0,0 +1,482 @@
...
+        if world_size < 2 or world_size % 2:
+            pytest.skip("MFSDP v2 EP requires an even world size of at least two.")
+        Utils.initialize_model_parallel(1, 1, expert_model_parallel_size=2)
+        self.pg_collection = ProcessGroupCollection.use_mpu_process_groups()
+        model_parallel_cuda_manual_seed(1234)
+
+    def teardown_method(self):
+        Utils.destroy_model_parallel()
+
+    def test_sharded_state_dict_rejects_expert_parallelism(self) -> None:
```
wujingyue, line 424: Got it -- I assume this is a TODO rather than rejection by contract? https://github.com/NVIDIA/Megatron-LM/pull/5956

tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py
```diff
@@ -0,0 +1,482 @@
...
+                "model": source_model.state_dict_for_save_checkpoint(),
+                "optimizer": source_optimizer.sharded_state_dict({}),
+            }
+            preprocess_state_dict_for_uneven_dtensor(save_state_dict)
+            dcp.save(save_state_dict, checkpoint_id=ckpt_dir)
+
+            destination_model, destination_optimizer = _build_model_and_optimizer(
+                config, self.pg_collection, zero_init=True
+            )
+            for _ in range(DESTINATION_STEPS):
```
wujingyue, line 322: ditto

tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py
```diff
@@ -0,0 +1,482 @@
...
+                "No rank had an empty local shard, so the placeholder path is untested for "
+                f"this config. empty_local_counts={empty_local_counts}"
+            )
+
+        with TempNamedDir(tmp_path_dist_ckpt / "fsdp_dtensor_optimizer", sync=True) as ckpt_dir:
+            save_state_dict = {
+                "model": source_model.state_dict_for_save_checkpoint(),
+                "optimizer": source_optimizer.sharded_state_dict({}),
+            }
+            preprocess_state_dict_for_uneven_dtensor(save_state_dict)
```
wujingyue, line 316: Are we moving away from this given https://github.com/NVIDIA/Megatron-LM/pull/6214?

wujingyue (commented):
LGTM otherwise

wujingyue (approved):

megatron/core/optimizer/fully_sharded_optimizer.py
ahmadki: Yes, but would prefer to do it in a new PR since it's `DistributedOptimizer` logic

tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py
ahmadki: done

tests/unit_tests/distributed/mfsdp_v2/test_optimizer_checkpoint.py
ahmadki: done

megatron/core/optimizer/fully_sharded_optimizer.py
ahmadki: `get_global_unique_param_name` looks the parameter up in `named_parameters()` which does de-duplication, so it returns the first registered name (stable one on every rank, save and load operation).

`DistributedOptimizer.param_to_name` is built  by inverting `dict(named_parameters())`, so v1 and v2 write a tied weight under the same key.

`test_tied_parameter_roundtrip` checks it end to end with two Linears sharing a weight, the model has both fc1.weight and fc2.weight, the optimizer has only one of the two, and the weight and its momentum come back bit-exact after a save and load.

megatron/core/optimizer/fully_sharded_optimizer.py
ahmadki: #7002 removed the filter, then #7216 restored it for TE < 2.18. I left a TODO on MFSDP v2 dropping TE < 2.18 instead of https://github.com/NVIDIA/TransformerEngine/pull/3212

PR #6222, opened 2026-08-03
Add disaggregated KV state handoff

wdykas:

# What does this PR do?
- Adds prefill-to-decode KV-cache handoff using NIXL or NCCL.
- Handles block ownership, prefix reuse, asynchronous transfers, TP/PP metadata, admission, release, and reset safety.

  ## Design overview

  This change adds native prefill/decode disaggregation to Megatron’s dynamic
  inference engine. The prefill engine computes the prompt, transfers the resulting
  KV state to a decode engine, and the decode engine resumes generation without
  recomputing the prompt.

  Client / Coordinator
          |
          | prompt request
          v
  +-------------------+
  |  Prefill engine   |
  |                   |
  | 1. Run prefill    |
  | 2. Pin KV blocks  |
  | 3. Build handoff  |
  |    metadata       |
  +---------+---------+
            |
            | block IDs, hashes, resume tokens,
            | transfer metadata
            v
  +-------------------+
  |   Decode engine   |
  |                   |
  | 4. Reuse cached   |
  |    prefix         |
  | 5. Allocate the   |
  |    missing suffix |
  | 6. Transfer KV    |
  | 7. Register hashes|
  | 8. Admit directly |
  |    into decode    |
  +---------+---------+
            |
            | transfer complete / release KV
            v
  +-------------------+
  | Prefill releases  |
  | pinned KV blocks  |
  +-------------------+

  ### Request lifecycle

  1. The prefill engine runs the prompt and captures the KV blocks, block-hash chain,
     transfer metadata, and token needed to resume generation.

  2. The prefill KV blocks receive a temporary reference before the request is
     removed:

     Active request owns block:       refcount = 1
     Temporary handoff retain:        refcount = 2
     Prefill request is removed:      refcount = 1
     Decode finishes transfer:        refcount = 0

     This prevents eviction or reuse while the decode engine is reading the blocks.

  3. The decode engine checks its local prefix cache before allocating destination
     blocks. Cached prefix blocks are reused when using the nixl backend(nccl doesn't support this because its a push backed), and only the missing suffix is
     transferred when supported by the backend.

  4. Transfer completion is tracked across the model-parallel ranks using a separate
     handoff control path. This avoids adding model-parallel collectives to the
     inference compute path.  This using zmq to notify when transfer is complete.

  5. Transfer polling does not mutate the active decode batch. A completed import is
     admitted at the normal scheduling boundary, after logits from the previous
     asynchronous forward have been consumed.

  6. The decode engine registers the transferred hashes and starts directly from the
     imported KV state. Precomputed hashes preserve the exact hash chain, including
     partial-tail state that cannot be reconstructed from complete prompt blocks
     alone.

  7. If the transferred token already satisfies the request’s generation limit, EOS
     condition, or stop sequence, the request completes immediately without launching
     another model forward.

  8. Once the decode transfer is complete, the coordinator sends a release
     notification to the prefill engine, which drops the temporary block references.

  ### Capacity and backpressure

  A completed transfer may remain pending if the decode batch is at its request or
  token limit. Its completion notification and allocated destination blocks remain
  owned until admission becomes possible.

  The engine treats an admissible import as schedulable work, ensuring an otherwise-
  idle decode engine enters the scheduling path and does not leave the import pending
  indefinitely.

  ### Backend behavior

  - NIXL: The decode engine initiates a one-sided pull into its selected destination
    blocks.

  - NCCL: Prefill sends and decode receives are matched as a two-sided transfer.
  - Allocation, ownership, cached-prefix matching, completion tracking, and decode
    admission are shared across backends.

  ### Important invariants

  - Prefill blocks remain pinned until decode reports transfer completion.
  - Transferred blocks are not visible through prefix caching until their contents
    are complete.

  - Transfer polling never changes the active batch.
  - A completion notification is consumed only when its import is admitted or failed.
  - Decode never recomputes the transferred prompt.
  - Normal non-disaggregated inference continues to use no-op handoff hooks.


megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -2795,10 +2883,18 @@ async def run_engine(self, *, loop: Optional[asyncio.AbstractEventLoop] = None):
                             and (
                                 self.context.get_active_request_count() > 0
                                 or self.waiting_request_ids
+                                or self.pending_kv_import_count > 0
                             )
                         )
                     )
-                await self.async_step()
+                self._poll_pending_kv_imports()
```
wdykas, line 3001: KV transfer handles only support polling. When a transfer is the only pending work, we sleep for 1 ms to avoid busy-spinning while keeping decode admission responsive. I dont know about a better way.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -2795,10 +2883,18 @@ async def run_engine(self, *, loop: Optional[asyncio.AbstractEventLoop] = None):
...
+                if self.context.get_active_request_count() > 0 or self.waiting_request_ids:
+                    await self.async_step()
+                else:
+                    # Transfer handles expose polling but no async completion
+                    # callback. Yield briefly without adding the normal 20 ms
+                    # idle-loop delay to decode admission.
+                    await asyncio.sleep(0.001)
```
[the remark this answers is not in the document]
nvcsathe: Is there an optimal here for polling frequency?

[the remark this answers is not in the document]
wdykas: This design I am not sure about, I think it should just be the smallest value possible

megatron/core/inference/engines/dynamic_engine.py
wdykas: maybe we add a back off or something as well ?

megatron/core/inference/disaggregation/inference_state_handoff.py
```diff
@@ -0,0 +1,843 @@
...
+    torch.distributed.all_gather_into_tensor(gathered_payload, local_payload, group=group)
+    return [
+        _unpack_int_records(rank_payload[:payload_size])
+        for rank_payload, payload_size in zip(
+            gathered_payload.view(world_size, max_payload_size).cpu().tolist(), payload_sizes
+        )
+    ]
+
+
+class InferenceStateHandoffMixin:
```
wdykas, line 56: I personally have never seen Mixin wording used in code. I think it may be from merging chaitras into mine but its also what both ai recommend so I am choosing to keep it =)

megatron/core/inference/disaggregation/inference_state_handoff.py
```diff
@@ -0,0 +1,750 @@
...
+
+    def _initialize_disaggregation_state(self) -> None:
+        """Initialize state without importing or constructing a transfer backend."""
+
+        self._pinned_handoff_blocks: Dict[int, list] = {}  # Request ID -> pinned KV block IDs.
+        self._kv_transfer_agent = None
+        self._kv_peer_metas = None  # KV descriptors for this PP stage's TP ranks.
+        self._pp_kv_peer_metas = None  # Each PP stage's set of TP KV descriptors.
+        self._deferred_kv_handoffs = deque()
+        self._pending_kv_imports = deque()
```
wdykas, line 67: not sure about a deque here but leaving for now because its cleaner but larger ones could slow down the queue

santhnm2 (approved):
Discussed some possible additional tests offline

megatron/core/inference/disaggregation/inference_state_handoff.py
```diff
@@ -0,0 +1,906 @@
...
+            queued_request_id = self.waiting_request_ids[0]
+            if queued_request_id != pending.request_id:
+                raise RuntimeError(
+                    "Decode-only admission expected the imported request at the waiting queue "
+                    f"head: expected {pending.request_id}, got {queued_request_id}"
+                )
+            self.waiting_request_ids.popleft()
+            self._validate_decode_ready_handoff(pending)
+            first_token = pending.resume_tokens[0]
+            if request.sampling_params.num_tokens_to_generate > 0:
```
nvcsathe, line 761: Is this the prefill worker generating the first few tokens and then handing off to decode?

megatron/core/inference/disaggregation/inference_state_handoff.py
```diff
@@ -0,0 +1,906 @@
...
+
+    def _try_start_kv_handoff_import(self, handoff: DeferredKvHandoff) -> bool:
+        """Reserve cache state and start one import, or return without mutation.
+
+        Side: decode engine; pull and push transport paths.
+        """
+
+        allocator = self.context.kv_block_allocator
+        cached_blocks = []
+        if not self._kv_transfer_agent.is_push:
```
nvcsathe, line 554: Is decode side prefix caching only enabled for NIXL?

shanmugamr1992 (approved):

megatron/core/inference/disaggregation/inference_state_handoff.py
wdykas: It just prefills which gives you the first next token which gets handed to decode side immediately for decoding

megatron/core/inference/disaggregation/inference_state_handoff.py
wdykas: its enabled but the optimization of only sending what the decode needs instead of the whole cache every time only works with pull based systems

sidsingh-nvidia (approved):

megatron/core/inference/disaggregation/inference_state_handoff.py
nvcsathe: Does this get complicated with MTP/speculative decoding? Because I think I set this to 0 in the dynamo integration so that generation is completely owned by the decode worker

wdykas: yeah we transfer speculative tokens also now

wdykas: That is all handled as "continuation_tokens" in the data class

nvcsathe (approved):

megatron/core/inference/disaggregation/inference_state_handoff.py
```diff
@@ -0,0 +1,906 @@
...
+
+        self._drain_deferred_kv_handoffs()
+        if not self._pending_kv_imports:
+            return 0
+        self._report_completed_kv_imports()
+        ready = 0
+        remaining = deque()
+        while self._pending_kv_imports:
+            pending = self._pending_kv_imports.popleft()
+            failed = self._handoff_completion_notifications.pop(pending.request_id, None)
```
lmcafee-nvidia: my understanding is that a request can be copied onto the decode worker even when the decode worker has no free active-request slot, which could be problematic. it looks like the KV-cache capacity is checked before the transfer, but `max_requests` is checked after the transfer. the transfer reports success only once, and this line removes that success before checking whether the decoder has an open slot. if the decoder is full, the request goes back to the pending list, but it cannot report success again. I think we need to keep the success state until the request starts or fails, so the client does not wait forever and the reserved memory can be released.

here's an example flow that I think would trigger this issue:
- the decoder has `max_requests=1`, and request A uses that slot.
- request B finishes its transfer and receives one success marker.
- the code removes that marker, sees A, and returns B to the pending list.
- request A finishes, but no new success marker is created for B.
- request B remains pending forever while its client waits.

megatron/core/inference/engines/dynamic_engine.py
```diff
@@ -2686,6 +2825,9 @@ def schedule_requests(self) -> int:
                 # Control signal: queue for second pass.
                 self._pending_signals.append(message)
 
+        self._poll_pending_kv_imports()
```
lmcafee-nvidia, line 2884: the async path keeps one batch’s logits until the next step uses them. this poll can add a new request before those previous logits, made only for the previously existing request set, are read. the new request must be delayed and only added during admission in the `no_overlap` path, so that logits stay coherent and we don't fail with a size error. here is the place that the new request can be safely added: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/inference/text_generation_controllers/text_generation_controller.py#L2858

megatron/core/inference/disaggregation/decode_admission.py
```diff
@@ -0,0 +1,136 @@
...
+    context.request_in_prefill_status_tensor[current_id] = 0
+    context.request_kv_block_counts[current_id] = len(all_block_ids)
+    context.request_last_kv_block_id[current_id] = all_block_ids[-1]
+    context.request_last_kv_block_offset[current_id] = (
+        prompt_length - 1 + input_token_count
+    ) % context.block_size_tokens
+
+    token_start = context.active_token_count
+    token_end = token_start + input_token_count
+    token_positions = torch.arange(
```
lmcafee-nvidia, line 106: I think we should make sure to finish the request immediately when the transferred token meets its length/end-token/stop-rule limit. for example, with `max_sequence_length=4` and a four-token prompt, this code schedules position 4 even though valid positions end at 3.

megatron/core/inference/disaggregation/inference_state_handoff.py
```diff
@@ -0,0 +1,906 @@
...
+                    )
+                request.ttft = first_token_event.timestamp - request.event_add_engine.timestamp
+
+                stop_word_hit = False
+                if request.stop_word_ids:
+                    stop_word_hit, _ = self._check_stop_words_for_request_post_append(request)
+                if first_token == request.sampling_params.termination_id or stop_word_hit:
+                    request.sampling_params.num_tokens_to_generate = len(request.generated_tokens)
+
+            admit_prefilled_decode(
```
lmcafee-nvidia: regarding my previous comment about immediately finishing a request that has met a termination limit, I believe this call should only be run conditional on if the request has not immediately finished.

megatron/core/inference/engines/dynamic_engine.py
wdykas: it this specific to async scheduling or just in general?

lmcafee-nvidia: yes, this exact logits-coherency problem is specific to async scheduling because that mode retains logits across steps; legacy scheduling consumes them within the same step. current HEAD makes polling readiness-only and admits the handoff through the no-overlap callback after the prior logits are consumed and cleared, so I think that this concern is addressed.

shanmugamr1992:
### NeMo-RL functional tests

cluster: `oci-hsg` · image: `nvcr.io/nvidian/nemo-rl:nightly` · updated 2026-08-13 17:19 UTC

**Exactly what was tested**

| Component | Ref | Commit |
| --- | --- | --- |
| megatron-core | #6222 head | `caab981b` |
| Megatron-Bridge | `refs/heads/main` — an override, **not** the Bridge NeMo-RL pins | `cf7cad53` |
| NeMo-RL | `nrlta/integration` | `d6d15ce6` |

Compared against a baseline of the same suite on megatron-core `main` at `80cda5db`.

| Test | Status | Comment |
| --- | --- | --- |
| `grpo_megatron_generation` | ✅ pass |  |
| `grpo_megatron_generation_topology` | ✅ pass |  |
| `grpo_megatron_generation_non_colocated` | ✅ pass |  |
| `grpo_megatron_generation_async` | ✅ pass |  |
| `grpo_megatron_generation_colocated_async` | ✅ pass |  |
| `grpo_megatron_generation_async_gym` | ✅ pass |  |
| `grpo_megatron_generation_topp_topk` | ⚠️ fail (pre-existing) | Also broken on megatron-core `main` (80cda5db), so this is not caused by this PR. Being fixed separately. Pre-existing on megatron-core main, not caused by any labeled PR. grpo_megatron_generation_topp_topk runs to completion but max(train/token_mult_prob_error) lands at 1.078047513961792 against a < 1.06 threshold. The value is bit-identical on Megatron-LM#5382 head eb01b689 and on main 3aee84c3 (both with the Bridge FSDP fix), so it is a deterministic train/generation logprob mismatch in the top-p/top-k sampling path rather than run-to-run noise. Marginal at ~1.7% over. No fix raised; needs an owner to decide whether the sampling path regressed or the threshold is stale. |
| `grpo_megatron_generation_colocated_gym` | ✅ pass |  |
| `grpo_megatron_generation_async_prefix_caching` | ⚠️ fail (pre-existing) | Also broken on megatron-core `main` (da18ed66), so this is not caused by this PR. Being fixed separately. L2 nano-3.5 prefix-caching GRPO on megatron generation exceeds train/token_mult_prob_error max<1.1 on megatron-core main as well. Not attributed to a labeled Megatron-LM PR; needs a dedicated NeMo-RL/mcore investigation (threshold vs real logprob mismatch). |
| `grpo_megatron_generation_async_mxfp8` | ⚠️ fail (pre-existing) | Also broken on megatron-core `main` (da18ed66), so this is not caused by this PR. Being fixed separately. L2 nano-3.5 MXFP8 GRPO on megatron generation exceeds train/token_mult_prob_error max<1.5 on megatron-core main as well. Not attributed to a labeled Megatron-LM PR; needs a dedicated NeMo-RL/mcore investigation. |

Tested PR head as-is without merging main (branch still conflicting; operator confirmed the conflicts are unit-test-only). Results are against the un-merged tip, not main+PR. Megatron-Bridge was pinned to its main branch because the NeMo-RL submodule pin still lags.

<sub>Posted by the nemo-rl-testing-agent. Re-runs edit this comment in place.</sub>

megatron/core/inference/disaggregation/decode_admission.py
wdykas: addressed

megatron/core/inference/disaggregation/inference_state_handoff.py
wdykas: adressed

megatron/core/inference/engines/dynamic_engine.py
wdykas: addressed

lmcafee-nvidia (approved):

kvareddy (approved):

wdykas:
/ok to test [3489132](https://github.com/NVIDIA/Megatron-LM/pull/6222/commits/3489132c5ec96b727d43f2b611a1934364d0c288)

PR #6229, opened 2026-08-03
[NCCL EP] Support eager mode and overflow handling

YangFei1990:

## Add eager mode and overflow handling to the ncclEP flex dispatcher. 

The ncclEP backend now selects between two modes based on `moe_expert_rank_capacity_factor`: unset selects eager, where TE sizes the receive buffer per step from the actual received-token count; setting it selects the existing static path, which keeps a fixed budget and stays CUDA-graph capturable. 

In static mode the dispatcher enables NCCL EP's drop-on-overflow policy and accumulates over_budget and the peak required capacity device-side per dispatch, so the happy path never syncs — a MAX all-reduce runs only once overflow is reported. On overflow, PagedStashRunner grows the receive budget to that observed peak, rebuilds the EP context once, and replays the step, note this is different compared with hybrid where it use eager mode for rerun and fallback to same capacity afterwards. 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1527,51 +1532,62 @@ def __init__(
...
-        if self.static_shape:
-            # static shape needs a fused grouped GEMM that consumes ragged per-expert counts on
-            # device (no host-side split narrowing): moe_grouped_gemm selects the grouped experts
-            # and use_transformer_engine_op_fuser fuses FC1+act+FC2 over them (fp8/fp4 via the CuTe
-            # DSL fused grouped MLP, bf16 via the op-fuser GroupedLinear grouped-tensor path).
+        # [TODO] Add support for drop and pad with the 'ncclep' backend.
+        if config.moe_pad_expert_input_to_capacity or config.moe_expert_capacity_factor is not None:
+            raise ValueError("drop and pad is not supported yet with the 'ncclep' backend.")
+
+        if self.eager:
```
Autumn1998, line 1556: So when we use zero copy now, do we still need to specify an appropriate recv-token-nums-per-rank? As I remember, BF16 used a symmetric memory pool, while FP8 required a fixed-size symmetric memory buffer. Why doesn’t BF16 zero-copy support eager mode either?

megatron/core/transformer/moe/paged_stash.py
Autumn1998, line 1134: This log maybe wrong if we rerun with larger stashing buffer? 🤔

megatron/core/transformer/moe/token_dispatcher.py
YangFei1990: We do not support zero-copy with eager mode at the moment due to the memory overhead introduced by the eager symm pool.

Autumn1998: Is this because the symmetric memory pool and the normal memory pool are independent of each other, so they cannot be reused simultaneously, leading to additional GPU memory usage?

YangFei1990: Yes symm buffer pool is independent with Pytorch normal pool. When using symm pool for eager, it will always be allocated to the outlier recv buffer size, which kind of defeats the purpose of eager.

YangFei1990: Double thinking, it is totally unacceptable for bf16 path, because the symm tensor will be saved per-layer per-mb as activations, so the pool will be super large. But for fp8 path, because only the quantized tensor will be save for backward, the symm tensor can be reused (in this case we allocate persistent tensor for graph mode), so the pool size probably can be acceptable, despite it is still allocated to the outlier recv size, but just for one tensor. 

I think we should support eager mode zero copy with fp8 path, but nccl ep internal does not support it yet. I will work with the team for it later.

megatron/core/transformer/moe/paged_stash.py
YangFei1990: I actually changed it to run eagerly during rerun. Because NCCL EP needs to rebootstrap every time when changing mode/recv size, so I originally choose to do static rerun with updated recv size. However, I found that it does not handle the stash overflow very well. For simplicity and also be consistent with hep, I change NCCL EP to do eager rerun as well, despite there will be one more bootstrap, but the logic is much simpler. Please check the updated code.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1596,29 +1610,46 @@ def setup_metadata(self, routing_map: torch.Tensor, probs: torch.Tensor):
...
+        # PagedStashRunner will set factor to None during rerun to enable eager mode
+        self.eager = self.moe_expert_rank_capacity_factor is None
+        # TODO: support eager mode with zero_copy
+        self.zero_copy = self.config.moe_ncclep_zero_copy and not self.eager
+        self._zc_quant = self.zero_copy and bool(self.config.fp8 or self.config.fp4)
         # NCCL EP's HT backend requires max_dispatch_tokens_per_rank to be a multiple of the HT
         # chunk size (64); ncclEpCreateGroup otherwise fails with "invalid usage".
         # (nccl_ep device/hybridep_adapter.cu).
         _HT_TOKENS_PER_CHUNK = 64
+        # TODO: support THD/Dynamic CP when different ranks might have different number of tokens
```
phu0ngng, line 1628: FYI, NCCL EP already supports this.
We will need to test and see if we need to make small fixes here and there in MCore and TE to enable this feature.

phu0ngng (approved):

YangFei1990:
/ok to test [8b34958](https://github.com/NVIDIA/Megatron-LM/pull/6229/commits/8b34958cb277cf4b607f6bdbda72532bdfea1376)

fanshiqing (approved):

Autumn1998 (approved):

Phlip79 (approved):

megatron/core/transformer/transformer_config.py
```diff
@@ -949,22 +949,12 @@ class TransformerConfig(ModelParallelConfig):
     moe_hybridep_num_sms_preprocessing: int = 108
     """Number of SMs to use for HybridEP preprocessing (metadata scan kernel)."""
 
-    moe_ncclep_static_shape: bool = False
```
jaredcasper, line 961: Does this need to be properly deprecated? How long has this been here?

YangFei1990: It just added a few weeks before in the previous nccl_ep PR. I think it is okay to remove it as nccl_ep is still in a pretty early stage so there is not a lot of usage.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1690,20 +1723,38 @@ def dispatch(
             recv_topk_weights=_NCCLEPManager._zc_recv_topk_weights_buf,
         )
         self.tokens_per_expert = tokens_per_expert.to(torch.int64)
+        if not self.eager:
+            # ep_prepare fills total_recv_tokens from the routing map, before any dropping, so
+            # it is the capacity this step actually needed.
+            total_recv_tokens = self._buffer.total_recv_tokens
+            self.over_budget |= total_recv_tokens > self._recv_capacity
+            torch.maximum(self.required_recv, total_recv_tokens, out=self.required_recv)
```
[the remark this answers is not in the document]
YangFei1990: This is not true. The `total_recv_tokens` is already padded total. The `grow_recv_capacity ` is guaranteed to upgrade the recv buffer to the desired size.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1690,20 +1723,38 @@ def dispatch(
...
+        the routing that overflowed instead of dropping again on the next step. Monotonic.
+        """
+        if self.config.moe_expert_rank_capacity_factor is None:
+            return
+        # Overflow means total_recv_tokens > _recv_capacity >= the current override, so the
+        # peak that triggered it is always strictly larger. Anything else is stale accounting.
+        assert new_capacity > (self._recv_capacity_override or 0), (
+            f"receive capacity must grow on overflow: peak {new_capacity} is not above the "
+            f"current override {self._recv_capacity_override}"
+        )
```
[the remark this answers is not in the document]
YangFei1990: Currently we only consider the case where `_max_tokens_per_rank `/`_recv_capacity_override `/ `_recv_capacity ` consistent across layers. Changing that is not supported by NCCL EP and there is already a TODO of THD support that is supposed to handle this issue.

megatron/core/transformer/moe/fused_a2a.py
```diff
@@ -635,29 +636,44 @@ def ensure_nccl_ep_bootstrapped(
...
     te_ep.ep_bootstrap(
         ep_group,
         num_experts=num_experts,
         max_tokens_per_rank=max_tokens_per_rank,
         recv_capacity_per_rank=recv_capacity_per_rank,
         hidden_dim=hidden_dim,
+        num_topk=num_topk,
         max_num_sms=num_sms,
         zero_copy=zero_copy,
+        drop_on_overflow=recv_capacity_per_rank is not None,
```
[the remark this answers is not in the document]
YangFei1990: Similar comment as https://github.com/NVIDIA/Megatron-LM/pull/6229#discussion_r3738011924, nccl ep is in early stage and it should be okay to have APIs changes

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1596,29 +1610,46 @@ def setup_metadata(self, routing_map: torch.Tensor, probs: torch.Tensor):
...
+            self._recv_capacity = None
+        else:
+            budget = int(
+                self._max_tokens_per_rank * self.router_topk * self.moe_expert_rank_capacity_factor
+            )
+            if self._recv_capacity_override is not None:
+                budget = max(budget, self._recv_capacity_override)
+            if self.alignment != 0:
+                budget += -budget % self.alignment
+            self._recv_capacity = budget
```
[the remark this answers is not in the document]
YangFei1990: In practice user won't set factor < 1, it is only for testing purpose

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -1082,10 +1092,41 @@ def check_moe_overflow(self):
...
+                required = torch.cat(per_layer).max().reshape(1)
+                torch.distributed.all_reduce(required, op=torch.distributed.ReduceOp.MAX)
+                self._required_recv_capacity = int(required.item())
+                comm_manager = self.moe_layers[0].token_dispatcher._comm_manager
+                denominator = getattr(comm_manager, '_max_tokens_per_rank', 0) * getattr(
+                    comm_manager, 'router_topk', 0
+                )
+                # Track the updated cap factor to log for user
+                if denominator:
+                    self._required_capacity_factor = self._required_recv_capacity / denominator
```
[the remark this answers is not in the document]
YangFei1990: Same as above comment, we now assume `_max_tokens_per_rank ` consistent across layers

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -1136,6 +1189,8 @@ def _try_copy_main_params(opt):
                 stage='training' if is_training else 'validation'
             )
 
+        nccl_ep_release_context()
```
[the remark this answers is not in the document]
YangFei1990: This is not true, even for stashing buffer overflow, we still want to re-bootstrap. The rerun is always in eager mode.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1596,29 +1610,46 @@ def setup_metadata(self, routing_map: torch.Tensor, probs: torch.Tensor):
...
+        """
+        Bootstrap NCCL EP and size the receive buffer on first use or after paged stash rerun
+        """
+        # A released context always needs a rebuild, and every mode change is paired with one.
+        if self._bootstrapped and is_nccl_ep_bootstrapped():
             return
+        # PagedStashRunner will set factor to None during rerun to enable eager mode
+        self.eager = self.moe_expert_rank_capacity_factor is None
+        # TODO: support eager mode with zero_copy
+        self.zero_copy = self.config.moe_ncclep_zero_copy and not self.eager
```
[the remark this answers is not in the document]
YangFei1990: No, the rerun is always eager, by design. Thus even user is using zero copy, it will be dismissed. I have considered to have a `_derive_mode ` helper previously, but it actually makes code more complicated. 

Also, I do not understand why this is marked as **[IMPORTANT Correctness]**

megatron/core/transformer/moe/paged_stash.py
```diff
@@ -1214,17 +1266,34 @@ def __call__(self, *args, **kwargs):
                         mlp.token_dispatcher._comm_manager.moe_expert_rank_capacity_factor = (
                             mlp.token_dispatcher.config.moe_expert_rank_capacity_factor
                         )
+                # Only after an actual rerun: rebootstrap NCCL EP
+                if num_tries > 1:
+                    for mlp in self.moe_layers:
+                        mlp.token_dispatcher.invalidate_ep_bootstrap()
+                    nccl_ep_release_context()
```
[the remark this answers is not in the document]
YangFei1990: No this is not redundant. `nccl_ep_release_context ` is global, but `invalidate_ep_bootstrap ` is per-layer. This is by design to have more fine-grained control.

megatron/core/transformer/moe/token_dispatcher.py
```diff
@@ -1997,7 +2051,28 @@ def check_over_budget(self):
...
+        """Grow the backend's receive budget to ``new_capacity`` so an overflowed step can be
+        replayed without dropping. No-op unless the backend supports it."""
+        if hasattr(self._comm_manager, 'grow_recv_capacity'):
+            self._comm_manager.grow_recv_capacity(new_capacity)
+
+    def invalidate_ep_bootstrap(self) -> None:
+        """Force the backend to re-bootstrap on its next dispatch. Call after changing the mode:
+        every manager must re-derive it, not just the one that happens to dispatch first."""
+        if hasattr(self._comm_manager, '_bootstrapped'):
+            self._comm_manager._bootstrapped = False
```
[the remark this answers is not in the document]
YangFei1990: This is to align with the existing code style of the flex dispatcher and comm manager

YangFei1990:
/ok to test [b993cfe](https://github.com/NVIDIA/Megatron-LM/pull/6229/commits/b993cfeb3f50cddd9e73e5b0a59d3c890e69a6fc)

jaredcasper (approved):

YangFei1990:
/ok to test [2f67d31](https://github.com/NVIDIA/Megatron-LM/pull/6229/commits/2f67d31eea17e6c9c6e7153403da5f76fe360bbf)

PR #6230, opened 2026-08-03
Release MFSDP storage after model deletion

### LINKED ISSUE #5616: MFSDP v2 should support HSDP and HFSDP placements

wujingyue:
MFSDP installed strong back-references from tensor-owned or runtime state to the module graph. Deleting a sharded model could therefore retain its persistent CUDA storage indefinitely.

Make the ownership graph acyclic: parameter ownership markers and post-accumulate hooks hold groups/modules weakly; parameter groups hold their owning modules weakly; the shared context holds its root and prefetch orders weakly; and module hooks dispatch from their runtime module argument rather than close over it. Internal APIs do not expose dereferenced owner or root modules, preventing accidental retention. Deleting a model now releases FSDP storage immediately, without explicit garbage collection or a teardown API.

Add an activation-path regression that runs a sharded model, deletes it, and requires device allocation to return below 1 MiB. It reproduces the leak on origin/main without the main-grad materialization changes.

### DIFF
--- a/tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
+++ b/tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
@@ -73,6 +73,18 @@ class MultiChildModel(nn.Module):
         return x
 
 
+class ElementwiseModel(nn.Module):
+    """Small activation path over a large FSDP-managed weight."""
+
+    def __init__(self, dim: int) -> None:
+        super().__init__()
+        self.weight = nn.Parameter(torch.ones(dim, dim))
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Apply the first weight row to an activation tensor."""
+        return torch.relu(x + self.weight[0])
+
+
 class TiedLM(nn.Module):
     """Tiny language model with shared input and output embedding weights."""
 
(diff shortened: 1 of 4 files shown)


shjwudp (approved):

yaoyu-33 (approved):

tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py
```diff
@@ -423,6 +435,28 @@ def test_forward_peak_memory_bounds_in_flight_child_all_gathers(distributed_setu
...
+    model = ElementwiseModel(dim=8192).to(dtype=torch.bfloat16, device=device)
+    fully_shard(model, mesh=mesh, placements=_flat_placements())
+
+    x = torch.ones(1, 8192, dtype=torch.bfloat16, device=device)
+    output = model(x)
+    torch.cuda.synchronize(device)
+    del output, x, model
+    torch.cuda.synchronize(device)
+
+    assert torch.cuda.memory_allocated(device) - allocated_before < 1024**2
```
skyw, line 456: Q: is pytorch guaranteed to update `memory_allocated` after dereferencing? regardless whether cache allocator's decision to release the memory or not.

wujingyue: Yes, unlike https://docs.pytorch.org/docs/2.13/generated/torch.cuda.memory.memory_reserved.html

jaredcasper (approved):

PR #6232, opened 2026-08-04
Fix GTP prefetch parameter readiness

JF-D:

## Summary

- attach a GTP-only readiness handle to each DDP parameter-gather bucket used by GTP weights
- finish the required DDP gather before eager GTP weight reads
- record capture-local readiness dependencies in `gtp_cuda_graphs.py` and finish them before replaying each local forward CUDA graph
- keep the existing DDP pre-forward gather policy shared by both paths

## Root cause

GTP can prefetch the next module's local weight shard before that module executes. With overlapped DDP parameter gathering, the ordinary module pre-forward hook therefore runs too late: GTP may read the distributed-optimizer parameter buffer while its bucket all-gather is still in flight. CUDA graph capture also suppresses the DDP hook, so the dependency must be retained and replayed explicitly.

Each forward graph capture now creates its own `GTPCaptureCommState`. The transient active state is used only while capturing; the yielded state is stored on that graph's `_CudaGraphRunner`, preventing handle sets from aliasing across graph instances.

### DIFF
--- a/megatron/core/distributed/distributed_data_parallel.py
+++ b/megatron/core/distributed/distributed_data_parallel.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
 
 import logging
+import weakref
 from contextlib import contextmanager
 from typing import Optional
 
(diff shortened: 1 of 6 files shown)


JF-D:
@claude strict-review

JF-D:
/ok to test https://github.com/NVIDIA/Megatron-LM/commit/0f18dfe3d10fb905326f2f1fe86978e8c3459652

fanshiqing:
about to close this one as this issue has been fixed in https://github.com/NVIDIA/Megatron-LM/pull/6388

PR #6234, opened 2026-08-04
[fix] Cleanup usage of pg_collection in gtp

fanshiqing:

# What does this PR do?
Refine usage of `pg_collection` in gtp, which is a planned following up tiny task of https://github.com/NVIDIA/Megatron-LM/pull/4967

### Pre-checks

### DIFF
--- a/megatron/core/distributed/finalize_model_grads.py
+++ b/megatron/core/distributed/finalize_model_grads.py
@@ -612,6 +612,19 @@ def finalize_model_grads(
         gtp_remat_group = parallel_state.get_gtp_weight_remat_group(check_initialized=False)
         egtp_remat_group = parallel_state.get_expert_gtp_weight_remat_group(check_initialized=False)
 
+    # A missing group would silently skip the gtp_remat-axis reduction below and train on
+    # wrong gradients, so fail loudly whenever the config says the axis is active.
+    for axis, group, axis_size in (
+        ('gtp_remat', gtp_remat_group, config.gtp_weight_remat_size),
+        ('expt_gtp_remat', egtp_remat_group, config.expert_gtp_weight_remat_size),
+    ):
+        if axis_size > 1:
+            found = 'None' if group is None else f'a size-{group.size()} group'
+            assert group is not None and group.size() == axis_size, (
+                f"{axis} is enabled (size={axis_size}) but pg_collection provides {found}. "
+                f"Pass a pg_collection carrying `{axis}` to finalize_model_grads."
+            )
+
     # Fence the current stream against all GTP backward grad work before the DP gradient sync.
     if config.gtp_weight_remat_size > 1 or config.expert_gtp_weight_remat_size > 1:
         from megatron.core.tensor_parallel.gtp_api import (
--- a/megatron/core/distributed/finalize_model_grads.py
+++ b/megatron/core/distributed/finalize_model_grads.py
@@ -492,7 +492,10 @@ _allreduce_layernorm_grads = _allreduce_non_tensor_model_parallel_grads
 
 
 def _allreduce_replicated_grads_over_gtp_remat_group(
-    model: List[torch.nn.Module], calculate_per_token_loss: bool = False
+    model: List[torch.nn.Module],
+    gtp_remat_group: Optional[torch.distributed.ProcessGroup],
+    egtp_remat_group: Optional[torch.distributed.ProcessGroup],
+    calculate_per_token_loss: bool = False,
 ):
     """Complete the gtp_remat / egtp_remat axis reduction for replicated parameters.
 
@@ -510,12 +513,6 @@ def _allreduce_replicated_grads_over_gtp_remat_group(
 
     No-op when GTP_remat is inactive (group size <= 1).
     """
-    pg_collection = ProcessGroupCollection.use_mpu_process_groups(
-        required_pgs=["gtp_remat", "expt_gtp_remat"]
-    )
-    gtp_remat_group = pg_collection.gtp_remat
-    egtp_remat_group = pg_collection.expt_gtp_remat
-
     dense_active = gtp_remat_group is not None and gtp_remat_group.size() > 1
     expert_active = egtp_remat_group is not None and egtp_remat_group.size() > 1
     if not dense_active and not expert_active:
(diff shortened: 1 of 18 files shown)


megatron/core/extensions/transformer_engine.py
```diff
@@ -1362,10 +1362,13 @@ def __init__(
         tp_group: Optional[torch.distributed.ProcessGroup] = None,
         stride: int = 1,
         name: str | None = None,
+        pg_collection: Optional[ProcessGroupCollection] = None,
```
fanshiqing, line 1365: @yashaswikarnati do we mean to set `pg_collection` as non-optional here? same question for the local path.

yashaswikarnati: i think for backward compatiblity we would need it to be optional

megatron/core/tensor_parallel/layers.py
```diff
@@ -935,6 +942,7 @@ def __init__(
         disable_grad_reduce: bool = False,
         tp_group: Optional[torch.distributed.ProcessGroup] = None,
         name: str | None = None,
+        pg_collection: Optional[ProcessGroupCollection] = None,
```
yashaswikarnati, line 937: I know this is tricky. we might have to remove tp_group in the arg as pg_collecton will anyway have it.. 
`
 tp_group: Optional[torch.distributed.ProcessGroup] = None,`

megatron/core/extensions/transformer_engine.py
yashaswikarnati: we might need to remove tp_group arg here as pg_collection should already have it. same comment as layers.py

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -604,12 +601,16 @@ def finalize_model_grads(
         # Full DP x CP x gtp_remat group: num_tokens (the per-token-loss divisor below) counts the
         # gtp_remat peers' distinct tokens. Falls back to replicate dp_cp when gtp is inactive.
         dp_cp_group = getattr(pg_collection, 'dp_cp_gtp_remat', None) or pg_collection.dp_cp
+        gtp_remat_group = getattr(pg_collection, 'gtp_remat', None)
```
yashaswikarnati, line 604: I dont know the downstream usage of these groups, so here is a naive question

Will none fallback be dangerous here ? like it may use the world process group.. . or should we assert that these groups are not none.

megatron/core/extensions/transformer_engine.py
fanshiqing: yeah, understood, but `remove tp_group arg` will not be included in this MR

megatron/core/distributed/finalize_model_grads.py
fanshiqing: > Will none fallback be dangerous here?

No, when `gtp_remat_group` or `egtp_remat_group` is None, the `_allreduce_replicated_grads_over_gtp_remat_group` will just do nothing for the corresponding group.

yashaswikarnati: so i meant the opposite. when _allreduce_replicated_grads_over_gtp_remat_group needs the group. but we pass None.

fanshiqing: Done. Add assert below when the config enables the axis but the group is missing, instead of treating it as "GTP inactive" and training on wrong gradients.

megatron/core/tensor_parallel/layers.py
fanshiqing: The goal of this MR is to resolve unsolved task left in GTP. Let's keep `removing tp_group` for a furure task is gtp independent and more relevant to backward compatibility

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -1614,7 +1623,7 @@ def forward_backward_helper_wrapper(
                 recv_next = True
                 if is_pp_last_stage(p2p_communicator.pp_group):
                     recv_next = False
-                (input_tensor, output_tensor_grad) = (
+                input_tensor, output_tensor_grad = (
```
yashaswikarnati, line 1626: just a nit; may  be lets not modify this if not necessary for this PR. uunless the linter is not happy

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -690,6 +690,15 @@ def _build_default_pg_collection() -> ProcessGroupCollection:
     pg_collection.dp = parallel_state.get_data_parallel_group(
         with_context_parallel=False, partial_data_parallel=False
     )
+    # gtp_remat axis: consumers read these with getattr and silently skip the gtp_remat
+    # reduction when absent, so populate them even when GTP_remat is inactive.
+    pg_collection.gtp_remat = parallel_state.get_gtp_weight_remat_group(check_initialized=False)
```
yashaswikarnati, line 695: would we need check_initialized=False here ? also seems pg_collection.dp_cp_gtp_remat  does not pass this though.

yashaswikarnati (approved):

megatron/core/pipeline_parallel/schedules.py
fanshiqing: yeah, these are reported by the online lint check, otherwise I really don't want to touch these GTP-unrelated stuffs...

megatron/core/pipeline_parallel/schedules.py
fanshiqing: because `get_data_parallel_group`  don't have the option for `check_initialized` (and the only group without this flag), becasue we ALWAYS requires this `dp_cp_gtp_remat` to be not None.

fanshiqing: set `check_initialized=False` just to align the above non-DP related groups like `embd` and `pos_embd`, where group=None is accepted is not activated.

if GTP is active while the group is not finalized, then gtp-guard in `finalize_model_grads` will take effect.

so just keep it as is, it costs nothing.

yaox12 (approved):

Phlip79 (approved):

kvareddy (approved):

mathemakitten (approved):

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -604,12 +601,29 @@ def finalize_model_grads(
...
+        egtp_remat_group = getattr(pg_collection, 'expt_gtp_remat', None)
     else:
         tp_group = parallel_state.get_tensor_model_parallel_group()
         pp_group = parallel_state.get_pipeline_model_parallel_group()
         embd_group = parallel_state.get_embedding_group(check_initialized=False)
         pos_emb_group = parallel_state.get_position_embedding_group(check_initialized=False)
         dp_cp_group = parallel_state.get_data_parallel_group(with_context_parallel=True)
+        gtp_remat_group = parallel_state.get_gtp_weight_remat_group(check_initialized=False)
```
deepakn94, line 612: @yashaswikarnati when are we getting rid of all of this fallback code?

deepakn94: Not only for GTP, but everywhere. It's an eyesore :P

deepakn94 (approved):

yaoyu-33 (approved):

PR #3339, opened 2026-02-10
docs: fix wrong documentation link in README.md

mental2008:
Hi team, I submitted this PR to correct a broken link to the documentation in the README.md file. The original URL was outdated and returned a "Page Not Found" error; it has been updated to point to the current, valid documentation page.

Example of the fix:
```markdown
<!-- Before -->
[Documentation](https://docs.nvidia.com/Megatron-Core/developer-guide/latest/index.html) ❌

<!-- After -->
[Documentation](https://docs.nvidia.com/megatron-core/developer-guide/latest/get-started/quickstart.html) ✅
```

This small but important fix helps maintain the project's professionalism and usability.

PR #3003, opened 2026-01-19
[MLA] fix: Pad V when Q/V head dims differ for THD

HollowMan6:
# What does this PR do ?

https://github.com/volcengine/verl/pull/4931

Resolves https://github.com/nvidia/megatron-lm/issues/1698

For MLA, we shall pad V when Q/V head dims differ for THD

Refer to
https://github.com/modelscope/ms-swift/blob/50d54089a824578f64bde37f8b47785d955ede4f/swift/megatron/init.py#L63-L177

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

## Contribution process

```mermaid
flowchart LR
    A[Pre-checks] --> B[PR Tests]
    subgraph Code Review/Approval
        C1[Expert Review] --> C2[Final Review]
    end
    B --> C1
    C2 --> D[Merge]
```

### Pre-checks

- [ ] I want this PR in a versioned release and have added the appropriate Milestone (e.g., `Core 0.8`)
- [x] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

The following process is enforced via the CODEOWNERS file for changes into `megatron/core`. For changes outside of `megatron/core`, it is up to the PR author whether or not to tag the Final Reviewer team.

<details>
<summary>For MRs into `main` branch</summary>

Feel free to message or comment the @mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

#### (Step 2): Collect the expert reviewers reviews

1. Attach the `Expert Review` label when your PR is ready for review.
2. GitHub auto-assigns expert reviewers based on your changes. They will get notified and pick up your PR soon.

:warning: Only proceed to the next step once all reviewers have approved, merge-conflict are resolved and the CI is passing.  
Final Review might get declined if these requirements are not fulfilled.

#### (Step 3): Final Review

1. Add `Final Review` label
2. GitHub auto-assigns final reviewers based on your changes. They will get notified and pick up your PR soon.

#### (Optional Step 4): Cherry-pick into release branch

[...]

PR #6241, opened 2026-08-04
Let MLA skip the thd V pad on the packed path

### LINKED ISSUE #6240: MLA pads V on the thd path unconditionally, which costs up to 87% of a step at long sequence

On the packed (`thd`) path, `_prepare_mla_core_attention_value` pads V up to the QK width whenever the two differ, and `_trim_mla_core_attention_output` trims the result back. For MLA the widths always differ — the RoPE dims ride on q/k but not on v — so every MLA `thd` layer goes through the pad.

The pad and the trim cancel out. What they cost is a V that is 50% wider than it needs to be through the attention kernel, plus the pad and the trim, and on some stacks a fallback to a slower backend as well.

Measured on a small MLA model with DeepSeek-V2-Lite attention dims (`qk_head_dim` 128, `qk_pos_emb_head_dim` 64, `v_head_dim` 128, `kv_lora_rank` 512, `q_lora_rank` None), 8 layers, hidden 2048, `ffn_hidden_size` 10944, bf16, `attention_dropout` 0. Whole training step — forward, backward, Adam — median of 40 steps after 15 warmup, two rounds per cell with the arm order reversed between them:

| tokens | packs | tokens/seq | H200 native | H200 padded | H200 | B300 native | B300 padded | B300 |
|---|---|---|---|---|---|---|---|---|
| 4096 | 4 | 1024 | 57.7, 58.0 | 61.0, 61.3 | **5.7%** | 57.1, 58.0 | 57.8, 58.3 | — |
| 8192 | 4 | 2048 | 85.3, 85.8 | 91.4, 92.0 | **7.2%** | 62.5, 63.6 | 68.6, 69.9 | **10.0%** |
| 16384 | 4 | 4096 | 156.1, 157.3 | 168.9, 169.1 | **7.8%** | 97.2, 97.3 | 125.4, 125.5 | **29.0%** |
| 16384 | 1 | 16384 | 214.7, 215.0 | 239.9, 240.0 | **11.7%** | 116.4, 116.5 | 218.3, 218.3 | **87.4%** |

Both stacks improve monotonically with sequence length, which is what you would expect if the saving is in attention: attention is O(s²), so its share of the step grows and the wasted width costs more. On B300 the 4096-token cell is a wash — the two rounds disagree on which arm is faster, so it is noise rather than a small win.

The two shapes differ because the backends differ. On H200 both arms run cuDNN fused attention, so the gap is the wasted width alone. On B300 the padded shape gets no fused backend at all and falls back to FlashAttention:

```
# native V (192/128)
Disabling FlashAttention 2 as it does not support MLA.
Selected backend = FusedAttention (sub-backend 1)

# padded V (192/192)
Disabling FusedAttention as no backend supports the provided input
Selected backend = FlashAttention (2.7.4)
```

That message comes from `tex.get_fused_attn_backend` returning `No_Backend` for 192/192 while accepting 192/128 — so padding V to match QK does not widen backend support there, it narrows it. The two stacks differ in both compute capability and TE version, so this data does not isolate which accounts for that difference.

Tag @NVIDIA/mcore-oncall

Let a model opt out of the pad. `need_v_pad` already has an escape hatch for `experimental_attention_variant`; this adds one more condition rather than a new mechanism.

A PR is attached that adds `MLATransformerConfig.mla_native_v_head_dim`, defaulting to `False` so nothing changes unless a model asks for it.

*Deciding automatically instead of by config.* Megatron-Core could skip the pad whenever the backend can take native V. That is the better end state, but it means querying TE's backend selection from Megatron-Core, and the answer varies with the stack — the two measured above already disagree about whether 192/192 is servable at all. Happy to pursue this instead if you prefer; the config flag is proposed because it is small and reviewable.

*Leaving it to FlashAttention 3.* FA3 does support mismatched head dims — TE's `_is_fa3_supported` admits `head_dim_qk` in `(128, 192]` with `head_dim_v` in `(96, 128]`, which covers this geometry. It does not address this. With FA3 3.0.0 installed and eligible on H200, TE still disables it in favour of fused attention ("Disabling FlashAttention to give FusedAttention preference on Hopper+ for performance reasons"), and TE gates FA3 to sm90 so it is unavailable on Blackwell.

*Keeping the pad as a fallback.* The pad is useful when fused attention cannot serve the layer, since it lets FlashAttention take over instead of falling to `UnfusedDotProductAttention`. That is why this is opt-in rather than a change of default.

**Additional context**

Pad-then-trim is an identity bit-exactly, not merely within tolerance: the padded columns of the attention output are sums of zeros, so the trim discards nothing, and every backward input matches. Verified in fp32 and bf16 across several shapes.

Where the two arms end up on different backends — the B300 stack — outputs and gradients agree to bf16 rounding rather than exactly: worst relative drift 5.7e-3 across forward, dq, dk and dv. On H200 both arms run the same kernel.

One measurement note, in case it saves someone else the detour: Megatron-Core defaults `attention_dropout` to 0.1, and on B300 a `thd` request with dropout is served by cuDNN's composite engine, where dropout mask generation costs several times the attention itself. Leaving that default in place reverses the comparison above. The numbers here use `attention_dropout=0`, which is what the models this geometry comes from actually set. Filed separately as NVIDIA/TransformerEngine#3312.

bzantium:

# What does this PR do?

Adds `MLATransformerConfig.mla_native_v_head_dim`, which lets an MLA model hand the attention backend its native V width on the `thd` path instead of padding V up to the QK width.

## Issue tracking

Linked issue: Fix #6240

### Pre-checks

---

## Why

`_prepare_mla_core_attention_value` pads V to the QK width whenever the two differ. For MLA they always differ, so every MLA `thd` layer is padded and its output trimmed back. The two cancel out, so what the pad costs is a 50% wider V through the attention kernel plus the pad and trim — and on some stacks a fallback to a slower backend as well.

Whole training step (forward, backward, Adam) on a small MLA model with DeepSeek-V2-Lite attention dims, 8 layers, hidden 2048, bf16, `attention_dropout` 0. Median of 40 steps after 15 warmup, two rounds per cell with the arm order reversed:

| tokens | packs | tokens/seq | H200 | B300 |
|---|---|---|---|---|
| 4096 | 4 | 1024 | 5.7% | — (noise) |
| 8192 | 4 | 2048 | 7.2% | 10.0% |
| 16384 | 4 | 4096 | 7.8% | 29.0% |
| 16384 | 1 | 16384 | 11.7% | 87.4% |

Both stacks improve monotonically with sequence length, which is what a saving in attention should look like. Absolute timings, backend logs and the reason the two stacks differ in shape are in the linked issue.

## What changes

`need_v_pad` gains one condition. It already skips the pad for `experimental_attention_variant`, so this extends an existing escape hatch rather than adding a mechanism.

Default is `False`: no existing model changes behaviour.

## Correctness

Pad-then-trim is an identity — the padded output columns are sums of zeros, so the trim discards nothing and every backward input is unchanged. Skipping it therefore only changes which backend runs.

The two backends are different kernels, so they agree to bf16 rounding rather than bit-exactly: worst relative drift 5.7e-3 across forward, dq, dk and dv on the geometry above.

## Test

`test_gpu_forward_thd_native_v_head_dim` is parametrized over both settings and asserts that the flag controls `need_v_pad`, that the prepared V has the expected width, and that the forward still produces correctly shaped output and bias. It sits next to the existing `test_gpu_forward_thd_qv_head_dim_mismatch`, which covers the padded path.

## Notes for reviewers

Deciding this automatically — skipping the pad whenever the backend can take native V — would be the better end state, but it requires querying TE's backend selection from Megatron-Core, where the answer varies with TE version, compute capability and cuDNN version. This PR takes the smaller step; happy to rework it if you prefer the automatic route.

The pad is still worth keeping as a fallback: when fused attention cannot serve a layer, padding lets FlashAttention take it instead of dropping to `UnfusedDotProductAttention`. That is why the flag is opt-in.

### DIFF
--- a/megatron/core/transformer/multi_latent_attention.py
+++ b/megatron/core/transformer/multi_latent_attention.py
@@ -90,6 +90,7 @@ def _prepare_mla_core_attention_value(parallel_attention, query, value, packed_s
         packed_seq_params is not None
         and packed_seq_params.qkv_format == "thd"
         and parallel_attention.config.experimental_attention_variant is None
+        and not parallel_attention.config.mla_native_v_head_dim
         and value is not None
         and query.shape[-1] != orig_v_dim
     )
--- a/megatron/core/transformer/multi_latent_attention.py
+++ b/megatron/core/transformer/multi_latent_attention.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import math
+import warnings
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, NoReturn, Optional, Union
 
@@ -82,6 +83,31 @@ if TYPE_CHECKING:
     from megatron.core.packed_seq_params import PackedSeqParams
 
 
+_WARNED_NATIVE_V_BELOW_SM90 = False
+
+
+def _warn_native_v_below_sm90():
+    """Warn once when the native V width is in use on a device with no kernel for it.
+
+    Below compute capability 9.0 neither cuDNN nor FlashAttention accepts unequal QK and V
+    widths, so the attention call falls back to the unfused path. That is two orders of
+    magnitude slower and produces no error of its own, so it is worth saying out loud.
+    """
+    global _WARNED_NATIVE_V_BELOW_SM90
+    if _WARNED_NATIVE_V_BELOW_SM90 or not torch.cuda.is_available():
+        return
+    if torch.cuda.get_device_capability() >= (9, 0):
+        return
+    _WARNED_NATIVE_V_BELOW_SM90 = True
+    major, minor = torch.cuda.get_device_capability()
+    warnings.warn(
+        f"mla_native_v_head_dim is enabled on compute capability {major}.{minor}, where no "
+        "attention backend has a kernel for unequal QK and V head dims. Attention will fall "
+        "back to the unfused path. Set mla_native_v_head_dim=False on this device.",
+        stacklevel=2,
+    )
+
+
 def _prepare_mla_core_attention_value(parallel_attention, query, value, packed_seq_params):
     """Prepare value tensor for MLA core attention THD execution."""
     orig_v_dim = value.shape[-1] if value is not None else None
@@ -97,6 +123,14 @@ def _prepare_mla_core_attention_value(parallel_attention, query, value, packed_s
     if need_v_pad:
         value = F.pad(value, [0, query.shape[-1] - orig_v_dim])
         padded_v_dim = value.shape[-1]
+    elif (
+        packed_seq_params is not None
+        and packed_seq_params.qkv_format == "thd"
+        and parallel_attention.config.mla_native_v_head_dim
+        and value is not None
+        and query.shape[-1] != orig_v_dim
+    ):
+        _warn_native_v_below_sm90()
     return value, need_v_pad, orig_v_dim, padded_v_dim
 
 
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -2980,6 +2980,12 @@ class MLATransformerConfig(TransformerConfig):
     v_head_dim: int = 128
     """Dimension of the head in the V projection."""
 
+    mla_native_v_head_dim: bool = False
+    """Whether to hand the attention backend its native V width on the thd path instead of
+    padding V up to the QK width. The pad and the matching output trim cancel out, so skipping
+    them only saves work: a narrower V through the kernel, and no pad or trim. In some
+    configurations the padded shape also gets no fused backend and falls back to a slower one."""
+
     normalization: str = "RMSNorm"
     """Default normalization layer for MLA models is RMSNorm."""
 
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -2980,11 +2980,10 @@ class MLATransformerConfig(TransformerConfig):
     v_head_dim: int = 128
     """Dimension of the head in the V projection."""
 
-    mla_native_v_head_dim: bool = False
-    """Whether to hand the attention backend its native V width on the thd path instead of
-    padding V up to the QK width. The pad and the matching output trim cancel out, so skipping
-    them only saves work: a narrower V through the kernel, and no pad or trim. In some
-    configurations the padded shape also gets no fused backend and falls back to a slower one."""
+    mla_native_v_head_dim: bool = True
+    """Pass V at its native width on the thd path instead of padding it to the QK width.
+    Faster from compute capability 9.0. Below that no backend has a kernel for unequal QK
+    and V widths, so attention falls back to the unfused path: set it False there."""
 
     normalization: str = "RMSNorm"
     """Default normalization layer for MLA models is RMSNorm."""
--- a/tests/unit_tests/transformer/test_multi_latent_attention.py
+++ b/tests/unit_tests/transformer/test_multi_latent_attention.py
@@ -438,6 +438,66 @@ class TestParallelMLAAttention:
                 assert output.shape[2] == transformer_config.hidden_size
                 assert bias.shape[0] == transformer_config.hidden_size
 
+    @pytest.mark.parametrize("mla_native_v_head_dim", [False, True])
+    def test_gpu_forward_thd_native_v_head_dim(self, mla_native_v_head_dim):
+        """Test that mla_native_v_head_dim controls the thd V pad without changing the output."""
+        if is_te_min_version("1.10.0"):
+            transformer_config = MLATransformerConfig(
+                num_layers=2,
+                hidden_size=12,
+                num_attention_heads=4,
+                use_cpu_initialization=True,
+                q_lora_rank=32,
+                kv_lora_rank=32,
+                qk_head_dim=128,
+                v_head_dim=64,
+                qk_pos_emb_head_dim=64,
+                mla_native_v_head_dim=mla_native_v_head_dim,
+                rope_type=self.transformer_config.rope_type,
+                rotary_base=self.transformer_config.rotary_base,
+                original_max_position_embeddings=self.transformer_config.original_max_position_embeddings,
+            )
+            attention = MLASelfAttention(
+                transformer_config,
+                get_mla_self_attn_submodules(),
+                layer_number=1,
+                attn_mask_type=AttnMaskType.causal,
+            )
+
+            sequence_length = 32
+            micro_batch_size = 1
+
+            attention.cuda().bfloat16()
+
+            hidden_states = torch.ones(
+                (sequence_length, micro_batch_size, transformer_config.hidden_size)
+            )
+            hidden_states = hidden_states.cuda().bfloat16()
+
+            with mock.patch.dict(
+                os.environ, {"NVTE_FUSED_ATTN": "1", "NVTE_FLASH_ATTN": "0"}, clear=False
+            ):
+                packed_seq_params = make_test_packed_seq_params(sequence_length=sequence_length)
+                query, _, value, _, _ = attention.get_query_key_value_tensors(
+                    hidden_states, None, None, packed_seq_params, None
+                )
+                assert query.shape[-1] != value.shape[-1]
+
+                prepared, need_v_pad, orig_v_dim, _ = mla_module._prepare_mla_core_attention_value(
+                    attention, query, value, packed_seq_params
+                )
+                assert need_v_pad is not mla_native_v_head_dim
+                assert prepared.shape[-1] == (
+                    orig_v_dim if mla_native_v_head_dim else query.shape[-1]
+                )
+
+                output, bias = attention(hidden_states, None, packed_seq_params=packed_seq_params)
+
+                assert output.shape[0] == sequence_length
+                assert output.shape[1] == micro_batch_size
+                assert output.shape[2] == transformer_config.hidden_size
+                assert bias.shape[0] == transformer_config.hidden_size
+
     def test_checkpointed_gpu_forward(self):
         if is_te_min_version("1.10.0"):
             transformer_config = self.transformer_config
--- a/tests/unit_tests/transformer/test_multi_latent_attention.py
+++ b/tests/unit_tests/transformer/test_multi_latent_attention.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
 
 import os
+import warnings
 from inspect import signature
 from unittest import mock
 
@@ -388,6 +389,8 @@ class TestParallelMLAAttention:
                 qk_head_dim=128,
                 v_head_dim=64,
                 qk_pos_emb_head_dim=64,
+                # Pinned: this test covers the padded path, which is no longer the default.
+                mla_native_v_head_dim=False,
                 rope_type=self.transformer_config.rope_type,
                 rotary_base=self.transformer_config.rotary_base,
                 original_max_position_embeddings=self.transformer_config.original_max_position_embeddings,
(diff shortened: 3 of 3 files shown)


LitLeo:
One concern is that mla_native_v_head_dim currently defaults to False, so the default THD behavior still pads V from 128 to 192.
It make sense to default mla_native_v_head_dim to True, while keeping False as an explicit compatibility fallback for older or unsupported backends.

janEbert (approved):
I agree that the decision on whether `mla_native_v_head_dim` should be True or False by default isn't easy. If we could automate the check, that would be ideal, but I don't see a good solution at this point without maybe exposing an explicit check from TE.

bzantium:
Good question, and it deserved a measurement rather than an opinion, so I ran one. The answer is that the default cannot flip: on A100 the native width has no fused kernel at all and falls back to the unfused backend.

Core-attention forward only, DeepSeek-V2-Lite geometry (`qk_head_dim=128`, `qk_pos_emb_head_dim=64`, `v_head_dim=128`, so QK is 192 wide and V is 128), 4096 tokens in 4 packs, bf16, 20 iterations after 3 warmup, same `nvidia/nemo:26.04` container on both machines. Times are per call.

**A100 (sm80)**

| backend | padded | native |
|---|---|---|
| `NVTE_FUSED_ATTN=1 NVTE_FLASH_ATTN=0` | 130.76 ms | 113.65 ms |
| `NVTE_FUSED_ATTN=0 NVTE_FLASH_ATTN=1` | **0.96 ms** | 113.60 ms |
| both enabled (what a real run does) | **0.96 ms** | **113.68 ms** |

**B300 (sm103)**

| backend | padded | native |
|---|---|---|
| `NVTE_FUSED_ATTN=1 NVTE_FLASH_ATTN=0` | 93.74 ms | **0.35 ms** |
| `NVTE_FUSED_ATTN=0 NVTE_FLASH_ATTN=1` | **0.45 ms** | 82.12 ms |
| both enabled | **0.45 ms** | **0.35 ms** |

The 80-130 ms entries are the unfused fallback — note that on A100 fused-native and flash-native land within 0.05 ms of each other, which is what you would expect if neither backend accepted the shape and both dropped to the same PyTorch path.

Read off the row that matters, the one with both backends enabled:

- **sm103: native is 1.29x faster.** Something in the fused path takes the unequal widths and the padded shape does not reach it.
- **sm80: native is 118x slower.** Neither backend produced a fast kernel for the unequal widths, so the pad is what buys A100 a fast kernel in the first place.

I am reporting the timings rather than the kernel selection: I inferred the fallback from the shape of the numbers rather than from backend logs, so treat "which kernel ran" as the likely explanation and the times as the measurement.

So the flag is not "skip work that cancels out" everywhere — it decides which kernel is reachable, and which one that is depends on the architecture. Defaulting it to `True` would leave every A100 MLA user two orders of magnitude slower in attention, with no error to tell them why. That is a worse failure than the one I was trying to avoid by keeping it opt-in.

What I will change based on this: the config docstring currently reads as though skipping the pad is unconditionally a saving, which the numbers above say is wrong. It should state that the flag is a win where the backend has a kernel for unequal QK and V widths and a large loss where it does not, and name the hardware I have actually measured.

If a maintainer wants this to be automatic rather than a knob, the honest way is for the selection to ask the backend whether it will fuse the given shape and choose accordingly, rather than for the config to guess. I did not find a way to query that from TE without running the layer, but I am happy to follow up if someone knows of one.

H200 (sm90) is missing: the container build on that cluster has not finished and the machine is currently unreachable. I will add the row when I can, since sm90 is the shape most people will actually be running, but it does not change the conclusion — sm80 alone rules the default out.

LitLeo:
> Good question, and it deserved a measurement rather than an opinion, so I ran one. The answer is that the default cannot flip: on A100 the native width has no fused kernel at all and falls back to the unfused backend.
> 
> Core-attention forward only, DeepSeek-V2-Lite geometry (`qk_head_dim=128`, `qk_pos_emb_head_dim=64`, `v_head_dim=128`, so QK is 192 wide and V is 128), 4096 tokens in 4 packs, bf16, 20 iterations after 3 warmup, same `nvidia/nemo:26.04` container on both machines. Times are per call.
> 
> **A100 (sm80)**
> 
> backend	padded	native
> `NVTE_FUSED_ATTN=1 NVTE_FLASH_ATTN=0`	130.76 ms	113.65 ms
> `NVTE_FUSED_ATTN=0 NVTE_FLASH_ATTN=1`	**0.96 ms**	113.60 ms
> both enabled (what a real run does)	**0.96 ms**	**113.68 ms**
> **B300 (sm103)**
> 
> backend	padded	native
> `NVTE_FUSED_ATTN=1 NVTE_FLASH_ATTN=0`	93.74 ms	**0.35 ms**
> `NVTE_FUSED_ATTN=0 NVTE_FLASH_ATTN=1`	**0.45 ms**	82.12 ms
> both enabled	**0.45 ms**	**0.35 ms**
> The 80-130 ms entries are the unfused fallback — note that on A100 fused-native and flash-native land within 0.05 ms of each other, which is what you would expect if neither backend accepted the shape and both dropped to the same PyTorch path.
> 
> Read off the row that matters, the one with both backends enabled:
> 
> * **sm103: native is 1.29x faster.** Something in the fused path takes the unequal widths and the padded shape does not reach it.
> * **sm80: native is 118x slower.** Neither backend produced a fast kernel for the unequal widths, so the pad is what buys A100 a fast kernel in the first place.
> 
> I am reporting the timings rather than the kernel selection: I inferred the fallback from the shape of the numbers rather than from backend logs, so treat "which kernel ran" as the likely explanation and the times as the measurement.
> 
> So the flag is not "skip work that cancels out" everywhere — it decides which kernel is reachable, and which one that is depends on the architecture. Defaulting it to `True` would leave every A100 MLA user two orders of magnitude slower in attention, with no error to tell them why. That is a worse failure than the one I was trying to avoid by keeping it opt-in.
> 
> What I will change based on this: the config docstring currently reads as though skipping the pad is unconditionally a saving, which the numbers above say is wrong. It should state that the flag is a win where the backend has a kernel for unequal QK and V widths and a large loss where it does not, and name the hardware I have actually measured.
> 
> If a maintainer wants this to be automatic rather than a knob, the honest way is for the selection to ask the backend whether it will fuse the given shape and choose accordingly, rather than for the config to guess. I did not find a way to query that from TE without running the layer, but I am happy to follow up if someone knows of one.
> 
> H200 (sm90) is missing: the container build on that cluster has not finished and the machine is currently unreachable. I will add the row when I can, since sm90 is the shape most people will actually be running, but it does not change the conclusion — sm80 alone rules the default out.

A100 GPUs are already relatively old and are gradually being phased out; Hopper and newer GPUs are now becoming mainstream. Starting with H100, both FlashAttention and
  cuDNN fused attention have ways to support the native V head dimension.

  In a few years, once A100 GPUs have largely been retired, this configuration option could be removed altogether. If it defaults to False, most users will have to
  explicitly set and maintain an additional configuration option.

  To accommodate A100 users, we could instead emit a warning when mla_native_v_head_dim=True is used on devices with compute capability below SM90.

bzantium:
H200 is done, and it agrees with you rather than with me. Same container and geometry as before. This is the row a real run takes, with both backends enabled:

| | padded | native | |
|---|---|---|---|
| A100 (sm80) | **0.96 ms** | 113.68 ms | native 118x slower |
| H200 (sm90) | 0.44 ms | **0.34 ms** | native 1.29x faster |
| B300 (sm103) | 0.45 ms | **0.35 ms** | native 1.29x faster |

Per backend:

| | fused only | | flash only | |
|---|---|---|---|---|
| | padded | native | padded | native |
| A100 (sm80) | 130.76 ms | 113.65 ms | 0.96 ms | 113.60 ms |
| H200 (sm90) | 0.44 ms | **0.34 ms** | 0.50 ms | 77.76 ms |
| B300 (sm103) | 93.74 ms | **0.35 ms** | 0.45 ms | 82.12 ms |

(The two 0.96 ms figures on sm80 are the same FlashAttention kernel running in both cases, not a copied number.)

cuDNN takes the native width from sm90 on. FlashAttention does not, on any of the three. That is why the pad is what makes A100 fast.

So my "the default cannot flip" was too strong. It rests entirely on sm80, and you are right that sm80 is on the way out. The narrower claim the numbers support: sm80 has no fast kernel for the native width, so there the flag is a 118x regression with nothing in the logs to explain it.

Your warning suggestion handles that, and it is better than what I proposed. An opt-in flag nobody finds helps nobody. A default plus a warning below sm90 puts the cost on the shrinking set of users it hurts, and makes it visible.

Pushed as `9500f5b`. The default is now `True`. A one-shot warning fires when the native width is in use below compute capability 9.0. The config docstring read as though skipping the pad were always a saving, and now says which capability it helps on and which it hurts.

`test_gpu_forward_thd_qv_head_dim_mismatch` is pinned to `False`, since it exists to cover the padded path and that is no longer what the default exercises. Two new tests cover the warning: that it fires below 9.0 and not at or above it, and that it fires once rather than once per layer.

Two things to flag.

The numerics change reaches everyone once it is the default. The two backends are different kernels and agree to bf16 rounding, not bit-exactly: worst relative drift 5.7e-3 across forward, dq, dk and dv on this geometry. Every existing MLA `thd` run shifts slightly on the next release. That looks like a release note to me, but it is a maintainer call.

I am reporting timings, not backend selection. The fallback is inferred from the shape of the numbers, since on A100 fused-native and flash-native land within 0.05 ms of each other. I did not read it out of backend logs.

janEbert:
Thank you for all the measurements and adjustments to the PR. I don't think it's as easy as an `sm` version check, since people could also set the `NVTE` envvars manually, and also on `sm89` I found that the unfused path is used no matter if we pad or keep V natively sized. That means that for my `sm89` setup, the native path is also faster again.

If we keep the configuration value, I'd personally be in favor of having the setting `True` without a warning, because it requires so much logic to get a truly faithful check (for reference, check out the [1.3k lines here in TE](https://github.com/NVIDIA/TransformerEngine/blob/a7aec214eb5c3969984a40c3accb6d66987d8f25/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L382-L1648)). However, ideally, we should just modify TE so it only pads if necessary and remove the padding logic from Megatron-LM entirely.

bzantium:
Agreed, and #3339 is the right place for it. The numbers from this PR line up with the heuristic there.

Mapping the three rows onto it: on A100 the native path is unfused, so it probes the padded params, finds a fused kernel and pads, which matches native being 118x slower. On H200 and B300 the native path is already fused and is not FA2, so it stays native, which matches native being 1.29x faster on both. The probe reproduces the table without an `sm` check.

That makes the knob here a stopgap at best. Either I simplify this PR to default `True` with no warning as you suggested and we drop it once a TE release carries #3339, or I close this and wait for #3339. The first only earns its keep if something needs to land before that TE version floor. Your call.

One thing worth keeping in mind for whenever #3339 takes over this decision: #3313 makes THD training with dropout prefer FA2 over FusedAttention on SM100/103, and under the new rule a native path landing on FA2 means pad. So the exact case this PR is about, skipping the V pad for MLA on THD, would go back to padding on Blackwell once dropout is on. Probably still the right trade since the slow cuDNN dropout kernels are what motivated #3313, but it does mean dropping the padding here is not a pure win in every configuration.

janEbert:
What do you think of adjusting this PR to instead remove the padding logic from Megatron-LM, so that once https://github.com/NVIDIA/TransformerEngine/pull/3339 is merged, the Megatron-LM code is cleaner? https://github.com/NVIDIA/TransformerEngine/pull/3313 is not negatively affected, since FA2 requires padding in any case (which is also handled/refactored in https://github.com/NVIDIA/TransformerEngine/pull/3339).

Basically revert the padding-related logic that https://github.com/NVIDIA/Megatron-LM/pull/3003 introduced.

bzantium:
Sounds good, that is cleaner. I will revert the padding logic #3003 added so the decision lives in TE.

I will hold it until #3339 lands unless you want it sooner.

janEbert:
However you like; we can only actually merge the change to Megatron-LM once https://github.com/NVIDIA/TransformerEngine/pull/3339 is merged and included as part of a release anyway.

bzantium:
@janEbert checking in on this one. https://github.com/NVIDIA/TransformerEngine/pull/3339 has not moved since Aug 13, so this PR is still parked on it.

Is #3339 still on track, or is it worth me pushing the Megatron-LM side now anyway? I can put up the #3003 revert so it is ready and reviewed, and we merge it whenever a TE release carries #3339. Happy to keep waiting too, just did not want it to go stale silently.

PR #6244, opened 2026-08-04
[dev] Add chunked optimizer-state and master-weight offload

hxbai:

# What does this PR do?
The existing optimizer-state offload path restores the complete optimizer state to GPU for every optimizer step, so it does not reduce peak optimizer-update memory.

This PR replaces that implementation with chunked GPU optimizer execution backed by pinned CPU canonical storage. Selected optimizer tensor states and master weights remain on CPU between updates. Tensor states are restored, updated, and offloaded in bounded chunks, while selected master weights use one full restore window.

External optimizers remain unaware of offloading; no Transformer Engine or Emerging Optimizers changes are required.

##  Acknowledgements and related work

This work benefited from our collaboration with @yanring (Zijie Yan) at Periodic Labs, including design discussions and his implementation with chunked optimizer-state and master-weight offloading.

See [#6544](https://github.com/NVIDIA/Megatron-LM/pull/6544), for a complementary implementation that chunks both optimizer states and master weights.

## Supported configurations

- Adam with `DistributedOptimizer`
- BF16 Muon with compact-layout `LayerWiseDistributedOptimizer`
- BF16 and FP8 parameter gather, including MXFP8 grad-buffer reuse
- Precision-aware FusedAdam
- Partial optimizer-state offload
- Distributed checkpoint save/load
- Checkpoint compatibility between offloaded and non-offloaded configurations

## User interface

The new options are:

```bash
--chunked-optimizer-state-offload
--optimizer-state-offload-chunk-size-mb <size>
--optimizer-state-offload-fraction <fraction>
```

## Issue tracking

Linked issue: 

### Pre-checks

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


megatron/core/optimizer/cpu_offloading/chunked_optimizer_state_offload.py
```diff
@@ -0,0 +1,988 @@
...
+                and left.device == right.device
+            )
+            if not compatible:
+                return False
+            # CUDA value comparison would synchronize the host once per chunk. Each subset
+            # starts from its own clone of the same metadata; commit the first result and only
+            # validate structural compatibility for CUDA tensor-valued fields. Snapshot/commit
+            # replaces tensor-valued group fields with clones, so identity-sensitive consumers
+            # must not retain references to those tensors; optimizer CUDA graphs are rejected.
+            return left.is_cuda or torch.equal(left, right)
```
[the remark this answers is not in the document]
hxbai: It is OK for now. I'll leave it here.

megatron/training/arguments.py
```diff
@@ -3747,15 +3786,41 @@ def _add_training_args(parser):
...
     group.add_argument(
         '--offload-optimizer-states',
         action='store_true',
         dest='offload_optimizer_states',
-        help='Offload optimizer states to CPU after each optimizer step and '
-        'reload them before the next optimizer step. '
-        'Only support TE FusedAdam optimizer.'
-        'Note that this still uses pure GPU optimizer instead of '
-        'HybridDeviceOptimizer for --optimizer-cpu-offload.',
+        help='Deprecated spelling for --chunked-optimizer-state-offload. It automatically '
```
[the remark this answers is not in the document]
hxbai: It is OK.

megatron/core/optimizer/optimizer_config.py
```diff
@@ -404,6 +415,62 @@ class OptimizerConfig:
...
+        if self.offload_optimizer_states:
+            warnings.warn(
+                "offload_optimizer_states is deprecated; use "
+                "chunked_optimizer_state_offload instead. The deprecated option now enables "
+                "the replacement with its legacy-equivalent defaults: chunk size 0 and "
+                "offload fraction 1.0 unless those settings were explicitly changed.",
+                FutureWarning,
+                stacklevel=2,
+            )
+            self.chunked_optimizer_state_offload = True
```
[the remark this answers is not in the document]
hxbai: same as above

megatron/core/optimizer/cpu_offloading/chunked_optimizer_state_offload.py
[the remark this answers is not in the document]
hxbai: same as above

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/6244/commits/a2157850d4c07704c468636f029c11ec9cd8b2c5

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/6244/commits/2c6276af70a5f5b6372aae12b0cee8d38c3b6438

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/6244/commits/707b8079608b63ce8e73a4d754535bd83efd6915

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/6244/commits/c6573f274512119106fab2810b86f0ecdbea0594

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -812,8 +809,42 @@ def __init__(
...
+                    param
+                    for group in self.shard_fp32_from_float16_groups
+                    for param in group
+                    if isinstance(param, torch.Tensor)
+                ]
+            state_dtypes = (
+                (self.config.exp_avg_dtype, self.config.exp_avg_sq_dtype)
+                if self.config.use_precision_aware_optimizer
+                else (torch.float32, torch.float32)
+            )
```
Wohox, line 842: **[Question — supported scope]** Are FP8 optimizer moments meant to be in scope?

The PR description lists "Precision-aware FusedAdam" without qualifying the moment dtype, `--exp-avg-dtype fp8` is a valid choice mapping to `torch.uint8`, and TE accepts it (`FusedAdam.__init__` allows fp32/fp16/bf16/uint8 moments). But `_is_offloadable_state` says otherwise:

> Shape is sufficient only for the currently admitted fp32/bf16 Adam moments ...

I couldn't find anything that rejects the combination — no dtype guard in `OptimizerConfig.__post_init__` or `validate_args`, and no test uses a non-fp32 moment dtype. If both are enabled, my reading of the round trip is:

1. TE stores `state[param]['exp_avg']` as a `Float8Tensor` with `shape == param.shape`, so `_is_offloadable_state` accepts it.
2. `_new_cpu_buffer` uses `tensor.dtype`, which for `Float8Tensor` is the nominal dtype (`Float8Quantizer.make_empty` defaults to fp32) — a 4x oversized pinned buffer.
3. `_prefetch_state` then installs a plain dense fp32 CUDA tensor into `state[param]['exp_avg']`, replacing the `Float8Tensor` while TE still holds `self._scales[param]['exp_avg']`.

`_state_bytes_per_param` would also sum `torch.uint8.itemsize == 1` against an fp32 staging buffer, under-bounding the chunk window 4x.

**Suggestion:** if FP8 moments are out of scope, could a guard go next to the existing optimizer assert in `OptimizerConfig.__post_init__`, and the PR description / `cpu_offloading/README.md` narrow "Precision-aware FusedAdam" to fp32/bf16 moments?

megatron/training/arguments.py
```diff
@@ -1855,6 +1855,16 @@ def validate_args(args, defaults={}):
...
 
+        if (
+            args.optimizer == 'muon'
+            and (args.chunked_optimizer_state_offload or args.offload_optimizer_states)
+            and args.optimizer_state_offload_fraction > 0.0
+        ):
+            assert args.use_distributed_optimizer, (
+                "Muon optimizer state offload requires --use-distributed-optimizer so non-Muon "
+                "parameter groups are routed through a sibling DistributedOptimizer"
+            )
```
Wohox, line 1871: **[SUGGESTION]** This guard falsely rejects the still-supported `--optimizer dist_muon` spelling.

The branch just above rewrites `args.optimizer` to `'muon'` and sets `use_layer_wise_distributed_optimizer = True` directly, without ever setting `use_distributed_optimizer` (`action='store_true'`, default `False`), so this assert fires.

Its stated goal is already met in that case: `training.py` forces `ddp_config.use_distributed_optimizer = True` whenever `use_layer_wise_distributed_optimizer` is set, which makes `use_separate_distributed_optimizer` True and routes non-Muon groups to a sibling `DistributedOptimizer`. `OptimizerConfig.__post_init__` likewise only requires `use_layer_wise_distributed_optimizer`.

Low impact — the message is self-explanatory and adding `--use-distributed-optimizer` is the documented migration anyway. But gating on the derived flag after the flip would express the actual requirement:

```python
if args.use_distributed_optimizer:
    args.use_layer_wise_distributed_optimizer = True
    args.use_distributed_optimizer = False

if (args.optimizer == 'muon' and (...) and args.optimizer_state_offload_fraction > 0.0):
    assert args.use_layer_wise_distributed_optimizer, (
        "Muon optimizer state offload requires the LayerWise distributed optimizer "
        "(--use-distributed-optimizer) so non-Muon parameter groups are routed through "
        "a sibling DistributedOptimizer"
    )
```

megatron/core/optimizer/cpu_offloading/chunked_optimizer_state_offload.py
```diff
@@ -0,0 +1,988 @@
...
+            param_bytes = self._estimated_state_bytes(param)
+            if current_params and current_bytes + param_bytes > self.chunk_size_bytes:
+                chunks.append(OptimizerStateChunk(tuple(current_params)))
+                current_params = []
+                current_bytes = 0
+            current_params.append(param)
+            current_bytes += param_bytes
+        if current_params:
+            chunks.append(OptimizerStateChunk(tuple(current_params)))
+        return tuple(chunks)
```
Wohox, line 270: **[SUGGESTION — observability]** The oversized-chunk fallback is silent; consider logging it once.

The `current_params and ...` guard is right and the atomic-parameter rule is necessary (Newton-Schulz needs whole matrices, and slicing a DistOpt shard would mean fabricating optimizer parameters). The issue is only that a parameter whose state alone exceeds `chunk_size_bytes` forms an oversized chunk with no signal: there is no logging anywhere in this module, and `validate_args` only warns for `chunk_size_mb == 0`.

That matters because `--optimizer-state-offload-chunk-size-mb` is what users size their memory budget from, and the help text promises "roughly 2x this value per active offload manager". When a single parameter overflows the target, the real steady-state window is `2 * max(chunk_size, largest_single_param_state)`.

`cpu_offloading/README.md` does mention that "a single atomic parameter bundle can exceed the target", but nothing surfaces it at runtime and `chunks` is not exposed in a form an operator can check.

Suggest a one-time `log_single_rank` at construction reporting how many parameters are individually oversized and the largest single-parameter state window.

megatron/core/optimizer/layer_wise_optimizer.py
```diff
@@ -480,6 +480,88 @@ def __init__(
...
+                shared_h2d_stream = torch.cuda.Stream() if offload_bytes_remaining > 0 else None
+                for optimizer, master_params, state_dtypes, child_bytes in child_offload_specs:
+                    if child_bytes == 0 or offload_bytes_remaining <= 0:
+                        child_fraction = 0.0
+                    elif offload_bytes_remaining >= child_bytes:
+                        child_fraction = 1.0
+                        offload_bytes_remaining -= child_bytes
+                    else:
+                        child_fraction = offload_bytes_remaining / child_bytes
+                        offload_bytes_remaining = 0
```
Wohox: **[SUGGESTION — simplification]** With offload enabled this chain always has exactly one child, so the cross-child budget split can be replaced by a single-child path.

Muon param groups all land in one `(opt_name, is_expert)` bucket — `optimizer/__init__.py:809` forces `is_expert=False` in layer-wise mode, and `emerging_optimizers.py:81-87` routes everything non-2D or embedding/output to `'adam'` — while the `assert type(raw_optimizer) is TensorParallelMuon` above rules out a second emerging-optimizer bucket. So `child_bytes` always cancels out and `child_fraction` reduces to `config.optimizer_state_offload_fraction`.

Suggest replacing the byte accounting and greedy loop with an explicit contract:

```python
offload_children = [opt for opt in optimizers if opt.optimizer is not None]
assert len(offload_children) <= 1, (
    "LayerWise chunked optimizer state offload assumes a single Muon child; "
    f"got {len(offload_children)}"
)
for optimizer in offload_children:
    ...  # existing assert + enable_chunked_optimizer_state_offload + deferred lifecycle
```

That also lets `_next_managed_optimizer_state_offload_index` (`:588-593`, always `{}`) and the `following` branch in `_before_child_step` (`:1096-1098`, never taken) go, along with the `offload_fraction` kwarg on `enable_chunked_optimizer_state_offload` — its docstring says it exists solely for this split.

One doc fix follows: `cpu_offloading/README.md:53-56` documents a `3 *` cross-child peak and a 768 MiB example; with a single child the real peak is `2 *`, and that is the number users size memory from.

Scope: only the LayerWise-internal multi-child path. The stream sharing in `ChainedOptimizer._share_optimizer_state_offload_streams` is genuinely reachable and should stay.

megatron/core/optimizer/layer_wise_optimizer.py
```diff
@@ -480,6 +480,88 @@ def __init__(
...
+                        param.numel()
+                        * (
+                            state_bytes_per_param
+                            + (param.element_size() if id(param) in master_param_ids else 0)
+                        )
+                        for param in unique_params.values()
+                    )
+                    child_offload_specs.append(
+                        (optimizer, master_params, state_dtypes, child_bytes)
+                    )
```
Wohox: **[SUGGESTION — observability]** On the Muon path, please emit a rank-0 log when the configured staging size cannot hold the largest parameter — here the flag can be **completely inert**, not just approximate.

Parameter granularity differs sharply between the two supported paths:

- Adam + `DistributedOptimizer`: parameters are byte-level shards of a DDP bucket, usually well under any sensible target.
- Muon + LayerWise decouple: parameters are **whole matrices**, since Newton-Schulz runs on the full matrix on one rank. A DeepSeek-V3 dense MLP at `7168 x 18432` has a 528 MiB fp32 momentum in one atomic unit.

If `chunk_size_bytes` falls below the *smallest* selected parameter's state, `_build_chunks` flushes on every iteration and each chunk ends up holding exactly one parameter. The plan is then set purely by parameter granularity, and every value from `1` up to the minimum parameter size produces an identical result: `64`, `256` and `16` all behave the same, with a real steady-state window of `2 x 528 MiB` in the example above. Nothing tells the user.

The largest single-parameter window is one pass over the child's `param_groups`:

```python
largest_state_bytes = max(
    (param.numel() * state_bytes_per_param for param in unique_params.values()), default=0
)
```

Suggest a `log_single_rank(logger, logging.WARNING, ...)` when `0 < chunk_size_bytes < largest_state_bytes`, naming both numbers. Reporting `largest_state_bytes` unconditionally at INFO would also help operators size the flag in the first place, since it is the real floor and is not derivable from the config.

megatron/core/optimizer/optimizer_config.py
```diff
@@ -404,6 +415,62 @@ class OptimizerConfig:
...
+            self.offload_optimizer_states = False
+
+        assert (
+            self.optimizer_state_offload_chunk_size_mb >= 0
+        ), "optimizer_state_offload_chunk_size_mb must be non-negative"
+        assert (
+            0.0 <= self.optimizer_state_offload_fraction <= 1.0
+        ), "optimizer_state_offload_fraction must be in [0.0, 1.0]"
+
+        if self.chunked_optimizer_state_offload and self.optimizer_state_offload_fraction > 0.0:
```
FDecaYed, line 438: [P1] Please assert out the unsupported precision-aware dtypes here rather than trying to support them in this PR.

`--chunked-optimizer-state-offload` currently accepts `main_params_dtype=fp16` and `exp_avg_dtype`/`exp_avg_sq_dtype` of `fp16` or `fp8`, but the offloader only preserves ordinary FP32/BF16 tensor state safely. For FP8 moments, `_new_cpu_buffer()` replaces TE’s `Float8Tensor` with a plain tensor and loses its quantizer metadata, so the next FusedAdam step fails. For FP16 moments/master weights, TE keeps the corresponding scale tensors on CUDA; once the state is CPU-canonical, checkpoint serialization attempts to combine CPU state with CUDA scales and fails with a device mismatch.

I suggest validating the precision-aware combination as:
- `main_params_dtype == torch.float32`
- `exp_avg_dtype` and `exp_avg_sq_dtype` in `(torch.float32, torch.bfloat16)`

FP8 model parameter gather can remain supported; this restriction is only for FP8/FP16 optimizer state storage. Please add a config-validation test for these rejected combinations.

hxbai: added the checks

FDecaYed (approved):

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/6244/commits/b9f7b6b5332af92f564f3fdb46064696a14e5b13

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/6244/commits/8bfbea070463674d994983f920b8081af28f3d8f

megatron/core/optimizer/distrib_optimizer.py
Wohox: This is resolved, same as Deyu's comment.

megatron/training/arguments.py
hxbai: fixed the checks

megatron/core/optimizer/cpu_offloading/chunked_optimizer_state_offload.py
hxbai: added the warning logs

megatron/core/optimizer/layer_wise_optimizer.py
hxbai: I simplified this to the supported single-child contract.

LayerWise offload now asserts that there is at most one non-empty TensorParallelMuon child and applies optimizer_state_offload_fraction directly to that child. I removed the cross-child byte-budget split, the per-wrapper offload_fraction override, and the unreachable next-child prefetch path.

Empty local Muon shards are still skipped, so ranks without locally owned Muon parameters do not construct an empty offloader or unnecessary copy streams. The outer ChainedOptimizer stream-sharing path remains unchanged.

I also updated the memory documentation: the LayerWise manager itself uses the normal two staging slots, while a split Muon + sibling DistributedOptimizer chain can still reach an aggregate window of roughly 3× the chunk size.

megatron/core/optimizer/layer_wise_optimizer.py
hxbai: Added the logs in `_build_chunks` as the one above.

Wohox (approved):
LGTM, thanks!

hxbai:
/ok to test https://github.com/NVIDIA/Megatron-LM/pull/6244/commits/2de7461a68ef8a0d4194b4659e64fb78e9a8063d

PR #6257, opened 2026-08-04
Document the parallel_state deprecation and its replacement

Connor-XY:
Global `parallel_state` accessors cannot select an independent model's parallel grid. Document caller-provided `ProcessGroupCollection` / `ProcessGroup` arguments as the migration path, mark 56 accessors with replacement guidance, and clarify where the compatibility shim remains appropriate.

Include current dense/expert and GTP-remat group mappings, global versus group-local rank conversion, optional-group and initialization semantics, and the limits of physical pipeline-stage helpers. This is documentation-only: executable ASTs for both Python files are unchanged from main (`d564dd01de45`).

Validation: Black 26.3.0, isort, Ruff, pylint, syntax, and `git diff --check` pass. The local mypy run reports the same 13 dependency/type diagnostics as main. No GPU execution is needed for this documentation-only change.

Related to #6307.

### DIFF
--- a/docs/developer/contribute.md
+++ b/docs/developer/contribute.md
@@ -47,6 +47,24 @@ File any bugs you find, keeping the following in mind:
 - Include commented-out code.
 - Attempt large architectural changes without first opening an issue to discuss.
 
+### Process groups in `megatron/core`
+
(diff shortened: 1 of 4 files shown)


Connor-XY:

PR #6258, opened 2026-08-04
Add a CI ratchet for global process-group reads in megatron/core

Connor-XY:
Core library code still contains calls that read the single global parallel grid. Add a stdlib-only CI ratchet to catch new direct accessors and `ProcessGroupCollection.use_mpu_process_groups()` calls while retaining the current migration baseline.

The baseline records 211 calls across 54 files (190 accessors and 21 compatibility shims). Entries use enclosing function/class and accessor counts, so line shifts do not require a refresh and additional identical calls still fail. `--update` removes stale entries and refuses additions; syntax errors also fail the scan. Imports and import aliases are resolved in their enclosing scopes, with parameters and simple assignments shadowing outer bindings. Global rank-list accessors are included. Dynamic lookups and aliases created through assignment are outside the checker's scope.

Run the checker and its standalone stdlib regression tests on synthetic PR branches and merge groups. Existing lifecycle functions, virtual-pipeline/memory-buffer accessors, `parallel_state.py`, and `process_groups_config.py` remain outside this gate.

Baseline audit against the old PR head: no newly grandfathered sites; PR #6888 removed the `parallel_state.get_data_parallel_rank()` read in `LanguageModule.tie_embeddings_and_output_weights_state_dict`, reducing the old 212-site baseline to 211. The comparison was repeated using the new scoped identities to ensure totals do not hide additions or moves. The three previously missed global-rank-list accessors have no current production callers, so detecting them does not grow the baseline.

Validation: 26 stdlib tests passed on Python 3.12.13; current-tree checker passed; isort, Black 26.3.0, Ruff, Pylint with CLI print allowed, workflow YAML parsing, and `git diff --check` passed. The pinned reusable preflight was read to verify outputs and read permissions. CI itself has not run for this refreshed commit.

Related: #6307.

The allowlist remains review-controlled: direct manual JSON additions can allow new calls. The checker rejects CLI-generated additions but does not compare the committed allowlist against the PR base.

### DIFF
--- a/.github/workflows/process-group-usage-check.yml
+++ b/.github/workflows/process-group-usage-check.yml
@@ -1,0 +1,49 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Process group usage check
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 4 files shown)


Connor-XY:

PR #6279, opened 2026-08-05
[Dev] Add TileLang DSA and indexer weights projection precision controls

buptzyb:
## Related PRs

- Main counterpart: #6842
- TileLang DSA source: #5049

## Summary

- Port the TileLang DSA backend from NVIDIA/Megatron-LM PR 5049 so the dev branch has a fused backend that accepts FP32 indexer projection weights.
- Add independent controls for whether `linear_weights_proj` participates in FP8/FP4 quantization and whether its output is BF16 or FP32, while preserving the existing defaults.
- Implement the non-quantized FP32-output path with BF16 parameters/inputs, TE `general_gemm` when supported, and a cached true-FP32 fallback.
- Preserve immediate/deferred WGrad, gradient-accumulation fusion, FSDP main-grad buffers, CUDA Graph capture, and checkpoint keys.
- Keep the cuDNN DSA backend on its required BF16 weights-tensor contract.

## Testing

- Megatron-LM autoformat check against `dev` (Black, isort, Pylint, Ruff): passed.
- AWS-PDX B300 precision suite: 22 passed, 1 skipped on each of 8 ranks; the skipped cuDNN parity case requires `flash_mla`, which was unavailable in that CI image.
- AWS-PDX B300 runtime probe: TE `general_gemm` selected on all 8 ranks for BF16 operands with FP32 output; no `torch.mm` fallback.
- Full-dependency cuDNN validation and 16-GPU GB200 proxy performance/NSys runs will be added while the PR is draft.


megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -1218,6 +1227,153 @@ class DSAttentionSubmodules:
...
+                getattr(linear, "wgrad_store", None) is not None
+                and linear.wgrad_store.delay_wgrad_compute()
+            ):
+                wgrad_func = functools.partial(
+                    _dsa_weights_proj_wgrad,
+                    weight=weight,
+                    linear=linear,
+                    is_first_microbatch=ctx.is_first_microbatch,
+                )
+                linear.wgrad_store.put([x, grad_output], wgrad_func)
```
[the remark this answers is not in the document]
buptzyb: Partly fixed and clarified in 147e65b. wgrad_store is production-reachable: TELinear inherits TE Linear, TE creates the store when delay_wgrad_compute=True, and the DSA production spec uses TELinear. The grep misses it because the attribute is defined in the Transformer Engine dependency. The second point did expose a real issue: the fused-WGrad dummy was only produced in the eager branch. Dummy creation now happens after the eager or deferred choice, so the delayed callback preserves the DDP hook contract. Added test_weights_projection_deferred_fused_wgrad_preserves_dummy_for_hook.

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -1218,6 +1227,153 @@ class DSAttentionSubmodules:
...
+            return output.reshape(*x_shape[:-1], weight.size(0)), True
+        except RuntimeError as error:
+            if not _dsa_weights_proj_te_gemm_is_unsupported(error):
+                raise
+            te_gemm_supported = False
+
+    # Do not emulate the requested precision by casting a low-precision output.
+    # Casting both operands makes the fallback a genuine FP32 linear operation.
+    output = torch.mm(x_2d.float(), weight.float().t())
+    return output.reshape(*x_shape[:-1], weight.size(0)), te_gemm_supported
```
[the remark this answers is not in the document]
buptzyb: Fixed in 147e65b. The generic markers not supported, unsupported, and not implemented were removed. Fallback is now limited to CUBLAS_STATUS_NOT_SUPPORTED plus the two known TE algorithm-selection capability failures; unrelated RuntimeErrors are re-raised. The fallback emits a rank-filtered warning once. Tests cover accepted capability errors, unrelated-error propagation, fallback caching, and warn-once behavior.

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -1322,22 +1478,43 @@ def __init__(
...
+            get_fp8_disabled_context(self.config)
+            if not self.config.dsa_indexer_weights_proj_use_quantization
+            else nullcontext()
+        )
+        with weights_proj_context:
+            if self.config.dsa_indexer_weights_proj_output_dtype == "fp32":
+                return _dsa_weights_projection_fp32(x, self)
+
+            weights, _ = self.linear_weights_proj(x)
+            return weights.to(dtype=torch.bfloat16)
```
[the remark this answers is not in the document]
buptzyb: Fixed in 147e65b. Both field docstrings now say that they apply only to DSAIndexer, and explicitly document that CSAIndexer keeps its FP8-disabled BF16 projection.

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -1218,6 +1227,153 @@ class DSAttentionSubmodules:
...
+                main_grad.copy_(grad_weight)
+            else:
+                main_grad.add_(grad_weight)
+            weight.main_grad = main_grad
+            if hasattr(weight, "grad_added_to_main_grad"):
+                weight.grad_added_to_main_grad = True
+            # The delayed-WGrad owner ignores this return value when accumulation is fused.
+            return grad_weight, None
+
+    return grad_weight.to(dtype=weight.dtype), None
```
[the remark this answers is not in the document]
buptzyb: Keeping this behavior intentionally. The relevant reference is TE Linear, not MCore tensor_parallel.layers: TE uses is_first_microbatch=True to overwrite main_grad for the first microbatch and accumulates subsequent microbatches; overwrite_main_grad remains the FSDP override. This custom FP32 autograd path bypasses TE forward and backward, so it must capture and advance that TE-owned flag itself. Also, local use_quantization=False does not imply that the enclosing model has no FP8 or FP4 recipe: set_is_first_microbatch() is still rearmed by the global config in that supported combination. The accumulation regression test verifies first-call overwrite followed by accumulation.

megatron/core/transformer/transformer_config.py
```diff
@@ -1665,6 +1675,32 @@ def __post_init__(self):
...
+                    "quantization is configured. Quantized TELinear does not expose a "
+                    "guaranteed true-FP32 output contract for this projection."
+                )
+            if (
+                self.dsa_indexer_weights_proj_output_dtype == "fp32"
+                and self.dsa_kernel_backend == "cudnn"
+            ):
+                raise ValueError(
+                    "dsa_indexer_weights_proj_output_dtype='fp32' is not supported by "
+                    "dsa_kernel_backend='cudnn', which requires a BF16 indexer weights tensor. "
```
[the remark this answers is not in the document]
buptzyb: Fixed in 147e65b. The three precision-contract checks now run in the unconditional part of TransformerConfig.__post_init__, instead of either attention-variant branch. This covers dsa, dsv4_hybrid, and a DSAIndexer selected through a custom ModuleSpec. Regression tests cover validation both with the hybrid variant and when the indexer is not selected.

megatron/core/transformer/experimental_attention_variant/dsa.py
```diff
@@ -1218,6 +1232,166 @@ class DSAttentionSubmodules:
...
+        return grad_input, grad_weight, None, None
+
+
+def _dsa_weights_projection_fp32(x: torch.Tensor, indexer: torch.nn.Module) -> torch.Tensor:
+    """Apply the non-quantized DSA indexer projection with a true FP32 output."""
+    linear = indexer.linear_weights_proj
+    weight = getattr(linear, "weight", None)
+    if weight is None:
+        raise RuntimeError("DSA indexer linear_weights_proj must expose a weight parameter.")
+    return _DSAWeightsProjection.apply(x, weight, linear, indexer)
```
hxbai, line 1399: This direct Function.apply call bypasses linear_weights_proj(...), so the module’s forward pre-hooks never run. With DDP overlap_param_gather, that hook calls finish_param_sync; fine-grained FSDP similarly relies on the leaf-module hook to unshard its parameters. The GEMM can therefore race with parameter gathering or read sharded storage.
Please invoke an equivalent of ensure_params_ready([weight]) before the GEMM and add coverage for overlap-param-gather/fine-grained FSDP. The main counterpart already includes this guard and a ready -> gemm ordering test.

megatron/core/transformer/transformer_config.py
```diff
@@ -1688,6 +1700,33 @@ def __post_init__(self):
                 )
                 self.dsa_kernel_backend = legacy_backend
 
+        if self.dsa_indexer_weights_proj_output_dtype not in ("bf16", "fp32"):
+            raise ValueError(
+                "dsa_indexer_weights_proj_output_dtype must be either 'bf16' or 'fp32', got "
+                f"{self.dsa_indexer_weights_proj_output_dtype!r}."
+            )
+        quantization_configured = bool(self.fp8 or self.fp4 or self.fp8_param or self.fp4_param)
```
hxbai: This validation only considers the global fp8, fp4, fp8_param, and fp4_param fields. However, GPT/Hybrid model initialization can later assign a matching per-module recipe to linear_weights_proj.te_quant_params.
The FP32 path then bypasses TELinear.forward, where that recipe would normally be applied. Consequently, a recipe that quantizes this projection can pass validation with use_quantization=True and output_dtype="fp32", but silently execute an unquantized direct GEMM.
Please either reject the combination based on the linear’s effective quantization configuration or implement an explicit recipe-aware override, with a regression test using quant_recipe.

megatron/core/transformer/experimental_attention_variant/ops/indexer.py
```diff
@@ -0,0 +1,131 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
hxbai, line 1: Can we change the `ops` dir name to a more meaningful one?

megatron/core/transformer/experimental_attention_variant/dsa.py
buptzyb: Fixed. Port #6388 from main and calls `ensure_params_ready([weight])` immediately before the direct FP32 GEMM.

For fine-grained M-FSDP v1, the FP32 path now gathers nested indexer parameters at the `DSAttention` boundary, since `forward_before_topk()` bypasses the indexer and projection module hooks.

megatron/core/transformer/transformer_config.py
buptzyb: Fixed. The config now rejects `use_quantization=True` with `output_dtype="fp32"` independently of the global FP8/FP4 flags, since a matching per-module `quant_recipe` can still quantize this TELinear.

megatron/core/transformer/experimental_attention_variant/ops/indexer.py
buptzyb: This is from the ported PR #5049. Should we better make the dir name change on the main branch?

hxbai: I see. It's OK for now.

hxbai (approved):

PR #4550, opened 2026-04-30
[training migration] Migrate mamba builder

maanug-nv:
# What does this PR do ?
This PR migrates the ModelConfig+ModelBuilder system built in Megatron-Bridge (https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2241), including the config and builder for Mamba/Hybrid model.
See original PR for more design details.

This should replace the existing `hybrid_builders.py` method of initializing MCore Hybrid models.

#4656 (which is still WIP) demonstrates how this will abstraction will be used and integrated into the setup/initialization phase of the training code.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

## Issue tracking

For PRs from open-source community contributors:

- **New features**: a linked issue is **required**. Please open a [feature request](https://github.com/NVIDIA/Megatron-LM/issues/new?template=feature_request.md) and reference it here before submitting the PR.
- **Small updates (bug fixes, minor improvements)**: a linked issue is **recommended** and will accelerate the PR review process.

Linked issue: 

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

[...]

PR #6286, opened 2026-08-05
Apply the layerwise param layout on the ModelBuilder DDP path

deepakn94:
## Summary

Config-container entry points (`pretrain_hybrid.py`, `pretrain_gpt.py`, …) wrap DDP through `ModelBuilder.build_distributed_models()` rather than `get_model()`. That path, in `megatron/training/models/dist_utils.py`, has no layerwise-optimizer handling: it never performs the `ddp_config.use_distributed_optimizer = True` restore that `wrap_model_chunks_with_ddp()` does at `megatron/training/training.py:1638`, and it only computes a `full_param_layout` when `use_distributed_optimizer` is already `True`.

Since `validate_args()` converts `--use-distributed-optimizer` into `use_layer_wise_distributed_optimizer` and clears the original (`megatron/training/arguments.py:1557-1559`), layerwise runs launched from these entry points silently fall back to the legacy path: `_shard_params_ping_pong` sharding, `allgather_params()` param sync, and an all-reduce instead of a reduce-scatter. It also trips an assert in `_ParamAndGradBucketGroup.start_grad_sync()` when combined with `--ddp-reduce-scatter-with-fp32-accumulation`, because that assert requires the reduce-scatter branch to have run:

```
File "megatron/core/distributed/param_and_grad_buffer.py", line 722, in start_grad_sync
    assert grad_reduce_handle is not None
AssertionError
```

The two DDP-wrapping paths were introduced independently on the same day — `wrap_model_chunks_with_ddp` in #4509 and `dist_utils._ddp_wrap` in #4550 — and diverged when #5516 routed config-container scripts to the latter.

## Changes

- `megatron/training/models/dist_utils.py`: `_ddp_wrap()` re-enables `ddp_config.use_distributed_optimizer`, tags params for buffer routing, and selects `LayerWiseDistributedOptimizer.compute_full_param_layout` when the layerwise optimizer is active with a param layout. The two public helpers thread the flags through.
- `megatron/training/models/{base,gpt,hybrid}.py`: forward the two new kwargs, mirroring how `overlap_param_gather_with_optimizer_step` already flows.
- `megatron/training/training.py`: `setup_model_and_optimizer()` passes them from `cfg.optimizer` / `args`.
- `examples/mimo/training/builder.py`: accepts the kwargs and raises `NotImplementedError` for layerwise, since MIMO wraps via its own `wrap_active_modules_with_ddp()` and would otherwise hit the same silent fallback.

Both new parameters default to current behavior, so existing callers are unaffected.

## Test plan

- [ ] `tests/unit_tests/training/models/`
- [ ] Muon run via `pretrain_hybrid.py` logs `use_distributed_optimizer=True` at DDP setup and `Using reduce-scatter for gradient reductions`


jaredcasper (approved):

maanug-nv (approved):

deepakn94:
/ok to test [56bf3b0](https://github.com/NVIDIA/Megatron-LM/pull/6286/commits/56bf3b099f7c2a45dc8f9afd2fefd6660987eab0)

PR #6293, opened 2026-08-05
Remove global process-group reads from Megatron Core

Connor-XY:
Core attention, transformer, model, and inference paths still read the global parallel grid when callers omit process groups. Require explicit collections at the migrated constructors and carry caller-owned groups through those paths. The base `LanguageModule` compatibility fallback remains; #6303 is the separate follow-up that makes that collection mandatory.

Refreshed through main `64d156734918e73238de5cad6d97dcfaa6ec1005`. This preserves current checkpoint, GTP, MTP, and multimodal behavior while updating affected callers. Six new LLaVA vision-CP calls needed explicit groups: three omitted the required argument, and three accidentally bound the padding count to the new group parameter. Pass the owning CP group by keyword, retain the padding argument separately, migrate new helper tests, and cover static, dynamic, and temporal vision paths with global CP access forbidden.

The migration also replaces ineffective `hasattr` checks with actual collection-membership or usable-group validation. T5 forwards its collection through the base, embedding, and output layers and validates tied pipeline endpoints; that prerequisite is isolated in #7487 and should land first.

Validation for this refresh:

- Repository formatting/lint passes. Mypy is advisory and remains non-clean.
- Ten source-isolated test cases pass on each of two real CPU/Gloo ranks, using actual forward/preprocessing/helper methods, real tensors, and real collectives. Six CP cases fail at the expected missing-group error in the preserved pre-fix merged source. Tests check supplied-group identity, static padding removal, dynamic partitioning, and all three temporal gathers with exact output/media-count comparisons. Unrelated import dependencies are substituted; this is not full-runtime or GPU acceptance.
- A paired four-GB200 comparison of this head against exact main `64d156734918e73238de5cad6d97dcfaa6ec1005` passes all 15 shared cases on every rank: dynamic vision preprocessing, profiling-group selection, partition logging, and distributed CP helpers. Both additional candidate static/temporal regressions pass on every rank. This uses actual repository imports/fixtures and CUDA/NCCL collectives; all test phases, selected IDs, rank counts, matching environments, and source manifests before/after execution are verified. It is cached nv26.04 compatibility evidence (PyTorch 2.12 development build/TE 2.14), separate from current CI acceptance.
- Earlier four-GPU comparisons and T5 validation belong to their recorded older revisions. The preceding frozen combined snapshot also passes its nine-case attention/LLaVA selection and T5 comparison on eight ranks; that evidence does not establish the later changes in this head. Broader current-head constructor coverage, current CI, and downstream acceptance remain pending.

Inventory remains 162 direct reads across 43 files, versus 211 across 54 files on current main. These are syntactic counts, not proof of complete independent-grid coverage. When #6258 lands, its strict ratchet needs reviewed allowlist substitutions for the export-converter and inference-config compatibility boundaries, plus removal of stale entries.

Keep this PR in draft until its constructor-contract validation and downstream review are complete. Related to #6307.

### DIFF
--- a/examples/multimodal/model.py
+++ b/examples/multimodal/model.py
@@ -1,23 +1,35 @@
 # Copyright (c) 2024-2026, NVIDIA CORPORATION.  All rights reserved.
-import warnings
 import logging
+import warnings
 from copy import deepcopy
 
 import torch
 from config import get_language_model_config, get_vision_model_config, get_vision_projection_config
-from layer_specs import (get_layer_spec, get_layer_spec_te, get_mlp_module_spec, get_norm_mlp_module_spec_te,
-                         get_hybrid_layer_spec_te)
+from layer_specs import (
+    get_hybrid_layer_spec_te,
+    get_layer_spec,
+    get_layer_spec_te,
+    get_mlp_module_spec,
+    get_norm_mlp_module_spec_te,
+)
 
 from megatron.core.models.multimodal.llava_model import IMAGE_TOKEN, LLaVAModel
 from megatron.core.models.vision.clip_vit_model import get_num_image_embeddings
+from megatron.core.process_groups_config import ProcessGroupCollection
+from megatron.core.utils import log_single_rank
 from megatron.training import get_args, get_tokenizer, print_rank_0
 from megatron.training.arguments import core_transformer_config_from_args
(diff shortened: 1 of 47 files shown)


Connor-XY:

PR #6303, opened 2026-08-05
Require an explicit process-group collection in LanguageModule

Connor-XY:
Core language-model constructors can silently acquire the global parallel grid when their caller omits or incompletely supplies `pg_collection`. Require an explicit collection in `LanguageModule`, preserve its identity, and use its TP group directly. Missing fields are rejected using actual collection membership. Live distributed models require participating TP/CP/PP groups, including singleton dimensions; embedding nonparticipants use explicit `None`, and offline construction may use explicit `None` fields. PyTorch's `NON_GROUP_MEMBER` sentinel is rejected.

Refresh onto main `5f51602f059a1bda38ee64bf676bf9a4bda96339`. Update repository model constructors, provider factories, keyword-expanded ModelOpt calls, and MIMO ModuleSpecs; preserve caller-owned collections in compatibility providers. Remove the three invalid `pg_collection` keywords previously added to RL mock models. Reconcile T5's base/embedding/output forwarding and tied-pipeline embedding-group checks with #7487.

Caller audit: 176 named constructor calls, seven explained keyword-expanded calls, and nine model ModuleSpecs. The audit found nine newly added direct constructors, five MIMO factories, and two ModelOpt provider paths that needed updates. This is source coverage, not proof of every external factory or runtime topology.

Validation: repository formatting/lint passes; all 76 changed Python files parse. Thirty actual-source CPU control-flow checks pass. The formatter treats mypy as advisory; diagnostics remain.

Paired eight-GB200 validation of this head against baseline `5f51602f059a1bda38ee64bf676bf9a4bda96339` verifies all ranks, complete test phases, matching environments, and source manifests before/after execution:

| Selected cases | Baseline, per rank | Candidate, per rank |
| --- | --- | --- |
| GPT/Bert constructor and forward; Hybrid constructor | 5 pass | 5 pass |
| T5 shared cases | 13 pass | 13 pass |
| Additional T5 collection/pipeline regressions | Not present | 13 pass |
| Focused MCore RL utilities | 23 pass | 23 pass |
| MIMO constructor/text/image/image+audio/state dictionary | 5 pass | 5 pass |
| GPT/T5 inference | 5 pass | 5 pass |
| Additional attention/LLaVA selection | 9 pass | 9 pass |
| Custom TP1/CP1/PP8 model | 1 pass | 1 pass |
| HybridBlock non-forward cases | 2 pass | 2 pass |
| HybridBlock GPU forward, native ARM overlay | 1 pass | 1 pass |
| Same 22 LanguageModule guard tests | 18 expected failures, 4 pass | 22 pass |
| Hybrid forward variants, native ARM overlay | 6 pass | 6 pass |

The guard comparison uses the exact candidate test file over unchanged baseline production. Those tests mock process groups and detect the intended missing-validation/global-fallback defects; T5 separately exercises real pipeline embedding synchronization. Selections overlap, so row counts are not an aggregate test total.

The two weighted-model constructor cases (GPT/MoE-GPT) also pass on both revisions in a one-process distributed world. The separate eight-rank weighted-checkpoint runs both hit stale-file-handle errors during shared-directory cleanup and became incomplete; neither establishes checkpoint acceptance. #7548 independently fixes a proven nonowner-finalizer deletion defect in the shared-directory test helper; its causal role in that distributed failure still needs a controlled rerun. #7546 independently repairs the custom-Hybrid fixture's inconsistent embedding-group creation order. The TP2/CP1/PP4 attempts failed during driver temporary-directory setup before the test body and remain unvalidated.

The seven Hybrid forward cases initially failed on both revisions because the cached ARM image contains an x86-64 causal-convolution extension. Building the same causal-conv1d 1.6.1 source for ARM and installing that exact verified artifact into fresh containers makes all seven cases pass on all eight ranks on both unchanged revisions. Wheel/member/installed-extension hashes, ARM ELF, direct CUDA forward/backward reference checks, normal MCore/Mamba2 imports, every rank's actual callable/source paths, and source manifests before/after execution are verified. Original failures remain recorded. This evidence covers the selected MCore/Mamba2 path; standalone Mamba1 and other SSM paths are not validated. All results use cached nv26.04 (PyTorch 2.12 development build/TE 2.14), separate from current CI acceptance.

Landing dependencies: #6257 documents the migration, #7487 isolates the T5 prerequisite, and #6293 migrates the wider constructor paths while retaining the base fallback. The current Megatron-Bridge compatibility providers also need [NVIDIA-NeMo/Megatron-Bridge#6151](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/6151) before this mandatory contract is consumed there. That draft passes focused source-isolated propagation checks; full downstream runtime and dependency-pin acceptance remain pending. Keep this PR in draft until distributed/current-CI validation and downstream coordination are complete.

Related to #6307.

### DIFF
--- a/examples/academic_paper_scripts/detxoify_lm/generate_samples_gpt.py
+++ b/examples/academic_paper_scripts/detxoify_lm/generate_samples_gpt.py
@@ -13,6 +13,7 @@ from megatron.training import get_args
 from megatron.training import get_tokenizer
 from megatron.training import print_rank_0
 from megatron.training.checkpointing import load_checkpoint
+from megatron.core.process_groups_config import ProcessGroupCollection
 from megatron.core import mpu
 from megatron.training.arguments import parse_and_validate_args
 from megatron.training.initialize import initialize_megatron
@@ -74,7 +75,9 @@ def model_provider(pre_process=True, post_process=True) -> GPTModel:
         share_embeddings_and_output_weights=not args.untie_embeddings_and_output_weights,
         position_embedding_type=args.position_embedding_type,
         rotary_percent=args.rotary_percent
-    )
+    ,
+                pg_collection=ProcessGroupCollection.use_mpu_process_groups(),
+            )
 
     return model
 
--- a/examples/academic_paper_scripts/detxoify_lm/generate_samples_gpt.py
+++ b/examples/academic_paper_scripts/detxoify_lm/generate_samples_gpt.py
@@ -13,6 +13,7 @@ from megatron.training import get_args
 from megatron.training import get_tokenizer
 from megatron.training import print_rank_0
 from megatron.training.checkpointing import load_checkpoint
+from megatron.core.process_groups_config import ProcessGroupCollection
 from megatron.core import mpu
 from megatron.training.arguments import parse_and_validate_args
 from megatron.training.initialize import initialize_megatron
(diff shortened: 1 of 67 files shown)


Connor-XY:

PR #4741, opened 2026-05-11
[training migration] Migrate GPT builder

maanug-nv:
# What does this PR do ?
Follow-up to #4550 for GPT Model.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact the @mcore-oncall.

## Issue tracking

For PRs from open-source community contributors:

- **New features**: a linked issue is **required**. Please open a [feature request](https://github.com/NVIDIA/Megatron-LM/issues/new?template=feature_request.md) and reference it here before submitting the PR.
- **Small updates (bug fixes, minor improvements)**: a linked issue is **recommended** and will accelerate the PR review process.

Linked issue: 

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment the [@mcore-oncall](https://github.com/orgs/NVIDIA/teams/mcore-oncall) to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

Once all required reviewers have approved, the `Approved` label is applied **automatically**.

### Merge

Any member of [mcore-engineers](https://github.com/orgs/NVIDIA/teams/mcore-engineers) will be able to merge your PR.

<details>
<summary>For MRs into `dev` branch</summary>
The proposed review process for `dev` branch is under active discussion.

[...]

PR #6306, opened 2026-08-05
Propagate rope_scaling_factor from args to GPT model construction

### LINKED ISSUE #6305: [BUG] --rope-scaling-factor is silently ignored: GPT models always use the default factor of 8.0
**Describe the bug**

`--rope-scaling-factor` is parsed but never forwarded to the model. `gpt_config_from_args` sets `rope_scaling` from `--use-rope-scaling` but not the factor, so `GPTModelConfig.rope_scaling_factor` keeps its dataclass default of 8.0 regardless of the command line. The same gap exists in `gpt_builders.py` (used by `train_rl.py` and the checkpoint converter tools) and in `megatron/post_training/model_builder.py`.

Training with the Llama 3.1 factor of 8.0 works by coincidence of defaults. Any other value, e.g. 32.0 for Llama 3.2 style scaling, silently trains with 8.0. The heterogeneous checkpoint path even validates `args.rope_scaling_factor` against the HF config's `rope_scaling["factor"]`, then drops the validated value anyway.

#2902 fixed this on the old file layout but was closed without being merged, and the training migration (#4741) carried the gap into the new `gpt_config_from_args` path.

```python
import sys
from megatron.training.arguments import parse_args, validate_args
from megatron.training.argument_utils import gpt_config_from_args

sys.argv = [
    'pretrain_gpt.py',
    '--num-layers', '2', '--hidden-size', '64', '--num-attention-heads', '4',
    '--seq-length', '128', '--max-position-embeddings', '131072',
    '--micro-batch-size', '1', '--vocab-size', '128256',
    '--position-embedding-type', 'rope', '--no-rope-fusion',
    '--use-rope-scaling', '--rope-scaling-factor', '32.0',
]
args = parse_args()
validate_args(args)
cfg = gpt_config_from_args(args)
print('args.rope_scaling_factor =', args.rope_scaling_factor)
print('cfg.rope_scaling         =', cfg.rope_scaling)
print('cfg.rope_scaling_factor  =', cfg.rope_scaling_factor)
```

Output on current main (3aee84c3):

```
args.rope_scaling_factor = 32.0
cfg.rope_scaling         = True
cfg.rope_scaling_factor  = 8.0
```

**Expected behavior**

`cfg.rope_scaling_factor == 32.0`, so the configured factor reaches `RotaryEmbedding`.

**Additional context**

`pretrain_vlm.py` and `examples/multimodal/model.py` have the same issue on the LLaVA path: they pass `language_rope_scaling` but never `language_rope_scaling_factor`. That is a separate call surface and not covered by the linked PR.

Alvorecer721:

# What does this PR do?

Forwards `rope_scaling_factor` from the parsed arguments to `GPTModel`, so `--rope-scaling-factor` actually takes effect.

## Issue tracking

Linked issue: Fixes #6305

## Problem

All three GPT construction sites pass `rope_scaling=args.use_rope_scaling` but not the factor, so `GPTModel` always falls back to its default of 8.0:

- `gpt_config_from_args` in `megatron/training/argument_utils.py` (pretrain_gpt.py, train_rl.py, inference utils)
- `gpt_builder` in `gpt_builders.py` (train_rl.py, checkpoint converter tools)
- the ModelOpt builder in `megatron/post_training/model_builder.py`

Runs using the Llama 3.1 factor of 8.0 are unaffected because the defaults happen to match, which is why this went unnoticed. Anything else, e.g. 32.0 for Llama 3.2 style scaling, silently trains with 8.0.

#2902 fixed this on the old layout but was closed unmerged; the training migration (#4741) kept the gap in the new path.

## Changes

- Forward `args.rope_scaling_factor` at the three construction sites (one line each).
- Regression test for the config path: `gpt_config_from_args` with a non-default factor must produce a config carrying that factor (`tests/unit_tests/training/models/test_gpt_builder.py`).
- Regression test for the builder path: `gpt_builder` must include `rope_scaling_factor` in its `GPTModel` call (`tests/unit_tests/models/test_gpt_model.py`).

## Testing

```bash
pytest -q tests/unit_tests/training/models/test_gpt_builder.py
pytest -q "tests/unit_tests/models/test_gpt_model.py::test_gpt_builder_forwards_rope_scaling_factor"
```

Both new tests fail before the fix (the config keeps 8.0; the kwarg is missing from the call) and pass after. All 58 tests in `test_gpt_builder.py` pass.

### Pre-checks


Alvorecer721:
@NVIDIA/mcore-oncall Could someone take a look at CI vetting for this when you get a chance? Small bug fix (three one-line propagation fixes plus regression tests) — the linked issue #6305 has a minimal repro. Happy to adjust anything if needed.

Phlip79 (approved):

jaredcasper (approved):

AAnoosheh (approved):

asolergi-nv (approved):

PR #6311, opened 2026-08-06
[Dev] Recompute EP A2A overlap activations per layer instead of per model chunk

Wohox:
### Summary
(#6429 main counterpart)
PR #5869 added full activation recompute to the EP A2A overlap path
(`--overlap-moe-expert-parallel-comm`) at **model-chunk / VPP-stage** granularity:
the whole stage runs its initial forward under `no_grad`, and every layer's
activations are rebuilt at once at backward time.

This PR moves that recompute to **layer** granularity. Only each recompute
segment's input tensor survives the forward->backward gap, and a segment is
replayed with grad enabled immediately before its own backward, so at most one
segment's activations are materialized at a time.

The chunk-level path is removed rather than kept alongside — there is a single
recompute path now.

### Semantics

For decoder layers, `recompute_method` / `recompute_num_layers` retain the same
high-level grouping as the non-overlap path:

* `uniform`: decoder layers in groups of `recompute_num_layers`; one segment per
  MTP depth (`recompute_num_layers` must be 1 when MTP is present).
* `block`: the first `recompute_num_layers` decoder layers, one segment each.
  Under overlap, every MTP depth is also recomputed as its own segment; the
  non-overlap MTP path instead warns and leaves MTP eager.
* Under FP8/FP4 with `block`, overlap recomputes the first configured decoder
  layers directly. Non-overlap checkpointing may shift that window past initial
  inputs that do not require gradients.

The validation exemptions that let the overlap path leave these unset are gone,
so both flags are required exactly as on the non-overlap path.

### Why a segment object rather than a `checkpoint()` wrapper

The non-overlap path can wrap a layer in `tensor_parallel.random.checkpoint`
because a layer is an opaque callable there. Under EP A2A overlap a layer is
decomposed into `ScheduleNode`s (attn / dispatch / mlp / combine) that the 1F1B
scheduler interleaves with *other microbatches'* nodes — which is the whole point
of the overlap path. `checkpoint()` needs its region to run as one contiguous
autograd call and to backward as one contiguous unit, so the replay has to be
driven by the scheduler at segment-backward time instead.

Two things follow from N segments per chunk instead of one:

* Each segment snapshots the mutable chunk-state fields (`input_ids`,
  `position_ids`, `padding_mask`, `mtp_hidden_states`, `mhc_multistream`). These
  are mutated in place during the forward, so by backward time they hold
  end-of-chunk values rather than what the segment saw.
* `mhc_multistream` is the only cross-layer detach bridge. Backward runs in
  reverse, so the MTP segment is replayed and backwarded *before* the decoder
  segment that produces the leaf, and that replay installs a freshly detached
  tensor. The gradient is carried across the replay explicitly.

`mtp_post_process` is inside the recompute scope: excluding it breaks the
`torch.cat(mtp_hidden_states)` graph and silently drops decoder weight gradients.

### Testing

Unit tests (`tests/unit_tests/a2a_overlap/`):

* `test_schedule_chunk_1f1b.py::test_1f1b_schedule_model_chunk_full_recompute`
  compares gradients from `TransformerModelChunkSchedulePlan.run` against the
  non-overlap `gpt_model.forward` reference, asserts that recompute activations
  are released and every configured MTP layer belongs to a segment, and sweeps
  `mtp_layers` 0/1, dispatcher configs, `get_valid_fp8_flags()`, and
  `(uniform,1) (uniform,2) (block,1) (block,3)`. Only invalid
  `uniform + MTP + recompute_num_layers != 1` cases are skipped; all `block + MTP`
  cases run even though the non-overlap reference keeps MTP eager.
* `test_mhc_schedule.py::test_full_recompute_carries_the_mhc_bridge_gradient`
  compares uniform/block MTP gradients against eager execution and asserts that
  the two decoder layers plus MTP form three independent recompute segments.

EOS H100 validation with the CI `mcore-ci-dev` image:
`146 passed, 10 skipped, 0 failed` in 8m41s. The ten skips are exactly the
invalid `uniform-2 + MTP` combinations.

Note that `get_valid_fp8_flags()` is architecture-gated: `blockwise` is Hopper
only and `mxfp8` is Blackwell only, so no single CI machine exercises both.

E2E, DeepSeek-V3-Lite-deter (the deterministic benchmark recipe from
megatron-moe-scripts: TP1 PP4 EP2 CP1, VPP2, MBS1 GBS16, seq 256, 16 layers,
flex/DeepEP dispatcher, `unfused` attention, 100 iterations, one node). Every
comparison below runs each arm twice and requires both arms to reproduce
themselves before the arms are compared:

| Config | non-overlap self-repro | overlap self-repro | overlap vs non-overlap |
|---|---|---|---|
| BF16, no recompute | 100/100 | 100/100 | 100/100 (two independent pairings) |
| BF16 + full recompute (uniform, 1) | 100/100 | 100/100 | 100/100 (two independent pairings) |
| Blockwise FP8 + full recompute | 100/100 | 100/100 | 100/100 (two independent pairings) |
| MXFP8 + full recompute (Blackwell) | 100/100 | 100/100 | 100/100 (two independent pairings) |

The rows above run the recipe's NonMTP branch. Repeating them on its MTP branch
(`--mtp-num-layers 1`, 8 layers, every layer MoE, `--pipeline-model-parallel-layout
Et|(t|)*6tmL` in place of VPP) — which is what exercises `mtp_post_process` inside
the recompute scope and the `mhc_multistream` gradient carry-over:

| Config (MTP) | non-overlap self-repro | overlap self-repro | overlap vs non-overlap |
|---|---|---|---|
| BF16, no recompute | 100/100 | 100/100 | 100/100 (two independent pairings) |
| BF16 + full recompute | 100/100 | 100/100 | 100/100 (two independent pairings) |
| Blockwise FP8 + full recompute | 100/100 | 100/100 | 100/100 (two independent pairings) |

Full recompute is itself bitwise neutral in this recipe: on the non-overlap arm,
recompute on and recompute off agree 100/100.

### Throughput and memory

Measured on the DeepSeek-V3 benchmark recipe rather than the proxy config the
bitwise runs use: 14 layers / 256 experts, hidden 7168, MLA, MTP 1, seq 2048,
64x H100, TP1 PP4 EP16 CP1 with `--pipeline-model-parallel-layout "Ett|(tt|)*6mL"`,
MBS1 GBS256, bf16, alltoall dispatcher, 30 iterations. Median of the steady-state
iterations; peak is `max allocated`.

| Arm | `CUDA_DEVICE_MAX_CONNECTIONS` | TFLOP/s/GPU | Peak allocated |
|---|---|---|---|
| no overlap, no recompute | 1 | 225.1 | 68,704 MB |
| no overlap + full recompute (uniform, 1) | 1 | 184.2 | 60,154 MB |
| **overlap + full recompute (uniform, 1)** | 1 | 191.1 | 60,518 MB |
| **overlap + full recompute (uniform, 1)** | **32** | **198.5** | 60,518 MB |

Full recompute costs 18% throughput and returns 8.5 GB (-12.4%). Against that
same-recompute baseline the overlap path returns **+7.8%** of the lost
throughput for +364 MB (+0.6%) of live memory.

The connection count matters and is easy to get wrong: `CUDA_DEVICE_MAX_CONNECTIONS=1`
serializes work onto a single hardware queue, so the A2A cannot actually run
alongside compute and the overlap gain reads as +3.7% instead of +7.8%. The
benchmark recipe this was transcribed from sets 1, so anyone re-measuring from a
stock recipe will understate this path.

`memory_reserved` tells a different story from `memory_allocated`: the overlap
arm reserves ~4.7 GiB more while live bytes differ by 364 MB. Most of that is
structural, not extra data. Blocks in PyTorch's caching allocator are keyed by
the stream they were allocated on and are never handed to another stream
(`c10/cuda/CUDACachingAllocator.cpp`: "Allocations are associated with a stream.
Once freed, blocks can be re-allocated on the same stream, but not on any other
stream"), so the communication stream gets its own segments -- 1,800 MB reserved
against an 899 MB working set here. The remainder is allocator high-water slack
established before the sampled window, which the snapshot cannot attribute.

One measured overhead worth naming: one `detached` tensor per recompute segment
(the post-attention hidden state, `seq x mbs x hidden x dtype`) stays live across
the forward->backward gap. Its only owner is on the C++ side -- zero gc-visible
referrers -- so clearing Python attributes does not reclaim it. It is released at
the segment replay, so this is a peak-memory cost rather than a growing leak.


Wohox:
/ok to test [3630ffe](https://github.com/NVIDIA/Megatron-LM/pull/6311/commits/3630ffeaedb294044b3951895547b41b73da17e0)

megatron/core/transformer/moe/README.md
```diff
@@ -436,6 +436,8 @@ EP All-to-All can consume 30-40% of training time without optimization. These fe
 
 > **Requirements for EP A2A Overlap**: `expert_model_parallel_size > 1`, CUDA_DEVICE_MAX_CONNECTIONS > 1.
 
+> **Full activation recompute with EP A2A Overlap**: `--recompute-granularity full` is supported and is applied per layer segment, so `--recompute-method` and `--recompute-num-layers` carry exactly the same meaning as without `--overlap-moe-expert-parallel-comm`. Both flags are required. Earlier releases accepted them being unset (the whole model chunk was recomputed as one unit); such configs now fail validation and must set the two flags explicitly.
```
lhb8125: **[IMPORTANT Compatibility]** This says the two flags have exactly the same meaning with and without overlap, but the implementations intentionally differ in two supported cases. Under `block`, non-overlap `MultiTokenPredictionLayer._checkpointed_forward` leaves MTP eager, while `_build_recompute_segments()` always adds an MTP segment. Under FP8/FP4, non-overlap also shifts the block window via `recompute_skip_num_layers`, whereas overlap explicitly does not. Those differences affect recompute compute/memory and benchmark expectations. Please scope this statement to decoder segmentation and document the MTP and FP8/FP4 exceptions; the PR description should be updated as well (it still says block leaves MTP eager and references the deleted `test_recompute_segments.py`).

tests/unit_tests/a2a_overlap/test_schedule_chunk_1f1b.py
```diff
@@ -203,83 +227,80 @@ def test_1f1b_schedule_model_chunk_with_padding_mask(
...
+        dispatcher_type,
+        flex_backend,
+        fp8_flag,
+        recompute_method,
+        recompute_num_layers,
+    ):
+        """Layer-level full recompute matches the non-overlap reference, which checkpoints
+        through checkpointed_forward / MTP _checkpointed_forward with the same config.
+        """
+        if mtp_layers > 0 and not (recompute_method == "uniform" and recompute_num_layers == 1):
```
lhb8125: **[IMPORTANT Test Coverage]** This skip removes every block+MTP case, even though a key change in this PR is to recompute MTP under `block`. The paths remain numerically comparable: recompute should be transparent even if one path leaves MTP eager, and the new mHC parity test already makes such a comparison. As written, block+MTP is only exercised by one BF16/alltoall mHC case, so block-1 recomputed-to-eager decoder handoff and flex/FP8 variants are untested. Please skip only invalid uniform+MTP configurations (`recompute_num_layers != 1`), run the block+MTP parity cases, and assert that the overlap plan contains the MTP segment.

megatron/core/transformer/moe/README.md
Wohox: Thanks, you’re right. I scoped the documentation to the decoder grouping semantics and explicitly documented both overlap-specific differences:

- MTP is recomputed as one segment per depth under `block`, while the non-overlap path leaves it eager.
- FP8/FP4 overlap does not apply the non-overlap checkpoint path’s `recompute_skip_num_layers` window shift.

I also updated the corresponding code comments and PR description, including removing the stale `test_recompute_segments.py` reference. Fixed in `3162a52fc`.

tests/unit_tests/a2a_overlap/test_schedule_chunk_1f1b.py
Wohox: Thanks, agreed. I changed the skip so that it applies only to invalid `uniform + MTP` configurations where `recompute_num_layers != 1`.

The `block-1` and `block-3` MTP parity cases now run across alltoall, flex/HybridEP, BF16, and all FP8 recipes available on the test architecture. The test also explicitly verifies that every configured MTP layer is built and assigned to a recompute segment.

I validated the updated file on EOS H100 with the CI `mcore-ci-dev` image: `146 passed, 10 skipped, 0 failed` in 8m41s. The ten skips are exactly the invalid `uniform-2 + MTP` cases. Fixed in `3162a52fc`.

lhb8125 (approved):
LGTM

PR #6313, opened 2026-08-06
Refactor hybrid layers to use per-layer configs

Phlip79:

# What does this PR do?

See Part 2 of [design doc](https://github.com/user-attachments/files/30916074/HybridModel.pdf).

Previous PRs: https://github.com/NVIDIA/Megatron-LM/pull/6410, https://github.com/NVIDIA/Megatron-LM/pull/6419, https://github.com/NVIDIA/Megatron-LM/pull/6423.

Follow-up PR (for inference): https://github.com/NVIDIA/Megatron-LM/pull/6824

This PR changes when and how layer configs are constructed. Instead of duplicating the same `TransformerConfig` for each layer (no matter the layer type), now each layer has its own config. Even after this PR, each layer's config will be identical. Each layer config just inherits `TransformerConfig`; however, this enables two features:

- future layers will add their arguments to its respective layer class, instead of adding an argument to `TransformerConfig`
- enable heterogeneous layer construction, as now we have a list of configs
  - in a future PR, we will add the ability to construct custom layer configs

Here is a simplified view of the new training workflow construction:

legacy_hybrid


megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -30,10 +30,6 @@
 )
 from megatron.core.inference.utils import device_memory_summary, tensor_swap
 from megatron.core.models.common.embeddings.rope_utils import apply_rotary_pos_emb
-from megatron.core.models.hybrid.hybrid_layer_allocation import (
-    Symbols,
-    get_layer_maps_from_layer_type_list,
-)
```
janEbert, line 39: Why was this moved?

Phlip79: Circular dependency reasons. It's not clean right now, I will change it.

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -81,8 +88,8 @@ def __init__(
         self,
         config: TransformerConfig,
         submodules: HybridStackSubmodules,
+        layer_config_list: list[HybridLayerConfig],
         pre_process: bool = True,
-        layer_type_list: Optional[list[str]] = None,
```
Phlip79, line 96: Should keep this, but deprecate it. Create a list of configs immediately.

janEbert (commented):
One general comment: I often see `type(obj) is LayerConfig` instead of `isinstance(obj, LayerConfig)`. The latter seems more canonical to me; is there a reason to prefer the former? (Note, I'm totally fine with keeping the `type` calls, just feels slightly less pythonic.)

Ah it's to explicitly avoid child classes also matching, isn't it? Could that bite us later?

megatron/core/models/hybrid/hybrid_layer_allocation.py
```diff
@@ -38,6 +47,37 @@ def name_sorted_valid_layer_symbols(cls) -> list[str]:
...
+
+
+class _HybridLayerConfigList(list[HybridLayerConfig]):
+    """Configs cloned from one legacy stack config.
+
+    The marker lets HybridStack preserve constructor-time mutations that previously
+    affected a shared config, without coupling independently supplied config lists.
+    """
+
+    synchronize_shared_config_mutations = True
```
janEbert: This attribute is only queried once, but the code never sets it to False. Do we need the `_HybridLayerConfigList` abstraction class surrounding it? In general I like the idea of having a type alias, but it's also not used for this.

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -438,17 +440,32 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
...
+                    raise NotImplementedError("GDN layers are not supported for inference.")
+                elif type(layer_config) is MLALayerConfig:
+                    # Hybrid MLA cache mapping was not supported by the legacy path.
+                    continue
+                elif type(layer_config) in (MLPLayerConfig, MoELayerConfig):
+                    continue
+                else:
+                    raise ValueError(
+                        f"Unexpected hybrid layer config type: {type(layer_config).__name__}"
+                    )
```
janEbert: Wouldn't it make more sense to update `get_layer_maps_from_layer_type_list` to work with the `layer_config_list` instead of strings instead of this rework of this specific code?

megatron/core/models/hybrid/hybrid_layer_allocation.py
```diff
@@ -298,45 +338,62 @@ def _validate_pattern(pattern: str, pattern_name: str, allow_pipe: bool = False)
...
-            )
+    _validate_segment_layer_symbols(segment)
 
-    # Disallow Attention + MLA/DSA hybridity.
-    if Symbols.ATTENTION in segment and (Symbols.DS_ATTENTION in segment or Symbols.MLA in segment):
-        raise ValueError("Not supported to have both Attention and MLA/DSA in one model")
+    layer_configs = _HybridLayerConfigList()
+    for layer_symbol in segment:
+        layer_config = deepcopy(config)
+        layer_config.__class__ = LAYER_SYMBOL_TO_CONFIG_CLASS[layer_symbol]
```
janEbert: I personally find this too hacky and would prefer a cleaner solution, e.g., a separate function that creates the class and then populates the attributes from the deep-copied config.

janEbert (approved):
Approved since I find the design is solid, but please take a look at the comments. :)

megatron/core/inference/config.py
```diff
@@ -73,12 +74,12 @@ def from_model(
             elif ssm_states_dtype is None:
                 ssm_states_dtype = model.config.params_dtype
             mamba_chunk_size = 128
-            for layer_type, layer in zip(decoder.layer_type_list, decoder.layers):
-                if layer_type == Symbols.MAMBA and hasattr(layer, 'mixer'):
+            for layer_config, layer in zip(layer_config_list, decoder.layers):
```
janEbert: Let's use `zip(..., strict=True)` here.

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -261,8 +315,8 @@ def mamba_state_shapes_per_request(self) -> Optional[Tuple[Tuple[int], Tuple[int
         Returns the Mamba conv and ssm states shapes per input sequence
         if this block contains Mamba layers (this may not be the case with PP > 1).
         """
-        for layer_type, layer in zip(self.layer_type_list, self.layers):
-            if layer_type == LayerSymbols.MAMBA:
+        for layer_config, layer in zip(self.layer_config_list, self.layers):
```
janEbert: Let's use `zip(..., strict=True)` here.

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -367,10 +421,10 @@ def get_inner_quant_context(config, layer_number):
                     use_inner_quantization_context=(use_inner_fp8_context or use_fp4_context),
                 )
             else:
-                for layer in self.layers:
+                for layer_config, layer in zip(self.layer_config_list, self.layers):
```
janEbert: Let's use `zip(..., strict=True)` here.

tests/unit_tests/models/test_hybrid_model.py
```diff
@@ -234,9 +237,54 @@ def test_constructor(self):
...
+        assert [type(config) for config in decoder.layer_config_list] == [
+            MambaLayerConfig,
+            AttentionLayerConfig,
+            MLPLayerConfig,
+        ]
+        assert len({id(config) for config in decoder.layer_config_list}) == 3
+        assert all(config is not self.model.config for config in decoder.layer_config_list)
+        assert all(
+            layer.config is layer_config
+            for layer, layer_config in zip(decoder.layers, decoder.layer_config_list)
```
janEbert: Let's use `zip(..., strict=True)` here.

megatron/core/inference/contexts/dynamic_context.py
Phlip79: Good idea. Updated.

tests/unit_tests/ssm/test_hybrid_layer_allocation.py
```diff
@@ -840,3 +988,47 @@ def test_mixed_dsa_and_mla(self):
...
+            get_layer_maps_from_layer_config_list([object()])
+
+    def test_rejects_ambiguous_config_types(self):
+        """A config matching multiple registered layer types is ambiguous."""
+
+        class AmbiguousLayerConfig(MambaLayerConfig, AttentionLayerConfig):
+            pass
+
+        with pytest.raises(ValueError, match="Ambiguous hybrid layer config type"):
+            get_layer_maps_from_layer_config_list([object.__new__(AmbiguousLayerConfig)])
```
janEbert: This is such a nice test! Could we either (1) enforce programatically that we cannot have a layer subclass from multiple configs types or (2) add an additional test that checks non-ambiguity for all existing config types (e.g., those in `LAYER_SYMBOL_TO_CONFIG_CLASS.values()`)?
(2) is probably much simpler to add and wouldn't pollute the main code.

megatron/core/models/hybrid/hybrid_layer_allocation.py
Phlip79: Agreed. I've updated the logic. I've added a `from_config` function to `TransformerConfig`.

tests/unit_tests/ssm/test_hybrid_layer_allocation.py
Phlip79: Added this check [here](https://github.com/NVIDIA/Megatron-LM/pull/6313/changes#diff-90fcc7bb8bf6a83170ae440a5004e191bb5854a6efd0c555ef974fd3fcd64c28R618-R622).

megatron/core/models/hybrid/hybrid_layer_allocation.py
Phlip79: `_HybridLayerConfigList` has been removed.

guihong-nv (approved):

cuichenx (approved):

megatron/core/models/hybrid/layers/utils.py
```diff
@@ -0,0 +1,137 @@
...
+    if Symbols.DS_ATTENTION in segment:
+        unsupported_features.append("DSA")
+    if has_mtp:
+        unsupported_features.append("MTP")
+
+    if not config.tp_comm_overlap or not unsupported_features:
+        return
+
+    config.tp_comm_overlap = False
+    warnings.warn(
```
jaredcasper: This should be an error.

megatron/core/models/hybrid/layers/utils.py
```diff
@@ -0,0 +1,137 @@
...
+        """Return valid layer symbols sorted by their public attribute names."""
+        valid_layer_attrs = []
+        for name, value in vars(cls).items():
+            if not name.startswith('_') and value in cls.VALID_LAYERS:
+                valid_layer_attrs.append((name, value))
+        valid_layer_attrs.sort()
+        return [value for (_, value) in valid_layer_attrs]
+
+
+def create_layer_config(config: TransformerConfig, layer_symbol: str) -> TransformerConfig:
```
jaredcasper, line 57: Feels like these two functions could be cleaned up and more easily expanded using a list of tuples that you just loop through?
```
symbols_to_configs = [(Symbols.MAMBA, MambaLayerConfig),
                      (Symbols.GDN, GDNLayerConfig),
...
]
```

megatron/core/models/hybrid/layers/utils.py
Phlip79: Updated to throw error and changed function name to `validate_tp_comm_overlap`.

megatron/core/models/hybrid/layers/utils.py
Phlip79: Replaced `VALID_LAYERS` with `LAYER_CONFIG_MAP`.

jaredcasper (approved):

PR #6315, opened 2026-08-06
[Dev] feat(multimodal): migrate Qwen3.5-VL to HybridModel

BestJuly:
# What

Migrate the `examples/multimodal_dev` Qwen3.5-VL language decoder from
`GPTModel` to `HybridModel`, ahead of the deprecation of the GPTModel code path.

# Why

The Qwen3.5-VL decoder is a hybrid GatedDeltaNet / full-attention stack. On the
GPT path that structure was expressed implicitly through
`--experimental-attention-variant gated_delta_net` plus
`--linear-attention-freq 4`, with the attention/MLP ordering baked into
`TransformerLayer`. `HybridModel` expresses it explicitly through
`--hybrid-layer-pattern`, which is where this model family belongs long term.

# Changes

**Model wiring (`examples/multimodal_dev`)**

- `MultimodalModel` / `Qwen35VLModel` now take `hybrid_stack_spec` and
  `hybrid_layer_pattern` instead of a `transformer_layer_spec` and an
  `mtp_block_spec`. `get_qwen35_vl_language_spec` is removed.
- Each historical Qwen block becomes two hybrid layers: a GatedDeltaNet (`G`)
  or full-attention (`*`) layer followed by a dense (`-`) or MoE (`E`) MLP
  layer. Three GDN blocks then one attention block reproduces the
  `--linear-attention-freq 4` cadence exactly.
- MTP depths replicate the last block, matching `get_gpt_mtp_block_spec`, which
  derived the MTP spec from the final decoder layer.
- Variant configs are converted from block counts to hybrid layer counts,
  because `HybridModel` counts one layer per pattern symbol.

**mRoPE support in `HybridModel` (`megatron/core/models/hybrid/hybrid_model.py`)**

- Build a `MultimodalRotaryEmbedding` when `position_embedding_type == 'mrope'`
  and compute sectioned (optionally raw, fused-kernel) freqs in `forward`,
  mirroring `GPTModel` line for line, including the
  `apply_rope_fusion` / `rotary_interleaved` / `fused_single_qkv_rope` /
  inference conditions and the flash-decode `NotImplementedError`.
- Widen the `HybridModelConfig.position_embedding_type` literal accordingly.

**GatedDeltaNet (`megatron/core/ssm/gated_delta_net.py`)**

- Import `fla.ops.cp.build_cp_context` separately from the rest of FLA so an FLA
  build without the CP extension still enables GatedDeltaNet, and raise a
  descriptive error only when chunkwise CP is actually requested.

**Launcher (`scripts/run_qwen35_vl.sh`)**

- Derive the pattern from `NUM_LAYERS`, `NUM_EXPERTS` and the new
  `MTP_NUM_LAYERS` knob; pass the MTP arguments only when MTP is enabled.
- `NUM_LAYERS` need not be a multiple of 4: trailing blocks stay GatedDeltaNet,
  which is what `--linear-attention-freq 4` produced, so shallow proxy runs keep
  working.

# Tests

- `examples/multimodal_dev/tests/test_hybrid_migration.py` — factory wiring plus
  structural equivalence between the hybrid pattern and the former GPT layout.
  The key check cross-validates `validate_segment_layers` (hybrid) against
  `get_linear_attention_pattern` (GPT) so a semantic drift on either side fails.
- `tests/unit_tests/models/test_hybrid_model.py` — `HybridModel` constructor with
  `position_embedding_type='mrope'`.
- `examples/multimodal_dev/tests/test_cp_thd_correctness.py` — ported to the
  hybrid stack spec.

# Checkpoint compatibility

The decoder checkpoint layout changes: one GPT decoder layer is split across two
hybrid layer indices, the final-norm key changes, and hybrid MTP uses a nested
`HybridStack`. An existing GPTModel-format Qwen3.5-VL checkpoint is therefore not
directly loadable. `tools/checkpoint/gpt_hybrid_conversion.py` does not cover
this case either — it explicitly rejects GDN (`G`) and MTP checkpoints. This is
documented in `examples/multimodal_dev/README.md`.

# Scope of risk

The `megatron/core` changes are additive: the mRoPE branches only run when
`position_embedding_type == 'mrope'`, which no existing `HybridModel`
configuration uses, and the FLA import change only alters behavior on builds
that lack `fla.ops.cp`. Everything else is confined to `examples/multimodal_dev`.


QiZhangNV:
While reviewing this migration, I noticed that the configured position embedding type does not match the model that is actually constructed.

`run_qwen35_vl.sh` sets `--position-embedding-type rope`, while `Qwen35VLModel` ignores that configuration and unconditionally constructs the decoder with `position_embedding_type="mrope"`. The runtime model therefore uses MRoPE, but the parsed arguments and checkpoint metadata describe ordinary RoPE. Configuration-driven tooling such as checkpoint converters, exporters, or generic model builders could consequently reconstruct the wrong decoder architecture.

This cannot be fixed by changing the recipe to `--position-embedding-type mrope` alone. The earlier MTP validation in `arguments.py` only accepts `rope` and `none`, so it rejects `mrope`, while the later validation explicitly lists `mrope` as supported. Please consolidate these two validation paths and make the recipe, model construction, and checkpoint metadata consistently describe the decoder as using `mrope`.

megatron/core/models/hybrid/hybrid_model.py
```diff
@@ -329,6 +335,26 @@ def __init__(
                 use_cpu_initialization=self.config.use_cpu_initialization,
                 cp_group=self.pg_collection.cp,
             )
+        elif self.position_embedding_type == 'mrope' and not self.config.multi_latent_attention:
+            self.rotary_pos_emb = MultimodalRotaryEmbedding(
```
hxbai, line 339: Should we pass `cp_group` to mrope like other PEs?

examples/multimodal_dev/models/qwen35_vl/configuration.py
```diff
@@ -302,8 +313,9 @@ def get_qwen35_vl_language_config(
     v = _VARIANT_CONFIGS[variant]
 
     kwargs = dict(
```
hxbai, line 315: I think `is_hybrid_model` also need to be set.

FDecaYed:
With #6083 now in `dev`, I suggest landing this HybridModel migration before #6081. That lets the PP work be rebased onto the final HybridModel interface instead of the GPTModel/MTP-spec path that this PR replaces.

Before the next exact-head CI, could you address the two blocking findings: pass the injected CP group to `MultimodalRotaryEmbedding` ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6315#discussion_r3903095114)), and validate that the effective `mrope_section` matches the Qwen3.5 architecture ([thread](https://github.com/NVIDIA/Megatron-LM/pull/6315#discussion_r3903099095))? Then please rebase onto the current `dev` and ping us for another review/CI run.

megatron/core/models/hybrid/hybrid_model.py
```diff
@@ -329,6 +335,26 @@ def __init__(
...
+                kv_channels=self.config.kv_channels,
+                rotary_percent=rotary_percent,
+                rotary_interleaved=self.config.rotary_interleaved,
+                seq_len_interpolation_factor=seq_len_interpolation_factor,
+                rotary_base=rotary_base,
+                interleaved_mrope=self.config.mrope_interleaved,
+            )
```
FDecaYed, line 346: Please pass `cp_group=self.pg_collection.cp` to `MultimodalRotaryEmbedding`, consistent with the rotary and YaRN constructors immediately above. MRoPE participates in context-parallel position handling, so leaving this unset makes the new HybridModel path depend on a global/default process group instead of the `ProcessGroupCollection` injected into this model. That can select the wrong CP group—or fail entirely—in isolated or non-default process-group configurations, and it works against the ongoing effort to remove direct MPU-global dependencies from Megatron Core library code. Please also add a focused constructor/forward test that supplies a non-default or mocked CP group and verifies that the same group reaches the MRoPE implementation.

examples/multimodal_dev/models/qwen35_vl/factory.py
```diff
@@ -6,19 +6,27 @@
...
+    ``mrope_interleaved`` goes the other way. ``--mrope-interleaved`` also exists
+    as a generated flag, but it is ``store_true`` with a ``False`` default, so a
+    recipe that forgot it would silently build a non-interleaved decoder.
+    Qwen3.5 always interleaves the T/H/W sections, so it is pinned here as an
+    architectural constant, alongside ``ROTARY_PERCENT`` / ``ROTARY_BASE`` in
+    ``model.py``. ``mrope_section`` cannot get the same treatment because
+    ``validate_args`` requires it on the CLI before the model is built.
     """
-    language_config.mrope_section = list(MROPE_SECTION)
     language_config.mrope_interleaved = True
```
FDecaYed, line 30: `mrope_section` is part of the Qwen3.5-VL architecture, not an arbitrary recipe-only tuning value. This factory pins `mrope_interleaved`, but currently accepts any CLI-provided section that passes the generic TransformerConfig checks; a section with the expected total width but the wrong temporal/height/width partition will therefore build an incompatible rotary layout. Please validate the effective `language_config.mrope_section` against the architecture value for this variant (currently `[11, 11, 10]`, or a variant-derived constant) before constructing the model, and raise a clear error on mismatch. It is fine to keep the value CLI-fed, but the model-specific factory should verify it. Please cover both the accepted architecture value and a same-sum-but-wrong partition in tests.

megatron/core/models/hybrid/hybrid_model.py
BestJuly: Good catch. This was a bug and have fixed and is covered by new tests.

examples/multimodal_dev/models/qwen35_vl/configuration.py
BestJuly: Agreed and have set this, and also deleted `linear_attention_freq` to match the logic.

megatron/core/models/hybrid/hybrid_model.py
BestJuly: Done, and has added unit tests, which passed in local GPU tests.

examples/multimodal_dev/models/qwen35_vl/factory.py
BestJuly: Good catch. Now wrong partition will be rejected for Qwen3.5-VL model.

FDecaYed (approved):
Re-reviewed the current head e9927943. The CP-group injection and architecture-specific mrope_section validation are correctly implemented with focused coverage, and the exact-head required CI checks pass. The failing codecov/patch status is advisory rather than a required dev check; its misses are concentrated in optional fused/inference and missing-extension defensive branches, so I do not consider it a merge blocker.

PR #6316, opened 2026-08-06
[Dev] Generalize FSDP persistent communication buffers

lhb8125:
## Summary

This PR generalizes Megatron-FSDP's persistent communication-buffer pools from a hard-coded double buffer to a configurable N-buffer pool.

- Add `fsdp_buffer_count` and keep its default at `2` to preserve the existing double-buffer behavior.
- Propagate the setting through Megatron-Core DDP configuration, the training CLI, and the standalone `fully_shard` APIs.
- Replace FSDP-unit-count-based prefetch checks with allocator-aware capacity checks that account for buffers that are already live and buffers that will be released before the next allocation.
- Allow combined 1F1B overlap to use persistent FSDP buffers when at least three slots are configured.

## Motivation

`fsdp_double_buffer` historically enables two persistent copies of each eligible Megatron-FSDP communication-buffer layout. Two copies are sufficient for conventional current-unit plus next-unit prefetch overlap, but combined 1F1B can require three concurrently live units:

1. a backward/recompute unit,
2. the current genuine-forward unit, and
3. its forward-prefetched successor.

The slots are shared runtime capacity; they are not statically reserved for those three roles.

The previous capacity check counted only the distinct FSDP unit IDs requested by the current all-gather call and assumed a global limit of two. That approximation is insufficient because:

- a requested bucket may already own a slot and therefore require no new allocation;
- other live buckets may occupy slots even when they are absent from the current request;
- lazily released buckets can be recycled before the next allocation and should count as available capacity; and
- weight, FP8-transpose-weight, gradient, and HSDP communication buffers use separate allocators, with capacity partitioned by bucket offset and, for max pools, dtype.

As a result, counting only requested FSDP units can either admit an allocation that does not fit the live pool state or reject a valid allocation whose releasable buffers will be recycled first.

## Configuration and API

`fsdp_double_buffer` remains the backward-compatible enable flag for Megatron-FSDP persistent communication-buffer pools. Despite the legacy name, the pool capacity is controlled by `fsdp_buffer_count`:

| Configuration | Behavior |
| --- | --- |
| `fsdp_double_buffer=False` | Use the existing non-persistent/dynamic allocation path. |
| `fsdp_double_buffer=True`, `fsdp_buffer_count=2` | Preserve conventional double buffering. |
| `fsdp_double_buffer=True`, `fsdp_buffer_count>=3` | Use an N-buffer persistent pool. |

The default remains `2`, so existing configurations retain their current allocation behavior and memory footprint.

Example training CLI configuration:

```bash
--fsdp-double-buffer \
--fsdp-buffer-count 3
```

The same capacity setting is exposed by:

- `megatron.core.distributed.DistributedDataParallelConfig`;
- the standalone Megatron-FSDP `DistributedDataParallelConfig`;
- `fully_shard_model(...)`;
- `fully_shard(...)`; and
- `MegatronFSDP(...)`.

When combined 1F1B overlap (`overlap_moe_expert_parallel_comm`) is used with persistent Megatron-FSDP buffers, `fsdp_buffer_count` must be at least `3`.

### Persistent-pool construction

The fixed `UB_BUFFER_NUM = 2` is replaced with `fsdp_buffer_count` when constructing the persistent allocators for:

- model weights;
- FP8 transpose weights;
- main gradients; and
- HSDP gradient communication.

Increasing the count adds another reusable buffer group to each applicable pool. The exact memory increase depends on the eligible bucket layouts, dtypes, and whether fixed-pool or max-pool allocation is selected.

### Allocator-aware capacity checks

`TemporaryBucketAllocator` now exposes a common `can_allocate(...)` query:

- dynamically sized allocators always return `True` because they do not have a fixed slot limit;
- `FixedPoolAllocator` checks capacity per bucket offset; and
- `MaxPoolAllocator` checks capacity per `(dtype, bucket_offset)` slot class.

Both fixed allocators account for:

- currently idle slots;
- live requested buckets that already own storage and require no new slot; and
- live buckets marked for lazy release, whose slots will be recycled before allocation.

`AllGatherPipeline._persistent_allocators_can_fit(...)` groups requested and releasable buckets by their actual allocator, including the forward-weight versus backward FP8-transpose distinction, and requires every affected pool to fit.

### All-gather and prefetch behavior

The capacity query is used in two places:

- If the buckets required by the current all-gather do not fit, the pipeline raises a descriptive error.
- If an optional prefetch candidate does not fit, the pipeline stops extending the prefetch window instead of overcommitting the pool.

Lazy-release bookkeeping remains separate from allocation: the capacity check predicts which slots will be available, and `async_bucket_gather()` then recycles those buckets immediately before fetching new bucket storage.

### Gradient-reduction limit

The gradient-reduction queue limit now uses `fsdp_buffer_count` instead of the previous hard-coded limit of two FSDP units.

## Backward compatibility

- The default capacity remains two.
- Existing `fsdp_double_buffer` users do not need to change their configurations.
- Existing standalone `fully_shard` callers receive the new argument with a default value, so current call sites remain source compatible.
- The non-persistent allocation path is unchanged when `fsdp_double_buffer` is disabled.
- This PR does not change model parameters, checkpoint formats, or numerical operations; it changes only temporary communication-buffer capacity and lifetime accounting.

### Unit coverage

Added CPU unit coverage for both persistent allocator implementations:

- `test_fixed_pool_capacity_includes_live_and_releasable_buffers`
- `test_max_pool_capacity_includes_live_and_releasable_buffers`

Each test verifies that:

1. a bucket already live in the pool requires no additional slot;
2. a new bucket is rejected when all matching slots are live; and
3. the same new bucket is accepted when one matching live bucket is marked releasable.

### Static validation

- `CHECK_ONLY=true BASE_REF=dev bash tools/autoformat.sh`: Black, isort, pylint, and ruff pass.
- Copyright checks pass for all changed Python files.

### Combined 1F1B / full-iteration CUDA Graph validation

Validated with the 61-layer DeepSeek-V3 combined-1F1B/full-iteration-CUDA-Graph configuration on GB200:

- SLURM job `2593409` completed all `11/11` iterations with the generalized pool and corrected live/releasable accounting.
- Earlier two-slot and live-accounting failures were reproduced in SLURM jobs `2592350` and `2592413`.

### Not run locally

The local toolkit lint environment does not contain PyTorch, so pytest and mypy import-based validation is deferred to CI or a PyTorch-enabled container.

Generated with Codex.


shjwudp (approved):
This PR adds the `fsdp_buffer_count` option and includes boundary checks. LGTM.

lhb8125:
/ok to test [3d1180a](https://github.com/NVIDIA/Megatron-LM/pull/6316/commits/3d1180abd18566f8769fd6f01e7b6d1f29c6d7f2)

Wohox (approved):
LGTM!

lhb8125:
/ok to test [f7b65d4](https://github.com/NVIDIA/Megatron-LM/pull/6316/commits/f7b65d4a2ea145ce13497b4df37572f681c6ac4e)

megatron/training/arguments.py
```diff
@@ -4260,10 +4260,20 @@ def _add_distributed_args(parser):
...
+        "user buffer registration and is enabled automatically when using NCCL user buffers.",
+    )
+    group.add_argument(
+        '--fsdp-buffer-count',
+        type=int,
+        default=2,
+        help="Number of persistent buffers in each Megatron FSDP communication pool. "
+        "The default of two provides conventional double buffering; combined 1F1B "
+        "overlap with forward prefetch requires at least three. A non-default value requires "
+        "--fsdp-double-buffer, --nccl-ub, or --megatron-fsdp-max-pool-double-buffer.",
```
hxbai: Should it be `--use-nccl-ub`?

lhb8125: Fixed in fe6c65925: updated the help text to use the actual CLI flag --use-nccl-ub. Thanks!

lhb8125:
/ok to test 
8b6c523

hxbai (approved):
LGTM

PR #6326, opened 2026-08-06
Add per-head Muon QKV orthogonalization and fix TP>num_query_groups with Muon. Generalize to MLA and GQA

mkhona-nvidia:
## Summary

- add opt-in `--muon-split-qkv-per-head` support so Muon orthogonalizes each physical Q, gate, K, and V head independently
- derive layouts from the owning attention layer for MHA, GQA, gated attention, heterogeneous layers, and MLA
- keep QKV parameters on the split path when TP fragments query-group blocks, including `TP > num_query_groups`
- reconstruct fragmented layouts across TP and GTP-remat, remove alignment padding before Newton-Schulz, and restore each rank's physical shard afterward
- preserve the old TP-aware path only for projection-level splits with complete local query groups and no GTP
- route per-head, GTP, and fragmented layouts through duplicated/non-TP Newton-Schulz; an explicit `muon_tp_mode=distributed` request warns once
- support `muon_tp_mode=auto` by explicitly selecting duplicated execution for reconstructed QKV layouts instead of forwarding `auto` to Emerging-Optimizers
- expose the behavior through Muon and AdaptiveMuon and add focused TP/GTP/MLA coverage

This branch is rebased onto current `main` at `c9e7ea765`.

## Root cause and behavior

The old projection-split path decided from each rank-local QKV row shard. When TP or GTP cut through a query-group block, the parameter could stop being tagged for QKV splitting, so the same logical weight received different Muon update rules under different parallel layouts.

Tagging now uses the owning layer's global QKV layout and accounts for TP, GTP-remat, and alignment padding. If a split cannot be performed locally, the optimizer reconstructs the logical rows before applying projection-level or per-head Newton-Schulz and then restores the TP/GTP shard.

Uniform per-head layouts use Emerging-Optimizers' batched 3D Newton-Schulz path. The `v0.3.0` dependency already pinned on current `main` provides that contract, so this PR no longer changes the dependency pin.

A genuine layer-layout mismatch still falls back to whole-matrix Muon rather than aborting. This primarily protects custom or heterogeneous modules whose parameter metadata does not describe the physical weight.

## MLA + Muon behavior change

MLA Q and KV up-projection weights now carry logical `qkv_layout` metadata. Because `muon_split_qkv=True` by default, existing MLA + Muon runs now orthogonalize those weights by logical projection slices rather than as one whole matrix. This intentionally changes Muon update geometry and training numerics by default.

Use `--muon-no-split-qkv` to retain the previous whole-matrix MLA behavior. There is currently no MLA + Muon functional golden test in the repository; future functional coverage will need golden values for the split-projection behavior.

## PR #7117 integration

This branch cherry-picks all three commits from [PR #7117](https://github.com/NVIDIA/Megatron-LM/pull/7117) in order and preserves Shiqing Fan's authorship:

- gathered QKV row accounting excludes GTP padding and is independent of rank-local shard divisibility
- tagging and execution both resolve the same GTP-remat process group, including partial process-group collections
- fallback diagnostics include the first parameter debug name
- the `tp_mode=auto` commit is now substantive: reconstructed per-head/global paths explicitly request duplicated execution and have a 4/8-GPU parity regression
- imported GTP tests exercise the public `TensorParallelMuon.orthogonalize()` path

## Validation

Post-rebase checks:

- `git diff --check upstream/main...HEAD`
- Python syntax compilation for every changed Python file
- isort 5.13.2 check
- Ruff 0.9.10 check
- verified 14 commits, each with a commit signature and DCO trailer
- verified the PR #7117 `tp_mode=auto` commit is non-empty and contains code, docs, and a distributed regression test

Distributed pytest was not run locally because this macOS workspace has neither PyTorch/CUDA nor the required 4/8-GPU environment. The checkout requires Black 26.3.0, which is not installed in this workspace; the available Black 24 refuses the repository's required-version pin.


megatron/core/optimizer/__init__.py
```diff
@@ -788,16 +792,77 @@ def _get_megatron_emerging_optimizer(
...
+                        f"global_rows={sum(global_split_shapes)}, "
+                        f"local_rows={param.shape[0]}, tp_size={tp_size}, "
+                        f"gtp_remat_size={gtp_size}, "
+                        f"gtp_pad_length={qkv_gtp_pad_length}"
+                    )
+                local_start = tp_rank * logical_tp_local_rows + gtp_rank * param.shape[0]
+                if config.muon_split_qkv_per_head:
+                    if qkv_gtp_pad_length > 0:
+                        param.qkv_split_shapes = qkv_split_shapes
+                        param.qkv_split_heads_are_complete = False
```
philipcmonk, line 874: IIUC the point of this branch is to set `param.qkv_split_heads_are_complete = False` so it forces us to remat along GTP/TP.  `param.qkv_split_shapes = qkv_split_shapes` is somewhat redundant/misleading, since in this branch it'll set it to the global shape, versus in the other branch if a head is not complete `qkv_split_shapes` will still be local in some sense.  It doesn't matter because if `qkv_split_heads_are_complete = False` then we never read this parameter, but it's inconsistent.

At minimum we could add a comment saying that `qkv_split_heads_are_complete` makes it fall back to `qkv_split_shapes_global` so `qkv_split_shapes` is irrelevant.  But probably `qkv_split_shapes` should be set to `None` to early flag any downstream usage.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -230,6 +287,130 @@ def scaled_orthogonalize_fn(
...
+        if self.split_qkv_per_head:
+            if grad.shape[0] != split_width:
+                raise RuntimeError(
+                    f"Muon per-head QKV split shape mismatch: grad_shape={tuple(grad.shape)}, "
+                    f"split_shapes={split_shapes}"
+                )
+            if len(set(split_shapes)) == 1:
+                # A 3D input selects Emerging-Optimizers' batched Newton-Schulz path.
+                head_rows = split_shapes[0]
+                return orthogonalize_fn(grad.view(len(split_shapes), head_rows, -1)).view_as(grad)
```
philipcmonk, line 410: Technically I don't think emerging-optimizers supports batched NS until v0.3.0 which is ahead of uv.lock, and I would expect it to crash before that.  Not sure what else might break if we upgrade that though.

megatron/core/optimizer/__init__.py
mkhona-nvidia: Agreed and fixed.
Padded per-head QKV parameters now set param.qkv_split_shapes = None.
A comment explains that qkv_split_heads_are_complete = False selects qkv_split_shapes_global.
The TP×GTP regression now explicitly tests the None local metadata path.

megatron/core/optimizer/emerging_optimizers.py
mkhona-nvidia: There's another PR that bumps it up, will add a guard though

philipcmonk (approved):

tests/unit_tests/generalized_tensor_parallel/test_tp_gtp.py
```diff
@@ -395,3 +400,103 @@ def test_pre_init_pads_per_tp_slice(self, tp_size, gtp_remat_size):
...
+    )
+
+    logical_rows = sum(split_shapes)
+    logical_tp_rows = logical_rows // tp_size
+    pad_length = 3
+    physical_tp_rows = logical_tp_rows + pad_length
+    gtp_local_rows = physical_tp_rows // gtp_remat_size
+    hidden_size = 4
+
+    global_grad = torch.arange(logical_rows * hidden_size, dtype=torch.float32, device="cuda").view(
```
skyw: Avoid ambiguous naming like "global", hard to interpret what it means.
If it is the convention of the code base, comment what it means.

megatron/core/optimizer/__init__.py
```diff
@@ -779,25 +783,108 @@ def _get_megatron_emerging_optimizer(
...
+                gtp_remat_group = (
+                    pg_collection.expt_gtp_remat
+                    if getattr(param, 'expert_tp', False)
+                    else pg_collection.gtp_remat
+                )
+                if getattr(param, 'is_gtp_weight_remat', False):
+                    gtp_size = get_pg_size(gtp_remat_group)
+                    gtp_rank = get_pg_rank(gtp_remat_group)
                 else:
+                    gtp_size = 1
```
skyw: Shouldn't need this path, get_pg_size and get_pg_rank returns 1 and 0 with None.

megatron/core/transformer/attention.py
```diff
@@ -136,6 +136,49 @@
...
+    same group. Standard fused QKV has one group per query group, while MLA up-projections have
+    one group per attention head.
+    """
+
+    num_groups: int
+    projection_split_shapes: tuple[int, ...]
+    per_head_split_shapes: tuple[int, ...]
+
+    @classmethod
+    def from_standard_attention_config(cls, config: TransformerConfig) -> 'QKVLayout':
```
skyw: Better to name it from_transfomer_config.

megatron/core/transformer/attention.py
```diff
@@ -136,6 +136,49 @@
...
+        projection_split_shapes += [config.kv_channels, config.kv_channels]
+        per_head_split_shapes += [config.kv_channels, config.kv_channels]
+        return cls(
+            num_groups=config.num_query_groups,
+            projection_split_shapes=tuple(projection_split_shapes),
+            per_head_split_shapes=tuple(per_head_split_shapes),
+        )
+
+    @classmethod
+    def from_repeated_splits(cls, num_groups: int, split_shapes: tuple[int, ...]) -> 'QKVLayout':
```
skyw: looks like from_splits? repeated is on output not input, i.e. not from.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -292,36 +556,31 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
             partition_dim = None
 
         if self.split_qkv and self.is_qkv_fn(p):  # type: ignore[misc]
-            grad_shape = grad.shape
+            if self.split_qkv_per_head:
+                return self._orthogonalize_qkv_per_head(p, grad, tp_group)
```
FDecaYed, line 560: I think we can keep the behavior simpler here: preserve the old TP Muon path only for the case it already supported correctly — projection-level QKV splitting where each TP rank has complete local query groups and GTP is not involved.

Per-head splitting and GTP should always use the new path. The same applies when projection-level QKV cannot be split locally. If `muon_tp_mode=distributed` is requested in those cases, please fall back to non-TP/duplicated NS and print a warning instead of silently ignoring the setting.

In short: keep the old behavior only for the old valid case; route all new cases（perhead/gtp/tp>group) through the new implementation. and print warning

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -230,6 +307,130 @@ def scaled_orthogonalize_fn(
...
+        if not split_shapes or any(size <= 0 for size in split_shapes):
+            raise RuntimeError(f"Muon QKV split shapes must be positive: {split_shapes}")
+        split_width = sum(split_shapes)
+        if self.split_qkv_per_head:
+            if grad.shape[0] != split_width:
+                raise RuntimeError(
+                    f"Muon per-head QKV split shape mismatch: grad_shape={tuple(grad.shape)}, "
+                    f"split_shapes={split_shapes}"
+                )
+            if len(set(split_shapes)) == 1 and _supports_batched_newton_schulz():
```
FDecaYed: Now that #6381 is in main, there is one compatibility case to handle: uniform per-head splits use 3D batched NS, but EO’s SYRK path only supports 2D input. `muon_split_qkv_per_head + muon_use_syrk` would therefore fail.

Could we use the batched path only when SYRK is disabled, and fall back to one 2D call per head when it is enabled? It would also be good to add a test for this flag combination.

Longer term, I would prefer MCore to pin and support a single EO version instead of adding more version-specific branches here, but that can be a follow-up.

megatron/core/transformer/multi_latent_attention.py
```diff
@@ -616,6 +616,15 @@ def __init__(
             name=(name + ".linear_kv_up_proj") if name is not None else None,
         )
 
+        q_up_proj = self.linear_q_proj if self.config.q_lora_rank is None else self.linear_q_up_proj
+        q_up_proj.weight.qkv_layout = QKVLayout.from_splits(
+            self.config.num_attention_heads,
+            (self.config.qk_head_dim, self.config.qk_pos_emb_head_dim),
```
buptzyb, line 622: With this layout, enabling `muon_split_qkv_per_head` appears to run separate Newton–Schulz updates for every head’s content-Q (`W^UQ`) and RoPE-Q (`W^QR`) slices. Is per-head Muon for `W^QR` required by a specific model or recipe? If not, do we need mixed-granularity layout metadata so that `W^UQ` can be split per head while `W^QR` retains its intended whole-projection treatment?

The same question applies to the corresponding FusedMLA and AbsorbedMLA layouts below.

mkhona-nvidia: We think per-head muon should apply to each attention head individually. So I would be biased towards splitting every type of head regardless of RoPE (note that latet SOTA hybrid models do not have RoPE) for both GQA and MLA. 

GLM-5 is where it was introduced: https://arxiv.org/html/2602.15763v2

and it has since been adopted by Kimi-K3. We expect it to be default for all models going forward

buptzyb: From the GLM-5 report I only found

> In the original recipe, we apply matrix orthogonalization to the up-projection matrices W^UQ, W^UK, W^UV for multi-head queries, keys, and values. Instead, we split these matrices into smaller matrices for different heads and apply matrix orthogonalization to these independent matrices.

So I'm not sure how W^QR is treated.

megatron/core/optimizer/emerging_optimizers.py
skyw: batched syrk is supported in EO now.

megatron/core/optimizer/emerging_optimizers.py
mkhona-nvidia: https://github.com/NVIDIA/Megatron-LM/pull/6326/changes/3a3d00cd072b6720a795e613a66889bf67c907be

This commit should address your comment

kwangjunahn:
Heads up, [!7117](https://github.com/NVIDIA/Megatron-LM/pull/7117) addresses the GTP portion of the shards layout problem. But it seems to me that the current PR solves the issue more generally

mkhona-nvidia:
> Heads up, [!7117](https://github.com/NVIDIA/Megatron-LM/pull/7117) addresses the GTP portion of the shards layout problem. But it seems to me that the current PR solves the issue more generally

thank you for pointing me to it. I posted about it but seems like it was dropped

mkhona-nvidia:
Rebased this PR's 14-commit stack onto current `main` (`c9e7ea765`) and force-pushed with an explicit lease.

History was preserved:
- PR #7117's three commits remain separate and authored by Shiqing Fan.
- The `tp_mode=auto` commit is no longer an empty placeholder. It now explicitly routes reconstructed per-head/global QKV Newton-Schulz through duplicated mode and adds a 4/8-GPU parity regression.
- All 14 rebased commits contain commit signatures and DCO trailers.

Post-rebase validation passed: Python syntax compilation for all changed Python files, Ruff 0.9.10, isort 5.13.2, and `git diff --check upstream/main...HEAD`.

Distributed pytest was not run locally because this macOS workspace has no PyTorch/CUDA or 4/8-GPU environment. Exact Black 26.3.0 is also not installed locally; the available Black 24 refuses the repository's required-version pin.

GitHub currently reports the PR mergeable.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -507,6 +702,82 @@ def scaled_orthogonalize_fn_with_gtp_remat(self, p, grad, tp_group, partition_di
...
+        if qkv_split_shapes is None:
+            raise RuntimeError("Muon per-head QKV split requested but qkv_split_shapes is not set")
+        if not qkv_split_shapes or any(size <= 0 for size in qkv_split_shapes):
+            raise RuntimeError(
+                f"Muon per-head QKV split shapes must be positive: {qkv_split_shapes}"
+            )
+
+        expected_rows = sum(qkv_split_shapes)
+        gathered_grad, tp_slice, gtp_slice = self._gather_qkv_grad(
+            p, grad, tp_group, expected_rows, gather_gtp=not use_local_layout
```
FDecaYed, line 739: Could you check whether `use_local_layout` is guaranteed to agree across all ranks in the GTP group? It is derived from rank-local head completeness, but controls participation in `all_gather`.

For example, consider an MLA per-head layout `[192, 64, 192, 64]` with TP=1, GTP=8, and no padding. Each rank owns 64 rows: ranks 3 and 7 contain complete RoPE-Q slices and skip the gather, while the other ranks contain partial content-Q slices and enter it. This appears to create mismatched collective participation and a potential hang.

Could you ensure the gather decision is consistent across the group and add a regression test for unequal MLA slice sizes?

asolergi-nv (approved):

megatron/core/optimizer/emerging_optimizers.py
mkhona-nvidia: `use_local_layout` was not guaranteed to agree across GTP ranks, so the unequal MLA layout could cause mismatched all_gather participation and hang

I’ve fixed this by disabling the local-layout optimization whenever dim 0 is sharded across GTP or TP. In those cases, every rank reconstructs the global QKV layout before per-head Newton–Schulz, regardless of whether its local shard happens to contain complete heads.

PR #6333, opened 2026-08-06
Refine MFSDP v2 configuration validation

### LINKED ISSUE #5614: Integrate MFSDP v2 with Megatron Core training

wujingyue:
## Summary

- Refine MFSDP v2 configuration validation: retain restrictions that are genuinely unsupported while allowing v2-native overlap, HSDP, and double-buffer behavior.
- Enable NCCL user buffers through `fully_shard_context(use_symmetric_memory=ddp_config.nccl_ub)`. Reject only the incompatible combination of NCCL user buffers with disabled symmetric registration.
- Accept the default `fsdp_all_gather_in_start_param_sync=True` as a v2 no-op. MFSDP v2 gathers parameters from forward pre-hooks, so `start_param_sync()` does not issue an all-gather.
- Add adapter coverage for construction-scoped symmetric memory.

## Validation

- Black check (locked version)
- `isort --check-only` and `ruff check` on the changed adapter and test files
- `python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/mfsdp_v2/test_mcore_adapter.py` (3 passed)
- `git diff --check`

### DIFF
--- a/megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
+++ b/megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
@@ -564,6 +564,11 @@ class FullyShardedDataParallelV2(_BaseDataParallel):
         placements = Placements(
             dp_axes=[0], parameter=[Flat()], gradient=[Flat()], optimizer=[Flat()]
         )
+        # NCCL symmetric memory requires UB. MFSDP v2 intentionally does not support UB
+        # without symmetric memory: it uses ncclCommRegister rather than the more performant
+        # ncclCommWindowRegister:
+        # https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/bufferreg.html#window-registration
+        use_symm_mem = ddp_config.nccl_ub
         with fully_shard_context(device=device):
             for submodule in reversed(list(module.modules())):
                 if submodule is module:
@@ -576,9 +581,14 @@ class FullyShardedDataParallelV2(_BaseDataParallel):
                         mesh=mesh,
                         placements=placements,
                         mixed_precision_policy=self.mp_policy,
+                        use_symm_mem=use_symm_mem,
                     )
             fully_shard(
-                module, mesh=mesh, placements=placements, mixed_precision_policy=self.mp_policy
+                module,
+                mesh=mesh,
+                placements=placements,
+                mixed_precision_policy=self.mp_policy,
+                use_symm_mem=use_symm_mem,
             )
         super().__init__(config=config, module=module)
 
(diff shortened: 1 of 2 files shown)


megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -642,12 +648,8 @@ def _validate_config(
...
             )
-        if ddp_config.num_distributed_optimizer_instances != 1:
-            raise ValueError("MFSDP v2 does not currently support HSDP.")
         if ddp_config.outer_dp_sharding_strategy != "no_shard":
             raise ValueError("MFSDP v2 does not currently support outer DP sharding.")
-        if ddp_config.overlap_grad_reduce or ddp_config.overlap_param_gather:
-            raise ValueError("MFSDP v2 does not currently support communication overlap modes.")
         if config.gradient_accumulation_fusion:
             raise ValueError("MFSDP v2 does not currently support gradient accumulation fusion.")
         if config.calculate_per_token_loss:
```
wujingyue, line 653: MFSDP v2 unconditionally overlaps reduce-scatter and parameter all-gathers, and always uses double buffering. We intentionally accept and ignore the legacy `overlap_grad_reduce`, `overlap_param_gather`, and `fsdp_double_buffer` flags for compatibility: their defaults are `False`, so rejecting those values would require every existing v2 caller to override them just to preserve current behavior. The tradeoff is that setting them to `False` does not disable v2 overlap or double buffering.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -657,16 +659,15 @@ def _validate_config(
         if config.cuda_graph_impl != "none" or ddp_config.megatron_fsdp_cuda_graph_mode:
             raise ValueError("MFSDP v2 does not currently support CUDA graphs.")
 
-        if ddp_config.fsdp_double_buffer:
-            raise ValueError("MFSDP v2 does not support fsdp_double_buffer.")
         if ddp_config.fsdp_db_use_persist_buf_on_alloc_fail:
-            raise ValueError("MFSDP v2 does not support fsdp_db_use_persist_buf_on_alloc_fail.")
-        if ddp_config.fsdp_all_gather_in_start_param_sync:
```
wujingyue, line 664: This flag doesn't make sense for MFSDP v2 because it doesn't rely on this flag to prefetch the first bucket.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -564,6 +564,7 @@ def __init__(
         placements = Placements(
             dp_axes=[0], parameter=[Flat()], gradient=[Flat()], optimizer=[Flat()]
         )
+        use_symm_mem = ddp_config.nccl_ub
```
skyw: Consider comment why symmetric memory depends on UB use here.

Q: Any reason not using long name "use_symmetric_memory"? the other PR has gone length to rename it that way.

skyw (approved):

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: > Q: Any reason not using long name "use_symmetric_memory"? the other PR has gone length to rename it that way.

Yes. I'll do that after the other PR (#6127) is merged.

wujingyue: Also added a code comment as you requested.

jaredcasper (approved):

ericharper (approved):

shanmugamr1992 (approved):

Issue #6325, opened 2026-08-06
Integrate MFSDP v2 with Megatron Bridge


wujingyue:
## Goal

Enable Megatron Bridge training recipes to use the experimental MFSDP v2 path.

## Scope

Build on the Megatron Core integration from #5614 and add the Bridge-side configuration and recipe plumbing needed to select MFSDP v2.

## Intended outcome

- Expose MFSDP v2 through an experimental Bridge configuration.
- Map the Bridge configuration to the MFSDP v2/MCore training path.
- Preserve existing FSDP behavior when MFSDP v2 is not selected.
- Add a minimal Bridge smoke test or recipe that verifies training can start with MFSDP v2 enabled.

PR #6336, opened 2026-08-07
fix(mfsdp): refresh V2 compute weights after optimizer step

wujingyue:
## Summary

- Refresh MFSDP V2 BF16 compute weights through `FullyShardedOptimizer`'s native post-step copy hook.
- Extract `sync_model_weights_from_main_weights()` so the checkpoint-load path and the `fully_shard_optimizer()` step post-hook refresh compute weights through one implementation instead of two.
- Extend the MFSDP V2 optimizer integration test from three to ten steps so missing compute-weight refreshes are detected.

## Root cause

[#5949](https://github.com/NVIDIA/Megatron-LM/pull/5949) moved main-weight to model-weight synchronization from pre-forward to optimizer post-step. `FullyShardedOptimizer` inherited the post-step extension point but left its implementation as a no-op, so optimizer weights updated while subsequent forwards continued to use stale compute weights.

The existing three-step integration test used a 1% relative tolerance. The resulting loss drift remained below that threshold, so the test passed despite the missing refresh.

## Also fixes a rank-order deadlock in `fully_shard_optimizer()`

The experimental `step_post_hook` collected parameter groups into a `set` and iterated it:

```python
fsdp_parameter_groups: set[FsdpParameterGroup] = set()
...
for parameter_group in fsdp_parameter_groups:
```

`FsdpParameterGroup` is `@dataclass(frozen=True, eq=False)`, so it hashes by `id()` and set iteration order varies per process. `sync_model_weight_from_main_weight()` issues a collective whenever `main_weight` and `model_weight` are distinct buffers, so ranks could issue the per-group collectives in different orders and hang.

Routing the hook through the shared helper fixes this as a consequence of the refactor rather than as a special case: the helper iterates the parameters, whose order is identical on every rank, and uses a set only to skip groups already synced.

The bug is latent on `main` today because it needs both conditions at once — two or more parameter groups **and** `main_weight is not model_weight`. No in-tree test combines them: configurations whose optimizer and parameter placements match (e.g. HSDP) alias the two buffers, so the sync early-returns without a collective.

## Validation

- `uv run python -m torch.distributed.run --nproc_per_node=2 -m pytest -q tests/unit_tests/distributed/mfsdp_v2/test_mcore_adapter.py::TestMcoreAdapter::test_build_train_and_step`
- Full `tests/unit_tests/distributed/mfsdp_v2/test_fully_shard.py` on 4x H100: 32 passed. The one failure, `test_overlaps_communication_and_compute[default]`, reproduces identically on unmodified `main` — it is a launch-bound profiler assertion, unrelated to this change.
- Deadlock repro: an HFSDP configuration (optimizer `[Flat, Flat]`, parameter `[Replicate, Flat]`) over three parameter groups on 4 GPUs hung indefinitely before this change and completes in ~9s after it.
- `ruff check` and `black --check` on edited files

Related to #5949 and #6325.

### DIFF
--- a/megatron/core/optimizer/fully_sharded_optimizer.py
+++ b/megatron/core/optimizer/fully_sharded_optimizer.py
@@ -8,6 +8,9 @@ import torch
 
 from ..config_logger import has_config_logger_enabled, log_config_to_disk
 from ..dist_checkpointing.mapping import ShardedStateDict
+from ..distributed.fsdp.src.megatron_fsdp.experimental.parameter_group import (
+    get_containing_parameter_group,
+)
 from ..transformer.module import MegatronModule
 from .grad_scaler import MegatronGradScaler
 from .optimizer import MixedPrecisionOptimizer
@@ -120,7 +123,16 @@ class FullyShardedOptimizer(MixedPrecisionOptimizer):
         """No-op: MFSDP v2 reduces directly into optimizer-visible sharded grads."""
 
     def _copy_main_params_to_model_params(self) -> None:
-        """No-op: MFSDP v2 currently syncs compute weights in its forward pre-hook."""
+        """Refresh MFSDP V2 compute weights after updating optimizer weights."""
+        # TODO(wujingyue): Reuse experimental.checkpoint.resync_compute_weights after #6024 merges.
+        seen_parameter_groups = set()
+        for parameter in self.get_parameters():
+            if (parameter_group := get_containing_parameter_group(parameter)) is None:
+                continue
+            if parameter_group in seen_parameter_groups:
+                continue
+            seen_parameter_groups.add(parameter_group)
+            parameter_group.sync_model_weight_from_main_weight()
 
     def _copy_model_params_to_main_params(self, state_dict=None) -> None:
         """No-op: model loads already write into MFSDP v2's main weights."""
--- a/megatron/core/optimizer/fully_sharded_optimizer.py
+++ b/megatron/core/optimizer/fully_sharded_optimizer.py
@@ -9,7 +9,7 @@ import torch
 from ..config_logger import has_config_logger_enabled, log_config_to_disk
 from ..dist_checkpointing.mapping import ShardedStateDict
 from ..distributed.fsdp.src.megatron_fsdp.experimental.parameter_group import (
-    get_containing_parameter_group,
+    sync_model_weights_from_main_weights,
 )
 from ..transformer.module import MegatronModule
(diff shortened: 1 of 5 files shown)


megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -120,7 +123,15 @@ def _copy_model_grads_to_main_grads(self) -> None:
         """No-op: MFSDP v2 reduces directly into optimizer-visible sharded grads."""
 
     def _copy_main_params_to_model_params(self) -> None:
-        """No-op: MFSDP v2 currently syncs compute weights in its forward pre-hook."""
+        """Refresh MFSDP V2 compute weights after updating optimizer weights."""
+        # TODO: Reuse experimental.checkpoint.resync_compute_weights after #6024 merges.
```
skyw: Nit: may be add name to the TODO.

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -120,7 +123,15 @@ def _copy_model_grads_to_main_grads(self) -> None:
...
+        parameter_groups = {
+            parameter_group
+            for parameter in self.get_parameters()
+            if (parameter_group := get_containing_parameter_group(parameter)) is not None
+        }
+        for parameter_group in parameter_groups:
```
skyw: Nit: single letter "s" difference is hard to spot and easily miss. May consider better name.

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -120,7 +123,15 @@ def _copy_model_grads_to_main_grads(self) -> None:
         """No-op: MFSDP v2 reduces directly into optimizer-visible sharded grads."""
 
     def _copy_main_params_to_model_params(self) -> None:
-        """No-op: MFSDP v2 currently syncs compute weights in its forward pre-hook."""
+        """Refresh MFSDP V2 compute weights after updating optimizer weights."""
+        # TODO: Reuse experimental.checkpoint.resync_compute_weights after #6024 merges.
+        parameter_groups = {
+            parameter_group
+            for parameter in self.get_parameters()
+            if (parameter_group := get_containing_parameter_group(parameter)) is not None
```
skyw: I like walrus :)

skyw (approved):

jaredcasper (approved):

ericharper (approved):

shanmugamr1992 (approved):

PR #6352, opened 2026-08-07
rl: add durable rollout bank for completed groups

lauradang:
Persist completed RolloutGroups to a write-through, per-group-fsync'd ledger the instant they assemble, so completed work survives a SIGKILL (the SLURM time limit) and is restored at resume instead of regenerated.

- types/: extract the rollout data models (Rollout, TokenRollout, RolloutGroup, Rollouts, GroupedRollouts, AgentBaseModel, and the EnvId/GroupsPerEnv/GroupQueuesPerEnv aliases) out of agent/api.py into a megatron.rl.types package; api.py re-exports them so existing imports keep working. RolloutGroup gains a uid; TokenRollout gains advantage_override; Rollout.reward widens to float | None.
- rollout_bank.py: the durable bank (append/restore/mark_consumed/ checkpoint), offset-indexed binary sidecars for token/logprob/mask arrays, per-record checksums, torn-write recovery, and compaction.
- api.py: the pipeline writes each completed group through the bank on assembly (injected, never a global).
- rl_utils.py: bank singleton + restore/injection in get_environment_rollouts, weight-balanced per env across steps.
- weighted_multi_task.py: env_group_targets + num_groups_per_env override so restored groups keep the per-env weight split.
- checkpointing.py: compact at the checkpoint boundary (sync path and async durability callback) so compacted-through T tracks the model.
- arguments.py: --rl-durable-rollout-bank / --rl-rollout-bank-dir / --rl-rollout-bank-max-bytes. No-op when unset.

Ported from jalbericiola/Megatron-LM#10. That PR was authored on a fork base whose in-flight tracker and speculative-rollout scaffolding are not on NVIDIA main, so the bank was reapplied against main's episode-based pipeline rather than cherry-picked.

Tests: unit + pipeline coverage for the bank round-trip, durability, compaction, restore balancing, and the type re-export (tests/unit_tests/ rl/test_rollout_bank.py, test_grouped_rollouts.py). 101 pass locally.

# What does this PR do?

## Issue tracking

Linked issue: 

### Pre-checks


janEbert:
Hey, your first commit (8c9a31051d222fc45187576ab82d7dc3bce6895b) needs to be signed with the DCO still.

megatron/rl/rl_utils.py
```diff
@@ -685,6 +927,12 @@ def get_environment_rollouts(
                                 assert False, "Unexpected group left in generator."
                             except StopAsyncIteration:
                                 break
+                    # Record consumption for every group handed to the trainer. On a
+                    # rollback (restart at T < this step) the marker > T rule restores
+                    # these; once the checkpoint advances past this step they are pruned.
+                    if bank is not None:
+                        for group in rollouts:
+                            bank.mark_consumed(group.uid, args.curr_iteration)
```
tdene: This repeatedly opens and flushes the same file in a loop. This is going to be very slow compared to just opening the file once at the head of the loop and flushing + closing at the end.

megatron/rl/rollout_bank.py
```diff
@@ -0,0 +1,755 @@
...
+    def _maybe_warn_over_cap(self) -> None:
+        if self.max_bytes <= 0 or self._bytes_written <= self.max_bytes:
+            return
+        if not self._warned_over_cap:
+            logger.warning(
+                "RolloutBank exceeded --rl-rollout-bank-max-bytes (%d > %d); will compact at the "
+                "next checkpoint. Consider a shorter checkpoint interval or lower --rl-generation-lag.",
+                self._bytes_written, self.max_bytes,
+            )
+            self._warned_over_cap = True
```
tdene, line 978: This never gets reset after compaction.

megatron/rl/rollout_bank.py
```diff
@@ -0,0 +1,755 @@
...
+            for group in groups:
+                self.append(group, uid=group.uid)  # reuses the write-through encoder + fsync
+        finally:
+            self._close_handles()
+            (self._seg_dir, self._collection_iter, self._seq,
+             self._tok_off, self._lp_off, self._mask_off,
+             self._ledger_f, self._tok_f, self._lp_f, self._mask_f) = saved
+
+    def _maybe_warn_over_cap(self) -> None:
+        if self.max_bytes <= 0 or self._bytes_written <= self.max_bytes:
```
tdene, line 969: `_bytes_written` is also never changed by compaction, even though compaction will change its measure.

megatron/rl/rollout_bank.py
```diff
@@ -0,0 +1,755 @@
...
+_TOKEN_DTYPE = np.int32
+_LOGPROB_DTYPE = np.float16
+_MASK_DTYPE = np.uint8
+
+# Bump whenever the manifest/ledger schema or any sidecar dtype/layout changes.
+# Readers intentionally fail closed; cross-version migration must be explicit.
+_FORMAT_VERSION = 1
+_MANIFEST = "MANIFEST.json"
+_LEDGER = "ledger.log"
+_CONSUMED = "consumed.log"
```
tdene, line 81: Should this also be compacted so that it does not grow without bound?

megatron/rl/agent/weighted_multi_task.py
```diff
@@ -189,9 +190,53 @@ async def get_reward_rollouts(self, request: RolloutRequest) -> list[Rollout]:
...
+    def _env_ids(self) -> list[str]:
+        """Per-agent env_ids (aligned with ``self.agents``)."""
+        env_ids = []
+        for i, (a, config) in enumerate(zip(self.agents, self.agent_configs)):
+            env_id = getattr(a, "env_id", None)
+            if not env_id and not config.evaluation_only:
+                raise ValueError(
+                    f"Active agent {i} ({type(a).__name__}) has no env_id; it is "
+                    f"required to weight-balance restored rollout-bank groups by env. "
+                    f"Set env_id for every non-evaluation agent in the environment config."
```
tdene: Do single-environment setups currently set `env_id`? If so, they would fail this check, when they shouldn't have to because single-env routing is trivial.

megatron/rl/agent/weighted_multi_task.py
```diff
@@ -189,9 +190,53 @@ async def get_reward_rollouts(self, request: RolloutRequest) -> list[Rollout]:
...
+
     async def get_grouped_rollouts(self, request: GroupedRolloutRequest):
         """Distribute grouped rollouts across sub-agents according to weights."""
-        agent_groups = self._distribute_counts(request.num_groups)
+        override: GroupsPerEnv | None = request.num_groups_per_env
+        if override is not None:
+            # Explicit per-env counts (e.g. the residual after injecting restored
+            # rollout-bank groups). Route each env's count to its agent(s).
+            env_ids = self._env_ids()
+            unknown = set[Any](override) - set(env_ids)
```
tdene: I'm not familiar with this `set [Any]` syntax, but from what I can see it's intended for type-hinting not in-line set creation like what's being used here?

megatron/rl/rl_utils.py
```diff
@@ -587,14 +661,119 @@ def get_rollout_generator(args, inference_interface, n_prompts, samples_per_grou
...
+
+    The plain drain: ``anext`` the generator a fixed number of times, with no
+    per-env routing. Correct whenever the generator already yields the right
+    per-env mix -- a one-shot generator built with ``num_groups_per_env``, or a
+    balanced streaming generator in steady state (no restore residual to satisfy).
+    """
+    return [loop.run_until_complete(anext(generator)) for _ in range(n_fresh)]
+
+
+def _pull_fresh_balanced(
```
tdene: I see no test for a streaming restore.

megatron/rl/types/rollout.py
```diff
@@ -0,0 +1,77 @@
...
+    generation_mask: list[list[bool]] | None = None
+    logprobs: list[list[float]] | None = None
+    env_id: str = ''
+    problem_id: str | None = None
+    policy_epoch: list[list[tuple[int, int]]]
+    kv_cache_epoch: list[list[tuple[int, int]]]
+    num_evictions: list[int]
+    # When set, replaces this rollout's group-normalized advantage with a fixed
+    # value after group normalization in calculate_grpo_advantages.
+    advantage_override: float | None = None
```
tdene: This looks like a test-only argument? If so, it does not belong in the main API file; the test can override it. It also doesn't seem to do anything as-is.

megatron/rl/rollout_bank.py
```diff
@@ -0,0 +1,755 @@
...
+        dumped = group.model_dump()
+        dumped.pop("uid", None)  # uid lives at the top level of the record
+        members = dumped.get("rollouts", [])
+        member_type = type(group.rollouts[0]).__name__ if group.rollouts else "Rollout"
+
+        # if the member type is TokenRollout and all the rollouts are token-typed, then set token_typed to True
+        token_typed = member_type == "TokenRollout" and all(
+            "trajectory" in m and m["trajectory"] and isinstance(m["trajectory"][0], list)
+            for m in members
+        )
```
tdene: You can do `isinstance(group.rollouts[0], TokenRollout)` directly.

megatron/rl/rl_utils.py
```diff
@@ -671,9 +904,18 @@ def get_environment_rollouts(
...
+                    # Otherwise pull n_fresh groups in completion order.
+                    if args.rl_partial_rollouts and residual_per_env is not None:
+                        fresh = _pull_fresh_balanced(
+                            loop, rollout_generator, residual_per_env,
+                            runtime_state.fresh_overflow,
+                        )
+                    else:
+                        fresh = _pull_fresh_inorder(loop, rollout_generator, n_fresh)
+                    # Restored groups first, then freshly generated ones.
+                    rollouts = list(inject) + fresh
```
tdene: There's a problem with this logic.

Let's say you're in a 2-env situation. You want 50% from env A and 50% from env B. You bank rollouts, and you end up with 100 env A rollouts and 2 env B rollouts.

This logic ensures that you consume a balanced amount, but on the production side you will continue to produce env A despite having way too many rollouts from it.

I don't think this should be fixed in this PR. I think this is a pervasive prior issue that can happen even without banking. But I think it should be documented.

tdene: Actually, I'm having second thoughts.

This logic will reliably result in the rollout bank never getting drained. The generator continues to generate rollouts in exactly the same proportion and amount as this logic drains.

I am pretty sure that any overflow is permanent. That is a problem that should be fixed in this PR.

tdene: I also don't think this env-balancing logic should exist on the consumer side.

It already exists on the producer side. The rollout bank is a producer as well. It should plug into `WeightedMultiTask`.

That lets you cut out `_pull_fresh_balanced`, `fresh_overflow`, `_plan_restore_injection`, `_empty_rollout_generator`, `_env_targets`, and `env_group_targets`, and `num_groups_per_env`.

Something like
```
correct_agent_groups = list(agent_groups)
for i, (eid, total) in enumerate(zip(env_ids, agent_groups)):
    cached = min(total, len(restored.get(eid, [])))
    correct_agent_groups[i] = total - cached
    for _ in range(cached):
        yield restored[eid].popleft()
```

megatron/rl/rollout_bank.py
```diff
@@ -0,0 +1,755 @@
...
+
+    @property
+    def _consumed_path(self) -> str:
+        return os.path.join(self.bank_dir, _CONSUMED)
+
+    def _read_manifest(self) -> Manifest:
+        try:
+            with open(self._manifest_path) as f:
+                manifest = json.load(f)
+        except (FileNotFoundError, json.JSONDecodeError):
```
tdene: `FileNotFoundError` and `json.JSONDecodeError` should not be treated the same. If the file is not found, the ledger doesn't exist. If we have a `json.JSONDecodeError`, the user should be told that there's a malformatted ledger and an error should be raised so that the user can try to manually recover it.

megatron/rl/rollout_bank.py
```diff
@@ -0,0 +1,755 @@
...
+            return None
+
+        group_dict = record["group"]
+        uid = record["uid"]
+        if record["kind"] == "inline":
+            cls = TokenRollout if record["member_type"] == "TokenRollout" else Rollout
+            members = [cls.model_validate(m) for m in group_dict["rollouts"]]
+            group = RolloutGroup(
+                rollouts=members,
+                batch_id=group_dict.get("batch_id", 0),
```
tdene, line 768: This `batch_id` will be stale and will collide.

Each new run starts counting from 0, even if it's a continuation run.

I think the correct fix is to fix this behavior, and make runs start counting from `current_iteration` instead.

This will affect B/B.

megatron/rl/rollout_bank.py
```diff
@@ -0,0 +1,755 @@
...
+            }
+        )
+        for seg in old_segments:
+            if seg != new_seg:
+                _rmtree(os.path.join(self.bank_dir, seg))
+
+        # Reopen the (now compacted) active segment for continued appends.
+        self._collection_iter = None
+        self._seg_dir = None
+        self.set_collection(iteration)
```
tdene: I'm pretty sure this will be off-by-1 for sync saves, and may be very off for async saves.

But also, I don't think these 3 lines do anything? Right after this happens, `get_environment_rollouts` runs `bank.set_collection(args.curr_iteration)`.

tdene (commented):
Left a couple of important comments.

Another one in the comments here: you are also recovering rollouts that were trained on since the last saved checkpoint.

The generator should not be allowed to run until we catch up. If we saved on ckpt 20, and crashed on ckpt 38, then ckpts 21 -> 37 should train without the generator ever engaging.

Look at #4125 as well. Perhaps this PR should be rebased on top of 4125 so that it can easily load data into the iterator without turning on the generator.

megatron/rl/agent/weighted_multi_task.py
lauradang: `set[Any](override)` is valid on the project's supported Python versions: `set[Any]` is a callable generic alias and behaves like `set(override)`. That said, the type argument is unnecessary and less readable, so I’ll simplify it to `set(override)`. Existing unknown-environment coverage already exercises the behavior.

megatron/rl/rollout_bank.py
lauradang: Yes—good catch. I’ve changed this so `consumed.log` is compacted together with the rollout data at checkpoint boundaries.

Previously there was one global marker log:

```text
<bank_dir>/
├── MANIFEST.json
├── consumed.log
└── gen-<iter>/
    ├── ledger.log
    ├── tokens.bin
    ├── logprobs.bin
    └── masks.bin
```

Besides growing without bound, that structure had a rollback-lifetime bug:

```text
checkpoint 20
X consumed at 25             -> consumed.log contains X:25
crash and recover from 20    -> 25 > 20, so X is correctly restored
new execution reaches 26
without consuming X          -> stale 25 <= 26, so X is incorrectly discarded
```

The problem is that the marker belongs to iteration 25 on the abandoned execution, but the global file makes it look applicable to iteration 26 on the restarted execution.

The bank now stores the ledger and its marker log in a manifest-selected generation:

```text
<bank_dir>/
├── MANIFEST.json             # selects the active generation and timeline
└── generations/
    └── generation-<uuid>/
        ├── consumed.log
        └── gen-<iter>/
            ├── ledger.log
            ├── tokens.bin
            ├── logprobs.bin
            └── masks.bin
```

At an ordinary checkpoint `T`, the bank writes a new generation containing:

- only groups not yet represented in checkpoint `T`; and
- one deduplicated marker per surviving group only when `marker > T`.

Markers at or before `T`, duplicates, and markers for groups no longer in the ledger are dropped. This bounds the active log to work newer than the checkpoint.

Recovery is intentionally different. Returning to the example, recovery from checkpoint 20 first uses `X:25` to determine that X must survive, then publishes:

```text
generation-<new-timeline>/
├── consumed.log              # empty
└── gen-000020/
    ├── ledger.log            # contains X
    └── ...
```

X is copied because checkpoint 20 did not train on it, but the old `X:25` marker is not copied because that consumption belonged to the abandoned timeline. A later checkpoint 26 therefore retains X unless the restarted execution actually consumes it.

Publication is crash-safe: the complete generation is written and `fsync`ed first, its directory is renamed into place, `MANIFEST.json` is atomically replaced to select it, and the old generation is reclaimed afterward. A crash therefore leaves either the complete old generation or the complete new one active.

megatron/rl/agent/weighted_multi_task.py
lauradang: Good catch. The in-tree single-environment agents currently define `env_id`, but single-env configurations should not be required to do so because routing is unambiguous. I updated this to require `env_id` only when multiple positive-weight agents are active. A single active environment now uses the existing empty-string `Rollout.env_id` default. I also added regression coverage for both the single-env compatibility case and the multi-env validation case.

megatron/rl/agent/weighted_multi_task.py
tdene: Okay! This was just the first time I had ever seen this syntax and it threw me off.

megatron/rl/rollout_bank.py
tdene: Sounds great!

I see `_restore_state` hasn't been adjusted to work with this new format yet, but you're in the middle of writing code right now.

I do want to point out something real though. I'm pretty sure you have an off-by-one error here, the marker is written using `curr_iteration` right after training, so groups consumed on iteration X are actually marked with an X+1 in the ledger.

Make sure your tests cover the case where a group is consumed on the checkpointed iteration itself; in fact you should probably parameterize your tests to have them check situations where groups get consumed on `CKPT-1`, `CKPT`, and `CKPT+1`, because there's so many moving parts in the training loop that there could be off-by-one errors in every direction.

megatron/rl/rollout_bank.py
lauradang: Yes, you are right, it can be very off for async saves especially. The problem is that I muddled the checkpoint method responsibility.

To solve that, I removed the reopen from `checkpoint()` entirely. Compaction now leaves the writer closed, and the next rollout collection explicitly selects the current iteration. I also updated direct unit-test callers to follow that lifecycle.

megatron/rl/rl_utils.py
lauradang: Good catch. I refactored restored groups to flow through the weighted producers and added `test_streaming_restored_backlog_drains_before_fresh_per_env`. It exercises the real pipeline with `streaming=True`, verifies the backlog drains across batches while preserving the per-environment mix, and confirms fresh generation resumes afterward.

megatron/rl/types/rollout.py
lauradang: Agreed. `advantage_override` was unused test-only API surface. I removed it from `TokenRollout` and removed the test that declared it. There are no remaining references.

megatron/rl/rl_utils.py
lauradang: You are right—the consumer-side approach made the overflow permanent. I have changed the implementation so restored groups participate on the producer side.

After rebasing onto the unified `RolloutPipeline`, `bank.recover()` installs per-environment queues through `WeightedMultiTask.set_restored_groups()`. The normal `rollout_allocations()` still creates the weighted environment slots. For each slot, `RolloutPipeline.stage_prepare()` calls `take_restored_group(env_id)` before `prepare_group_rollout()`. If a restored group is available, it fills that slot directly and fresh generation for that environment is skipped.

In your 100-A/2-B, 50/50 example, the first batch consumes two restored groups from each environment. Subsequent batches consume two restored A groups and generate two fresh B groups. No fresh A groups are produced until the restored A queue is empty, so the backlog decreases rather than being replenished.

I also removed the recovery-specific consumer machinery: `_pull_fresh_balanced`, `fresh_overflow`, `_plan_restore_injection`, `_empty_rollout_generator`, `_env_targets`, `env_group_targets`, and `num_groups_per_env`.

There are regression tests for both non-streaming substitution and streaming backlog draining; they verify that the A producer is not called while restored A groups remain and resumes only after that queue is empty. Thanks for catching this.

megatron/rl/rl_utils.py
```diff
@@ -770,6 +836,13 @@ def get_environment_rollouts(
                                 assert False, "Unexpected group left in generator."
                             except StopAsyncIteration:
                                 break
+                    # Record consumption for every group handed to the trainer. On a
+                    # rollback (restart before this update) the marker > T rule restores
+                    # these; once a checkpoint includes the update they are pruned.
+                    if bank is not None:
+                        bank.mark_consumed_many(
+                            (group.uid for group in rollouts), args.curr_iteration + 1
```
tdene, line 881: There's one problem with this `args.curr_iteration + 1` rule.

Inside GRPO, the same data can be reused for multiple training iterations. Specifically, one batch of data feeds `grpo_iterations * (grpo_prompts_per_step * grpo_group_size // global_batch_size)` number of training iterations.

So I think the `+1` in here needs to change to `+(grpo_iterations * (grpo_prompts_per_step * grpo_group_size // global_batch_size))`.

tdene (approved):

megatron/rl/rl_utils.py
tdene: Yeah, I'm quite confident that you need a `+(grpo_iterations * (grpo_prompts_per_step * grpo_group_size // global_batch_size))` here, and the problem is solved.

But I would factor that equation out into a helper method, because around line 2300 we already have:
```
    global_batches_per_collection = (grpo_prompts_per_step * grpo_group_size) // global_batch_size
    if (
        buffered_rollouts is None or
    ):
```

So a helper method can be called in both places.

lauradang: I don't think this is needed given the current assertions here: https://github.com/NVIDIA/Megatron-LM/blob/d42df820852e54e6d748463cdf914f8acf641387/megatron/training/arguments.py#L720-L737. This code asserts that our save intervals must be aligned with the complete collection window. e.g. in the following example, checkpoint intervals must be intervals of 8 (number of optimizer passes per batch essentially)

```
Optimizer step 0: samples   0–127  ┐
Optimizer step 1: samples 128–255  │ first GRPO pass
Optimizer step 2: samples 256–383  │
Optimizer step 3: samples 384–511  ┘
 
(the assertions ensure we can't save a checkpoint here)

Optimizer step 4: samples   0–127  ┐
Optimizer step 5: samples 128–255  │ second GRPO pass
Optimizer step 6: samples 256–383  │
Optimizer step 7: samples 384–511  ┘

Durable checkpoint 8 is written
Optimizer step 8: generate new rollouts
```

jaredcasper (approved):

tests/unit_tests/rl/test_rollout_bank.py
```diff
@@ -0,0 +1,302 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+"""Parameterized coverage for durable rollout-bank persistence and replay."""
```
asolergi-nv, line 3: asyncio import missing?

megatron/rl/types/rollout.py
```diff
@@ -0,0 +1,71 @@
...
+    rollout_status: str = 'ok'
+    failure_reason: str | None = None
+
+
+class TokenRollout(AgentBaseModel):
+    """Tokenized representation of a language-based Rollout."""
+
+    trajectory: list[list[int]]
+    reward: list[float] | float
+    generation_mask: list[list[bool]] | None = None
```
asolergi-nv, line 36: Is it me or are we missing `generation_cap`?

lauradang: yes, thanks for catching that!

asolergi-nv (approved):

chtruong814:
The unit tests in the merge queue failed due to some flaky behavior.  Pulling it out of the queue to retry.  Will open separate PR To resolve the specific test.

Issue #4933, opened 2026-05-22
🐛 CI failure: MultiTokenPredictionLayer._checkpointed_forward() got unexpected kwarg 'padding_mask'


ko3n1g:
**Describe the bug**

CI test `tests/unit_tests/transformer/test_multi_token_prediction.py::TestMultiTokenPrediction` is failing across many parametrizations with a `TypeError` on the recompute path:

```
megatron/core/transformer/multi_token_prediction.py:1301: TypeError
E   TypeError: MultiTokenPredictionLayer._checkpointed_forward() got an unexpected keyword argument 'padding_mask'
```

Failing nodes (all in the same job):

- `TestMultiTokenPrediction::test_forward_backward[{1,2,4}-{1,2,4}-True]` (9 combinations)
- `TestMultiTokenPrediction::test_fp8_support[True]`
- `TestMultiTokenPrediction::test_packed_sequences_with_full_recompute`

Tag @NVIDIA/mcore-oncall to get oncall's attention to this issue.

**Root cause (likely)**

The call site at `megatron/core/transformer/multi_token_prediction.py:1301` passes `padding_mask=padding_mask` into `self._checkpointed_forward(...)`, but the method definition at `megatron/core/transformer/multi_token_prediction.py:1093` does not declare a `padding_mask` parameter.

`git blame` points to two recent landings that don't compose:

- `2d1fa8d372` ([#2645](https://github.com/NVIDIA/Megatron-LM/pull/2645), 2026-05-14, @Connor-XY) added `padding_mask=padding_mask` to the call site.
- `2b77d32b1e` ([#4593](https://github.com/NVIDIA/Megatron-LM/pull/4593), 2026-05-21, @BestJuly) refactored `_checkpointed_forward` without including `padding_mask` in the new signature.

The recompute branch (`config.recompute_granularity == 'full' and self.training`) crashes on every call.

**Failing run**

| Field | Value |
|-------|-------|
| PR    | [#4931: test: enable NVTE_CUTEDSL_FUSED_GROUPED_MLP via pytest fixture](https://github.com/NVIDIA/Megatron-LM/pull/4931) (surfaced here; the PR itself does not touch MTP) |
| Run   | [26290097610](https://github.com/NVIDIA/Megatron-LM/actions/runs/26290097610) |
[...]

PR #6367, opened 2026-08-08
fix(mtp): thread padding_mask through _checkpointed_forward (backport to core_r0.18.0)

gaohongkui:
## Describe your changes

Backport of the `padding_mask` plumbing in `MultiTokenPredictionLayer._checkpointed_forward` to
`core_r0.18.0`, so that MTP works with `recompute_granularity='full'` on a released
megatron-core.

### The problem

On this branch, `MultiTokenPredictionLayer.forward` calls

```python
self._checkpointed_forward(..., padding_mask=padding_mask, ...)   # line 1301
```

but `_checkpointed_forward` (line 1093) does not declare `padding_mask`. Every MTP run with
`recompute_granularity == 'full'` therefore aborts at the first forward:

```
TypeError: MultiTokenPredictionLayer._checkpointed_forward() got an unexpected
           keyword argument 'padding_mask'
```

This is #4933 — two landings that do not compose (#2645 added the call-site kwarg, #4593
refactored the method without it). That issue is still open, and reports it as a CI failure.

**It is not only a CI failure: it is in a release.** `core_v0.18.2` was tagged 2026-07-20 from
this branch — two months after #4933 was filed — and is still the newest release tag. So MTP
with full activation recomputation is unusable on every released megatron-core.

### Why not cherry-pick

`main` gained the parameter in ffd66a3e6bfe ("Roll input IDs for MTP labels", #3457), where it
is incidental to an unrelated feature. Cherry-picking that into a release branch would drag the
feature along, so this is the minimal equivalent: thread `padding_mask` through exactly the six
places `main` threads it, and nowhere else.

* `_checkpointed_forward` signature, positioned after `attention_mask` as on `main`
* `custom_forward`'s parameter list
* the `_proj_and_transformer_layer` call inside `custom_forward`
* both checkpoint entry points (`te_checkpoint`, `tensor_parallel.checkpoint`), **positionally** —
  `padding_mask` is a tensor / `None`, so per this method's own docstring it belongs with the
  positional args rather than the closure
* the `recompute_method == 'block'` fallback path

`_proj_and_transformer_layer` on this branch **already accepts `padding_mask`**, so nothing else
needs to change: the parameter was simply never plumbed through the recompute path.

6 insertions, 0 deletions.

## Testing

* `padding_mask` now appears 8 times inside `_checkpointed_forward`, identical to `main`.
* The file compiles; the change is pure parameter plumbing with no behavioural branch.
* The failure mode it removes was reproduced on 4x8 H20 with a 40-layer MoE
  (`mtp_num_layers=1`, `recompute_granularity='full'`, TP2/PP2/CP4/EP8, 64K context,
  megatron-core 0.18.2 + torch 2.11): training aborts at the first forward with the `TypeError`
  above. With the parameter accepted, MTP trains and MTP losses are reported normally.

I could not exercise this exact patched file end-to-end in that cluster (site-packages there is
root-owned and the container is not privileged, so the installed module cannot be replaced), so
the end-to-end evidence above comes from making `_checkpointed_forward` accept the kwarg at
runtime rather than from this file. The file-level change is byte-equivalent in effect to
`main`'s.

## Checklist

- [x] Read the [contributing guidelines](https://github.com/NVIDIA/Megatron-LM/blob/main/CONTRIBUTING.md)
- [x] Self-reviewed the diff
- [x] Existing issue referenced (#4933)

AI assistance (Claude) was used to locate the root cause and prepare this patch; the diff is
small and was reviewed line by line against `main`.


gaohongkui:
For reviewers: this is a straight backport of the `padding_mask` threading from **`ffd66a3e6` ("Roll input IDs for MTP labels", #3457, 2026-06-03)** onto `core_r0.18.0`. I bisected the file's history on `main` to confirm the attribution — `4c6360260` (2026-05-22) still lacks `padding_mask` in the `_checkpointed_forward` signature, `ffd66a3e6` is the first commit that has it.

That commit threads the parameter through the same five places this PR does (signature, `custom_forward` positional and kwarg, and both `checkpoint_handler()` entry points), and it also added `test_packed_sequences_with_full_recompute` — i.e. upstream already treats packed sequences + `recompute_granularity=full` as a supported combination that must not crash.

Why the backport matters: no released megatron-core carries the fix. `core_v0.18.0` (2026-06-22) and `core_v0.18.2` (2026-07-20, the newest release) both still declare `_checkpointed_forward` without `padding_mask` while `MultiTokenPredictionLayer.forward` passes it, so any `recompute_granularity=full` + MTP run on a released 0.18.x raises `TypeError`. Tracked as #4933.

This PR is still waiting on `copy-pr-bot` validation before any workflow can run, so if a vetter could `/ok to test`, that would be much appreciated.

PR #6373, opened 2026-08-09
Fix hidden checkpoint parameter handling in local CUDA graphs

JF-D:

# What does this PR do?

Fixes local CUDA-graph training when `CheckpointWithoutOutput` recomputation uses parameters from nested modules that are not visible in the outer autograd graph.

## Problem                                                                                                                                               

  `CheckpointWithoutOutput` runs its recompute function during backward. Parameters
  used only by that function do not have autograd edges from the outer forward
  outputs.

  CUDA-graph capture previously discovered `params_to_backprop` only by traversing
  the autograd graph from the captured forward outputs. Consequently, a parameter
  used exclusively by checkpoint recomputation could be omitted from the captured
  backward inputs.
 
  This caused two related problems:

  1. The nested recompute backward accumulated the parameter gradient into
     `param.grad`, but `torch.autograd.grad` returned `None` for the corresponding
     outer parameter input. The gradient was therefore not transferred to
     `param.main_grad`.

  2. Because the parameter was not passed as a synthetic input to the CUDA-graph
     replay node, replay had no autograd edge to its `AccumulateGrad` node. The
     normal DDP hook registered on that node did not fire, leaving the DDP bucket
     with an incomplete ready count.

  This can result in missing gradients, incorrect gradient accumulation, or DDP
  waiting for a bucket that never becomes ready.

  Example:
  ```
  def forward(self, x):
      checkpoint = CheckpointWithoutOutput()


      hidden = checkpoint.checkpoint(self.checkpointed, x)
      output = self.projection(hidden)

      checkpoint.discard_output_and_register_recompute(output)
      return output
  ```
  `self.projection.weight` is used by the outer forward autograd graph, so it can be 
    discovered by traversing backward from output. In contrast, `self.checkpointed.weight`
    is closed over by the checkpoint recompute function. Its computation is deferred until
    the nested recompute backward and therefore has no AccumulateGrad edge in the 
    outer graph. Traversing only the captured forward outputs consequently discovers
   `projection.weight` but misses `checkpointed.weight`, even though backward eventually
    produces a gradient for it.

  ## Solution

  ### Automatically discover hidden checkpoint dependencies

  During CUDA-graph warmup, `CheckpointWithoutOutput` bypasses checkpointing and
  executes its recompute function with autograd enabled. Parameters used by nested
  recompute modules are therefore visible through their `AccumulateGrad` nodes.

  This change:                                                                                                                                             

  - Traverses the warmup autograd graph to discover these parameters.
  - Filters the result to parameters recursively owned by the captured base module.
  - Merges the warmup-discovered parameters with those reachable from the actual
    captured forward outputs.
  - Preserves `base_module.parameters()` ordering so capture and replay use a stable
    positional input layout.
  - Runs at least one internal warmup for training graphs, even when
    `cuda_graph_warmup_steps=0`, because one discovery pass is required for correct
    checkpoint handling.

  This discovery is automatic and does not require checkpoint call sites or
  individual modules to declare additional dependencies.

  ### Capture gradients produced by nested recomputation

  For a parameter hidden behind `CheckpointWithoutOutput`,
  `torch.autograd.grad(..., allow_unused=True)` preserves its requested output slot
  but returns `None` for that slot. Its nested backward instead writes the real
  gradient to `param.grad`.

  Backward capture now transfers that contribution into `param.main_grad`.

  The returned `wgrad` and `param.grad` are handled independently. This is
  important when the same parameter is used both outside and inside checkpoint
  recomputation:

  - `wgrad` represents uses visible to the outer autograd graph.
  - `param.grad` represents uses performed by the nested recompute backward.

  Both are legitimate contributions and must be accumulated exactly once.

  ### Preserve normal DDP bucket readiness during replay

  All discovered parameters are passed as synthetic inputs to
  `_CudagraphReplayNode.apply`. For ordinary DDP parameters, this creates replay
  autograd edges to the existing `AccumulateGrad` nodes where DDP registered its
  hooks.

  For checkpoint parameters whose gradients were already transferred from
  `param.grad` to `param.main_grad`, replay sets
  `grad_added_to_main_grad=True` before the normal DDP hook runs. The hook then:

  - Skips adding the same gradient to `main_grad` a second time.
  - Clears `param.grad`.
  - Still calls `register_grad_ready()`, allowing the DDP bucket to be dispatched.

  GTP parameters retain their existing explicit hook path in
  `_CudagraphReplayNode.backward`.

### Pre-checks

### DIFF
--- a/megatron/core/transformer/cuda_graphs.py
+++ b/megatron/core/transformer/cuda_graphs.py
@@ -864,11 +864,9 @@ class _CudagraphReplayNode(torch.autograd.Function):
         runner.bwd_graph_replay_complete_event.record(torch.cuda.current_stream())
         for param in runner.params_to_backprop:
             param._cudagraph_wgrad_ready_event = runner.bwd_graph_replay_complete_event
-            if id(param) in runner.checkpoint_param_grad_ids:
-                # Only nested checkpoint backward leaves param.grad after replay; ordinary
-                # parameters use the returned wgrad and leave it unset. Mark that gradient as
-                # already added so the standard DDP hook skips a duplicate add, then clears
-                # param.grad and registers readiness. The GTP path below does both explicitly.
+            if id(param) in runner.param_ids_to_skip_ddp_grad_add_on_replay:
+                # The backward graph already added param.grad to main_grad. Restore the
+                # Python-side flag so DDP skips that add, clears param.grad, and marks readiness.
                 param.grad_added_to_main_grad = True
         runner.status = _GraphStatus.FWD_READY
 
@@ -939,6 +937,7 @@ class _CudaGraphRunner(torch.nn.Module):
         self.fp4_runtime_enabled = None
         self.deallocate_pipeline_outputs = False
         self.num_warmup_steps = 0
+        self.needs_recompute_param_discovery = False
         self.use_stream = False
         self.gtp_remat = False
         self.fwd_side_streams = []
@@ -970,6 +969,9 @@ class _CudaGraphRunner(torch.nn.Module):
             self.backward_retain_grad = self.base_module.config.cuda_graph_retain_backward_graph
             self.deallocate_pipeline_outputs = self.base_module.config.deallocate_pipeline_outputs
             self.num_warmup_steps = self.base_module.config.cuda_graph_warmup_steps
+            self.needs_recompute_param_discovery = (
+                self.grad_enabled and self.base_module.config.recompute_granularity is not None
+            )
             self.fp8_enabled = self.base_module.config.fp8 is not None
             self.fp4_enabled = self.base_module.config.fp4 is not None
             self.fp8_runtime_enabled = None
@@ -1081,12 +1083,10 @@ class _CudaGraphRunner(torch.nn.Module):
         else:
             return nullcontext()
 
-    def get_connected_param_ids(self, outputs):
-        """Find base-module parameters reachable from the autograd graphs of ``outputs``.
-
-        The traversal identifies leaf parameters through their AccumulateGrad nodes and returns
-        only IDs of parameters recursively owned by ``base_module``.
-        """
+    def get_connected_params(self, outputs):
+        """Iterate through the autograd graph of 'outputs' and returns all parameters connected.
+        In theory this should return all parameters that return a nonzero wgrad when computing
+        the backward pass of 'outputs'."""
         # Flatten outputs and start traversal from roots that require gradients
         args = (outputs,) if torch.is_tensor(outputs) else outputs
         stack = [
@@ -1104,18 +1104,7 @@ class _CudaGraphRunner(torch.nn.Module):
                     p_ids.add(id(fn.variable))
                 stack.extend(f for f, _ in fn.next_functions if f)
 
-        module_param_ids = {id(param) for param in self.base_module.parameters()}
-        return p_ids & module_param_ids
-
-    def get_connected_params(self, outputs, warmup_param_ids=()):
-        """Build the ordered parameter inputs required by the captured backward graph.
-
-        This combines parameters reachable from the captured outputs with parameters discovered
-        while warmup bypassed CheckpointWithoutOutput. The result follows ``base_module`` parameter
-        order so the capture and replay input layouts are stable.
-        """
-        p_ids = self.get_connected_param_ids(outputs)
-        p_ids.update(warmup_param_ids)
+        # Return module params that were found in the graph, preserving original order
         return tuple(p for p in self.base_module.parameters() if id(p) in p_ids)
 
     def _weakref_forward_buffers(self, preserve_forward_to_backward_lifetimes: bool) -> None:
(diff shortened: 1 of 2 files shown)


JF-D:
@claude strict-review

kvareddy:
@jiemingz  can you please take a look at this MR?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1648,6 +1694,12 @@ def replay_graph_capture(self, is_first_microbatch, args, kwargs):
 
         inp_tensors = self.get_tensors(args, kwargs, check_types=False)
         if self.grad_enabled:
+            # Passing parameters as synthetic replay-node inputs creates autograd edges to their
```
jiemingz: this comment seems not necessary?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1280,7 +1299,16 @@ def _resolve_input_buffer(ten):
...
             # warmup again as case graph capture mode may execute a different codepath
             _set_warmup_start()
-            for _ in range(self.num_warmup_steps):
+
+            warmup_param_ids = None
+            # The outer autograd graph of CheckpointWithoutOutput has no edges to parameters used
+            # only by its recompute function. Warmup bypasses the checkpoint and runs that function
+            # with autograd, exposing those parameters through AccumulateGrad. Always run at least
+            # one training warmup so they can be included in the captured backward graph inputs.
+            num_internal_warmups = (
```
jiemingz: can you additionally guard only if the transformer config has recompute set? Ie creating a flag like`needs_recompute_param_discovery`

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1295,6 +1323,8 @@ def clone_ten(ten):
                 if self.grad_enabled:
                     warmup_outputs = self.get_tensors(warmup_outputs)
                     warmup_outputs = tuple(o for o in warmup_outputs if o.requires_grad)
+                    if warmup_param_ids is None:
+                        warmup_param_ids = self.get_connected_param_ids(warmup_outputs)
```
jiemingz: params are static, why can't we just do:
`self.params_to_backprop =  self.get_connected_param_ids(warmup_outputs)` ?

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1466,9 +1498,23 @@ def create_bwd_graph(self):
             n_act_grads = sum(
                 1 for i in self.fwd_graph_input_surface[: self.num_dgrads] if i.requires_grad
             )
+            self.checkpoint_param_grad_ids = set()
```
jiemingz: isnt this true broadly? if any function (recompute or not) returns a param.grad we need to accumulate it to main_grad here. The backward func returns `None` grads so all gradients have to be consumed in the graph.

JF-D: Yes, the tracking set serves a narrower replay purpose and has been renamed to param_ids_to_skip_ddp_grad_add_on_replay. It records only parameters whose param.grad was already transferred to main_grad inside the captured backward graph, but the grad_added_to_main_grad is not set by the DDP hook replay.

JF-D:
@claude strict-review

megatron/core/transformer/cuda_graphs.py
```diff
@@ -1466,9 +1482,17 @@ def create_bwd_graph(self):
             n_act_grads = sum(
                 1 for i in self.fwd_graph_input_surface[: self.num_dgrads] if i.requires_grad
             )
+            self.param_ids_to_skip_ddp_grad_add_on_replay = set()
             for param, wgrad in zip(self.params_to_backprop, grad_inputs[n_act_grads:]):
-                if wgrad is not None and not getattr(param, 'grad_added_to_main_grad', False):
+                grad_already_accumulated = getattr(param, 'grad_added_to_main_grad', False)
+                if wgrad is not None and not grad_already_accumulated:
                     param.main_grad.add_(wgrad)
+                if param.grad is not None and not grad_already_accumulated:
```
jiemingz, line 1485: do we even handle cases where `param.grad_added_to_main_grad = False` ? Because the backward replay returns no gradients, meaning all grads need to accumulated at this point. 
`param.grad_added_to_main_grad = False` means it needs to be accumulated by DDP but we return no grads.

JF-D: grad_added_to_main_grad is default to be False, and it's used to skip the accumulation of DDP hook when set to be true. Here, `grad_added_to_main_grad = False` means the grad hasn't been added to main_grad, so we added them into main grad in the bwd graph. And the grad_added_to_main_grad will be set to true afterwards.

JF-D: Here https://github.com/NVIDIA/Megatron-LM/pull/6373/changes#diff-a3f9fb509ad46e999a7d8e1b6850c40d2ca0a85344c749096008fefebf51ab08R867-R870 and https://github.com/NVIDIA/Megatron-LM/pull/6373/changes#diff-a3f9fb509ad46e999a7d8e1b6850c40d2ca0a85344c749096008fefebf51ab08R887-R888, we will set the grad_added_to_main_grad to be true.

JF-D:
@claude strict-review

jiemingz (approved):

yaoyu-33 (approved):

PR #6379, opened 2026-08-09
Balance LayerWise optimizer shards by Newton-Schulz cost, not parameter size

### LINKED ISSUE #6390: [feat] GTP+Muon

deepakn94:
## Problem

`LayerWiseDistributedOptimizer` assigns whole Muon matrices to `dp_size` shards, packing them largest-first onto the least-loaded shard. It measures "loaded" by `numel`, which breaks under GTP: a GTP-sharded weight stores only its local shard, but Newton-Schulz runs on the full all-gathered matrix, so its real cost is `gtp_remat_size` times what its `numel` suggests.

Ordering by `numel` therefore starts with the cheapest work and leaves the most expensive matrices to fill in wherever there is room, which piles them onto shards that are already busy. Every rank waits for the slowest, so the Muon step is gated on whichever shard collected them.

## Changes

1. `_ns_compute_cost()` estimates Newton-Schulz cost from the full post-all-gather shape (GTP shards along dim 0): `max(M, N) * min(M, N)^2`, the dominant term in the orthogonalization.
2. Order and assign by that cost instead of `numel`.
3. Track `shard_compute_loads` persistently across buckets, so expensive matrices spread over the whole buffer rather than clustering inside each bucket. `shard_cursors` still resets per bucket, which preserves the memory layout. A per-bucket numel cap (`_NUMEL_EPSILON = 0.3`) bounds the extra padding this can introduce, and assignment falls back to the previous least-numel rule when no shard satisfies the cap.
4. Correct the bucket-cutter comment from #5415. See Notes.

## Results

Measured end-to-end on Ultra-half (54-layer hybrid Mamba-MoE, 1024 GB200 GPUs, Muon + GTP=64): **2080 -> 1707 ms/iter, a 19% throughput improvement.**

## Tests

`tests/unit_tests/distributed/test_layer_wise_param_layout.py` gains `TestComputeBalancedLayout`, built so `numel` and cost disagree: three GTP-sharded matrices whose Newton-Schulz cost is 16x the largest dense matrix but whose `numel` is the smallest in the set. Numel-ordered placement puts all three on one shard for a 3.77x compute imbalance; cost-ordered placement spreads them across three shards for 1.33x. The test asserts both the spread and the imbalance bound, so it fails against the previous heuristic.

The pre-existing cases in `TestSizeMatchingLayout` all use 1-D params, where `_ns_compute_cost` falls back to `nelement()` and cost equals `numel`. They therefore pass identically under both heuristics and cannot detect this change; a class docstring now says so.

## Notes

`_emit_bucket` sorts a chunk before packing it, while the incremental estimate in the bucket cutter walks params in backprop order. #5415's comment claimed the two mirror each other. They do not, and after this PR they also differ in what they measure. Equal-sized params make both differences vanish; mixed sizes can send params to different shards, so absorbing occasionally enlarges a bucket instead of filling it.

@kunlunl raised this in review of #5415 with a counterexample, now pinned as `test_mixed_sizes_can_absorb_into_larger_bucket`: `dp_size=2`, backprop-order numels `[192, 192, 256, 128, 128, 192]`, `bucket_size=448` emits 1408 elements where closing at the threshold emits 1280. Documented rather than fixed, since an exact estimate would mean re-sorting the chunk on every param, and the target case is equal-sized expert matrices.

This helps without GTP too, though less dramatically. Newton-Schulz cost is `numel * min(M, N)`, so `numel` misvalues a matrix by its aspect ratio: across this model's shapes that factor ranges from 512 to 10240, meaning a tall-thin matrix and a square one of the same byte count do very different amounts of work. Balancing `numel` equalises bytes per shard; balancing cost equalises the Newton-Schulz FLOPs that actually gate the step. Without GTP the two orderings largely agree and the gain is small, but it grows with `dp_size`, where there is less slack to absorb a misranked matrix, and it is never worse.

The shared-embedding branch is untouched.

### DIFF
--- a/megatron/core/optimizer/layer_wise_optimizer.py
+++ b/megatron/core/optimizer/layer_wise_optimizer.py
@@ -213,15 +213,19 @@ class LayerWiseDistributedOptimizer(ChainedOptimizer):
            followed by an isolated bucket for that embedding alone.
         2. When the chunk's total numel reaches ``bucket_size`` (or all
            params have been consumed), bin-pack the chunk into ``dp_size``
-           shards via greedy LPT — sort by numel descending and assign each
-           param to the shard with the smallest current load.
+           shards: sort by estimated Newton-Schulz compute cost descending and
+           assign each param to the shard with the smallest accumulated compute
+           load, subject to a per-bucket numel cap that bounds shard-imbalance
+           padding. Compute loads persist across buckets so expensive
+           (GTP-sharded) matrices spread over the whole buffer instead of
+           clustering inside each bucket.
         3. Pad each shard to ``max(shard_cursors)`` aligned to
            :meth:`_shard_divisor`, then emit the bucket.
 
         Each bucket therefore spans a contiguous backprop range so that
         ``overlap_grad_reduce`` can dispatch the bucket's reduce-scatter as
         soon as the bucket's backward segment finishes — preserving the
-        original DDP overlap semantics.  LPT bin-packing keeps shards close
+        original DDP overlap semantics.  Greedy bin-packing keeps shards close
         to balanced; for uniform transformer blocks where ``params_per_layer
         * num_layers`` is a multiple of ``dp_size`` the packing is perfect.
 
--- a/megatron/core/optimizer/layer_wise_optimizer.py
+++ b/megatron/core/optimizer/layer_wise_optimizer.py
@@ -360,8 +360,11 @@ class LayerWiseDistributedOptimizer(ChainedOptimizer):
         chunk_params: List[torch.nn.Parameter] = []
         chunk_numel = 0
         chunk_max_param = 0
-        # Mirror _emit_bucket's greedy LPT placement incrementally so we can
-        # decide, per param, whether it still fits in the current bucket.
+        # Approximate _emit_bucket's placement so we can decide, per param,
+        # whether it still fits in the current bucket. _emit_bucket assigns by
+        # Newton-Schulz compute cost while this tracks numel, so the two can
+        # disagree. The estimate errs optimistic, which makes a bucket slightly
+        # larger than predicted rather than producing an invalid layout.
         shard_loads = [0] * dp_size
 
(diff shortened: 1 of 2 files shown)


deepakn94:
Both addressed in 8d204a8.

**1. Stale docstring.** Fixed, close to your suggested wording. Also changed "LPT bin-packing keeps shards close to balanced" a few lines below to "Greedy bin-packing", since that sentence carried the same assumption.

**2. Cross-bucket persistence untested.** You were right that `bucket_size=None` only exercised the single-bucket path. Added `test_gtp_params_land_on_different_shards_in_each_bucket`: four buckets, each holding one GTP-sharded matrix plus three dense params of *identical numel*, so numel cannot distinguish them. `shard_cursors` resets per bucket, so every bucket starts identical under a numel-keyed rule and all four expensive matrices would go to shard 0. Because `shard_compute_loads` carries over, each bucket sees the previous ones' cost and picks a different shard. Verified the counterfactual: with loads reset per bucket the assignment is `[0, 0, 0, 0]`; with persistence it is `[0, 1, 2, 3]`.

Also added `test_param_larger_than_cap_is_still_placed` for the branch you mentioned: `dp_size=2` with numels `[1024, 64, 64]` gives a cap of 748.8, so the 1024 param disqualifies every shard and takes the `candidates`-empty fallback. Confirmed it reaches that branch rather than passing for some other reason.

I did not add an assertion on the padding staying within the ~1.3x bound. The cap governs candidate *eligibility* rather than the emitted buffer, and the fallback can exceed it by design, so a bound assertion would either be vacuous or wrong. The persistence test does assert the bucket count, which pins the packing shape.

Phlip79 (approved):

jaredcasper (approved):

deepakn94:
/ok to test [4c62a5c](https://github.com/NVIDIA/Megatron-LM/pull/6379/commits/4c62a5c7334aebd6183730ef4db87ad93715da98)

kunlunl (approved):

PR #6380, opened 2026-08-09
GTP+Muon: support 'blockwise' and 'distributed' tp_mode

### LINKED ISSUE #6390: [feat] GTP+Muon

deepakn94:
## Summary

`TensorParallelMuon.scaled_orthogonalize_fn_with_gtp_remat` had one hard-coded strategy for GTP-sharded momenta: all-gather along the GTP dim, run whole-matrix Newton-Schulz redundantly on every GTP rank, slice the local shard back out. That ignores `self.tp_mode`, which already selects between `blockwise`, `duplicated` and `distributed` for TP-sharded params. This branches the GTP path on it so both axes behave consistently:

- **blockwise**: NS locally on the `[M/g, K]` GTP shard, no GTP collective.
- **duplicated**: the previous behavior, now explicit rather than implicit.
- **distributed**: NS distributed over GTP via the small-Gram all-reduce, so no rank does redundant full-matrix work.

Under `distributed` with both GTP and TP active, NS can only be distributed over one (group, dim) pair, so the code distributes over the larger group and all-gathers the smaller one first.

## Default `muon_tp_mode` changes from `blockwise` to `duplicated`

`blockwise` orthogonalizes each shard on its own, so the update rule, and with `scale_mode="spectral"` the effective learning rate, move whenever the parallelism config changes. `duplicated` and `distributed` orthogonalize the whole matrix and are TP-invariant.

This is wider than the GTP path: it changes behavior for any Muon run with TP > 1. Golden values are unaffected, since all six Muon functional tests run at TP and expert TP of 1, where the two modes are the same computation, and every Muon unit test sets `muon_tp_mode` explicitly. It also aligns the training-config default with `TensorParallelMuon.__init__`, which already defaulted to `duplicated`.

## Tests

`tests/unit_tests/generalized_tensor_parallel/test_gtp_muon.py` asserts per-shard parity against a full-matrix reference for all three modes:

- `test_gtp_only_parity` — TP1, `distributed` and `duplicated`, at GTP4 and GTP8.
- `test_row_parallel` / `test_col_parallel` — TP on dim 1 / dim 0, at `(tp, gtp)` of (2,2), (2,4), (4,2).
- `test_gtp_blockwise_mode` — TP1, at GTP4 and GTP8.

Every test in this directory required exactly 4 ranks while the unit-test recipe launches 8 GPUs, so a 4-GPU box was the only place they ran. Both shapes are now covered, and TP2xGTP4 / TP4xGTP2 additionally cover `tp_size != gtp_remat_size`, the case equal sizes cannot reach. Note this does not make the directory run in CI: `HAVE_GTP` is false in the CI container, whose TE has no hook registry, so the module-level guard fires before any world-size check. These need a GTP-capable 8-GPU node to exercise.

Parity is asserted at `num_ns_steps=1`: at more steps the distributed result is still mathematically identical, but the past-convergence NS coefficients amplify the ~1e-6 fp difference to ~1e-2 by step 5. The `blockwise` expected value is rebuilt from `newton_schulz` and `get_muon_scale_factor` rather than from `scaled_orthogonalize_fn`, so it does not restate the production branch and it pins the scale factor to the shard row count.

`test_gtp_muon_dcp.py` gains the standard `HAVE_GTP` module-level skip guard used elsewhere in the directory.

## Notes

Original implementation by @fanshiqing; both sign-offs are on the first commit.

### DIFF
--- a/megatron/core/optimizer/emerging_optimizers.py
+++ b/megatron/core/optimizer/emerging_optimizers.py
@@ -312,12 +312,17 @@ class TensorParallelMuon(OrthogonalizedOptimizer):
             return gathered_grad[gtp_rank * shard_size : (gtp_rank + 1) * shard_size].contiguous()
 
         # distributed: NS via the small-Gram all-reduce (no redundant full-matrix NS).
-        is_tp_active = (
+        # A momentum with both TP and GTP as sharding axes takes two communication steps: an
+        # all-gather that eliminates one axis, then the Gram all-reduce that distributes NS over
+        # the other. With GTP as the only sharding axis, the Gram all-reduce is the only
+        # communication needed. partition_dim is what says whether TP is a sharding axis here,
+        # the same signal scaled_orthogonalize_fn and newton_schulz_tp key off.
+        needs_two_step_communication = (
             partition_dim is not None and tp_group is not None and get_pg_size(tp_group) > 1
         )
 
-        if not is_tp_active:
-            # GTP-only: distribute NS over the GTP group on the local dim-0 row shard.
+        if not needs_two_step_communication:
+            # GTP is the only sharding axis: distribute NS over it on the local dim-0 row shard.
             return self.scaled_orthogonalize_fn(grad, gtp_remat_group, partition_dim=0)
 
         # GTP + TP: distributed NS can only operate over one (group, dim) at a
--- a/megatron/core/optimizer/emerging_optimizers.py
+++ b/megatron/core/optimizer/emerging_optimizers.py
@@ -258,13 +258,24 @@ class TensorParallelMuon(OrthogonalizedOptimizer):
             scaled_orthogonalize_fn=scaled_orthogonalize_fn,
         )
 
-    def scaled_orthogonalize_fn_with_gtp_remat(self, p, grad, tp_group, partition_dim):
-        """All-gather grad along GTP_remat/EGTP_remat dim 0, orthogonalize, then slice back.
+    @staticmethod
+    def _all_gather_tensor(t, group, dim):
+        """All-gather equal-size shards of ``t`` over ``group`` and concat along ``dim``."""
+        shards = [torch.empty_like(t) for _ in range(get_pg_size(group))]
+        torch.distributed.all_gather(shards, t.contiguous(), group)
+        return torch.cat(shards, dim=dim)
 
-        GTP_remat shards weights along dim 0 independently of TP's partition_dim. Newton-Schulz
-        needs the full weight matrix, so we reconstruct the GTP_remat dimension before running
-        the TP-aware orthogonalization, then extract the local GTP_remat shard from the result.
-        When GTP_remat is inactive this is a plain passthrough to scaled_orthogonalize_fn.
+    def scaled_orthogonalize_fn_with_gtp_remat(self, p, grad, tp_group, partition_dim):
+        """Orthogonalize a (possibly GTP-sharded) momentum, then reshard.
+
+        When GTP is inactive this is a plain passthrough to ``scaled_orthogonalize_fn``.
+        Otherwise, ``self.tp_mode`` controls how GTP sharding is handled:
+
+        - **blockwise**: orthogonalize the local GTP shard independently, no collective.
+        - **duplicated**: all-gather over GTP, run whole-matrix NS (TP-aware), reshard.
+        - **distributed**: distribute NS over GTP via small-Gram all-reduce. When both
+          GTP and TP are active, NS is distributed over the larger group to minimize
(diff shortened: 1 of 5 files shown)


megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -230,13 +230,24 @@ def scaled_orthogonalize_fn(
             scaled_orthogonalize_fn=scaled_orthogonalize_fn,
         )
 
+    @staticmethod
+    def _all_gather_tensor(t, group, dim):
+        """All-gather equal-size shards of ``t`` over ``group`` and concat along ``dim``."""
+        shards = [torch.empty_like(t) for _ in range(get_pg_size(group))]
```
fanshiqing, line 264: This line might introduce (1) worse memory fragmentaion and (2) cuda graphs imcompatibility and also (3) performance overhead for multiple small allocations

Can we figure out how to perform memory reuse or at least allocate single contiguous buffer instead?

deepakn94: Let me look into this.

mkhona-nvidia (approved):

deepakn94:
/ok to test [f9be365](https://github.com/NVIDIA/Megatron-LM/pull/6380/commits/f9be3657bd9dbff635608d020d5d6e7f73eb0c05)

jaredcasper (approved):

megatron/core/optimizer/emerging_optimizers.py
deepakn94: I can try doing this in a separate PR. Outside the scope of this PR IMO.

Phlip79 (approved):

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -256,21 +267,56 @@ def scaled_orthogonalize_fn_with_gtp_remat(self, p, grad, tp_group, partition_di
...
-        shards = [torch.empty_like(grad) for _ in range(gtp_remat_size)]
-        torch.distributed.all_gather(shards, grad, gtp_remat_group)
-        gathered_grad = torch.cat(shards, dim=0)
 
-        gathered_grad = self.scaled_orthogonalize_fn(gathered_grad, tp_group, partition_dim)
+        if self.tp_mode == "blockwise":
+            # Local block NS on this rank's GTP row-shard (shape [M/gtp_remat_size, K]):
+            # partition_dim=None makes scaled_orthogonalize_fn run a plain Newton-Schulz on
+            # the shard with no GTP/TP collective.
+            return self.scaled_orthogonalize_fn(grad, tp_group, None)
```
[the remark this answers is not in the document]
deepakn94: Fixed the default to `duplicated` in this PR.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -256,21 +267,56 @@ def scaled_orthogonalize_fn_with_gtp_remat(self, p, grad, tp_group, partition_di
...
 
-        shard_size = gathered_grad.shape[0] // gtp_remat_size
-        return gathered_grad[gtp_rank * shard_size : (gtp_rank + 1) * shard_size].contiguous()
+        gathered_grad = self._all_gather_tensor(grad, smaller_group, smaller_dim)
+        orthogonalized_grad = self.scaled_orthogonalize_fn(gathered_grad, larger_group, larger_dim)
+        shard_size = orthogonalized_grad.size(smaller_dim) // get_pg_size(smaller_group)
+        reshard_rank = get_pg_rank(smaller_group)
+        return orthogonalized_grad.narrow(
+            smaller_dim, reshard_rank * shard_size, shard_size
+        ).contiguous()
```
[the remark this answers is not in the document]
deepakn94:
```
Take P a permutation matrix and look at one NS step, X ← aX + b(XX^T)X + c(XX^T)²X:

(PX)(PX)^T = P (XX^T) P^T
(PXX^TP^T)(PX)  = P (XX^T) P^T P X = P (XX^T) X     since P^T P = I
(PXX^TP^T)²(PX) = P (XX^T)² X                        same cancellation

So NS_step(PX) = P · NS_step(X). The P^T from the Gram always cancels against the P on the trailing factor. Normalization is fine too, since ‖PX‖_F = ‖X‖_F. Each step is equivariant, and a composition of equivariant maps is equivariant, so it holds for any num_ns_steps and any coefficient schedule.
```

cspades (approved):

megatron/core/optimizer/emerging_optimizers.py
fanshiqing: okay, that's fine

fanshiqing (approved):

PR #6381, opened 2026-08-09
Add --muon-use-syrk option for Triton SYRK kernel in Newton-Schulz

deepakn94:
## Summary

Adds an opt-in `--muon-use-syrk` flag that routes the Gram-matrix computation inside Newton-Schulz through the Triton SYRK kernel in `emerging_optimizers`. The Gram matrix `X @ X.T` is symmetric, so SYRK computes roughly half the FLOPs of the equivalent GEMM.

Off by default, so behavior is unchanged unless the flag is set.

## Wiring

- `OptimizerConfig.muon_use_syrk` (default `False`).
- `--muon-use-syrk` in `_add_regularization_args`.
- `use_syrk` parameter on `TensorParallelMuon.__init__` and `TensorParallelAdaptiveMuon.__init__`.

No plumbing code is needed between the config and the optimizer: `_kwargs_from_config` already maps `muon_<name>` config fields onto matching `__init__` parameters, so `muon_use_syrk` reaches `use_syrk` automatically.

## Version compatibility

`newton_schulz_tp` — the only Newton-Schulz entry point Megatron calls — gained the `use_syrk` kwarg in **emerging_optimizers 0.4.0**:

| emerging_optimizers | `newton_schulz_tp` accepts `use_syrk` |
|---|---|
| v0.2.0 | no |
| v0.3.0 | no |
| v0.3.1 | no |
| **v0.4.0** | **yes** |

Note that `use_syrk` appears on the *non-TP* `newton_schulz` as far back as v0.2.0, so the presence of the name in `muon_utils.py` is not a reliable signal; only the TP variant matters here.

The check is a version gate built on a new `get_emerging_optimizers_version()` / `is_emerging_optimizers_min_version()` pair in `megatron/core/utils.py`, following the existing `is_te_min_version` / `is_mamba_min_version` / `is_causal_conv1d_min_version` helpers. The minimum is spelled `0.4.0.dev0` so pre-release builds of that line are accepted, matching how the TransformerEngine minimums are written elsewhere in the tree.

**Requesting `--muon-use-syrk` against an older release raises `ValueError` at optimizer construction** rather than warning and silently falling back. A perf flag that appears to take effect but doesn't is worse than a hard failure, and the fallback made it impossible to tell from the logs whether SYRK was actually running.

## Dependency pin

`pyproject.toml` currently pins `emerging_optimizers` at `rev = "v0.2.0"`, so `--muon-use-syrk` will raise under the locked dependency until that pin moves to v0.4.0. This is intentional for now: the flag is opt-in and aimed at runs on newer builds, and bumping the pin pulls in everything else that changed across 0.2.0 → 0.4.0 plus a `uv.lock` regeneration, which belongs in its own PR.

### DIFF
--- a/megatron/core/optimizer/emerging_optimizers.py
+++ b/megatron/core/optimizer/emerging_optimizers.py
@@ -17,7 +17,13 @@ import torch
 from torch.optim.optimizer import ParamsT
 
 from megatron.core.process_groups_config import ProcessGroupCollection
-from megatron.core.utils import get_pg_rank, get_pg_size, log_single_rank
+from megatron.core.utils import (
+    get_emerging_optimizers_version,
+    get_pg_rank,
+    get_pg_size,
+    is_emerging_optimizers_min_version,
+    log_single_rank,
+)
 
 from .optimizer_config import ParamKey, ParamPredicate
 
@@ -29,8 +35,6 @@ try:
         get_muon_scale_factor,
     )
     from emerging_optimizers.orthogonalized_optimizers.muon_utils import NSCoeffT, newton_schulz_tp
-
-    _NS_TP_SUPPORTS_SYRK = "use_syrk" in inspect.signature(newton_schulz_tp).parameters
 
     # It is necessary to import optimizers for the registry to work.
     from emerging_optimizers.scalar_optimizers import Lion  # pylint: disable=unused-import
@@ -44,6 +48,12 @@ except ImportError:
 
 
 logger = logging.getLogger(__name__)
+
+# newton_schulz_tp() gained the use_syrk kwarg in emerging_optimizers 0.4.0. Earlier releases
+# expose use_syrk on the non-TP newton_schulz() only, so 0.3.x still rejects it here. Spelled
+# ".dev0" so pre-release builds of that line are accepted too, matching how the TE minimums
+# elsewhere in the tree are written.
+_SYRK_MIN_EO_VERSION = "0.4.0.dev0"
 
 
 def get_supported_coefficient_types() -> tuple[str, ...]:
@@ -184,12 +194,11 @@ class TensorParallelMuon(OrthogonalizedOptimizer):
     ) -> None:
         if num_ns_steps < 1:
             raise ValueError(f"num_ns_steps must be at least 1, got {num_ns_steps}")
-        if use_syrk and not _NS_TP_SUPPORTS_SYRK:
-            log_single_rank(
-                logger,
-                logging.WARNING,
-                "use_syrk requested but the installed emerging_optimizers does not support it; "
-                "falling back to standard GEMM.",
+        if use_syrk and not is_emerging_optimizers_min_version(_SYRK_MIN_EO_VERSION):
+            raise ValueError(
+                f"use_syrk requires emerging_optimizers >= {_SYRK_MIN_EO_VERSION}, but "
+                f"{get_emerging_optimizers_version()} is installed. Upgrade "
+                "emerging_optimizers or drop --muon-use-syrk."
             )
 
         def scaled_orthogonalize_fn(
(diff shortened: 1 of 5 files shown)


megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -30,6 +30,8 @@
     )
     from emerging_optimizers.orthogonalized_optimizers.muon_utils import NSCoeffT, newton_schulz_tp
 
+    _NS_TP_SUPPORTS_SYRK = "use_syrk" in inspect.signature(newton_schulz_tp).parameters
```
mkhona-nvidia: What is this

mkhona-nvidia (approved):

megatron/core/optimizer/emerging_optimizers.py
deepakn94: Versions prior to 0.3.0 don't have SYRK supported. I guess maybe I should do a version check directly.

cspades (approved):

Phlip79 (approved):

jaredcasper (approved):

megatron/core/optimizer/emerging_optimizers.py
deepakn94: Done — replaced the `inspect.signature` probe with a real version check.

It turned out to be v0.4.0 rather than v0.3.0. `use_syrk` does appear in `muon_utils.py` as far back as v0.2.0, but only on the non-TP `newton_schulz`; `newton_schulz_tp` (the only Newton-Schulz entry point Megatron calls) gained it at v0.4.0. Checked tag by tag:

| emerging_optimizers | `newton_schulz_tp` accepts `use_syrk` |
|---|---|
| v0.2.0 / v0.3.0 / v0.3.1 | no |
| v0.4.0 | yes |

The gate is now `is_emerging_optimizers_min_version("0.4.0.dev0")`, backed by a new `get_emerging_optimizers_version()` / `is_emerging_optimizers_min_version()` pair in `megatron/core/utils.py` following the existing `is_te_min_version` / `is_mamba_min_version` helpers. It raises rather than warning-and-falling-back, so a no-op perf flag can not go unnoticed.

deepakn94:
Addressed the review feedback and rebased onto latest `main`.

**@claude's two items:**

1. **`TensorParallelAdaptiveMuon` docstring** — added the missing `use_syrk` entry after `tp_mode`.

2. **Test coverage for the version gate** — added two parametrized tests over both `TensorParallelMuon` and `TensorParallelAdaptiveMuon`:
   - `test_muon_use_syrk_rejected_on_old_emerging_optimizers` — asserts the `ValueError` fires when the version helper reports an older release. This is the only reachable branch under CI today given the `v0.2.0` pin, so it was worth covering.
   - `test_muon_use_syrk_default_off_ignores_version` — asserts the gate does *not* fire on the default path, so an old `emerging_optimizers` still constructs fine when `use_syrk` is left off.

**Correction to the version number.** The gate is `0.4.0.dev0`, not `0.3.0` as I said in the earlier thread. `use_syrk` exists on the non-TP `newton_schulz` back to v0.2.0, but `newton_schulz_tp` — the only Newton-Schulz entry point Megatron calls — only gained it at v0.4.0. A `0.3.0` minimum would have let v0.3.x through the check and then failed with `TypeError: newton_schulz_tp() got an unexpected keyword argument 'use_syrk'` at the first optimizer step, which is worse than the warning this replaces.

Note `pyproject.toml` still pins `emerging_optimizers` at `rev = "v0.2.0"`, so `--muon-use-syrk` raises under the locked dependency until that pin moves to v0.4.0. Left for a separate PR since it needs a `uv.lock` regeneration and pulls in everything else that changed across 0.2.0 → 0.4.0.

PR #6385, opened 2026-08-10
Telemetry/otel lens upstream

rhewett-nv:
## What this adds

This PR instruments Megatron's training, checkpointing, and inference paths with
OpenTelemetry spans and metrics. It builds on the core telemetry module (added
separately) and is purely the call-site instrumentation: roughly 53 named spans
across ten source files, organised as a hierarchical name taxonomy so a name
prefix selects a subtree. The main families are `megatron.startup.*` (12 spans:
imports, arg parsing, `initialize_megatron`, JIT fusion setup, model init,
dataloader construction, weight-hash check, sniff test), `megatron.checkpoint.*`
(13: the exposed save path, state-dict generation, IO write, async finalize,
load, plus the memory-reclaim and GC work that blocks the training loop), and
`megatron.train.*` (9: per-iteration spans, the one-off first iteration, param-norm
reduction, logging, CUDA-graph capture). Finer-grained groups cover per-layer
attention/MLP/Mamba breakdown, microbatch forward/backward, P2P communication,
gradient sync, and activation offload.

## How it's controlled

Everything is gated behind span *groups* selected at runtime with
`--otel-span-groups`, which resolves either a preset (`default`, `per_step`,
`profiling`, `all`) or an explicit comma-separated list. A disabled group costs
nothing: `managed_span` checks the group before creating a span, and the
`trace_fn` decorator returns the wrapped function unchanged, so the instrumentation
is a genuine no-op rather than a cheap-but-nonzero one. `nemo-lens` itself is an
optional dependency — when it is absent every call site falls back to no-op shims,
and telemetry setup returns a null handle. With telemetry off, this change does not
alter training behaviour: no additional collectives, no barriers, no change to RNG
state, numerics, checkpoint contents, or control flow.

## Metrics, inference, and goodput

Alongside spans, the per-iteration logging path records OTel metrics — step
duration, loss, throughput (TFLOP/s), gradient norm, learning rate, tokens/sec,
skipped iterations, and peak memory — read from the values already computed for
the existing log line, so nothing new is calculated on the critical path. The
inference server is instrumented to the GenAI semantic conventions: request spans
carry `gen_ai.operation.name`, `gen_ai.provider.name`, model, max tokens,
temperature, and sampling parameters, and W3C trace context is extracted from
inbound request headers so a caller's trace continues into Megatron. Spans that
represent resiliency overhead — checkpoint save/load, restart-path startup work —
opt in to an `is_goodput_span` attribute, which lets a collector route the goodput
subset without needing to know Megatron's span taxonomy.

## Design notes

The training loop is emitted as one trace per checkpoint interval rather than a
single run-long trace, with each interval linked back to the startup trace and to
its predecessor. This keeps individual traces small enough for backends to render,
and means a run killed mid-flight still exports every completed interval instead of
losing one enormous unclosed span. Startup is likewise its own trace. Run identity
(`run_uuid`, SLURM job/array/SLUID, restart count, GPU identity) rides on resource
attributes, so traces from separate incarnations of the same job correlate without
a shared parent span. Spans that outlive the function that opens them use the
explicit `start_span` API with `try/finally` cleanup so a raising body cannot leak
a span or mis-parent later ones.


Phlip79:
This PR is too large. Can you please have this PR add the telemetry logic and documention? Then a subsequent PR to use the telemetry logic?

deepakn94:
Closing in favor of #6438

PR #6388, opened 2026-08-10
[fix] GTP+DDP potential data-race: publish DDP params before GTP prefetch reads them

### LINKED ISSUE #6389: [fix] GTP+DDP potentail data-race in the fwd pass
## Issue description: GTP `ahead-of-consume` weight prefetch vs. DDP's `asynchronous parameter all-gather`.

- With ``overlap_param_gather=True`` DDP publishes a bucket group only from the forward pre-hook
of a module owning one of its parameters. 
- GTP prefetches ahead of that module, and its gather input is ``next_w.data`` -- a view into the bucket's ``param_data``. 
- So when the consumed weight and the prefetch target sit in different bucket groups, GTP reads a buffer whose all-gather is still in flight or not yet dispatched, and whose post-gather quantize has not run: it gathers
last iteration's weights.

fanshiqing:

# Summary
[fix] GTP+DDP potential data-race: publish DDP params before GTP prefetch reads them

- BUG: GTP prefetches ahead of the module owning the weight, and its all-gather input is `param.data` -- a view into the DDP bucket buffer. With `--overlap-param-gather` that bucket may still be mid-gather, so GTP **_silently gathers the PREVIOUS iteration's weights_**. Wrong results, no crash.

- FIX: backends attach a callable under `PARAM_READY_ATTR`; consumers call `ensure_params_ready()` before reading. DDP registers one per bucket group and stays GTP-agnostic. GTP calls it on forward gathers only.

# Note

Originally reported by @JF-D and has one preliminary fix in https://github.com/NVIDIA/Megatron-LM/pull/6232; both sign-offs are on the commit.

This MR steps further from #6232:
- `DDP is GTP agnostic`. distributed_data_parallel.py has zero GTP references (grep-verified). 
- **Recompute excluded** — the baseline includes recompute-forward, which runs inside backward, where publishing could dispatch an all-gather into the grad-aliased buffer (--reuse-grad-buf-for-mxfp8-param-ag).
- `disable_forward_pre_hook(param_sync=False)` — the baseline returns early and never drains an in-flight gather; ours drains it (with skip_next=True).
- `Stale marker on re-wrap` — the baseline's persists and can point at the wrong bucket group; ours clears it.
- `Weakrefs` — the baseline holds bucket_group strongly; ours holds both weakly.
- `Hot path ~2.7× cheaper` — 190 ns vs ~517 ns (the baseline rebuilds a list+set per call and checks graph-capture before the published check). ≈ 5 ms → 1.8 ms/iter at 600 weights × 16 microbatches.
- More rubut tests— ours: real 4-rank repro, verified to fail with the fix disabled, plus grouped-chain and recompute coverage. 

### Pre-checks


megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1223,6 +1223,18 @@ def _all_gather_weight(self, async_op, fwd, nvtx_label=None):
...
+        # 0. Publish the shards before reading them. EVERY chain prefetches ahead of the module
+        #    that owns the target weight -- default chains by one weight, grouped-expert chains by
+        #    one MoE block -- so that module's DDP pre-hook may not have run and the shard can
+        #    still hold last iteration's values. Reading further ahead widens the window, not
+        #    creates it. No-op when no backend owns the parameter, or it is already published.
+        #
+        #    FORWARD ONLY. Backward re-reads what forward published; recompute is a forward
+        #    replayed inside backward, where publishing could dispatch an all-gather into the
+        #    buffer that aliases grads under --reuse-grad-buf-for-mxfp8-param-ag.
+        if fwd and not in_fp8_activation_recompute_phase():
```
JF-D: Can we also check if it is in activation recompute phase in other precisions, like bf16 or fp16?

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -334,9 +386,24 @@ def __init__(
...
                 for bucket in bucket_group.buckets:
                     for param in bucket.params_list:
                         self.param_to_bucket_group[param] = bucket_group
+                        if readiness is not None:
+                            setattr(param, PARAM_READY_ATTR, readiness)
+                        elif hasattr(param, PARAM_READY_ATTR):
+                            # Re-wrapping a model chunk: a previous DDP may have left a marker
+                            # pointing at ITS bucket group. This DDP owns the parameter now and
+                            # publishes nothing, so the stale callback must go.
+                            delattr(param, PARAM_READY_ATTR)
```
JF-D: Do we need to consider the case that rewraps the same model chunk after changing a parameter’s requires_grad value? The current cleanup only visits parameters included in the new DDP bucket lists. A parameter managed by the previous DDP but excluded from the new buckets after requires_grad=False would retain its old PARAM_READY_ATTR.

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -334,9 +386,24 @@ def __init__(
         # Create map from param to bucket group, used in pre_hook.
         for bucket_groups in [self.bucket_groups, self.expert_parallel_bucket_groups]:
             for bucket_group in bucket_groups:
+                # One readiness callback per bucket group, stamped on all its params, so a
+                # consumer reading param.data outside the owning module's pre-hook can publish it
+                # first. Backend-agnostic: DDP never learns which consumers use it.
+                readiness = (
```
deepakn94: I think the name on this should be better. `is_ready_callback`?

fanshiqing: `is_ready_callback`: the `is_` prefix conventionally marks a boolean predicate, which is not the case here

`ready_callback` seems looks better, I will take this one

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
fanshiqing: Sure. This func name (defined in TE) is just misleading, it already covers it and actually never gated by 'fp8'

fanshiqing: You can also check with TE: https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/distributed.py#L241-L286

That's why GTP also works with bf16 :)

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: sorry but when should it happen for `the case that rewraps the same model chunk after changing a parameter’s requires_grad value?`

JF-D (approved):

megatron/core/utils.py
```diff
@@ -3118,3 +3118,40 @@ def deprecate_inference_params(inference_context, inference_params):
         )
         return inference_params
     return inference_context
+
+
+#: Attribute a parameter-sharding backend sets on each parameter it publishes asynchronously.
+#: See :func:`ensure_params_ready`.
+PARAM_READY_ATTR = "_ensure_param_ready"
```
deepakn94: "PARAM_READY_CALLBACK"?

megatron/core/utils.py
```diff
@@ -3118,3 +3118,40 @@ def deprecate_inference_params(inference_context, inference_params):
...
+    asynchronously, so only the owning module's forward pre-hook makes ``param.data`` valid. Any
+    consumer reading it earlier -- ahead of that module, or from another stream -- calls this
+    first. Backends mark their params with :data:`PARAM_READY_ATTR`; unmarked params no-op, so
+    neither side needs to know about the other.
+
+    Hot path: called once per weight per forward, so weights x microbatches times per iteration.
+    Only the first microbatch actually publishes a bucket, so nearly every call is a no-op --
+    hence it allocates nothing unless one call spans two or more buckets.
+    """
+    called = None  # last callback fired
```
deepakn94: Can you briefly summarize the logic in this function? I don't quite follow.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1223,6 +1223,18 @@ def _all_gather_weight(self, async_op, fwd, nvtx_label=None):
 
         weights = self._weights
 
+        # 0. Publish the shards before reading them. EVERY chain prefetches ahead of the module
+        #    that owns the target weight -- default chains by one weight, grouped-expert chains by
+        #    one MoE block -- so that module's DDP pre-hook may not have run and the shard can
```
deepakn94: Is this trying to say we need to wait until the params we depend on are ready?

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
deepakn94: "fp8_activation_recompute_phase" seems very specific, how about if we're doing activation recomputation with BF16?

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -11,14 +12,65 @@
...
-from ..utils import log_single_rank
+from ..utils import PARAM_READY_ATTR, log_single_rank
 from .data_parallel_base import _BaseDataParallel
 from .distributed_data_parallel_config import DistributedDataParallelConfig
 from .param_and_grad_buffer import _ParamAndGradBuffer, group_params_for_buffers, partition_buckets
 
 logger = logging.getLogger(__name__)
 
 
+class _BucketParamReadiness:
```
deepakn94: Let's add `Callback` in the name here.

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -11,14 +12,65 @@
...
+        if bucket_group is None:
+            return
+
+        # HOT PATH: already published. True for every microbatch after the first one of an
+        # iteration, so it precedes the DDP deref and the graph-capture query.
+        if bucket_group.param_gather_dispatched and bucket_group.param_gather_handle is None:
+            return
+
+        ddp = self._ddp()
+        if ddp is None or is_graph_capturing():
```
deepakn94: Add comment as to why you can essentially no-op here.

megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -11,14 +12,65 @@
...
+            return
+
+        # `ddp.remove_forward_pre_hook_handles` empty means the caller removed the pre-hooks and
+        # now drives param sync itself (``disable_forward_pre_hook(param_sync=False)``), so we
+        # must not touch its dispatch schedule.
+        owns_schedule = bool(ddp.remove_forward_pre_hook_handles)
```
deepakn94: Better variable name? Not clear who is owning the schedule...`ddp`?

megatron/core/utils.py
fanshiqing: Summarize: for all params in one ddp-bucket-group, the `param_ready_callback` should be only called only once.
I have simplified this for-loop to make it looks clear, with no significant increasement in pytorch ops.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
fanshiqing: Yes, just refined the docs and make it more highlight and easier to understand

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
fanshiqing: `fp8_activation_recompute_phase` it's name is really misleading... it also works for bf16 too, not guarded by fp8-dtype under the hood. 

it's a TE side's function

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: Done

megatron/core/distributed/distributed_data_parallel.py
fanshiqing: yeah, that should looks better

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
fanshiqing: here is related implementation in TE: https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/distributed.py#L284-L286

it's not gated by fp8, and I also have verified locally for bf16 this function also works

Phlip79 (approved):

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
deepakn94: Can we alias it in our code to make things easier to understand? That is, do something like `from transformer_engine.pytorch.distributed import in_fp8_activation_recompute_phase as in_activation_recomputate_phase`?

megatron/core/utils.py
```diff
@@ -3118,3 +3118,28 @@ def deprecate_inference_params(inference_context, inference_params):
         )
         return inference_params
     return inference_context
+
+
+#: Attribute a parameter-sharding backend sets on each parameter it publishes asynchronously.
+#: See :func:`ensure_params_ready`.
+PARAM_READY_CALLBACK_ATTR = "_ensure_param_ready"
```
deepakn94: Make  the attribute also have "callback" in the name? "_ensure_param_ready_callback"?

deepakn94 (approved):

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
fanshiqing: good idea~

fanshiqing: Done~

PR #5936, opened 2026-07-21
feat(transformer): add native manifold hyper-connections

FDecaYed:
## What

Add the model-neutral native PyTorch manifold hyper-connection primitives:

- expand/reduce residual-stream operations and learned mappings;
- Sinkhorn projection and initialization controls;
- selective output-discarding recompute through CheckpointManager;
- TransformerConfig fields and compatibility validation;
- focused numerical, gradient, recompute, and serialization tests.

This slice deliberately excludes fused kernels and excludes TransformerBlock, TransformerLayer, GPTModel, HybridModel, pipeline, MTP, and recipe wiring. Those concerns belong to smaller dependent slices.

## Stack

- Depends on #5935.
- Temporary review base: pull-request/5935.
- Files changed contains only the mHC primitive/config/test slice.
- Before #5935 merges, retarget and refresh this PR to main to preserve review history.

## Provenance

- Reconstructs the native mHC work from #2943 and the current main proposal #4531.
- Refactors the same implementation captured in #5795 (commit 07c0d088).
- Original implementation by the #2943/#4531 authors; DSv4 adaptation by @hxbai.
- Feature cutoff: 2026-07-21. GPTModel integration is intentionally excluded.

## Testing

- uv run isort --check-only on all changed Python files
- uv run ruff check on all changed Python files
- uv run python -m compileall -q on all changed Python files
- Focused tests are included; local execution requires the project Torch runtime and is left to stack CI.

## Stack follow-up

- #5943 contains the PP/VPP tensor-shape slice and temporarily targets this PR mirror.

PR #6401, opened 2026-08-10
Add fused mHC support for HybridModel

FDecaYed:
# Add fused mHC support for HybridModel

## Summary

Add optional fused manifold hyper-connection (mHC) execution and complete HybridModel/MTP integration. This includes fused kernels with native fallback, Hybrid layer execution, selective-recompute handling, MTP multistream propagation, selective-FP32 mapping, and focused unit coverage.

Related to #5795.

## Scope

- Add optional Triton/cuTile fused mHC operations with numerical and gradient tests.
- Integrate fused mHC execution with HybridModel layers and MTP.
- Preserve mHC multistream and raw-branch behavior through MTP, Hidden State Mixing, and selective recompute.
- Keep the wrapper in `megatron/core/models/hybrid/layers/hybrid_hyper_connection.py` while preserving state-dict key paths.
- Preserve wrapped-GDN dynamic-inference state discovery and ModelOpt Hybrid mHC+MTP construction on current `main`.
- Use the final `enable_mhc_connections` / `mhc_num_residual_streams` APIs and preserve #4531 fail-fast guards.
- Preserve the selective-FP32 mapping correctness from dev #6172 (`d8b71082ea7033974703a101ded49a359b773458`).

## Non-goals

- Pipeline/virtual-pipeline mHC transport, full recompute, FP32 residual mode, or fused TP inference.
- Hash MoE routing; that remains a separate follow-up.
- GPTModel support, DSv4 attention, recipes, or unrelated CUDA-graph expansion.

## Review boundary and dependencies

#5929 and #4531 are merged in `main`. The branch is rebased onto `main@dba084bf7a7d2b43c7317cfefb4be35e1a298941` and contains four signed residual commits:

- `151248d25628b59b65f3844c0640f2c660f5121d` — fused mHC and HybridModel/MTP integration;
- `9a880f441e04c445ed6f38f110ad31292625427e` — determinism metadata;
- `80720c869911bb46bcbe3e134b8b6f4e4670949a` — Hybrid wrapper module move;
- `4f2e0e3b69f9366b7faa5b703cd5ecb58769a09b` — current-main GDN and ModelOpt adaptations.

Please review the complete PR diff: **17 files, +6,045/-250**.

## Provenance

Reconstructed from the mHC behavior tracked by #5795, the final native implementation in #4531, and the earlier focused recut #5936. Original implementation credit remains with the #2943/#3430/#4531 contributors and the #5795 DSv4 adaptation authors.

## Validation

Passed locally on the final changeset:

- clean worktree;
- valid GPG signatures and required `Signed-off-by` trailers on all four commits;
- `git diff --check`;
- Python parsing and Ruff on all 17 changed Python files;
- isort 5.13.2 on all changed Python files;
- Black 26.3.1 on all changed Python files;
- focused line-length lint.

The macOS host has no usable Torch/CUDA runtime, so no local GPU/runtime result is claimed. Full functional CI is required on exact head `4f2e0e3b69f9366b7faa5b703cd5ecb58769a09b`. No dependency or lock-file change is part of this branch.

### DIFF
--- a/megatron/core/fusions/fused_mhc_kernels.py
+++ b/megatron/core/fusions/fused_mhc_kernels.py
@@ -12,6 +12,11 @@ Four fused operations:
   - h_aggregate:         weighted n-stream -> 1-stream aggregation
   - h_post_bda:          fused H_res.T @ residual + H_post * (x + bias)
   - proj_rms_compute_h:  fused projection + RMS normalization + compute_h
+
+Determinism:
+  The native, Triton, and cuTile backends are currently tagged ``unknown``.
+  Numerical and gradient parity are covered, but bit-exact repeatability has
+  not been certified. The accelerated backends use timing-based autotuning.
 """
 
 import logging
@@ -29,6 +34,16 @@ from megatron.core._rank_utils import log_single_rank, safe_get_rank
 
 logger = logging.getLogger(__name__)
 LOG2E = math.log2(math.e)
+
+#: Bit-exact determinism status for each concrete mHC backend. This uses the
+#: same three-value vocabulary proposed for operation-backend metadata while
+#: the ops registry is not available on this PR's base. ``auto`` is a
+#: per-operation selection policy, not a backend of its own.
+MHC_BACKEND_DETERMINISM: dict[str, str] = {
+    "native": "unknown",
+    "triton": "unknown",
+    "cutile": "unknown",
+}
 
 
 def _env_flag(name: str) -> bool:
--- a/megatron/core/fusions/fused_mhc_kernels.py
+++ b/megatron/core/fusions/fused_mhc_kernels.py
@@ -2,10 +2,11 @@
 
 """Fused kernels for mHC (Manifold-Constrained Hyper-Connections).
 
-Uses Triton and cuda.tile (cuTile) kernels when available, with PyTorch
-reference implementations as fallback.  Reference (non-fused) implementations
-live in ``megatron.core.transformer.hyper_connection`` and are used when fused
-kernels are unavailable or when the ``use_fused_mhc`` config flag is False.
+Uses Triton and cuda.tile (cuTile) kernels according to an explicit backend
+policy. With the default ``auto`` policy, unavailable accelerated operations
+fall back to PyTorch reference implementations. Reference implementations live
+in ``megatron.core.transformer.hyper_connection`` and are also used when the
+``use_fused_mhc`` config flag is False.
 
 Four fused operations:
   - sinkhorn:            Sinkhorn-Knopp projection to doubly stochastic matrix
@@ -25,7 +26,7 @@ import os
 import shutil
 import subprocess
 import warnings
-from typing import Optional, Tuple
+from typing import Literal, Optional, Tuple
 
 import torch
 from torch import Tensor
@@ -37,7 +38,7 @@ LOG2E = math.log2(math.e)
 
 #: Bit-exact determinism status for each concrete mHC backend. This uses the
 #: same three-value vocabulary proposed for operation-backend metadata while
-#: the ops registry is not available on this PR's base. ``auto`` is a
+#: the ops registry is unavailable. ``auto`` is a
 #: per-operation selection policy, not a backend of its own.
 MHC_BACKEND_DETERMINISM: dict[str, str] = {
     "native": "unknown",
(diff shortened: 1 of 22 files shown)


zhongbozhu (approved):

guihong-nv:
@FDecaYed For the new fused mHC kernel, could we add some tags to clarify if it's deterministic or not?

FDecaYed:
@guihong-nv Thanks for the suggestion. I added explicit determinism metadata in `2651afeebeaa7b1f41945203818302445bdde334` for the native, Triton, and cuTile mHC backends; all three are currently tagged `unknown` because numerical and gradient parity is covered, but bit-exact repeatability has not yet been certified.

This is a non-functional metadata-and-coverage-only change and does not alter backend selection or kernel execution. The ready-PR auto-sync briefly queued a CI run for the new SHA, so I cancelled it rather than spend a full CI cycle on this change. Once all review feedback is complete, I will consolidate any final changes and run the final full functional CI on that head.

guihong-nv (approved):

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -52,6 +65,201 @@ class HybridStackSubmodules:
     mtp_block_spec: Optional[ModuleSpec] = None
 
 
+class HyperConnectionHybridLayer(MegatronModule):
```
Phlip79: We want to ensure `hybrid_block.py` remains generic. How about we put this class in  a new directory and file `megatron/core/models/hybrid/layers/hybrid_hyper_connection.py`?

Phlip79 (approved):

ChenhanYu (approved):

megatron/core/fusions/fused_mhc_kernels.py
```diff
@@ -0,0 +1,3043 @@
...
+    "cutile": "unknown",
+}
+
+
+def _env_flag(name: str) -> bool:
+    return os.getenv(name, "0").lower() in ("1", "true", "yes", "on")
+
+
+def _forced_backend() -> Tuple[str, Optional[Exception]]:
+    value = os.getenv("MHC_FORCE_BACKEND", "auto").strip().lower()
```
jaredcasper: We do not use environment variables to configure the code paths! This needs to be a configuration option.

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -120,6 +132,16 @@ def __init__(
         if getattr(self.config, "mla_down_proj_fusion", False):
             submodules = self._fuse_mla_down_proj(submodules)
 
+        # Final #4531 rejects mHC on a plain TransformerLayer because it cannot
```
jaredcasper: Don't reference PRs in comments. I shouldn't have a look up a PR to understand what the code is doing.

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -120,6 +132,16 @@ def __init__(
...
+        # consume n-stream inputs directly. HybridStack's layer-boundary wrapper
+        # aggregates to a single stream before calling the inner layer, so build
+        # that inner layer with the mHC flag disabled and retain the original
+        # config on HyperConnectionHybridLayer itself.
```
jaredcasper: Why not just fix TransformerLayer to do the right thing when enable_mhc_connections = True?

megatron/core/models/hybrid/hybrid_block.py
```diff
@@ -134,7 +156,7 @@ def __init__(
                 if layer_type == LayerSymbols.MAMBA:
                     layer = build_module(
                         submodules.mamba_layer,
-                        config=self.config,
+                        config=inner_layer_config,
```
jaredcasper: This is confusing if I'm not doing mHC, it's not an inner layer there, yet it is using something called inner_layer_config.

megatron/core/fusions/fused_mhc_kernels.py
FDecaYed: Backend selection now comes from `TransformerConfig.mhc_fused_backend` and is threaded through each fused mHC op. The environment-variable switches are removed. `auto` keeps per-op selection; explicit backends fail if unavailable and never switch to another accelerated backend.

megatron/core/models/hybrid/hybrid_block.py
FDecaYed: Removed the PR-specific comment together with the config-copy workaround. This is part of the same ownership fix explained in the related `TransformerLayer` thread.

megatron/core/models/hybrid/hybrid_block.py
FDecaYed: The check was at the wrong ownership boundary. `TransformerBlock` directly feeds n-stream state and should require an mHC-capable layer, while `HybridStack` applies one `HyperConnectionHybridLayer` per symbol and passes single-stream state to the inner layer. The check is now in `TransformerBlock`, and the copied config is removed. This does not add GPTModel wiring; `HyperConnectionTransformerLayer` is existing generic mHC infrastructure.

megatron/core/models/hybrid/hybrid_block.py
FDecaYed: Removed `inner_layer_config` and the copied config as part of the same ownership fix described in the related `TransformerLayer` thread. `HybridStack` now always passes `self.config` and only adds the wrapper when mHC is enabled.

jaredcasper (approved):

PR #6403, opened 2026-08-10
Add HybridModel hash MoE routing

FDecaYed:
# Add HybridModel hash MoE routing

HybridModel can route the first configured MoE layers through a checkpointed token-to-expert table while retaining learned gate weights. Token IDs follow the existing model/layer/MLP path through recompute and partial CUDA-graph execution. Later MoE layers and MTP continue to use learned routing.

This PR retains its existing Hybrid mHC norm-recompute, MTP metric-depth and padding-mask integration fixes. Review follow-ups correct score normalization, preserve RouterReplay, update shared ShortcutMoE callers, and reject unsupported configurations. The vocabulary is resolved once after tokenizer initialization; generic config helpers and model entry points do not take a separate tokenizer-vocabulary argument.

The scope remains Hybrid Hash routing and its existing execution contracts. It does not add a new mHC implementation, GPTModel integration, CP/THD scheduling, Variable-token HybridEP, new inference backends, recipes, or complete Flextron/KD/VLM support. The round-robin table remains a placeholder without a load-balance or performance claim.

Validation: source-contract CPU checks cover route outputs/gradients against public score semantics, real RouterReplay actions, startup normalization, reshape callers, mHC wrapper and SP/PP mask handling. Formatting/static checks and the kernel determinism coverage gate pass. GPU/TE/distributed execution and exact-head CI remain pending; standard pytest collection on the local host is blocked by the missing Triton dependency.


megatron/core/models/hybrid/hybrid_model.py
```diff
@@ -492,6 +492,36 @@ def forward(
             # decoder will get hidden_states from encoder.input_tensor
             decoder_input = None
 
+        hash_input_ids = input_ids
```
Phlip79: Can you add a comment here explaining what this code is used for?

megatron/core/models/hybrid/hybrid_model.py
```diff
@@ -527,6 +557,10 @@ def forward(
         #   be None, so this assert will succeed.
         # assert attention_mask is None, "The attention mask is ignored and should be set to None"
 
+        decoder_extra_block_kwargs = {}
+        if self.config.moe_n_hash_layers > 0 and hash_input_ids is not None:
+            decoder_extra_block_kwargs['input_ids'] = hash_input_ids
+
```
Phlip79: Can we just do this?

```suggestion
        input_ids = None
        if self.config.moe_n_hash_layers > 0 and hash_input_ids is not None:
            input_ids = hash_input_ids

```

megatron/core/models/hybrid/hybrid_model.py
FDecaYed: Added a comment explaining the hash-routing input and sequence-parallel sharding.

megatron/core/models/hybrid/hybrid_model.py
FDecaYed: Removed decoder_extra_block_kwargs and now pass the IDs directly. Kept a separate hash_input_ids because it may be sequence-sharded, while the original input_ids are still used by MTP/loss.

Phlip79 (approved):

megatron/core/transformer/transformer_config.py
```diff
@@ -866,6 +867,15 @@ class TransformerConfig(ModelParallelConfig):
     If negative, generates bias once per layer and reuses it (abs value is std).
     This is an experimental feature for benchmarking purposes."""
 
+    moe_n_hash_layers: int = 0
+    """Number of leading MoE layers that use hash-based routing.
+    In HybridModel this counts MoE positions in the layer pattern rather than
+    all hybrid symbols. Other transformer stacks use the layer number directly."""
+
+    actual_vocab_size: Optional[int] = None
```
jaredcasper: This is a confusing name. Why would you need an "actual" vocab size over "vocab size". Either the name needs to be more descriptive or the docstring needs more explanation.

megatron/core/transformer/transformer_layer.py
```diff
@@ -457,12 +458,15 @@ def __init__(
...
                 "`as_mlp_submodule` classmethod instead.",
             )
-        self.mlp = submodules.mlp(
+        mlp_kwargs = dict(
             config=self.config,
             pg_collection=pg_collection,
             is_mtp_layer=self.is_mtp_layer,
             name=(name + ".mlp") if name is not None else None,
         )
+        if hash_moe_layer_threshold is not None:
```
jaredcasper: What if hash_moe_layer_threshold is not None, but this is just a normal mlp layer without this arg?

megatron/core/transformer/transformer_layer.py
```diff
@@ -687,6 +698,7 @@ def _forward_attention(
         packed_seq_params: Optional[PackedSeqParams] = None,
         sequence_len_offset: Optional[Tensor] = None,
         padding_mask: Optional[Tensor] = None,
+        input_ids: Optional[Tensor] = None,
```
jaredcasper, line 708: Why does the attention need a separate input_ids that are forwarded to the moe layers? This smells weird.

megatron/training/argument_utils.py
```diff
@@ -306,6 +306,7 @@ def core_transformer_config_from_args(args, config_class=None):
     kw_args['pipeline_dtype'] = args.params_dtype
     kw_args['batch_p2p_comm'] = not args.overlap_p2p_comm
     kw_args['num_moe_experts'] = args.num_experts
+    kw_args['actual_vocab_size'] = args.padded_vocab_size
```
zhongbozhu: what is `actual_vocab_size`?

megatron/core/transformer/transformer_config.py
```diff
@@ -866,6 +867,15 @@ class TransformerConfig(ModelParallelConfig):
     If negative, generates bias once per layer and reuses it (abs value is std).
     This is an experimental feature for benchmarking purposes."""
 
+    moe_n_hash_layers: int = 0
```
zhongbozhu, line 881: Is deepseekv4 the only model using this?

megatron/core/transformer/transformer_config.py
zhongbozhu: agreed that this name is quite confusing

FDecaYed: Renamed this to hash_moe_vocab_size and made it TP-independent by deriving it from the tokenizer vocab. Added a TP1-to-TP8 distributed-checkpoint load test for tid2eid.

megatron/core/transformer/transformer_layer.py
FDecaYed: Fixed. hash_moe_layer_threshold is now passed only when the MLP builder constructs a MoELayer.

megatron/core/transformer/transformer_layer.py
FDecaYed: It is not consumed by attention. TransformerLayer keeps it in the shared forward signature because attention and MLP are split for CUDA graphs; only _forward_mlp passes it to the hash router. Clarified this in the docstring.

megatron/training/argument_utils.py
FDecaYed: Changed this to use the TP-independent tokenizer vocab instead of padded_vocab_size; the config field is now hash_moe_vocab_size.

megatron/core/transformer/transformer_config.py
FDecaYed: Currently HybridModel-only, with DSv4 as the first user. GPTModel cannot forward token IDs, so it now rejects moe_n_hash_layers > 0 at construction.

dimapihtar (approved):

guihong-nv (approved):

maanug-nv (approved):

zhongbozhu (approved):
LGTM. Nit: We might need to ban Hash routing X RouterReplay

megatron/core/distributed/finalize_model_grads.py
```diff
@@ -319,7 +319,11 @@ def reset_model_temporary_tensors(config: TransformerConfig, model: List[torch.n
     """
     for model_chunk in model:
         for module in get_attr_wrapped_model(model_chunk, 'modules')():
-            if config.moe_router_enable_expert_bias and hasattr(module, 'expert_bias'):
+            if (
+                config.moe_router_enable_expert_bias
+                and hasattr(module, 'expert_bias')
+                and module.expert_bias is not None
+            ):
```
yashaswikarnati, line 326: curios why we need this now 
module.expert_bias is not None

FDecaYed: Hash layers disable expert bias even when it is enabled globally, leaving both buffers as `None`. This guard skips them during reset. [`test_hash_router_without_expert_bias_is_ignored`](https://github.com/NVIDIA/Megatron-LM/blob/d8ddd86cff412243f0c6f96ef48571f29aad9ce1/tests/unit_tests/distributed/test_finalize_model_grads.py#L58) covers this case.

fanshiqing (approved):

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -146,7 +151,13 @@ class RouterBuilder(Protocol):
...
 
     def __call__(
-        self, /, *, config: TransformerConfig, pg_collection: ProcessGroupCollection | None
+        self,
+        /,
+        *,
+        config: TransformerConfig,
+        pg_collection: ProcessGroupCollection | None,
+        is_mtp_layer: bool = False,
+        hash_moe_layer_threshold: int | None = None,
```
yashaswikarnati, line 160: instead of hash_moe_layer_threshold, just a suggestion may be and have the Hybrid model builder pass the flag like  use_hash_routing  might be cleaner

megatron/core/transformer/moe/router.py
```diff
@@ -263,6 +274,32 @@ def __init__(
         if self.config.moe_enable_routing_replay:
             self.router_replay = RouterReplay()
 
+    def set_layer_number(self, layer_number: int):
```
yashaswikarnati, line 277: seems this function is doing more than just set layer number.

megatron/core/transformer/moe/router.py
```diff
@@ -263,6 +274,32 @@ def __init__(
...
+        hash_moe_layer_threshold = self.hash_moe_layer_threshold
+        if hash_moe_layer_threshold is None:
+            hash_moe_layer_threshold = self.config.moe_n_hash_layers
+        self.is_hash_layer = (
+            not self.is_mtp_layer
+            and hash_moe_layer_threshold > 0
+            and layer_number <= hash_moe_layer_threshold
+        )
+        if not self.is_hash_layer:
+            return
```
yashaswikarnati, line 289: to  have clear separations, may be we shjould have init hash routing function and not do this in set layer number. then seems like we dont have to override 

```
if self.is_hash_layer:
    self._initialize_hash_routing()
```

megatron/core/transformer/moe/router.py
FDecaYed: Extracted the hash-only state setup into `_initialize_hash_routing()` in 7633788d31. `set_layer_number()` now assigns the layer number, determines hash eligibility, and delegates initialization.

megatron/core/transformer/moe/router.py
FDecaYed: Implemented the helper in 7633788d31: `tid2eid` creation and expert-bias cleanup now live in `_initialize_hash_routing()`. The `set_layer_number()` override retains eligibility selection; existing mappings, MTP exclusion, and replay layer-number handling are preserved.

megatron/core/tokenizers/utils/build_tokenizer.py
```diff
@@ -146,7 +144,8 @@ def vocab_size_with_padding(orig_vocab_size, args, logging_enabled=True):
     return after
 
 
-def _set_padded_vocab_size(args, tokenizer):
-    """Sets padded vocab size if None."""
-    if getattr(args, "padded_vocab_size", None) is None:
+def _set_vocab_sizes(args, tokenizer):
+    """Record the tokenizer vocabulary and set the padded model vocabulary if needed."""
+    args.tokenizer_vocab_size = tokenizer.vocab_size
```
jaredcasper: We should be moving away from using a global args... why are we adding state to it here?

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -258,9 +270,16 @@ def __init__(
...
-        self.router = self.submodules.router(
-            config=self.config, pg_collection=pg_collection, is_mtp_layer=is_mtp_layer
-        )
+        router_kwargs = {
+            "config": self.config,
+            "pg_collection": pg_collection,
+            "is_mtp_layer": is_mtp_layer,
+        }
+        if hash_moe_layer_threshold is not None:
+            router_kwargs["hash_moe_layer_threshold"] = hash_moe_layer_threshold
```
jaredcasper: Why is this optionally added, don't all routers now take this? If it's None just pass it in?

megatron/core/models/hybrid/layers/hybrid_hyper_connection.py
```diff
@@ -69,6 +71,9 @@ def _call_inner_layer(
                 packed_seq_params=packed_seq_params,
                 padding_mask=padding_mask,
             )
+            if input_ids is not None:
+                layer_kwargs["input_ids"] = input_ids
```
jaredcasper: Do not all inner layers take input_ids? Why are we optionally adding this if it's not None instead of just passing None?

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -463,13 +482,23 @@ def setup_delayed_wgrad_for_dispatch_backward_overlap(self):
...
+    ):
         """Compute token routing for preprocessing.
 
         This method uses the router to determine which experts to send each token to,
         producing routing probabilities and a mapping.
         """
-        probs, routing_map = apply_module(self.router)(hidden_states, padding_mask)
+        if input_ids is None:
+            probs, routing_map = apply_module(self.router)(hidden_states, padding_mask)
+        else:
```
jaredcasper: Again here, if input_ids is None what does it hurt to just always pass it in?

megatron/core/transformer/transformer_config.py
```diff
@@ -876,6 +877,15 @@ class TransformerConfig(ModelParallelConfig):
...
```
jaredcasper: no need to shorten "num", how about `moe_num_hash_layers` or `hash_moe_num_layers` to make a common prefix?

megatron/core/transformer/transformer_layer.py
```diff
@@ -443,6 +444,13 @@ def __init__(
         from megatron.core.extensions.transformer_engine import TEFusedMLP
         from megatron.core.transformer.moe.moe_layer import MoELayer
 
+        mlp_builder = (
+            submodules.mlp.module if isinstance(submodules.mlp, ModuleSpec) else submodules.mlp
+        )
+        while isinstance(mlp_builder, functools.partial):
```
jaredcasper: This is ugly and error prone, is there a better way to do this? An attribute on the builder that says if its moe or not or something?

At the very least we should just make mlp take in the argument and error out so all "MLPs" take the same argument, rather than changing the arguments based on this inspection.

megatron/core/tokenizers/utils/build_tokenizer.py
FDecaYed: Removed the write to `args.tokenizer_vocab_size` and now pass the actual vocabulary size explicitly when constructing `config.hash_moe_vocab_size`.

megatron/core/transformer/moe/moe_layer.py
FDecaYed: Removed the conditional kwargs construction and now always forward `hash_moe_layer_threshold` unchanged, including `None`.

megatron/core/models/hybrid/layers/hybrid_hyper_connection.py
FDecaYed: The `TransformerLayer` branch now always forwards `input_ids`, including `None`; the Mamba branch omits it because its forward does not accept the argument.

megatron/core/transformer/moe/moe_layer.py
FDecaYed: Simplified this to a single router call that always passes `input_ids=input_ids`.

megatron/core/transformer/transformer_config.py
FDecaYed: Renamed it to `moe_num_hash_layers`, with the corresponding CLI flag `--moe-num-hash-layers`.

megatron/core/transformer/transformer_layer.py
FDecaYed: Unified the MLP builder interface to always forward `hash_moe_layer_threshold`, removing type checks and wrapper inspection. Dense MLP and TE fused MLP accept `None`/`0` and explicitly reject positive values.

ChenhanYu (approved):

YangFei1990 (approved):

megatron/core/transformer/transformer_layer.py
```diff
@@ -991,7 +1022,7 @@ def _forward_pre_mlp_layernorm(self, hidden_states: Tensor):
 
         return pre_mlp_layernorm_output
 
-    def _maybe_unflatten_for_moe(self, hidden_states, padding_mask, packed_seq_params):
+    def _maybe_unflatten_for_moe(self, hidden_states, padding_mask, input_ids, packed_seq_params):
```
tdene, line 1024: The inputs/outputs of `_maybe_unflatten_for_moe` have changed, but the callsites of `ShortcutMoE` inside `megatron/core/models/hybrid/shortcut_block.py` have not been updated, so anyone using those classes will experience a `TypeError`.

Note that this `ShortcutMoe` got merged while this PR was being worked on, in #6959

megatron/core/transformer/moe/router.py
```diff
@@ -746,17 +791,83 @@ def _apply_expert_bias(
...
+            raise ValueError(f"Invalid score_function: {self.score_function}")
+
+        # Hidden states are flattened from [sequence, batch, hidden], whereas
+        # model token IDs arrive as [batch, sequence].
+        flat_ids = input_ids.T.reshape(-1)
+        assert flat_ids.numel() == logits.shape[0], (
+            f"input_ids contains {flat_ids.numel()} tokens, but router logits contain "
+            f"{logits.shape[0]}."
+        )
+        default_top_indices = self.tid2eid[flat_ids].long()
```
tdene: `tid2eid` is defined as having `hash_moe_vocab_size` number of rows.

`flat_ids` has `padded_vocab_size` number of rows. `padded_vocab_size` > `hash_moe_vocab_size`, so anything in-between will cause a CUDA error.

My suggested solution here is to just clamp with something like `default_top_indices = self.tid2eid[flat_ids.clamp(0, self.tid2eid.shape[0] - 1)].long()`. You can probably write it cleaner than I did there.

There should also be a unit test that can catch this kind of failure.

megatron/core/transformer/moe/router.py
```diff
@@ -746,17 +791,83 @@ def _apply_expert_bias(
...
+            return scores.gather(1, default_top_indices), default_top_indices
+
+        if self.router_replay is not None:
+            probs, top_indices = self.router_replay.get_replay_topk(
+                scores, self.topk, default_compute_topk=_compute_hash_topk
+            )
+        else:
+            probs, top_indices = _compute_hash_topk(scores, self.topk)
+        if self.score_function != "softmax" and self.topk > 1:
+            probs = probs / (probs.sum(dim=-1, keepdim=True) + 1e-20)
```
tdene: Is this discrepancy intended?

With `moe_router_score_function="softmax"`, `moe_router_pre_softmax=False`, the two paths calculate weights differently.

The learned path does top-k over the logits, then softmax over the selected weights, causing the weight probs to sum to 1.

This hash path does softmax over ALL weights, then gathers the entries at the selected experts. Renormalization doesn't happen because it's gated by `score_function != "softmax"`. The weight probs don't sum to 1. Unlike with the training path, there is no `moe_router_pre_softmax` conditional here.

This seems like a bug?

megatron/elastification/pretrain_hybrid_flex.py
tdene, line 1: This reads like an attempt to add compatibility with flextron.

But flextron computes routing via `topk_softmax_with_capacity`, not `routing()`, so we don't get hash layer functionality while still claiming `is_hash_layer=True`.

Also flextron has a `router.local_tokens_per_expert += routing_map.sum(dim=0)` call that will cause a crash since `_initialize_hash_routing` sets `router.local_tokens_per_expert` to `None`.

megatron/training/argument_utils.py
```diff
@@ -274,7 +274,15 @@ def _get_field_docstrings(self, src_cfg_class: type) -> dict[str, str]:
         return field_docstrings
 
 
-def core_transformer_config_from_args(args, config_class=None):
+def core_transformer_config_from_args(
+    args, config_class=None, *, tokenizer_vocab_size: int | None = None
+):
+    """Build a transformer config with an explicitly supplied tokenizer vocabulary.
+
+    Hash routing needs the vocabulary including added tokens, before TP-dependent
```
jaredcasper: I don't like having a special case as argument to this function. It's not the place to handle special cases like this, especially since in all the places you call it you pass None if hash_moe_vocab_size is set... that makes this confusing and non-intuitive.

Can you just set args.hash_moe_vocab_size from the tokenizer after parsing args and making the tokenizer or something?

pretrain_hybrid.py
```diff
@@ -112,7 +113,12 @@ def get_batch(data_iterator, vp_stage=None):
     """Generate a batch."""
 
     args = get_args()
-    config = core_transformer_config_from_args(args)
+    config = core_transformer_config_from_args(
+        args,
+        tokenizer_vocab_size=(
+            get_tokenizer().vocab_size if getattr(args, 'moe_num_hash_layers', 0) else None
```
jaredcasper: Having this at every call site is ugly, and is confusing because just looking at it from the call site, why would I only set tokenizer_vocab_size if moe_num_hash_layers is set, that seems very odd.

megatron/core/tokenizers/utils/build_tokenizer.py
jaredcasper: I appreciate doing this explicitly in the config, but the way that it is handled in the args -> config version needs improvement.

megatron/core/models/hybrid/hybrid_model.py
```diff
@@ -509,6 +566,39 @@ def forward(
...
+            and decoder_input is not None
+            and padding_mask.shape[1] != decoder_input.shape[0]
+        ):
+            padding_mask = (
+                tensor_parallel.scatter_to_sequence_parallel_region(
+                    padding_mask.transpose(0, 1).contiguous(), group=self.pg_collection.tp
+                )
+                .transpose(0, 1)
+                .contiguous()
+            )
```
tdene, line 640: This works for PP=1.

For SP + PP>1, on all stages except the first one, we come into this comparison with `decoder_input == None`, but `padding_mask != None` and needs to be scattered.

I'm pretty sure the fix is small; instead of `and decoder_input is not None and padding_mask.shape[1] != decoder_input.shape[0]`, it would be `ref_tensor = decoder_input if decoder_input is not None else self.decoder.input_tensor` before the if-statement, and then `and ref_tensor is not None and and padding_mask.shape[1] != ref_tensor.shape[0]` inside the if-statement.

Also, this same logic is broken in `gpt_model.py`. It is out-of-scope for this PR to fix it there, so there's no requirement for you to fix it there as well. But could you please at least add a `TODO` in the code or a GitHub issue as a reminder that PP + SP + TP + MoE is broken?

megatron/post_training/model_builder.py
```diff
@@ -496,7 +501,20 @@ def modelopt_gpt_hybrid_builder(
             ), "ModelOpt Distillation currently incompatible with interleaved pipeline schedule."
 
         teacher_config_raw = _load_teacher_model_config(args.export_kd_teacher_load)
-        teacher_config = core_transformer_config_from_args(teacher_config_raw)  # convert to TransformerConfig
+        # The teacher owns its vocabulary metadata; it need not match the student's tokenizer.
+        teacher_vocab_size = getattr(teacher_config_raw, 'hash_moe_vocab_size', None)
+        if teacher_vocab_size is None:
+            # Compatibility with checkpoints saved before tokenizer metadata was passed explicitly.
+            teacher_vocab_size = getattr(teacher_config_raw, 'tokenizer_vocab_size', None)
```
tdene: I think part of this is outdated code: `tokenizer_vocab_size` was added in this PR and got removed in 12a34a8276, so there is no need to have compatibility with it.

megatron/post_training/model_builder.py
```diff
@@ -496,7 +501,20 @@ def modelopt_gpt_hybrid_builder(
...
+        if teacher_vocab_size is None:
+            teacher_vocab_size = teacher_config_raw.vocab_size
+        if teacher_vocab_size is None:
+            # Without teacher-specific metadata, _load_teacher_model_config defaults to
+            # the student's architecture. Reuse its already resolved vocabulary as well.
+            teacher_vocab_size = config.hash_moe_vocab_size
```
tdene: I think the order on these is wrong. `teacher_config_raw.vocab_size` is the unpadded vocab size from the student's args; `config.hash_moe_vocab_size` is the correct vocab size, so it should take precedence.

Combined with the other comment here, I would expect this block to be
```
teacher_vocab_size = getattr(teacher_config_raw, 'hash_moe_vocab_size', None)
if teacher_vocab_size is None:
    teacher_vocab_size = config.hash_moe_vocab_size
if teacher_vocab_size is None:
    teacher_vocab_size = teacher_config_raw.vocab_size
if teacher_vocab_size is None and getattr(teacher_config_raw, 'moe_num_hash_layers', 0):
    teacher_vocab_size = get_tokenizer().vocab_size
```

megatron/core/transformer/transformer_config.py
```diff
@@ -2971,6 +2981,41 @@ def __post_init__(self):
...
             )
 
+        if self.moe_num_hash_layers > 0:
+            if self.moe_shortcut_connection:
+                raise ValueError(
+                    "ShortcutMoE does not yet forward the token IDs required for hash MoE routing."
+                )
+            assert (
+                self.hash_moe_vocab_size is not None and self.hash_moe_vocab_size > 0
+            ), "hash_moe_vocab_size must be positive when moe_num_hash_layers > 0."
```
tdene, line 3018: I can't find a better place to attach this comment: `pretrain_vlm.py` also needs to be updated, because some VLMs use hybrid models.

megatron/core/transformer/transformer_config.py
```diff
@@ -2971,6 +2981,41 @@ def __post_init__(self):
...
+            assert (
+                not self.overlap_moe_expert_parallel_comm
+            ), "overlap_moe_expert_parallel_comm does not support hash MoE layers yet."
+            log_single_rank(
+                logger,
+                logging.WARNING,
+                "Hash MoE initialized with a placeholder round-robin token-to-expert table. "
+                "Load a trained table from a checkpoint or provide a workload-aware "
+                "initialization before training.",
+            )
```
tdene: This warning will fire on every microbatch, spamming the logs.

Could you move it inside `_initialize_hash_routing`, inside the `if self.tid2eid is None:` block, instead?

megatron/core/transformer/moe/router.py
```diff
@@ -265,6 +276,34 @@ def __init__(
...
+            and hash_moe_layer_threshold > 0
+            and layer_number <= hash_moe_layer_threshold
+        )
+        if self.is_hash_layer:
+            self._initialize_hash_routing()
+
+    def _initialize_hash_routing(self) -> None:
+        """Initialize the hash lookup table and disable learned-routing expert bias."""
+        if self.tid2eid is None:
+            token_ids = torch.arange(self.config.hash_moe_vocab_size, device=self.weight.device)
```
tdene, line 317: `self.weight.device` is CPU. Can we build this directly on GPU instead with `device=torch.cuda.current_device()`?

megatron/core/transformer/moe/router.py
```diff
@@ -265,6 +276,34 @@ def __init__(
         if self.config.moe_enable_routing_replay:
             self.router_replay = RouterReplay()
 
+    def set_layer_number(self, layer_number: int):
+        """Set the layer number and initialize hash routing for eligible layers."""
+        super().set_layer_number(layer_number)
+        hash_moe_layer_threshold = self.hash_moe_layer_threshold
+        if hash_moe_layer_threshold is None:
+            hash_moe_layer_threshold = self.config.moe_num_hash_layers
```
tdene, line 298: `hash_moe_layer_threshold` is a layer number. `moe_num_hash_layers` is a layer count. For a hybrid model, those are not equivalent.

This should not be a problem today, as far as I can tell, because `HybridModel` will always set `hash_moe_layer_threshold`. But anything that doesn't go through that code in `HybridModel`, like some code that goes only through `HybridStack`, would break here.

I can't fully trace out the code-paths to be able to tell how relevant this danger is, but it does exist.

megatron/core/transformer/moe/router.py
```diff
@@ -265,6 +276,34 @@ def __init__(
...
+            self._initialize_hash_routing()
+
+    def _initialize_hash_routing(self) -> None:
+        """Initialize the hash lookup table and disable learned-routing expert bias."""
+        if self.tid2eid is None:
+            token_ids = torch.arange(self.config.hash_moe_vocab_size, device=self.weight.device)
+            expert_offsets = torch.arange(self.topk, device=token_ids.device)
+            self.tid2eid = ((token_ids[:, None] + expert_offsets) % self.num_experts).to(
+                torch.int32
+            )
```
tdene, line 321: I can't fully trace out this code, but it seems like this hurts expert load balancing. If `num_local_experts > topk`, for example, are we not sending the entire top-k window to the same rank (two ranks if it straddles a boundary). The way I read this, this doesn't balance experts well when we have multiple expert copies?

In particular, padding tokens (more common with sequence padding) will always have `token_id == 0`, so all padding will always be routed to rank 0. That sounds bad.

megatron/core/transformer/moe/router.py
```diff
@@ -806,8 +928,15 @@ def routing(self, logits: torch.Tensor, padding_mask: Optional[torch.Tensor] = N
                 pad_to_capacity=self.config.moe_pad_expert_input_to_capacity,
             )
 
-        # Apply each aux loss type and attach aux loss autograd function to probs
-        if self.training and torch.is_grad_enabled() and self.is_aux_loss_enabled():
+        # Hash routing assignments come from tid2eid rather than the learned logits, so a
+        # learned top-k aux loss would optimize routing decisions that were never dispatched.
+        # Keep aux loss enabled for the non-hash MoE layers that share this configuration.
+        if (
+            not self.is_hash_layer
```
tdene: This causes a denominator mismatch in the metrics, right? `load_balancing_loss` is divided by the number of layers in the metrics, so this gate will make the denominator smaller when not all MoE layers hash.

The actual training will be fine, but the metrics will report that with MoE hashing turned on the `load_balancing_loss` suddenly changes (because the denominator changed).

megatron/core/transformer/moe/router.py
```diff
@@ -265,6 +276,34 @@ def __init__(
...
+            hash_moe_layer_threshold = self.config.moe_num_hash_layers
+        self.is_hash_layer = (
+            not self.is_mtp_layer
+            and hash_moe_layer_threshold > 0
+            and layer_number <= hash_moe_layer_threshold
+        )
+        if self.is_hash_layer:
+            self._initialize_hash_routing()
+
+    def _initialize_hash_routing(self) -> None:
```
tdene, line 307: Nothing ever asserts `topk <= num_experts`. If that's the case, the logic in this method breaks down. Can we put an `assert self.topk <= self.num_experts` here?

megatron/core/models/hybrid/layers/hybrid_hyper_connection.py
```diff
@@ -58,17 +58,21 @@ def _call_inner_layer(
         packed_seq_params: Optional[PackedSeqParams],
         packed_sequence_cp_metadata: Optional[PackedSequenceCPMetadata],
         padding_mask: Optional[Tensor],
+        input_ids: Optional[Tensor] = None,
+        mhc_recompute_manager=None,
```
tdene: I don't think this is ever used, nor ever provided

megatron/core/transformer/transformer_layer.py
```diff
@@ -460,12 +462,14 @@ def __init__(
...
             )
-        self.mlp = submodules.mlp(
+        mlp_kwargs = dict(
             config=self.config,
             pg_collection=pg_collection,
             is_mtp_layer=self.is_mtp_layer,
             name=(name + ".mlp") if name is not None else None,
+            hash_moe_layer_threshold=hash_moe_layer_threshold,
         )
+        self.mlp = submodules.mlp(**mlp_kwargs)
```
tdene: Since we're doing a `**dict(kwargs)` here, should we just inline it and eliminate the `mlp_kwargs` helper variable?

megatron/core/transformer/moe/moe_layer.py
```diff
@@ -126,7 +126,12 @@ def __call__(
 class RouterInterface(Protocol):
     """Interface for the router used in an MoELayer."""
 
-    def forward(self, input: torch.Tensor, /) -> tuple[torch.Tensor, torch.Tensor]:
+    def forward(
```
tdene, line 129: Can you bring the `/` back?

megatron/core/transformer/transformer_layer.py
```diff
@@ -474,6 +478,10 @@ def __init__(
 
         self.is_moe_layer = isinstance(self.mlp, MoELayer)
 
+        # Split Hybrid mHC runs these base-layer norms through the wrapper's recompute manager.
+        self.mhc_checkpoint_input_layernorm = not isinstance(self.input_layernorm, IdentityOp)
+        self.mhc_checkpoint_pre_mlp_layernorm = not isinstance(self.pre_mlp_layernorm, IdentityOp)
```
tdene, line 482: I don't have a better place to anchor this comment: these two flags are now computed in TransformerLayer's init, so the code in HyperConnectionTransformerLayer's init that also computes them is redundant.

megatron/core/transformer/transformer_layer.py
FDecaYed: All MLP builders now accept the same optional `hash_moe_layer_threshold`. Dense MLP and TE fused MLP accept `None`/`0` and reject positive values; builder inspection is removed.

megatron/core/tokenizers/utils/build_tokenizer.py
FDecaYed: The Core tokenizer utility now matches main. `hash_moe_vocab_size` is initialized once in the training bootstrap after tokenizer creation, preserving explicit values; config helpers no longer take a special vocabulary argument.

megatron/training/argument_utils.py
FDecaYed: Done. Set `args.hash_moe_vocab_size` once after tokenizer initialization, preserving explicit values, and removed the extra vocabulary argument from the config helpers.

pretrain_hybrid.py
FDecaYed: Removed these repeated call-site branches. The callers now use the normalized config field directly.

megatron/core/transformer/transformer_layer.py
FDecaYed: Updated both ShortcutMoE calls for the new arguments and return values. The existing packed/unpacked tests now use the production reshape helper.

megatron/core/transformer/moe/router.py
FDecaYed: `flat_ids` contains one entry per token, not per padded vocabulary row. Valid token IDs use the tokenizer vocabulary; TP padding does not create new token IDs. Changed lookup to `index_select` so negative and out-of-range IDs are rejected rather than clamped to another token’s experts. Both invalid-ID cases were checked on CPU.

megatron/core/transformer/moe/router.py
FDecaYed: Fixed. Hash routing now honors `moe_router_pre_softmax`: post-softmax normalizes the selected logits. Intermediate score calculations stay in FP32, and the existing tests compare outputs and gradients against the public routing function while retaining RouterReplay.

megatron/elastification/pretrain_hybrid_flex.py
FDecaYed: Added an explicit rejection before Flextron installs the router hooks. This PR does not add Flextron hash-routing support.

megatron/core/models/hybrid/hybrid_model.py
FDecaYed: Fixed using `decoder_input` or `self.decoder.input_tensor` as the reference. Extended the existing SP test to cover later PP stages, and left a TODO for the corresponding GPTModel fix.

megatron/post_training/model_builder.py
FDecaYed: Removed the obsolete compatibility block. The ModelOpt builder now matches main.

megatron/post_training/model_builder.py
FDecaYed: Removed this special teacher-vocabulary block. The shared startup path resolves `hash_moe_vocab_size` before config conversion; teacher config keeps its existing normalized-args inheritance and explicit overrides.

megatron/core/transformer/transformer_config.py
FDecaYed: Vocabulary propagation no longer needs an entry-point-specific argument: it is resolved in the shared tokenizer startup path. Full VLM hash execution has not been validated in this PR.

megatron/core/transformer/transformer_config.py
FDecaYed: Moved the warning into `_initialize_hash_routing`, under `if self.tid2eid is None:`.

megatron/core/transformer/moe/router.py
FDecaYed: Kept the table on `self.weight.device` to preserve CPU initialization. It is a registered buffer and moves with the router through `.cuda()`/`.to()`.

megatron/core/transformer/moe/router.py
FDecaYed: Added a fail-fast check: a Hybrid hash router must receive an explicit global layer threshold. It no longer falls back to the MoE layer count; explicit `0` remains valid and MTP stays on learned routing.

megatron/core/transformer/moe/router.py
FDecaYed: Agreed, the placeholder round-robin table does not guarantee balanced EP traffic, especially for repeated or padding IDs. Kept the initialization warning and checkpoint-loaded table path; changing the mapping policy needs separate workload validation.

megatron/core/transformer/moe/router.py
FDecaYed: I checked the training caller and metrics aggregation: the denominator remains the total configured MoE layer count, including hash layers. Hash layers contribute no load-balancing loss, but do not reduce that denominator. Left the aggregation unchanged.

megatron/core/transformer/moe/router.py
FDecaYed: Added config validation requiring `1 <= moe_router_topk <= num_moe_experts` before table initialization.

megatron/core/models/hybrid/layers/hybrid_hyper_connection.py
FDecaYed: Removed the unused `mhc_recompute_manager` argument from `_call_inner_layer`. The manager remains on the raw branch where it is used.

megatron/core/transformer/transformer_layer.py
FDecaYed: Inlined the builder call and removed `mlp_kwargs`.

megatron/core/transformer/moe/moe_layer.py
FDecaYed: Restored the positional-only `/`.

megatron/core/transformer/transformer_layer.py
FDecaYed: Removed the duplicate assignments from `HyperConnectionTransformerLayer`; the base class initializes both flags.

megatron/core/transformer/moe/router.py
tdene: Sorry, I got confused and misdiagnosed.

The problem I was looking at was not with the row count, it's about the tokens being generated living in the padded-vocab region. I was specifically worried about inference outputs making it back into the next forward pass.

After a conversation with the inference team, I've been reminded this risk has been analyzed before and is not relevant.

Resolving!

megatron/core/transformer/moe/router.py
tdene: Fair!

megatron/core/transformer/moe/router.py
tdene: My bad; what I meant to say is that the numerator changes, so we need to change the denominator to match it.

The denominator is always `num_moe_layers`, while the numerator excludes hashed layers. So to get an accurate metric we should set the denominator `num_moe_layers - num_hashed_layers`, because that's the number of contributions to the numerator we have.

The only good way I see of fixing this is to pass `num_hash_layers` into `megatron/core/transformer/moe/moe_logging.py` (next to where `num_hash_layers` is passed in), and for the metrics `["load_balancing_loss", "seq_load_balancing_loss", "global_load_balancing_loss"]` to divide by `num_moe_layers - num_hash_layers`.

tdene: Approving, but I still think this needs to be fixed.

tdene (approved):

jaredcasper (approved):

PR #6402, opened 2026-08-10
Add native DeepSeek-V4 hybrid attention orchestration

FDecaYed:
# Add native DeepSeek-V4 hybrid attention orchestration

## Summary

Wire the D1 SBHD CSA primitives into native DeepSeek-V4 attention orchestration, module specs, grouped output, MTP propagation, and indexer-loss scheduling/logging.

## Scope

- Add `DSv4HybridAttention` and `DSv4HybridSelfAttention` SBHD construction and execution.
- Add the DSv4 module-spec builder without depending on later HybridModel construction. The existing `models/gpt` filename is a shared experimental-attention factory and does not add GPTModel construction support.
- Propagate attention/MTP layer identity needed by the indexer-loss tracker.
- Use explicit `ProcessGroupCollection` inputs for cross-pipeline logging behavior.
- Add per-layer CSA ratio parsing/validation and grouped output projection configuration.
- Require `dsa_kernel_backend='none'` for this native slice; D3 adds the fused backend.
- Rebase the required `AbsorbedMLA` integration on the current-main implementation instead of replacing it with a dev copy.

## Non-goals

- Concrete fused CSA kernels; D3 owns them.
- HybridModel symbol allocation and construction; D4 owns them.
- Packed THD, context parallelism, CUDA-graph execution, or recipes.
- GPTModel construction support.
- Changes to ordinary DSA semantics that are not architecture-neutral and independently necessary.

## Review boundary and dependencies

Requires [#6400](https://github.com/NVIDIA/Megatron-LM/pull/6400) (D1) and P6/#5944. The local branch currently carries the complete D1 prerequisite line through `c463cbed9919`; this will still be an independent PR targeting `main`.

- Own commit: `0ecb394a67d2ae2b89c20d8d489c48fa5b5b1ef9`
- Exact own diff: **14 files, +1,650/-28**
- Temporary carried commit boundary: `c463cbed9919bcc662fd5b4282546367f1852c32`

Please review the own commit only. After D1 and #5944 merge, it will be replayed directly onto then-current `main`.

## Provenance

[...]

PR #6020, opened 2026-07-24
[Dev] Migrate main-first DSA support to dev

buptzyb:
## Summary

Bring the ordinary DS3.2/GLM DeepSeek Sparse Attention implementation from `main` into `dev` while preserving the existing DSv4/CSA/HCA feature set.

## Changes

- Keep the `main` DSA layout and APIs as the ordinary-DSA baseline, including the `none`, `tilelang`, and `cudnn` backend routing, cuDNN kernels, masking, layout handling, and native-parity behavior.
- Preserve the dev-only DSv4/CSA/HCA paths in CSA-specific modules, including fused and unfused kernels, C/H/W variants, packed/compressed THD, CP, MTP, mHC, CUDA Graph, FP8, HybridModel, and checkpoint compatibility.
- Reuse the unchanged `FusedDSAIndexerLoss` for both ordinary DSA and compressed-THD CSA by lowering CSA layout metadata at the call site. Remove the duplicate CSA loss implementation and ambiguous CSA kernel names.
- Keep `absorbed_mla` main-first while retaining dev combined/split K/V checkpoint loading, dynamic CP, MTP, and HybridModel behavior.
- Make `dsa_kernel_backend` the canonical DSA/CSA backend control. Retain `apply_dsa_kernel_fusion` and `--no-dsa-kernel-fusion` as deprecated DSv4 compatibility adapters.
- Preserve the latest dev deferred-weight-gradient contract. Indexer-owning layers flush all indexer linears, while IndexShare layers without local indexer weights safely no-op.

## File ownership and mapping

| Resulting file | Source of truth | Migration rule |
| --- | --- | --- |
| `dsa.py` | `main:dsa.py` | Keep the main implementation and API. Add only dev architecture-neutral integration for MTP layer numbering, packed-THD RoPE bounds, loss tracking/CUDA Graph behavior, and deferred-weight-gradient flushing. |
| `dsa_kernels.py` | `main:dsa_kernels.py` | Keep the main backend router exactly: `none`, `tilelang`, and `cudnn`. |
| `dsa_cudnn_kernels.py` | `main:dsa_cudnn_kernels.py` | Copy the main cuDNN implementation unchanged. |
| `dsa_indexer_loss.py` | `main:dsa_indexer_loss.py` | Copy the main indexer-loss helpers unchanged. |
| `dsa_layout.py` | `main:dsa_layout.py` | Keep main layout semantics unchanged; only the required copyright header differs. |
[...]

PR #6405, opened 2026-08-10
Add HybridModel DeepSeek-V4 layer construction

FDecaYed:
# Add HybridModel DeepSeek-V4 layer construction

## Summary

Add HybridModel symbols, specs, ratio handling, and construction for DeepSeek-V4 after the native attention and mHC runtime contracts are available.

## Scope

- Extend Hybrid layer allocation with C/H/W for compression ratios 4/128/0 while preserving the existing DSA `D` and MLA `+` symbols.
- Add CSA/HCA/window submodule slots and construct the corresponding D2 attention layers.
- Add the Transformer Engine `hybrid_dsv4_stack_spec` callable and callable-spec handling in training and builder entry points.
- Infer `dsv4_hybrid` and MLA use from C/H/W patterns.
- Normalize either compact C/H/W-only or legacy full-layer `csa_compress_ratios`, validating each ratio against its symbol.
- Add focused allocation, argument-normalization, stack-spec, and builder coverage.

Production diff: **6 files, +195/-20**. Test diff: **3 files, +153/-136**.

## Non-goals

- Reimplementing mHC or Hash routing; those remain P2 and P4.
- Concrete fused CSA kernels; D3 is an independent sibling and is not a dependency.
- Packed THD, context parallelism, CUDA-graph lifecycle, FLOP/accounting, functional recipes, or GPTModel support.

## Review boundary and dependencies

Requires [#6402](https://github.com/NVIDIA/Megatron-LM/pull/6402) (D2) and [#6401](https://github.com/NVIDIA/Megatron-LM/pull/6401) (P2); D3 is not a dependency. The local branch is built on D2 and temporarily carries:

- `993fe75090` — patch-equivalent to the current #4531 prerequisite;
- `2ebb42e43c` — patch-equivalent to P2;
- `a82126a5621b09cfda5abaa1c16b90640c10bedb` — this PR’s own change.

The exact own diff is **9 files, +348/-156**. Please review the last commit only. After D2 and P2 merge, the own commit will be replayed directly onto then-current `main`.

## Provenance

Reconstructed from current HybridModel APIs plus the post-[#6020](https://github.com/NVIDIA/Megatron-LM/pull/6020) DSv4 residual. [#5795](https://github.com/NVIDIA/Megatron-LM/pull/5795) is the historical behavior/test oracle, [#6372](https://github.com/NVIDIA/Megatron-LM/pull/6372) defines the current CSA namespace, and [#5960](https://github.com/NVIDIA/Megatron-LM/pull/5960) supplies attention-teacher correctness inherited through D1/D2.

## Validation

Passed locally:

- clean worktree and no dependency/lock-file changes;
- signed-commit verification and required `Signed-off-by` trailer;
- isort, Ruff, Python bytecode compilation, `git diff --check`, and focused API/import scans;
- focused AST/argument behavior checks.

Black’s whole-file checks for `megatron/training/argument_utils.py` and `megatron/training/arguments.py` retain the same baseline formatting failures; the D4 hunks introduce no new formatting regression. Runtime pytest is blocked on the bare macOS environment without the Linux CUDA stack. Focused GPU tests and full functional CI remain required after publication.

## Current functional constraint

C/H/W layers currently require explicitly selecting the callable `hybrid_dsv4_stack_spec` with `transformer_impl='transformer_engine'`. Omitting the DSv4 spec or selecting an inference-optimized/default stack can leave the new submodule slots as `IdentityOp`; those configurations are unsupported and must not be advertised. A later early-rejection check is optional and does not block this construction slice.

## Sibling rebase constraint

D4 can merge without [#6404](https://github.com/NVIDIA/Megatron-LM/pull/6404) (D3). Both touch `megatron/training/argument_utils.py` and `megatron/training/arguments.py`; if D3 merges first, D4’s rebase must retain D3’s omitted-versus-explicit backend resolution while adding the C/H/W inference and ratio normalization here.

## Local artifact

[D4-hybrid-dsv4-layer-construction.patch](../patches/D4-hybrid-dsv4-layer-construction.patch), generated from the own commit only. Verify it against [SHA256SUMS](../patches/SHA256SUMS).

## Before publication

After D2 and P2 merge, replay only D4’s own commit onto current `main`, preserve any already-merged D3 backend-default logic, confirm the final HybridModel/mHC call signatures and explicit-spec constraint, then rerun focused construction/forward tests and full functional CI.

### DIFF
--- a/megatron/core/models/hybrid/hybrid_layer_specs.py
+++ b/megatron/core/models/hybrid/hybrid_layer_specs.py
@@ -1,4 +1,5 @@
 # Copyright (c) 2023-2026, NVIDIA CORPORATION. All rights reserved.
+from dataclasses import replace
 from functools import partial
 
 from megatron.core.extensions.transformer_engine import (
@@ -30,6 +31,9 @@ from megatron.core.tensor_parallel import (
 )
 from megatron.core.transformer.attention import SelfAttention, SelfAttentionSubmodules
 from megatron.core.transformer.enums import AttnMaskType
+from megatron.core.transformer.experimental_attention_variant import (
+    deepseek_v4_hybrid_attention_module_specs as dsv4_module_specs,
+)
 from megatron.core.transformer.experimental_attention_variant.absorbed_mla import (
     AbsorbedMLASelfAttention,
     AbsorbedMLASelfAttentionSubmodules,
@@ -53,6 +57,7 @@ from megatron.core.transformer.multi_token_prediction import (
     MultiTokenPredictionLayerSubmodules,
 )
 from megatron.core.transformer.spec_utils import ModuleSpec
+from megatron.core.transformer.transformer_config import TransformerConfig
 from megatron.core.transformer.transformer_layer import (
     MoETransformerLayer,
     TransformerLayer,
@@ -427,3 +432,36 @@ mamba_stack_spec = hybrid_stack_spec
 mamba_inference_stack_spec = hybrid_inference_stack_spec
 gdp_stack_spec = gated_delta_product_stack_spec
 gdp_inference_stack_spec = gated_delta_product_inference_stack_spec
+
+
+def hybrid_dsv4_stack_spec(config: TransformerConfig) -> ModuleSpec:
+    """Build a HybridStack with fixed-ratio DSv4 C/H/W attention layers."""
+    assert (
+        config.transformer_impl == "transformer_engine"
+    ), "DSv4 HybridModel currently supports only the transformer-engine implementation."
+
+    from megatron.core.extensions.transformer_engine_spec_provider import TESpecProvider
+
+    dsv4_attention = dsv4_module_specs.get_dsv4_hybrid_module_spec_for_backend(
+        config=config, backend=TESpecProvider()
+    )
+
+    def wrap_dsv4_layer(compress_ratio: int) -> ModuleSpec:
+        attention = replace(
+            dsv4_attention,
+            params={**(dsv4_attention.params or {}), "compress_ratio": compress_ratio},
+        )
+        return ModuleSpec(
+            module=TransformerLayer,
+            submodules=TransformerLayerSubmodules(
+                input_layernorm=TENorm, self_attention=attention, self_attn_bda=get_bias_dropout_add
+            ),
+        )
+
+    submodules = replace(
+        hybrid_stack_spec.submodules,
+        csa_layer=wrap_dsv4_layer(compress_ratio=4),
+        hca_layer=wrap_dsv4_layer(compress_ratio=128),
+        window_layer=wrap_dsv4_layer(compress_ratio=0),
+    )
+    return replace(hybrid_stack_spec, submodules=submodules)
--- a/megatron/core/models/hybrid/hybrid_layer_specs.py
+++ b/megatron/core/models/hybrid/hybrid_layer_specs.py
@@ -435,7 +435,7 @@ gdp_inference_stack_spec = gated_delta_product_inference_stack_spec
 
 
 def hybrid_dsv4_stack_spec(config: TransformerConfig) -> ModuleSpec:
-    """Build a HybridStack with fixed-ratio DSv4 C/H/W attention layers."""
+    """Build a HybridStack with DSv4 C/H/W attention layers."""
     assert (
         config.transformer_impl == "transformer_engine"
     ), "DSv4 HybridModel currently supports only the transformer-engine implementation."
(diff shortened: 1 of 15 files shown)


megatron/core/models/hybrid/hybrid_layer_specs.py
```diff
@@ -427,3 +432,28 @@ def _get_gated_delta_product_mamba_layer_spec(in_proj, out_proj):
 mamba_inference_stack_spec = hybrid_inference_stack_spec
 gdp_stack_spec = gated_delta_product_stack_spec
 gdp_inference_stack_spec = gated_delta_product_inference_stack_spec
+
+
+def hybrid_dsv4_stack_spec(config: TransformerConfig) -> ModuleSpec:
```
Phlip79, line 437: Is this necessary?

FDecaYed: Yes, with the current spec API. The shared `hybrid_stack_spec` leaves `csa_layer` as `IdentityOp`, so this fills in the C/H/W attention spec while reusing the other layer and MTP specs without modifying the shared spec. The DSv4 spec factory needs the config to select the optional Q/K layer norms. Compression ratios still come from the individual layer configs.

Phlip79 (approved):

zhongbozhu (approved):

yaoyu-33:
This is now the concrete blocker for the Megatron-Bridge MCore-main bump in NVIDIA-NeMo/Megatron-Bridge#6008. On MCore main at `8190837c`, both H100 and GB200 DeepSeek L0 reach model construction and fail with:

> `TransformerLayer does not implement mHC residual streams`

Bridge correctly enables DeepSeek-V4 mHC, but current main has the DSv4 attention orchestration without the C/H/W HybridModel/HybridStack construction added here. Disabling mHC would change model semantics, and Bridge cannot patch the vendored MCore implementation. The migration is also tracked in NVIDIA-NeMo/Megatron-Bridge#4713.

This PR's functional matrix is green and it has two approvals; could a required code owner please complete the remaining approval/merge gate? Once merged, I can refresh #6008 to the resulting MCore main SHA and re-run its exact-head L0/full validation.

maanug-nv (approved):

megatron/core/models/hybrid/hybrid_layer_specs.py
jaredcasper: No! Specs do not rely on configuration. Period. They are statically defined in this file alone.

In the hybrid model the spec says what module to build for a given layer. `csa_layer` should not be `IdentityOp`, it should say what module to use to build a CSA layer. There should be no such thing as a "spec factory". This is the same path we went down with the GPT model and I do not want to go down this path on the Hybrid Model.

megatron/core/models/hybrid/hybrid_layer_specs.py
```diff
@@ -427,3 +432,28 @@ def _get_gated_delta_product_mamba_layer_spec(in_proj, out_proj):
...
+    dsv4_layer = ModuleSpec(
+        module=TransformerLayer,
+        submodules=TransformerLayerSubmodules(
+            input_layernorm=TENorm,
+            self_attention=dsv4_attention,
+            self_attn_bda=get_bias_dropout_add,
+        ),
+    )
+
+    submodules = replace(hybrid_stack_spec.submodules, csa_layer=dsv4_layer)
```
jaredcasper, line 458: We should not be changing specs like this. Specs are simple. They are defined once and used.

megatron/training/argument_utils.py
```diff
@@ -274,6 +274,79 @@ def _get_field_docstrings(self, src_cfg_class: type) -> dict[str, str]:
         return field_docstrings
 
 
+def _normalize_dsv4_hybrid_csa_compress_ratios(
```
jaredcasper, line 277: It feels like this should be in a deepseek v4 specific file, not in generic argument_utils.py.

megatron/training/argument_utils.py
```diff
@@ -508,7 +589,10 @@ def hybrid_config_from_args(
             not transformer_cfg.inference_fuse_tp_communication
         ), "inference_fuse_tp_communication is not supported for HybridModel"
     elif args.spec is not None:
-        kwargs["hybrid_stack_spec"] = import_module(args.spec)
+        hybrid_stack_spec = import_module(args.spec)
+        if callable(hybrid_stack_spec) and not isinstance(hybrid_stack_spec, ModuleSpec):
```
jaredcasper, line 593: A spec is by definition NOT a callable.

megatron/core/models/hybrid/hybrid_layer_specs.py
FDecaYed: Defined CSA directly in the static Hybrid spec; config-dependent Q/K norm selection now happens during module construction.

megatron/core/models/hybrid/hybrid_layer_specs.py
FDecaYed: Removed runtime spec replacement; the DSv4 name now aliases the static Hybrid spec.

megatron/training/argument_utils.py
FDecaYed: Moved DSv4 compression-ratio normalization into `megatron/training/models/deepseek_v4.py`.

megatron/training/argument_utils.py
FDecaYed: Removed config-aware spec calls; both Hybrid entry points now require a static `ModuleSpec`.

jaredcasper (approved):

PR #6415, opened 2026-08-10
Add megatron.core.telemetry base layer and observability guide

deepakn94:
## What does this PR do?

Adds `megatron/core/telemetry/` — the dependency-free foundation for Megatron's OpenTelemetry integration — its unit tests, and the `docs/user-guide/observability/` guide.

This is a **split of #6385** ("Telemetry/otel lens upstream") intended to make that PR smaller and cheaper to review. Original changes by @rhewett-nv.

### Contents

| File | Purpose |
|---|---|
| `megatron/core/telemetry/span_groups.py` | `MegatronSpanGroup` — the Megatron-specific span groups and the `default` / `per_step` / `all` granularity presets |
| `megatron/core/telemetry/training_metrics.py` | `record_training_metrics()` and the `megatron.training.*` metric instruments |
| `megatron/core/telemetry/fallbacks.py` | No-op `trace_fn` / `managed_span` / `span_cm` / `is_span_group_enabled` / `safe_set_span_attributes` |
| `megatron/core/telemetry/README.md` | Module overview and status |
| `docs/user-guide/observability/` | Six-page Observability user guide |
| `tests/unit_tests/telemetry/` | 83 unit tests |

### Why this splits cleanly

Nothing in this package imports anything from Megatron. Its only externals are `nemo-lens` and `opentelemetry`, both optional and both behind `try/except ImportError`. When neither is installed, `fallbacks` supplies no-op decorators and context managers, `span_groups` falls back to a local `SpanGroup` stub, and `record_training_metrics()` returns immediately. Call sites in the follow-up PR can therefore import from here unconditionally.

Nothing imports this package yet — that is deliberate. It is the base layer, and the call-site instrumentation lands in #6385.

### A note on the docs

The six guide pages are here, but the `docs/index.md` toctree entry that publishes them **stays in #6385**. Docs build nightly and on release, so wiring them into the toctree now would publish pages telling users to set `MEGATRON_OTEL_ENABLED=1` and pass `--otel-enabled` before either exists on main. A note on the section index says as much for anyone reading them in the repository meanwhile. They go live the moment #6385 adds the 14-line toctree hunk.

Changes made while moving them:

- `extending.md` and `metrics.md` both instructed readers to add metric modules under `megatron/core/instruments/`, a directory that has never existed in this repository — corrected to `megatron/core/telemetry/training_metrics.py`, which this PR adds.
- Dropped `observability-stack.md`. It walked through a `docker-compose.otel.yml` at the repository root and an `observability/` tree of collector, Prometheus, and Grafana configs, none of which exist here or in #6385. Readers wanting a local backend are pointed at the lens backends guide instead.
- The pages picked up the NVIDIA copyright headers that the other 26 files under `docs/` carry.

### Still in #6385

The call-site instrumentation (training loop, checkpointing, pipeline schedules, P2P, DDP, Mamba/transformer layers, inference server), the `--otel-*` CLI flags, the `MEGATRON_OTEL_*` environment variables, the `megatron-core[otel]` install extra, and the `docs/index.md` toctree entry.

When #6385 rebases onto this, its eight `from megatron.core.telemetry._fallbacks import ...` sites need updating to `megatron.core.telemetry.fallbacks`.

This PR requires only `@NVIDIA/core-adlr` / `@NVIDIA/core-nemo`, versus ten CODEOWNERS groups for #6385 as it stands. `docs/user-guide/**` and `tests/unit_tests/telemetry/` have no CODEOWNERS entries, so they add no reviewers.

## Issue tracking

Linked issue: split out of #6385

### Pre-checks

### Tests

`tests/unit_tests/telemetry/` — 83 tests, all CPU-only, no distributed setup, and no optional dependency required. They cover:

- Group constants, uniqueness, and non-collision with the base lens groups.
- Preset nesting: `default` stays coarse-grained, `per_step` adds step-level groups, the verbose per-layer and per-marker groups (`microbatch`, `layer`, `activation_offload`, `trace_region`) remain opt-in under `all` only.
- Instrument creation, per-meter caching, and that the weak-keyed cache does not retain meters.
- Value handling: `None` skipped, `skipped_iters=0` not exported, zero recorded for gauges, `grad_norm` coerced to `float`.
- Failure paths: a broken meter logs and is not cached, and the recorder no-ops without `opentelemetry`.

Assertions that depend on the real `nemo.lens` being present are gated, so the suite is meaningful with or without it rather than skipping wholesale.


megatron/core/telemetry/fallbacks.py
Phlip79, line 1: Can we avoid having files starting with `_`?

Phlip79 (approved):

megatron/core/telemetry/fallbacks.py
deepakn94: Renamed to `megatron/core/telemetry/fallbacks.py` in b5aa0de — agreed, the underscore was wrong for a module that call sites across `megatron/core` and `megatron/training` are meant to import from.

Note for #6385: its eight `from megatron.core.telemetry._fallbacks import ...` sites will need the same rename when it rebases onto this.

jaredcasper (approved):

deepakn94:
/ok to test [d98c555](https://github.com/NVIDIA/Megatron-LM/pull/6415/commits/d98c555840f59b05de90e89a10d5da9bac6a1321)

PR #6416, opened 2026-08-10
Defer prefix-cache hit accounting until admission succeeds

lmcafee-nvidia:

## What does this PR do?

This PR fixes two related prefix-cache accounting errors. Engine-wide hit and matched-block counters now update only after a request successfully allocates its uncached tail, and `request.num_cached_tokens` now records prompt tokens whose prefill was actually skipped rather than every matching KV block. This keeps internal statistics and OpenAI cached-token usage accurate without changing allocation or model outputs.

## Example failure scenario(s)

**Scenario 1 (failed admission)**

- Cache: 2 reusable blocks.
- Request A: 3-block prompt; 0 free tail blocks.
- Attempt 1 fails; old totals add 1 hit and 2 blocks.
- Retry succeeds; old totals add the same reuse again.
- Fix reports 1 hit and 2 matched blocks after retry.

**Scenario 2 (partial Mamba coverage)**

- Block size: 256 tokens.
- Request A matches 3 KV blocks and 1 Mamba block.
- The engine skips 256 tokens and recomputes 512.
- The old response reports 768 cached tokens.
- The fix reports 256 cached tokens.

## How was this bug discovered?

A prefix-cache pressure test forced a request to match saved blocks and then fail allocation for its uncached tail. Exact before-and-after checks showed that the rejected attempt changed accounting and the retry counted the same reuse again. During review, Siddharth Singh noticed that matching KV blocks can exceed the prompt work actually skipped; hybrid Mamba checks confirmed the mismatch, with `prefix_skip_tokens` as the correctness oracle.

## Dependency and scope

This is an independent production-fix PR against `main`. `prefix_cache_hits` and `prefix_cache_blocks_matched` continue to describe successful KV reuse, while `request.num_cached_tokens` describes prompt prefill actually skipped after Mamba limits and minimum-prefill backoff. The change affects internal accounting and user-visible cached-token usage metadata; it does not change block allocation, generated tokens, log probabilities, or model numerics.

### Pre-checks

- [ ] I have run the autoformatter on my PR

## Validation

Exact signed tip: `5c30ae89e4a37dcb003ba658493e554919bc1c13`

- DFW Slurm job `17583336`: passed at the exact signed tip
  - failed-admission rollback and retry accounting
  - hybrid Mamba no-skip and partial-skip accounting
  - 3 passed, 0 failed, and 0 skipped on each of 8 ranks
- `git diff --check`, isort, Ruff, and Python bytecode compilation passed on the three changed files.
- Black passed on both changed production files; the unit-test file retains its pre-existing current-main Black delta to avoid unrelated churn.


santhnm2 (approved):

kvareddy (approved):

megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -3216,6 +3207,14 @@ def add_request(
                     self.kv_block_allocator.block_ref_counts[matched_tensor] -= 1
                 raise BlockOverflowError(req.request_id)
 
+        # Track prefix cache hits only after allocation succeeds. num_cached_tokens
+        # accumulates across successful prefill chunks; a failed admission can be
+        # retried and must not count the same cached prefix twice.
+        if num_matched_blocks > 0:
+            self.prefix_cache_hits += 1
+            self.prefix_cache_blocks_matched += num_matched_blocks
+            req.num_cached_tokens += num_matched_blocks * self.block_size_tokens
```
sidsingh-nvidia: Should we use `+=prefix_skip_tokens` instead? I believe that is the true amount of compute skipped because of mamba.

sidsingh-nvidia (changes requested):

megatron/core/inference/contexts/dynamic_context.py
lmcafee-nvidia: thanks! fixed this in the latest commit

sidsingh-nvidia (approved):

PR #6417, opened 2026-08-10
Preserve greedy temperature in RL inference

lmcafee-nvidia:

## What does this PR do?

This PR preserves an explicit `temperature=0.0` when an RL request passes through `MegatronLocal`, while retaining `1.0` as the default when temperature is unspecified. The old truth-value fallback treated both `None` and floating-point zero as missing; the fix uses an explicit `None` check so only an omitted value receives the default.

## Example failure scenario(s)

**Scenario (explicit zero temperature)**

- Request A sets `temperature=0.0` for greedy output.
- The old code treats `0.0` as missing and sends `1.0`.
- The model can sample instead of choosing the most likely token.
- The fix preserves `0.0` and keeps the request greedy.

## How was this bug discovered?

While making RL inference comparisons deterministic, the setup requested `temperature=0.0` and inspected the keyword arguments forwarded by `MegatronLocal`. It observed `1.0`, although an explicit zero should have been forwarded unchanged. The detection oracle is the mocked client's exact received temperature for both relevant cases: an omitted temperature must become `1.0`, while an explicit `0.0` must remain `0.0`.

## Issue tracking

Linked issue: N/A

### Pre-checks

## Validation

Exact signed tip: `abcdda5160de412644a7bf131bb2b93c7e6407f4`

- DFW Slurm job `17582623`: passed at the exact signed tip
  - omitted temperature defaults to `1.0`
  - explicit `temperature=0.0` remains `0.0`
  - 2 passed, 0 failed, and 0 skipped on each of 8 ranks
- isort and Black passed on the changed-import test file.
- Ruff and Python bytecode compilation passed on both changed Python files.
- `git diff --check 3c04d2bd2255c9652a687c3d5a5b9636467696db...abcdda5160de412644a7bf131bb2b93c7e6407f4` passed.


megatron/rl/inference/megatron.py
```diff
@@ -55,10 +55,11 @@ async def base_generate(self, request: InferenceRequest) -> InferenceResponse:
         # Things that may be problematic when doing this switch
         # - Add BOS token
         # - Skip prompt logprobs
+        temperature = request.generation_args.temperature
         response = await client.chat.completions.create(
             model="",
             messages=[message.model_dump() for message in request.prompt],
-            temperature=request.generation_args.temperature or 1.0,
+            temperature=1.0 if temperature is None else temperature,
             top_p=request.generation_args.top_p or 0.0,
```
tdene, line 65: Is this line also dealing with the same problem? If so, can you bundle a fix into this PR?

tdene (approved):

kvareddy (approved):

megatron/rl/inference/megatron.py
lmcafee-nvidia: No—the fallback here is also `0.0`, so an explicit `top_p=0.0` remains `0.0`. Unlike temperature, this expression does not replace the caller's explicit value with a different value, so I kept this PR scoped to the temperature fix.

Issue #40, opened 2020-08-06
Scaling up to GPT-3 size (175B)


nakosung:
Will it be straightforward to scale Megatron LM up to 175B?

Issue #60, opened 2021-01-05
OOM when training the same size of gpt2(2.6B) with mp=2 dp=8 with 64*V100(32GB)


Hanlard:
**We train gpt2(2.6B) with following parameters ，but OOM**

using world size: 16 and model-parallel size: 2 
using torch.float16 for parameters ...
-------------------- arguments --------------------
  adam_beta1 ...................... 0.9
  adam_beta2 ...................... 0.999
  adam_eps ........................ 1e-08
  adlr_autoresume ................. False
  adlr_autoresume_interval ........ 1000
  apply_query_key_layer_scaling ... False
  apply_residual_connection_post_layernorm  False
  attention_dropout ............... 0.1
  attention_softmax_in_fp32 ....... False
  batch_size ...................... 8
  bert_load ....................... None
  bias_dropout_fusion ............. False
  bias_gelu_fusion ................ False
  block_data_path ................. None
  checkpoint_activations .......... True
  checkpoint_num_layers ........... 1
  clip_grad ....................... 1.0
  data_impl ....................... mmap
  data_path ....................... /raid/gpt3-train-data/filterBy256-100G-notag_text_document
  DDP_impl ........................ local
  distribute_checkpointed_activations  True
  distributed_backend ............. nccl
  dynamic_loss_scale .............. True
  eod_mask_loss ................... False
  eval_interval ................... 10000
  eval_iters ...................... 10
  exit_interval ................... None
  faiss_use_gpu ................... False
  finetune ........................ False
  fp16 ............................ True
  fp16_lm_cross_entropy ........... True
  fp32_allreduce .................. False
  hidden_dropout .................. 0.1
  hidden_size ..................... 1920
  hysteresis ...................... 2
  ict_head_size ................... None
  ict_load ........................ None
  indexer_batch_size .............. 128
  indexer_log_interval ............ 1000
  init_method_std ................. 0.02
  layernorm_epsilon ............... 1e-05
  lazy_mpu_init ................... None
  load ............................ checkpoints/gpt2_64_xxxM
  local_rank ...................... 0
  log_interval .................... 5
  loss_scale ...................... None
  loss_scale_window ............... 1000
  lr .............................. 0.00015
  lr_decay_iters .................. 70000
  lr_decay_style .................. cosine
  make_vocab_size_divisible_by .... 128
  mask_prob ....................... 0.15
  max_position_embeddings ......... 1024
  merge_file ...................... bpe_3w_new/merges.txt
[...]

Issue #39, opened 2020-08-02
Is there a pretrained 8.3B parameter model?


shamoons:
I believe this repo links to the 345M parameter model. Is there any way to get the 8.3B parameter one?

zainsarwar865:
Same. I need the 8.3B one as well.

tkwa:
This is marked as completed but I can't find the 8.3B model on the NGC catalog. Was it never released or removed again?

Issue #38, opened 2020-07-07
FileExistsError when training with a shared file-system


g-karthik:
When training on a multi-node cluster with a shared file-system, I observe the following:

```
initializing model parallel cuda seeds on global rank 0, model parallel rank 0, and data parallel rank 0 with model parallel seed: 3952 and data parallel seed: 1234
configuring data
Traceback (most recent call last):
File "/fsx/DeepSpeedExamples/Megatron-LM/pretrain_gpt2.py", line 707, in <module>
    main()
File "/fsx/DeepSpeedExamples/Megatron-LM/pretrain_gpt2.py", line 652, in main
    args.eod_token = get_train_val_test_data(args)
  File "/fsx/DeepSpeedExamples/Megatron-LM/pretrain_gpt2.py", line 598, in get_train_val_test_data
    args)
  File "/fsx/DeepSpeedExamples/Megatron-LM/configure_data.py", line 34, in apply
    return make_loaders(args)
  File "/fsx/DeepSpeedExamples/Megatron-LM/configure_data.py", line 170, in make_loaders
    train, tokenizer = data_utils.make_dataset(**data_set_args)
  File "/fsx/DeepSpeedExamples/Megatron-LM/data_utils/__init__.py", line 93, in make_dataset
    datasets = [get_dataset_from_path(p) for p in path]
  File "/fsx/DeepSpeedExamples/Megatron-LM/data_utils/__init__.py", line 93, in <listcomp>
    datasets = [get_dataset_from_path(p) for p in path]
  File "/fsx/DeepSpeedExamples/Megatron-LM/data_utils/__init__.py", line 83, in get_dataset_from_path
    make_lazy(path_, text.X, data_type='data')
  File "/fsx/DeepSpeedExamples/Megatron-LM/data_utils/lazy_loader.py", line 51, in make_lazy
    os.makedirs(lazypath)
  File "/usr/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/fsx/datasets/openwebtext/openwebtext.lazy'

```

I don't see the same error when training on a single independent node (with its own file-system) though. Do you have a clean fix for this?

PR #6418, opened 2026-08-10
Add prefix cache stress test foundation

lmcafee-nvidia:
## What does this PR do?

This PR adds a unit-only differential interaction foundation for prefix caching. It verifies cache-policy behavior directly at the inference-context layer, then runs real models with prefix caching disabled and enabled while another inference feature remains active in both runs.

The foundation adds:

- 10 context cases covering five prefix-sharing behaviors under both `LRU` and `REF_ZERO`;
- eight real-engine rows covering TP4, PP4, EP4 with a real MoE layer, chunked prefill, FlashInfer fused RoPE, MTP depth two, hybrid Mamba, and hybrid GDP;
- feature-owned runtime witnesses, so a row only counts when a real cache hit and the claimed companion path execute together; and
- test-builder support for an explicitly requested hidden size and MoE expert count, which lets the fused-RoPE and EP rows instantiate the intended models.

An audit found that earlier coverage could pass after merely enabling prefix caching: it did not consistently create reusable prefixes, exhaust the allocator, prove eviction or physical reuse, or show that a companion kernel or topology processed a cache-hit request.

The matrix therefore has two layers:

1. The inexpensive context matrix crosses every selected sharing behavior with both eviction policies. This isolates matching, reference-count, eviction, and reuse semantics without model execution.
2. The real-engine matrix selects the major execution and state owners that can interact with cached prefill: tensor, pipeline, and expert parallelism; chunked scheduling; fused rotary position encoding; speculative decoding; and hybrid Mamba and GDP recurrent state. Each row forces the companion path to execute on a request that reused a prefix.

This is **differential interaction coverage**, not a universal four-cell Boolean 2x2 or a covering array over every inference flag. In each real-engine row, the primary flag is compared as `enable_prefix_caching=False` versus `enable_prefix_caching=True`, while the selected companion configuration stays active in both runs. Companion-disabled cells are not repeated, and each expensive engine feature uses one representative cache policy because the five-by-two context matrix already exercises both policy semantics exhaustively.

### Context behavior and policy matrix (10 cases)

Prefix caching is enabled in all 10 cases. Every behavior below is run once with `prefix_caching_eviction_policy=LRU` and once with `prefix_caching_eviction_policy=REF_ZERO`, under deliberate allocator exhaustion.

| Behavior forced | LRU case | REF_ZERO case | Expected prefix behavior |
| --- | --- | --- | --- |
| Exact three-block follower | `exact-prefix-lru` | `exact-prefix-ref_zero` | Match all three blocks and report all three as cached. |
| Two matching blocks followed by one divergent block | `partial-prefix-lru` | `partial-prefix-ref_zero` | Match only the first two blocks. |
| Fully disjoint three-block follower | `missing-prefix-lru` | `missing-prefix-ref_zero` | Match no blocks and report zero cached tokens. |
| Three concurrent followers sharing one producer | `concurrent-sharing-lru` | `concurrent-sharing-ref_zero` | Increase and then release the exact shared-block references. |
| One cached and one fresh request in the same batch | `mixed-cached-fresh-lru` | `mixed-cached-fresh-ref_zero` | Keep each request's query length and cached-token accounting independent. |

After each behavior, a divergent three-block request fills the allocator. LRU cases must evict exactly the retained least-recently-used block IDs. REF_ZERO cases must remove released hashes immediately and physically recycle the released producer block IDs.

### Real-engine differential matrix (eight rows)

Every row executes two otherwise equivalent sessions: cache off (`enable_prefix_caching=False`) and cache on (`enable_prefix_caching=True`). The companion settings listed below remain active in both sessions.

| Row | Companion configuration active in both sessions | Cache policy in the treatment |
| --- | --- | --- |
| `tp4-ref-zero` | GPT with `tensor_model_parallel_size=4`, `pipeline_model_parallel_size=1`, and `expert_model_parallel_size=1` | `REF_ZERO` |
| `pp4-lru` | GPT with `pipeline_model_parallel_size=4`, `tensor_model_parallel_size=1`, and `expert_model_parallel_size=1` | `LRU` |
| `ep4-moe-lru` | GPT MoE with `expert_model_parallel_size=4` and four real experts; TP=PP=1 | `LRU` |
| `gpt-chunked-lru` | GPT with `enable_chunked_prefill=True`, `max_tokens=264`, and an extended 781-token follower that must continue prefill after its hit | `LRU` |
| `gpt-fused-rope-ref-zero` | GPT with `position_embedding_type="rope"`, `use_flashinfer_fused_rope=True`, hidden size 64, four attention heads, and head dimension 16 | `REF_ZERO` |
| `mtp2-ref-zero` | GPT with `num_speculative_tokens=2`, two MTP layers, and `materialize_only_last_token_logits=False` | `REF_ZERO` |
| `hybrid-mamba-ref-zero` | Hybrid Mamba/attention/MLP model; the cache-on session allocates durable recurrent-state prefix storage with `prefix_caching_mamba_gb=0.2` | `REF_ZERO` |
| `hybrid-gdp-lru` | Hybrid GDP/attention/MLP model with `ssm_mixer="gdp"`; the cache-on session allocates durable recurrent-state prefix storage with `prefix_caching_mamba_gb=0.2` | `LRU` |

The common workload uses 256-token KV blocks and three donor/follower/divergent-pressure waves, for nine requests per session. Base prompts contain 517 tokens, each request greedily generates four tokens with the Torch sampling backend and `top_k=1`, and the allocator is deliberately driven to exhaustion. The chunked follower is longer so the hit-bearing request itself must continue in another prefill chunk.

The GDP row uses FP32 recurrent state in both sessions, following the existing GDP prefix-cache tests, so BF16 state rounding cannot hide or create a cache-on/cache-off difference.

## How are results checked?

The context cases check exact matched-block lists, cached-token and query-length accounting, concurrent reference counts, hash discoverability, exact LRU victims, and exact REF_ZERO block-ID recycling.

For every real-engine row, the cache-on run must reach zero free blocks and record at least three cache hits, six matched blocks, and three prefix-coordination waits. LRU rows must evict earlier hashes; REF_ZERO rows must reuse physical block IDs across waves. The cache-off run must record zero prefix hits.

Each companion has a runtime witness tied to cached work:

- TP counts real row-parallel reductions on the cache-hit step.
- PP counts the pipeline-parallel forward path on the cache-hit step.
- EP records real MoE dispatches and confirms that a cached follower's request ID participated.
- Chunking confirms that the hit-bearing follower continues prefill.
- Fused RoPE counts the real FlashInfer rotary application and verifies its cache is on CUDA.
- MTP confirms that a cached request generates during a decode step with real speculative proposals.
- Mamba confirms matched blocks and a real `restore_to_live` call.
- GDP counts a real `GatedDeltaProductMixer` prefill on the cache-hit step, confirms matched blocks, and records a real `restore_to_live` call on that same step.

Only after those witnesses pass does the test compare outputs. For all nine requests, cache off and cache on must have identical generated token IDs and text, complete successfully, and produce exactly four tokens. Generated log probabilities use `rtol=2e-2` and `atol=5e-2`. Non-chunked rows also compare generated top-5 shape, selected-token membership and score consistency, ranked values, and near-tied cutoff candidates within tolerance. The chunked row still compares generated log probabilities but sets `top_n_logprobs=0` because intermediate chunking currently emits an extra top-N entry.

## Scope

This PR contains only shared unit-test infrastructure and coverage. It has no Megatron Core production changes, functional tests, functional configurations, runners, prompts, recipes, invocations, or goldens.

The shared matrix deliberately sets `skip_prompt_log_probs=True`; complete prompt-logprob parity and the HTTP result-field regression belong to dependent replacement PR #60, which supersedes historical PR #41. Focused Mamba publication and checkpoint reconstruction regressions remain in dependent PRs #38 and #39 with their production fixes. Closed PRs #40 and #42 are historical and are not part of the active split. Production defects exposed while building the aggregate are not fixed in this foundation.

## Validation

- DFW H100 unit job `17575659` at `b2de5e6b79ca78bdb0fc771b4426afe9debbfdc7`
- Exact 8-rank focused command passed on every rank: `18 passed, 61 deselected, 0 failed, 0 skipped` per rank; `hybrid-gdp-lru` ran and passed rather than skipping
- `git diff --check`, isort, Black, Ruff, and `py_compile` passed locally and on DFW


santhnm2 (approved):
LGTM. If this gets merged after [!6780](https://github.com/NVIDIA/Megatron-LM/pull/6780) could you add GDP coverage as well?

wdykas (approved):

kvareddy (approved):

lmcafee-nvidia:
@santhnm2, #6780 is now merged, so I merged current `main` and added the requested GDP row. It runs real GDP inference with prefix caching off and on under LRU pressure, proves that a cache-hit request executed GDP prefill and restored recurrent state, and checks token, text, and log-probability parity for all nine requests.

DFW H100 job `17575659` passed at `b2de5e6b79ca78bdb0fc771b4426afe9debbfdc7`: all 18 selected tests passed on all eight ranks with no failures or skips. Could you please take another look at the GDP addition?

PR #6421, opened 2026-08-10
build(container): support non-root CI execution

ko3n1g:
## Background / motivation

- ARC and sandboxed repository CI materialize the immutable test commit at `/opt/megatron-lm` and may execute it as a non-root UID.
- The dev image previously exposed `uv` only below `/root` and did not prepare the writable source/cache paths needed by that runtime.
- Existing image consumers still require root to remain the default runtime user.

## What changed

- Install `uv` directly on the system path, from a checksum-verified release archive.
- Add UID/GID 65532 and prepare its home, source, test-data, cache, and virtual-environment paths.
- Own the non-root cache root itself, not only its pre-created children, so a caller may create new cache subdirectories under it.
- Validate the non-root executable/write contract during the build, then restore root as the final user of both `main` and `jet`.
- Replace hard-coded `/root/.netrc` paths in 45 H100/GB200 recipes with `${HOME}/.netrc`.

## Details

- Runtime callers opt into UID/GID 65532 and provide `HOME` plus cache environment variables; the image does not impose that policy on other consumers.
- Both public `main` and internal `jet` stages end with `USER root`.
- `/opt/venv` directory ownership permits CI cases that uninstall preinstalled packages while keeping the baked environment executable.
- `install -d` applies ownership only to the leaves it creates, so naming `/tmp/regent-cache/pip` alone left the cache root itself `root:root 0755`. A non-root caller could write every pre-created leaf and still fail to create anything new below the root — which is what a caller-provided `XDG_CACHE_HOME` or `HF_HOME` does on first use. The cache root is now named explicitly.
- The non-root proof block probes directory creation under the cache root, not only file writes into an already-owned leaf, so the layer that grants the permission is the layer that proves it.
- The change covers `docker/Dockerfile.ci.dev` and the 45 recipes that create or remove `.netrc` credentials.
- The older LTS image is intentionally unchanged.

## Tested

- Exact branch head: `4cd0f556e40fe333a0bda69ad2a83c089bfe7147`.
- `docker buildx build --check --target main --build-arg FROM_IMAGE_NAME="$(cat docker/.ngc_version.dev)" --build-arg IMAGE_TYPE=dev -f docker/Dockerfile.ci.dev .` — `Check complete, no warnings found.`
- `git diff --check github/main...HEAD` — passed.
- `git verify-commit HEAD` — valid GPG signature; DCO sign-off present.
- Full `--target main` image build from `nvcr.io/nvidia/pytorch:26.06-py3` in the sandboxed CI image builder — succeeded, including the in-Dockerfile non-root proof layer. Static inspection of the resulting image config reports `User: root`, so the published default is unchanged.
- Runtime verification of that image on Kubernetes under the security context CI actually applies (`runAsUser`/`runAsGroup` 65532, `runAsNonRoot: true`, `capabilities.drop: [ALL]`, `seccompProfile: RuntimeDefault`, pod `fsGroup: 65532`):
  - A container with no security context starts as `uid=0(root)`.
  - As 65532: `$HOME`, `/opt/venv` and its `site-packages`, `/opt/megatron-lm`, `/workspace/data/cache`, `/home/gitlab-runner/data/gpt3_data`, and the cache root plus its `pip`/`uv`/`xdg` children are all writable.
  - As 65532: `uv --version`, `python -V`, `import torch`, and `uv pip list` all succeed against the baked environment.
- That runtime check is what surfaced the cache-root ownership gap: before this fix the same probe reported `FAIL write /tmp/regent-cache (owner=root:root 755)` and could not create the cache subdirectory the runtime asks for on first use.


docker/Dockerfile.ci.dev
```diff
@@ -145,4 +216,6 @@ RUN --mount=type=secret,id=JET_INDEX_URLS \
     uv pip install --no-cache-dir --upgrade "setuptools>=80"
     uv pip install --no-cache-dir --upgrade $JET_INDEX_URLS "jet-client~=4.0" 
 EOF
+
+USER root
```
thomasdhc, line 220: [P3] Findings: `main` chowns every `/opt/venv` directory to the runtime UID and proves the non-root contract, but the root `pip`/`uv pip install` commands in this derived `jet` stage then create root-owned package directories (`one_logger/`, `jet_client/`, `setuptools/` and their `*.dist-info/`), so a non-root caller of the internal image cannot uninstall or modify exactly those packages. Proposal: repeat `find "${UV_PROJECT_ENVIRONMENT}" -type d -exec chown "${RUNTIME_UID}:${RUNTIME_GID}" {} +` in the `jet` stage before this final `USER root`.

thomasdhc (approved):
LGTM.

### Review notes
- Both `uv` 0.7.2 archive checksums match the upstream release manifests, no `/root/.netrc` reference remains anywhere, and every recipe `script` block still renders with `${HOME}` intact after the brace escaping.
- Review limits: the non-root contract is exercised only by the in-Dockerfile probe layer, and `spec.script_setup` is rendered by no in-repo formatter, so JET-side handling of its single-brace `${HOME}` was not verifiable here.

PR #6422, opened 2026-08-10
Add tensor-parallel Muon Hyperball optimizer

mkhona-nvidia:
## Summary

- add `TensorParallelMuonHT`, registered as `--optimizer muon_ht`
- implement Hyperball update and weight normalization through Emerging Optimizers' pre/post weight-update extension points
- initialize every Muon-managed matrix to the configured radius in Megatron's model-initialization lifecycle, before mixed-precision copies are created
- compute logical Frobenius norms across dense/expert TP and GTP-remat shards without reducing replica-only axes
- require an explicit `ProcessGroupCollection` throughout MuonHT; the new path never falls back to `use_mpu_process_groups`
- keep the fixed radius as configuration rather than per-parameter optimizer state, avoiding scalar distributed-checkpoint entries
- keep the implementation summary in the `TensorParallelMuonHT` docstring; no optimizer-directory README is added

## Update rule

For Muon's orthogonalized update `u_t`, learning rate `eta_t`, configured radius `R`, and `Normalize(X) = X / ||X||_F`:

```
W_(t+1) = R * Normalize(W_t - eta_t * R * Normalize(u_t))
```

The pre-update hook rescales the logical update norm to `R`; the post-update hook projects the logical updated weight back to `R`. Numerical-zero decisions consistently use `norm < hyperball_eps`, without dtype-relative tolerances. This replaces weight decay, so `muon_ht` requires `weight_decay=0`.

## Integration and initialization

`TensorParallelMuonHT` subclasses Megatron's `TensorParallelMuon`, which subclasses Emerging Optimizers' `OrthogonalizedOptimizer`. It reuses momentum, Newton-Schulz orthogonalization, QKV handling, and TP/GTP-aware Muon behavior while implementing the constraint through `pre_weight_update_fn_inplace` and `post_weight_update_fn_inplace`.

Megatron owns the initialization contract: when `muon_ht_radius` is configured and parameter initialization is enabled, model construction scales each Muon-managed matrix to `R` before mixed-precision wrapping. Checkpoint-oriented construction with `perform_initialization=False` does not mutate parameters.

## Sharding and checkpoint behavior

- norm sums use `all_reduce(SUM)` over the applicable `tp`/`expt_tp` and `gtp_remat`/`expt_gtp_remat` groups
- TP and GTP-remat axes are reduced sequentially to cover the logical tensor's Cartesian shard product
- replica-only axes are excluded
- MuonHT requires an explicit process-group collection at construction, initialization, and factory entry points
- the configured radius is not stored as per-parameter optimizer state, so distributed checkpointing does not encounter scalar state for sharded parameters

## Tests added

- update normalization and post-update radius preservation
- full optimizer-step radius preservation without persistent radius state
- fixed-radius, weight-decay, epsilon, and required-process-group validation
- model initialization filters exactly Muon-managed parameters
- model initialization rejects norms below epsilon and accepts norms equal to epsilon
- model lifecycle initialization runs before distributed/mixed-precision wrapping
- checkpoint-oriented `perform_initialization=False` leaves parameters unchanged
- optimizer factory construction and explicit-process-group enforcement
- multi-rank TP logical norm normalization
- multi-rank GTP-remat logical norm normalization

## Validation

Passed locally:

- isort 5.13.2 import check
- Black 25.11.0 compatibility check (the requested Black 26 runtime is unavailable locally)
- Ruff 0.9.10 on touched Python files
- Python 3.12 warnings-as-errors byte compilation
- `git diff --check`

CUDA/multi-rank pytest cases were not executed locally because this host does not provide PyTorch or a CUDA test environment; they are included for repository CI.

## Research context

- [ELR/Hyperball overview](https://hy.tencent.ai/research/elr)
- [Emerging Optimizers](https://github.com/NVIDIA-NeMo/Emerging-Optimizers)


megatron/core/optimizer/README_MuonHT.md
```diff
@@ -0,0 +1,153 @@
+<!---
```
philipcmonk: The filename `optimizer/README.md` suggests it applies to the whole directory.  Maybe `optimizer/README_MuonHT.md` or something like that?

philipcmonk (commented):
The sharding and checkpointing logic looks fine to me.  I imagine this will be reasonably fast as-is, but if calculating the per-parameter global frobenius norm ends up being slow, we could batch the parameters together and call `multi_tensor_l2norm`.

mchrzanowski (approved):

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -107,13 +108,16 @@ class EmergingOptimizerEntry:
         optimizer_cls: The torch optimizer class.
         init_state_fn: Lazily initialises optimizer state (needed for checkpoint formats).
         config_to_kwargs: ``(config, model_chunks, pg_collection) -> dict`` of constructor kwargs.
+        param_init_fn: Optional parameter initialization performed by the optimizer factory before
```
skyw: Why this is needed? is it for distributed checkpoint or something else?
The unspoken convention for distributed checkpoint is to have a _init_group function in emerging optimizers that megatron can call.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -355,6 +361,184 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
 
+def _muon_ht_is_tensor_parallel_shard(p: torch.Tensor) -> bool:
+    """Return whether ``p`` contains a unique tensor-parallel shard."""
+    is_parallel = getattr(p, "tensor_model_parallel", None)
+    if is_parallel is not None:
+        return bool(is_parallel)
+    return getattr(p, "partition_dim", None) not in (None, -1)
+
+
+def _muon_ht_is_expert_parameter(p: torch.Tensor) -> bool:
```
skyw: Q: No existing helper function does this? there should be multiple throughout the code base.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -355,6 +361,184 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+    """Return whether ``p`` uses the expert tensor-parallel topology."""
+    return bool(getattr(p, "expert_tp", False) or not getattr(p, "allreduce", True))
+
+
+def _muon_ht_parameter_shard_groups(
+    p: torch.Tensor, pg_collection: ProcessGroupCollection | None
+) -> tuple[torch.distributed.ProcessGroup, ...]:
+    """Return process groups spanning all unique shards of ``p``."""
+    if pg_collection is None:
```
skyw: New code shouldn't allow pg_collection to be None.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -355,6 +361,184 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+    if radius is None:
+        raise ValueError("TensorParallelMuonHT requires hyperball_radius to be configured")
+    if not math.isfinite(radius) or radius < eps:
+        raise ValueError(
+            f"hyperball_radius must be finite and at least hyperball_eps={eps}, got {radius}"
+        )
+
+
+@torch.no_grad()  # type: ignore[misc]
+def _initialize_muon_ht_parameters(param_groups: list[dict], kwargs: Dict[str, Any]) -> None:
```
skyw: This seems to be specific to this optimizer, no need to change the general interface just for it.

tests/unit_tests/test_emerging_optimizers.py
```diff
@@ -516,6 +706,103 @@ def test_muon_optimizer_blockwise_mode_different_result(self):
...
+    def setup_and_teardown(self):
+        """Use the full test world as one GTP-remat shard group."""
+        world = int(os.getenv('WORLD_SIZE', '1'))
+        Utils.initialize_model_parallel(gtp_remat_size=world)
+        yield
+        Utils.destroy_model_parallel()
+
+    def test_muon_ht_uses_global_norm_for_gtp_shards(self):
+        """Hyperball norms should include every unique GTP-remat shard."""
+        pg_collection = ProcessGroupCollection.use_mpu_process_groups()
```
skyw: New code should not use use_mpu_process_groups fallback.

megatron/core/optimizer/emerging_optimizers.py
```diff
@@ -355,6 +361,184 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor, **kwargs: Any) -> t
...
+
+        for group in self.param_groups:
+            if group["weight_decay"] != 0.0:
+                raise ValueError(
+                    "TensorParallelMuonHT replaces weight decay with Hyperball norm control; "
+                    "set weight_decay=0.0"
+                )
+            for p in group["params"]:
+                p_norm = self._global_frobenius_norm(p, p)
+                dtype_rtol = torch.finfo(p.dtype).eps if p.is_floating_point() else 0.0
```
skyw: don't use floating point type specifics.

PR #6428, opened 2026-08-11
Test streaming responses with async scheduling

lmcafee-nvidia:
## What does this PR do?

This PR adds a standalone end-to-end functional test for streaming responses under asynchronous scheduling. It launches the real OpenAI-compatible inference server once with legacy scheduling and once with asynchronous scheduling, then sends the same synchronized concurrent workload to both.

The test adds:

- one self-contained server runner that sends six ordinary requests and one streaming request at the same time;
- real HTTP and server-sent-event coverage through the production server;
- checks for incremental text, present and finite per-token log probabilities, and terminal response reconstruction;
- exact cross-mode comparison of the streamed text, token IDs, finish state, and chunk sizes; and
- one H100 functional-test recipe.

The async-scheduling unit matrix exercises scheduler behavior in process, while streaming unit tests exercise individual client and response-formatting components. The remaining gap was the boundary between them: a real server had not been tested while asynchronous scheduling served a streaming request under concurrent admission pressure.

The test therefore uses a focused differential comparison. It runs the same stream-enabled workload once with `--inference-dynamic-batching-async-sched-mode=legacy` and once with `--inference-dynamic-batching-async-sched-mode=async`. Six simultaneous non-streaming requests keep the server busy while the streaming request progresses, and the server admits at most two requests at a time.

This is the separately approved end-to-end functional exception, not another pairwise-matrix layer. It does not repeat stream-disabled cells and receives no pairwise-coverage credit.

### Scheduler comparison (two sessions)

Both sessions use the same local checkpoint and tokenizer fixture, deterministic settings, TP2/PP2/DP2 topology, prompts, and synchronized request release. Only the scheduling mode changes.

| Session | Scheduling mode | Concurrent workload |
| --- | --- | --- |
| Control | `legacy` | Six ordinary requests plus one streaming request |
| Treatment | `async` | The identical six ordinary requests plus one streaming request |

### Request mix (seven requests per session)

All seven requests use greedy generation, ignore EOS, and request log-probability data. A barrier releases them together against a two-request active-capacity limit.

| Request type | Count | Generated-token counts | Streaming interval |
| --- | ---: | --- | ---: |
| Ordinary completion | 6 | 24, 6, 18, 8, 20, 10 | — |
| Streaming completion | 1 | 16 | 2 |

## How are results checked?

Each ordinary request must return one nonempty completion, finish because it reaches its requested length, and contain the expected number of token IDs and finite log probabilities.

The streaming request must use the SSE content type, emit valid `data:` records and a final `[DONE]` marker, produce multiple one- or two-token deltas totaling 16 tokens, and emit exactly one terminal result. Every delta and the terminal result must contain the expected number of finite log probabilities. Joining the text deltas must reproduce the terminal text exactly.

Finally, the legacy and async streamed results must agree exactly in text, token IDs, finish reason, delta sizes, and response structure. Numeric log-probability values are validated independently in each session because different dynamic batch shapes can produce small BF16 score differences; exact generated content is compared only for the streaming request under test.

## Scope

This PR adds one functional-test runner and one H100 recipe. It changes no production code, existing functional test, prompt, or golden output. The runner supplies the explicit architecture used by the repository's existing Qwen3-8B functional tests because the converted checkpoint does not store model arguments.

The selected checkpoint and tokenizer are locally provisioned test fixtures, not subjects of coverage. The behavior under test is the interaction between real HTTP streaming and asynchronous scheduling. The six ordinary completions provide concurrent load and are validated within each session; only the streaming response is the cross-mode oracle.

## Validation

- DFW H100 job `15516850` at `dc2e9a61ae6922faa9f3fbae9d73c172a17f0fe4`: Python compilation, Black, isort, Ruff, YAML parsing, and recipe discovery passed; the recipe expands to exactly one H100 product
- GitHub CI job `100709334870` first exercised the real launcher at aggregate commit `77e29726d74bba86d04d57986f0cca6d958f3f02` and exposed that the original converted checkpoint does not supply model arguments
- Aggregate GitHub CI run `33774739901` passed the complete unit phase, then SSE job `100732593457` exposed an unavailable tokenizer-cache path before server readiness
- Aggregate GitHub CI run `33782241164` passed the complete unit fanout, while SSE job `100790043543` proved that the mounted general Hugging Face cache did not contain the original remote tokenizer fixture
- In that GitHub CI environment, Qwen job `100732593042` successfully loaded `/mnt/artifacts/model/qwen3-8b-dist` and its local Hugging Face tokenizer
- Standalone DFW jobs `17729790` and `17730212` at aggregate commit `3bec18c2d40babb7769bea58c7a92f7c387d37c2` proved that both scheduling modes satisfy every HTTP/SSE invariant, while the former full-response oracle fails at different batch-sensitive values across runs
- DFW diagnostic job `17732029` at the same aggregate commit audited every field twice: streamed text, token IDs, finish state, and chunk sizes matched exactly, while cross-mode log-probability differences reached `0.0618`
- Exact source head `ac8a329b0645a25619c2ae9bde2979bfd2209612` narrows only the invalid cross-mode oracle and strengthens independent score validation; it changes no production code or workload
- Exact-head DFW job `17733210`: the standalone legacy-versus-async test passed twice (`148s` and `123s`) with all seven requests and all required mode and parity markers
- `git diff --check`, Python compilation, isort, Black, and Ruff pass on the exact head


tests/functional_tests/test_cases/gpt/gpt_inference_server_async_sched_tp2_pp2_dp2_qwen3_8b_sse/serve_sse.py
```diff
@@ -0,0 +1,361 @@
...
+from collections.abc import Callable
+from typing import Any
+
+READINESS_MARKER = "Running on http"
+READINESS_TIMEOUT_S = 600
+REQUEST_TIMEOUT_S = 180
+SHUTDOWN_TIMEOUT_S = 60
+SERVER_HOST = "0.0.0.0"
+SERVER_PORT = 5000
+STREAMING_INTERVAL = 2
```
nvcsathe, line 34: Is there any reason why the streaming interval is 2?

nvcsathe (approved):

kvareddy (approved):

tests/functional_tests/test_cases/gpt/gpt_inference_server_async_sched_tp2_pp2_ep2_gptoss_20b_sse/serve_sse.py
```diff
@@ -0,0 +1,361 @@
...
+from collections.abc import Callable
+from typing import Any
+
+READINESS_MARKER = "Running on http"
+READINESS_TIMEOUT_S = 600
+REQUEST_TIMEOUT_S = 180
+SHUTDOWN_TIMEOUT_S = 60
+SERVER_HOST = "0.0.0.0"
+SERVER_PORT = 5000
+STREAMING_INTERVAL = 2
```
lmcafee-nvidia: `2` is intentional: it is the smallest non-default interval, so the test exercises buffered multi-token SSE deltas rather than only the default one-token-per-event path. With 16 generated tokens, it still produces multiple deltas and verifies that their text, token IDs, and logprobs reconstruct the terminal response.

PR #6429, opened 2026-08-11
[Main] Recompute EP A2A overlap activations per layer segment (supersedes #5920)

Wohox:

> Main-branch counterpart of dev PR NVIDIA/Megatron-LM#6311.
>
> **This supersedes and fully covers NVIDIA/Megatron-LM#5920** (the main-branch
> port of the older, model-chunk-granularity design). #5920 can be closed in
> favour of this PR — everything it added to `main` is contained here, with the
> recompute unit changed from *VPP stage* to *layer segment*.

# What does this PR do ?

Adds full activation recompute support for the EP A2A overlap path
(`--overlap-moe-expert-parallel-comm`), which currently hard-asserts
`recompute_granularity=full` off.

Recompute is applied at **layer-segment** granularity: a segment's layers drop
their activations as soon as that segment has finished forwarding, and the
segment is replayed with grad enabled immediately before its own backward (whose
nodes release again as they backward). So only one segment's activations are
materialized at a time, and only each segment's input tensor survives the
forward→backward gap.

## Issue tracking

Linked issue: 

## Semantics

`recompute_method` / `recompute_num_layers` mean exactly the same thing with and
without `--overlap-moe-expert-parallel-comm`, mirroring
`megatron.core.recompute.checkpointed_forward` (decoder) and
`MultiTokenPredictionLayer._checkpointed_forward` (MTP):

* `uniform`: decoder layers in groups of `recompute_num_layers`; one segment per
  MTP layer (`recompute_num_layers` must be 1 when MTP is present, as upstream).
* `block`: the first `recompute_num_layers` decoder layers, one segment each;
  trailing decoder layers stay eager.

Both flags are therefore **required** on this path, exactly as on the non-overlap
path — no validation exemption is introduced. (#5920 did add such an exemption,
because it recomputed the whole model chunk as a single unit; that is gone here.)

## Design

* The initial forward of a recomputed layer's nodes runs under `no_grad`
  (`ScheduleNode.forward_no_grad`), so no autograd graph / saved activation is
  retained. `pre_process` / `post_process` keep their graphs.
* `RecomputeSegment` captures, at the head layer's forward: the segment input
  tensor, an RNG snapshot, and a snapshot of the mutable `ModelChunkState`
  fields (`input_ids` / `position_ids` / `padding_mask` / `mtp_hidden_states`) —
  these are mutated in place during the forward (MTP rolls the tokens), so by
  backward time they hold end-of-chunk values rather than what the segment saw.
* At the tail layer's backward, `RecomputeSegment.recompute()` replays the
  segment by re-invoking the **same** `ScheduleNode` objects in forward order, so
  each node repopulates its own `inputs` / `output` / `detached` / `layer_state`.
  The activation→node mapping is therefore implicit and needs no bookkeeping.
* `RecomputeSegment.release_activations()` unsets the segment's nodes'
  `inputs` / `output` / `detached` as soon as its tail layer has forwarded,
  rather than at the end of the chunk forward — otherwise each recomputed layer
  would keep one set of boundary tensors pinned for the remainder of the forward.
  Only the references are dropped; the storages are deliberately left intact,
  since the replay rebuilds the autograd graph on top of those very tensors and
  resizing them (the trick `ScheduleNode`'s `free_input` path and
  `tensor_parallel.random.checkpoint` use) corrupts it. On the way back,
  `ScheduleNode._release_state()` already frees each replayed node as it
  backwards.
* The A2A issued by a replay is **exposed** (not overlapped) by design; the
  normal forward and normal backward A2A overlap is fully preserved.

### Why a segment object rather than a `checkpoint()` wrapper

The non-overlap path can wrap a layer in `tensor_parallel.random.checkpoint`
because a layer is an opaque callable there. Under EP A2A overlap a layer is
decomposed into `ScheduleNode`s (`pre_dispatch_computation` / `moe_dispatch` /
`mlp` / `moe_combine` / `mtp_post_process`) that the 1F1B scheduler interleaves
with *other microbatches'* nodes — which is the whole point of the overlap path.
`checkpoint()` needs its region to run as one contiguous autograd call and to
backward as one contiguous unit, so the replay has to be driven by the scheduler
at segment-backward time instead.

`mtp_post_process` is inside the recompute scope: it builds a
`torch.cat(mtp_hidden_states)` over the MTP `pre_dispatch_computation` node's
detached input, so excluding it would anchor that `cat` on a `no_grad` leaf and
silently drop the decoder's gradient through MTP.

## Constraints (asserted)

Under `overlap_moe_expert_parallel_comm` + `recompute_granularity=full`:

* `distribute_saved_activations` must be off — the replay is hand-rolled rather
  than a checkpoint primitive, so the retained segment input is not TP-sharded.
* `attention_dropout == 0` and `hidden_dropout == 0` — the forward RNG stream is
  interleaved across microbatches in the 1F1B schedule and cannot yet be
  faithfully replayed for nonzero dropout.
* Delayed-scaling FP8 (`fp8` set and `fp8_recipe='delayed'`) is rejected — the
  amax history is persistent across microbatches, and this path does not enter
  TE's `recompute_phase` machinery that `te_checkpoint` uses to snapshot and
  restore FP8 metadata, so the replayed forward would update `amax_history` a
  second time and silently diverge. The current-scaling recipes (`tensorwise` /
  `mxfp8` / `blockwise`) re-derive their scales from the current tensor each
  pass, so the replay reproduces them exactly; those are the recipes the parity
  tests sweep. The assert is gated on `fp8` being set, because `fp8_recipe`
  defaults to `'delayed'` in BF16 configs where no FP8 autocast runs at all.
* Not yet supported (asserted off, left as follow-up): Megatron FSDP, CUDA graphs
  (`cuda_graph_impl != "none"`), `moe_paged_stash`, and
  `fine_grained_activation_offloading` (the offload manager pairs a
  forward-pushed group with the backward that pops it; the replay pushes its
  groups immediately before its own backward, which does not fit that pairing).

The last one is main-only — `dev` has no fine-grained activation offloading, so
it has no counterpart in #6311.

## Differences from the dev PR (#6311)

* Node naming: the attention node is `pre_dispatch_computation` on `main`
  (`attn` on `dev`), and `TransformerLayerNode` lives in
  `megatron/core/models/common/utils.py`.
* `main` has no multi-hyper-connection (mHC) support, so the `mhc_multistream`
  detach-bridge gradient carry-over from #6311 has no counterpart here and is
  omitted.
* `main`'s token dispatchers have no `reset_transient_forward_state()` (that
  arrived on `dev` via #5869), so the per-layer dispatcher routing metadata is
  not additionally freed across the gap. This is a memory refinement only —
  the replay repopulates it unconditionally — and is left as a follow-up.
* Adds the `fine_grained_activation_offloading` assert described above.
* Releases each segment's activations at its own tail-layer forward rather than
  at the end of the chunk forward, and adds the four-microbatch parity case.
  #6311 still does both at chunk granularity / one microbatch; it will be
  brought in line separately.

## Test results

Unit tests, 8×H100 (EOS), CI container (`docker/Dockerfile.ci.dev`, `--target
main`) via the same `tests/unit_tests/run_ci_test.sh` entry point CI uses:

`tests/unit_tests/a2a_overlap/test_schedule_chunk_1f1b.py` — **156 passed,
40 skipped, 0 failed** (9m35s).

Whole-directory regression (the change touches shared code: `ScheduleNode`, the
layer schedule plan), bucket `tests/unit_tests/a2a_overlap/**/*.py` —
**251 passed, 41 skipped, 2 deselected, 0 failed** (19m00s), covering
`test_schedule_chunk_1f1b.py`, `test_schedule_layer_1f1b.py`,
`test_delay_wgrad_compute.py` and `test_fsdp_1f1b_overlap.py`.

The new cases are
`TestA2AOverlap::test_1f1b_schedule_model_chunk_full_recompute` and
`..._full_recompute_multi_microbatch`, which compare gradients from
`TransformerModelChunkSchedulePlan.run` against the non-overlap
`gpt_model.forward` reference under the *same* recompute config, parametrized
over `mtp_layers` 0/1, `get_valid_dispatcher_configs()` and
`get_valid_fp8_flags()`. The first sweeps the full segmentation matrix
(`(uniform,1) (uniform,2) (block,1) (block,3)`) at one microbatch; the second
runs four microbatches over `(uniform,1)` and `(block,1)`, so the same modules
go through forward → recompute → backward repeatedly as in a real 1F1B steady
state and grads accumulate across the four rounds on both paths before the
bitwise compare. One microbatch cannot catch state the replay leaves behind for
the next round to read (FP8 scaling metadata, a segment's RNG snapshot, a node
whose `forward_no_grad` was not restored). The 40 skips are the
MTP × non-`(uniform,1)` combinations, where the non-overlap reference warns and
skips MTP recompute and the two paths are not comparable.

Beyond gradient parity the tests also assert the feature actually happened:
after each chunk forward, every recomputed node has `inputs is None` and
`output is None` while its segment's input tensor is retained. Recompute is
numerically transparent, so a segment that quietly kept its graph would still
compare equal.

The whole file carries `pytestmark = pytest.mark.flaky_in_dev`, so these runs
used `--environment lts` (marker `not flaky`) to actually execute them; under
`--environment dev` the file is deselected wholesale.

The two pre-existing tests in the file were refactored onto a shared
`run_two_chunk_parity()` helper (no behavioural change) so the new test reuses
the same 1F1B pattern.

For the algorithm's bitwise E2E evidence (DeepSeek-V3-Lite deterministic recipe:
BF16 / blockwise-FP8 / MXFP8, with and without MTP, overlap vs non-overlap
100/100 bitwise-identical) see the dev PR NVIDIA/Megatron-LM#6311, which
implements the same segmentation and replay.

## Benchmark: this recompute path against the non-overlap one

The question this section answers is narrow on purpose: **with full recompute
enabled, is the EP A2A overlap schedule still faster than the non-overlap path
it has to compete with?** The exposed replay A2A is a real cost, so it is worth
showing that it does not eat the overlap gain.

Three arms, all on this branch, 64×H100 (8 nodes). DeepSeek-V3 shape at 14
layers with the real 256-expert MoE and MTP 1, TP1 / PP4 / EP16, pipeline layout
`Ett|(tt|)*6mL`, alltoall dispatcher, seq 2048, MBS 1, GBS 256 (16 microbatches),
BF16, 30 iterations with the first 5 dropped as warmup. `recompute` means
`recompute_granularity=full`, `recompute_method=uniform`,
`recompute_num_layers=1`.

| | TFLOP/s/GPU (median) | iter (ms, mean) | peak allocated | peak reserved | device memory used |
|---|---|---|---|---|---|
| B. **this path** — overlap on, recompute on | **199.5** | 2839 | 60,378 MB | 68,392 MB | 74,803 MB |
| C. non-overlap baseline — overlap off, recompute on | 185.2 | 3044 | 60,154 MB | 62,570 MB | 68,973 MB |
| A. overlap on, recompute off (context) | 32.8 | 16985 | 70,279 MB | 75,594 MB | 79,677 MB |

### B vs C — the result this PR rests on

**+7.7 % throughput** (199.5 vs 185.2 TFLOP/s/GPU) at the same recompute
setting. The segment replay's exposed A2A does not cancel the overlap gain, so
enabling full recompute on the overlap path leaves it ahead of the non-overlap
path rather than behind it.

The two arms reach the same *allocated* peak (60,378 vs 60,154 MB, +0.4 %):
recompute drops the same activations either way, and the segment inputs the
replay retains are small next to what is dropped. That equality is also a useful
check on the feature — the segmentation is not quietly retaining more than the
non-overlap checkpoint path does. The overlap arm's cost shows up in *reserved*
memory (+5.8 GB), which is the schedule's own multi-stream working set rather
than anything recompute adds.

One methodological caveat: `CUDA_DEVICE_MAX_CONNECTIONS` is 32 on the overlap
arms and 1 on the non-overlap arm. `moe/README.md` lists `> 1` as a requirement
of EP A2A overlap, so running the overlap arm at 1 would measure a configuration
the feature does not support; 1 is the baseline recipe's value for everything
else. B vs C is therefore a best-configuration comparison rather than a
single-variable one.

### Arm A is context, not a result

Arm A is included only to show why recompute is on at all here. Without it this
configuration reaches 79.7 GB of device memory on the heaviest PP stage — 98 %
of the 80 GB card — and throughput collapses to a memory-bound 32 TFLOP/s/GPU,
flat across all 30 iterations. That reproduced exactly on a second run on a
disjoint node set (peak allocated identical to the MB), so it is the allocator
working against a full card rather than a fluke.

**The 6× gap between A and B is a memory-pressure effect and should not be read
as "recompute makes training 6× faster."** What it does establish is that at
this scale the configuration does not practically run without recompute.
Isolating the replay A2A's own cost would need a configuration where the
recompute-off arm fits comfortably; that measurement is not included here.

### Pre-checks
- [x] I have added relevant documentation (inline comments / docstrings)

### DIFF
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -3291,18 +3291,6 @@ class TransformerConfig(ModelParallelConfig):
                     'attention_dropout==0 and hidden_dropout==0 (RNG replay across the '
                     'interleaved 1F1B schedule is not yet supported for nonzero dropout)'
                 )
-                # CUDA graphs capture a fixed node sequence with static buffers, which
-                # the inserted no_grad forward + backward-time replay does not fit.
-                assert self.cuda_graph_impl == "none", (
-                    'overlap_moe_expert_parallel_comm full recompute is not yet supported '
-                    'together with CUDA graphs (cuda_graph_impl != "none").'
-                )
-                # Paged stash and full recompute are redundant activation-memory
-                # strategies, and stash/restore conflicts with the recompute release.
-                assert not self.moe_paged_stash, (
-                    'overlap_moe_expert_parallel_comm full recompute is not yet supported '
-                    'together with moe_paged_stash.'
-                )
                 # The offload manager pairs a forward-pushed group with the backward that
                 # pops it. Under full recompute the initial forward saves nothing and the
                 # replay pushes its groups immediately before its own backward, which does
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -3263,15 +3263,62 @@ class TransformerConfig(ModelParallelConfig):
                 'flex',
             ], 'overlap_moe_expert_parallel_comm is supported with alltoall/flex token dispatcher'
 
-            assert (
-                self.recompute_granularity != 'full'
-            ), 'disable full recomputation when enabling overlap_moe_expert_parallel_comm'
-            assert (
-                self.recompute_method is None
-            ), 'disable recomputation method when enabling overlap_moe_expert_parallel_comm'
-            assert (
-                self.recompute_num_layers is None
-            ), 'recompute_num_layers must be None when enabling overlap_moe_expert_parallel_comm'
+            if self.recompute_granularity == 'full':
+                # Full recompute runs per layer segment; recompute_method /
+                # recompute_num_layers keep their non-overlap meaning and are validated
+                # by the shared checks in __post_init__. See
+                # megatron/core/models/common/model_chunk_schedule_plan.py.
+                # The replay is hand-rolled rather than a checkpoint primitive, so it
+                # cannot shard the retained segment input across TP ranks.
+                assert not self.distribute_saved_activations, (
+                    'overlap_moe_expert_parallel_comm full recompute does not support '
+                    'distribute_saved_activations: the segment replay does not go through '
+                    'a checkpoint primitive, so the retained input is not sharded.'
+                )
+                # MTP is segmented under both methods here, so unlike
+                # MultiTokenPredictionLayer._checkpointed_forward there is no 'block'
+                # exemption. The 'uniform' restriction below is config parity with that
+                # function's own assert, not a constraint of this schedule - multi-layer
+                # decoder segments hand the MTP bridge over correctly.
+                if self.mtp_num_layers and self.recompute_method == 'uniform':
+                    assert (
+                        self.recompute_num_layers == 1
+                    ), 'recompute_num_layers must be 1 for MTP recompute'
+                # With dropout the forward RNG stream is interleaved across microbatches
+                # in the 1F1B schedule and cannot yet be faithfully replayed.
+                assert self.attention_dropout == 0.0 and self.hidden_dropout == 0.0, (
+                    'overlap_moe_expert_parallel_comm full recompute currently requires '
+                    'attention_dropout==0 and hidden_dropout==0 (RNG replay across the '
+                    'interleaved 1F1B schedule is not yet supported for nonzero dropout)'
+                )
+                # CUDA graphs capture a fixed node sequence with static buffers, which
+                # the inserted no_grad forward + backward-time replay does not fit.
+                assert self.cuda_graph_impl == "none", (
+                    'overlap_moe_expert_parallel_comm full recompute is not yet supported '
+                    'together with CUDA graphs (cuda_graph_impl != "none").'
+                )
+                # Paged stash and full recompute are redundant activation-memory
+                # strategies, and stash/restore conflicts with the recompute release.
+                assert not self.moe_paged_stash, (
+                    'overlap_moe_expert_parallel_comm full recompute is not yet supported '
+                    'together with moe_paged_stash.'
+                )
+                # The offload manager pairs a forward-pushed group with the backward that
+                # pops it. Under full recompute the initial forward saves nothing and the
+                # replay pushes its groups immediately before its own backward, which does
+                # not fit that pairing.
+                assert not self.fine_grained_activation_offloading, (
+                    'overlap_moe_expert_parallel_comm full recompute is not yet supported '
+                    'together with fine_grained_activation_offloading.'
+                )
+            else:
+                assert self.recompute_method is None, (
+                    'disable recomputation method when enabling ' 'overlap_moe_expert_parallel_comm'
+                )
+                assert self.recompute_num_layers is None, (
+                    'recompute_num_layers must be None when enabling '
+                    'overlap_moe_expert_parallel_comm'
+                )
             assert (
                 "moe" not in self.recompute_modules
             ), 'disable moe in recompute_modules when enabling overlap_moe_expert_parallel_comm'
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -3291,6 +3291,23 @@ class TransformerConfig(ModelParallelConfig):
                     'attention_dropout==0 and hidden_dropout==0 (RNG replay across the '
                     'interleaved 1F1B schedule is not yet supported for nonzero dropout)'
                 )
+                # Delayed-scaling FP8 accumulates amax into a persistent amax_history
+                # across microbatches. Both the initial no_grad forward and the
+                # backward-time recompute run under the delayed FP8 autocast, and this
+                # path does not enter TE's activation-recompute (recompute_phase)
+                # machinery that te_checkpoint uses to snapshot and restore FP8 metadata,
+                # so the recompute would update amax_history a second time and silently
+                # diverge. Current-scaling recipes (tensorwise / mxfp8 / blockwise) derive
+                # scales from the current tensor each pass, so the replayed forward
+                # re-derives identical scales; only delayed scaling is rejected here.
+                # self.fp8 is checked because fp8_recipe defaults to 'delayed' in BF16
+                # configs, where no FP8 autocast runs at all.
+                assert not (self.fp8 and self.fp8_recipe == Fp8Recipe.delayed), (
+                    'overlap_moe_expert_parallel_comm full recompute is not yet supported '
+                    'with delayed-scaling FP8 (fp8_recipe="delayed"): the backward-time '
+                    'recompute would double-update the persistent amax_history. Use a '
+                    'current-scaling recipe (tensorwise / mxfp8 / blockwise) instead.'
+                )
                 # The offload manager pairs a forward-pushed group with the backward that
                 # pops it. Under full recompute the initial forward saves nothing and the
                 # replay pushes its groups immediately before its own backward, which does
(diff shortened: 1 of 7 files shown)


megatron/core/transformer/transformer_config.py
```diff
@@ -2750,15 +2750,62 @@ def _scope_to_str(s):
...
+                # With dropout the forward RNG stream is interleaved across microbatches
+                # in the 1F1B schedule and cannot yet be faithfully replayed.
+                assert self.attention_dropout == 0.0 and self.hidden_dropout == 0.0, (
+                    'overlap_moe_expert_parallel_comm full recompute currently requires '
+                    'attention_dropout==0 and hidden_dropout==0 (RNG replay across the '
+                    'interleaved 1F1B schedule is not yet supported for nonzero dropout)'
+                )
+                # CUDA graphs capture a fixed node sequence with static buffers, which
+                # the inserted no_grad forward + backward-time replay does not fit.
+                assert self.cuda_graph_impl == "none", (
```
zhongbozhu: can you share why this is not playing well with cuda graph?

megatron/core/transformer/transformer_config.py
```diff
@@ -2750,15 +2750,62 @@ def _scope_to_str(s):
...
+                    'overlap_moe_expert_parallel_comm full recompute is not yet supported '
+                    'together with CUDA graphs (cuda_graph_impl != "none").'
+                )
+                # Paged stash and full recompute are redundant activation-memory
+                # strategies, and stash/restore conflicts with the recompute release.
+                assert not self.moe_paged_stash, (
```
zhongbozhu: same comment as above

megatron/core/transformer/transformer_config.py
```diff
@@ -2750,15 +2750,62 @@ def _scope_to_str(s):
...
+                )
+                # The offload manager pairs a forward-pushed group with the backward that
+                # pops it. Under full recompute the initial forward saves nothing and the
+                # replay pushes its groups immediately before its own backward, which does
+                # not fit that pairing.
+                assert not self.fine_grained_activation_offloading, (
+                    'overlap_moe_expert_parallel_comm full recompute is not yet supported '
+                    'together with fine_grained_activation_offloading.'
+                )
+            else:
```
zhongbozhu, line 3293: can you explain the asserts under this else?

Wohox: Recompute method and recompute_num_layers shouldn't be assigned when disabling full recomupte.

megatron/core/transformer/transformer_config.py
Wohox: Good catch, this is a slop, current implementation should work with full-iteration cuda graph + paged stash, but doesn't work with partial cuda graph(same as EP overlap=False case).

Removed this assertion in 3ffbe326e

megatron/core/transformer/transformer_config.py
Wohox: Removed in 3ffbe326e. same reason as above~

zhongbozhu:
https://github.com/NVIDIA/Megatron-LM/commit/74dc1b570330d430d3c2a0b87dba2cefd7e630b8 changed API of 1f1b, please rebase accordingly, thanks~

zhongbozhu (approved):

Phlip79 (approved):

Issue #46, opened 2020-10-02
Malicious domain name in openwebtext URL list


abhinavs95:
I was following the instructions for preparing openwebtext dataset using instructions here: https://github.com/NVIDIA/Megatron-LM/blob/main/tools/openwebtext/README.md

In the URL list downloaded from the link in Step 1 of Download the dataset, one of the domain names ("horsefucker.org") is associated with a known C&C server. This caused security vulnerability on my system. The blacklist_urls list must be updated with this domain name so that its filtered before data download begins.

jon-barker:
Will update blacklist

jon-barker:
Fixed: https://github.com/NVIDIA/Megatron-LM/commit/780978175455afadc78eceea2c8f63ee4fceec7b - thanks for highlighting this

Issue #45, opened 2020-09-26
Fully half precision optimizer


eric-haibin-lin:
Hi,

Has anyone implemented/tried fully half-precision optimizer with Megatron? I see from the GPT-3 paper:

> ... implemented an early version of the codebase, and developed the memory optimizations for fully
half-precision training.

It looks like Megatron FP16_Optimizer is still using mixed precision. Has anyone looked into this before? This would allow training these big models with much less memory ... Thanks!

Issue #44, opened 2020-09-25
Incorporating Megatron-ML with DeepSpeed


msmolyak:
Microsoft incorporated Megatron-LM with their DeepSpeed project in [DeepSpeedExamples](https://github.com/microsoft/DeepSpeedExamples) project. The combination of the two projects features increased speed and lower memory requirements compared to standalone Megatron-LM model. However, the version of Megatron-LM used in DeepSpeedExamples dates to February of 2020 and as such lacks the latest updates of your product. Have you considered supporting an up-to-date version of Megatron-LM integrated with DeepSpeed project?

jon-barker:
Closing. This issue should be addressed to the DeepSpeed developers at Microsoft: https://github.com/microsoft/DeepSpeed/

Issue #43, opened 2020-09-21
Pretrain and generate


king-menin:
if i run with
`python -m torch.distributed.launch --nproc_per_node 16 pretrain_gpt2.py --model_parallel_size==16`
and after run generate:
`python generate_samples.py`
 i have error while initialization: size mismatch for transformer.layers.15.attention.dense.weight: copying a param with shape torch.Size([1024, 64]) from checkpoint, the shape in current model is torch.Size([1024, 1024]).
can i load model on one GPU and train distributed on 16 GPU with --model_parallel_size==16?
thank you!

PR #6430, opened 2026-08-11
Add async scheduling pairwise unit coverage

lmcafee-nvidia:
## What does this PR do?

This PR adds a runtime-backed unit-test foundation for asynchronous dynamic inference scheduling. It compares the same real-engine workload under the legacy scheduler and the asynchronous scheduler across 27 selected companion configurations covering admission pressure, caching, CUDA graphs, sampling, request behavior, model implementations, attention modes, and distributed topologies.

Every matrix row runs the companion configuration twice:

- baseline: `async_sched_mode=LEGACY`
- treatment: `async_sched_mode=ASYNC`

All other flags, requests, seeds, and model settings stay the same. For example, the prefix-cache row first serves a shared-prefix workload with legacy scheduling and then serves the identical workload with async scheduling; both runs have prefix caching enabled. The test checks that changing only the scheduler preserves the required results while runtime counters prove that both prefix reuse and the async scheduling phases actually executed.

This is a targeted pairwise matrix, not a universal Boolean 2x2 over every flag. A row tests the scheduler choice against one deliberately active companion configuration; it does not normally repeat a companion-disabled cell.

The matrix starts from the supported inference settings whose execution, request-local state, cache ownership, batch shape, or distributed communication can interact with scheduling. The selected surface includes:

- engine behavior: request/token capacity, chunked prefill, prefix caching, KV residency, CUDA graphs, and token events;
- request behavior: sampling filters, logprob modes, output-length controls, BOS insertion, termination IDs, stop sequences, and speculative decoding;
- model and kernel behavior: dense GPT, hybrid Mamba, Transformer Engine, inference-optimized layers, BF16/FP8, sliding-window attention, and softmax sinks;
- parallel behavior: tensor, pipeline, sequence, data, and expert parallelism plus NCCL expert dispatch.

A configuration counts only when the unit test can force the production path to run, record a feature-owned runtime witness, and compare scheduler behavior or a precise invariant. Merely setting a flag, inspecting a configured object, or observing generic model execution does not count. Related settings are grouped into one diagnostic owner row when they naturally execute together; configurations requiring special GPU topology have separate rows.

The companion-disabled cells are omitted because every row already has a legacy-scheduler control with the same companion active. Repeating each row with the companion disabled would test the scheduler alone again without proving an additional interaction.

Focused regressions that depend on production fixes remain in their dependent PRs: fused RoPE is in fork #45, and raw-prompt top-N plus heterogeneous MTP/logprob coverage is in fork #46. Direct pending-logit lifecycle and metadata-compaction invariants are in fork #43 and fork #44. GPTOSS SSE remains the separately approved end-to-end exception in NVIDIA/Megatron-LM#6428.

## What combinations are tested?

Each row below is the two-run `LEGACY` versus `ASYNC` scheduler comparison described above. The listed companion flags remain enabled in both runs.

### Core engine and cache behavior

| # | Scenario | Active companion configuration |
| ---: | --- | --- |
| 1 | Dense eager baseline | GPT, BF16, Torch sampling with `top_k=1`, PERSIST KV mode, static KV pointers, generated-token events, last-token logits, and no CUDA graphs |
| 2 | Chunked prefill under capacity | `enable_chunked_prefill=True`, `context_max_requests=4`, and `context_max_tokens=8` |
| 3 | Prefix cache with REF_ZERO | `enable_prefix_caching=True`, REF_ZERO eviction, 256-token blocks, 512-token shared prefixes, `context_max_tokens=768`, and `max_sequence_length=544` |
| 4 | Prefix cache with LRU and chunking | Prefix caching plus chunked prefill, LRU eviction, 256-token blocks, 512-token shared prefixes, `context_max_tokens=384`, and `max_sequence_length=544` |

### CUDA-graph behavior

| # | Scenario | Active companion configuration |
| ---: | --- | --- |
| 5 | Decode-only block graphs | Four forced graphs, block scope, `use_cuda_graphs_for_non_decode_steps=False`, exponential sizing, and `cuda_graph_max_tokens=16` |
| 6 | Bounded mixed-prefill graphs | Four forced graphs, `cuda_graph_mixed_prefill_count=2`, and `cuda_graph_max_tokens=16` |
| 7 | Layer-scope all-prefill graphs | Four forced graphs, layer scope, `cuda_graph_all_prefills=True`, mixed-prefill count 2, linear sizing, and `cuda_graph_max_tokens=24` |

### Sampling and logprobs

| # | Scenario | Active companion configuration |
| ---: | --- | --- |
| 8 | Torch temperature and top-k | Torch sampling, `temperature=0.7`, and `top_k=8` |
| 9 | Torch top-p | Torch sampling, `temperature=1.1`, `top_k=0`, and `top_p=0.85` |
| 10 | FlashInfer top-k plus top-p | FlashInfer sampling, `temperature=0.8`, `top_k=12`, and `top_p=0.9` |
| 11 | Processed generated logprobs | `return_log_probs=True`, `logprobs_mode=processed_logprobs`, `skip_prompt_log_probs=True`, and generated `top_n_logprobs=3` |
| 12 | FlashInfer without truncation | FlashInfer sampling, `temperature=0.8`, `top_k=0`, and `top_p=0.0` |

### Request behavior and speculation

| # | Scenario | Active companion configuration |
| ---: | --- | --- |
| 13 | Mixed output-length controls | One zero-output request, one `num_tokens_total` request, and ordinary `num_tokens_to_generate` requests |
| 14 | String prompts with BOS | String tokenization with `add_BOS=True` |
| 15 | Mixed termination values | One request uses a token first observed from deterministic generation as its termination ID; the others keep termination disabled |
| 16 | Retained stop sequence | One request uses a two-token stop sequence with `detokenize_stop_sequence=True` |
| 17 | Stripped stop sequence | One request uses a two-token stop sequence with `detokenize_stop_sequence=False` |
| 18 | MTP depth one | `num_speculative_tokens=1` |

### Model, precision, implementation, and attention

| # | Scenario | Active companion configuration |
| ---: | --- | --- |
| 19 | Hybrid Mamba | Hybrid model provider with live Mamba-state compaction |
| 20 | Transformer Engine | `transformer_impl=transformer_engine` |
| 21 | Inference-optimized layers | `transformer_impl=inference_optimized` |
| 22 | FP8 | `fp8=True`, with the FP8 production path required to execute |
| 23 | All-layer SWA with off-by-one sink | `window_size=(4,0)` and `softmax_type=off-by-one` |
| 24 | Alternating SWA with learnable sink | `window_size=(4,0)`, `window_attn_skip_freq=2`, and `softmax_type=learnable` |

### Distributed topology

| # | Scenario | Active companion configuration |
| ---: | --- | --- |
| 25 | TP2 + PP2 + SP + DP2 | `tensor_model_parallel_size=2`, `pipeline_model_parallel_size=2`, sequence parallelism enabled, DP-rank seed offsets enabled, `temperature=0.8`, and `top_k=8`; requires exactly eight GPUs |
| 26 | MoE + EP2 + NCCL dispatch | `expert_model_parallel_size=2`, NCCL inference token dispatch, and the inference-optimized implementation; requires at least four GPUs |
| 27 | Inference-optimized TP2 + SP | TP2, sequence parallelism, inference-optimized layers, shared sampling seeds across DP ranks, `temperature=0.8`, and `top_k=8`; requires at least four GPUs |

## How are results checked?

The common workload contains seven requests with prompt lengths `(9, 5, 12, 7, 10, 6, 11)` and output lengths `(10, 5, 9, 7, 11, 6, 8)`. They arrive in three waves, `(0,1,2)`, `(3,4)`, and `(5,6)`, with capacity for only four active requests. This forces waiting, arrival during decoding, completion of a middle request while others survive, metadata compaction, both async phase shapes (overlap and no-overlap), and a complete engine drain. The dense eager/PERSIST row also performs an actual `suspend()`/`resume()` cycle and verifies that valid pending async logits and the static KV-memory pointer are both retained across it.

For deterministic rows, the async result must match the legacy result in completion status, output length, exact generated tokens, finite logprobs within the configured tolerance, and top-N content where requested. Event rows also compare event sequences. For stochastic sampling and FP8 rows, cross-scheduler checks require valid completion, shapes, lengths, token ranges, and finite values; a second ASYNC run must then reproduce the first ASYNC run exactly.

Every row also checks the shared async runtime contract and its own feature-specific witness. Examples include nonzero prefix hits and skipped prefill tokens, executed CUDA-graph dimensions, FlashInfer sampling calls, Mamba metadata compaction, SWA and sink kernels, FP8 forwards, TP/SP collectives, pipeline logits broadcasts, or matched NCCL expert dispatch/combine calls. A configured feature that never reaches its owning production boundary fails the test instead of receiving coverage credit.

Four negative guards verify that unsupported routing replay, mismatched MTP depth, processed logprobs with speculative decoding, and skipped async bookkeeping are rejected explicitly. They document unsupported boundaries and are not counted among the 27 positive pairwise rows.

## Scope

This is the unit-only shared foundation. Production fixes and the regressions that expose them are split into dependent follow-up PRs.

Relative to `main`, it changes only:

- `tests/unit_tests/inference/engines/test_dynamic_engine.py` (+34/-4)
- `tests/unit_tests/inference/engines/test_dynamic_engine_async_sched.py` (+1577/-5)

The net unit-test growth is +1,602 lines. It adds no production, functional-test, or recipe changes.

## Validation

Exact signed tip: `1668169d2607f666894b088f288ad48f6e70600e`

- DFW Slurm job `17644824` at executable predecessor `55ad007efcad8f2055b4151ef6324bd70b1c048e`: passed
  - exact 8-rank suite: 94 passed, 175 deselected, 0 failed/skipped on each rank
  - pytest runtime: 135.65–135.94 seconds across ranks
  - complete `uv`/torchrun wall time: 189.391 seconds
  - static checks passed: `git diff --check`, isort, Black, Ruff, and Python compilation
- The only subsequent change documents why the test helper forwards `num_moe_experts`; the same static checks passed at the current tip


tests/unit_tests/inference/engines/test_dynamic_engine.py
```diff
@@ -295,38 +298,42 @@ def _build_inference_context(
...
+            sampling_backend=test_config.sampling_backend,
+            async_sched_mode=test_config.async_sched_mode,
+            logprobs_mode=test_config.logprobs_mode,
+        )
+        unknown_overrides = set(test_config.inference_config_overrides) - set(
+            InferenceConfig.__annotations__
+        )
+        if unknown_overrides:
+            raise ValueError(f"Unknown InferenceConfig overrides: {sorted(unknown_overrides)}")
+        inference_config_kwargs.update(test_config.inference_config_overrides)
```
santhnm2: Why are these changes necessary?

tests/unit_tests/inference/engines/test_dynamic_engine.py
```diff
@@ -363,7 +370,11 @@ def _build_test_env(cls, test_config):
                 params_dtype=torch.bfloat16,
                 num_layers=4,
                 mtp_num_layers=test_config.num_speculative_tokens,
-                hidden_size=128 if test_config.fp8 else 32,
+                hidden_size=(
+                    test_config.hidden_size
+                    if test_config.hidden_size is not None
```
santhnm2: Why would the hidden size ever not be provided?

tests/unit_tests/inference/engines/test_dynamic_engine.py
```diff
@@ -405,11 +428,17 @@ def _build_test_env(cls, test_config):
                 window_attn_skip_freq=test_config.window_attn_skip_freq,
             )
             if test_config.fp8 or test_config.transformer_impl == "transformer_engine":
-                layer_spec = get_gpt_layer_with_transformer_engine_spec()
+                layer_spec = get_gpt_layer_with_transformer_engine_spec(
+                    num_experts=transformer_config.num_moe_experts
```
santhnm2, line 664: Does this need to be specified explicitly here? Can we just set it in `get_gpt_layer_with_transformer_engine_spec`?

tests/unit_tests/inference/engines/test_dynamic_engine_async_sched.py
santhnm2, line 1: This appears to be comprehensive but what is the overhead of running such a large test suite? Is this realistic for a per-PR unit test suite?

kvareddy (approved):

tests/unit_tests/inference/engines/test_dynamic_engine.py
lmcafee-nvidia: You're right—the generic override dictionary was unnecessary. I removed it and restored the direct `InferenceConfig(...)` construction. The four matrix-specific settings are now explicit typed test-config fields, so this no longer needs custom unknown-key validation.

tests/unit_tests/inference/engines/test_dynamic_engine.py
lmcafee-nvidia: It should not be optional in this shared harness. I removed `hidden_size` from this PR and restored the existing `128 if fp8 else 32` behavior. The fused-RoPE child PR, which is the only case needing another size, will own an explicit non-optional field.

tests/unit_tests/inference/engines/test_dynamic_engine.py
lmcafee-nvidia: Good question. `get_gpt_layer_with_transformer_engine_spec` does not have access to `TransformerConfig`; it only receives explicit arguments. If `num_experts` is omitted, it defaults to `None`, which definitively constructs a dense MLP even when `num_moe_experts=2` and `expert_model_parallel_size=2` are set in the configuration. Previously, those settings initialized expert-parallel infrastructure, but the GPT model did not contain MoE layers or execute expert routing. Forwarding the configured expert count makes the layer specification match the model configuration, and the test verifies actual MoE execution through balanced dispatch/combine calls. Or did you have something else in mind that I might be missing?

tests/unit_tests/inference/engines/test_dynamic_engine_async_sched.py
lmcafee-nvidia: I measured the exact eight-rank selection at the current signed tip. Every rank passed 94 tests with 175 deselected and no failures or skips. Pytest took 135.65–135.94 seconds (about 2 minutes 16 seconds); the complete `uv`/torchrun wrapper took 189.391 seconds, and the slowest individual case took 8.23 seconds. The 27 positive rows each perform their LEGACY/ASYNC comparison inside the unit suite, without adding a functional-test job. At this measured runtime, I believe it is realistic for per-PR testing. I also added these timings to the Validation section.

tests/unit_tests/inference/engines/test_dynamic_engine.py
santhnm2: This makes sense, could add a comment documenting the reasoning?

santhnm2 (approved):

tests/unit_tests/inference/engines/test_dynamic_engine.py
lmcafee-nvidia: Added immediately before the layer-spec selection in `1668169d`. It now documents that the factories do not receive `TransformerConfig`, so `num_moe_experts` must be forwarded explicitly to make the MoE test configurations construct MoE layers instead of a dense MLP.

PR #6436, opened 2026-08-11
ci(auth): treat svcnemo-autobot as internal

ko3n1g:
## Background / motivation

- svcnemo-autobot is an internal service account but cannot enroll in NVIDIA GitHub SSO.
- The SSO-only check classifies its fork PRs as external.

## What changed

- Add svcnemo-autobot to the internal user allowlist.
- Document that is_member includes allowlisted internal accounts.

## Details

- .github/actions/check-nvidia-sso-membership/action.yml returns is_member=true, is_org_member=false, and an empty user_orgs value for svcnemo-autobot.
- Preserve SSO membership behavior for every other username.

## Tested

- git diff --check — passed.
- yq metadata assertion — composite action YAML parsed and expected check-membership step was present.
- Bash behavioral harness extracted the check-membership script with yq — allowlisted service account, SSO member, and external user assertions passed.
- actionlint — repository-wide check reports existing failures in unchanged workflow files; this PR does not modify .github/workflows.

### DIFF
--- a/.github/actions/check-nvidia-sso-membership/action.yml
+++ b/.github/actions/check-nvidia-sso-membership/action.yml
@@ -1,5 +1,5 @@
 name: 'Check NVIDIA SSO Membership'
-description: 'Check if a GitHub username exists in the NVIDIA SSO users list from github-audits'
+description: 'Check if a GitHub username is in the NVIDIA SSO users list or internal allowlist'
 author: 'NVIDIA'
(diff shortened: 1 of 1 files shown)


ko3n1g:
@svcnemo-autobot cherrypick this PR to MLM dev

thomasdhc (approved):

PR #5678, opened 2026-07-06
Remove torch cuda empty cache

shurkat-nvidia:
- [x] I, the PR author, have personally reviewed every line of this PR. (I have written every line of this PR.)

# What does this PR do?

The PR removes an expensive `torch.cuda.empty_cache()` call within `save_checkpoint_and_time()`. This call took multiple seconds on every checkpoint and also increased the size of the `generate_state_dict()` function by multiple seconds. To avoid OOM errors, the code also modifies the Nemotron3 Super script to 1) use SHM to avoid expensive `rebuild_cuda_tensor()` calls during the IPC handoff and 2) garbage collect the torch allocator at 80% GPU allocation with a lower overhead than a raw `empty_cache()` call, and to use larger blocks to prevent fragmentation in the first place.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact @NVIDIA/mcore-oncall.

## Issue tracking

For PRs from open-source community contributors:

- **New features**: a linked issue is **required**. Please open a [feature request](https://github.com/NVIDIA/Megatron-LM/issues/new?template=feature_request.md) and reference it here before submitting the PR.
- **Small updates (bug fixes, minor improvements)**: a linked issue is **recommended** and will accelerate the PR review process.

Linked issue: 

### Pre-checks

- [ ] (N/A) I have added relevant unit tests
- [ ] (N/A) I have added relevant functional tests
- [ ] (N/A) I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] (N/A) I have added relevant documentation
- [x] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment @NVIDIA/mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

[...]

PR #6438, opened 2026-08-11
Instrument Megatron-LM with OpenTelemetry traces and metrics

deepakn94:
## What does this PR do?

Wires the span groups and metric instruments added in #6415 into the code that produces them, and adds the configuration surface to turn it all on.

**This replaces #6385**, rebased onto current main now that #6415 has merged. Original changes by @rhewett-nv — please close #6385 in favour of this one.

### What's instrumented

| Subsystem | File |
|---|---|
| Training loop, model init, evaluation | `megatron/training/training.py` |
| Checkpoint save/load, async checkpointing | `megatron/training/checkpointing.py`, `async_utils.py` |
| Pipeline schedules, microbatch forward/backward | `megatron/core/pipeline_parallel/schedules.py` |
| P2P send/recv | `megatron/core/pipeline_parallel/p2p_communication.py` |
| Gradient sync | `megatron/core/distributed/distributed_data_parallel.py` |
| Activation offload/reload | `megatron/core/pipeline_parallel/fine_grained_activation_offload.py` |
| Transformer and Mamba layers | `megatron/core/transformer/transformer_layer.py`, `megatron/core/ssm/mamba_layer.py` |
| Dataset builder | `megatron/core/datasets/blended_megatron_dataset_builder.py` |

Adds `--otel-enabled`, `--otel-service-name`, `--otel-span-groups`, the `MEGATRON_OTEL_*` environment variables, and telemetry setup/teardown in `megatron/training/global_vars.py`. Publishes the Observability guide by adding its `docs/index.md` toctree entry.

### Cost when disabled

Spans are gated per group: a disabled group costs a frozenset lookup and no span object is allocated. Non-exporting ranks resolve to an empty group set. Every `nemo.lens` import falls back to the no-ops in `megatron.core.telemetry.fallbacks`, so nothing here requires nemo-lens to be installed.

## Changes since the first review

Thanks @ahmadki and @santhnm2 — all four review threads are addressed.

- **Inference-server instrumentation reverted.** `megatron/core/inference/text_generation_server/text_generation_server.py` is byte-identical to main again; this PR no longer touches that module. Per @santhnm2 it is effectively deprecated, so the GenAI-semconv spans and metrics will be ported to `dynamic_text_gen_server/text_generation_server.py` as a follow-up rather than reworked here.
- **The nemo.lens test suite is gone.** Removed all seven files (`conftest.py`, `test_config`, `test_handle`, `test_helpers`, `test_instrumentation`, `test_instruments`, `test_propagation`). Per @ahmadki they test the Lens library rather than Megatron and belong upstream. They were inherited verbatim from #6385. The four telemetry test files already on main from #6415 are untouched and still pass.
- **`RandomIdGenerator` monkeypatch removed** from `_set_telemetry`. It reached into `opentelemetry.sdk` process-globally to work around identically-seeded span IDs across data-parallel ranks; nemo-lens handles that in its own provider setup.
- **nemo-lens pin bumped** to `b85578fc` in both the `dev` group and the `otel` extra.
- **Rebased onto main** across 68 commits. Two conflicts: `distributed_data_parallel.py` (kept both main's `PARAM_READY_CALLBACK_ATTR` import and the lens `trace_fn` fallback) and `training.py`, where main had removed `torch.cuda.empty_cache()` from `save_checkpoint_and_time` in #5678 — adopted that removal rather than resurrecting the call.
- **Ran `tools/autoformat.sh`**, which #6385 never had; black reformatted 9 files and isort fixed 6.

## :warning: `opentelemetry-api` pin widened — needs a look

`uv.lock` had to be regenerated for the new dependency, and doing so surfaced an **unsatisfiable conflict that exists on main today**:

- `dev` pinned `opentelemetry-api~=1.33.1`, capping OpenTelemetry at 1.33.x.
- That forces `opentelemetry-exporter-otlp-proto-grpc <=1.34.1`, which needs `protobuf>=3.19,<6.0`.
- `nvidia-resiliency-ext==0.6.0` needs `grpcio-tools>=1.76.0`, which needs `protobuf>=6.31.1`.
- Irreconcilable, so `megatron-core[dev]` and `megatron-core[otel]` could not resolve together.

nemo-lens is not the constraint; it asks only for `opentelemetry-api>=1.20.0`. The cap was. The conflict is invisible on main only because nothing currently pulls the `otel` extra.

Fix: `opentelemetry-api~=1.33.1` -> `>=1.33.1,<2`. **This moves every dev environment from OpenTelemetry 1.33.1 to 1.44.0**, not just telemetry runs, which is the main thing worth reviewing here.

Two notes on the lockfile:

- Generated with **uv 0.7.2**, the version CI pins in both `setup-uv` (`cicd-main.yml:396`) and `Dockerfile.ci.dev`. Verified in the CI container with `uv lock --check` (exit 0) and `uv sync --locked --only-group linting` (exit 0), which is what the linting job runs.
- The diff shows `revision = 3` -> `revision = 2`. Main's lock was generated by a newer uv than CI pins, so 0.7.2 rewrites it in its own format. Pre-existing inconsistency, not introduced here.

## Issue tracking

Linked issue: replaces #6385

### Pre-checks

### Verified locally

- Sphinx builds clean with `--fail-on-warning` (the flag the docs job uses), including the newly published Observability toctree.
- black, isort, ruff, and pylint (10.00/10) pass on every changed file in `megatron/core`.
- All changed Python files compile.
- The 83 telemetry unit tests from #6415 still pass against this branch.

### Known follow-ups

**Extracting the span-lifecycle helpers out of `training.py` — immediate follow-up, already planned.**

This PR adds 1,281 lines to `megatron/training/training.py` (now 5,603 lines), and a large share of that is telemetry bookkeeping rather than training logic:

- 11 span-lifecycle functions in a contiguous block (`_start_otel_job_spans`, `_end_otel_startup_span`, `_start_otel_train_span`, `_reroot_otel_interval`, `_maybe_reroot_otel_interval`, `_end_otel_interval_span`, `_end_otel_train_span`, `_end_otel_job_spans`, `_otel_telemetry_active`, `_otel_mark_goodput`, `_backdated_otel_span`)
- 9 module-level `_otel_*` globals holding span handles, context tokens and idempotence flags
- the SIGTERM handler, force-flush, graceful-drain and exit-hook installer

Roughly 400-500 lines will move to a dedicated module, leaving `training.py` with imports plus the call-site usage that should have been the only footprint all along: 21 `with _otel_managed_span(...)`, 6 `_otel_sg_enabled(...)`, and a handful of attribute stamps.

Deliberately **not** done in this PR: it would rewrite the exact `training.py` diff reviewers have already read, and the span lifecycle has no unit-test coverage here (the four telemetry tests on main cover `megatron/core/telemetry`, not this), so a move touching `global` mutation ordering and atexit/SIGTERM idempotence would be unguarded. As a standalone follow-up the diff is almost entirely a move, reviewable with `git diff -M`, and can land with tests for the lifecycle.

**Also queued:** port the GenAI-semconv inference instrumentation to `dynamic_text_gen_server/`, and restore the inference sections of the Observability guide alongside it (they were removed here in f007fd2, since this PR no longer ships that instrumentation).


yashaswikarnati (approved):

Phlip79 (approved):

Phlip79 (approved):

yashaswikarnati (approved):

megatron/core/inference/text_generation_server/text_generation_server.py
santhnm2, line 1: This file is pretty much deprecated, changes should go in `megatron/core/inference/text_generation_server/dynamic_text_gen_server/text_generation_server.py`

pyproject.toml
```diff
@@ -97,6 +97,7 @@ dev = [
     "tensorstore~=0.1,!=0.1.46,!=0.1.72",
     "multi-storage-client~=0.50",
     "opentelemetry-api~=1.33.1",
+    "nemo-lens @ git+https://github.com/NVIDIA-NeMo/Lens.git@ecb0d15a93a49c45cb7189e792a91087f8fe1ace",
```
ahmadki:
```suggestion
    "nemo-lens @ git+https://github.com/NVIDIA-NeMo/Lens.git@b85578fc2b736a1804705e537001b5f45e9c715d",
```

pyproject.toml
```diff
@@ -115,6 +116,10 @@ dev = [
     "zstandard",
 ]
 
+otel = [
+    "nemo-lens[sdk] @ git+https://github.com/NVIDIA-NeMo/Lens.git@ecb0d15a93a49c45cb7189e792a91087f8fe1ace",
```
ahmadki:
```suggestion
    "nemo-lens[sdk] @ git+https://github.com/NVIDIA-NeMo/Lens.git@b85578fc2b736a1804705e537001b5f45e9c715d",
```

megatron/training/global_vars.py
```diff
@@ -323,6 +347,213 @@ def _ensure_var_is_not_initialized(var, name):
...
+    if not os.environ.get('OTEL_SERVICE_NAME', '').strip():
+        config.service_name = 'megatron-lm'
+    if getattr(args, 'otel_enabled', False):
+        config.enabled = True
+    if getattr(args, 'otel_service_name', None):
+        config.service_name = args.otel_service_name
+    if getattr(args, 'otel_span_groups', None):
+        config.span_groups = args.otel_span_groups
+
+    if config.enabled:
```
ahmadki: `RandomIdGenerator` is now integrated with Lens, the entire block should go

ahmadki (changes requested):

ahmadki (changes requested):
The entire test suit should be dropped, these tests belong to Lens - some of them are even direct copies

megatron/training/global_vars.py
deepakn94: Removed in 4de231f. Agreed it did not belong here: it reached into `opentelemetry.sdk` to monkeypatch `RandomIdGenerator` process-globally, and nemo-lens now handles seed-independent span IDs in its own provider setup.

deepakn94:
@ahmadki both addressed in 4de231f:

- **Test suite dropped.** Removed all seven files (`conftest.py`, `test_config`, `test_handle`, `test_helpers`, `test_instrumentation`, `test_instruments`, `test_propagation`). You were right that they belong to Lens; every one imports exclusively from `nemo.lens.*` and none exercised Megatron code. They came across verbatim from #6385 — I only ever reformatted them, since that PR had never been run through `autoformat.sh`.

  The four files remaining under `tests/unit_tests/telemetry/` are untouched: they landed on main in #6415, import `megatron.core.telemetry` directly, and cover the span groups, metric instruments, and no-op fallbacks. All 83 still pass after the deletion (the removed `conftest.py` only reset global OTel providers, which none of them use).

- **`RandomIdGenerator` patch removed** from `_set_telemetry`.

- **Both `nemo-lens` pin suggestions** accepted and merged.

Still open: @santhnm2's note that the instrumentation should move to `dynamic_text_gen_server/text_generation_server.py`. Looking at that next.

megatron/core/inference/text_generation_server/text_generation_server.py
deepakn94: Good catch — reverted `text_generation_server.py` to main in f521a8e, so this PR no longer touches the deprecated module at all.

Porting the GenAI-semconv spans and metrics to `dynamic_text_gen_server/text_generation_server.py` will come as a follow-up rather than being reworked here, to keep this PR to the training-loop instrumentation.

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
chtruong814, line 2: This change is due to the uv version used. Probably one of my team had a higher uv version and it updated to revision 3. It's the lockfile format version.  It's ok to downgrade here since it matches the uv version we use in the Dockerfile anyway.

pyproject.toml
```diff
@@ -96,7 +96,8 @@ dev = [
     "einops~=0.8",
     "tensorstore~=0.1,!=0.1.46,!=0.1.72",
     "multi-storage-client~=0.50",
-    "opentelemetry-api~=1.33.1",
+    "opentelemetry-api>=1.33.1,<2",
+    "nemo-lens @ git+https://github.com/NVIDIA-NeMo/Lens.git@b85578fc2b736a1804705e537001b5f45e9c715d",
```
chtruong814: For using a git commit, I recommend moving the git ref to the `tool.uv.sources` section like the rest of these:
https://github.com/NVIDIA/Megatron-LM/blob/main/pyproject.toml#L231

So you'd just reference "nemo-lens" here and add the commit in that later section.

pyproject.toml
```diff
@@ -115,6 +116,10 @@ dev = [
     "zstandard",
 ]
 
+otel = [
+    "nemo-lens[sdk] @ git+https://github.com/NVIDIA-NeMo/Lens.git@b85578fc2b736a1804705e537001b5f45e9c715d",
+]
```
chtruong814, line 121: Similar feedback as the one above, please just reference `nemo-lens[sdk]` here and move the git coomit to later section at `uv.sources`
https://github.com/NVIDIA/Megatron-LM/blob/main/pyproject.toml#L231

So it might look like this overall:
```
 [project.optional-dependencies]
  dev = [
      # ...
      "nemo-lens",
  ]

  otel = [
      "nemo-lens[sdk]",
  ]

  [tool.uv.sources]
  nemo-lens = { git = "https://github.com/NVIDIA-NeMo/Lens.git", rev = "b85578fc2b736a1804705e537001b5f45e9c715d" }
```

chtruong814 (commented):
Only reviewed the uv changes.  The revision change in the uv lock file is fine.

Would recommend changing how the git references are done in the pyproject.toml.

megatron/training/arguments.py
```diff
@@ -2509,6 +2509,33 @@ def _add_logging_args(parser):
     log_factory = ArgumentGroupFactory(LoggerConfig, exclude = ["log_throughput_to_tensorboard", "throughput_window_size", "memory_keys", "log_l2_norm_grad_to_tensorboard", "log_runtime_to_tensorboard", "runtime_time_unit", "filter_warnings", "modules_to_filter", "set_level_for_all_loggers", "save_config_filepath"])
     group = log_factory.build_group(parser, title="logging")
 
+    otel_group = parser.add_argument_group(title='opentelemetry')
```
asolergi-nv, line 2532: @maanug-nv This won't be directly exposed in MBridge, right? Should we include OTel in [`LoggerConfig`](https://github.com/NVIDIA/Megatron-LM/blob/e60e42fa1e82c2b9e2fb4145f93e446394447bda/megatron/training/config/training_config.py#L244)?

megatron/training/training.py
```diff
@@ -272,13 +282,413 @@ def set_startup_timestamps(program_start=None, main_entry=None):
...
+#            `- whatever runs live before train()'s loop (weight-hash check, sniff test,
+#               cuda-graph setup) -- they attach to the current context
+#   trace C_n: megatron.train       (one per checkpoint interval, each its own trace --
+#              see _reroot_otel_interval; C_0 Links back to trace B, C_n Links to C_n-1)
+#
+# The launch_script/container_load children are only emitted when this process actually
+# owns that window (the launch script passed presrun through AND we're not under NVRx,
+# where nvrx.cold_start covers it -- see pretrain_hybrid.py). When they are emitted the
+# startup trace is anchored at launch_script_start so they tile inside it; otherwise it
+# is anchored at program_start.
```
asolergi-nv, line 359: Do we need ALL these comments & details here?

asolergi-nv (approved):
There are no critical issues to address, but I wonder if we should start regulating which comments belong in Python files and which ones should be in the documentation.

ahmadki (approved):

rhewett-nv:
> There are no critical issues to address, but I wonder if we should start regulating which comments belong in Python files and which ones should be in the documentation.

Comments used to be for people but now they are for agents.  Having the documentation closer to the code seems to work better for them...

pyproject.toml
deepakn94: Done in da38988 — moved the ref to `[tool.uv.sources]` for both, so the revision lives in one place now instead of two:

```toml
dev  = [ ..., "nemo-lens" ]
otel = [ "nemo-lens[sdk]" ]

[tool.uv.sources]
nemo-lens = { git = "https://github.com/NVIDIA-NeMo/Lens.git", rev = "b85578fc2b736a1804705e537001b5f45e9c715d" }
```

Worth noting `uv.lock` came back **byte-identical** — uv records the same `source = { git = ..., rev = ... }` entry either way, so this is purely declaration style with no effect on resolution. Re-verified with uv 0.7.2 in the CI container: `uv lock --check` and `uv sync --locked --only-group linting` both exit 0.

This supersedes @ahmadki's two accepted suggestions, which bumped the rev inline in both spots — same revision, just no longer duplicated.

uv.lock
deepakn94: Thanks, that confirms it — I had flagged the `revision = 3` -> `2` line in the PR description as something reviewers might query. Generated with uv 0.7.2 to match the pin in both `Dockerfile.ci.dev` and `setup-uv` in `cicd-main.yml`, so it should stay at 2 until that pin moves.

jaredcasper (approved):

chtruong814 (approved):

PR #6440, opened 2026-08-11
Fix async eviction with hidden chunked prefill

lmcafee-nvidia:
## Summary

- allow async overlap bookkeeping to handle a chunked-prefill request that is temporarily outside the active request window after eviction
- preserve the invariant that the chunked-prefill request has at most one active row
- add a regression test covering eviction followed by the next overlap step

## Background

Under KV-memory pressure, the no-overlap schedule can evict an active decode request while retaining a chunked-prefill request outside the active request window. The next overlap step does not include that hidden request in its active request IDs, so requiring exactly one matching row raises an assertion before inference can continue.

## Validation

- the focused regression test fails on unmodified `main` at the original assertion
- HSG Nano TP4/EP4 pressure benchmark on unmodified `main` failed at step 4500, immediately after an eviction, with `AssertionError: The active chunked-prefill request must have exactly one row.`
- 5 focused controller tests pass on DFW with the fix
- the same HSG Nano TP4/EP4 pressure benchmark completes with the fix after 13 evictions (`4588.8 tok/s`)
- `git diff --check` and Python syntax checks pass

### DIFF
--- a/megatron/core/inference/text_generation_controllers/text_generation_controller.py
+++ b/megatron/core/inference/text_generation_controllers/text_generation_controller.py
@@ -2179,9 +2179,10 @@ class TextGenerationController:
                 active_request_ids == context.chunked_prefill_request_id, as_tuple=True
             )[0]
             assert (
-                chunked_prefill_rows.numel() == 1
-            ), "The active chunked-prefill request must have exactly one row."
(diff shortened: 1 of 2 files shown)


lmcafee-nvidia:
Closing in favor of a replacement PR focused on the underlying hidden chunked-prefill overlap condition.

PR #6484, opened 2026-08-12
Expose MFSDP v2 custom-schedule lifecycle

shjwudp:
## Summary

Part 1 of the MFSDP v2 combined-1F1B split. This PR contains only the generic experimental MFSDP lifecycle required by external fine-grained schedules.

- add a `register_hooks` option so an external scheduler can disable automatic forward/backward execution hooks while retaining state-dict load safety
- expose explicit `FsdpModule.unshard()` and `reshard()` operations
- expose parameter-completion-based post-backward callback registration, including Transformer Engine delayed-wgrad hooks
- allow manual backward schedules to opt out of the autograd final callback
- preserve NVIDIA main's `SchedulePolicy` prefetch behavior for the default automatic-hook path
- add operation-level NVTX ranges and focused lifecycle/callback coverage

## Split

The MCore adapter and `models/common` integration live in draft PR #6692 and depend on this PR.

## Validation

- DCO: passed
- Black 26.3.0: passed
- isort 5.13.2: passed
- Ruff 0.9.10: passed
- Python compilation and `git diff --check`: passed
- GitHub linting: passed
- focused GPU/unit-test CI on rebased commit `cff249e1969b`: in progress

### DIFF
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
@@ -110,6 +110,7 @@ def fully_shard(
     mixed_precision_policy: MixedPrecisionPolicy | None = None,
     grad_divisor: int = 1,
     schedule_policy: SchedulePolicy = SchedulePolicy(),
+    register_hooks: bool = True,
 ) -> None:
     """Apply FSDP to a module in place.
 
@@ -134,6 +135,10 @@ def fully_shard(
             ``grad_divisor=ep_size`` makes up the difference. Dense parameters see only
             their own rank's tokens and need no divisor.
         schedule_policy: Communication scheduling policy for this FSDP module.
+        register_hooks: Whether to register the automatic forward and backward execution
+            hooks on ``module``. Disable this when an external scheduler invokes the
+            corresponding FSDP lifecycle methods explicitly. The state-dict safety hook
+            is registered independently.
     """
     if isinstance(module, FsdpModule):
         raise ValueError("This module is already managed by FSDP.")
@@ -164,6 +169,7 @@ def fully_shard(
             grad_divisor=grad_divisor,
             schedule_policy=schedule_policy,
             use_symmetric_memory=context.use_symmetric_memory,
+            register_hooks=register_hooks,
         )
     except Exception:
         module.__class__ = original_cls
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
@@ -15,6 +15,7 @@
 """Module mixin for the minimal Megatron-FSDP path."""
 
 import enum
+from collections.abc import Callable
 from typing import Literal, cast
 from weakref import ref
 
@@ -159,6 +160,7 @@ class FsdpModule:
     _parameter_groups: tuple[FsdpParameterGroup, ...]
     _context: FsdpContext
     _trainable_parameter_countdown: Countdown
+    _post_backward_hook_registered: bool
     _is_root: bool
     _num_trainable_parameters: int
     _schedule_policy: SchedulePolicy
@@ -183,6 +185,7 @@ class FsdpModule:
         grad_divisor: int = 1,
         schedule_policy: SchedulePolicy = SchedulePolicy(),
         use_symmetric_memory: bool = False,
+        register_hooks: bool = True,
     ) -> None:
         """Initialize FSDP runtime state on an already-constructed module."""
         self._context = context
@@ -191,6 +194,7 @@ class FsdpModule:
         self._unshard_event = None
         self._phase = FsdpModule.Phase.RESTING
         self._schedule_policy = schedule_policy
+        self._post_backward_hook_registered = False
         owned_parameters = _collect_owned_parameters(self)
         if grad_divisor <= 0:
             raise ValueError(f"grad_divisor must be positive, got {grad_divisor}.")
@@ -222,7 +226,9 @@ class FsdpModule:
                 if group.requires_grad
             )
         )
-        self._register_hooks()
+        self.register_load_state_dict_pre_hook(FsdpModule._pre_load_state_dict)
+        if register_hooks:
+            self._register_hooks()
         context.register_module(self)
 
     @property
@@ -262,7 +268,6 @@ class FsdpModule:
 
     def _register_hooks(self) -> None:
         module = cast(nn.Module, self)
-        module.register_load_state_dict_pre_hook(FsdpModule._pre_load_state_dict)
         # Use PyTorch's callback module argument instead of capturing self so
         # these hooks do not retain a deleted FSDP module.
         module.register_forward_pre_hook(
@@ -274,12 +279,28 @@ class FsdpModule:
         module.register_full_backward_pre_hook(
             lambda hooked_module, _grad_output: cast(FsdpModule, hooked_module).pre_backward()
         )
+        self.register_post_backward_hook(FsdpModule.post_backward)
+
+    def register_post_backward_hook(
+        self, post_backward_hook: Callable[["FsdpModule"], None]
+    ) -> None:
+        """Register a callback to run when this module's backward is complete.
+
+        Args:
+            post_backward_hook: Callback receiving this FSDP module after all of its
+                trainable parameters have accumulated gradients.
+        """
+        if self._post_backward_hook_registered:
+            raise RuntimeError("This FSDP module already has a post-backward hook registered.")
+
+        module = cast(nn.Module, self)
         if self._trainable_parameter_countdown.initial_value == 0:
             module.register_full_backward_hook(
-                lambda hooked_module, _grad_input, _grad_output: cast(
-                    FsdpModule, hooked_module
-                ).post_backward()
+                lambda hooked_module, _grad_input, _grad_output: post_backward_hook(
+                    cast(FsdpModule, hooked_module)
+                )
             )
+            self._post_backward_hook_registered = True
             return
 
         # Gradient reduction for trainable parameters is parameter-completion
@@ -293,7 +314,7 @@ class FsdpModule:
             if module is None:
                 return
             if module._trainable_parameter_countdown.decrement():
-                module.post_backward()
+                post_backward_hook(module)
 
         for group in self._parameter_groups:
             if not group.requires_grad:
@@ -315,6 +336,7 @@ class FsdpModule:
                 parameter_module.register_wgrad_accumulation_and_reduce_hooks(
                     lambda parameter=parameter: grad_hook(parameter)
                 )
+        self._post_backward_hook_registered = True
 
     @staticmethod
     def _pre_load_state_dict(
@@ -352,25 +374,31 @@ class FsdpModule:
         if self.phase is not FsdpModule.Phase.BACKWARD:
             self.phase = FsdpModule.Phase.FORWARD
         torch.cuda.nvtx.range_push(self._nvtx_label("forward"))
-        allgather_stream = context.allgather_stream
-        current_stream = context.current_stream()
 
         if self.is_root():
-            allgather_stream.wait_stream(current_stream)
+            context.allgather_stream.wait_stream(context.current_stream())
 
+        self.unshard(prefetch="forward" if not is_recomputing else "none")
+
+    def unshard(self, prefetch: Literal["forward", "backward", "none"] = "none") -> None:
+        """Unshard this FsdpModule's parameter groups immediately."""
+        torch.cuda.nvtx.range_push(self._nvtx_label("unshard"))
         self._unshard_parameter_groups()
         assert self._unshard_event is not None
         # Compute waits only for this FsdpModule's all-gather (the prefetch below is
         # issued afterwards, so it is free to run concurrently with this FsdpModule).
-        current_stream.wait_event(self._unshard_event)
+        self.context.current_stream().wait_event(self._unshard_event)
 
-        # Activation recomputation runs forward hooks inside backward. Do not
-        # prefetch the next module in forward order: its backward may already
-        # be complete, so no later backward hook would reshard it.
-        if not is_recomputing:
+        context = self.context
+        if prefetch == "forward":
             self._prefetch_parameter_groups(
                 context.forward_order, self._schedule_policy.forward_prefetch_size
             )
+        elif prefetch == "backward":
+            self._prefetch_parameter_groups(
+                context.backward_order, self._schedule_policy.backward_prefetch_size
+            )
+        torch.cuda.nvtx.range_pop()
 
     def _prefetch_parameter_groups(
         self, order: IndexedOrder["FsdpModule"], prefetch_size: int | None
@@ -412,9 +440,15 @@ class FsdpModule:
         # post_backward() will reshard them after gradient reduction.
         is_recomputing = self.phase is FsdpModule.Phase.BACKWARD or _is_in_backward()
         if not is_recomputing:
-            self._reshard_parameter_groups()
+            self.reshard()
         if self.phase is FsdpModule.Phase.FORWARD:
             self.phase = FsdpModule.Phase.RESTING
+        torch.cuda.nvtx.range_pop()
+
+    def reshard(self) -> None:
+        """Reshard this FsdpModule's parameter groups."""
+        torch.cuda.nvtx.range_push(self._nvtx_label("reshard"))
+        self._reshard_parameter_groups()
         torch.cuda.nvtx.range_pop()
 
     def _reshard_parameter_groups(self) -> None:
@@ -435,14 +469,22 @@ class FsdpModule:
                 group.release_unsharded_storage()
             self._unshard_event = None
 
-    def pre_backward(self) -> None:
-        """Prepare full parameters and prefetch the next FsdpModule in backward order."""
+    def pre_backward(self, register_final_callback: bool = True) -> None:
+        """Prepare full parameters and prefetch the next FsdpModule in backward order.
+
+        Args:
+            register_final_callback: Whether to finalize through the autograd engine.
+                Manual backward schedules (e.g. 1F1B EP overlap schedule) finalize
+                explicitly in ``post_backward()``, so they pass False to avoid
+                installing an autograd callback outside the backward pass.
+        """
         self.phase = FsdpModule.Phase.BACKWARD
-        torch.cuda.nvtx.range_push(self._nvtx_label("backward"))
         context = self.context
+        torch.cuda.nvtx.range_push(self._nvtx_label("backward"))
         current_stream = context.current_stream()
         if self.is_root():
-            context.register_post_backward_final_callback()
+            if register_final_callback:
+                context.register_post_backward_final_callback()
             # Fork the reduce-scatter stream from the current stream once, at the
             # start of backward, so every module's post-backward reduce-scatter is
             # part of any active CUDA-graph capture. A stream only joins the
@@ -452,23 +494,18 @@ class FsdpModule:
             # fork each preceding module issues before its collective.
             context.reduce_scatter_stream.wait_stream(current_stream)
 
-        self._unshard_parameter_groups()
-        assert self._unshard_event is not None
-        current_stream.wait_event(self._unshard_event)
-
-        self._prefetch_parameter_groups(
-            context.backward_order, self._schedule_policy.backward_prefetch_size
-        )
+        self.unshard(prefetch="backward")
 
     def post_backward(self) -> None:
         """Reduce gradients and return parameters to their sharded resting state."""
-        self._reshard_parameter_groups()
+        self.reshard()
         self._reduce_gradient_groups()
         self.phase = FsdpModule.Phase.RESTING
         torch.cuda.nvtx.range_pop()
 
     def _reduce_gradient_groups(self) -> None:
         """Pack gradients and immediately launch their reduce-scatters."""
+        torch.cuda.nvtx.range_push(self._nvtx_label("gradient_reduce"))
         context = self.context
         reduce_scatter_stream = context.reduce_scatter_stream
         current_stream = context.current_stream()
@@ -486,6 +523,7 @@ class FsdpModule:
             reduce_scatter_stream.wait_stream(current_stream)
             with torch.cuda.stream(reduce_scatter_stream):
                 group.reduce_partial_gradients(partial_grad, self.context.is_last_microbatch)
+        torch.cuda.nvtx.range_pop()
 
     @property
     def parameter_groups(self) -> tuple[FsdpParameterGroup, ...]:
@@ -501,9 +539,9 @@ class FsdpModule:
             for parameter in group.fsdp_parameters
         )
 
-    def _nvtx_label(self, phase: Literal["forward", "backward"]) -> str:
-        name = self.name if self.name else "<root>"
-        return f"MFSDP {name} {phase}"
+    def _nvtx_label(self, operation: str) -> str:
+        module_name = self.name or "<root>"
+        return f"MFSDP {module_name} {operation}"
 
 
 def _collect_backward_order(module: nn.Module, order: IndexedOrder["FsdpModule"]) -> None:
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
@@ -160,7 +160,7 @@ class FsdpModule:
     _parameter_groups: tuple[FsdpParameterGroup, ...]
     _context: FsdpContext
     _trainable_parameter_countdown: Countdown
-    _post_backward_hook_registered: bool
+    _grad_reduction_callback_registered: bool
     _is_root: bool
     _num_trainable_parameters: int
     _schedule_policy: SchedulePolicy
@@ -194,7 +194,7 @@ class FsdpModule:
         self._unshard_event = None
         self._phase = FsdpModule.Phase.RESTING
         self._schedule_policy = schedule_policy
-        self._post_backward_hook_registered = False
+        self._grad_reduction_callback_registered = False
         owned_parameters = _collect_owned_parameters(self)
         if grad_divisor <= 0:
             raise ValueError(f"grad_divisor must be positive, got {grad_divisor}.")
@@ -290,7 +290,7 @@ class FsdpModule:
             post_backward_hook: Callback receiving this FSDP module after all of its
                 trainable parameters have accumulated gradients.
         """
-        if self._post_backward_hook_registered:
+        if self._grad_reduction_callback_registered:
             raise RuntimeError("This FSDP module already has a grad-reduction callback registered.")
 
         module = cast(nn.Module, self)
@@ -300,7 +300,7 @@ class FsdpModule:
                     cast(FsdpModule, hooked_module)
                 )
             )
-            self._post_backward_hook_registered = True
+            self._grad_reduction_callback_registered = True
             return
 
         # Gradient reduction for trainable parameters is parameter-completion
@@ -336,7 +336,7 @@ class FsdpModule:
                 parameter_module.register_wgrad_accumulation_and_reduce_hooks(
                     lambda parameter=parameter: grad_hook(parameter)
                 )
-        self._post_backward_hook_registered = True
+        self._grad_reduction_callback_registered = True
 
     @staticmethod
     def _pre_load_state_dict(
@@ -381,7 +381,15 @@ class FsdpModule:
         self.unshard(prefetch="forward" if not is_recomputing else "none")
 
     def unshard(self, prefetch: Literal["forward", "backward", "none"] = "none") -> None:
-        """Unshard this FsdpModule's parameter groups immediately."""
+        """Unshard this FsdpModule's parameter groups immediately.
+
+        External schedulers invoking this directly (rather than through the
+        automatic ``pre_forward`` hook) must first synchronize the all-gather
+        stream on the root by calling
+        ``context.allgather_stream.wait_stream(context.current_stream())``
+        before this when ``self.is_root()``; the automatic forward path
+        performs that root sync in ``pre_forward()`` immediately before this.
+        """
         torch.cuda.nvtx.range_push(self._nvtx_label("unshard"))
         self._unshard_parameter_groups()
         assert self._unshard_event is not None
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
@@ -15,7 +15,8 @@
 """Module mixin for the minimal Megatron-FSDP path."""
 
 import enum
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
+from contextlib import contextmanager
 from typing import Literal, cast
 from weakref import ref
 
@@ -375,6 +376,9 @@ class FsdpModule:
         is_recomputing = self.phase is FsdpModule.Phase.BACKWARD or _is_in_backward()
         if self.phase is not FsdpModule.Phase.BACKWARD:
             self.phase = FsdpModule.Phase.FORWARD
+        # forward/backward each span multiple lifecycle methods (pre_forward ->
+        # post_forward and pre_backward -> post_backward), so they keep explicit
+        # push/pop instead of a single context scope.
         torch.cuda.nvtx.range_push(self._nvtx_label("forward"))
 
         if self.is_root():
@@ -392,23 +396,22 @@ class FsdpModule:
         before this when ``self.is_root()``; the automatic forward path
         performs that root sync in ``pre_forward()`` immediately before this.
         """
-        torch.cuda.nvtx.range_push(self._nvtx_label("unshard"))
-        self._unshard_parameter_groups()
-        assert self._unshard_event is not None
-        # Compute waits only for this FsdpModule's all-gather (the prefetch below is
-        # issued afterwards, so it is free to run concurrently with this FsdpModule).
-        self.context.current_stream().wait_event(self._unshard_event)
+        with self._nvtx_range("unshard"):
+            self._unshard_parameter_groups()
+            assert self._unshard_event is not None
+            # Compute waits only for this FsdpModule's all-gather (the prefetch below is
+            # issued afterwards, so it is free to run concurrently with this FsdpModule).
+            self.context.current_stream().wait_event(self._unshard_event)
 
-        context = self.context
-        if prefetch == "forward":
-            self._prefetch_parameter_groups(
-                context.forward_order, self._schedule_policy.forward_prefetch_size
-            )
-        elif prefetch == "backward":
-            self._prefetch_parameter_groups(
-                context.backward_order, self._schedule_policy.backward_prefetch_size
-            )
-        torch.cuda.nvtx.range_pop()
+            context = self.context
+            if prefetch == "forward":
+                self._prefetch_parameter_groups(
+                    context.forward_order, self._schedule_policy.forward_prefetch_size
+                )
+            elif prefetch == "backward":
+                self._prefetch_parameter_groups(
+                    context.backward_order, self._schedule_policy.backward_prefetch_size
+                )
 
     def _prefetch_parameter_groups(
         self, order: IndexedOrder["FsdpModule"], prefetch_size: int | None
@@ -457,9 +460,8 @@ class FsdpModule:
 
     def reshard(self) -> None:
         """Reshard this FsdpModule's parameter groups."""
-        torch.cuda.nvtx.range_push(self._nvtx_label("reshard"))
-        self._reshard_parameter_groups()
-        torch.cuda.nvtx.range_pop()
+        with self._nvtx_range("reshard"):
+            self._reshard_parameter_groups()
 
     def _reshard_parameter_groups(self) -> None:
         """Reshard parameter groups and release unsharded storage after compute.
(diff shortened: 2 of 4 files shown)


wujingyue (commented):
Thanks for the PR! Can you think about how to https://google.github.io/eng-practices/review/developer/small-cls.html so I can start review? 

One high-level question: 

> With delay_wgrad_compute=True, the per-param post_accumulate_grad_hook fires when activation gradients are ready, before backward_dw() writes weight gradients.

Really? I believe `post_accumulate_grad_hook` always goes with weight gradients.

shjwudp:
> Thanks for the PR! Can you think about how to https://google.github.io/eng-practices/review/developer/small-cls.html so I can start review?
> 
> One high-level question:
> 
> > With delay_wgrad_compute=True, the per-param post_accumulate_grad_hook fires when activation gradients are ready, before backward_dw() writes weight gradients.
> 
> Really? I believe `post_accumulate_grad_hook` always goes with weight gradients.

@wujingyue Agreed — this PR is currently rather large because it originated from the prototype. Although I have already simplified parts of it, it is still structured as an all-in-one change.

I have been thinking about splitting it into the following parts:

1. A design document.
2. FSDP module support for optionally registering fine-grained FSDP hooks.
3. A follow-up change, after (2) lands, to integrate this with MCore 1F1B overlap.

Does this breakdown sound reasonable to you?

Regarding:

> With `delay_wgrad_compute=True`, the per-param `post_accumulate_grad_hook` fires when activation gradients are ready, before `backward_dw()` writes weight gradients.

Yes, that is correct. `backward_dw()` does not trigger `post_accumulate_grad_hook`. The hook is not a general observer of changes to `parameter.grad`; it is triggered only by `AccumulateGrad`.

I also put together a minimal reproduction to verify this behavior:
https://gist.github.com/shjwudp/d95722088ea45cb1a2f906b964fb914e

shjwudp:
Created a copy of the design doc to make commenting easier:
https://docs.google.com/document/d/1Jb_6IN1uJFTRFlI5-3zp_WsOPB8jqUqgupiaG6zpPn0/edit?usp=sharing

wujingyue (commented):
> Does this breakdown sound reasonable to you?

Sounds good to me!

wujingyue (commented):
> I also put together a minimal reproduction to verify this behavior:
https://gist.github.com/shjwudp/d95722088ea45cb1a2f906b964fb914e

Seems pretty contrived. Is `backward_dw` from TE? Could you change the repro to use a TE layer so I can verify the actual pattern?

shjwudp:
@wujingyue Thanks — you were right that the zero-placeholder detail in the
mock was contrived. I added a Transformer Engine version using a real
`transformer_engine.pytorch.Linear` with the configuration supported by this
PR (`delay_wgrad_compute=True`, `fuse_wgrad_accumulation=False`):

https://gist.github.com/shjwudp/d95722088ea45cb1a2f906b964fb914e#file-delay_wgrad_hook_order_te-py

I ran it with Transformer Engine `2.14.1+366798ef`. The observed order is:

```text
PyTorch post-accumulate hook during backward(): weight.grad=None
After backward(): weight.grad=None
TE wgrad callback during backward_dw(): weight.grad=materialized (nonzero=4096)
After backward_dw(): weight.grad=materialized (nonzero=4096)
PASS with Transformer Engine 2.14.1+366798ef
```

So the real TE behavior confirms the ordering, with one correction to the
mock: the PyTorch hook fires during `backward()` while `weight.grad` is still
`None`, not a zero placeholder. Then `backward_dw()` assigns the real gradient
directly and does not retrigger the PyTorch hook; TE's dedicated wgrad callback
fires after materialization. This is why the FSDP reduction/release must be
scheduled explicitly after `backward_dw()`.

wujingyue:
Thanks for the reproducer—confirmed locally with TE `2.14.1+366798ef`.

With `delay_wgrad_compute=True`, TE’s custom `_Linear.backward()` queues the wgrad inputs/GEMM closure instead of running it ([source](https://github.com/NVIDIA/TransformerEngine/blob/366798ef8a0a00d8f2c1650d11e7e623d7c33e26/transformer_engine/pytorch/module/linear.py#L887-L900)). `wgrad` was initialized to `None` ([source](https://github.com/NVIDIA/TransformerEngine/blob/366798ef8a0a00d8f2c1650d11e7e623d7c33e26/transformer_engine/pytorch/module/linear.py#L764-L768)) and is returned as the first backward result—the gradient for the first forward input, `weight` ([source](https://github.com/NVIDIA/TransformerEngine/blob/366798ef8a0a00d8f2c1650d11e7e623d7c33e26/transformer_engine/pytorch/module/linear.py#L977-L982)).

PyTorch nevertheless invokes the weight’s `post_accumulate_grad_hook`, while `weight.grad` remains `None`. `Linear.backward_dw()` later executes the saved GEMM and directly assigns `weight_tensor.grad` ([source](https://github.com/NVIDIA/TransformerEngine/blob/366798ef8a0a00d8f2c1650d11e7e623d7c33e26/transformer_engine/pytorch/module/base.py#L1571-L1582)); because that occurs outside autograd, it does not rerun PyTorch’s post-accumulate hook.

So `post_accumulate_grad_hook` is not a usable “wgrad is materialized” signal for TE’s delayed-wgrad mode. The TE `register_wgrad_accumulation_and_reduce_hooks()` callback is the correct notification point for MFSDP v2 to begin gradient handling.

wujingyue:
Can `delay_wgrad_compute=True` be supported separately as a prerequisite of 1f1b? That can be the first PR while I review the rest.

shjwudp:
> Can `delay_wgrad_compute=True` be supported separately as a prerequisite of 1f1b? That can be the first PR while I review the rest.

In theory, that would be a cleaner split. But `delay_wgrad_compute=True` currently depends on the 1F1B EP-overlap, as enforced [here](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/transformer_config.py#L3024). Supporting it independently would require additional scheduling work rather than simply splitting out the existing code. So we can't.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/docs/1f1b_ep_overlap_design.md
```diff
@@ -0,0 +1,361 @@
...
+### Key flags
+
+| Flag | What it does |
+|---|---|
+| `--overlap-moe-expert-parallel-comm` | Enables the combined 1F1B schedule and fine-grained FSDP hooks. |
+| `--delay-wgrad-compute` | TE layers postpone weight gradient computation until `backward_dw()`. FSDP must defer `reduce_grad()` until after `backward_dw()`. |
+
+### Why `delay_wgrad_compute` matters for FSDP
+
+`post_accumulate_grad_hook` observes execution of the parameter's autograd
```
wujingyue: Have you considered TE's register_wgrad_accumulation_and_reduce_hooks? Some other code in MCore uses it. 

Here's a brief draft: https://github.com/NVIDIA/Megatron-LM/pull/6697

shjwudp: This is related to the 1F1B scheduling design. My understanding is that this should be fine as long as TE’s `register_wgrad_accumulation_and_reduce_hooks` covers all post-backward use cases.

But seems like MCore also have 1F1B-specific [backward_dw](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/models/common/utils.py#L411) and [post-backward scheduling hook](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/models/common/utils.py#L365).

I believe @Wohox would be better placed to explain the rationale behind this part.

Wohox: In Megatron FSDP v1, The [post backward reshard hook](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/models/common/utils.py#L365) must be called manually because 1F1B overlaping bypass the FSDP unit module backward(Autograd is bypassed) and calls dispatch/experts/combine in 3 callable functions. However, if we registered the hooks to experts instead of MoE or Transformer layer, I think the autograd based backward reshard hook should work.

one possible refactor would be to integrate post backward reshared hooks into backward dw's gradient accumulation hook, They should be fired at the same time, the only issue is these two hooks are separately managed by Mcore and TE.

shjwudp: I understand that if we want to use `register_wgrad_accumulation_and_reduce_hooks`, we also need to handle this case:
https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/models/common/utils.py#L319-L320

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
```diff
@@ -107,6 +107,7 @@ def fully_shard(
     mesh: DeviceMesh,
     placements: Placements,
     mixed_precision_policy: MixedPrecisionPolicy | None = None,
+    skip_forward_backward_hooks: bool = False,
```
wujingyue: I thought we agreed on still invoking the hooks but have options to only skip prefetching (e.g. https://github.com/NVIDIA/Megatron-LM/pull/6905) and/or gradient reduction?

wujingyue (commented):
Can we separate the adapter changes to a different PR? The megatron_fsdp/experimental changes look more reasonable for me to start with.

wujingyue:
> Can we separate the adapter changes to a different PR? The megatron_fsdp/experimental changes look more reasonable for me to start with.

I missed #6927. Thanks!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
shjwudp: Currently this PR use fine-grained hooks replace M-FSDP v2 default hooks. Keep M-FSDP v2 default hooks encouter technical challenges:

1. **1F1B** breaks down the model into sub-layer schedules, which can lead to the loss handle of some modules. In the worst case, it may lose the `FsdpModule` handle entirely, preventing the corresponding FSDP unshard and reshard operations from completing. For example, 1F1B maintains Attn and MLP, but loses the transformerlayer handler. Further complicating matters, we cannot assume that `fully_shard` always wraps a TransformerLayer. Fine-grained hooks provide links from any submodule to the `FsdpModule`, simplifying 1F1B applications.
2. The state management (`FsdpModule.phase`) of 1F1B hooks differs from the default management of M-FSDP v2 because multiple micro-batch forward and backward transitions occur simultaneously. If supported, defensive programming might need to be added to the M-FSDP v2 hooks. Introduce complicate.

> Can we separate the adapter changes to a different PR?

Yes, split this PR to #6927 and #6692. This PR is use for reference.

wujingyue:
> Yes, split this PR to https://github.com/NVIDIA/Megatron-LM/pull/6927 and https://github.com/NVIDIA/Megatron-LM/pull/6692. This PR is use for reference.

Converted to draft given the comment above

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
wujingyue: Didn't we converge on removing this flag but letting FsdpModule expose lower-level APIs like unshard and reshard? (That would look good to me)

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue, line 1: As discussed offline, I'd love this to be moved as close to 1f1b scheduling as possible. It's more like making 1f1b support FsdpModule than making MFSDP support 1f1b.

wujingyue (commented):
Thanks for the PR!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
shjwudp: Removed. Would it make sense to optionally skip prefetching in the 1F1B case?

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
shjwudp: I’m wondering whether it would make more sense to place the fine-grained hooks in [mcore_fsdp_adapter.py](https://github.com/NVIDIA/Megatron-LM/pull/6484/files/9ef6db1ebf192928e64500d5016fe7edffe0aafa#diff-8d3ed77aead9b21b07dfd6c4df4a3b0db421c1f29cf1d1ff52fcc72f8e2eccbf), so they can serve as common utilities for MCore.

Also, regarding the generic scheduler (for optimize prefetch and avoid over-reshard) we discussed earlier, I think it would be better to implement it here, since this component is more tightly coupled with M-FSDP.

tests/unit_tests/distributed/mfsdp_v2/test_mcore_nd_parallel.py
```diff
@@ -0,0 +1,425 @@
...
+
+        try:
+            model, optimizer = _make_model_and_optimizer(
+                use_mfsdp_v2=use_mfsdp_v2,
+                overrides={
+                    "micro_batch_size": cls.MICRO_BATCH_SIZE,
+                    "global_batch_size": cls.GLOBAL_BATCH_SIZE,
+                    "vocab_size": cls.VOCAB_SIZE,
+                    "padded_vocab_size": cls.VOCAB_SIZE,
+                    "untie_embeddings_and_output_weights": True,
```
shjwudp: TODO: `post_backward` can miss gradients for tied weights.

`register_post_accumulate_grad_hook` is safe for ordinary, non-1F1B training because PyTorch accumulates gradient contributions from both uses of the tied weight before running its `AccumulateGrad` node:

```text
GraphTask output contribution  ─┐
                                ├─ shared AccumulateGrad → one completion hook
GraphTask embedding contribution┘
```

In contrast, combined 1F1B executes these contributions in separate graph tasks:

```text
GraphTask A → output contribution    → completion hook
GraphTask B → LayerNorm gradients
GraphTask C → embedding contribution → completion hook again
```

As a result, the first `post_backward` invocation may observe only the output-side contribution, before the embedding-side gradient has been accumulated.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
wujingyue: > Removed

Thanks! 

> Would it make sense to optionally skip prefetching in the 1F1B case?

Doesn't look relevant to me. In 1F1B, we never call TransformerLayer so the hooks (e.g. FsdpModule.pre_forward) won't be invoked anyway. I think a better way is to expose lower-level FsdpModule APIs like unshard_parameter_groups (which don't prefetch) so 1F1B can call them directly. 

Bottom line: I'd like 1F1B to invoke what it needs, instead of invoking a wider-than-necessary API and asking that API to skip certain things. 

What do you think?

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: > they can serve as common utilities for MCore

There seem to be many places in Megatron-LM that "can serve as common utilities". Where is MoE 1f1b right now so I can take a look at how to support FsdpModule from there. 

> regarding the generic scheduler (for optimize prefetch and avoid over-reshard)

I don't think that affects this PR. But if you really want to prefetch my thought on this, I'd probably create a separate file for that scheduler. **One** reason is that mcore_fsdp_adapter.py is getting a bit too large -- >1000 lines of code.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
shjwudp: Although the `pre_forward` hook of `TransformerLayer` is not called, `GroupedLinear` is also an `FsdpModule`, and in some special cases, `Embedding` might be as well.

Therefore, if we do not skip the forward/backward hooks, we should at least skip prefetching to avoid incorrect unsharding.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
shjwudp: > Where is MoE 1f1b right now so I can take a look at how to support FsdpModule from there.

The 1F1B implementation in MCore mainly consists of two files:

https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/pipeline_parallel/combined_1f1b.py

https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/models/common/model_chunk_schedule_plan.py

Simply put, `combined_1f1b` implements the pipeline parallelism "step" abstraction, while `model_chunk_schedule_plan` implements how the 1F layer, 1B layer - TransformerLayerNode computation, and how the 1F chunk, 1B chunk - VPP chunk perform computations.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
wujingyue: > GroupedLinear is also an FsdpModule

Thanks -- I'm sure I missed that. 

From yesterday's discussion, I thought you don't want these submodules to be FsdpModules. Instead, you want them to trigger the **parent** module's (i.e. TransformerLayer) pre-forward hooks because their parameters are often leaked (e.g. token_dispatcher uses shared_expert_weights). Is this no longer ture? 

https://github.com/NVIDIA/Megatron-LM/blob/a5cdf57d9be925b3d4b2f64a46b1057e442e1c86/megatron/core/models/common/model_chunk_schedule_plan.py#L285 seems to call `mlp.forward()` without triggering the hooks. So are you going to change that to `mlp()`? 

Also, does this mean you do need https://github.com/NVIDIA/Megatron-LM/pull/6697 because GroupedLinear, an FsdpModule, will be `delay_wgrad`ed?

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: > model_chunk_schedule_plan

Yeah, that sounds like the right place to schedule FsdpModules as well as normal modules.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
shjwudp: Thanks, I see the point. I’ll check whether any 1F1B calls might unexpectedly trigger `FsdpModule`’s `pre_forward` or `pre_backward` hooks. If that cannot happen, then the `skip_prefetching` switch would primarily serve as a safety guard.

That said, I think the switch could still be useful: it gives users an explicit guarantee that, when using `FsdpModule`, no prefetching-related operations will be performed.

> Also, does this mean you do need #6697 because `GroupedLinear`, an `FsdpModule`, will be delay_wgraded?

Yes. The current 1F1B implementation depends on #6697. 🙌

wujingyue: > Yes. The current 1F1B implementation depends on https://github.com/NVIDIA/Megatron-LM/pull/6697. 🙌

OK -- I'll merge it

wujingyue: > I think the switch could still be useful

Agreed. I'm just trying to YAGNI.

shjwudp: > `f_input = f_layer.mlp.forward(f_input)` seems to call mlp.forward() without triggering the hooks. So are you going to change that to mlp()?

The naming is confusing, it will indeed trigger a GroupLinear `pre_forward` hook:

```
  f_layer.mlp                           # TransformerLayerNode, not MoELayer
    .forward(...)
      → submodule_moe_forward(...)
        → layer.mlp.routed_experts_compute(...)  # direct method call
          → apply_module(layer.mlp.experts)(...) # equivalent to experts(...)
```

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
shjwudp: Looks good. I’ve split the fine-grained hooks into a separate PR: #6692.

wujingyue: 🙇

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -346,26 +351,38 @@ def pre_forward(self) -> None:
...
         current_stream = context.current_stream()
 
         if self.is_root():
             allgather_stream.wait_stream(current_stream)
 
+        self.unshard(
+            prefetch="forward" if not context.custom_schedule and not is_recomputing else "none"
+        )
+
+    def unshard(self, prefetch: Literal["forward", "backward", "none"] = "none") -> None:
```
wujingyue, line 389: Looks great!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -346,26 +351,38 @@ def pre_forward(self) -> None:
...
         if self.phase is not FsdpModule.Phase.BACKWARD:
             self.phase = FsdpModule.Phase.FORWARD
-        torch.cuda.nvtx.range_push(self._nvtx_label("forward"))
+        if not context.custom_schedule:
+            torch.cuda.nvtx.range_push(self._nvtx_label("forward"))
         allgather_stream = context.allgather_stream
         current_stream = context.current_stream()
 
         if self.is_root():
             allgather_stream.wait_stream(current_stream)
```
wujingyue:
```suggestion

if self.is_root():
context.allgather_stream.wait_stream(context.current_stream())
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -346,26 +351,38 @@ def pre_forward(self) -> None:
...
+        # A reentrant checkpoint recomputes before the child module's backward-pre
+        # hook runs. The active autograd GraphTask identifies that recomputation.
+        if prefetch == "forward":
+            next_module = self.context.forward_order.next_item(self)
             if next_module is not None:
                 next_module._unshard_parameter_groups()
+        elif prefetch == "backward":
+            next_module = self.context.backward_order.next_item(self)
+            if next_module is not None:
+                next_module._unshard_parameter_groups()
```
wujingyue, line 423: There's a bit duplication here. Consider:

```
next_module = None
if forward:
  next_module = ...
elif backward:
  next_module = ...

if next_module is not None:
  next_module._unshard_parameter_groups()
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -465,15 +494,18 @@ def _reduce_gradient_groups(self) -> None:
...
+        torch.cuda.nvtx.range_pop()
 
     @property
     def parameter_groups(self) -> tuple[FsdpParameterGroup, ...]:
         """Parameter groups owned by this FsdpModule."""
         return self._parameter_groups
 
-    def _nvtx_label(self, phase: Literal["forward", "backward"]) -> str:
+    def _nvtx_label(
+        self, operation: Literal["forward", "backward", "unshard", "reshard", "gradient_reduce"]
```
wujingyue:
```suggestion
        self, operation: str
```

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -465,15 +494,18 @@ def _reduce_gradient_groups(self) -> None:
...
+    ) -> str:
         name = self.name if self.name else "<root>"
```
wujingyue:
```suggestion
        module_name = self.name or "<root>"
```

tests/unit_tests/distributed/mfsdp_v2/test_annotation.py
```diff
@@ -17,13 +17,13 @@
     fully_shard_context,
 )
 
-_NVTX_LABEL_PATTERN = re.compile(r"MFSDP (.+) (forward|backward)")
+_NVTX_LABEL_PATTERN = re.compile(r"MFSDP (.+) (forward|backward|unshard|reshard|gradient_reduce)")
```
wujingyue: Let's simplify the pattern. Don't check for a fixed set of event names for flexibility.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
wujingyue: Also, would you like this to be inside `def unshard` so it can be reused by 1f1b? Or will you end up synchronizing differently?

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
```diff
@@ -64,6 +64,7 @@ def fully_shard_context(
     *,
     use_symmetric_memory: bool = False,
     unify_communication_stream: bool = False,
+    custom_schedule: bool = False,
```
wujingyue: I think `fully_shard(register_hooks=True)` would work better. By default, we register. Otherwise, the user is responsible for providing the hooks, which can of course reuse `FsdpModule`'s lower-level API such as `unshard` and `reshard`. 

This essentially goes back to my earlier comment: https://github.com/NVIDIA/Megatron-LM/pull/6484#discussion_r3920197023. Instead of adding a knob to turn off half of the logic scattered various places, I'd rather 1f1b compose the right hooks from ground up, cherrypicking whatever it likes.

wujingyue (commented):
LGTM otherwise!

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
shjwudp: I’d prefer to handle it separately. For the root module, stream synchronization should occur only during the pre-forward phase, rather than inside unshard. That keeps the synchronization point explicit and avoids coupling unshard to behavior that may differ for 1F1B.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
shjwudp: I made a few changes. `register_hooks` now controls all hooks except the state_dict hooks. Do you think this approach is acceptable?

I think we should continue to register the state_dict hooks by default, since they are required to ensure that `FsdpModule` is initialized correctly during training setup.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -274,12 +279,28 @@ def _register_hooks(self) -> None:
         module.register_full_backward_pre_hook(
             lambda hooked_module, _grad_output: cast(FsdpModule, hooked_module).pre_backward()
         )
+        self.register_post_backward_hook(FsdpModule.post_backward)
```
wujingyue, line 285: Why? I intentionally avoided `register_post_backward_hook` because it's fired at the wrong time: https://github.com/NVIDIA/Megatron-LM/blob/95387a1e4a864781b893abc607cdaec635ee75f4/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py#L285-L288

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -435,14 +469,22 @@ def _reshard_parameter_groups(self) -> None:
                 group.release_unsharded_storage()
             self._unshard_event = None
 
-    def pre_backward(self) -> None:
-        """Prepare full parameters and prefetch the next FsdpModule in backward order."""
+    def pre_backward(self, register_final_callback: bool = True) -> None:
```
wujingyue:
```suggestion
    def pre_backward(self) -> None:
```
Why this knob? I thought when register_hooks=False 1f1b will run its own pre_backward hook, not even calling this.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
shjwudp: This is actually a newly added method on `FsdpModule`, rather than a native PyTorch hook. It may be clearer to rename it to `register_grad_reduction_callback`.

The purpose of this function is to expose a method for the 1F1B scheduler to register a post-backward callback. Within that callback, the scheduler can control whether resharding should be performed and potentially defer gradient reduction.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
shjwudp: Correct, this option is no longer needed; I'll revert it.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -222,7 +226,9 @@ def __init__(
                 if group.requires_grad
             )
         )
-        self._register_hooks()
+        self.register_load_state_dict_pre_hook(FsdpModule._pre_load_state_dict)
```
wujingyue, line 232: Code-comment why this is done unconditionally.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
wujingyue: My bad -- I confused it with the one from torch.nn.Module. 

Naming-wise, let's keep `register_post_backward_hook`. `register_grad_reduction_callback` is inaccurate because the callback also reshards, and it's unclear from the name **when** this callback will be fired.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -159,6 +160,7 @@ class Phase(enum.Enum):
     _parameter_groups: tuple[FsdpParameterGroup, ...]
     _context: FsdpContext
     _trainable_parameter_countdown: Countdown
+    _post_backward_hook_registered: bool
```
wujingyue:
```suggestion
```
I'd remove this state. 

You probably added it for error checking, but I'd rather error checking be done by 1f1b. This way, FsdpModule can stay flexible to accept multiple hooks.

Also, it's weird that **only** the post backward hook checks this, not other hooks.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -452,23 +494,18 @@ def pre_backward(self) -> None:
...
         """Reduce gradients and return parameters to their sharded resting state."""
-        self._reshard_parameter_groups()
+        self.reshard()
         self._reduce_gradient_groups()
         self.phase = FsdpModule.Phase.RESTING
         torch.cuda.nvtx.range_pop()
 
     def _reduce_gradient_groups(self) -> None:
         """Pack gradients and immediately launch their reduce-scatters."""
+        torch.cuda.nvtx.range_push(self._nvtx_label("gradient_reduce"))
```
wujingyue: Nit: prefer a context scope so it works better with early return. Can be done in a separate PR.

tests/unit_tests/distributed/mfsdp_v2/test_annotation.py
```diff
@@ -76,18 +76,52 @@ def parse_nvtx_label(label: str) -> tuple[str, str]:
...
-        name, phase = label_stack.pop()
-        events.append(NvtxEvent("pop", name, phase))
+        name, operation = label_stack.pop()
+        events.append(NvtxEvent("pop", name, operation))
 
     monkeypatch.setattr(torch.cuda.nvtx, "range_push", record_push)
     monkeypatch.setattr(torch.cuda.nvtx, "range_pop", record_pop)
 
 
+def _lifecycle_events(events: list[NvtxEvent]) -> list[NvtxEvent]:
```
wujingyue:
```suggestion
def _forward_backward_events(events: list[NvtxEvent]) -> list[NvtxEvent]:
```

tests/unit_tests/distributed/mfsdp_v2/test_context.py
```diff
@@ -158,6 +158,90 @@ def test_nested_prefetch_orders_use_dfs(distributed_setup):
     assert list(context.backward_order) == [model, model.right, model.left, model.left.inner]
 
 
+def test_fully_shard_can_disable_execution_hooks(distributed_setup):
```
wujingyue, line 161: https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html

I'd probably remove the test, because it's too coupled with the implementation of register_hooks and adds little value. Also, it relies on private attributes (such as _forward_hooks), which worries me. 

I'm sure you'll exercise register_hooks=False well enough in your upcoming 1f1b PRs.

tests/unit_tests/distributed/mfsdp_v2/test_context.py
```diff
@@ -158,6 +158,90 @@ def test_nested_prefetch_orders_use_dfs(distributed_setup):
...
+    model = nn.Linear(4, 4).to(device)
+
+    with fully_shard_context(device=device):
+        fully_shard(model, mesh=mesh, placements=_flat_placements(), register_hooks=False)
+
+    parameters = [
+        fsdp_parameter.unsharded
+        for group in model.parameter_groups
+        if group.requires_grad
+        for fsdp_parameter in group.fsdp_parameters
```
wujingyue, line 195: https://google.github.io/styleguide/pyguide.html#27-comprehensions--generator-expressions

tests/unit_tests/distributed/mfsdp_v2/test_context.py
```diff
@@ -158,6 +158,90 @@ def test_nested_prefetch_orders_use_dfs(distributed_setup):
...
+        assert not fsdp_module._forward_hooks
+        assert not fsdp_module._backward_pre_hooks
+        assert not fsdp_module._backward_hooks
+        assert len(fsdp_module._load_state_dict_pre_hooks) == 1
+        for group in fsdp_module.parameter_groups:
+            for fsdp_parameter in group.fsdp_parameters:
+                assert not getattr(fsdp_parameter.unsharded, "_post_accumulate_grad_hooks", None)
+
+
+def test_register_grad_reduction_callback_waits_for_all_parameter_gradients(distributed_setup):
```
wujingyue: https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html

Either remove this test entirely or change it to focus on user-observable behavior. For example, you could register hooks manually (like https://github.com/NVIDIA/Megatron-LM/blob/b1fe7599e18a213292600177ad7ff290a1127160/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py#L268-L276) and then verify in your custom `post_backward` callback that all parameters have `.grad`.

tests/unit_tests/distributed/mfsdp_v2/test_context.py
```diff
@@ -234,3 +318,33 @@ def test_fully_shard_rejects_child_from_another_context(distributed_setup):
...
+        fully_shard(model, mesh=mesh, placements=_flat_placements())
+
+    reduce_calls = []
+    original_reduce_gradient_groups = model._reduce_gradient_groups
+
+    def record_reduce_gradient_groups() -> None:
+        reduce_calls.append(None)
+        original_reduce_gradient_groups()
+
+    monkeypatch.setattr(model, "_reduce_gradient_groups", record_reduce_gradient_groups)
```
wujingyue, line 327: Not sure what this test tries to verify. Given the reliance on private methods, this probably falls again into https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html

wujingyue (commented):
Almost LGTM! Tests can be cleaned up quite a bit.

tests/unit_tests/distributed/mfsdp_v2/test_context.py
shjwudp: Thanks, I learned something new!

tests/unit_tests/distributed/mfsdp_v2/test_context.py
shjwudp: This is used to check whether `reduce_gradients` is still effective when one module processes multiple sets of data at the same time. I can verify this part in a subsequent 1F1B PR.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
shjwudp: Handle by the commit https://github.com/NVIDIA/Megatron-LM/pull/6484/changes/54cc853ab9c4b652c5dff555249bb7c239de79bc

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/module.py
```diff
@@ -452,40 +498,37 @@ def pre_backward(self) -> None:
...
         self._reduce_gradient_groups()
         self.phase = FsdpModule.Phase.RESTING
         torch.cuda.nvtx.range_pop()
 
     def _reduce_gradient_groups(self) -> None:
         """Pack gradients and immediately launch their reduce-scatters."""
-        context = self.context
-        reduce_scatter_stream = context.reduce_scatter_stream
-        current_stream = context.current_stream()
+        with self._nvtx_range("gradient_reduce"):
```
wujingyue:
```suggestion
        with self._nvtx_range("reduce_gradients"):
```
or `gradient_reduction` to fix the grammar.

wujingyue (approved):
🚢

PR #6486, opened 2026-08-12
Enable Pipeline Parallelism Support for M-FSDP v2

shjwudp:
## Summary

Under virtual pipeline parallelism, `wrap_model_chunks_with_ddp` wraps each chunk in its own `FullyShardedDataParallelV2`, and each adapter `__init__` previously opened a fresh `fully_shard_context`. Every chunk therefore got independent communication streams, microbatch state, and prefetch orders, so the combined-1F1B schedule could not coordinate all-gathers or reduce-scatters across chunks.

## Changes

- `fully_shard_context` gains `reuse_existing`: when a context is already active on the same device, it is yielded instead of creating a new one, and only the outermost scope finalizes it. Nesting rejection remains the default.
- The MFSDP v2 adapter opens its construction scope with `reuse_existing`, so it joins an ambient training-loop context when present.
- `wrap_model_chunks_with_ddp` wraps the per-chunk wrapping loop in one `fully_shard_context` when DP is the MFSDP v2 adapter and there is more than one chunk, so all VPP chunks share one context (streams, orders, `is_last_microbatch`) and finalize is deferred to the outer scope.
- ModelBuilder path (`prepare_existing_model_chunks_for_distributed_training`) mirrors the same ambient-context fix.
at do not follow the static forward/backward orders.

## Tests

Unit coverage for trace, replay wrap-around, orientation-aware prefetch, divergence re-trace, stale release, and nested reuse scopes sharing the outer context.

## Validation (GPU test results)

Tested on EOS H100 with `deepseek_v3_proxy` + real DeepSeek-V3 SlimPajama, seed 1234, PP=2/VPP=2, TP1/EP1/CP1 (DP=4), MBS=1/GBS=8, seq 4096, 1000 iterations, tensorboard + W&B.

- **1000-step real-data convergence — M-FSDP v2 ≡ DistOpt.** Both PP=2/VPP=2 configs converge to the same val-loss at every 100-iter eval; max diff `~0.02` (~0.3% relative), both **0 NaN**, exit 0. Final val-loss @iter1000: **M-FSDP v2 6.023** vs **DistOpt 6.019** (lm-loss @1000: 6.490 vs 6.485; both start ~11.93 → ~6.0). W&B runs: `nidznep9` (M-FSDP v2) / `6d21q4g7` (DistOpt), project `jianbinc-mfsdp-v2-vpp-benchmark`.


megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
```diff
@@ -65,6 +65,7 @@ def fully_shard_context(
     *,
     use_symmetric_memory: bool = False,
     unify_communication_stream: bool = False,
+    reuse_existing: bool = False,
```
wujingyue: This looks fishy... Why (why do we **need** this) and why not (why not other approaches)?

shjwudp: The goal here is to share a single `FsdpContext` across VPP chunks, thereby avoiding an excessive number of communication streams, potential CCA fragmentation, and other issues introduced by having multiple `FsdpContext` instances.

wujingyue: > share a single FsdpContext across VPP chunks

Why can't we share them by wrapping everything in the same FsdpContext?

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -727,9 +729,13 @@ def _validate_config(
             # silently inflated norm.
             raise ValueError("MFSDP v2 does not currently support mtp_detach_heads.")
 
+        # Pipeline parallelism is supported with MFSDP v2 as long as a single shared
+        # FsdpContext is opened across all VPP model chunks (see
+        # wrap_model_chunks_with_ddp / dist_utils._ddp_wrap). The data-parallel mesh is
+        # built solely from dp_cp, so DP sharding is unaffected by how layers are split
+        # across pipeline stages or virtual stages.
         unsupported_parallelisms = [
```
guihong-nv: Enabling PP introduces duplicate embedding/output weights, but FullyShardedOptimizer.get_grad_norm() counts both copies in its WORLD reduction. Its replication correction only covers DP. A four-rank reproduction returned 7.071 instead of 5, causing excessive gradient clipping. Preserve the shared-parameter marker and exclude duplicate copies from gradient statistics.

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
shjwudp: Good point—thanks! Removed `reuse_existing`.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
shjwudp: Thanks! When `share_embeddings_and_output_weights=False`, counting the embedding and output-weight gradient norms twice is definitely incorrect.

Thanks for pointing this out—I’ll address it.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue, line 69: Question beyond this PR: do we still need this? IIUC, megatron_fsdp is packaged with megatron-core unconditionally.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -648,7 +655,15 @@ def __init__(
             forward_prefetch_size=ddp_config.suggested_communication_unit_size,
             backward_prefetch_size=ddp_config.suggested_communication_unit_size,
         )
-        with fully_shard_context(device=device, use_symmetric_memory=ddp_config.nccl_ub):
+        # Join the caller's ambient context when one is active (VPP chunks); otherwise
+        # open and finalize our own.
+        active_context = current_fully_shard_context()
```
wujingyue, line 660: Note: we could pass in the context from the constructor if we didn't have to pay the "inheritance tax" -- The API is quite coupled with DDP at this very moment.

tests/unit_tests/distributed/mfsdp_v2/test_mfsdp_v2_vpp_shared_context.py
wujingyue, line 1: test_pipeline_parallel.py

tests/unit_tests/distributed/mfsdp_v2/test_mfsdp_v2_vpp_shared_context.py
```diff
@@ -0,0 +1,110 @@
...
+from megatron.core.process_groups_config import ProcessGroupCollection
+from megatron.core.transformer import TransformerConfig
+from tests.unit_tests.test_utilities import Utils
+
+pytestmark = pytest.mark.skipif(
+    torch.cuda.device_count() < 2, reason="Requires at least 2 CUDA devices."
+)
+
+
+class _VppChunk(nn.Module):
```
wujingyue: Can you avoid abbreviation? Make code friendly to new readers.

tests/unit_tests/distributed/mfsdp_v2/test_mfsdp_v2_vpp_shared_context.py
```diff
@@ -0,0 +1,110 @@
...
+from megatron.core.distributed.fsdp.src.megatron_fsdp.experimental.fully_shard import (
+    current_fully_shard_context,
+)
+from megatron.core.distributed.fsdp.src.megatron_fsdp.experimental.module import FsdpModule
+from megatron.core.process_groups_config import ProcessGroupCollection
+from megatron.core.transformer import TransformerConfig
+from tests.unit_tests.test_utilities import Utils
+
+pytestmark = pytest.mark.skipif(
+    torch.cuda.device_count() < 2, reason="Requires at least 2 CUDA devices."
```
wujingyue: Don't do this globally -- it's sticky. Move it to individual tests even though it leads to duplication.

tests/unit_tests/distributed/mfsdp_v2/test_mfsdp_v2_vpp_shared_context.py
```diff
@@ -0,0 +1,110 @@
...
+
+            # The shared context was finalized exactly once after the loop, so it is
+            # ready to drive forward/backward without an extra finalize.
+            contexts[0].ensure_finalized()
+
+            # The helper left no ambient context behind: the chunks joined its scope and
+            # that scope, which owned the only finalize call, has since exited.
+            assert current_fully_shard_context() is None
+        finally:
+            Utils.destroy_model_parallel()
```
wujingyue: Move init to setup and this to teardown.

tests/unit_tests/distributed/mfsdp_v2/test_mfsdp_v2_vpp_shared_context.py
```diff
@@ -0,0 +1,110 @@
...
+            pg_collection = ProcessGroupCollection.use_mpu_process_groups()
+
+            # virtual_pipeline_model_parallel_size > 1 yields multiple chunks on this
+            # rank; model two VPP sub-stages here. The ambient fully_shard_context is
+            # opened for every MFSDP v2 wrap, regardless of the chunk count.
+            chunks = [
+                _VppChunk().to(device=device, dtype=torch.bfloat16) for _ in range(2)
+            ]
+
+            wrapped = wrap_model_chunks_with_ddp(
```
wujingyue:
```suggestion
            wrapped_chunks = wrap_model_chunks_with_ddp(
```

tests/unit_tests/distributed/mfsdp_v2/test_mfsdp_v2_vpp_shared_context.py
```diff
@@ -0,0 +1,110 @@
...
+
+            # Every chunk is an MFSDP v2 wrapper whose root is an FsdpModule.
+            assert len(wrapped) == 2
+            for chunk in wrapped:
+                assert isinstance(chunk, FullyShardedDataParallelV2)
+                assert isinstance(chunk.module, FsdpModule)
+
+            # The whole point of the feature: all VPP chunks share ONE context.
+            contexts = [chunk.module.context for chunk in wrapped]
+            assert contexts[0] is contexts[1]
```
wujingyue: A bit fragile because people could update the test to use more chunks but not this check. Something like `for each context assert it's the same as the first` would work better.

wujingyue (approved):

wujingyue:
Looks great! I only read the FSDP part. Someone else needs to review the PP logic.

guihong-nv (approved):

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: Are you saying there's a bug in FullyShardedOptimizer.get_grad_norm() when handling tied weights? Can you file a bug with a minimal, non-mcore repro so I can take a look? It doesn't seem to be PP-specific and can affect everything.

shjwudp: In MCore’s DP path, [get_grads_for_grad_norm](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/optimizer/optimizer.py#L260) selects the gradients eligible for gradient-norm computation. For shared weights, it uses [param_is_not_shared](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/module.py#L26) to identify shared parameters; gradients associated with shared parameters are excluded from the calculation.

Phlip79 (approved):

shjwudp:
@NVIDIA/training-adlr @NVIDIA/training-nemo @NVIDIA/core-adlr — could you please review this PR when you have a chance? Thanks!

megatron/core/pipeline_parallel/schedules.py
```diff
@@ -197,7 +197,6 @@ def deallocate_output_tensor(out, deallocate_pipeline_outputs=False):
 
     # Base case: deallocate tensor
     assert isinstance(out, torch.Tensor), "expected Tensor, found %s." % type(out).__name__
-    assert out._base is None, "counter-productive to free a view of another tensor."
```
jaredcasper, line 200: Why was this removed?

maanug-nv (approved):

megatron/core/pipeline_parallel/schedules.py
shjwudp: The line was removed because it asserts a precondition that the schedule can legitimately violate.

With VPP + MFSDP v2, the FSDP module’s PyTorch backward hook causes a chunk’s output hidden states to become a view (`grad_fn=BackwardHookFunctionBackward` with `_base` set). As a result, this assertion aborts on the first warmup call to `deallocate_output_tensor`.

Minimal reproducer:

```python
import torch
import torch.nn as nn

m = nn.Linear(1024, 1024)
m.register_full_backward_pre_hook(
    lambda module, grad_output: None
)  # Mirrors what FsdpModule does.

x = torch.randn(4096, 1024, requires_grad=True)
y = m(x)

print(y._base is None, type(y.grad_fn).__name__, tuple(y._base.shape))
# False BackwardHookFunctionBackward (4096, 1024)
```

So, in this configuration, I remove this line to make `deallocate_output_tensor` handle output tensors that are views rather than assuming `_base is None`.

jaredcasper: But then isn't the correct thing to do not try to deallocate it? If this is a view and we set out.data to an empty tensor, what happens to the original tensor. This feels like it could easily introduce a very nasty hard to find bug. The assert is there for a reason, just because you have a case that fires it doesn't mean it's okay to just remove the assert.

yashaswikarnati: +1, i think we would need this assert.

shjwudp: Hi @jaredcasper , @yashaswikarnati  — thanks for the review!

I removed the check because I wanted to keep the deallocate-output optimization
enabled for M-FSDP v2. In the configuration I tested, `out._base` is present and
swapping `out.data` leaves the base tensor untouched, so it looked safe. But you're
right that it isn't safe in general: if an operator saves the output in its
autograd context, the swap silently corrupts it and produces a correctness issue.
Deleting the check was not the right fix.

I've changed it to disable deallocate-output for M-FSDP v2 instead, which is
semantically unambiguous and safe.

Fix commit: https://github.com/NVIDIA/Megatron-LM/pull/6486/changes/0fd684e79c6b1ff3862ca09692f6f1c297a10907

jaredcasper (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
```diff
@@ -80,7 +92,8 @@ def fully_shard_context(
             communication stream to reduce peak transient memory. See
             https://github.com/NVIDIA/Megatron-LM/issues/6471.
     """
-    if _FSDP_CONTEXT.get() is not None:
+    existing = _FSDP_CONTEXT.get()
+    if existing is not None:
```
wujingyue: Does this change anything?

wujingyue (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/fully_shard.py
shjwudp: Fixed.

PR #6489, opened 2026-08-12
fix(optimizer): support MFSDP v2 gradient clipping across dense and expert meshes

wujingyue:
## Summary

MFSDP v2 gradients are DTensors, and dense and expert gradients do not share a device
mesh: with EP=2 over eight ranks the dense gradients span all eight while the expert
gradients span the four-rank expert-DP stripe. The shared gradient-statistics helpers
assume a single mesh, so `FullyShardedOptimizer` computes its own statistics from each
gradient's placements.

- `clip_grads.py`: accept BF16 DTensor gradients in the FP32 clipping assert.
- `optimizer.py`: route `clip_grad_norm` through `self.get_grad_norm()` so a subclass
  clips by the norm it reports. The base implementation is unchanged.
- `optimizer/__init__.py`: reduce MFSDP v2 gradient statistics over the world group.
  `get_grad_norm` divides each contribution by its replication factor, so summing over
  every rank is correct by construction. Inert today, since the two groups coincide.
- `fully_sharded_optimizer.py`: override `get_grad_norm` and `count_zeros`, and drop the
  `__fsdp_param__` marker.
- `mcore_fsdp_adapter.py`: reject `mtp_detach_heads`. Behavior change — that config
  previously trained, with the silently inflated group norm described below.
- tests: dense and expert-parallel clipping coverage.

## Why override `get_grad_norm` instead of inheriting it

The inherited path ends in `get_grad_norm_fp32`, which needs local tensors. Given
DTensors it asserts they all share one device mesh — dense and expert do not, so EP
fails outright — and when it does find a mesh group it reduces over that group *in
addition to* the grad-stats group. For MFSDP v2 those are the same group, so the squared
norm is summed twice and the reported norm is inflated by `sqrt(world_size)`, silently.

That unwrapping previously depended on the `__fsdp_param__` marker — a side channel,
and one more thing to keep in sync. Reading each gradient's `placements` drops both the
marker and the single-mesh assumption, and divides out replicated axes instead of
double-counting them (inert today; it matters under HSDP).

## Validation

On 8 GPUs:

| Suite | Result |
| --- | --- |
| `distributed/mfsdp_v2/` | 113 passed |
| `test_optimizer.py` | 33 passed, 35 skipped |

`black`, `isort` and `pylint` clean at the versions CI pins.

The world-group commit came after those runs and was checked separately on 4 ranks
(`test_optimizer` 2 passed, `test_mcore_adapter` 7 passed). Note that no test can
distinguish the two groups today: MFSDP v2 rejects tensor, pipeline and context
parallelism, so the data-parallel group already spans the world.

### DIFF
--- a/megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
+++ b/megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
@@ -636,6 +636,15 @@ class FullyShardedDataParallelV2(_BaseDataParallel):
         if not hasattr(pg_collection, 'dp_cp'):
             raise ValueError("MFSDP v2 requires an explicit dp_cp process group.")
 
+        if config.mtp_detach_heads:
+            # Detached MTP heads are tagged into a separate gradient-norm group, which
+            # the shared helpers reduce with get_grad_norm_fp32. That path derives a
+            # data-parallel group from the gradients' DTensor mesh and reduces over it
+            # in addition to the grad-stats group, double-counting for MFSDP v2 whose
+            # mesh already is the grad-stats group. Reject it rather than report a
+            # silently inflated norm.
+            raise ValueError("MFSDP v2 does not currently support mtp_detach_heads.")
+
         unsupported_parallelisms = [
             "tensor_model_parallel_size",
             "pipeline_model_parallel_size",
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
@@ -232,6 +232,8 @@ class FsdpParameterGroup:
             sharded_parameter = nn.Parameter(
                 self.main_weight.get_dtensor(index), requires_grad=parameter.requires_grad
             )
+            # Route optimizer gradient statistics through the FSDP local-shard path.
+            sharded_parameter.__fsdp_param__ = True
             if main_grad_dtype:
                 sharded_parameter.grad_dtype = main_grad_dtype
             setattr(sharded_parameter, _CONTAINING_PARAMETER_GROUP_ATTR, ref(self))
--- a/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
+++ b/megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
@@ -232,8 +232,6 @@ class FsdpParameterGroup:
             sharded_parameter = nn.Parameter(
                 self.main_weight.get_dtensor(index), requires_grad=parameter.requires_grad
             )
-            # Route optimizer gradient statistics through the FSDP local-shard path.
-            sharded_parameter.__fsdp_param__ = True
             if main_grad_dtype:
                 sharded_parameter.grad_dtype = main_grad_dtype
             setattr(sharded_parameter, _CONTAINING_PARAMETER_GROUP_ATTR, ref(self))
--- a/megatron/core/optimizer/fully_sharded_optimizer.py
+++ b/megatron/core/optimizer/fully_sharded_optimizer.py
@@ -19,6 +19,28 @@ from .optimizer import MixedPrecisionOptimizer
 from .optimizer_config import OptimizerConfig
 
 
+def count_replication(tensor: DTensor) -> int:
+    """Return how many ranks hold an identical copy of ``tensor``'s local shard.
+
+    A sharded mesh axis holds disjoint pieces that must all be counted; a replicated
+    axis holds identical copies that must be counted once, so a gradient statistic
+    summed over the grad-stats group has to divide by this.
+
+    MFSDP v2 gradients are always DTensors, so this takes one rather than accepting
+    a plain tensor and guessing a layout for it.
+    """
+    replication = 1
+    for axis, placement in enumerate(tensor.placements):
+        if placement.is_replicate():
+            replication *= tensor.device_mesh.size(axis)
+        elif placement.is_partial():
+            raise RuntimeError(
+                "MFSDP v2 gradient is still Partial when gradient statistics are taken; "
+                "the reduction must be finalized first."
+            )
+    return replication
+
+
 class FullyShardedOptimizer(MixedPrecisionOptimizer):
     """MCore optimizer wrapper for MFSDP-owned sharded parameters and gradients.
 
@@ -107,28 +129,6 @@ class FullyShardedOptimizer(MixedPrecisionOptimizer):
         """Build a sharded optimizer state dict."""
         raise NotImplementedError("MFSDP v2 optimizer checkpointing is not yet supported.")
 
-    @staticmethod
-    def count_replication(tensor: DTensor) -> int:
-        """Return how many ranks hold an identical copy of ``tensor``'s local shard.
-
-        A sharded mesh axis holds disjoint pieces that must all be counted; a replicated
-        axis holds identical copies that must be counted once, so a gradient statistic
-        summed over the grad-stats group has to divide by this.
-
-        MFSDP v2 gradients are always DTensors, so this takes one rather than accepting
-        a plain tensor and guessing a layout for it.
-        """
-        replication = 1
-        for axis, placement in enumerate(tensor.placements):
-            if placement.is_replicate():
-                replication *= tensor.device_mesh.size(axis)
-            elif placement.is_partial():
-                raise RuntimeError(
-                    "MFSDP v2 gradient is still Partial when gradient statistics are taken; "
-                    "the reduction must be finalized first."
-                )
-        return replication
-
     def get_grad_norm(self):
         """Compute the global gradient L2 norm from each gradient's own DTensor layout.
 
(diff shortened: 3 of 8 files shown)


shjwudp (approved):

megatron/core/distributed/fsdp/src/megatron_fsdp/experimental/parameter_group.py
```diff
@@ -223,6 +223,8 @@ def __init__(
             sharded_parameter = nn.Parameter(
                 self.main_weight.get_dtensor(index), requires_grad=parameter.requires_grad
             )
+            # Route optimizer gradient statistics through the FSDP local-shard path.
```
wujingyue: This trick has been used for V1 and works. But I'm also exploring alternatives...

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -105,6 +132,82 @@ def sharded_state_dict(
...
+        )
+        for parameter in self.get_parameters():
+            # MFSDP v2 reduces into parameter.grad; it never populates decoupled_grad,
+            # which is a v1 param-and-grad-buffer concept.
+            grad = parameter.grad
+            if grad is None:
+                continue
+            replication = count_replication(grad)
+            local_grad = grad.to_local()
+            if local_grad.numel():
```
skyw: Prefer more explicit `local_grad.numel() > 0` or `local_grad.numel() is not None`

Q: would numel() return None? or 0 for empty tensor?

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -105,6 +132,82 @@ def sharded_state_dict(
...
+                total_norm_squared += local_grad.float().pow(2).sum() / replication
```
skyw, line 168: Nit: total_norm_squared  is not entirely accurate. total_grad_square_sum is more accurate. Give it is probably widely used already, fine to keep as it is.

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -105,6 +132,82 @@ def sharded_state_dict(
...
+
+        torch.distributed.all_reduce(
+            total_norm_squared,
+            op=torch.distributed.ReduceOp.SUM,
+            group=self.get_grad_stats_parallel_group(),
```
skyw, line 173: Q: get_grad_stats_parallel_group is verified to always return the right group? e.g. when experts and attention are in different mesh?

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -105,6 +132,82 @@ def sharded_state_dict(
...
+        total_zeros = torch.zeros((), dtype=torch.float32, device=torch.cuda.current_device())
+        for parameter in self.get_parameters():
+            grad = parameter.grad
+            if grad is None:
+                continue
+            replication = count_replication(grad)
+            local_grad = grad.to_local()
+            if local_grad.numel():
+                zeros = local_grad.numel() - torch.count_nonzero(local_grad)
+                total_zeros += zeros.float() / replication
```
skyw, line 196: Q: Is this always divisible? If yes, any reason to not use integer type? total_zeros can have fraction in floating point.

skyw (approved):

megatron/core/optimizer/fully_sharded_optimizer.py
wujingyue: Done

megatron/core/optimizer/fully_sharded_optimizer.py
wujingyue: I'll keep the name as is. The containing function `get_grad_norm` is pretty clear that the norm here is grad norm.

megatron/core/optimizer/fully_sharded_optimizer.py
wujingyue: Good question -- that's what I've been struggling to get correct in this PR and why I don't like the base implementation. 

At this moment, `get_grad_stats_parallel_group` is hardcoded to dp_cp for MFSDP v2, which happens to be the right process group for both attention and MoE. After we switch to use single-grouped expert weights, the disciplined way is to simply pass in the full group of the mesh, e.g. `mesh.mesh.flatten().tolist()`.

megatron/core/optimizer/fully_sharded_optimizer.py
wujingyue: Good catch! No, it's not always divisible. This is probably fine for count_zeros which is only used for debugging, before I can figure out a better way.

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
```diff
@@ -624,6 +636,15 @@ def _validate_config(
...
             raise ValueError("MFSDP v2 requires an explicit dp_cp process group.")
 
+        if config.mtp_detach_heads:
+            # Detached MTP heads are tagged into a separate gradient-norm group, which
+            # the shared helpers reduce with get_grad_norm_fp32. That path derives a
+            # data-parallel group from the gradients' DTensor mesh and reduces over it
+            # in addition to the grad-stats group, double-counting for MFSDP v2 whose
+            # mesh already is the grad-stats group. Reject it rather than report a
+            # silently inflated norm.
+            raise ValueError("MFSDP v2 does not currently support mtp_detach_heads.")
```
Phlip79, line 646: Are there plans to support this in the future?

Phlip79 (approved):

yaoyu-33 (approved):

megatron/core/distributed/fsdp/mcore_fsdp_adapter.py
wujingyue: Not yet. I have some ideas but I'd love to see a concrete example before I speculate a fix. Who should I talk to about MTP?

megatron/core/optimizer/optimizer.py
```diff
@@ -379,13 +379,11 @@ def clip_grad_norm(self, clip_grad: float) -> float:
...
         params = self.get_parameters()
-        if params:
-            grads_for_norm = self.get_grads_for_grad_norm()
-        else:
-            grads_for_norm = []
-        grad_norm = get_grad_norm_fp32(
-            grads_for_norm, grad_stats_parallel_group=self.get_grad_stats_parallel_group()
-        )
+        # Go through get_grad_norm() so a subclass that computes the norm differently
+        # clips by the value it reports. The base implementation is unchanged. Called
```
jaredcasper: Weird to say "unchanged" in a comment, without context of being in a PR what does that even mean?

megatron/core/optimizer/fully_sharded_optimizer.py
```diff
@@ -105,6 +132,82 @@ def sharded_state_dict(
...
+        Each rank contributes ``||local||^2`` divided by the product of its replicated
+        mesh-axis sizes. A sharded axis holds disjoint pieces that must all be added; a
+        replicated axis holds identical copies that must be counted once. Summing that
+        over the grad-stats group is then exact, because every shard is held by exactly
+        one rank in that group.
+
+        No duplicate filtering is needed here. ``get_grad_norm_fp32``'s shared/TP/GTP
+        filters exist to count replicas once along the model-parallel axes, and MFSDP v2
+        rejects TP, PP and CP sizes above 1 (see ``mcore_fsdp_adapter``), so those axes
+        are trivial and every parameter is unique. Revisit if v2 gains TP composability.
```
jaredcasper: Should this assert here that TP, PP, and CP are 1... if that changes in mcore_fsdp_adapter but not here could lead to trouble?

megatron/core/optimizer/optimizer.py
wujingyue: Agreed — that comment described the change rather than the resulting code, so it doesn't read correctly outside the PR. Removed in `5c8bd7f96`.

On why the delegation is there: routing through `get_grad_norm()` gives `FullyShardedOptimizer` an interception point so it can compute the norm from each gradient's DTensor layout. It doesn't affect any other optimizer.

1. The other classes that reach `clip_grad_norm` are `Float16OptimizerWithFloat16Params`, `DistributedOptimizer` (both via `MixedPrecisionOptimizer.step`) and `FP32Optimizer`. None of them overrides `get_grad_norm`, so they resolve to `MegatronOptimizer.get_grad_norm` — the same `get_grads_for_grad_norm()` + `get_grad_norm_fp32()` that was inlined here. The dropped `if params:` guard was redundant, since `get_grads_for_grad_norm()` already returns `[]` for an empty param list.
2. The classes that *do* override `get_grad_norm` — `ChainedOptimizer`, `LayerWiseDistributedOptimizer`, `MimoOptimizer` — never call `clip_grad_norm`. Each has its own `step` that calls `get_grad_norm()` directly and clips inline.

One knock-on worth flagging: `clip_grad_norm` now requires `get_grad_norm` on `self`, which is why the duck-typed mock in `test_mtp_grad_clipping_uses_separate_norms` had to borrow it as well.

More broadly, we're paying some [inheritance tax](https://medium.com/@b.stoilov/why-you-should-think-twice-before-using-inheritance-573545dcd8e1) here: the flow of `def step` jumps back and forth between base and subclass. Out of scope for this PR — happy to follow up if you have a cleanup direction.

megatron/core/optimizer/fully_sharded_optimizer.py
wujingyue: Thanks -- this paragraph is yet another slop and I removed it.  

The code here is in fact general enough to work with TP/CP/PP parameters as long as they are DTensors. Being DTensors is enforced in `def count_replication`.

jaredcasper (approved):

PR #6497, opened 2026-08-12
Inference: Fixes for large scale and concurrency

sidsingh-nvidia:

# What does this PR do?

## Coordinator wire-format change for faster routing
Splits the coordinator's ZMQ wire format into **multipart frames**: a small, constant-size metadata frame followed by opaque body frames. The coordinator unpacks and repacks only the metadata frame and forwards the bodies untouched. The metadata is of a small constant size, independent of the sequence length.
Todo - apply the same fix for multimodal. Move large multimodal data to the body from the metadata.

## Hash prompts on the client, not the coordinator
Hashing is embarassingly parallel and can be handled on frontend replicas - each of which spawns a client. 

## Apply a weight versioned salt to prefix hashes in the engine
This prevents prompts in RL from matching with stale prefix caches. 

## Change routing cost function
A linear combination of the prefix match score and the relative load on each engine. Works well in practice and is able to balance load and affinity.

Load is measured against the **fleet mean** rather than against `max_requests`. Against a configured ceiling a sizeable imbalance stays a small fraction of it, so affinity keeps winning however lopsided the fleet gets; against the mean the term vanishes while ranks are even and grows only as they diverge — the drain at the end of a batch, when work should spread. It subtracts rather than blending, so a full cache hit cannot cancel the load term and strand work on a saturated rank.

Prefix depth is now graded per rank and counted forward from the first block. The previous match credited only the ranks holding the *deepest* cached block, so a rank holding 8 of 10 blocks scored the same as one holding none, leaving the load term nothing to trade against. Counting forward also stops a rank being credited for a deep block whose prefix has been evicted — unusable KV, since the hashes chain.

## Expire coordinator hash entries after a TTL
The coordinator's hash table is a guess about what each engine still holds: it sees blocks being routed, never blocks being evicted, so the guess only gets more optimistic and the table grows without bound. Entries untouched for `prefix_cache_ttl_seconds` (default 300) are dropped. The queue is insertion-ordered, so the sweep stops at the first live entry and costs what it evicts rather than the size of the table.

## Behaviour changes worth flagging

**`prefix_caching_routing_alpha` changes meaning.** It was a blend weight in `[0, 1]`; it is now a dimensionless, uncapped coefficient on the load term. A value tuned for the old scoring does not carry over — and at `1.0` a single request of imbalance across two ranks exactly cancels a full cache hit, so the default stays `0.5`.

**`detokenize_generations` is a new `SamplingParams` field, defaulting to `True`.** The coordinator used to detokenize every reply unconditionally. Now that reply bodies are opaque frames, it detokenizes only when asked — defaulting on so no existing client changes behaviour. Clients that detokenize for themselves (the HTTP frontend does) can opt out and let the coordinator forward the body without decoding and re-encoding it.

New knobs: `--inference-dynamic-batching-prefix-cache-ttl-seconds`.

## Issue tracking

Linked issue: 

### Pre-checks

Existing coordinator, client, streaming, engine, and prefix-caching-coordinator unit tests were updated to the new framing, and roughly 25 tests added across the pieces:

- **Client-side hashing** — the client hashes and salts multimodal with the media key, string prompts are left unhashed, `LOAD_BALANCED` skips hashing, prefix routing reads those hashes without decoding the prompt (using a deliberately undecodable prompt frame, so a decode attempt would raise), an empty hash list is not treated as a fallback and a null frame is.
- **Weight salt** — generations are disjoint, epoch 0 is byte-identical to unsalted, media identity stays distinct within a generation, a text request admitted after a resume cannot match blocks from before it, and caller-supplied hashes are adopted rather than re-salted (a disaggregated handoff depends on this).
- **Cost function** — one test per property: alpha=0 is pure affinity, mild imbalance does not overturn a hit, a saturated rank loses to an idle one, a balanced fleet ignores load entirely, deeper prefixes outscore shallower ones, a deep block without its prefix earns nothing, and ties go to the least loaded rank.
- **TTL** — expiry, refresh-on-reuse, an expired entry ceasing to win routing, and survival across an engine removal (which renumbers rank indices, so the expiry queue is deliberately keyed on the hash rather than the rank).
- **Framing robustness** — a malformed submission is dropped rather than raising out of the event loop, and `handle_engine_reply` is covered in both the detokenizing and forward-untouched cases.

Verified against `main` at `a3344a6f`: across all of `tests/unit_tests/inference/`, including `engines/`, the set of failing tests is identical with and without this branch (218 either side), and this branch passes 25 more tests than `main`.

### Code review

#### Step 2: Final Review

#### Step 3: Approved

### Merge


wdykas (approved):
Looks good from disag side

santhnm2 (approved):
LGTM, but as a suggestion for a future PR: the design here is passing around lists of frames where frames are arbitrary packed bytes, which requires indexing of the frame list at certain indices and doesn't assign any semantic meaning to each value in the list. Would it make sense to have a wrapper class so we could refer to the header frame more explicitly?

sidsingh-nvidia:
@santhnm2 that's a fair critique. I can impose the following structure on all messages
1. each message can only consist of two frames - [metadata, payload]
2. metadata and payload will be dicts with bespoke keys for easy readabiity - this can remove seemingly arbitrary indexing.

santhnm2:
> @santhnm2 that's a fair critique. I can impose the following structure on all messages
> 
> 1. each message can only consist of two frames - [metadata, payload]
> 2. metadata and payload will be dicts with bespoke keys for easy readabiity - this can remove seemingly arbitrary indexing.

I think this makes sense. An even tighter contract might be a custom dataclass with `metadata` and `payload` fields, but that's a small design decision I'll defer to you

wdykas:
> Since we wanna pack metadata and payload separately, we will have to separate them out prior to the send operation.

do you know what we are allowed to serialize with msg pack? I thought it was real stingy

kvareddy (approved):

nvcsathe (approved):

shanmugamr1992 (approved):

sidsingh-nvidia:
/ok to test [e4edf35](https://github.com/NVIDIA/Megatron-LM/pull/6497/commits/e4edf351c3ab4784db16e99060817bc2bc9eb55e)

sidsingh-nvidia:
/ok to test [ee1c419](https://github.com/NVIDIA/Megatron-LM/pull/6497/commits/ee1c4194a1aa3c43cc7838714c6053d10e81ca61)

sidsingh-nvidia:
/ok to test [95a59ce](https://github.com/NVIDIA/Megatron-LM/pull/6497/commits/95a59ce3e4cc4316a6a5f9f569e48b3184e5ad9b)

jaredcasper (approved):

asolergi-nv (approved):

sidsingh-nvidia:
/ok to test [cdddde3](https://github.com/NVIDIA/Megatron-LM/pull/6497/commits/cdddde3562e3ead60991af2c0dbc809b6c955ba9)

cspades:
/ok to test [e15f6a8](https://github.com/NVIDIA/Megatron-LM/pull/6497/commits/e15f6a8143fc6497d3b44a60d3012b93259f3aa6)

sidsingh-nvidia:
/ok to test [a8125e4](https://github.com/NVIDIA/Megatron-LM/pull/6497/commits/a8125e4544921648438687af1ad0e4b06813588e)

sidsingh-nvidia:
/ok to test [fbb4d54](https://github.com/NVIDIA/Megatron-LM/pull/6497/commits/fbb4d54350de3ab6eef9000dc17c0e4e77118614)

sidsingh-nvidia:
/ok to test [c7bb112](https://github.com/NVIDIA/Megatron-LM/pull/6497/commits/c7bb112f77bee09f037b1e7b51ea471b62478575)

PR #5382, opened 2026-06-16
Add generic interface for SSM inference

santhnm2:
- [x] I, the PR author, have personally reviewed every line of this PR.

# What does this PR do ?
Makes the following refactoring changes (no functionality changes):
1. Factors out common decode / prefill split logic for linear attention dynamic inference into `ssm_inference.py`
2. Re-organizes Mamba2 dynamic inference code to use the generic `ssm_decode` and `ssm_prefill` interface functions
3. Quarantines the legacy static Mamba2 inference code into `_static_decode` and `_static_prefill` functions.

:warning: For major changes (either in lines of code or in its impact), please make sure to first share a design doc with the team. If you're unsure what's the best way to do so, contact @NVIDIA/mcore-oncall.

## Issue tracking

For PRs from open-source community contributors:

- **New features**: a linked issue is **required**. Please open a [feature request](https://github.com/NVIDIA/Megatron-LM/issues/new?template=feature_request.md) and reference it here before submitting the PR.
- **Small updates (bug fixes, minor improvements)**: a linked issue is **recommended** and will accelerate the PR review process.

Linked issue: 

### Pre-checks

- [ ] I have added relevant unit tests
- [ ] I have added relevant functional tests
- [ ] I have added proper typing to my code [Typing guidelines](https://docs.python.org/3/library/typing.html)
- [ ] I have added relevant documentation
- [ ] I have run the [autoformatter.sh](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/autoformat.sh) on my PR

### Code review

Feel free to message or comment @NVIDIA/mcore-oncall to help accelerate your merge into main. The less complex your PR is, the faster it will be approved and merged!

All PRs start as **draft**. If you open a non-draft PR, it will be automatically converted to draft.

#### Step 1: Mark PR as "Ready for Review"

1. When your PR is ready, click **Ready for Review**.
2. An oncall reviewer is auto-assigned and expert reviewers are notified based on your changes.
   - Some PRs may jump straight to step 2. This is determined by `.github/CODEOWNERS`.

:warning: Only mark as ready once merge-conflicts are resolved and the CI is passing.
Final Review might get declined if these requirements are not fulfilled.

#### Step 2: Final Review

For PRs that change `megatron/core`, once all expert reviewers have approved, the `Final Review` label is applied **automatically** and final reviewers are assigned.

For PRs outside `megatron/core`, this step is skipped.

#### Step 3: Approved

[...]

PR #6499, opened 2026-08-13
Add GDN dynamic inference

anlthms:

# What does this PR do?

This implementation adds GDN prefill and single-token decode to the dynamic inference engine. It reuses the generic recurrent-mixer control flow introduced by [PR #5382](https://github.com/NVIDIA/Megatron-LM/pull/5382) and uses the inference kernels from flash-linear-attention (FLA) 0.5.1.

### Pre-checks

## Notes
For a detailed description of the changes, please see [Gated DeltaNet Dynamic Inference](https://docs.google.com/document/d/19vAUjIKocJmN89z2DsIGziE1P4RiuikzPl79XEuRWqE/edit?tab=t.0#heading=h.3jjrj6nrzhe4) (internal document)

### DIFF
--- a/megatron/core/inference/contexts/dynamic_context.py
+++ b/megatron/core/inference/contexts/dynamic_context.py
@@ -441,21 +441,22 @@ class DynamicInferenceContext(BaseInferenceContext):
                     "boundaries are not rounded between decode chunks."
                 )
 
-            # For hybrid models, the layer map converts the global layer index to the
-            # corresponding attention layer index or Mamba layer index depending on the
-            # layer type.
-            attention_layer_map, dsa_layer_map, gdn_layer_map, mamba_layer_map = (
-                operator.itemgetter(
-                    Symbols.ATTENTION, Symbols.DS_ATTENTION, Symbols.GDN, Symbols.MAMBA
-                )(get_layer_maps_from_layer_type_list(mamba_inference_state_config.layer_type_list))
-            )
-
-            if len(gdn_layer_map) > 0:
-                raise NotImplementedError("GDN layers are not supported for inference.")
+            # Mamba and GDN use the same slot-indexed recurrent-state cache contract. Build
+            # one map in global layer order; independently generated per-symbol maps both
+            # start at zero and would alias if they were simply unioned.
+            attention_layer_map, dsa_layer_map = operator.itemgetter(
+                Symbols.ATTENTION, Symbols.DS_ATTENTION
+            )(get_layer_maps_from_layer_type_list(mamba_inference_state_config.layer_type_list))
+            recurrent_layer_map = {}
+            for global_layer_idx, layer_type in enumerate(
+                mamba_inference_state_config.layer_type_list
+            ):
+                if layer_type in (Symbols.MAMBA, Symbols.GDN):
+                    recurrent_layer_map[global_layer_idx] = len(recurrent_layer_map)
 
             self.num_attention_layers = len(attention_layer_map) + len(dsa_layer_map)
-            self.num_mamba_layers = len(mamba_layer_map)
-            self.layer_map = attention_layer_map | dsa_layer_map | mamba_layer_map
+            self.num_mamba_layers = len(recurrent_layer_map)
+            self.layer_map = attention_layer_map | dsa_layer_map | recurrent_layer_map
         else:
             # The layer map is the identity function for pure Transformer models.
             # Use the same per-PP-rank layer count as TransformerBlock (handles
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,6 @@ dev = [
     "multi-storage-client~=0.50",
     "opentelemetry-api>=1.33.1,<2",
     "nemo-lens",
-    "flash-linear-attention==0.5.1",
     "megatron-energon[av_decode]~=7.0",
     "av",
     "flashinfer-python>=0.5.0,<0.7.0",
@@ -134,6 +133,7 @@ te = [
 ssm = [
     "mamba-ssm",
     "causal-conv1d~=1.5",
+    "flash-linear-attention==0.5.1",
 ]
 
 [dependency-groups]
--- a/tests/unit_tests/inference/engines/test_dynamic_engine.py
+++ b/tests/unit_tests/inference/engines/test_dynamic_engine.py
@@ -16,6 +16,7 @@ from unittest import mock
 import msgpack
 import pytest
 import torch
+import torch.nn.functional as F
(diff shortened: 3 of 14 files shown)


megatron/core/inference/contexts/dynamic_context.py
```diff
@@ -438,21 +438,22 @@ def __init__(self, model_config: TransformerConfig, inference_config: InferenceC
...
+            # start at zero and would alias if they were simply unioned.
+            attention_layer_map, dsa_layer_map = operator.itemgetter(
+                Symbols.ATTENTION, Symbols.DS_ATTENTION
+            )(get_layer_maps_from_layer_type_list(mamba_inference_state_config.layer_type_list))
+            recurrent_layer_map = {}
+            for global_layer_idx, layer_type in enumerate(
+                mamba_inference_state_config.layer_type_list
+            ):
+                if layer_type in (Symbols.MAMBA, Symbols.GDN):
+                    recurrent_layer_map[global_layer_idx] = len(recurrent_layer_map)
```
santhnm2, line 455: Not really related to this PR, but I'm curious why GDN has a specific layer type but GDP did not - should we be unifying all of these linear attention variants under a single layer type?

tests/unit_tests/inference/engines/test_dynamic_engine.py
```diff
@@ -432,7 +434,8 @@ def _build_test_env(cls, test_config):
                 mtp_block_spec=mtp_block_spec,
                 position_embedding_type=test_config.position_embedding_type,
             ).cuda()
-        elif test_config.model_provider == "hybrid":
+        elif test_config.model_provider in ("hybrid", "gdn"):
```
santhnm2, line 438: I think `model_provider` will be "hybrid" for both mamba and GDN, can we parametrize on a different variable instead?

tests/unit_tests/ssm/conftest.py
santhnm2, line 1: Are these changes necessary? Seems like an artifact of how we're running the tests rather than something fundamental?

tests/unit_tests/ssm/test_gated_delta_net_inference.py
```diff
@@ -0,0 +1,180 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+"""Focused kernel and CUDA-graph tests for Gated DeltaNet dynamic inference."""
```
santhnm2, line 3: Is cuda graph functionality intended to work out of the box? For Mamba2 at least we needed to modify the kernel to support masking computation when we pass -1 values in the `batch_indices` tensor, do the GDN kernels already support this?

pyproject.toml
santhnm2, line 1: Was this change intentional?

megatron/core/inference/contexts/dynamic_context.py
anlthms: good question! I don’t know why GDP doesn’t have its own symbol. Symbol.GDN was already there and being used, so it made sense to just follow the pattern for this work. GDP piggybacks on `Symbol.Mamba`, I think.

tests/unit_tests/inference/engines/test_dynamic_engine.py
anlthms: when `model_provider` is set to “hybrid”, it really means Mamba. it is just a local test selector, so I am not sure if it is worth renaming.

tests/unit_tests/ssm/conftest.py
anlthms: this is fixing a crash that I observed. without this, every worker would quietly default to CUDA device 0 (unless something explicitly set the current device to something else).  with this new fixture, the workers pick the device indicated by LOCAL_RANK as one would want.

tests/unit_tests/ssm/test_gated_delta_net_inference.py
anlthms: writes use `tensor_masked_update`, which explicitly ignores -1.

for reads, there is a call to `batch_indices.clamp(min=0)` before gathering. this is present for both `ssm_decode()` and `ssm_prefill()` inside gdn.py.

pyproject.toml
anlthms: intentional. FLA was previously a dev dependency. that seemed wrong as it is a requirement for GDN (just like causal-conv1d, which was already under the ssm section). but let me know if you think FLA belongs in dev.

tests/unit_tests/ssm/test_gated_delta_net_inference.py
santhnm2: I see, so this is basically relying on the fact that multiple readers of the value at index 0 is safe as long as we don't have multiple writers. I think that's ok for now, but we should maybe document this explicitly.

pyproject.toml
santhnm2: It seems like we should bump this, but maybe @NVIDIA/mcore-oncall has thoughts here?

santhnm2 (approved):

kvareddy (approved):

tests/unit_tests/ssm/test_gated_delta_net_inference.py
anlthms: > as long as we don't have multiple writers

@santhnm2, to clarify: writes go through `tensor_masked_update`, which in turn calls `_tensor_masked_update_kernel_2d` (or 3d/4d). these kernels ignore an index of -1 with an explicit return.

```
    if target_idx == -1:
        return
```
(it is possible I misunderstood your comment)

santhnm2: Sorry let me clarify - my understanding is the `clamp` on `batch_indices` is basically forcing all padding entries to read from slot 0 and then do some dummy computation based on that value, but the result of the dummy computation is not written back to slot 0 because `tensor_masked_update` prevents writes on -1 slots. So what I meant by not having multiple writers is slot 0 will have a max of 1 writer (i.e., a real request mapped to slot 0) so this is safe.

The semantics for Mamba2 (and eventually GDP) are that within the kernel itself we skip computation for the -1 indices and directly write out 0 values rather than proceeding with the dummy computation. But the end result is effectively the same other than the output activation tensor having nonzero values in the padding slots. These padding slot outputs are ignored anyway so there should be no downstream effect.

anlthms: I get it now. the way Mamba2 handles padding sounds cleaner. probably worth changing this in another PR.

Phlip79 (approved):

PR #6506, opened 2026-08-13
Support mixed-precision optimizers in MFSDP v2

### LINKED ISSUE #5614: Integrate MFSDP v2 with Megatron Core training

wujingyue:
## Summary

- Enable `--use-precision-aware-optimizer` with Megatron-FSDP v2 so Transformer Engine FusedAdam receives `exp_avg_dtype` and `exp_avg_sq_dtype`.
- Keep MFSDP v2 on its FP32 main weights and direct reduced gradients; it does not use the V1-only `decoupled_grad` or FusedAdam master-weight path.
- Cast MFSDP v2 gradients to the parameter-data dtype for non-precision-aware optimizers before MCore clipping and stepping, then restore the original gradient dtype. This allows FusedSGD and other regular optimizers to use BF16 reduced gradients with FP32 main weights.

## Why

MFSDP v2 already has FP32 optimizer parameters and may reduce gradients in BF16. FusedAdam can consume that mixed-dtype pair directly, while regular optimizers such as FusedSGD require matching parameter and gradient dtypes. The MCore wrapper now selects the appropriate behavior from `use_precision_aware_optimizer`.

## Validation

- `python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/mfsdp_v2/test_mcore_adapter.py::TestMcoreAdapter::test_build_train_and_step --capture=fd`
- `python -m torch.distributed.run --nproc-per-node 2 -m pytest -q tests/unit_tests/distributed/mfsdp_v2/test_mcore_adapter.py::TestMcoreAdapter::test_fused_sgd_casts_mismatched_grads --capture=fd`
- `git diff --check`

### DIFF
--- a/megatron/core/optimizer/__init__.py
+++ b/megatron/core/optimizer/__init__.py
@@ -573,7 +573,9 @@ def _get_megatron_optimizer_based_on_param_groups(
                 # Otherwise, master weight will be managed by TransformerEngine.
                 # Delayed scaling is an exception because casting as well as the computation
                 # of the scaling factor can be conducted in the adam kernel.
-                if config.use_precision_aware_optimizer_no_fp8_or_ds_fp8:
+                if config.use_precision_aware_optimizer_no_fp8_or_ds_fp8 and not isinstance(
+                    model_chunks[0], FullyShardedDataParallelV2
+                ):
                     kwargs.update(
                         {
                             "master_weights": True,
@@ -1052,6 +1054,10 @@ def get_megatron_optimizer(
     if is_mfsdp_v2:
         if config.use_distributed_optimizer:
             raise ValueError("MFSDP v2 currently requires use_distributed_optimizer=False.")
+    elif config.use_precision_aware_optimizer and not config.use_distributed_optimizer:
+        raise ValueError(
+            "--use-precision-aware-optimizer only supported with distributed optimizer"
+        )
 
     # Separate out first model chunk if overlapping param AG with optimizer step.
     if config.overlap_param_gather_with_optimizer_step:
@@ -1142,6 +1148,7 @@ def get_megatron_optimizer(
             if (
                 not USING_PYTORCH_OPTIMIZER
                 and config.use_precision_aware_optimizer
+                and not is_mfsdp_v2
                 and getattr(optimizer_part.optimizer, "master_weights", None) is not None
             ):
                 # NOTE(@cspades): FusedAdam is provided Megatron-FSDP's main weights as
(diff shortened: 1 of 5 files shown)


megatron/core/optimizer/__init__.py
```diff
@@ -573,7 +573,10 @@ def _get_megatron_optimizer_based_on_param_groups(
                 # Otherwise, master weight will be managed by TransformerEngine.
                 # Delayed scaling is an exception because casting as well as the computation
                 # of the scaling factor can be conducted in the adam kernel.
-                if config.use_precision_aware_optimizer_no_fp8_or_ds_fp8:
+                if (
+                    config.use_precision_aware_optimizer_no_fp8_or_ds_fp8
+                    and not isinstance(model_chunks[0], FullyShardedDataParallelV2)
```
wujingyue: the `isinstance(model_chunks[0], FullyShardedDataParallelV2)` check is added here because MFSDP v2 still wants to respect `exp_avg_dtype` and `exp_avg_sq_dtype` for backward compatibility. 

I think a better way is to set `exp_avg_dtype` and `exp_avg_sq_dtype` regardless of `use_precision_aware_optimizer`. That would change the non-MFSDP-v2 behavior, so I didn't do that in this PR.

megatron/training/training.py
```diff
@@ -2097,12 +2096,13 @@ def get_megatron_ddp_config(args: argparse.Namespace) -> DistributedDataParallel
         kwargs["megatron_fsdp_main_params_dtype"] = args.megatron_fsdp_main_params_dtype
         kwargs["megatron_fsdp_main_grads_dtype"] = args.megatron_fsdp_main_grads_dtype
         kwargs["megatron_fsdp_grad_comm_dtype"] = args.megatron_fsdp_grad_comm_dtype
-        kwargs["megatron_fsdp_use_decoupled_grad"] = args.use_precision_aware_optimizer
-        if args.use_megatron_fsdp and args.megatron_fsdp_version == 2:
-            # MFSDP v2 gathers parameters from module hooks rather than the V1
-            # start_param_sync path, so disable the V1-only startup all-gather knob.
```
wujingyue, line 2102: This is no longer necessary. With https://github.com/NVIDIA/Megatron-LM/pull/6333, MFSDP v2 ignores the flag either way.

skyw (approved):
LGTM

Phlip79 (approved):

deepakn94 (approved):

maanug-nv (approved):

PR #6512, opened 2026-08-13
Parallelism aware Implementation of Hidden State Mixing for the MTP

YanivDorGalron:
This PR implements HSM (hidden states mixing) for the MTP.
HSM create a set of aligned hidden states - hidden stated that predict the same token position, and sample within this set for each position in the sequence. 

In order to create aligned hidden states the previous mtp outputs needs to be rolled by -1 every iterations. This require passing information from various ranks depending on the parallelism and how the sequence is scatter across ranks. This implementation takes that into account, construct a sequence layout according to TP/SP/CP/packed sequence and pass only the relevant information to the relevant rank

### DIFF
--- a/megatron/core/transformer/multi_token_prediction.py
+++ b/megatron/core/transformer/multi_token_prediction.py
@@ -65,6 +65,36 @@ else:
 from megatron.core.transformer.pipeline_parallel_layer_layout import PipelineParallelLayerLayout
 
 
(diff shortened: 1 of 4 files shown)


YanivDorGalron:
replaced with PR https://github.com/NVIDIA/Megatron-LM/pull/6566

PR #6528, opened 2026-08-14
Fix missing golden values updates

janEbert:
https://github.com/NVIDIA/Megatron-LM/pull/5916 missed some GitLab-only tests, so we add the missing updates using pipelines from the exact commit that was merged.

^ cross-checked every number with my slide rule

Joking aside, I made sure all files updated here weren't touched between #5916 and this commit (f4dadcfb13). Need to make sure to update in case this isn't merged soon.


janEbert:
The `nan`s that the lint complains about were included in the files previously and not added through these changes.

janEbert:
Applied same solution as https://github.com/NVIDIA/Megatron-LM/pull/5916.

chtruong814 (approved):

PR #6535, opened 2026-08-14
[GTP] Symmetric Memory Registration for the GTP/EGTP ReduceScatter path

### LINKED ISSUE #6540: [feat] GTP+Symmetric Memory Registration buffer

prajwal1210:

# What does this PR do?
Adds symmetric memory registration support for the wgrad tensors in GTP to enable symmetric memory NCCL kernels. 

## Summary

Symmetric memory registration allows for the use of faster LSA/GIN based and CE based collectives in NCCL to improve communication performance for GTP. Specifically, for reduce scatter, NCCL uses an NVLS kernel within an NVLink domain and a RailA2A kernel across IB/Network. These kernels provide better numerical error compared to the legacy NCCL kernels and we want to enable them for the GTP/EGTP group as an alternative to a2a + local fp32 reduction.

For reduce-scatter, only the input needs to be backed by symmetric memory registration but output does not.

Based on these requirements, this MR enables symmetric memory backed tensors gated by two flags:
Flag | Scope
-- | --
--gtp-remat-nccl-ub | dense GTP remat group
--gtp-expert-remat-nccl-ub | routed-expert (EGTP) group
These are independent of independent of `--use-nccl-ub` and off by default.

Support for AG will be added in a separate MR. This MR is one-half of https://github.com/NVIDIA/Megatron-LM/pull/6155 (split for easier review).

## Design

One `ncclMemAlloc`-backed `torch.cuda.MemPool` per GTP process group, created and registered
once by `register_gtp_symm_pool(group)` (`gtp_symmetric_memory.py`). PyTorch's
ProcessGroupNCCL segment hook then auto-registers (`ncclCommWindowRegister` with
`NCCL_WIN_COLL_SYMMETRIC`) every later allocation made under `gtp_symm_pool_ctx(group)`, so
the context manager itself is collective-free and capture-safe. Only the send side of the
reduce-scatter needs a window; the output is the ordinary sharded `main_grad`.

Every buffer the wgrad reduce-scatter touches, where it comes from, and what puts it in the pool:

| Buffer | Where it comes from | What puts it in the pool |
|---|---|---|
| **RS send (ungraphed chains)** | `symmetric_wgrad_pool` (`RegisteredLIFOPool`), padded full shape; producers write into it directly — TE modules via the `grad_buffer()` protocol, Megatron-native linears via an `out=` matmul on the `get_wgrad_tensor()` view | `alloc()` under the pool context |
| **RS send (graphed chains)** | persistent graph wgrad ring slot, allocated pre-capture in `allocate_graph_wgrad_rings` | slot `torch.empty` under `gtp_symm_pool_ctx` |
| **RS recv** | the param's sharded `main_grad` | nothing — send-side-only registration suffices |
| **fp32-accum a2a scratch** | plain wgrad pool | nothing — a registered group takes precedence over `--gtp-remat-reduce-scatter-with-fp32-accumulation` and skips the all-to-all entirely (no scratch drawn, so enabling both costs no extra memory on registered groups) |

Pools are registered at `configure_gtp_remat_from_recipe` time (groups resolved via
`resolve_gtp_remat_group(pg_collection, ...)`) and torn down by `deregister_gtp_symm_pools()`
at the end of `pretrain()`: LIFO buffers dropped first, a synchronize to quiesce outstanding
work, then window deregistration.

## Testing 
All GTP UTs pass 

#### Convergence
Screenshot 2026-08-20 at 10 02 26 PM

## Issue tracking

Linked issue: 

### Pre-checks

### DIFF
--- a/docs/api-guide/core/generalized_tensor_parallel.md
+++ b/docs/api-guide/core/generalized_tensor_parallel.md
@@ -379,8 +379,8 @@ it** — a different collective over a different process group, so enable either
 ### 2.7 NCCL symmetric-memory wgrad reduce-scatter (optional)
 
 ```bash
---gtp-remat-nccl-ub        # dense gtp_remat group          default: off
---gtp-expert-remat-nccl-ub       # routed-expert egtp_remat group  default: off
+--gtp-remat-nccl-ub         # dense gtp_remat group           default: off
+--gtp-expert-remat-nccl-ub  # routed-expert egtp_remat group  default: off
 ```
 
 **Allocates the wgrad reduce-scatter send buffers from an NCCL-window-registered memory pool on the gtp_remat / egtp_remat group, so NCCL runs the reduce-scatter as a single symmetric device kernel — NVLS multimem within an NVLink domain, rail kernels when the group spans domains — instead of a ring.** Only the send side needs registration; the sharded output lands in the ordinary `main_grad`.
@@ -399,10 +399,10 @@ it** — a different collective over a different process group, so enable either
   (when the wgrad dtype matches `main_grad`). The untied embedding's wgrad is materialized by
   `F.embedding`'s own backward and pays one copy into the buffer.
 - **FP32-accumulation interplay.** A registered pool takes precedence over §2.6 on its group:
-  NCCL symmetric reduce-scatters provide equivalent numerics with better performance, so the
-  group keeps the symmetric reduce-scatter and the fp32-accum all-to-all applies only to axes
-  without a pool. E.g. `--gtp-remat-nccl-ub` + §2.6 gives a symmetric dense-GTP reduce-scatter and
-  the fp32-accum all-to-all on the EGTP axis.
+  NVLS symmetric reduce-scatters accumulate in fp32 in-switch (NCCL's `multimem.ld_reduce` uses
+  `.acc::f32` for bf16), so the group keeps the symmetric reduce-scatter and the fp32-accum
+  all-to-all applies only to axes without a pool. E.g. `--gtp-remat-nccl-ub` + §2.6 gives a
+  symmetric dense-GTP reduce-scatter and the fp32-accum all-to-all on the EGTP axis.
 - **Independent of `--use-nccl-ub`.** That flag registers DP-group (DDP bucket) buffers; these
   flags cover the gtp_remat axes only.
 
--- a/docs/api-guide/core/generalized_tensor_parallel.md
+++ b/docs/api-guide/core/generalized_tensor_parallel.md
@@ -48,6 +48,7 @@ Both `GTP_remat` collectives are prefetched one step ahead, so they overlap the 
     - [2.4 Minimal end-to-end example](#24-minimal-end-to-end-example)
     - [2.5 Tuning knobs](#25-tuning-knobs)
     - [2.6 FP32-accumulation wgrad reduce-scatter (optional)](#26-fp32-accumulation-wgrad-reduce-scatter-optional)
+    - [2.7 NCCL symmetric-memory wgrad reduce-scatter (optional)](#27-nccl-symmetric-memory-wgrad-reduce-scatter-optional)
   - [3. Implementation details](#3-implementation-details)
     - [3.1 GTP\_remat architecture (Mcore ↔ TE integration)](#31-gtp_remat-architecture-mcore--te-integration)
       - [What the flags do under the hood](#what-the-flags-do-under-the-hood)
@@ -362,7 +363,8 @@ it** — a different collective over a different process group, so enable either
   applies either way.
 - **Auto-bypass at axis size ≤ 2.** The gate reads the per-chain group, so each axis decides
   independently: a `GTP_remat=8 × EGTP_remat=2` run gets FP32 accumulation on the dense weights
-  and the plain reduce-scatter on the experts.
+  and the plain reduce-scatter on the experts. A group with a registered symmetric pool (§2.7)
+  also bypasses — the pool takes precedence.
 - **Scratch lifetime.** The buffer comes from GTP's wgrad pool rather than a fresh `empty_like`,
   and is returned only once the handle is waited — it is the *input* to the deferred FP32 sum.
 - **Batched (grouped / routed-expert) path.** The all-to-alls share one `ncclGroupStart/End` via
@@ -371,6 +373,36 @@ it** — a different collective over a different process group, so enable either
   behind it in one composite handle — which is why the all-to-alls are issued with
   `async_op=True`: for this primitive that flag defers the sum, it does not merely return a
   handle. (DDP's own flag sidesteps all this by asserting a single bucket.)
+
+### 2.7 NCCL symmetric-memory wgrad reduce-scatter (optional)
+
+```bash
+--gtp-nccl-ub        # dense gtp_remat group          default: off
+--egtp-nccl-ub       # routed-expert egtp_remat group  default: off
+```
+
+**Allocates the wgrad reduce-scatter send buffers from an NCCL-window-registered memory pool on the gtp_remat / egtp_remat group, so NCCL runs the reduce-scatter as a single symmetric device kernel — NVLS multimem within an NVLink domain, rail kernels when the group spans domains — instead of a ring.** Only the send side needs registration; the sharded output lands in the ordinary `main_grad`.
+
+| | |
+|---|---|
+| **Use when** | NVLS-capable systems (NVSwitch, NCCL with symmetric-kernel support) where the gtp_remat wgrad reduce-scatter is exposed |
+| **Skip when** | `--disable-symmetric-registration` is set (asserted incompatible) |
+| **Gain** | in-switch reduction: fewer SMs and lower latency per reduce-scatter |
+| **Cost** | a persistent registered pool of unsharded-wgrad-sized buffers per group, plus a one-time registration warmup; deregistered at shutdown |
+
+**Behaviour notes**
+
+- **Zero-copy producers.** Wgrads are written straight into the registered buffer — TE modules
+  via the `DistributedWeight.grad_buffer` protocol, Megatron-native linears via an `out=` matmul
+  (when the wgrad dtype matches `main_grad`). The untied embedding's wgrad is materialized by
+  `F.embedding`'s own backward and pays one copy into the buffer.
+- **FP32-accumulation interplay.** A registered pool takes precedence over §2.6 on its group:
+  NCCL symmetric reduce-scatters provide equivalent numerics with better performance, so the
+  group keeps the symmetric reduce-scatter and the fp32-accum all-to-all applies only to axes
+  without a pool. E.g. `--gtp-nccl-ub` + §2.6 gives a symmetric dense-GTP reduce-scatter and
+  the fp32-accum all-to-all on the EGTP axis.
+- **Independent of `--use-nccl-ub`.** That flag registers DP-group (DDP bucket) buffers; these
+  flags cover the gtp_remat axes only. Enable either, both, or neither.
 
 ---
 
--- a/docs/api-guide/core/generalized_tensor_parallel.md
+++ b/docs/api-guide/core/generalized_tensor_parallel.md
@@ -231,7 +231,7 @@ The table below covers every GTP-related CLI flag and Python knob. "Required" me
 | `--expert-tensor-parallel-num-weight-shards` | **Required** | MoE models (to shard routed-expert weights) | — | Total ETP×EGTP_remat shards per expert weight; EGTP_remat degree = value ÷ ETP. Independent of dense axis. [§2.2](#22-required-flags) |
 | `--gtp-remat-reduce-scatter-with-fp32-accumulation` | **Optional** | BF16 wgrads **and** GTP_remat axis ≥ 4 | off | Replaces the ring RS with an all-to-all + local FP32 sum to eliminate per-hop rounding error. Auto-bypassed at axis size ≤ 2. [§2.6](#26-fp32-accumulation-wgrad-reduce-scatter-optional) |
 | `--gtp-remat-nccl-ub` | **Optional** | For enabling symmetric-memory NCCL kernels on supported systems | off | Enables symmetric memory registration for the dense gtp_remat wgrad reduce-scatter path. Takes precedence over fp32-accum on its group; incompatible with `--disable-symmetric-registration`. [§2.7](#27-nccl-symmetric-memory-wgrad-reduce-scatter-optional) |
-| `--egtp-remat-nccl-ub` | **Optional** | For enabling symmetric-memory NCCL kernels on supported systems | off | Enables symmetric memory registration for the routed-expert egtp_remat wgrad reduce-scatter path. [§2.7](#27-nccl-symmetric-memory-wgrad-reduce-scatter-optional) |
+| `--gtp-expert-remat-nccl-ub` | **Optional** | For enabling symmetric-memory NCCL kernels on supported systems | off | Enables symmetric memory registration for the routed-expert egtp_remat wgrad reduce-scatter path. [§2.7](#27-nccl-symmetric-memory-wgrad-reduce-scatter-optional) |
 | `--gtp-remat-opt-in-modules` | **Optional** | MoE models with large `--moe-latent-size` | `[]` | Space-separated list of module tokens to opt in to GTP_remat sharding. Currently supported: `moe_latent_proj` (shards `fc1_latent_proj` / `fc2_latent_proj`; only beneficial when the latent size is large enough to amortize the all-gather). [§1.5](#15-opt-in-minimally-invasive-integration) |
 | `--fp8-param-gather` | **Required** | GTP + `--fp8-recipe mxfp8` | off | Gathers native MXFP8 shard directly; without it the grad-buffer reuse path is unavailable and `arguments.py` asserts. Always paired with `--reuse-grad-buf-for-mxfp8-param-ag`. [§1.3](#13-low-precision-gather-native-fp8--nvfp4-param) |
 | `--reuse-grad-buf-for-mxfp8-param-ag` | **Required** | GTP + `--fp8-recipe mxfp8` | off | Reuses the grad buffer for the MXFP8 all-gather (MXFP8 cannot map into the contiguous param buffer). Must accompany `--fp8-param-gather`. [§1.3](#13-low-precision-gather-native-fp8--nvfp4-param) |
@@ -380,7 +380,7 @@ it** — a different collective over a different process group, so enable either
 
 ```bash
 --gtp-remat-nccl-ub        # dense gtp_remat group          default: off
---egtp-remat-nccl-ub       # routed-expert egtp_remat group  default: off
+--gtp-expert-remat-nccl-ub       # routed-expert egtp_remat group  default: off
 ```
 
 **Allocates the wgrad reduce-scatter send buffers from an NCCL-window-registered memory pool on the gtp_remat / egtp_remat group, so NCCL runs the reduce-scatter as a single symmetric device kernel — NVLS multimem within an NVLink domain, rail kernels when the group spans domains — instead of a ring.** Only the send side needs registration; the sharded output lands in the ordinary `main_grad`.
--- a/docs/api-guide/core/generalized_tensor_parallel.md
+++ b/docs/api-guide/core/generalized_tensor_parallel.md
@@ -230,6 +230,8 @@ The table below covers every GTP-related CLI flag and Python knob. "Required" me
 | `--tensor-parallel-num-weight-shards` | **Required** | Always, to activate dense GTP | — | Total TP×GTP_remat shards per dense weight; GTP_remat degree = value ÷ TP. Must be ≥ TP and divisible by it. [§2.2](#22-required-flags) |
 | `--expert-tensor-parallel-num-weight-shards` | **Required** | MoE models (to shard routed-expert weights) | — | Total ETP×EGTP_remat shards per expert weight; EGTP_remat degree = value ÷ ETP. Independent of dense axis. [§2.2](#22-required-flags) |
 | `--gtp-remat-reduce-scatter-with-fp32-accumulation` | **Optional** | BF16 wgrads **and** GTP_remat axis ≥ 4 | off | Replaces the ring RS with an all-to-all + local FP32 sum to eliminate per-hop rounding error. Auto-bypassed at axis size ≤ 2. [§2.6](#26-fp32-accumulation-wgrad-reduce-scatter-optional) |
+| `--gtp-remat-nccl-ub` | **Optional** | For enabling symmetric-memory NCCL kernels on supported systems | off | Enables symmetric memory registration for the dense gtp_remat wgrad reduce-scatter path. Takes precedence over fp32-accum on its group; incompatible with `--disable-symmetric-registration`. [§2.7](#27-nccl-symmetric-memory-wgrad-reduce-scatter-optional) |
+| `--egtp-remat-nccl-ub` | **Optional** | For enabling symmetric-memory NCCL kernels on supported systems | off | Enables symmetric memory registration for the routed-expert egtp_remat wgrad reduce-scatter path. [§2.7](#27-nccl-symmetric-memory-wgrad-reduce-scatter-optional) |
 | `--gtp-remat-opt-in-modules` | **Optional** | MoE models with large `--moe-latent-size` | `[]` | Space-separated list of module tokens to opt in to GTP_remat sharding. Currently supported: `moe_latent_proj` (shards `fc1_latent_proj` / `fc2_latent_proj`; only beneficial when the latent size is large enough to amortize the all-gather). [§1.5](#15-opt-in-minimally-invasive-integration) |
 | `--fp8-param-gather` | **Required** | GTP + `--fp8-recipe mxfp8` | off | Gathers native MXFP8 shard directly; without it the grad-buffer reuse path is unavailable and `arguments.py` asserts. Always paired with `--reuse-grad-buf-for-mxfp8-param-ag`. [§1.3](#13-low-precision-gather-native-fp8--nvfp4-param) |
 | `--reuse-grad-buf-for-mxfp8-param-ag` | **Required** | GTP + `--fp8-recipe mxfp8` | off | Reuses the grad buffer for the MXFP8 all-gather (MXFP8 cannot map into the contiguous param buffer). Must accompany `--fp8-param-gather`. [§1.3](#13-low-precision-gather-native-fp8--nvfp4-param) |
@@ -377,15 +379,15 @@ it** — a different collective over a different process group, so enable either
 ### 2.7 NCCL symmetric-memory wgrad reduce-scatter (optional)
 
 ```bash
---gtp-nccl-ub        # dense gtp_remat group          default: off
---egtp-nccl-ub       # routed-expert egtp_remat group  default: off
+--gtp-remat-nccl-ub        # dense gtp_remat group          default: off
+--egtp-remat-nccl-ub       # routed-expert egtp_remat group  default: off
 ```
 
 **Allocates the wgrad reduce-scatter send buffers from an NCCL-window-registered memory pool on the gtp_remat / egtp_remat group, so NCCL runs the reduce-scatter as a single symmetric device kernel — NVLS multimem within an NVLink domain, rail kernels when the group spans domains — instead of a ring.** Only the send side needs registration; the sharded output lands in the ordinary `main_grad`.
 
 | | |
 |---|---|
-| **Use when** | NVLS-capable systems (NVSwitch, NCCL with symmetric-kernel support) where the gtp_remat wgrad reduce-scatter is exposed |
+| **Use when** | Systems with NCCL symmetric memory support |
 | **Skip when** | `--disable-symmetric-registration` is set (asserted incompatible) |
 | **Gain** | in-switch reduction: fewer SMs and lower latency per reduce-scatter |
 | **Cost** | a persistent registered pool of unsharded-wgrad-sized buffers per group, plus a one-time registration warmup; deregistered at shutdown |
@@ -399,10 +401,10 @@ it** — a different collective over a different process group, so enable either
 - **FP32-accumulation interplay.** A registered pool takes precedence over §2.6 on its group:
   NCCL symmetric reduce-scatters provide equivalent numerics with better performance, so the
   group keeps the symmetric reduce-scatter and the fp32-accum all-to-all applies only to axes
-  without a pool. E.g. `--gtp-nccl-ub` + §2.6 gives a symmetric dense-GTP reduce-scatter and
+  without a pool. E.g. `--gtp-remat-nccl-ub` + §2.6 gives a symmetric dense-GTP reduce-scatter and
   the fp32-accum all-to-all on the EGTP axis.
 - **Independent of `--use-nccl-ub`.** That flag registers DP-group (DDP bucket) buffers; these
-  flags cover the gtp_remat axes only. Enable either, both, or neither.
+  flags cover the gtp_remat axes only.
 
 ---
 
--- a/megatron/core/extensions/transformer_engine.py
+++ b/megatron/core/extensions/transformer_engine.py
@@ -454,7 +454,7 @@ def _gtp_pre_init(
     return shard_out, gtp_ctx
 
 
-def _gtp_attach_post_init(module, gtp_ctx, is_grouped=False, is_expert=False):
+def _gtp_attach_post_init(module, gtp_ctx, is_grouped=False):
     """Attach the GTP surface to a pre-sharded TE module's weights and restore logical out_features.
 
     ``is_grouped=True`` for GroupedLinear (per-expert weight0..N, coalesced AG via weight_list).
@@ -466,9 +466,7 @@ def _gtp_attach_post_init(module, gtp_ctx, is_grouped=False, is_expert=False):
     # super().__init__): downstream code reads it, e.g. the grouped-MLP fusion gate checks
     # fc1.out_features == 2 * fc2.in_features (a shard-sized fc1 would silently disable fusion).
     module.out_features = logical_out_features
-    attach_gtp_to_presharded_module(
-        module, gtp_remat_group, pad_length, is_grouped=is_grouped, is_expert=is_expert
-    )
+    attach_gtp_to_presharded_module(module, gtp_remat_group, pad_length, is_grouped=is_grouped)
 
 
 @contextmanager
@@ -501,7 +499,7 @@ def _init_gtp_remat_context(
         out_split_size=out_split_size,
     )
     yield out_features
-    _gtp_attach_post_init(module, gtp_ctx, is_grouped=is_grouped, is_expert=is_expert)
+    _gtp_attach_post_init(module, gtp_ctx, is_grouped=is_grouped)
 
 
 def split_te_layernorm_column_parallel_linear(
--- a/megatron/core/extensions/transformer_engine.py
+++ b/megatron/core/extensions/transformer_engine.py
@@ -1502,7 +1502,6 @@ class TELayerNormColumnParallelLinear(te.pytorch.LayerNormLinear):
             output_size,
             gtp_remat_group,
             extra_kwargs,
-            is_expert=is_expert,
             rng_via_kwarg=False,
             out_split_size=self.tp_size,
         )
--- a/megatron/core/extensions/transformer_engine.py
+++ b/megatron/core/extensions/transformer_engine.py
@@ -454,7 +454,7 @@ def _gtp_pre_init(
     return shard_out, gtp_ctx
 
 
-def _gtp_attach_post_init(module, gtp_ctx, is_grouped=False):
+def _gtp_attach_post_init(module, gtp_ctx, is_grouped=False, is_expert=False):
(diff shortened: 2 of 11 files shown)


megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,198 @@
...
+        return None
+    pool = get_gtp_symm_pool(group)
+    if group.group_name in _registered:
+        return pool
+    # NCCL creates communicators lazily on the first collective; run one tiny all-reduce
+    # so the registration below sees an initialized communicator.
+    warmup = torch.zeros(1, device=torch.cuda.current_device())
+    dist.all_reduce(warmup, group=group)
+    nccl_allocator.register_mem_pool(pool, group, symmetric=True)
+    _registered[group.group_name] = group
```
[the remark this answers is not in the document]
prajwal1210: This is a redundant fix. The all-reduce itself is a barrier synchronizing ranks on the group.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1720,13 +1807,22 @@ def batched_all_gather_and_prefetch(self, **kwargs):
...
+        if _use_symmetric_wgrad_buffer(self):
+            # Lifecycle invariant: get_wgrad_tensor -> GEMM -> _prepare consumes the slot -> RS.
+            if self._wgrad_symm_slot is not None:
+                raise RuntimeError(
+                    "[GTP] get_wgrad_tensor called again before the previous symmetric "
+                    f"wgrad buffer was consumed ({self._debug_name})."
+                )
+            buf = _alloc_symmetric_wgrad_buffer(self, self.main_grad.dtype, self.device)
+            self._wgrad_symm_slot = buf
+            return buf[: self._unsharded_shape[0]]
```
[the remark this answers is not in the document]
prajwal1210: This will only happen in the case there is a error between the get_wgrad_tensor and the RS being run. It is better to deliberately error here than to let something weird happen imo.

docs/api-guide/core/generalized_tensor_parallel.md
```diff
@@ -372,6 +371,36 @@ it** — a different collective over a different process group, so enable either
   `async_op=True`: for this primitive that flag defers the sum, it does not merely return a
   handle. (DDP's own flag sidesteps all this by asserting a single bucket.)
 
+### 2.7 NCCL symmetric-memory wgrad reduce-scatter (optional)
+
+```bash
+--gtp-nccl-ub        # dense gtp_remat group          default: off
+--egtp-nccl-ub       # routed-expert egtp_remat group  default: off
```
fanshiqing: nit: 
- `gtp-nccl-ub` -> `gtp-remat-ub`, 
- `egtp-nccl-ub` -> `egtp-remat-ub`

docs/api-guide/core/generalized_tensor_parallel.md
```diff
@@ -372,6 +371,36 @@ it** — a different collective over a different process group, so enable either
...
+| **Gain** | in-switch reduction: fewer SMs and lower latency per reduce-scatter |
+| **Cost** | a persistent registered pool of unsharded-wgrad-sized buffers per group, plus a one-time registration warmup; deregistered at shutdown |
+
+**Behaviour notes**
+
+- **Zero-copy producers.** Wgrads are written straight into the registered buffer — TE modules
+  via the `DistributedWeight.grad_buffer` protocol, Megatron-native linears via an `out=` matmul
+  (when the wgrad dtype matches `main_grad`). The untied embedding's wgrad is materialized by
+  `F.embedding`'s own backward and pays one copy into the buffer.
+- **FP32-accumulation interplay.** A registered pool takes precedence over §2.6 on its group:
```
fanshiqing, line 401: it's an attractive feature becuause now we are suffuring from the perf regression from all2all kernnels @deepakn94 @kvareddy

docs/api-guide/core/generalized_tensor_parallel.md
fanshiqing: pls also add these two args into the table in the section 2.1 [knob summary ](https://github.com/NVIDIA/Megatron-LM/blob/main/docs/api-guide/core/generalized_tensor_parallel.md#21-knob-summary)

docs/api-guide/core/generalized_tensor_parallel.md
```diff
@@ -372,6 +371,36 @@ it** — a different collective over a different process group, so enable either
...
+--gtp-nccl-ub        # dense gtp_remat group          default: off
+--egtp-nccl-ub       # routed-expert egtp_remat group  default: off
+```
+
+**Allocates the wgrad reduce-scatter send buffers from an NCCL-window-registered memory pool on the gtp_remat / egtp_remat group, so NCCL runs the reduce-scatter as a single symmetric device kernel — NVLS multimem within an NVLink domain, rail kernels when the group spans domains — instead of a ring.** Only the send side needs registration; the sharded output lands in the ordinary `main_grad`.
+
+| | |
+|---|---|
+| **Use when** | NVLS-capable systems (NVSwitch, NCCL with symmetric-kernel support) where the gtp_remat wgrad reduce-scatter is exposed |
+| **Skip when** | `--disable-symmetric-registration` is set (asserted incompatible) |
```
fanshiqing, line 391: Previous this knob  `--disable-symmetric-registration` is working with `--use-nccl-ub`.
Now we expect it to work with `--gtp-remat-nccl-ub` too, so you also need to update the [doc string](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/arguments.py#L3041) of `--disable-symmetric-registration` to make it clear

docs/api-guide/core/generalized_tensor_parallel.md
```diff
@@ -372,6 +371,36 @@ it** — a different collective over a different process group, so enable either
...
+  via the `DistributedWeight.grad_buffer` protocol, Megatron-native linears via an `out=` matmul
+  (when the wgrad dtype matches `main_grad`). The untied embedding's wgrad is materialized by
+  `F.embedding`'s own backward and pays one copy into the buffer.
+- **FP32-accumulation interplay.** A registered pool takes precedence over §2.6 on its group:
+  NCCL symmetric reduce-scatters provide equivalent numerics with better performance, so the
+  group keeps the symmetric reduce-scatter and the fp32-accum all-to-all applies only to axes
+  without a pool. E.g. `--gtp-nccl-ub` + §2.6 gives a symmetric dense-GTP reduce-scatter and
+  the fp32-accum all-to-all on the EGTP axis.
+- **Independent of `--use-nccl-ub`.** That flag registers DP-group (DDP bucket) buffers; these
+  flags cover the gtp_remat axes only. Enable either, both, or neither.
```
fanshiqing: nit: just remove redandant `Enable either, both, or neither.`

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -575,7 +642,9 @@ def _gtp_slice_one_param(param, gtp_remat_group, *, name="<unnamed>"):
     return gtp_shard
 
 
-def _gtp_attach_attrs(gtp_shard, gtp_remat_group, *, is_grouped=False, expert_idx=0):
+def _gtp_attach_attrs(
+    gtp_shard, gtp_remat_group, *, is_grouped=False, expert_idx=0, is_expert=False
```
fanshiqing: what's the necessarity of passing a new `is_expert` here?

In the vanilla API design, `is_grouped` is set to True **if and only if** `is_experts=True`, thus we don't need pass one-extra args.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -488,6 +533,28 @@ def configure_gtp_remat_from_recipe(
...
         update_gtp_config(pad_for_alignment=16)
 
+    if reduce_scatter_with_fp32_accumulation and (gtp_nccl_ub or egtp_nccl_ub):
+        log_single_rank(
+            logger,
+            logging.WARNING,
+            "--gtp-nccl-ub/--egtp-nccl-ub take precedence over "
+            "--gtp-remat-reduce-scatter-with-fp32-accumulation on their groups: NCCL "
+            "symmetric reduce-scatters provide equivalent numerics with better performance.",
+        )
```
fanshiqing: What about replace weak `warning` with `assert` instead?
We don't expect any user to enable `gtp-remat-reduce-scatter-with-fp32-accumulation` and `gtp-remat-nccl-ub` at the same time. And we should make sure they are explicitly be aware of it in case of abuse

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -488,6 +533,28 @@ def configure_gtp_remat_from_recipe(
...
+    # resolve_gtp_remat_group prefers the caller's pg_collection (custom groups included) and
+    # falls back to the MPU globals; register_gtp_symm_pool no-ops on None/single-rank groups.
+    # Local import: process_groups_config reaches parallel_state, which imports this module.
+    if gtp_nccl_ub or egtp_nccl_ub:
+        from megatron.core.process_groups_config import resolve_gtp_remat_group
+
+        if gtp_nccl_ub:
+            register_gtp_symm_pool(resolve_gtp_remat_group(pg_collection, is_expert=False))
+        if egtp_nccl_ub:
+            register_gtp_symm_pool(resolve_gtp_remat_group(pg_collection, is_expert=True))
```
fanshiqing: move L536-556 outside of `generalized_tensor_parallelism.py`, it's super awkard to call `resolve_gtp_remat_group` inside the core implementation of gtp
- L536-L543 into arguments.py (gtp related checks in one [place](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/arguments.py#L1464-L1536))
- L550-L556: what about some places in `training.py`? before the training-loop starts.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -594,6 +663,9 @@ def _gtp_attach_attrs(gtp_shard, gtp_remat_group, *, is_grouped=False, expert_id
         gtp_shard.chain_id = GTPChain.UNGRAPHED.value
     gtp_shard.group = gtp_remat_group
     gtp_shard.gtp_remat_size = gtp_remat_group.size()
+    # gtp_smr: draw this param's reduce-scatter send buffers from the per-group
+    # window-registered pool. Expert vs dense comes from the caller's is_expert.
+    gtp_shard.gtp_smr = GTP_CONFIG.egtp_nccl_ub if is_expert else GTP_CONFIG.gtp_nccl_ub
```
fanshiqing: What's the necessity of keeping `gtp_smr` been stored per GTP sharded params?
- gtp_shard already has attributes to identify if it's experts or not (by attr `is_routed_expert`)
- then read from the global sharded  `GTP_CONFIG.egtp_nccl_ub` or `GTP_CONFIG.gtp_nccl_ub` is just fine

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
fanshiqing, line 606: GTP use `is_grouped=True/False` to identify if a sharded param is expert or not.
for `is_group=True`, it will further set `is_routed_expert` and `expert_idx` to identify the expert-param-oder in the grouped-expert-params list.

Thus we don't need to passin the new added `is_expert` arg here.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1734,13 +1808,22 @@ def batched_all_gather_and_prefetch(self, **kwargs):
...
-        """Return a logical-shape view of stable ring storage or ordinary scratch.
-
-        Capture ownership is registered later, when the final RS input is selected.
-        """
+        """Return a logical-shape view of stable ring storage or ordinary scratch."""
         ring_slot = getattr(self, "_gtp_graph_wgrad_ring_slot", None)
         if ring_slot is not None:
             return self._gtp_graph_wgrad_ring_view
+
+        # TODO: Merge the ring wgrad slot and symmetric wgrad slot into a single slot.
```
fanshiqing, line 1768: Please help merge these slot in this MR if that's not heavy

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -307,6 +308,38 @@ class GTPWeightState(Enum):
...
+    """True when this weight's wgrad reduction runs as the fp32-accum all-to-all.
+
+    Size <= 2 gains nothing (one addition) and still costs the scratch, so it bypasses.
+    The exact complement of _use_symmetric_wgrad_buffer: a symmetric send buffer takes
+    precedence, so a weight never routes to both.
+    """
+    return (
+        GTP_CONFIG.reduce_scatter_with_fp32_accumulation
+        and weight.group.size() > 2
+        and not _use_symmetric_wgrad_buffer(weight)
```
fanshiqing: We should assert at the very begging (in arguments.py) to make sure `reduce_scatter_with_fp32_accumulation=True` and `gpt-remat-use-symmetric-buffer=True` should never be allowed.

So that we can keep the code simple and  get rid of this checks

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -307,6 +308,38 @@ class GTPWeightState(Enum):
 _GTP_GROUPED_BUF_PARITY_COUNTER: Dict[tuple, int] = {}
 
 
+def _use_symmetric_wgrad_buffer(weight) -> bool:
+    """True when this weight's reduce-scatter should send from the window-registered pool.
+
+    A registered pool takes precedence over fp32-accum on its group: the group keeps the
+    plain (symmetric) reduce-scatter, and fp32-accum applies only to unregistered groups.
+    """
+    return weight.gtp_smr and is_gtp_symm_pool_registered(weight.group)
```
fanshiqing: is there any special case that `GTP_CONFIG.egtp_nccl_ub` cannot satisfy the usecase here?
if not, pls don't set an extra 'gtp_smr` attribute for each gpt-sharded-params which is not that necessary

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -468,10 +507,14 @@ def configure_gtp_remat_from_recipe(
     fp8=False,
     calculate_per_token_loss=False,
     reduce_scatter_with_fp32_accumulation=False,
+    gtp_nccl_ub=False,
+    egtp_nccl_ub=False,
+    pg_collection=None,
```
fanshiqing: let's try not to passing `pg_collection` stuffs into gtp core file

megatron/core/tensor_parallel/gtp_api.py
```diff
@@ -9,6 +9,10 @@
 module uses GTP symbols without TE.
 """
 
+# Outside the HAVE_TE guard on purpose: gtp_symmetric_memory has no TE dependency, and
+# shutdown must be able to deregister pools even where the TE-backed surface is unavailable.
+from megatron.core.tensor_parallel.gtp_symmetric_memory import deregister_and_clear_gtp_symm_pools
```
fanshiqing: this line of import should also be covered by the follwoing `try...except...` 

because this import is only meaningful when `HAVE_GTP=True`

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
prajwal1210: So currently, the plan was to use a2a + fp32 local reduction for the egtp group and the symmetric kernel for the gtp group. The underlying symmetric kernel for RS over IB in NCCL is not tuned and sees really bad perf. That needs to be fixed first.

docs/api-guide/core/generalized_tensor_parallel.md
prajwal1210: Currently, GTP UB always does symmetric registration. I don't think there is any benefit of using ub without registration so I would argue against splitting the behavior.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
prajwal1210: This will be pretty substantial. The two slots have conflicting requirements. LIFO exists to reduce memory and re-use symmetric slots whereas ring slots are for partial CG correctness at the expense for more memory I believe. A fused allocator is ideal but needs to careful designed.

megatron/core/extensions/transformer_engine.py
```diff
@@ -1477,6 +1477,7 @@ def __init__(
             output_size,
             gtp_remat_group,
             extra_kwargs,
+            is_expert=is_expert,
```
JF-D: It seems this `is_expert` isn't used anywhere else?

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1939,9 +1978,8 @@ def _reduce_scatter(self, wgrads, async_op, nvtx_label=None):
             rs_ctx = nullcontext()
 
         with rs_ctx:
-            # Size <= 2 gains nothing (one addition) and still costs the scratch, so it bypasses.
             # self.group is per chain, so each axis decides on its own.
-            if GTP_CONFIG.reduce_scatter_with_fp32_accumulation and self.group.size() > 2:
+            if _use_fp32_accum_rs(self):
```
JF-D: We only add one condition here, but why do we want to make it a separate function?

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -2072,6 +2134,12 @@ def wgrad_reduce_scatter(self, wgrad, nvtx_label=None):
             if poolable:
                 for buf in wgrads:
                     _wgrad_pool_put(buf)
+            for buf in rs_inputs:
+                # Dual release, mirroring _release_comm_scratch: each free is tag-gated
+                # and takes only its own pool's buffers.
+                if poolable:
+                    _wgrad_pool_put(buf)
```
JF-D: Previously we didn't put the input wgrad buffer (rs_inputs) back. We should do this for both input and output wgrads, even if there is no symmetric wgrad buffer, right?

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
fanshiqing: Ah...okay, then just keep it as is.

docs/api-guide/core/generalized_tensor_parallel.md
fanshiqing: That's fine. I found you already added assert for that case.

docs/api-guide/core/generalized_tensor_parallel.md
deepakn94: Yes, but we are still verifying that this approach has acceptable numerics.

megatron/training/arguments.py
```diff
@@ -3038,7 +3052,13 @@ def _add_distributed_args(parser):
                        'which is improving the performance of the overlapped computation.')
     group.add_argument('--disable-symmetric-registration', action='store_true', dest='disable_symmetric_registration',
                        default=False, help='Disable symmetric (window) registration for NCCL userbuffer registration.'
-                       'This option will force to use conventional (local) userbuffer registration when use-nccl-ub is set.')
+                       'This option will force to use conventional (local) userbuffer registration when use-nccl-ub is set. '
+                       'Cannot be combined with --gtp-remat-nccl-ub/--egtp-remat-nccl-ub, which require symmetric windows.')
+    group.add_argument('--gtp-remat-nccl-ub', action='store_true', dest='gtp_remat_nccl_ub',
+                       default=False, help='Register the wgrad reduce-scatter send buffers with NCCL symmetric '
+                       'memory on the GTP group, independent of --use-nccl-ub (which covers the DP group).')
+    group.add_argument('--egtp-remat-nccl-ub', action='store_true', dest='egtp_remat_nccl_ub',
```
deepakn94: Nit: Can we leave the prefix of the argument as "--gtp" (but specify elsewhere in the argument that this is for expert buffers).

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,207 @@
...
+from megatron.core.utils import is_torch_min_version, log_single_rank
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Registries (process-global, keyed by group.group_name)
+# ---------------------------------------------------------------------------
+
+# One MemPool per GTP/EGTP group, created once.
+_pools: "dict[str, torch.cuda.MemPool]" = {}
```
deepakn94: Could we also make this `_pools: typing.Dict[str, torch.cuda.MemPool] = {}`?

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,207 @@
...
+
+# Groups whose pool registration is live. Maps name -> group (not a set) because
+# teardown needs the group object to deregister.
+_registered: "dict[str, object]" = {}
```
deepakn94: `_registered: typing.Dict[str, Any] = {}`

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,207 @@
...
+
+# ---------------------------------------------------------------------------
+# Pool lifecycle: create -> warm -> register -> allocate-into -> query -> deregister
+# ---------------------------------------------------------------------------
+
+
+def get_gtp_symm_pool(group: dist.ProcessGroup) -> torch.cuda.MemPool:
```
deepakn94: If this is an internal method, precede with underscore?

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,207 @@
...
+    """Return the per-group ``ncclMemAlloc``-backed MemPool, creating it once."""
+    name = group.group_name
+    pool = _pools.get(name)
+    if pool is None:
+        nccl_allocator.init()
+        pool = nccl_allocator.create_nccl_mem_pool(symmetric=True)
```
deepakn94, line 58: Creating the pool is not sufficient to register it?

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,207 @@
...
+    ``None`` and single-rank groups, which are never registered."""
+    return group is not None and group.size() > 1 and group.group_name in _registered
+
+
+# ---------------------------------------------------------------------------
+# RS send-buffer LIFO: recycled window-registered scratch for wgrad reduce-scatters
+# ---------------------------------------------------------------------------
+
+
+class RegisteredLifoPool:
```
deepakn94: Nit: capitalize LIFO?

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,207 @@
...
+    """A recycling cache of window-registered buffers, one free list per group.
```
deepakn94, line 114: Also, why LIFO?

prajwal1210: I think even FIFO works here. The LIFO/FIFO choice is per key (dtype, size, group) anyway. I preferred LIFO because despite the number of buffers per key (if every we over-allocated), LIFO leads to the same buffer being re-used for the same operation whereas FIFO would change the order of buffer use.

megatron/core/tensor_parallel/gtp_symmetric_memory.py
prajwal1210: No, creating a pool just selects the memory allocator backend. Registration is a collective operation.

docs/api-guide/core/generalized_tensor_parallel.md
```diff
@@ -372,6 +376,36 @@ it** — a different collective over a different process group, so enable either
...
+--gtp-expert-remat-nccl-ub       # routed-expert egtp_remat group  default: off
+```
+
+**Allocates the wgrad reduce-scatter send buffers from an NCCL-window-registered memory pool on the gtp_remat / egtp_remat group, so NCCL runs the reduce-scatter as a single symmetric device kernel — NVLS multimem within an NVLink domain, rail kernels when the group spans domains — instead of a ring.** Only the send side needs registration; the sharded output lands in the ordinary `main_grad`.
+
+| | |
+|---|---|
+| **Use when** | Systems with NCCL symmetric memory support |
+| **Skip when** | `--disable-symmetric-registration` is set (asserted incompatible) |
+| **Gain** | in-switch reduction: fewer SMs and lower latency per reduce-scatter |
```
fanshiqing, line 392: Can you provide detailed expected SM cost saving at least for Blackwell + NVLink domain? 
Something like [these](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/distributed/distributed_data_parallel_config.py#L121-L135):

```
| Communication domain | use_sharp | SM usage of "AG/RS" |
      |----------------------|-----------|---------------------|
      | NVL                  | N/A       | 4 / 5               |
      | NVL+IB               | False     | 16 / 16             |
      | NVL+IB               | True      | 6 / 6               |
      | IB                   | False     | 1 / 4               |
      | IB                   | True      | 1 / 1               |
```
we have discussed it before  and I remembered you mentioned the true SM cost is not as low as what's the above table showed.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -432,12 +445,18 @@ class GTPRematConfig:
...
+    gtp_expert_remat_nccl_ub: bool = False
     # Persistent wgrad slots per scheduling/shape domain for partial-CG asynchronous reduce-scatter.
     # Two slots cover the usual case of one same-key writer per graph. A graph containing multiple
     # same-key writers may need more slots to keep all in-flight RS inputs distinct.
     # TODO: Infer each domain's ring size automatically.
     graph_wgrad_ring_size: int = 2
 
+    # Symmetric-memory registration: back the wgrad reduce-scatter send buffers with
+    # ncclMemAlloc pools registered on the GTP / EGTP group, so NCCL can use its
+    # symmetric (NVLS) kernels. Independent of --use-nccl-ub, which covers the DP group.
```
fanshiqing: These 3 line comments should also been moved to some other places for now.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -2007,9 +2049,29 @@ def _prepare_wgrad_reduce_scatter_inputs(self, wgrads):
...
+
+                    # Alias check and copy
+                    if wgrad.data_ptr() != symm_slot.data_ptr():
+                        symm_slot[: weight._unsharded_shape[0]].copy_(wgrad)
+                        _wgrad_pool_put(wgrad)
+
+                    # Required to free the slot after the RS is waited on.
+                    wgrads[index] = symm_slot
+                    prepared.append(symm_slot)
+                    continue
```
fanshiqing, line 2082: this function `_prepare_wgrad_reduce_scatter_inputs` introduced by Jiangfei which supports the partial CG overlapping makes the original implementation much more complex and also more python overhead when CG is not introduced:

when getattr(weight, "_gtp_graph_wgrad_ring_slot", None) is None:
   if self.pad_lengh = 0: do nothing
   if self.pad_length > 0: one line for-loop as below.
```
if self.pad_length > 0:
    wgrads = [torch.nn.functional.pad(w, (0, 0, 0, self.pad_length)) for w in wgrads]
```

I will submit a following MR to reduce the GTP implementation related python overhead.
for the current MR: please keep try to keep 2057-2073 and seperate function to make it clear easy to maintain.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -307,6 +311,15 @@ class GTPWeightState(Enum):
 _GTP_GROUPED_BUF_PARITY_COUNTER: Dict[tuple, int] = {}
 
 
+def _alloc_symmetric_wgrad_buffer(weight, dtype, device) -> torch.Tensor:
+    """Padded send buffer from the registered LIFO, pad tail zeroed (the RS sends the
```
fanshiqing: Can you help explain a bit why it's LIFO here? what's the benefit or any other design considerations?

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,208 @@
...
+ProcessGroupNCCL hook window-registers every allocation made inside
+``gtp_symm_pool_ctx(group)``, which lets NCCL run its symmetric / NVLS kernels on
+those buffers.
+
+Two parts:
+  - Pool lifecycle: create, register, query, and tear down the per-group pools,
+    plus the allocation context ``gtp_symm_pool_ctx``.
+  - ``symmetric_wgrad_pool`` (a ``RegisteredLIFOPool``): recycled, window-registered
+    send buffers for the wgrad reduce-scatter.
+
```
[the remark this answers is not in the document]
prajwal1210: The comment was wrong. The HOOK can be set later and is so by the nccl allocator. NVLS ENABLE is not a requirement from experiments.

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,208 @@
...
+        torch.cuda.synchronize()
+    # Deregister while the recycled send buffers are still alive. Their memory keeps
+    # the pool non-empty, so deregister_mem_pool (which skips empty pools) always runs.
+    for name in sorted(_registered):
+        nccl_allocator.deregister_mem_pool(_pools[name], _registered[name])
+    # Only now drop the buffers and the pools; the windows are gone, so the memory
+    # is safe to release.
+    symmetric_wgrad_pool.clear()
+    _registered.clear()
+    _pools.clear()
```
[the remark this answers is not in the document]
prajwal1210: This is not valid. We don't want to clear memory that is owned by other GTP components. This function still deregisters the pool so all ring buffers will also be deregistered. It should only clear the buffers it owns.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1734,13 +1759,22 @@ def batched_all_gather_and_prefetch(self, **kwargs):
...
+        if is_gtp_symm_pool_registered(self.group):
+            # Lifecycle invariant: get_wgrad_tensor -> GEMM -> _prepare consumes the slot -> RS.
+            if self._wgrad_symm_slot is not None:
+                raise RuntimeError(
+                    "[GTP] get_wgrad_tensor called again before the previous symmetric "
+                    f"wgrad buffer was consumed ({self._debug_name})."
+                )
+            buf = _alloc_symmetric_wgrad_buffer(self, self.main_grad.dtype, self.device)
+            self._wgrad_symm_slot = buf
+            return buf[: self._unsharded_shape[0]]
```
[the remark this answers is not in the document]
prajwal1210: Resolved by updating the comment. The other suggestions are wrong and can lead to silent corruptions. It is better to raise an error as it is not expected to get a wgrad but not call reduce scatter. If reduce scatter is trivially on a group or delayed, it should be handled in that case separately that the buffer is freed.

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,213 @@
...
+    recycled send buffers. Allocations owned by others (e.g. graph wgrad ring slots)
+    are not freed here. Call on all ranks before teardown; no-op if never registered."""
+    # Wait for all GPU work to finish first: a kernel or collective still reading
+    # pool memory would fault once the windows go away.
+    if torch.cuda.is_available() and torch.cuda.is_initialized():
+        torch.cuda.synchronize()
+    # Deregister while the recycled send buffers are still alive. Their memory keeps
+    # the pool non-empty, so deregister_mem_pool (which skips empty pools) always runs.
+    for name in sorted(_registered):
+        nccl_allocator.deregister_mem_pool(_pools[name], _registered[name])
```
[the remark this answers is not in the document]
prajwal1210: This is not critical. Each rank in the gtp group would have similar symmetric pools.  Pool registration is a collective too and everyone in the PG has to do it.

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,213 @@
...
+    # Only now drop the buffers and the pools; the windows are gone, so the memory
+    # is safe to release.
+    symmetric_wgrad_pool.clear()
+    _registered.clear()
+    _pools.clear()
```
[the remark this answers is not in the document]
prajwal1210: GTP symmetric should not clear GRAPH_WGRAD_RINGS but it should be cleared in the reset_gtp_state. Will do that.

megatron/training/training.py
```diff
@@ -2185,6 +2196,11 @@ def _build_model_wrapper(wrap_with_ddp: bool):
             ),
         )
 
+        if getattr(args, 'gtp_remat_nccl_ub', False):
+            register_gtp_symm_pool(resolve_gtp_remat_group(pg_collection, is_expert=False))
+        if getattr(args, 'gtp_expert_remat_nccl_ub', False):
+            register_gtp_symm_pool(resolve_gtp_remat_group(pg_collection, is_expert=True))
```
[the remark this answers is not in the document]
prajwal1210: This is not valid. This cannot literally happen:
"if the dense group has size > 1 while the dense axis is inactive, the pool is registered "
Also the register function checks whether the size > 1 or not.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1839,6 +1865,11 @@ def _release_comm_scratch(self, attrs=("_wgrad_input_bufs", "_rs_a2a_bufs")):
             if not _chain_is_graphed(self.chain_id):
                 for buf in bufs:
                     _wgrad_pool_put(buf)
+            for buf in bufs:
+                # Return symm pool buffers (tag-gated no-op for plain and ring buffers).
+                # Unconditional on chain kind: the pool's captured alloc/free pops replay
+                # at stable addresses.
+                symmetric_wgrad_pool.free(buf)
```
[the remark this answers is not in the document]
prajwal1210: Will add a comment saying why the ordering gaurantee is not needed.

megatron/core/tensor_parallel/gtp_symmetric_memory.py
```diff
@@ -0,0 +1,213 @@
...
+
+def _get_gtp_symm_pool(group: dist.ProcessGroup) -> torch.cuda.MemPool:
+    """Return the per-group ``ncclMemAlloc``-backed MemPool, creating it once."""
+    name = group.group_name
+    pool = _pools.get(name)
+    if pool is None:
+        nccl_allocator.init()
+        pool = nccl_allocator.create_nccl_mem_pool(symmetric=True)
+        _pools[name] = pool
+    return pool
```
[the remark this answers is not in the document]
prajwal1210: This is existing code and has not been changed in the PR. We call nccl_allocator in Megatron itself that sets these flags.

fanshiqing:
@prajwal1210 Please don't forget to run all GTP related UTs offline (which are not covered by online UTs yet due to higher TE version) to make all existing GTP UTs are green:

```
torchrun --nproc-per-node 4 -m pytest tests/unit_tests/generalized_tensor_parallel/ -v
```

prajwal1210:
> @prajwal1210 Please don't forget to run all GTP related UTs offline (which are not covered by online UTs yet due to higher TE version) to make all existing GTP UTs are green:
> 
> ```
> torchrun --nproc-per-node 4 -m pytest tests/unit_tests/generalized_tensor_parallel/ -v
> ```

Yeah I have run all the UTs and they pass.

fanshiqing (approved):

jiemingz:
In the interest of time I think this is fine to merge but I think we should seriously think about cleaning up GTP, this PR essentially tacks on specialized path for symmetric buffers and its getting pretty disorganized.

jiemingz (approved):

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -1997,22 +2041,48 @@ def _reduce_scatter(self, wgrads, async_op, nvtx_label=None):
...
+                    if wgrad.data_ptr() != symm_slot.data_ptr():
+                        symm_slot[: weight._unsharded_shape[0]].copy_(wgrad)
+                        if not _chain_is_graphed(self.chain_id):
+                            _wgrad_pool_put(wgrad)
+
+                    send_bufs.append(symm_slot)
+                    release_bufs.append(symm_slot)
+                    continue
+
+                release_bufs.append(wgrad)
```
JF-D, line 2084: why do we release the wgrad befoe padding, instead of the padded wgard?

prajwal1210: This line is when the wgrad is not in the symmetric registration path. In that case, you took the buffer from _wgrad_input_bufs and that's what needs to be released. The padding creates a copy anyway using torch.nn.functional.pad later.

JF-D (approved):

ericharper (approved):

maanug-nv (approved):

PR #6546, opened 2026-08-14
Enable MXFP8 2D weight quantization

### LINKED ISSUE #6363: [feat] GTP+2D Mxfp8
New 2D quant for mxfp8 support in TE: https://github.com/NVIDIA/TransformerEngine/pull/2634
- 2D quant is for only for weight tensors: `W, W^T = 2D quantized in blk 32x32`
- Activations are quantized using 1D scales (1x32 elements): `X, X^T = 1D quantized in blk 1x32`

dingqingy-nv:
## What does this PR do?

Enable Transformer Engine's 2D MXFP8 weight quantization from Megatron Core while keeping 1D MXFP8 as the default.

- Add the `mxfp8_2d_quantization` transformer configuration and generated `--mxfp8-2d-quantization` CLI option.
- Forward the option to TE's `MXFP8BlockScaling(enable_2d_quantization=True)` recipe in global, per-module, fused grouped-MLP, and shared-expert paths.
- Reject incompatible recipes and report a clear error when the installed TE does not explicitly support the 2D option.
- Reject `mxfp8_2d_quantization` together with `moe_single_grouped_weight`, which is not covered by TE's 2D implementation.

## Why is this needed?

Transformer Engine added 2D MXFP8 weight quantization in NVIDIA/TransformerEngine#2634, but Megatron Core currently has no configuration path to select it.

The default is unchanged: without `--mxfp8-2d-quantization`, MXFP8 continues using 1D quantization.

## Validation

- Focused GB300 validation: 14 tests passed on four ranks, covering:
  - recipe and configuration propagation;
  - strict TE capability detection;
  - 1D and 2D native-MXFP8 GTP forward/backward;
  - aligned and unaligned GTP weight shapes.
- Full offline GTP unit-test suite: 126 tests passed on four GB300 GPUs, including existing native-MXFP8 parameter-gather and distributed-checkpoint coverage.
- Transformer Engine feature validation: 135 feature tests, 6 grouped-tensor tests, and 8 GroupedLinear cases passed on GB300.
- Two-node, 11-layer real-data proxy:
  - 1D and 2D native-MXFP8 + GTP + AdamW showed matching initial convergence, memory, and performance.
  - A 2D run saved at step 250 and resumed to step 500 consistently with the uninterrupted run. This checkpoint test used the separate GTP/GDP checkpoint fix from NVIDIA/Megatron-LM#6503.
- A 76-layer TP1 evaluation on 256 GB300 GPUs showed matching performance and memory between 1D and 2D MXFP8.

## Scope notes

- Requires a Transformer Engine build containing NVIDIA/TransformerEngine#2634.
- The separate GTP/GDP checkpoint-layout fix is provided by NVIDIA/Megatron-LM#6503 and is not part of this PR.

## Related issues

Fix #6363

### DIFF
--- a/megatron/core/fp8_utils.py
+++ b/megatron/core/fp8_utils.py
@@ -749,7 +749,10 @@ def is_mxfp8_output_proj_active(config) -> bool:
 
 if HAVE_TE:
     from megatron.core import parallel_state
-    from megatron.core.extensions.transformer_engine import TEDelayedScaling
+    from megatron.core.extensions.transformer_engine import (
+        TEDelayedScaling,
+        get_mxfp8_block_scaling_recipe,
+    )
 
     def get_fp8_recipe(config: TransformerConfig):
         """Return fp8 recipe.
@@ -785,8 +788,10 @@ if HAVE_TE:
                     fp8_format=fp8_format
                 )
             elif config.fp8_recipe == Fp8Recipe.mxfp8:
-                fp8_recipe = transformer_engine.common.recipe.MXFP8BlockScaling(
-                    fp8_format=fp8_format, fp8_dpa=config.fp8_dot_product_attention
+                fp8_recipe = get_mxfp8_block_scaling_recipe(
+                    mxfp8_2d_quantization=config.mxfp8_2d_quantization,
+                    fp8_format=fp8_format,
+                    fp8_dpa=config.fp8_dot_product_attention,
                 )
             elif config.fp8_recipe == Fp8Recipe.custom:
                 assert config.fp8_quantizer_factory is not None
--- a/megatron/core/fp8_utils.py
+++ b/megatron/core/fp8_utils.py
@@ -347,11 +347,9 @@ def _get_custom_recipe(quantizer_factory_python_path: str) -> Union[Fp8Recipe, F
     try:
         custom_recipe = transformer_engine.common.recipe.CustomRecipe(qfactory=quantizer_factory)
     except AttributeError:
-        raise ValueError(
-            """CustomRecipe recipe is not available in this version of 
+        raise ValueError("""CustomRecipe recipe is not available in this version of 
             Transformer Engine. Please make sure you are using TE version 
-            >= 2.9.0.dev0."""
-        )
+            >= 2.9.0.dev0.""")
     return custom_recipe
 
 
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -1483,6 +1483,10 @@ class TransformerConfig(ModelParallelConfig):
                     "mxfp8_2d_quantization requires fp8_recipe='mxfp8', "
                     f"got '{self.fp8_recipe}'."
                 )
+            if self.moe_single_grouped_weight:
+                raise ValueError(
+                    "mxfp8_2d_quantization does not support moe_single_grouped_weight."
+                )
 
         if self.fp8_output_proj:
             if not self.fp8:
--- a/megatron/core/transformer/transformer_config.py
+++ b/megatron/core/transformer/transformer_config.py
@@ -585,6 +585,10 @@ class TransformerConfig(ModelParallelConfig):
     uses delayed scaling recipe, 3) 'mxfp8' for Blackwell architecture only,
     4) 'blockwise' for blockwise scaling recipe, 5) 'custom' for custom quantization recipe."""
 
+    mxfp8_2d_quantization: bool = False
+    """If True, use 2D block scaling for forward-pass linear and grouped-linear weights.
+    Requires fp8_recipe='mxfp8' and a Transformer Engine build with 2D MXFP8 support."""
+
     fp8_param: bool = False
     """If set, keep the parameters in fp8 precision to save memory. This option must be used
     together with fp8 mode (i.e., TransformerConfig.fp8 is not None). Note that not all parameters
@@ -1470,6 +1474,15 @@ class TransformerConfig(ModelParallelConfig):
 
         if self.fp8_param and not self.fp8:
             raise ValueError("fp8_param must be used together with fp8 mode.")
+
+        if self.mxfp8_2d_quantization:
+            if not self.fp8:
+                raise ValueError("mxfp8_2d_quantization must be used together with fp8 mode.")
+            if self.fp8_recipe != Fp8Recipe.mxfp8:
+                raise ValueError(
+                    "mxfp8_2d_quantization requires fp8_recipe='mxfp8', "
+                    f"got '{self.fp8_recipe}'."
+                )
 
         if self.fp8_output_proj:
             if not self.fp8:
--- a/tests/unit_tests/generalized_tensor_parallel/test_gtp_basics.py
+++ b/tests/unit_tests/generalized_tensor_parallel/test_gtp_basics.py
@@ -813,9 +813,7 @@ def _worker_mxfp8_linear(rank, world_size, port, mxfp8_2d_quantization):
     batch, in_f, out_f = 32, 64, 128  # out_f % (16*world_size)==0 → no padding
     dtype = torch.bfloat16
     gtp_remat_group = dist.new_group(list(range(world_size)))
-    recipe = get_mxfp8_block_scaling_recipe(
-        mxfp8_2d_quantization=mxfp8_2d_quantization
-    )
+    recipe = get_mxfp8_block_scaling_recipe(mxfp8_2d_quantization=mxfp8_2d_quantization)
     layer = _make_native_fp8_gtp_linear(in_f, out_f, gtp_remat_group, dtype, recipe)
 
     # The weight IS the native FP8 shard: a QuantizedTensor with the GTP surface attached.
@@ -864,9 +862,7 @@ def _worker_mxfp8_linear_unaligned(rank, world_size, port, mxfp8_2d_quantization
     batch = 32
     dtype = torch.bfloat16
     gtp_remat_group = dist.new_group(list(range(world_size)))
-    recipe = get_mxfp8_block_scaling_recipe(
-        mxfp8_2d_quantization=mxfp8_2d_quantization
-    )
+    recipe = get_mxfp8_block_scaling_recipe(mxfp8_2d_quantization=mxfp8_2d_quantization)
     layer = _make_native_fp8_gtp_linear(in_f, out_f, gtp_remat_group, dtype, recipe)
     assert layer.weight.pad_length == 8, f"expected pad 8, got {layer.weight.pad_length}"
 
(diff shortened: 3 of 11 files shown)


megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -3111,6 +3111,13 @@ def copy_group_params(model_groups, shard_main_groups):
                     if self._is_distopt_quantized_param(model_param):
                         if self._is_grouped_quantized_tensor(model_param):
                             dequantized_model_param = model_param.float()
+                        elif getattr(model_param, "is_gtp_weight_remat", False):
```
zhongbozhu: This condition doesn't seem to be a nice fit as the `elif` after `_is_grouped_quantized_tensor`, can you explain more here? 

I believe @fanshiqing is also working on GTP + single grouped weight, so we might need to make the code logic clearer.

megatron/core/optimizer/distrib_optimizer.py
```diff
@@ -3111,6 +3111,13 @@ def copy_group_params(model_groups, shard_main_groups):
...
+                            # TE dispatches dequantization on the exact quantized tensor class.
+                            # Native-FP8 GTP params use a dynamic subclass, so temporarily expose
+                            # their TE base class at this model-only checkpoint boundary.
+                            from ..tensor_parallel.gtp_api import dequantize_gtp_native_fp8
+
+                            dequantized_model_param = dequantize_gtp_native_fp8(model_param)
```
zhongbozhu: When will this dequantize happen? Ideally when using fp8 param, even if the TE parameter is in bf16 which is then quantized to fp8, the bf16 copy is still gonna exist, we should ideally dequantize from that not form fp8.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -680,6 +680,24 @@ def attach_gtp_to_presharded_module(module, gtp_remat_group, pad_length, is_grou
 # Cache of dynamic ``GTP_<Fp8TensorClass>`` subclasses, keyed by the FP8 base class.
 _GTP_NATIVE_FP8_SUBCLASSES: Dict[type, type] = {}
 
+
+def _gtp_native_fp8_detach(self):
```
zhongbozhu: Do we really need this code? seems very hacky :)

megatron/core/transformer/transformer_config.py
```diff
@@ -1471,6 +1475,15 @@ def __post_init__(self):
         if self.fp8_param and not self.fp8:
             raise ValueError("fp8_param must be used together with fp8 mode.")
 
+        if self.mxfp8_2d_quantization:
```
zhongbozhu, line 1478: we need more asserts here: for example moe single weight is not supported

megatron/core/fp8_utils.py
```diff
@@ -263,7 +263,12 @@ def copy_tensors_to_quantized_params(params: List[torch.Tensor], srcs: List[torc
     # Equivalent to dst.copy_(src), but entered directly instead of via the aten::copy_ op,
     # QuantizedTensor.__torch_dispatch__ (type and usage checks) and dst.quantize_(src).
     for src, quantizer, dst in zip(srcs_to_cast, quantizers, dsts_to_cast):
-        quantizer.update_quantized(src, dst)
+        update_dst = dst
+        if HAVE_TE_MXFP8TENSOR and isinstance(dst, MXFP8Tensor) and type(dst) is not MXFP8Tensor:
```
zhongbozhu: why is this condition triggered in the first place? feels like we are patching for another issue here

megatron/core/optimizer/distrib_optimizer.py
dingqingy-nv: This only happens for model only load and in the [lossy fallback path](https://github.com/dingqingy-nv/Megatron-LM/blob/dingqingy/mxfp8-2d-mcore/megatron/training/checkpointing.py#L2951-L2952) where `state_dict` is not provided. e.g. with `--load`, `--no-load-optim`, and **without** `--load-main-params-from-ckpt`

is this lossy fallback a solid use case? if not we should probably consider a separate PR to avoid ever entering this path.

this part of the code will not be executed in other use case:
- full training resume (with `--load`), here fp32 master weight is obtained directly from optimizer ckpt. (e.g. `optimizer.load_state_dict(state_dict["optimizer"])`)
- model only load from ckpt tensor (`--no-load-optim`, `--load-main-params-from-ckpt`), [fp32 master weight is directly obtained from state_dict](https://github.com/dingqingy-nv/Megatron-LM/blob/dingqingy/mxfp8-2d-mcore/megatron/core/optimizer/distrib_optimizer.py#L3107-L3109)

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -680,6 +680,22 @@ def attach_gtp_to_presharded_module(module, gtp_remat_group, pad_length, is_grou
...
+    """Detach a native-FP8 GTP weight without dropping its dynamic subclass.
+
+    Some TE quantized tensors construct their concrete base class in ``detach``. PyTorch requires
+    a tensor subclass's ``detach`` result to have the exact same type when wrapping it in a
+    ``Parameter`` (for example during ``Module._apply``). Reconstruct with the dynamic GTP class
+    and copy the GTP-only runtime attributes that TE's ``make_like`` does not know about.
+    """
+    detached = type(self).make_like(self)
+    for name, value in self.__dict__.items():
+        if name not in detached.__dict__:
```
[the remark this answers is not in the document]
dingqingy-nv: fixed

tests/unit_tests/generalized_tensor_parallel/test_gtp_basics.py
```diff
@@ -817,6 +818,24 @@ def _worker_mxfp8_linear(rank, world_size, port):
...
+    assert type(rewrapped_w) is type(w), "Parameter rewrap must preserve the GTP MXFP8 type"
+
+    # DDP's native FP8 parameter-gather path requantizes BF16 bucket slices into the dynamic GTP
+    # MXFP8 parameter after every all-gather. TE's C++ update accepts only its exact base tensor
+    # type, so MCore must present a storage-sharing base view at that boundary.
+    gathered_w = torch.zeros(w.shape, dtype=dtype, device="cuda")
+    copy_tensors_to_quantized_params([w], [gathered_w])
+    for data in (w._rowwise_data, w._columnwise_data):
+        if data is not None:
+            assert torch.count_nonzero(data) == 0, "requantization did not update shared storage"
```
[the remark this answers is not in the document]
dingqingy-nv: fixed

tests/unit_tests/generalized_tensor_parallel/test_gtp_basics.py
```diff
@@ -807,7 +808,7 @@ def _worker_mxfp8_linear(rank, world_size, port):
     batch, in_f, out_f = 32, 64, 128  # out_f % (16*world_size)==0 → no padding
     dtype = torch.bfloat16
     gtp_remat_group = dist.new_group(list(range(world_size)))
-    recipe = MXFP8BlockScaling()
+    recipe = MXFP8BlockScaling(enable_2d_quantization=True)
```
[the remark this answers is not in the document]
dingqingy-nv: fixed

tests/unit_tests/test_fp8_param.py
```diff
@@ -546,6 +547,169 @@ def test_mxfp8(self, tp_size, dp_overlap):
...
+            source = torch.randn(shape, dtype=torch.bfloat16, device="cuda", generator=generator)
+            # Make neighboring block scales observably different.
+            source[0, 0] = 64.0
+            source[block_size, block_size] = -16.0
+
+            quantizer = MXFP8Quantizer(
+                fp8_dtype=tex.DType.kFloat8E4M3,
+                rowwise=True,
+                columnwise=True,
+                with_2d_quantization=True,
```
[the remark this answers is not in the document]
dingqingy-nv: fixed

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
fanshiqing: if GTP can works well 1D mxfp8, then why the baseline mechanism cannot be compatible with 2d mxfp8? what's the root cause for introducing this extra new func?

fanshiqing:
`GTP+DCP+2d mxfp8`: has the ckpt saving/loading also been well tested too?

fanshiqing:
Also, pls make sure the existing all GTP related UTs are not broken: currently GTP relies on high `TE>=2.19`, which is higher than the online CI/CD's TE version. Thus these GTP UTs are just been skipped. We have to trigger these UTs offline manually before the online CI/CD upgrade the TE version.

How to run GTP related UTs locally ([doc](https://github.com/NVIDIA/Megatron-LM/blob/main/docs/api-guide/core/generalized_tensor_parallel.md#4-testing)):
```
torchrun --nproc-per-node 4 -m pytest tests/unit_tests/generalized_tensor_parallel/ -v
```

megatron/core/optimizer/distrib_optimizer.py
fanshiqing: +1

One of the bottom lines is to make dist-opt be GTP agnostic.
Need to figure out a better solution if possible.

megatron/core/tensor_parallel/generalized_tensor_parallelism.py
```diff
@@ -720,6 +738,9 @@ def _gtp_native_fp8_subclass(base_cls: type) -> type:
         # Share the (class-level) prefetch chain-state dicts with GTPShardedParam.
         ns["_chain_state"] = GTPShardedParam._chain_state
         ns["_recompute_chain_state"] = GTPShardedParam._recompute_chain_state
+        # TE quantized tensor detach implementations may hardcode their base class. Preserve the
+        # exact dynamic type required when PyTorch rewraps the tensor as a Parameter.
+        ns["detach"] = _gtp_native_fp8_detach
```
zhongbozhu: same comment above ^

dingqingy-nv:
@fanshiqing 
chpt loading and saving is validated with matching loss between
- train from scratch for 500 iterations
- train from scratch for 250 iterations and save ckpt for one run, and resume ckpt for another 250 iterations in another run.

GTP unit tests are validated too offline.

both are commented in PR descriptions too.

zhongbozhu (approved):
LGTM, please share the TE PR dependency in this PR too @dingqingy-nv

deepakn94 (approved):

yaox12 (approved):

kunlunl (approved):

PR #6552, opened 2026-08-15
Add Tensor Metrics system for training

philipcmonk:
This PR introduces a general training metrics subsystem.  This is a reimplementation and generalization of #5492.

There are dozens of metrics that we want to keep track of during training, and currently we have to thread each one separately through the model.  This is bug-prone and delicate, and it multiplies the complexity of making other changes that affect the sharding of tensors.

Here, we add a system that decouples obtaining a tensor from the metric computation itself.  This lets users easily add metrics without changes to MCore.

Check the committed README at [megatron/training/tensor_metrics.md](https://github.com/philipcmonk/Megatron-LM/blob/pmonk/tensor-metrics/megatron/training/tensor_metrics/README.md) for an overview of the system and how to add metrics.

This is a large PR, I recommend it be reviewed in pieces (in whatever order is natural to you):

- `megatron/training/tensor_metrics/{core,observer}.py`: The core system, with Observer and Executor implementations plus the TensorMetric interface.
- `megatron/training/tensor_metrics/{definitions,examples,router_metrics}.py`: Initial set of implemented metrics.
- `megatron/training/training.py`: How the observer is called from the training loop.
- `megatron/core/parameter_names.py` and `megatron/training/tensor_metrics/optimizer_sources.py`: Parameter/wgrad discovery across optimizers, canonical naming, manifests, and placement metadata.
- Remainder of `megatron/core/*`: How activation tensors are observed.
- `tests/*`: Unit tests.

This is a draft while the design is reviewed.  The implementation is generally complete and I've started using it to debug issues, but details are not fully stable yet.


janEbert:
I think this needs to be split up, just based on the 8.9k LOC difference. :d
Is it possible in any way to abstract different features included here into different PRs?

philipcmonk:
I think the natural boundaries are probably these:

- All MCore changes (new parameter names module, new tensor observation module, hooks into various places to observe tensors): ~500 lines + 500 lines tests
- TensorMetric framework: 900 lines + 500 lines tests
- Training integration: 1400 lines + 1500 lines tests
- Metric definitions: 2300 lines + 1200 lines tests

We could make those as four separate PRs or combine some of the training PRs since they'll probably have the same reviewers.

philipcmonk:
This is now split up into
- #6647
- #6648 
- #6649 
- #6650

PR #6566, opened 2026-08-16
Add hidden state mixing to multi-token prediction

YanivDorGalron:
Add opt-in Hidden State Mixing (HSM) for Multi-Token Prediction layers.

HSM uniformly selects, per token, from the main-model hidden state and outputs accumulated from earlier MTP depths. It is enabled with `--mtp-hsm` and only runs during training when at least two MTP layers are configured.

## Implementation

- Add the default-off `TransformerConfig.mtp_hsm` option and generated `--mtp-hsm` argument.
- Roll older history entries before each MTP depth so candidates correspond to the same target-token offset.
- Respect packed-sequence boundaries during history alignment.
- Preserve activation dtype and gradient flow using indexed selection.
- Use MCore's data-parallel RNG tracker for reproducible sampling.
- Generate a full logical-sequence sampling mask and give SP/CP token owners disjoint slices, while replicated TP ranks retain matching selections.
- Fall back to the newest hidden state when a sampled rolled candidate contains the zero sentinel produced at an invalid local boundary.
- Keep evaluation and existing configurations unchanged.

## Compatibility

HSM is disabled by default. Existing training and inference behavior is unchanged unless `--mtp-hsm` is enabled.

The feature requires at least two MTP layers; otherwise it is disabled during argument validation.


YanivDorGalron:
1. Regarding “roll_tensor is sequence-parallel unaware, but HSM rolls SP-sharded hidden states” - this is a known behaviour, while roll_tensor is only local aware (does not take other sequence chunks that are not on current rank) we concluded it only affect 1 hidden state at each rank. In order to maintain integrity of selected hidden states we later replace invalid boundray hidden states that were replaced with zeros with the last and available hiddenstate
2. Regarding “The DP RNG tracker gives every DP rank the same mask” this is intentional - in order to be able to perform activation checkpointing we want to seed the random sampling. Since each rank get the same seed we decided to sample indices for the entire sequence and slice only the relevant parts in the current rank.
3. Regarding “The all-zero sentinel is ambiguous.. padded batches or masked positions could have zero vectors” - if thats the case then also their last hidden state will be zero and no change will happen


Add a commit for preventing calculating sum when not needed.

megatron/training/arguments.py
```diff
@@ -859,6 +859,11 @@ def validate_args(args, defaults={}):
             + f"The supported position embedding types are rope and none."
         )
 
+    if getattr(args, 'mtp_hsm', False) and not (
+        args.mtp_num_layers and args.mtp_num_layers >= 2
+    ):
+        args.mtp_hsm = False
```
fanshiqing, line 894: Pls at least at explicit warning msg here instead of setting the args silently

megatron/core/transformer/transformer_config.py
```diff
@@ -87,6 +87,9 @@ class TransformerConfig(ModelParallelConfig):
     This prevents MTP loss gradients from flowing back to the main model,
     only training the MTP heads themselves."""
 
+    mtp_hsm: bool = False
+    """Enable uniform per-token Hidden State Mixing for MTP layers."""
```
fanshiqing: plase also add the 'only runs during training when at least two MTP layers are configured.' along side with the existing comments

megatron/core/transformer/multi_token_prediction.py
```diff
@@ -1855,12 +1915,50 @@ def forward(
         if self.config.mtp_detach_heads:
             hidden_states = hidden_states.detach()
 
+        hsm_enabled = self.config.mtp_hsm and self.training and self.config.mtp_num_layers >= 2
```
fanshiqing: do we really need to check `self.config.mtp_num_layers >= 2` here again?
in `trainsformer_config.py` has already the guard for `mtp_num_layers` which will flip the user-passed-in `mtp_hsm` if `mtp_num_layers < 2`

megatron/training/arguments.py
deepakn94: Use `warn_on_rank0` for this.

cuichenx, line 894: can we add this guard to transformer_config instead or in addition. some downstream workloads don't use arguments.py but transformer_config is a common path for everyone

YanivDorGalron:
I’ve updated the HSM implementation based on the review feedback.

### Implementation

HSM remains disabled by default, runs only during training, and requires at least two MTP layers.

It uses a dedicated checkpointable RNG stream:

- Replicated TP ranks make the same selections.
- SP/CP owners select different sequence regions.
- DP replicas receive different randomness.
- Activation recomputation replays the same selections.

Before every MTP depth, all older hidden states are rolled once while the newest state remains unchanged. This aligns every candidate with the current target token. The stacked rolled tensor is reused directly during mixing, and `return_sum=False` avoids the unused roll reduction.

When a roll has no valid local continuation—at an SP seam, document end, or padded position—the location is zero-filled and mixing falls back to the newest hidden state.

### Local packed-sequence construction

`PackedSeqParams` arrives in global sequence coordinates, but HSM may be rolling only the local CP/SP shard. `_packed_seq_params_for_local_hsm_roll` converts those global boundaries into that local coordinate system.

For ordinary packed input:

1. Document boundaries are scaled into CP-local coordinates.
2. For zigzag CP ownership, a midpoint boundary is added between the two locally adjacent chunks when they are not globally adjacent.
3. The boundaries are clipped to the current SP window.
4. The SP window offset is subtracted, producing shard-local boundaries.

The final CP rank is treated specially because its two zigzag chunks are globally adjacent, so adding a midpoint there would incorrectly introduce a boundary.

For genuine padding, `cu_seqlens_q_padded` describes physical storage and CP ownership, while `cu_seqlens_q` describes the valid tokens. For each document, we:

1. Split its padded length into the `2 * cp_size` zigzag chunks.
2. Identify the two chunks owned by the current CP rank.
3. Record both the valid-token end and physical chunk end.
4. Concatenate those boundaries in local ownership order.
5. Clip them to the current SP shard.

This makes both the valid document end and the padded tail explicit roll boundaries.

The genuine-padding check intentionally uses tensor identity:

```python
cu_seqlens_q_padded is not cu_seqlens_q
```

Production sometimes passes the same tensor through both fields. That represents padded metadata structurally, but not different padded and unpadded layouts. Using identity also avoids a device synchronization from comparing tensor values.

After translation, the padded fields, `total_tokens`, and `seq_idx` are cleared because they still refer to global coordinates.

### Why `cp_group` is omitted

When SP is active—or when genuine padding has already been translated—the tensor and its boundaries are fully local.

Passing `cp_group` to `roll_tensor` at that point would apply CP interpretation a second time. Its CP path assumes that the input contains the standard pair of complete zigzag chunks and performs neighbor exchange accordingly. That assumption is no longer valid after SP sharding or local padded-boundary construction and can insert a token from an unrelated sequence position.

We therefore pass:

```python
cp_group=None
```

The local packed roll respects the translated boundaries, zeroes locations without a local continuation, and lets HSM fall back to the newest state. This is intentional: we prefer a safe local fallback over mixing a plausible but incorrectly aligned historical state.

TP, CP, and DP groups are otherwise provided through `ProcessGroupCollection`; no new global process-group reads were added.

### Tests

The alignment tests encode the global token position into every hidden-state value. Stub MTP layers increment the value, allowing each rolled and mixed state to be matched to its exact expected target token.

Every parallel layout runs with both random selection and forced-oldest selection, ensuring the deepest history and boundary paths execute. Coverage includes:

- TP, SP, CP=1/2/4, and combined TP+SP+CP
- Packed and unpacked input
- Production-style padded metadata
- Genuine padding with different valid and physical boundaries
- SP seams, CP zigzag seams, document ends, and padded tails
- Batch size 2, sequence length 8192, and backward gradients
- RNG behavior across TP, SP/CP, DP, and activation recomputation
- Real GPT/Transformer Engine forward and backward with recomputation on/off

The distributed test utility now uses global `RANK` for distributed initialization and `LOCAL_RANK` for CUDA device selection, enabling the eight-rank test across two four-GPU nodes.

### Results

```text
Focused HSM/rolling: 100 passed
CI-dev MTP file:     140 passed, 19 deselected
hsm_subset=0
full_file=0
```

The 19 deselected cases are the existing `flaky_in_dev` tests, matching the standard dev CI selection.

The remaining intentional limitations are that SP seams fall back to the newest state instead of fetching history across ranks, and history rolling remains \(O(D^2)\) in MTP depth.

cuichenx (approved):

fanshiqing (approved):

jaredcasper (approved):

jaredcasper:
/ok to test [9b9ded1](https://github.com/NVIDIA/Megatron-LM/pull/6566/commits/9b9ded1d5c012ef020c91f0e88a5dfa8aed4f292)

deepakn94:
/ok to test [1b3ae3c](https://github.com/NVIDIA/Megatron-LM/pull/6566/commits/1b3ae3cd11d406dc08ccee523a1c8035f99e5ca3)

